PR #19820, opened 2025-08-08
[ty] add nested bindings to the semantic index

oconnor663:
This PR replaces #19703. It has two main components:

1. Add free variable tracking to the `SemanticIndexBuilder`, and resolve free variables in `pop_scope`. This lets us determine which variables resolve where in the semantic index building pass, without requiring a second pass and without requiring us to defer function bodies. Previously we were doing some of this in `infer.rs` as a poor man's second pass, particularly `infer_nonlocal`, which used to flag invalid `nonlocal` statements. That's now a semantic syntax error in the builder, and `infer_nonlocal` is deleted. In my first pass at #19703 I was recording several new types of metadata in the index itself, including all nonlocal resolutions and nested references. But based on feedback from Micha and Carl, I've pared that down substantially to avoid wasting space. In this PR the only new metadata in the semantic index (specifically in the symbol tables) is `nested_scopes_with_bindings`, which should be empty for most scopes. (The new free variables collection lives in the builder and isn't persisted in the index.)
2. Now that `nested_scopes_with_bindings` is available, take advantage of it to consider all nested bindings when inferring the "public" type of a variable (i.e. its type as viewed from other scopes or other modules). This is done in `place_by_id`. Previously `infer_place_load` (which calls `place` -> `place_by_id`) was computing a union of all the local inferred types it encountered when walking enclosing scopes, but that was missing bindings in sibling/cousin scopes. Now we see those bindings, and the union builder in `infer_place_load` is deleted.

Examples:

```py
def _():
    x = 1
    def _():
        nonlocal x
        x = 2
        def _():
            nonlocal x
            x = 3
    def _():
        # Previously: Literal[1]
        # Now: Literal[1, 2, 3, 4]
        reveal_type(x)
        def _():
            nonlocal x
            x = 4

y = 1
def f():
    global y
    y = 2
def g():
    # Previously: Unknown | Literal[1]
    # Now: Unknown | Literal[1, 2]
    reveal_type(y)
```

Now that we can see nested bindings of globals, I'm going to put up a [follow-up PR on top of this one](https://github.com/astral-sh/ruff/pull/19821) that deletes `infer_global` and allows `global` statements that don't match an explicit symbol in the module scope. I don't know if we'll want that change, so I'm keeping it out of this PR.

[...]

PR #19821, opened 2025-08-08
[ty] now that inference sees nested bindings, allow unresolved globals

oconnor663:
This is an optional follow-up PR on top of https://github.com/astral-sh/ruff/pull/19820. Now that we track all bindings from nested scopes, it could be practical to allow `global` variables without an explicit definition in the module scope. For example:

```py
def f():
    global x
    x = 1

def g():
    print(x)
```

Before:

```
warning[unresolved-global]: Invalid global declaration of `x`
 --> test.py:2:12
  |
1 | def f():
2 |     global x
  |            ^ `x` has no declarations or bindings in the global scope
3 |     x = 1
  |
info: This limits ty's ability to make accurate inferences about the boundness and types of global-scope symbols
info: Consider adding a declaration to the global scope, e.g. `x: int`
info: rule `unresolved-global` is enabled by default

error[unresolved-reference]: Name `x` used when not defined
 --> test.py:6:11
  |
5 | def g():
6 |     print(x)
  |           ^
  |
info: rule `unresolved-reference` is enabled by default
```

After:

```
All checks passed!
```

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/star.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/star.md
@@ -1299,15 +1299,12 @@ reveal_type(Nope)  # revealed: Unknown
 ## `global` statements in non-global scopes
 
 Python allows `global` statements in function bodies to add new variables to the global scope, but
-we require a matching global binding or declaration. We lint on unresolved `global` statements, and
 we don't include the symbols they might define in `*` imports:
 
 `a.py`:
 
 ```py
 def f():
-    # error: [unresolved-global] "Invalid global declaration of `g`: `g` has no declarations or bindings in the global scope"
-    # error: [unresolved-global] "Invalid global declaration of `h`: `h` has no declarations or bindings in the global scope"
     global g, h
 
     g = True
--- a/crates/ty_python_semantic/resources/mdtest/scopes/global.md
+++ b/crates/ty_python_semantic/resources/mdtest/scopes/global.md
@@ -241,26 +241,32 @@ def f():
 # TODO: reveal_type(x)  # revealed: Unknown | Literal["1"]
 ```
 
-## Global variables need an explicit definition in the global scope
+## Global variables don't need an explicit definition in the global scope
 
 You're allowed to use the `global` keyword to define new global variables that don't have any
-explicit definition in the global scope, but we consider that fishy and prefer to lint on it:
+explicit definition in the global scope:
 
 ```py
-x = 1
-y: int
-# z is neither bound nor declared in the global scope
-
 def f():
-    global x, y, z  # error: [unresolved-global] "Invalid global declaration of `z`: `z` has no declarations or bindings in the global scope"
+    global x
+    x = 42
+
+def g():
+    print(x)  # allowed, resolves to the global `x` defined by `f`
+
+def h():
+    print(y)  # error: [unresolved-reference]
 ```
 
-You don't need a definition for implicit globals, but you do for built-ins:
+However, this only affects the "public" type of the global. It's still considered unbound when
+module-scope code refers to it locally.
 
 ```py
 def f():
-    global __file__  # allowed, implicit global
-    global int  # error: [unresolved-global] "Invalid global declaration of `int`: `int` has no declarations or bindings in the global scope"
+    global x
+    x = 42
+
+print(x)  # error: [unresolved-reference]
 ```
 
 ## References to variables before they are defined within a class scope are considered global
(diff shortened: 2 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/import/star.md
```diff
@@ -1299,15 +1299,12 @@ reveal_type(Nope)  # revealed: Unknown
 ## `global` statements in non-global scopes
 
 Python allows `global` statements in function bodies to add new variables to the global scope, but
-we require a matching global binding or declaration. We lint on unresolved `global` statements, and
 we don't include the symbols they might define in `*` imports:
```
oconnor663, line 1302: We could change this (CPython does include these globals in `*` imports), but we'd need to make `ExportFinder` walk function and class bodies, which it currently doesn't. (And doesn't want to?)

oconnor663:
That makes sense, and I think it's reasonable to close this as a won't-fix-for-now-at-least.

But now you've got me thinking about how we currently define the "public type". Right now we have two notions, as I understand them:

- the "local type"(?) taking account of control flow through the local scope
- the "public type" taking account of all reachable bindings in the current scope, and now/soon unioning that with all reachable bindings in nested scopes

What if there was a third notion?

- the union of all reachable bindings in nested scopes _only_ (not including the current scope)

Then maybe when inferring the local type (i.e. calling `infer_local_place_load` from `infer_place_load`), we could fall back to nested bindings _only_ (not later bindings in the current scope) when a local variable is unbound or possibly unbound? For example, we currently do this:

```py
def f():
    global x
    x = 42
reveal_type(x)  # Unknown (unbound)
f()
reveal_type(x)  # Unknown (unbound)              <-- false positive
if secrets.randbelow(2):
    x = 1
reveal_type(x)  # Literal[1] (possibly unbound)  <-- incomplete
x = 2
reveal_type(x)  # Literal[2]
f()
reveal_type(x)  # Literal[2]                     <-- incorrect
```

But if we wanted to, I think we could achieve this without any new analysis:

```py
def f():
    global x
    x = 42
reveal_type(x)  # Literal[42] (possibly unbound)  <-- false negative
f()
reveal_type(x)  # Literal[42] (possibly unbound)
if secrets.randbelow(2):
    x = 1
reveal_type(x)  # Literal[42, 1] (possibly unbound)
x = 2
reveal_type(x)  # Literal[2]
f()
reveal_type(x)  # Literal[2]                      <-- incorrect
```

Would that be worth pursuing? My default reaction is actually no (this is making things more complicated, and I'm not sure how common it is to see code that benefits from any of this), but let me know what you think.

crates/ty_python_semantic/resources/mdtest/import/star.md
carljm: > CPython does include these globals in `*` imports

Well, maybe -- it does if the function that sets the global has been executed before the import occurs, otherwise it doesn't. So I guess we could consider it a "possibly bound" symbol.

carljm:
I think we probably should fix the _incorrect_ results in the above examples (not necessarily right now, but at some point). I'm not that tempted by the exact change proposed above because it doesn't achieve that. (It's not clear to me why the nonlocal writes should be considered if the name is locally unbound, but not otherwise.)

The _simplest_ (but imprecise) way to do that is to union all inferred types of names that have nonlocal bindings with a possible binding of `Unknown`. This is basically "pretend we can't see what exactly the nonlocal bindings are, but we know they exist and therefore the name can be bound to _some_ new value from outside the scope", which is how we handle external writes to fully public scopes like classes or modules (where it really isn't feasible for us to see all the possible writes). But it feels weird to do this when in this case we clearly can see the precise nonlocal writes, and we do take them into account in the public type, just not in the local types.

The other approach (which would not be that hard; you were going to do it in the previous PR until I said not to bother) would be to _always_ include nonlocal bindings as part of the possible type for any use of a local name. This is not wrong, but I'm worried about false positives if we just do that much without more precision.

We could perhaps improve this by considering the creation of any nested scope with any nonlocal writes to a symbol to be a conditional write to that symbol in the local scope's control flow. That would mean we would only consider those nonlocal writes in places in control flow that are reachable from the definition of the nested scope. I think this is probably the best answer, but it definitely makes it a bit more complicated to implement.

I don't think we should prioritize any of this right now; I'd rather have our priorities driven by real-world bug reports. But I will open an issue to track it.

carljm:
Opened https://github.com/astral-sh/ty/issues/958

PR #19822, opened 2025-08-08
[ty] Send a single request for registrations/unregistrations

dhruvmanila:
## Summary

This is a small refactor to update the server to send a single request to perform registrations and unregistrations of dynamic capabilities.

## Test Plan

Existing E2E test cases pass, add a new test case to verify multiple registrations.

### DIFF
--- a/crates/ty_server/src/session.rs
+++ b/crates/ty_server/src/session.rs
@@ -700,7 +700,12 @@ impl Session {
         }
 
         for unregistration in &unregistrations {
-            self.registrations.remove(&unregistration.method);
+            if !self.registrations.remove(&unregistration.method) {
+                tracing::debug!(
+                    "Unregistration for `{}` was requested, but it was not registered",
+                    unregistration.method
+                );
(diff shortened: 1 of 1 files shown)


crates/ty_server/src/session.rs
```diff
@@ -587,134 +586,132 @@ impl Session {
...
+        unregistrations: Vec<Unregistration>,
+    ) {
+        if unregistrations.is_empty() {
+            return;
+        }
 
-        if !registered {
-            self.registrations.insert(Rename::METHOD.to_string());
+        for unregistration in &unregistrations {
+            self.registrations.remove(&unregistration.method);
```
MichaReiser: Should we error/warn if there's an `unregistration` that isn't present in `self.registrations`?

dhruvmanila: I don't know if that's disallowed or not as there's nothing like that mentioned in the spec but I think it would be good to have it at the debug level.

PR #19824, opened 2025-08-08
[ty] Improve performance of subtyping and assignability checks for protocols

MichaReiser:
## Summary

It seems unnecessary to write to `seen` if we have a cache entry (in which case we pop the seen immediately).

I doubt this has any meaningful performance impact but it makes it clearer that `cache` is a superset of `seen`.

## Test Plan

`cargo test`


AlexWaygood:
> I doubt this has any meaningful performance impact

it actually has a 3% perf improvement on `DateType`, a codebase where typechecking necessitates a large number of expensive assignability checks between protocol types! https://codspeed.io/astral-sh/ruff/branches/micha%2Fcyclic-small-cleanup?runnerMode=Instrumentation

PR #19677, opened 2025-08-01
[ty] support `kw_only=True` for `dataclass()` and `field()`

thejchap:
## Summary
https://github.com/astral-sh/ty/issues/111

adds support for `@dataclass(kw_only=True)` (https://docs.python.org/3/library/dataclasses.html)

## Test Plan
- new mdtests
- triaged conformance diffs (notes here: https://diffswarm.dev/d-01k2gknwyq82f6x17zqf3apjxc)
- `mypy_primer` no-op

PR #19825, opened 2025-08-08
[ty] diagnostic for dataclass field order

PrettyWood:
## Summary
Emit diagnostic when defining a field without a default after a field with a default, see [existing TODO](https://github.com/astral-sh/ruff/blob/dbc137c9516808baa292da7a928e50ba36a14d39/crates/red_knot_python_semantic/resources/mdtest/dataclasses.md?plain=1#L120)

Screenshot 2025-08-08 at 13 46 29

Part of https://github.com/astral-sh/ty/issues/111

## Test Plan
Updated Markdown tests


carljm:
Looking at the conformance suite results, I think we need to take `kw_only` into account here (keyword-only fields don't participate in this check, since they go to the end of the `__init__` signature and are not positional). I think also `init=False` fields should be ignored?

So I think we should land https://github.com/astral-sh/ruff/pull/19677 and then take those into account.

PrettyWood:
Great thank you so much for the review + improvements @carljm! I’m still getting familiar with the codebase, so your feedback is really helpful.
You’re right about `kw_only` and `init=False` fields. I reckon we should merge this first and I'll come back to this PR afterward. Tell me if I can help

carljm:
We already supported `init` arg, and I just merged the PR for `kw_only` support, so I think we are ready to add consideration of those to this PR! If you are up for doing that, that would be great.

PrettyWood:
Sure I'll work on it this weekend

PrettyWood:
@carljm Should be good. Feel free to update directly the branch!

PrettyWood:
@carljm Ok I'll take a step back on this PR. Probably best to first look at https://github.com/astral-sh/ty/issues/1047 (I suggested a fix https://github.com/astral-sh/ruff/pull/19986) and wait for feedbacks from others.
Thank you very much for the very precise reviews.

carljm:
@PrettyWood Your fix for astral-sh/ty#1047 looked good, I merged it!

FWIW I don't think there's any need to take a step back here or wait for other input, especially with that fix merged. I think this PR is close, we just need to add some tests for the cases I mentioned above (where there is dataclass-level `kw_only` kwarg or dataclass-transform-level `kw_only_default` kwarg) and figure out how to handle them best. If you don't have time and would rather hand this off to someone else, that's of course fine, just let us know. But if you want to propose a refactor that allows us to handle those cases, I think that would be the next step here.

Really appreciate your contributions!

PrettyWood:
Perfect thank you! Happy to keep working on it for sure! I'll work on the follow-up tomorrow

PrettyWood:
@carljm I kept extending `own_fields` directly (since I already added some contextual logic with `_: KW_ONLY` that sets `field.kw_only = Some(true)`). I don’t think we need a separate “processed fields” yet (if ever) but if you’d prefer me to split it I'll do it

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -1400,19 +1400,33 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    if field.is_kw_only_sentinel(self.db()) {
+                        kw_only_sentinel_fields.push(name);
+                        continue;
+                    }
+                    if !field.init || field.kw_only == Some(true) {
+                        continue;
+                    }
+
+                    // In bind.rs, we set `Type::unknown()` if no `default` (or `default_factory`) has been set
+                    if field.default_ty.is_some_and(|t| !t.is_unknown()) {
```
PrettyWood: This is to support 

```py
@dataclass(kw_only=True)
class KwOnlyClassWithPositionalField:
    x: int = 1
    y: str = field(kw_only=False, default="hello")
    # error: [dataclass-field-order] "Required field `z` cannot be defined after fields with default values"
    z: float = field(kw_only=False)
```

because when `dataclasses.field` is used, `default` is not `None` but `Some(Dynamic(Unknown))`

TBH I think it's a bit weird. I would expect in this example to have both `x` and `y`
with `default_ty = None`

```py
@dataclass
class A:
    x: int
    y: int = field(metadata={"random": "info"})
```

But it meant refactoring `FieldInstance` to have `default_type` being an `Option`
So in the meantime I think it's ok like this. Happy to go further if needed

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -487,6 +565,23 @@ a = A(1, 2)
 a = A(x=1, y=2)
 ```
 
+All fields participate in positional field ordering checks:
```
carljm: Shouldn't this say the opposite? No fields (of a `kw_only=True` dataclass) participate in field ordering checks.

crates/ty_python_semantic/src/types/infer.rs
carljm: I do think we should fix this properly in this PR by making `default_ty` be an `Option`. It is possible for `default_ty` to be `Unknown` for other reasons, like if there was an unknown import for example. These cases shouldn't be treated as "no default".

I think this problem is currently causing false positives in the ecosystem report for this PR, for example [in trio](https://github.com/python-trio/trio/blob/main/src/trio/_channel.py#L152) where we consider the return type of `attrs.Factory` to be `Unknown` (due to limitations in our current constraint solver), and so we treat that wrongly as "no default value".

carljm: Looking at the conformance suite results, the one remaining false positive we add in this PR is at https://github.com/python/typing/blob/main/conformance/tests/dataclasses_usage.py#L186

I think if we fix `default_ty` to be an `Option`, then we can easily fix this false positive also, by ensuring that we set it to `Some` if `default_factory` is specified. We could for now just make that `Some(Unknown)`, although I think adding proper support for `default_factory` would actually be quite easy -- we just need to take the argument type of `default_factory` and `Type::try_call` it with no arguments, and use the return type of that call. This could also be done as a separate PR, though I'd probably prefer that PR to land first in that case, so we avoid adding a new false positive with this diagnostic.

crates/ruff_benchmark/benches/ty_walltime.rs
```diff
@@ -142,7 +142,7 @@ static FREQTRADE: std::sync::LazyLock<Benchmark<'static>> = std::sync::LazyLock:
             max_dep_date: "2025-06-17",
             python_version: PythonVersion::PY312,
         },
-        400,
+        600,
```
carljm: I think we should revert this change, and if it causes problems that's a sign that means this PR isn't ready to land yet. Freqtrade does not have a large number of true positives for this diagnostic, so if we are causing a large number of new diagnostics, they are almost certainly false positives, and we don't want to land new diagnostics with large numbers of false positives on real ecosystem code.

The freqtrade false positives are weird, because I would have expected them to be fixed with this PR; they are false positives on lines like https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/rpc/api_server/api_schemas.py#L19, because we see previous fields on [pydantic.BaseModel](https://github.com/pydantic/pydantic/blob/f7a9b73517afecf25bf898e3b5f591dffe669778/pydantic/main.py#L121), and for some reason we aren't respecting `kw_only_default` [on their dataclass-transform metaclass](https://github.com/pydantic/pydantic/blob/f7a9b73517afecf25bf898e3b5f591dffe669778/pydantic/_internal/_model_construction.py#L78).

Boiling this down to [a minimal example](https://play.ty.dev/15f3c09e-ad52-45d3-8a1c-1d7b1bf64d8f), it looks to me like our support for `kw_only_default` just currently doesn't work for metaclass-based dataclass-transforms, only for decorator-based ones.

Unfortunately I think this needs to be fixed (whether in this PR or a separate one that lands first) before we can land this PR; it just causes too many false positives in real projects otherwise.

carljm (changes requested):
Thank you! The updates look great, and I agree we don't need a new abstraction layer (yet?) -- it's fine to just ensure that the fields returned from `own_fields` already have a fully correct `kw_only` value. My main concern is just that we be clear about the contract: do the Field objects returned from `own_fields` represent exactly what is explicitly specified in that Field definition, or do they represent our full understanding of the field, including what's inherited from dataclass arguments and dataclass-transform arguments? I think it's fine if it's the latter, so long as that's consistent.

Sorry for requesting changes again -- in many cases we are happy to land PRs with remaining TODOs, but adding new false positives in the ecosystem is something we are very reluctant to do, and at the moment this PR adds a lot of them. It turns out we need to improve our support for other dataclass features significantly in order to add this diagnostic without false positives.

crates/ruff_benchmark/benches/ty_walltime.rs
PrettyWood: Yes sorry this should have been reverted.
And no problem happy to take the time to fully tackle everything. I'll look at it tomorrow

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
PrettyWood: 🙈 oops

PrettyWood:
> think it's fine if it's the latter, so long as that's consistent.

For me yes. Happy to add a docstring on `own_fields` to make it explicit

> I think if we fix default_ty to be an Option

Perfect glad we agree. I didn't want to put too much in a single PR but you're right, better that than having new false positives. I'll work on it tomorrow

PrettyWood:
@carljm Better :) Commits should be self-explanatory.

carljm:
Thank you! The conformance suite results look great here now! The ecosystem results look mostly great :) The one case where it looks like we are still adding false positives is that [we don't support `field_specifiers` on `dataclass_transform`](https://github.com/astral-sh/ty/issues/1068), so we interpret `= attrs.field(...)` as a default value rather than as a field specifier, leading to the false positives we see on trio, svcs, and the attrs tests.

Ideally, we would support `field_specifiers` before we land this, so we avoid adding false positives. Trying to decide if we can go in the other direction and land this first (just to avoid collecting merge conflicts here in the meantime), but I think it would make it important that we address `field_specifiers` soon.

PrettyWood:
I'll fix the `field_specifiers` in a dedicated PR and we'll rebase/merge this one after

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -110,6 +110,35 @@ CustomerModel(id=1, name="Test")
 CustomerModel()
 ```
 
+### Metaclass with `kw_only_default=True`
```
carljm: This test looks great. Nit: I think this section is to show basic usage, and arguments are tested below. I think we should move this down with the rest of the `kw_only_default` tests below, and just add an example with metaclass there. (I think the example here with decorator is probably redundant with the test we already have down there, and we don't need both?)

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -231,6 +260,28 @@ class CustomerModel:
 c = CustomerModel(1, "Harry")
```
carljm, line 247: The test immediately above here I thought should be fixed by logic you added in this PR, but then I realized why it isn't -- commented below.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2468,9 +2477,24 @@ impl<'db> ClassLiteral<'db> {
...
+                        .dataclass_params(db)
+                        .is_some_and(|params| params.contains(DataclassParams::KW_ONLY))
+                        || self.try_metaclass(db).is_ok_and(|(_, transformer_params)| {
+                            transformer_params.is_some_and(|params| {
+                                params.contains(DataclassTransformerParams::KW_ONLY_DEFAULT)
+                            })
+                        })
+                    {
+                        // `@dataclass(kw_only=True)` or `@dataclass_transform(kw_only_default=True)`
+                        field.kw_only = Some(true);
```
carljm: There's a test (I commented above) that I would expect this logic to fix, but it doesn't fix it. And I think I understand why. Currently `DataclassParams` is just a bitset, but this isn't really adequate when we have to compose it with `DataclassTransformerParams`. With the bitset representation, we can't distinguish "no `kw_only` argument" from `kw_only=False` -- they are both represented by the bit being off. I think we will need to change `DataclassParams` so that at least `kw_only` can be tri-valued. Because "no `kw_only` arg" should default to the dataclass-transform's `kw_only_default` value, but `kw_only=False` should potentially override it.

I think this doesn't need to be fixed in this PR, but it does impact the correctness of this code, so I think we should add a TODO comment here.
 
```suggestion
                        // TODO `kw_only=False` should override `kw_only_default=True`, this will require
                        // a tri-valued representation for `DataclassParams::KW_ONLY`
                        .is_some_and(|params| params.contains(DataclassParams::KW_ONLY))
                        || self.try_metaclass(db).is_ok_and(|(_, transformer_params)| {
                            transformer_params.is_some_and(|params| {
                                params.contains(DataclassTransformerParams::KW_ONLY_DEFAULT)
                            })
                        })
                    {
                        // `@dataclass(kw_only=True)` or `@dataclass_transform(kw_only_default=True)`
                        field.kw_only = Some(true);
```

sharkdp:
I started rebasing this branch, but the changes on `main` in the meantime were too large so I did not preserve the commit structure and recreated it from scratch instead. I preserved the original state of the PR [here](https://github.com/astral-sh/ruff/tree/dataclass-field-order-backup).

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -110,14 +110,93 @@ reveal_type(D.__init__)  # revealed: (self: D, x: int) -> None
...
+    y: str = field(kw_only=True)
+    z: float = field(kw_only=True, default=2.0)
+
+@dataclass
+class BadWithKwOnly:
+    x: int = 1
+    y: str = field(kw_only=True)
+    # error: [dataclass-field-order] "Required field `z` cannot be defined after fields with default values"
+    z: float
+```
```
carljm, line 184: I feel like there's a missing case here to demonstrate that a field with `kw_only=True` and a default value does not prevent a subsequent required field:

```py
@dataclass
class AlsoGoodWithKwOnly:
    x: int = field(kw_only=True, default=1)
    y: str
```

sharkdp:
> Thank you! The conformance suite results look great here now! The ecosystem results look mostly great :) The one case where it looks like we are still adding false positives is that [we don't support `field_specifiers` on `dataclass_transform`](https://github.com/astral-sh/ty/issues/1068), so we interpret `= attrs.field(...)` as a default value rather than as a field specifier, leading to the false positives we see on trio, svcs, and the attrs tests.
>
> Ideally, we would support `field_specifiers` before we land this, so we avoid adding false positives. Trying to decide if we can go in the other direction and land this first (just to avoid collecting merge conflicts here in the meantime), but I think it would make it important that we address `field_specifiers` soon.

We do now support `field_specifiers` since https://github.com/astral-sh/ruff/pull/20888, and indeed, all false positives are gone! Ecosystem report looks great now (the `unused-type-ignore-comment` changes are related to this PR, everything else is due to flaky diagnostics).


I'll look into resolving the remaining review comments soon.

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
sharkdp: Added

crates/ty_python_semantic/src/types/class.rs
sharkdp: > With the bitset representation, we can't distinguish "no kw_only argument" from kw_only=False -- they are both represented by the bit being off. I think we will need to change DataclassParams so that at least kw_only can be tri-valued. Because "no kw_only arg" should default to the dataclass-transform's kw_only_default value, but kw_only=False should potentially override it.

This issue has been resolved in the meantime (and I added some new tests cases for it). We still use a bitset, but we always set it to the final "computed" value. So for `kw_only_default=True` and no `kw_only` parameter in the field-specifier, we set `kw_only=True`. When `kw_only_default` and `kw_only` are unset, we set it to the implicit default (`kw_only=False`).

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -667,9 +667,9 @@ class Person:
     age: int = field(default=0)
     tags: list[str] = field(default_factory=list)
     email: str = field(kw_only=True)
-    internal_notes: str = field(alias="notes")
+    internal_notes: str = field(alias="notes", default="")
```
sharkdp, line 670: The new check founds some bugs :smile:

sharkdp:
The new diagnostics for `attrs` dataclasses are concerning, looking into it.

sharkdp:
> The new diagnostics for `attrs` dataclasses are concerning, looking into it.


The fix in [2ebd8d0](https://github.com/astral-sh/ruff/pull/19825/commits/2ebd8d083b4e7b78c56f72505c2cbb4d38acce1c) got rid of some of these.

This leaves us with:


> ```diff
> + tests/test_hooks.py:152:13: error[dataclass-field-order] Required field `y` cannot be defined after fields with default values
> + tests/test_hooks.py:154:48: error[missing-argument] No argument provided for required parameter `y`
> + tests/test_next_gen.py:142:14: error[unresolved-attribute] Object of type `dataclasses.Field` has no attribute `validator`
> + tests/test_next_gen.py:380:14: error[unresolved-attribute] Object of type `dataclasses.Field` has no attribute `validator`
> ```

These would require an attrs plugin. `field_order` is not a standard feature.

https://github.com/python-attrs/attrs/blob/c44b8b0052ca3a4594219efa702129e1537d8966/tests/test_hooks.py#L152
https://github.com/python-attrs/attrs/blob/c44b8b0052ca3a4594219efa702129e1537d8966/tests/test_hooks.py#L154
https://github.com/python-attrs/attrs/blob/c44b8b0052ca3a4594219efa702129e1537d8966/tests/test_next_gen.py#L142

> ```diff
> + tests/test_make.py:829:13: error[dataclass-field-order] Required field `val4` cannot be defined after fields with default values
> ```

~~False positive, looking into it...~~. Oh, this happens because we check attrs with Python 3.9, and kw_only was not introduced back then. That took me a while to understand. So this is fine, I guess.

https://github.com/python-attrs/attrs/blob/c44b8b0052ca3a4594219efa702129e1537d8966/tests/test_make.py#L829


> ```diff
> + homeassistant/components/device_tracker/legacy.py:298:5: error[dataclass-field-order] Required field `name` cannot be defined after fields with default values
> + homeassistant/components/device_tracker/legacy.py:299:5: error[dataclass-field-order] Required field `platform` cannot be defined after fields with default values
> + homeassistant/components/device_tracker/legacy.py:300:5: error[dataclass-field-order] Required field `config` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:465:5: error[dataclass-field-order] Required field `created_at` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:466:5: error[dataclass-field-order] Required field `disabled_by` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:467:5: error[dataclass-field-order] Required field `id` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:468:5: error[dataclass-field-order] Required field `identifiers` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:469:5: error[dataclass-field-order] Required field `labels` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:470:5: error[dataclass-field-order] Required field `modified_at` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:471:5: error[dataclass-field-order] Required field `name_by_user` cannot be defined after fields with default values
> + homeassistant/helpers/device_registry.py:472:5: error[dataclass-field-order] Required field `orphaned_timestamp` cannot be defined after fields with default values
> ```

These seem correct to me. The fields are preceded by a `LEGACY_SETUP: Final[tuple[str, ...]] = ("...", …)` attribute which seems like it should be a normal field with a default, despite the `Final` qualifier. Other type checkers also error on something like …

```py
@dataclass
class C:
    x: Final[int] = 1
    y: str
```

… and do consider `x` to be a normal field.

See e.g. here: https://github.com/home-assistant/core/blob/78415bc1ffb8ea436b96ef383e108c38384be5d8/homeassistant/components/device_tracker/legacy.py#L298

PR #19827, opened 2025-08-08
Update salsa

MichaReiser:
## Summary

Update salsa to pull in https://github.com/salsa-rs/salsa/pull/961

## Test Plan

`cargo test`


MichaReiser:
The performance regression is a bit unfortunate but it's hard to tell if it's because the validation is now more correct or if the code is actually slower (it's fairly likely that we took some short-cuts in the old version that weren't safe and could lead to stale results)

PR #19833, opened 2025-08-08
[ty] Handle cycles when finding implicit attributes

dcreager:
The [minimal reproduction](https://gist.github.com/dcreager/fc53c59b30d7ce71d478dcb2c1c56444) of https://github.com/astral-sh/ty/issues/948 is an example of a class with implicit attributes whose types end up depending on themselves. Our existing cycle detection for `infer_expression_types` is usually enough to handle this situation correctly, but when there are very many of these implicit attributes, we get a combinatorial explosion of running time and memory usage.

Adding a separate cycle handler for `ClassLiteral::implicit_attribute` lets us catch and recover from this situation earlier.

Closes https://github.com/astral-sh/ty/issues/948

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -89,6 +89,26 @@ fn inheritance_cycle_initial<'db>(
     None
 }
 
+fn implicit_attribute_recover<'db>(
+    _db: &'db dyn Db,
+    _value: &PlaceAndQualifiers<'db>,
+    _count: u32,
+    _class_body_scope: ScopeId<'db>,
+    _name: String,
+    _target_method_decorator: MethodDecorator,
+) -> salsa::CycleRecoveryAction<PlaceAndQualifiers<'db>> {
+    salsa::CycleRecoveryAction::Iterate
+}
+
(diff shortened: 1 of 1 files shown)


carljm (approved):
Nice find! I suspect the win here is not "catching the cycle sooner" but that we are now memoizing a chunk of work that we did not memoize before, and presumably that work was previously being repeated on every cycle iteration, and now is not.

This is a good reminder that there are potentially significant performance wins hiding in "add memoization" that aren't necessarily obvious until we hit a pathological case.

MichaReiser:
This does seem to increae memory usage a fair bit. I'm not suggesting that this wasn't the right fix. But it might have been interesting to measure the total memory usage on a large project and comparing it to main (by using `TY_MEMORY_REPORT=full`).

AlexWaygood:
> The [minimal reproduction](https://gist.github.com/dcreager/fc53c59b30d7ce71d478dcb2c1c56444) of [astral-sh/ty#948](https://github.com/astral-sh/ty/issues/948) is an example of a class with implicit attributes whose types end up depending on themselves.

Is it worth adding something like this as a micro-benchmark, to ensure that we don't regress on this in the future?

PR #19838, opened 2025-08-08
[ty] Perform assignability etc checks using new `Constraints` trait

dcreager:
"Why would you do this? This looks like you just replaced `bool` with an overly complex trait"

Yes that's correct!

This should be a no-op refactoring. It replaces all of the logic in our assignability, subtyping, equivalence, and disjointness methods to work over an arbitrary `Constraints` trait instead of only working on `bool`.

The methods that `Constraints` provides looks very much like what we get from `bool`. But soon we will add a new impl of this trait, and some new methods, that let us express "fuzzy" constraints that aren't always true or false. (In particular, a constraint will express the upper and lower bounds of the allowed specializations of a typevar.)

Even once we have that, most of the operations that we perform on constraint sets will be the usual boolean operations, just on sets. (`false` becomes empty/never; `true` becomes universe/always; `or` becomes union; `and` becomes intersection; `not` becomes negation.) So it's helpful to have this separate PR to refactor how we invoke those operations without introducing the new functionality yet.

Note that we also have translations of `Option::is_some_and` and `is_none_or`, and of `Iterator::any` and `all`, and that the `and`, `or`, `when_any`, and `when_all` methods are meant to short-circuit, just like the corresponding boolean operations. For constraint sets, that depends on being able to implement the `is_always` and `is_never` trait methods.


crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -488,23 +501,30 @@ impl<'db> Signature<'db> {
...
-                .unwrap_or(Type::unknown())
-                .is_equivalent_to(db, other_type.unwrap_or(Type::unknown()))
-        };
+    pub(crate) fn is_equivalent_to_impl<C: Constraints<'db>>(
+        &self,
+        db: &'db dyn Db,
+        other: &Signature<'db>,
+        visitor: &IsEquivalentVisitor<'db, C>,
+    ) -> C {
+        let mut result = C::always(db);
```
dcreager: This is a common pattern in this PR, to implement short-circuiting when we can't (easily) use `when_all` or `when_any`. This method is (and was) a large intersection. Before, we'd use early `return false` to implement the short-circuiting. With `bool`, if any particular element doesn't short-circuit, there's only one possible "current pending result" — `true`. So we didn't need to explicitly track it, and down at the end of the method, where we know made it all the way through without ever getting a short-circuited `false`, we can return `true` directly.

But with constraint sets, we have to create a `result` accumulator, and explicitly intersect each element's constraints into the result, since we might end up with something that is neither "always true" nor "always false". I've structured the return values of `intersect` and `union` so that the check always has the form

```rust
if short_circuit {
    return result;
}
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -1417,69 +1443,79 @@ impl<'db> Type<'db> {
...
                         })
-                    }) =>
+                    })
+                    .is_never(db) =>
             {
-                true
+                // TODO: The repetition here isn't great, but we really need the fallthrough logic,
+                // where this arm only engages if it returns true (or in the world of constraints,
+                // not false). Once we're using real constraint sets instead of bool, we should be
+                // able to simplify the typevar logic.
```
carljm, line 1495: I'm curious how you think the logic can be simplified "once we're using real constraint sets"? To me it appears kind of the opposite -- this duplication is here precisely because we are trying to be correct in a future with real constraint sets, where we need to return the actual accumulated constraints from the inner `has_relation_to_impl` checks. In this PR, with just bool, we can pass all tests without this duplication, by just changing the sense of the match arm `if` clause back to a positive `.is_always(db)` instead of a negated `.is_never(db)`, and making the body of the match arm be `C::always(db)`.

carljm: Considering this case is actually making me wonder about a more fundamental thing: what about a case where this match arm returns some constraints (neither always nor never), but some later match arm would also have returned some constraints, had we reached it? In an ideal world, wouldn't we OR those constraint sets, if we're trying to express the conditions under which the relation would be true?

I guess in the real world, we just pick one of them and don't find all solutions?

crates/ty_python_semantic/src/types.rs
```diff
@@ -1583,42 +1622,37 @@ impl<'db> Type<'db> {
...
-                if callable.is_function_like(db)
-                    && KnownClass::FunctionType
-                        .to_instance(db)
-                        .has_relation_to_impl(db, target, relation, visitor) =>
-            {
-                true
+            (Type::Callable(callable), _) if callable.is_function_like(db) => {
+                KnownClass::FunctionType
+                    .to_instance(db)
+                    .has_relation_to_impl(db, target, relation, visitor)
```
carljm, line 1686: I guess the change in semantics is safe here because the very next match arm unconditionally returns false/never for `(Type::Callable(_), _)`

crates/ty_python_semantic/src/types.rs
```diff
@@ -1294,47 +1298,67 @@ impl<'db> Type<'db> {
     /// intersection simplification dependent on the order in which elements are added), so we do
     /// not use this more general definition of subtyping.
     pub(crate) fn is_subtype_of(self, db: &'db dyn Db, target: Type<'db>) -> bool {
+        self.when_subtype_of(db, target)
+    }
+
+    fn when_subtype_of<C: Constraints<'db>>(self, db: &'db dyn Db, target: Type<'db>) -> C {
```
carljm, line 1338: Do you foresee us needing this genericity over `C: Constraints` long-term? Or is this just needed for smoothing the transition from bool to real constraint sets, and once we've done that, these methods will in practice be monomorphic over just that "constraint sets" implementation of the trait?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -601,34 +603,41 @@ impl<'db> Specialization<'db> {
...
                 }
                 TypeVarVariance::Contravariant => {
                     other_type.has_relation_to_impl(db, *self_type, relation, visitor)
                 }
-                TypeVarVariance::Bivariant => true,
+                TypeVarVariance::Bivariant => C::always(db),
             };
-            if !compatible {
-                return false;
+            if result.intersect(db, compatible) {
```
carljm: I find the return value of `intersect` counter-intuitive here. IIUC this is saying "if this intersection is empty" but it looks like it might be saying "if there is a non-empty intersection".

I think I would find it clearer if `intersect` just returned `self` and we had to explicitly call `is_never()` on it -- but maybe that makes other short-circuiting cases more painful?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,159 @@
+//! Constraints under which type properties hold
+
+use crate::Db;
+
+/// Encodes the constraints under which a type property (e.g. assignability) holds.
+pub(crate) trait Constraints<'db>: Clone + Sized {
+    /// Returns a constraint set that never holds
+    fn never(db: &'db dyn Db) -> Self;
```
carljm: There's the obvious name tension with `Type::Never` here, but I'm not sure I have a better name to offer.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,159 @@
...
+                return result;
+            }
+        }
+        result
+    }
+
+    fn when_all<'db, C: Constraints<'db>>(self, db: &'db dyn Db, mut f: impl FnMut(T) -> C) -> C {
+        let mut result = C::always(db);
+        for child in self {
+            if result.intersect(db, f(child)) {
```
carljm: If these two cases are what we mean when we say above that we use the return types of `intersect` and `union` for short-circuiting, then my opinion in my other comment is unchanged: I think it would be clearer, and not particularly painful, for `union` and `intersect` to return `self` and require an explicit chained call to `.is_always()` or `.is_never()` if you want to use those methods in boolean context.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -112,83 +113,89 @@ impl<'db> CallableSignature<'db> {
...
+    }
+
+    fn is_subtype_of_impl<C: Constraints<'db>>(&self, db: &'db dyn Db, other: &Self) -> C {
+        self.has_relation_to_impl(
             db,
-            &self.overloads,
-            &other.overloads,
+            other,
             TypeRelation::Subtyping,
+            &HasRelationToVisitor::new(C::always(db)),
```
carljm: Thanks for also increasing our visitor coverage in this diff!

crates/ty_python_semantic/src/types/signatures.rs
carljm: Aha, and these cases are why you want `union` and `intersect` to return a somewhat-counter-intuitive boolean? But even here, it looks like you'd only need to call `is_never()` in one place, inside `types_inconsistent`.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -488,23 +501,30 @@ impl<'db> Signature<'db> {
...
-                .is_equivalent_to(db, other_type.unwrap_or(Type::unknown()))
-        };
+    pub(crate) fn is_equivalent_to_impl<C: Constraints<'db>>(
+        &self,
+        db: &'db dyn Db,
+        other: &Signature<'db>,
+        visitor: &IsEquivalentVisitor<'db, C>,
+    ) -> C {
+        let mut result = C::always(db);
+        let mut types_inconsistent =
```
carljm: Not sure I see a good reason for the rename from `check_types` to `types_inconsistent` (with the reversed sense). I usually think negatives in method names should be avoided, and `if types_inconsistent(...)` doesn't seem better to me than `if !check_types(...)`

carljm (approved):
This is quite cool! Very impressed with how you wrangled the Rust API to make the translation fairly mechanical in many cases, even though you're adding significant capability.

crates/ty_python_semantic/src/types/generics.rs
dcreager: > I think I would find it clearer if `intersect` just returned `self` and we had to explicitly call `is_never()` on it -- but maybe that makes other short-circuiting cases more painful?

It would need to both update `self` in-place (since it's an accumulator) and also return a clone that the caller would then call `is_never` on.  I thought that could be inefficient once we have non-`bool` implementations, and that it would be better for this to return a `bool`.

Maybe if I rename the method to `intersect_short_circuits`, so that it's clearer at the call site what the return value represents?

crates/ty_python_semantic/src/types/signatures.rs
dcreager: > Aha, and these cases are why you want `union` and `intersect` to return a somewhat-counter-intuitive boolean?

Yep, exactly

> But even here, it looks like you'd only need to call `is_never()` in one place, inside `types_inconsistent`.

That's true for this case, but there are other places where we're doing the short circuit check several times in a single method, so I wanted it to be as ergonomic as possible.

(But more importantly, the efficiency bit that I mention in the other comment thread)

crates/ty_python_semantic/src/types.rs
dcreager: Yes that's right, and I wanted to avoid the repetition like I have up in the typevar arm

crates/ty_python_semantic/src/types.rs
dcreager: Originally I thought that we would eventually be able to remove the trait. But in Discord @AlexWaygood asked about whether this PR's change would be compatible with returning a `Result` from `has_relation_to`, etc. I think we could do that by implementing the trait for that `Result` type, too (possibly wrapping the impl of the `T` inside the result). So...I'm not sure? Hopefully but maybe not.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: `unsatisfiable` or `never_satisfiable`, but that's longer to type :smile:

crates/ty_python_semantic/src/types/signatures.rs
dcreager: This was an attempt to be consistent that everywhere that does a short-circuit check has the same syntactic form, with a positive `if` condition check.  Maybe `type_check_short_circuits`, similar to my other rename suggestion, to keep the positive `if` condition but so that there's no negative in the name?

crates/ty_python_semantic/src/types/generics.rs
carljm: Can't it return `&Self` without needing a clone?

dcreager: Hmm, maybe, I can try that.  Does that work if the receiver is `&mut self`?

crates/ty_python_semantic/src/types.rs
dcreager: My thinking is not that the stuff inside the match arm would change all that much. The part that I find too complex in its current state is that so much depends on the _order_ of the match arms. My hunch is that the fall-through here won't be needed, and the ordering will no longer matter as much, because we'll be able to better combine the results of each recursive call with the other surrounding constraints.

crates/ty_python_semantic/src/types/generics.rs
dcreager: > Does that work if the receiver is `&mut self`?

Looks like [it does](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1aeb86e1df6189c64731585587673f9d)!

crates/ty_python_semantic/src/types/signatures.rs
dcreager: These now return more-intuitive values

crates/ty_python_semantic/src/types/generics.rs
dcreager: Updated to return `&Self`, and to use explicit `is_never` and `is_always` calls at the call sites

crates/ty_python_semantic/src/types/constraints.rs

crates/ty_python_semantic/src/types/signatures.rs
dcreager: The name change is reverted back to `check_types`.  I also opted to keep the meaning of the returned boolean be the same, so the call sites are back to being `!check_types`, as before. That means that there's technically a double-negation happening, but that doesn't bother me.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Returns a constraint set that always holds
+    fn always(db: &'db dyn Db) -> Self;
+
+    /// Returns whether this constraint set never holds
+    fn is_never(&self, db: &'db dyn Db) -> bool;
+
+    /// Returns whether this constraint set always holds
+    fn is_always(&self, db: &'db dyn Db) -> bool;
+
+    /// Updates this constraint set to hold the union of itself and another constraint set. Returns
```
```suggestion
    /// Updates this constraint set to hold the union of itself and another constraint set.
```

carljm (approved):
Looks great!

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: > `unsatisfiable` or `never_satisfiable`, but that's longer to type 😄

I sort-of prefer those, actually! `unsatisfiable` doesn't seem too verbose to me. Not a big deal though.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,153 @@
+//! Constraints under which type properties hold
```
AlexWaygood, line 1: It doesn't have to be added in this PR, but I'd love some more documentation about what a constraints set is exactly, with some examples about when it might matter that we return a constraints set rather than a `bool` from `has_relation_to_impl`. As it is, I think I'd feel very nervous editing this file if I needed to make updates for a PR of my own; I don't feel like I have a confident enough understanding of everything that's going on here

crates/ty_python_semantic/src/types/constraints.rs
dcreager: No problem, it's just a couple of `sed` commands to make the change!

For the predicates, I've gone with `is_{always,never}_satisfied`. For the constructors, I want them to be consistent, so I went with `unsatisfied`/`always_satisfied`.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Done! 

> when it might matter that we return a constraints set rather than a `bool` from `has_relation_to_impl`

This part relates to @carljm's question above https://github.com/astral-sh/ruff/pull/19838#discussion_r2289557108, about whether the trait will stick around long-term or not. My goal is that the `bool` impl will go away soon (likely in the very next PR https://github.com/astral-sh/ruff/pull/19997)

AlexWaygood: Thank you, that module doc-comment is exactly what I was looking for!

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+//!
+//! For "concrete" types (which contain no type variables), type properties like assignability have
+//! simple answers: one type is either assignable to another type, or it isn't. (The rules for
```
```suggestion
//! simple answers: one type is either assignable to another type, or it isn't. (The _rules_ for
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+//! comparing two particular concrete types can be rather complex, but the _answer_ is a simple
+//! "yes" or "no".)
+//!
+//! These properties are more complex when type variables are involved, because there are (usually)
+//! many different concrete types that a typevar can be specialized to, and the type property might
+//! hold for some specializations, but not for others. That means for types that include typevars
+//! "is this type assignable to another" no longer makes sense as a question. The better question
+//! is "under what constraints is this type assignable to another".
```
```suggestion
//! hold for some specializations, but not for others. That means that for types that include
//! typevars, "Is this type assignable to another?" no longer makes sense as a question. The better
//! question is: "Under what constraints is this type assignable to another?".
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+//!
+//! This module provides the machinery for representing the "under what constraints" part of that
+//! question. An individual constraint restricts the specialization a single typevar to be within a
```
```suggestion
//! question. An individual constraint restricts the specialization of a single typevar to be within a
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+//! particular lower and upper bound. You can then build up more complex constraint sets using
+//! union, intersection, and negation (just like types themselves).
```
```suggestion
//! union, intersection, and negation operations (just like types themselves).
```

PR #19806, opened 2025-08-07
[ty] Fix a few more diagnostic differences from Ruff

ntBre:
## Summary

Fixes the remaining range reporting differences between the `ruff_db` diagnostic rendering and Ruff's existing rendering, as noted in https://github.com/astral-sh/ruff/pull/19415#issuecomment-3160525595.

This PR is structured as a series of three pairs. The first commit in each pair adds a test showing the previous behavior, followed by a fix and the updated snapshot. It's quite a small PR, but that might be helpful just for the contrast.

You can also look at [this range](https://github.com/astral-sh/ruff/pull/19415/files/052e656c6c1bd7586cbf97f049415d7ace08c631..c3ea51030d38d481aa10855ca1d2556e391ff4cb) of commits from #19415 to see the impact on real Ruff diagnostics. I spun these commits out of that PR.

## Test Plan

New `ruff_db` tests

PR #19839, opened 2025-08-08
Avoid underflow in default ranges before a BOM

ntBre:
Summary
--

This fixes a regression caused by the BOM handling in #19806. Most diagnostics already account for the BOM in their ranges, but those that use `TextRange::default` to mean the beginning of the file do not, causing an underflow in `RenderableAnnotation::new` when subtracting the BOM-shifted `snippet_start` from the annotation range.

I ran into this when trying to run benchmarks on CPython in preparation for caching work. The file `cpython/Lib/test/bad_coding2.py` was causing a crash because it had a default-range `I002` diagnostic, with a BOM.

https://github.com/astral-sh/ruff/blob/7cc3f1ebe9386e77e7009bc411fc6480d3851015/crates/ruff_linter/src/rules/isort/rules/add_required_imports.rs#L122-L126

The fix here is just to saturate to zero instead of panicking. I considered adding a `TextRange::saturating_sub` method, but I wasn't sure it was worth it for this one use. I'm happy to do that if preferred, though.

Saturating seemed easier than shifting the affected annotations over, but that could be another solution.

Test Plan
--

A new `ruff_db` test that reproduced the issue and manual testing against the CPython file mentioned above

### DIFF
--- a/crates/ruff_db/src/diagnostic/render.rs
+++ b/crates/ruff_db/src/diagnostic/render.rs
@@ -710,7 +710,11 @@ impl<'r> RenderableAnnotation<'r> {
     /// lifetime parameter here refers to the lifetime of the resolver that
     /// created the given `ResolvedAnnotation`.
     fn new(snippet_start: TextSize, ann: &'_ ResolvedAnnotation<'r>) -> RenderableAnnotation<'r> {
-        let range = ann.range - snippet_start;
+        // This should only ever saturate if a BOM is present _and_ the annotation range points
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):

MichaReiser:
Is this marked as internal because the change hasn't been released yet?

ntBre:
> Is this marked as internal because the change hasn't been released yet?

Yep, that was my thinking. Otherwise it's a bug.

Issue #6262, opened 2023-08-02
exclude pattern with `*` in file name incorrectly matches entire file path


DetachHead:
```
ruff --exclude foo*.py
```
this incorrectly excludes the file `foo/bar/baz.py`.

since `excludes` takes globs (as mentioned [here](https://beta.ruff.rs/docs/configuration/#pyprojecttoml-discovery)), i would expect it to only match python files with names starting with `foo`.

i'm on windows (which uses `\` as path separators), that may be related

charliermarsh:
We use `globset` which lets `*` match the path separator by default, but is configurable to disable that behavior (https://docs.rs/globset/latest/globset/#syntax).

@zanieb - could use a second opinion, any objection to changing that?

zanieb:
What are you expected to do to get the desired behavior without `literal_separator` set? What's the point of `**` then?

I'm a little confused that this is the default in `globset` @burntsushi is probably best for a second opinion :D

It'd be a breaking change for us to change this, but I think it'd probably be best to match Unix shell behavior.

BurntSushi:
On mobile so apologies for short reply.

Globbing is pretty tricky because it depends on the context in which it is run. For example, at a shell, you might expect `*.py` to only match Python files in the current directory, but in a `.gitignore` file you would expect (and is what happens) `*.py` to exclude both `foo.py` and `bar/foo.py`. In the gitignore world, you would restrict your glob to the current directory (relative to the `.gitignore` file) by prefixing the glob with a `/`. So, `/*.py` would match `foo.py` but not `bar/foo.py`.

So whether `*.py` is supposed to match a path separator or not depends a lot on the context in which a glob is applied. But generally speaking, for exclude rules I would try to follow what gitignore does. (With that said, I haven't looked at in detail the context here, so take what I'm saying with a grain of salt.)

charliermarsh:
Thanks! I'm impressed that you can manage code formatting on mobile.

charliermarsh:
We have kind of a weird thing going on, which I think I followed from Flake8 or Black. When given a pattern like `foo*.py`, we match against _both_ the basename and the absolute path (when testing to exclude a given file or directory).

So it may make sense to turn on `literal_separator`, since we'd still match `*.py` on the cases described above. Or we may want to reconsider our behavior here more holistically.

BurntSushi:
> Or we may want to reconsider our behavior here more holistically.

👍

tylerlaprade:
[...]

PR #19841, opened 2025-08-09
[ty] ensure explicitly provided folders to the cli are fully checked

leandrobbraga:
When users directly specify directories on the command line (e.g., `ty check dir2`), they expect those directories to be fully processed, including all their contents. However, the directory contents were still subject to include rules in `pyproject.toml`, leading to confusing behavior.

## Example
```
dir1/
└─ test.py
dir2/
└─ test.py
```
```
# pyproject.toml
[tool.ty.src]
include = ["dir1"]
```

```
> ty check dir1 -v
INFO Indexed 1 file(s) in 0.005s
All checks passed!

> ty check dir2 -v 
INFO Indexed 0 file(s) in 0.005s #
WARN No python files found under the given path(s)
All checks passed!

> ty check dir2/test.py -v
INFO Indexed 1 file(s) in 0.003s
All checks passed!
```

This behavior would happen since `ty` would only consider the exact provided path as *required* to be checked, all the folders contents were still subject to the include rules in `pyproject.toml`.

This change implicitly adds any CLI directory argument to the include configuration as a `dir/**/*` glob pattern.

Fixes https://github.com/astral-sh/ty/issues/936


crates/ty/src/args.rs
```diff
@@ -177,6 +177,43 @@ impl CheckCommand {
...
+            let glob_patterns: Vec<RelativeGlobPattern> = {
+                self.paths
+                    .iter()
+                    .filter_map(|path| {
+                        let path = path.as_std_path();
+                        if !path.is_dir() {
+                            return None;
+                        }
+
+                        let Ok(canonical_path) = dunce::canonicalize(path) else {
```
MichaReiser: Can you tell me what the motivation is for canonicalizing the input paths?

MichaReiser (commented):
Nice. I'll take a closer look next week

crates/ty/src/args.rs
leandrobbraga: I was trying to solve the issue where the user provides paths such as:
- ".."
- "dir/../../other_dir"

Those paths fails as `invalid-glob`.

I could just reject any path containing ".." as well.

I'm not very used to working with paths, specially on other platforms such as Window. So I might be doing something that should be easy in a hard way.

onerandomusername:
I feel like astral already has a specification with the includes of ruff, it feels to me that this command should have very similar behavior to `ruff check .`, and `ruff format .`

leandrobbraga:
> I feel like astral already has a specification with the includes of ruff, it feels to me that this command should have very similar behavior to `ruff check .`, and `ruff format .`

It seems a little bit unsatisfying to me as well. I did some testing with ruff and this was the behavior:
```
a/
└─ t.py
b/
└─ t.py
```
```
# pyproject.toml
[tool.ruff]
include = ["a/*.py"]
```

```
> ruff check a -v
(...)
[2025-08-10][08:38:29][ruff::commands::check][DEBUG] Checked 1 files in: 756.468µs
All checks passed!

> ruff check b -v
(...)
warning: No Python files found under the given path(s)
All checks passed
```
The same is valid for `ruff format`.

Maybe the current behavior should be kept, what do you think @MichaReiser?

leandrobbraga:
Another interesting thing is that `ruff`'s include is slightly different in behavior from `ty`. From the [site](https://docs.astral.sh/ruff/configuration/#default-inclusions):

```
[tool.ruff]
include = ["pyproject.toml", "src/**/*.py", "scripts/**/*.py"]
```

> Warning
> Paths provided to include must match files. For example, include = ["src"] will fail since it matches a directory.

In `ty`, the configuration would be:
```
[tool.ty.src]
include = ["src"]
```

crates/ty/src/args.rs
MichaReiser: You can use `SystemPath::absolute("/foo//test/.././bar.rs", "/tmp")` to make a path absolute without calling `canonicalize` (which also resolves symlinks which we don't want to do here)

MichaReiser:
The new include and exclude syntax are intentionally different from Ruff's because Ruff's syntax proved to be confusing for users (https://github.com/astral-sh/ruff/issues/6262). Ruff also doesn't support negative exclude patterns.

crates/ty/src/args.rs
MichaReiser: We also need to escape globs, but let's wait until we agreed on the overall design

leandrobbraga: Ok, I just replaced dunce::canonicalize to SystemPath::absolute for now.

MichaReiser:
This is still on my mind. This week has just been a lot busier than I expected and I haven't found the time to think about the desired behavior (you raised very good concerns)

leandrobbraga:
I'm closing this MR since the related issue has also been closed.

MichaReiser:
Thanks for your work here and sorry that I never managed to get back to it.

PR #19811, opened 2025-08-07
[ty] Make `NominalInstanceType` internally wrap an enum

AlexWaygood:
**Stacked on top of https://github.com/astral-sh/ruff/pull/19669; review that PR first!**

---

## Summary

This implements @carljm's suggestion from https://github.com/astral-sh/ruff/pull/19669#issuecomment-3164171196, which appears to buy back almost all the performance and memory regressions from #19669. If we like this approach, I'll merge this PR into that branch shortly before merging that PR. (I'm keeping them as separate PRs for now, for ease of review.)

## Test Plan

All existing tests pass. I don't believe this has any impact on semantics. There is still a small primer report on this PR, but it's now a bit of a mystery to me: I can't reproduce any of the "diagnostics going away" on the target branch of this PR, and I can reproduce all of the "diagnostics being added" on the target branch of this PR. I've seen slightly confusing results from mypy_primer in the past for stacked PRs; I suspect there's a subtle bug in the workflow somewhere that leads to it reporting a diagnostic diff even when there isn't one in cases like this.

PR #19844, opened 2025-08-10
[ty] Add `static-frame` as a walltime benchmark

AlexWaygood:
## Summary

An early version of https://github.com/astral-sh/ruff/pull/19811 dramatically increased the time it took ty to type-check the [static-frame](https://github.com/static-frame/static-frame) codebase: in mypy_primer runs, the execution time went from around 1 minute to around 10 minutes. However, on that early version of #19811, most benchmarks showed big improvements, with only a few benchmarks showing regressions of 1-2%.

The latest version of #19811 no longer exhibits a large increase in execution time when type-checking static-frame, but I find it somewhat concerning that none of our benchmarks flagged the early version; I only accidentally noticed the regression due to the fact that mypy_primer runs were taking much longer to complete! This PR therefore adds static-frame as a walltime benchmark, since it obviously contains Python code patterns that don't currently have good coverage in our ty benchmark suite. The benchmark setup is basically copied from [the project's mypy_primer setup](https://github.com/hauntsaninja/mypy_primer/blob/3cc3235fb80746b03d197fbbce485e7407858c81/mypy_primer/projects.py#L1530-L1537).

## Test Plan

`cargo bench -p ruff_benchmark --bench=ty_walltime`

### DIFF
--- a/crates/ruff_benchmark/benches/ty_walltime.rs
+++ b/crates/ruff_benchmark/benches/ty_walltime.rs
@@ -218,6 +218,24 @@ static TANJUN: std::sync::LazyLock<Benchmark<'static>> = std::sync::LazyLock::ne
     )
 });
 
+static STATIC_FRAME: std::sync::LazyLock<Benchmark<'static>> = std::sync::LazyLock::new(|| {
+    Benchmark::new(
+        RealWorldProject {
+            name: "static-frame",
+            repository: "https://github.com/static-frame/static-frame",
+            commit: "34962b41baca5e7f98f5a758d530bff02748a421",
+            paths: vec![SystemPath::new("static_frame")],
+            // N.B. `arraykit` is installed as a dependency during mypy_primer runs,
+            // but it takes much longer to be installed in a Codspeed run than it does in a mypy_primer run
+            // (seems to be built from source on the Codspeed CI runners for some reason).
+            dependencies: vec!["numpy"],
+            max_dep_date: "2025-08-09",
+            python_version: PythonVersion::PY311,
+        },
+        500,
+    )
+});
+
 #[track_caller]
 fn run_single_threaded(bencher: Bencher, benchmark: &Benchmark) {
     bencher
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Hmm, I'm not sure why (as it's not the case for me locally), but it looks like the `arraykit` dependency of `static-frame` is being built from source in CI when preparing the environment for `static-frame`. `numpy` needs to be present in the build environment for building from source to work (which adds a fair amount of complexity to the benchmark setup). Building `arraykit` from source also means that setting up the environment for static-frame appears to take a fair bit longer than any other walltime benchmark (23.18 seconds). So this might not be worth it.

Having said all that, the walltime-benchmarks CI job is still completing in roughly the same amount of time as the instrumented-benchmarks CI job.

MichaReiser:
Looking at the install and wall time, it suggests that the wall time benchmark now takes 1 minute longer to run, reaching 10 minutes (for both wall and instrumented benchmarks). This is definitely reaching the point where it feels very long if you need the benchmark numbers. Unfortunately, codspeed also doesn't support splitting the benchmarks.

> Having said all that, the walltime-benchmarks CI job is still completing in roughly the same amount of time as the instrumented-benchmarks CI job.

Hehe, and the next instrumentation benchmark addition will not be much longer than the wall time run ;) 

I'll reach out to codspeed to see if there's anything we can do to improve our benchmark runtime. Are there any benchmark that we could remove in favor of `static-frame`?

MichaReiser:
Could we install some of the dependencies as part of the `setup-uv` build step?

AlexWaygood:
It actually looks like the big slowdown on the "bad version" of #19811 is reproducible even without `arraykit` installed, so the simplest solution here is probably to avoid that dependency (though I'm still not sure why the C extension is being built from source in the context of the Codspeed run, since it just downloads a built wheel from PyPI when I install arraykit locally, and mypy_primer also manages to use a built wheel).

I'll push that change and see how long the walltime benchmark takes then.

AlexWaygood:
Okay, on the latest version of this PR, the walltime job is completing in 10m7s overall, which appears ~indistinguishable to the time it takes to complete on `main`

MichaReiser (approved):

PR #19848, opened 2025-08-10
Update goto range for attribute access to only target the attribute

MatthewMckee4:
## Summary

We previously highlighted the whole attribute access for goto definition. This does not align with other language servers (rust-analyzer, python, etc).

More related discussion has been had here https://github.com/astral-sh/ty/issues/169.

But I think for goto definition it seems unintuitive to highlight the whole expression

## Test Plan

Update ty_ide tests

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -557,7 +557,10 @@ impl GotoTarget<'_> {
 impl Ranged for GotoTarget<'_> {
     fn range(&self) -> TextRange {
         match self {
-            GotoTarget::Expression(expression) => expression.range(),
+            GotoTarget::Expression(expression) => match expression {
+                ast::ExprRef::Attribute(attribute) => attribute.attr.range,
+                _ => expression.range(),
+            },
             GotoTarget::FunctionDef(function) => function.name.range,
             GotoTarget::ClassDef(class) => class.name.range,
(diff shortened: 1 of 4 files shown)


MichaReiser:
Thank you. This makes sense to me. I still think it should be consistent with how we highlight attribute access in diagnostics, where other Python LSPs also only highlight the attribute part (and not the object). Which makes sense to me, but it seems we don't have alignment as a team.

AlexWaygood:
I'm not sure I see a strong argument for consistency here -- to me, the range we show for an invalid attribute-access operation seems like a pretty different concept to the range we use as the target for jumping to the original definition/declaration of an attribute. I don't have a very strong opinion, though.

MichaReiser:
I guess that depends on the reasoning why it's preferable to use `self.x` for diagnostics but not for go-to definition. For me, it seems to be similar enough in concept to at least treat it the same (which other type checkers do).

I did a quick check and TypeScript, Pyright, and Rust all highlight the attribute portion only in diagnostics and in goto definition. I still think we should just change our diagnostics. 

Maybe I should stop arguing and just merge this to have another reason why we should change the diagnostic ;)

MichaReiser:
I'm okay going forward as it brings me closer to my goal, changing the range on the diagnostic and it matches my expectation for go to definition ;) It's also something that's easy to revert. Thank you

PR #19850, opened 2025-08-10
[ty] print diagnostics with fully qualified name to disambiguate some cases

leandrobbraga:
There are some situation that we have a confusing diagnostics:

## Class with same name from different modules

```python
import pandas
import polars

df: pandas.DataFrame = polars.DataFrame()
```

This yields the following error:

**Actual:**
error: [invalid-assignment] "Object of type `DataFrame` is not assignable to `DataFrame`"
**Expected**:
error: [invalid-assignment] "Object of type `polars.DataFrame` is not assignable to `pandas.DataFrame`"

## Nested classes

```python
from enum import Enum

class A:
    class B(Enum):
        ACTIVE = "active"
        INACTIVE = "inactive"

class C:
    class B(Enum):
        ACTIVE = "active"
        INACTIVE = "inactive"
```

**Actual**:
error: [invalid-assignment] "Object of type `Literal[B.ACTIVE]` is not assignable to `B`"
**Expected**:
error: [invalid-assignment] "Object of type `Literal[my_module.C.B.ACTIVE]` is not assignable to `my_module.A.B`"

# Solution

In this MR we added an heuristics to detect when to use a fully qualified name:
- There is an invalid assignment and;
- They are two different classes;
- They are not builtin classes (int, ellipsis)

The fully qualified name always includes:
- module name
- nested classes name
- actual class name

# Note

There was no `QualifiedDisplay` so I have to implement it from scratch. I'm very new to the codebase, so I might done things inefficient and I appreciate the feedback. 

Fixes https://github.com/astral-sh/ty/issues/848


leandrobbraga:
It seems there are some false positives, I'll work on those later.

EDIT: it was an important check that I accidentally deleted while refactoring

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1816,16 +1816,57 @@ pub(super) fn report_invalid_assignment(
...
+    db: &'db dyn crate::Db,
+) -> Option<crate::types::class::ClassLiteral<'db>> {
+    match ty {
+        Type::ClassLiteral(class) => Some(class),
+        Type::NominalInstance(instance) => match instance.class {
+            crate::types::class::ClassType::NonGeneric(class) => Some(class),
+            crate::types::class::ClassType::Generic(alias) => Some(alias.origin(db)),
+        },
+        Type::EnumLiteral(enum_literal) => Some(enum_literal.enum_class(db)),
+        _ => None,
```
AlexWaygood: There are a few more variants here where the class can pop up in the display:

```suggestion
        Type::NominalInstance(instance) => type_to_class_literal(Type::from(instance.class)),
        Type::EnumLiteral(enum_literal) => Some(enum_literal.enum_class(db)),
        Type::GenericAlias(alias) => Some(alias.origin(db)),
        Type::ProtocolInstance(ProtocolInstanceType { inner: Protocol::FromClass(class) }) => Some(class),
        Type::TypedDictType(typed_dict) => Some(typed_dict.defining_class(db)),
        Type::SubclassOf(subclass_of) => type_to_class_literal(Type::from(subclass_of.subclass_of().into_class()?)),
        _ => None,
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -59,6 +65,89 @@ impl fmt::Debug for DisplayType<'_> {
...
 
+#[derive(Copy, Clone)]
+pub struct QualifiedDisplayType<'db> {
+    ty: &'db Type<'db>,
+    db: &'db dyn Db,
+}
+
+impl Display for QualifiedDisplayType<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self.ty {
```
AlexWaygood: I'd also add branches for `GenericAlias`, `ProtocolInstance`, `SubclassOf` and `TypedDictType` here

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -59,6 +65,89 @@ impl fmt::Debug for DisplayType<'_> {
...
+            Type::ClassLiteral(literal) => {
+                write!(f, "{}", self.get_qualified_representation(*literal))
```
AlexWaygood: this can be slightly more performant than using the macro (though I'd still use the macro whenever it makes the code cleaner, e.g. in your `EnumLiteral` branch below)

```suggestion
                f.write_str(&self.get_qualified_representation(*literal))
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -59,6 +65,89 @@ impl fmt::Debug for DisplayType<'_> {
...
+            }
+            Type::NominalInstance(instance) => match instance.class {
+                ClassType::NonGeneric(class) => {
+                    write!(f, "{}", self.get_qualified_representation(class))
```
```suggestion
                    f.write_str(&self.get_qualified_representation(class))
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -59,6 +65,89 @@ impl fmt::Debug for DisplayType<'_> {
...
+                }
+                ClassType::Generic(alias) => {
+                    write!(
+                        f,
+                        "{}",
+                        self.get_qualified_representation(alias.origin(self.db))
+                    )
```
```suggestion
                    f.write_str(
                        &self.get_qualified_representation(alias.origin(self.db))
                    )
```

AlexWaygood (commented):
Thank you! A few quick things I noticed:

leandrobbraga:
I had to rebase to solve conflicts, but I think I addressed all the suggestions.

I tried doing some testing to the `<locals of function 'f'>` but I couldn't come up with a working test for this situation.

crates/ty_python_semantic/src/types/display.rs
leandrobbraga: Was this that you had in mind?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -162,3 +162,93 @@ reveal_type(x)  # revealed: Foo
...
+        pass
+
+class C:
+    class B:
+        pass
+
+a: A.B = C.B()  # error: [invalid-assignment] "Object of type `mdtest_snippet.C.B` is not assignable to `mdtest_snippet.A.B`"
+```
+
+### Class from different modules with identical names
```
carljm: Can we also add some tests similar to these where the name that needs qualifying is a nested type, e.g. an assignment fails between `list[mod1.A]` and `list[mod2.A]`. (It's OK if these tests remain TODO in this PR, but I want to record the fact that we still need to do this.)

This should be easy enough to support in the display implementation (especially if we refactor it as I've suggested); it will be a bit harder to detect when types need qualifying. It will probably require a more involved implementation of `are_these_types_ambiguously_named` which we call after knowing that we need to emit an assignment error, which actually descends into generic contexts when the outer types are matching generic classes. (Rather than just extracting the two top-level class types as the current implementation does.)

(There might be another possible approach where `has_relation_to` actually tracks the inner type mismatch and returns this information. It's _possible_ this would mesh nicely with https://github.com/astral-sh/ruff/pull/19838, but I don't think we should pay much extra cost to track information we only will need if emitting an error; it's usually better to do some extra work after we know there's an error. So I'd go with the above approach instead.)

carljm (changes requested):
Thank you for working on this! 

Sorry for coming in on the late side with a new suggestion, but I think we should take a slightly different approach in the `display.rs` part of the implementation. Rather than introducing a totally separate `QualifiedDisplayType` struct, I think we should have a `qualified: bool` attribute on the existing `DisplayType` and `DisplayRepresentation structs, which we can respect whenever we are emitting the name of a class or function. (The current PR doesn't seem to handle functions, and it doesn't need to be done in this PR, but at some point I think we'll want to do this for function names, too.) I think this will make it much easier to share parts of the implementation that are not related to whether or not names are qualified, without code duplication.

The `qualified` flag could be set using a builder pattern, like `ty.display(db).qualified()` vs just `ty.display(db)` for the un-qualified display of a type. Though we may also want some form that allows easily passing a boolean value; this will be convenient for when we want to pass along the "qualified" value to display of a nested inner type.

The recently-merged implementation of a "multiline" display option (which probably needs conflict resolution with this PR anyway) is a good model to follow here.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
leandrobbraga: I used the same approach as `multline` and it ended a much simpler design, I already pushed the commit to this PR.

leandrobbraga: I added a test such as: `dataframes: list[a.DataFrame] = [b.DataFrame()]` and left it as TODO like you mentioned, I can work on it on a different PR.

This assignment is not even failing, probably because of https://github.com/astral-sh/ty/issues/543

carljm: Yes, that's why the assignment doesn't fail. I updated the test to use a function argument instead of an inferred literal container type to avoid this issue -- it's better anyway if the test is more independent from the details of container inference.

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1943,18 +1945,58 @@ pub(super) fn report_invalid_assignment(
...
+            ClassType::Generic(alias) => Some(alias.origin(db)),
+        },
+        Type::TypedDict(typed_dict) => match typed_dict.defining_class() {
+            ClassType::NonGeneric(class) => Some(class),
+            ClassType::Generic(alias) => Some(alias.origin(db)),
+        },
+        Type::SubclassOf(subclass_of) => {
+            type_to_class_literal(Type::from(subclass_of.subclass_of().into_class()?), db)
+        }
+        _ => None,
```
carljm, line 1999: It would be great to have test coverage of all these match arms. I have Claude working on this right now, we'll see how it does.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -103,6 +122,50 @@ struct DisplayRepresentation<'db> {
...
+                    }
+                }
+                ScopeKind::Function => {
+                    if let Some(function_def) = node.as_function(&module_ast) {
+                        name_parts.push(format!(
+                            "<locals of function '{}'>",
+                            function_def.name.as_str()
+                        ));
+                    }
+                }
```
carljm, line 154: I don't think we have test coverage of this case. It would require trying to assign a class within the same function where it's defined, as there's no way to reference a class defined inside of a function outside that function. I'll see if Claude can add this next.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -120,13 +183,27 @@ impl Display for DisplayRepresentation<'_> {
                         .expect("Specialization::tuple() should always return `Some()` for `KnownClass::Tuple`")
                         .display_with(self.db, self.settings)
                         .fmt(f),
-                    (ClassType::NonGeneric(class), _) => f.write_str(class.name(self.db)),
+                    (ClassType::NonGeneric(class), _) => {
+                        if self.settings.qualified {
+                            f.write_fmt(format_args!("{}.{}", self.class_parents(class).join("."), class.name(self.db)))
+                        } else {
+                            f.write_str(class.name(self.db))
+                        }
```
carljm: It seems like we do this in several places -- we should define a function for it.

carljm (commented):
Looking good! I think we should increase our test coverage a bit, and try to reduce code boilerplate.

I think ultimately in order to support nested types we will probably need to change the settings from just having a `qualified` boolean, to instead having a vector of types whose names should be qualified, so that we can identify the right types to qualify even in nested display. But this can be a separate PR.

Issue #8074, opened 2023-10-19
Allow safe fixes to be added in patch releases


zanieb:
Adding new fixes isn't really a breaking change necessitating a minor version bump, however we have it categorized as such in our [current versioning policy](https://docs.astral.sh/ruff/versioning/#version-changes). Somewhat asymmetrically, we allow unsafe fixes to be added in patch releases. The idea was originally that unsafe fixes required opt-in so it was okay to add them in patch releases. In general, I think new fixes should be allowed in any release.

The downside here is that users may need to change their configuration to include new fixes as unfixable on patch releases if they do not want the fix. This goes against the intent to have patch releases be free of configuration changes. However, these fixes are unlikely to be applied in a context without manual review.

Avasam:
There's also the risk that a "safe" fix is actually unsafe and that wasn't caught until out of preview.

charliermarsh:
I'd like to elevate `SIM110` and `SIM111`. I'll probably add more over time.

Avasam:
> There's also the risk that a "safe" fix is actually unsafe and that wasn't caught until out of preview.

#9156 is a perfect example of what I mean here.

zanieb:
I don't think we have a clear consensus that this is worth changing yet. Removing this from our v0.2.0 milestone.

charliermarsh:
I still want this :sob:

MichaReiser:
To me the discussion should be whether we allow adding fixes in releases in general:
IMO, the bar for adding a manual fix in a patch release is lower than that for adding a safe fix or an unsafe fix. I'm also leaning towards the bar for adding an unsafe fix being lower than that for adding a safe fix.

AlexWaygood:
> There's also the risk that a "safe" fix is actually unsafe and that wasn't caught until out of preview.

To @Avasam's point here -- the fixes for RUF022 and RUF023 were both marked as "safe", and were included in the most recent patch release -- but in fact, they had the potential to introduce syntax errors in some unusual edge cases: https://github.com/astral-sh/ruff/issues/8402#issuecomment-1916203707; https://github.com/astral-sh/ruff/issues/8402#issuecomment-1916213693. (The rules themselves were still in preview here, but this is a good example of the way that fixes we think are safe might not actually be safe.)

charliermarsh:
A few reactions (and not a decision):

[...]

PR #19851, opened 2025-08-10
[`flake8-simplify`] Implement fix for `maxsplit` without separator (`SIM905`)

RazerM:
**Stacked on top of #19849; diff will include that PR until it is merged.**

---

## Summary

As part of #19849, I noticed this fix could be implemented.

## Test Plan

Tests added based on CPython behaviour.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM905.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM905.py
@@ -166,3 +166,7 @@ r"""first
 print("S\x1cP\x1dL\x1eI\x1fT".split())
 print("\x1c\x1d\x1e\x1f>".split(maxsplit=0))
 print("<\x1c\x1d\x1e\x1f".rsplit(maxsplit=0))
+
+# leading/trailing whitespace should not count towards maxsplit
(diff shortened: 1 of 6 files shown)


MichaReiser:
Could you rebase your PR now that https://github.com/astral-sh/ruff/pull/19849 is merged. It will make it easier for us to review your changes. Thank you

ntBre:
The more I think about this, the more I'm leaning towards using preview to be safe. @MichaReiser do you feel strongly one way or the other? I only found 99 results in a [code search](https://github.com/search?q=%2F%5B%27%22%5D%5C.r%3Fsplit%5C%28maxsplit%3D%5B1-9%5D%2B%5C%29%2F+language%3APython&type=code), so maybe it's quite a niche pattern anyway.

The preview-gate itself would be easy enough, but we also need a new function in `preview.rs` and to add the test case to the flake8-simplify `preview_rules` function, if we go that route.

MichaReiser:
I think we have to, according to our versioning policy. See https://github.com/astral-sh/ruff/issues/8074 why we might want to change that.

ntBre:
Sounds good, @RazerM do you mind gating this behind preview?

RazerM:
@ntBre I think I've done it correctly, but let me know.

ntBre (approved):
Excellent, thank you!

PR #19862, opened 2025-08-11
[ty] Cache dynamic callable type

MichaReiser:
## Summary

Cache the `CallableType` for `Type::dynamic` to reduce the congestion around salsa's `intern_id` lock. 

https://github.com/astral-sh/ty/issues/968

This reduces check time on a large project from 25s to 18s on my mac mini. The memory increase should be proportional to the amount of dynamic types (which should be fixed once we eliminated all Todo types)

## Test Plan

`cargo test`


crates/ty_python_semantic/src/types.rs
```diff
@@ -6403,11 +6402,17 @@ pub enum DynamicType {
...
+impl<'db> DynamicType {
     #[expect(clippy::unused_self)]
     fn normalized(self) -> Self {
         Self::Any
     }
+
+    #[salsa::tracked]
+    fn into_callable(self, db: &'db dyn Db, parameters: Parameters<'db>) -> Type<'db> {
+        CallableType::single(db, Signature::new(parameters, Some(Type::Dynamic(self))))
+    }
```
AlexWaygood: Caching the method definitely makes sense if it improves performance this much, but I think I'd prefer for this to be a method on `CallableType` rather than a method on `DynamicType`. I'm not sure it makes sense to think of this as transforming a `DynamicType` "into a `CallableType`": what we're really doing is creating a `CallableType` that returns a certain `DynamicType`.

Also, could we add a comment saying why we specifically cache the constructor if we're constructing a `CallableType` that returns a `DynamicType`, rather than generally caching the `CallableType::single()` method?

AlexWaygood (commented):
Nice!

MichaReiser:
Okay, my latest version undos the entire performance improvement because calling the cached query requires interning on its own, lol

crates/ty_python_semantic/src/types.rs
```diff
@@ -6403,7 +6410,7 @@ pub enum DynamicType {
     TodoUnpack,
 }
 
-impl DynamicType {
+impl<'db> DynamicType {
```
```suggestion
impl DynamicType {
```

MichaReiser:
This does help performance but there's still a congested lock because non-argument queries in salsa also require interning (it's just that the lock is now sharded by dynamic type which improves performance somewhat). Addressing the todo's will also mean that the locks become more congested again. So maybe this is just not worth it (But the perf improvement is massive, ~40 -> 30s or even less on my linux machine)

crates/ty_python_semantic/src/types.rs
```diff
@@ -8046,6 +8048,105 @@ impl<'db> CallableType<'db> {
         ))
     }
 
+    /// Create a callable type [`Parameters::gradual`] with a dynamic return type.
+    ///
+    /// This is a hack to work around a congested lock in Salsa's interning.
+    /// See <https://github.com/astral-sh/ty/issues/968>.
+    ///
+    /// Each of the variant uses their own query to achieve an extra degree of sharding (Salsa
```
AlexWaygood, line 8056:
```suggestion
    /// Each variant uses its own query to achieve an extra degree of sharding (Salsa
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -8046,6 +8048,105 @@ impl<'db> CallableType<'db> {
         ))
     }
 
+    /// Create a callable type [`Parameters::gradual`] with a dynamic return type.
```
AlexWaygood, line 8051:
```suggestion
    /// Create a callable type with ``*args: Any, **kwargs: Any`` as the parameter spec
    /// and a dynamic return type.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -8046,6 +8048,105 @@ impl<'db> CallableType<'db> {
...
+                        Signature::new(
+                            Parameters::unknown(),
+                            Some(Type::Dynamic(DynamicType::TodoUnpack)),
+                        ),
+                    )
+                }
+                unpack(db)
+            }
+        }
+    }
```
AlexWaygood, line 8148: I think these should all use `Parameters::gradual_form()` rather than `Parameters::unknown()`, if we want to keep the semantics exactly as they are on `main`. The former creates a `*args: Any, **kwargs: Any` parameter spec, the second creates a `*args: Unknown, **kwargs: Unknown` parameter spec. Practically speaking these are usually treated the same by ty, but `Unknown` normalizes to `Any`, and it could also have an effect on diagnostics messages.

MichaReiser:
Starring more at this. The sharding in Salsa isn't working because all threads look up the same few `CallableType`,s which, because they're the same, are all stored in the same shard. 

It might be worth to use an `RwLock` in the interning table. But I don't know how easily this is possible, due to the LRU

MichaReiser:
@carljm had a much better idea. I'll open a new PR

carljm:
I assume this won't show much if any perf impact after the PR that special-cases Dynamic in reachability constraint evaluation?

MichaReiser:
> I assume this won't show much if any perf impact after the PR that special-cases Dynamic in reachability constraint evaluation?

No, I wanted to close this PR (I thought I did)

PR #19867, opened 2025-08-11
[ty] Short circuit `ReachabilityConstraints::analyze_single` for dynamic types

MichaReiser:
## Summary

Fixes https://github.com/astral-sh/ty/issues/968

This reduces the wall time for a large project from ~24s to ~19s. I plan to do another perf recording to verify that the lock congestion is entirely gone but I first need to walk the dog :)

See inline comment for more details.

## Test Plan

`cargo nextest`

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
+++ b/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
@@ -823,7 +823,7 @@ impl ReachabilityConstraints {
                 // doesn't help much.
                 // See <https://github.com/astral-sh/ty/issues/968>.
                 if matches!(ty, Type::Dynamic(_)) {
-                    return Truthiness::AlwaysTrue.negate_if(!predicate.is_positive);
+                    return Truthiness::AlwaysFalse.negate_if(!predicate.is_positive);
                 }
 
                 let overloads_iterator =
--- a/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
+++ b/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
@@ -815,6 +815,17 @@ impl ReachabilityConstraints {
                 // very large in number, since we add them on all statement level function calls.
                 let ty = infer_expression_type(db, callable);
 
+                // Short-circuit for well known types that are known not to return `Never` when called.
+                // Without the short-circuit, we've seen that threads keep blocking each other
+                // because they all try to acquire Salsa's `CallableType` lock that ensures each type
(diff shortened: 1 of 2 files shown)


MichaReiser:
Nice, this is even a win for non concurrent runs. All credit goes to @carljm for the much better fix than what I came up with

dcreager (approved):

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
```diff
@@ -815,6 +815,17 @@ impl ReachabilityConstraints {
...
 
+                // Short-circuit for well known types that are known not to return `Never` when called.
+                // Without the short-circuit, we've seen that threads keep blocking each other
+                // because they all try to acquire Salsa's `CallableType` lock that ensures each type
+                // is only interned once. The lock is so heavily congested because there are only
+                // very few dynamic types, in which case Salsa's sharding the locks by value
+                // doesn't help much.
+                // See <https://github.com/astral-sh/ty/issues/968>.
+                if matches!(ty, Type::Dynamic(_)) {
+                    return Truthiness::AlwaysTrue.negate_if(!predicate.is_positive);
```
carljm: Sorry, I gave you the wrong info here, I had the sense of this constraint reversed in my head. This needs to return `Truthiness::AlwaysFalse`, not `Truthiness::AlwaysTrue`! I suspect the mypy-primer run is taking so long because there are now tons more bogus diagnostics 😆 

The fact that no test failed on this change also suggests we need an explicit mdtest like this:

```py
from typing import Never, Any

def f(func: Any) -> Never:
    func()
```

Where an `invalid-return-type` is expected (because the call to `func()` is _not_ terminal). That test fails on this PR currently.
```suggestion
                    return Truthiness::AlwaysFalse.negate_if(!predicate.is_positive);
```

carljm (changes requested):
Perf results look amazing! Hopefully they persist once we get the logic right 😆 (I expect they will)

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
MichaReiser: Don't worry, it does. The most congested lock is now for `IntersectionType`:

image

And the wall clock time on the large project on my linux machine goes from 42s .. to 16s which is rather ridiculous

MichaReiser:
Okay, the single threaded benchmarks don't improve by that much anymore. But it's still a huge improvement for multi threaded workloads (with many cores... not what github actions provides)

carljm (approved):
Awesome!

Issue #18198, opened 2025-05-19
Speed up caching


ntBre:
It looks like the caching slowdown is not recent, it's been in a pretty steady decline since that 6x number was recorded. These tags aren't exact, just the tag before the commit I was bisecting:

| Tag     | Speedup |
|---------|---------|
| 0.11.10 | 1.3     |
| 0.9.2   | 1.4     |
| 0.7.3   | 1.7     |
| 0.5.7   | 1.7     |
| 0.5.0   | 1.7     |
| 0.4.5   | 2.6     |
| 0.3.0   | 3.4     |
| 0.0.290 | 3.6     |
| 0.0.240 | 6.1     |

`Speedup` is the relative speedup reading from cache, as reported in the hyperfine summary above.

I couldn't actually build 0.0.240 because of a git dep on libCST that doesn't exist anymore, so that number is from `CONTRIBUTING.md`. 0.0.290 is the earliest version I could build easily because it added the `ruff_linter` crate, which is where I had my CPython clone.

I added a panic after the caching early return in `lint_path` and can confirm that the cache is working too. (I also tried adding a  panic _before_ the early return, right at the top of `lint_path` and it definitely panicked)

_Originally posted by @ntBre in https://github.com/astral-sh/ruff/issues/18051#issuecomment-2891351620_

ntBre:
That's a good idea, I can definitely extend the table. From what I remember while running these, the absolute times were growing too, but it's better to be sure.

ntBre:
This _was_ interesting. These numbers are from my laptop and retrieving the github releases instead of building from source, so the exact ratios are a little different, but the trend is the same.

| Version | Cold | Cache | Speedup |
|---------|------|-------|---------|
| 0.1.0 | 298.5 | 73.3 | 4.07 |
| 0.2.0 | 309.0 | 72.9 | 4.24 |
| 0.3.0 | 264.2 | 73.0 | 3.62 |
| 0.4.0 | 175.9 | 73.2 | 2.40 |
| 0.5.0 | 237.6 | 125.0 | 1.90 |
| 0.6.0 | 224.4 | 122.6 | 1.83 |
| 0.7.0 | 224.8 | 122.2 | 1.84 |
| 0.8.0 | 207.8 | 102.6 | 2.02 |
| 0.9.0 | 211.7 | 102.4 | 2.07 |
| 0.10.0 | 293.9 | 186.4 | 1.58 |
[...]

PR #19869, opened 2025-08-11
Don't cache files with diagnostics

ntBre:
Summary
--

To take advantage of the new diagnostics, we need to update our caching model to include all of the information supported by `ruff_db`'s diagnostic type. Instead of trying to serialize all of this information, Micha suggested simply not caching files with diagnostics, like we already do for files with syntax errors. This PR is an attempt at that approach.

This has the added benefit of trimming down our `Rule` derives since this was the last place the `FromStr`/`strum_macros::EnumString` implementation was used, as well as the (de)serialization macros and `CacheKey`.

Test Plan
--

Existing tests, with their input updated not to include a diagnostic, plus a new test showing that files with lint diagnostics are not cached.

Benchmarks
--

In addition to tests, we wanted to check that this doesn't degrade performance too much. I posted part of this new analysis in https://github.com/astral-sh/ruff/issues/18198#issuecomment-3175048672, but I'll duplicate it here. In short, there's not much difference between `main` and this branch for projects with few diagnostics (`home-assistant`, `airflow`), as expected. The difference for projects with many diagnostics (`cpython`) is quite a bit bigger (~300 ms vs ~220 ms), but most projects that run ruff regularly are likely to have very few diagnostics, so this may not be a problem practically. 

I guess GitHub isn't really rendering this as I intended, but the extra separator line is meant to separate the benchmarks on `main` (above the line) from this branch (below the line).

| Command                                                       | Mean [ms] | Min [ms] | Max [ms] |
|:--------------------------------------------------------------|----------:|---------:|---------:|
| `ruff check cpython --no-cache --isolated --exit-zero`        |     322.0 |    317.5 |    326.2 |
| `ruff check cpython --isolated --exit-zero`                   |     217.3 |    209.8 |    237.9 |
| `ruff check home-assistant --no-cache --isolated --exit-zero` |     279.5 |    277.0 |    283.6 |
| `ruff check home-assistant --isolated --exit-zero`            |      37.2 |     35.7 |     40.6 |
| `ruff check airflow --no-cache --isolated --exit-zero`        |     133.1 |    130.4 |    146.4 |
| `ruff check airflow --isolated --exit-zero`                   |      34.7 |     32.9 |     41.6 |
|:--------------------------------------------------------------|----------:|---------:|---------:|
| `ruff check cpython --no-cache --isolated --exit-zero`        |     330.1 |    324.5 |    333.6 |
| `ruff check cpython --isolated --exit-zero`                   |     309.2 |    306.1 |    314.7 |
| `ruff check home-assistant --no-cache --isolated --exit-zero` |     288.6 |    279.4 |    302.3 |
| `ruff check home-assistant --isolated --exit-zero`            |      39.8 |     36.9 |     42.4 |
| `ruff check airflow --no-cache --isolated --exit-zero`        |     134.5 |    131.3 |    140.6 |
| `ruff check airflow --isolated --exit-zero`                   |      39.1 |     37.2 |     44.3 |

I had Claude adapt one of the [scripts](https://github.com/sharkdp/hyperfine/blob/master/scripts/plot_whisker.py) from the hyperfine repo to make this plot, so it's not quite perfect, but maybe it's still useful. The table is probably more reliable for close comparisons. I'll put more details about the benchmarks below for the sake of future reproducibility.

image

<details><summary>Benchmark details</summary>
<p>

The versions of each project:
- CPython: 6322edd260e8cad4b09636e05ddfb794a96a0451, the 3.10 branch from the contributing docs
- `home-assistant`: 5585376b406f099fb29a970b160877b57e5efcb0
- `airflow`: 29a1cb0cfde9d99b1774571688ed86cb60123896

The last two are just the main branches at the time I cloned the repos.

I don't think our Ruff config should be applied since I used `--isolated`, but these are cloned into my copy of Ruff at `crates/ruff_linter/resources/test`, and I trimmed the `./target/release/` prefix from each of the commands, but these are builds of Ruff in release mode.

And here's the script with the `hyperfine` invocation:

```shell
#!/bin/bash

cargo build --release --bin ruff

# git clone --depth 1 https://github.com/home-assistant/core crates/ruff_linter/resources/test/home-assistant
# git clone --depth 1 https://github.com/apache/airflow crates/ruff_linter/resources/test/airflow

bin=./target/release/ruff
resources=./crates/ruff_linter/resources/test
cpython=$resources/cpython
home_assistant=$resources/home-assistant
airflow=$resources/airflow

base=${1:-bench}

hyperfine --warmup 10 --export-json $base.json --export-markdown $base.md \
		  "$bin check $cpython --no-cache --isolated --exit-zero" \
		  "$bin check $cpython --isolated --exit-zero" \
		  "$bin check $home_assistant --no-cache --isolated --exit-zero" \
		  "$bin check $home_assistant --isolated --exit-zero" \
		  "$bin check $airflow --no-cache --isolated --exit-zero" \
		  "$bin check $airflow --isolated --exit-zero"
```

I ran this once on `main` (`baseline` in the graph, top half of the table) and once on this branch (`nocache` and bottom of the table).

</p>
</details>

### DIFF
--- a/crates/ruff/src/cache.rs
+++ b/crates/ruff/src/cache.rs
@@ -525,7 +525,7 @@ mod tests {
         assert_eq!(cache.changes.lock().unwrap().len(), 0);
 
         let mut paths = Vec::new();
-        let mut errors = Vec::new();
+        let mut paths_with_diagnostics = Vec::new();
         let mut expected_diagnostics = Diagnostics::default();
         for entry in fs::read_dir(&package_root).unwrap() {
             let entry = entry.unwrap();
@@ -560,7 +560,7 @@ mod tests {
                 )
                 .unwrap();
                 if !diagnostics.inner.is_empty() {
-                    errors.push(path.clone());
+                    paths_with_diagnostics.push(path.clone());
                 }
                 paths.push(path);
                 expected_diagnostics += diagnostics;
@@ -573,10 +573,10 @@ mod tests {
         let cache = Cache::open(package_root.clone(), &settings);
         assert_ne!(cache.package.files.len(), 0);
 
-        errors.sort();
+        paths_with_diagnostics.sort();
 
         for path in &paths {
-            if errors.binary_search(path).is_ok() {
+            if paths_with_diagnostics.binary_search(path).is_ok() {
                 continue; // We don't cache files with diagnostics.
             }
 
--- a/crates/ruff/src/cache.rs
+++ b/crates/ruff/src/cache.rs
@@ -289,8 +289,8 @@ impl Cache {
         });
     }
 
-    pub(crate) fn set_linted(&self, path: RelativePathBuf, key: &FileCacheKey) {
-        self.update(path, key, ChangeData::Linted);
+    pub(crate) fn set_linted(&self, path: RelativePathBuf, key: &FileCacheKey, yes: bool) {
+        self.update(path, key, ChangeData::Linted(yes));
     }
 
     pub(crate) fn set_formatted(&self, path: RelativePathBuf, key: &FileCacheKey) {
@@ -456,15 +456,15 @@ struct Change {
 
 #[derive(Debug)]
 enum ChangeData {
-    Linted,
+    Linted(bool),
     Formatted,
 }
 
 impl ChangeData {
     fn apply(self, data: &mut FileCacheData) {
         match self {
-            ChangeData::Linted => {
-                data.linted = true;
+            ChangeData::Linted(yes) => {
+                data.linted = yes;
             }
             ChangeData::Formatted => {
                 data.formatted = true;
@@ -496,7 +496,7 @@ mod tests {
     use ruff_python_ast::{PySourceType, PythonVersion};
     use ruff_workspace::Settings;
 
-    use crate::cache::{self, FileCache, FileCacheData, FileCacheKey};
+    use crate::cache::{self, ChangeData, FileCache, FileCacheData, FileCacheKey};
     use crate::cache::{Cache, RelativePathBuf};
     use crate::commands::format::{FormatCommandError, FormatMode, FormatResult, format_path};
     use crate::diagnostics::{Diagnostics, lint_path};
@@ -676,8 +676,17 @@ mod tests {
         assert_eq!(results.inner.len(), 1, "Expected one F822 diagnostic");
         assert_eq!(
             cache.changes.lock().unwrap().len(),
-            0,
-            "Files with diagnostics should not be cached"
+            1,
+            "Files with diagnostics still trigger change events"
+        );
+        assert!(
+            cache
+                .changes
+                .lock()
+                .unwrap()
+                .last()
+                .is_some_and(|change| matches!(change.new_data, ChangeData::Linted(false))),
+            "Files with diagnostics are marked as unlinted"
         );
 
         cache.persist().unwrap();
--- a/crates/ruff/src/cache.rs
+++ b/crates/ruff/src/cache.rs
@@ -20,7 +20,6 @@ use ruff_cache::{CacheKey, CacheKeyHasher};
 use ruff_linter::package::PackageRoot;
 use ruff_linter::{VERSION, warn_user};
 use ruff_macros::CacheKey;
-use ruff_notebook::NotebookIndex;
 use ruff_workspace::Settings;
 use ruff_workspace::resolver::Resolver;
 
@@ -292,13 +291,8 @@ impl Cache {
         });
     }
 
-    pub(crate) fn update_lint(
-        &self,
-        path: RelativePathBuf,
-        key: &FileCacheKey,
-        data: LintCacheData,
-    ) {
-        self.update(path, key, ChangeData::Lint(data));
+    pub(crate) fn set_linted(&self, path: RelativePathBuf, key: &FileCacheKey) {
+        self.update(path, key, ChangeData::Linted);
     }
 
     pub(crate) fn set_formatted(&self, path: RelativePathBuf, key: &FileCacheKey) {
@@ -333,23 +327,15 @@ pub(crate) struct FileCache {
 }
 
 impl FileCache {
-    /// Convert the file cache into `Diagnostics`, using `path` as file name.
-    pub(crate) fn to_diagnostics(&self, path: &Path) -> Option<Diagnostics> {
-        self.data.lint.as_ref().map(|lint| {
-            let diagnostics = Vec::new();
-            let notebook_indexes = if let Some(notebook_index) = lint.notebook_index.as_ref() {
-                FxHashMap::from_iter([(path.to_string_lossy().to_string(), notebook_index.clone())])
-            } else {
-                FxHashMap::default()
-            };
-            Diagnostics::new(diagnostics, notebook_indexes)
-        })
+    /// Convert the file cache into an empty `Diagnostics`, if it has been linted.
+    pub(crate) fn to_diagnostics(&self) -> Option<Diagnostics> {
+        self.data.linted.then(Diagnostics::default)
     }
 }
 
 #[derive(Debug, Default, bincode::Decode, bincode::Encode)]
 struct FileCacheData {
-    lint: Option<LintCacheData>,
+    linted: bool,
     formatted: bool,
 }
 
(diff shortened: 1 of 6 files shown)


MichaReiser:
I don't mind the regression for CPython too much. It has a vast amount of diagnostics which should be uncommon. What I'm more surprised by is the regression for homeassitant and airflow. Do they have any diagnostics, is it just noise, or do you know where the regression is comming from?

crates/ruff/src/cache.rs
```diff
@@ -412,86 +387,17 @@ pub(crate) fn init(path: &Path) -> Result<()> {
...
-    pub(super) messages: Vec<CacheMessage>,
-    /// Source code of the file.
-    ///
-    /// # Notes
-    ///
-    /// This will be empty if `messages` is empty.
-    pub(super) source: String,
     /// Notebook index if this file is a Jupyter Notebook.
     #[bincode(with_serde)]
     pub(super) notebook_index: Option<NotebookIndex>,
```
MichaReiser: What's preventing us from removing the Notebook index too? (in which case `FileCacheData` only needs to store whether a file had any violations)

ntBre:
> I don't mind the regression for CPython too much. It has a vast amount of diagnostics which should be uncommon. What I'm more surprised by is the regression for homeassitant and airflow. Do they have any diagnostics, is it just noise, or do you know where the regression is comming from?

I _think_ it's mostly noise (you can see CPython without a cache was 8 ms slower in the second run), but they both have some diagnostics too. `home-assistant` has 21 diagnostics, and `airflow` has 143, compared to 15,577 in `cpython`.

I'll take another look at the notebook index too.

crates/ruff/src/cache.rs
ntBre: Ah of course, thanks. It kind of looked used, but obviously we won't actually look up a notebook in this map if the file has no diagnostics. I replaced `LintCacheData` with a `linted` field to mirror the `formatted` field.

I also updated `lint_path` (and `lint_stdin`) to avoid constructing a `NotebookIndex` that wouldn't be cached (e2bc85e56a). Otherwise the `same_results` cache tests were failing with differences in the `NotebookIndex` field. At first I just fixed this in the test, but it seemed like a small optimization in the real code anyway.

crates/ruff/src/cache.rs
```diff
@@ -339,42 +327,15 @@ pub(crate) struct FileCache {
...
-            };
-            let notebook_indexes = if let Some(notebook_index) = lint.notebook_index.as_ref() {
-                FxHashMap::from_iter([(path.to_string_lossy().to_string(), notebook_index.clone())])
-            } else {
-                FxHashMap::default()
-            };
-            Diagnostics::new(diagnostics, notebook_indexes)
-        })
+    /// Convert the file cache into an empty `Diagnostics`, if it has been linted.
+    pub(crate) fn to_diagnostics(&self) -> Option<Diagnostics> {
```
MichaReiser: This feels a bit overkill, given that it always returns an empty `Diagnostics` or `None`. I think I'd change this to a `bool` which should also allow simplifying the call site, e.g. the following can be removed entirely:


```
// `FixMode::Generate` and `FixMode::Diff` rely on side-effects (writing to disk,
// and writing the diff to stdout, respectively). If a file has diagnostics, we
// need to avoid reading from and writing to the cache in these modes.
```

crates/ruff/src/cache.rs
```diff
@@ -695,11 +573,11 @@ mod tests {
...
         assert_ne!(cache.package.files.len(), 0);
 
-        parse_errors.sort();
+        errors.sort();
 
         for path in &paths {
-            if parse_errors.binary_search(path).is_ok() {
-                continue; // We don't cache parsing errors.
+            if errors.binary_search(path).is_ok() {
+                continue; // We don't cache files with diagnostics.
```
MichaReiser: Nit: Maybe rename `errors` to `paths_with_diagnostics` to make it clear that this isn't storing the errors, but is a set of paths

crates/ruff/src/diagnostics.rs
```diff
@@ -326,8 +326,8 @@ pub(crate) fn lint_path(
     let diagnostics = result.diagnostics;
 
     if let Some((cache, relative_path, key)) = caching {
-        // We don't cache parsing errors.
-        if !has_error {
+        // We don't cache errors.
```
MichaReiser:
```suggestion
        // We don't cache files with diagnostics.
```

I think this allows us to now remove `result.has_syntax_errors` all together. It's only used in the benchmarking code but we can just get that information from the parsed module instead.

crates/ruff/src/diagnostics.rs
```diff
@@ -337,22 +337,22 @@ pub(crate) fn lint_path(
...
             } {
-                cache.update_lint(
-                    relative_path.to_owned(),
-                    &key,
-                    LintCacheData::from_diagnostics(
-                        &diagnostics,
-                        transformed.as_ipy_notebook().map(Notebook::index).cloned(),
-                    ),
-                );
+                cache.set_linted(relative_path.to_owned(), &key);
```
MichaReiser: I think this is fine, given that this is pre-existing code. But which code path sets `linted` to false if a file has diagnostics (how do we clear the state?)

crates/ruff/src/diagnostics.rs
```diff
@@ -337,22 +337,22 @@ pub(crate) fn lint_path(
...
             }
         }
     }
 
-    let notebook_indexes = if let SourceKind::IpyNotebook(notebook) = transformed {
-        FxHashMap::from_iter([(path.to_string_lossy().to_string(), notebook.into_index())])
-    } else {
+    // Avoid constructing a map when there are no diagnostics. The index is only used for rendering
+    // diagnostics anyway. This mirrors our caching behavior, which does not preserve an empty index
+    // either.
```
MichaReiser: This feels a bit subtle and it might not be evident for callers that `notebook_indexes` is empty if there are no diagnostics. We should document this constraint in `Diagnostics` or consider moving it into a constructor within `Diagnostics`.

crates/ruff/src/diagnostics.rs
ntBre: I think the current assertion in the benchmark is a bit stronger than what we get from the parsed module because it also includes semantic syntax errors, but I'm not sure that was intentional. The comment on the assertion just says "parse errors," so it's probably fine to assert only on those.

ntBre: Do we need some kind of assertion on the benchmark result to make sure it doesn't get optimized out? If we assert on the parsed module, `result` becomes unused.

We can at least trim down the `LinterResult` methods to `has_no_syntax_errors` and `has_invalid_syntax`. Those are both only used in one place:

https://github.com/astral-sh/ruff/blob/ad28b80f967282aaaaaff59c722faf087125ffc1/crates/ruff_benchmark/benches/linter.rs#L91-L92

(after removing the `!` and using `has_no_syntax_errors`)

https://github.com/astral-sh/ruff/blob/ad28b80f967282aaaaaff59c722faf087125ffc1/crates/ruff_server/src/fix.rs#L78-L81

MichaReiser: >  The comment on the assertion just says "parse errors," so it's probably fine to assert only on those.

Yeah, I think that's fine. The reason I added the assertion originally because we used to download the test files from Github and I once copied the wrong link (not the raw.github but the regular github.com link). The result was that we downloaded and benchmarked the HTML files instead of the Python files... whoops :) 

That's why this assertion is there

MichaReiser: I think you could return `LintResult` and criterion will automatically wrap it in a hint::black_box? (and also drop it)

crates/ruff/src/diagnostics.rs
ntBre: Ah, I think you're right. I guess this is different from `formatted` since that always gets set to `true` whether the file was reformatted or already formatted. I reworked this section a bit to set `linted` to `diagnostics.is_empty() && use_fixes` where `use_fixes` is the result of the `match` above this.

crates/ruff/src/diagnostics.rs
ntBre: We could also just do this filtering step in the test (or adjust the assertion not to check the notebook index), if you prefer. This didn't show any difference in the benchmarks, so it's not really helpful as an optimization, especially if it's confusing.

ntBre: I did this in 13d7cf9197 but happy to revert and document/update this code instead if you prefer it.

ntBre: I'm realizing now that we probably _also_ need the docs since it still affects `Diagnostics` created from the cache...

ntBre:
`internal` might not be the right label here, but I don't really want it in the release notes. I certainly hope it doesn't cause any user-noticeable slowdown.

PR #19870, opened 2025-08-11
[ty] Fix tool name is None when no ty path is given in ty_benchmark

nguu0123:
## Summary

When running the ty_benchmark, I found out that the Ty Tool name is None when no ty_path is given as str(None)='None'
image

## Test Plan
Minor fix, tested local
image


carljm (approved):

carljm:
Thank you!

PR #19790, opened 2025-08-06
[ty] Track heap usage of salsa structs

ibraheemdev:
## Summary

This is the last big remaining item for our Salsa memory reports.

PR #19872, opened 2025-08-11
[ty] Reduce memory usage of `TupleSpec` and `TupleType`

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/956. Use boxed slices rather than Vecs where possible for types interned in the salsa database, to save on memory.

## Test Plan

Existing tests

### DIFF
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -114,7 +114,7 @@ use crate::types::generics::{GenericContext, bind_typevar};
 use crate::types::instance::SliceLiteral;
 use crate::types::mro::MroErrorKind;
 use crate::types::signatures::{CallableSignature, Signature};
-use crate::types::tuple::{Tuple, TupleSpec, TupleType};
+use crate::types::tuple::{Tuple, TupleSpec, TupleSpecBuilder, TupleType};
 use crate::types::unpacker::{UnpackResult, Unpacker};
 use crate::types::{
     CallDunderError, CallableType, ClassLiteral, ClassType, DataclassParams, DynamicType,
@@ -10042,7 +10042,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     return result;
                 }
 
-                let mut element_types = TupleSpec::with_capacity(elements.len());
+                let mut element_types = TupleSpecBuilder::with_capacity(elements.len());
 
                 // Whether to infer `Todo` for the whole tuple
                 // (see docstring for `element_could_alter_type_of_whole_tuple`)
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Not sure if this is worth it — CI reports no significant memory-usage reductions in the [memory.txt projects](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/primer/memory.txt), and codspeed reports 1% regressions on several benchmarks.

Tomorrow I'll try getting some memory-usage reports for projects that we know have some complicated tuple types in them, such as colour-science and static-frame. Opening up for review now anyway, to see what folks think of the approach!

MichaReiser:
You probably want to wait for https://github.com/astral-sh/ruff/pull/19790 (or rebase on top) to get a full picture. The memory report today only accounts for queries (heap and stack size) and the stack size of tracked struct, but it doesn't account for the heap size of tracked structs. That means, the only memory reduction you'd see is from reducing the stack memory from  48 + `stack_size::<T>` to 32 + `stack_size::<T>`. That on its own might still be significant but only if we create a large number of tuple specs (and intern them)

MichaReiser:
This is the full memory report comparison from running on a large project. You can see how the tuple type's stack size has become much smaller. It's just not significant in overall terms because there aren't many enough (or put differently, some other queries use up so much space that the tuple type improvement is negligible). I suggest re-running the analysis with #19790 

https://www.diffchecker.com/ZkHpFHCN/

Edit: The reason why we have different counts for some ingredients is because ty still panics when analysing some files and that leads to non-deterministic results.

MichaReiser:
I just ran main with the salsa struct memory usage information and the total memory usage for tuples is only:

```
`TupleType`                                        metadata=6.65MB   fields=13.88MB  count=118827
```

So i don't think this will lead to a very significant improvement

AlexWaygood:
I tried running memory reports on colour-science and static-frame, which were codebases I thought might be interesting because we know operations on tuple types are quite hot when typechecking those codebases. The top-level memory usage for both those codebases is the same on `main` and with this PR. The memory taken up by `TupleType` does go down, but it's a small fraction of the overall memory usage

<table>
<tr>
 <td>Project
 <td>TupleType memory usage (main)
 <td>TupleType memory usage (this PR)
 <td>Total memory usage (main and this PR)
<tr>
 <td>static-frame
 <td>metadata: 0.98MB<br>fields: 1.91MB
 <td>metadata: 0.98MB<br>fields: 1.58MB
 <td>348MB
<tr>
 <td>colour
 <td>metadata: 0.12MB<br>fields: 0.44MB
 <td>metadata: 0.12MB<br>fields: 0.39MB
 <td>298MB
</table>

(All measurements done by executing `TY_MEMORY_REPORT=full cargo run --release --manifest-path ../ruff/Cargo.toml -p ty check --quiet`)

That does seem like a pretty decent reduction in memory usage on tuples in `static-frame`, even if it's a small slice of the overall pie. Using boxed slices is also consistent with what we do elsewhere for our Salsa-interned structs, and is more expressive of the fact that types are immutable once they're stored in the Salsa cache. I think the main question is whether this is worth the increase in code complexity -- I could go either way on that.

Performance is in the noise -- some benchmarks are showing tiny speedups, other ones tiny slowdowns.

MichaReiser:
Eh, sphinx has a memory reduction 😆

PR #19865, opened 2025-08-11
[ty] add `total`, `Required` & `NotRequired` support for `TypedDict`

PrettyWood:
⚠️ To review after https://github.com/astral-sh/ruff/pull/19810
(only the last 6 commits are for this PR)

part of https://github.com/astral-sh/ty/issues/154

## Summary
This PR implements complete support for `Required` and `NotRequired` type qualifiers in `TypedDict` classes, along with proper inheritance behavior and the `total=` parameter.
Support both constructor calls and dict literal syntax

```python
from typing import TypedDict
from typing_extensions import Required, NotRequired

class PartialUser(TypedDict, total=False):
    name: Required[str]      # Required despite total=False
    age: int                 # Optional due to total=False
    email: NotRequired[str]  # Explicitly optional (redundant)

class User(TypedDict):
    name: Required[str]      # Explicitly required (redundant)
    age: int                 # Required due to total=True
    bio: NotRequired[str]    # Optional despite total=True

# Valid constructions
partial = PartialUser(name="Alice")  # name required, age optional
full = User(name="Bob", age=25)      # name and age required, bio optional

# Inheritance maintains original field requirements
class Employee(PartialUser):
    department: str                  # Required (new field)
    # name: still Required (inherited)
    # age: still optional (inherited)

emp = Employee(name="Charlie", department="Engineering")  # ✅
Employee(department="Engineering")  # ❌
e: Employee = {"age": 1}  # ❌
```

Screenshot 2025-08-11 at 22 02 57

## Test Plan
Updated Markdown tests

PR #19874, opened 2025-08-12
[ty] typecheck dict methods for `TypedDict`

PrettyWood:
⚠️ To review after https://github.com/astral-sh/ruff/pull/19865
 (only the last 3 commits are for this PR)
 
part of https://github.com/astral-sh/ty/issues/154

## Summary

Typecheck `get()`, `setdefault()`, `pop()` for `TypedDict`

```py
from typing import TypedDict
from typing_extensions import NotRequired

class Employee(TypedDict):
    name: str
    department: NotRequired[str]

emp = Employee(name="Alice", department="Engineering")

emp.get("name")
emp.get("departmen", "Unknown")
emp.pop("department")
emp.pop("name")
```

Screenshot 2025-08-12 at 11 42 12

## Test Plan

Updated Markdown tests


crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2299,49 +2299,132 @@ impl<'db> ClassLiteral<'db> {
...
+                                Parameter::positional_only(Some(Name::new_static("self")))
+                                    .with_annotated_type(instance_ty),
+                                Parameter::positional_only(Some(Name::new_static("key")))
+                                    .with_annotated_type(key_type),
+                            ]),
+                            Some(return_type),
+                        );
+
+                        // `.get()` with a default value returns the field type
+                        // (we could prevent this method for required fields but for now let's do that)
```
sharkdp: I think that would be more of an opinionated lint rule, rather than a hard error. So I suggest we remove that comment for now:
```suggestion
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2299,49 +2299,132 @@ impl<'db> ClassLiteral<'db> {
...
+                            Parameters::new([
+                                Parameter::positional_only(Some(Name::new_static("self")))
+                                    .with_annotated_type(instance_ty),
+                                Parameter::positional_only(Some(Name::new_static("key")))
+                                    .with_annotated_type(key_type),
+                                Parameter::positional_only(Some(Name::new_static("default")))
+                                    .with_annotated_type(field.declared_ty),
+                            ]),
+                            Some(field.declared_ty),
+                        );
```
sharkdp: If the return type is the same (and should stay the same, see my other comment), could we merge these two overloads (with and without default value) into a single overload where the `default` parameter has a default value?

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2299,49 +2299,132 @@ impl<'db> ClassLiteral<'db> {
...
+                        // `.get()` with a default value returns the field type
+                        // (we could prevent this method for required fields but for now let's do that)
+                        let get_with_default_sig = Signature::new(
+                            Parameters::new([
+                                Parameter::positional_only(Some(Name::new_static("self")))
+                                    .with_annotated_type(instance_ty),
+                                Parameter::positional_only(Some(Name::new_static("key")))
+                                    .with_annotated_type(key_type),
+                                Parameter::positional_only(Some(Name::new_static("default")))
+                                    .with_annotated_type(field.declared_ty),
```
sharkdp: I'm wondering if this signature is too strict? Take this example:
```py
from typing import TypedDict, reveal_type

class T(TypedDict, total=False):
    a: str

def _(t: T):
    reveal_type(t.get("a", 0))
```

mypy, pyright, and pyrefly all agree that this code is fine and that the type of the `.get` call should be `str | Literal[0]` (or `str | int`). Instead, with this synthesized overload here, we get an error and a return type of `Unknown`:

```
info[revealed-type]: Revealed type
 --> /home/shark/playground/test.py:9:17
  |
8 | def _(t: T):
9 |     reveal_type(t.get("a", 0))
  |                 ^^^^^^^^^^^^^ `Unknown`
  |

error[invalid-argument-type]: Argument is incorrect
 --> /home/shark/playground/test.py:9:28
  |
8 | def _(t: T):
9 |     reveal_type(t.get("a", 0))
  |                            ^ Expected `str`, found `Literal[0]`
  |
```

I have not checked if this behavior is specified somewhere (we should do that), but if not, it feels safer to go with the behavior or the other type checkers here and return a union of the declared field type and the type of the default argument. This would probably require us to make that overload generic.


A similar comment might also apply to `.pop` and `.setdefault`?

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -6176,6 +6176,60 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
         let callable_type = self.infer_maybe_standalone_expression(func);
 
+        // Special handling for `TypedDict` method calls
+        if let ast::Expr::Attribute(ast::ExprAttribute { value, attr, .. }) = func.as_ref() {
+            let value_type = self.expression_type(value);
+            if let Type::TypedDict(typed_dict_ty) = value_type {
```
sharkdp, line 6277: This won't capture all `.pop` calls (imagine e.g. `method = person.pop; method("some_required_key")`, but it's still a very nice touch!

It's not impossible to make this work for all calls, I think. But it would require a significant rework of our call machinery, and I don't think it's important enough to do.

Even if we miss some calls to these methods by explicitly matching on `Expr::Attribute`, it's not like we won't emit an error. We will only fall back to a worse error message (e.g. no-matching-overload), which seems fine to me if it's relatively rare. I think this behavior might be worth documenting in a test, though.

sharkdp:
I'm currently looking into making the required changes here.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7528,6 +7528,28 @@ pub struct BoundTypeVarInstance<'db> {
...
+                None, // definition
+                None, // _bound_or_constraints
+                Some(variance),
+                None, // _default
+                TypeVarKind::Pep695,
+            ),
+            BindingContext::Synthetic,
+        )
+    }
+
```
sharkdp, line 7552: @dcreager Would be great if you could take a short look at this new method here.. and the new constructor in `generics.rs`. Mainly to check my wording in the doc comments / function names.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2331,49 +2331,182 @@ impl<'db> ClassLiteral<'db> {
...
+                                Parameter::positional_only(Some(Name::new_static("default")))
+                                    .with_annotated_type(Type::TypeVar(t_default)),
+                            ]),
+                            Some(UnionType::from_elements(
+                                db,
+                                [Type::unknown(), Type::TypeVar(t_default)],
+                            )),
+                        )
+                    }))
+                    .collect();
```
dcreager: nit: `CallableSignature::from_overloads` takes in an iterator, so I think you can skip the `collect` here

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2331,49 +2331,182 @@ impl<'db> ClassLiteral<'db> {
...
+                            ]),
+                            Some(UnionType::from_elements(
+                                db,
+                                [field.declared_ty, Type::TypeVar(t_default)],
+                            )),
+                        );
+
+                        [pop_sig, pop_with_default_sig]
+                    })
+                    .collect::<Vec<_>>();
```
dcreager: and here

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2331,49 +2331,182 @@ impl<'db> ClassLiteral<'db> {
...
+                                    .with_annotated_type(instance_ty),
+                                Parameter::positional_only(Some(Name::new_static("key")))
+                                    .with_annotated_type(key_type),
+                                Parameter::positional_only(Some(Name::new_static("default")))
+                                    .with_annotated_type(field.declared_ty),
+                            ]),
+                            Some(field.declared_ty),
+                        )
+                    })
+                    .collect::<Vec<_>>();
```
dcreager: and here

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -128,6 +128,14 @@ impl<'db> GenericContext<'db> {
         Self::new(db, variables)
     }
 
+    /// Creates a generic context from a list of (synthetic) `BoundTypeVarInstance`s.
```
dcreager: nit: I'd suggest removing the "(synthetic)" part, since that doesn't have to be a limitation on when you can use this constructor.  And I think you could simplify `from_type_params` and `normalized_impl` to delegate to this constructor to do the collecting.

dcreager (commented):
Just a couple of comments re the generics/typevar changes.

sharkdp:
Ok, after adding some fallback overloads for `.get`, we're now down to a reasonable number of ecosystem changes. 

```diff
discord.py (https://github.com/Rapptz/discord.py)
- discord/activity.py:882:38: warning[unused-ignore-comment] Unused blanket `type: ignore` directive
- discord/member.py:390:41: warning[unused-ignore-comment] Unused blanket `type: ignore` directive
- Found 533 diagnostics
+ Found 531 diagnostics
```

Correct, they are `.pop` ing a required key, and silence that with `# type: ignore`.

```diff
openlibrary (https://github.com/internetarchive/openlibrary)
+ openlibrary/plugins/worksearch/code.py:447:34: error[invalid-argument-type] Argument to function `__new__` is incorrect: Expected `Iterable[Unknown]`, found `list[str] | None`
+ openlibrary/plugins/worksearch/code.py:447:61: error[invalid-argument-type] Argument to function `__new__` is incorrect: Expected `Iterable[Unknown]`, found `list[str] | None`
- Found 709 diagnostics
+ Found 711 diagnostics
```

True positives. They have a key with value type `list[str] | None`, and then use `.get("that_key", [])` in an attempt to replace `None` with `[]`. But that doesn't work. `get` won't replace values of type `None` with the default. It only replaces missing keys with the default.

```diff
meson (https://github.com/mesonbuild/meson)
+ mesonbuild/interpreter/interpreter.py:1804:69: error[invalid-argument-type] Argument to bound method `__init__` is incorrect: Expected `dict[str, str] | None`, found `dict[OptionKey, str | int | list[str]]`
+ mesonbuild/modules/python.py:169:42: error[invalid-key] Invalid key access on TypedDict `ExtensionModuleKw`: Unknown key "limited_api"
- Found 808 diagnostics
+ Found 810 diagnostics
```

Both look like true positives to me

```diff
prefect (https://github.com/PrefectHQ/prefect)
+ src/prefect/runner/storage.py:851:13: error[invalid-return-type] Return type does not match returned value: expected `str`, found `str | (Any & ~AlwaysFalsy & ~Secret[Unknown]) | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)`
+ src/prefect/runner/storage.py:858:13: error[invalid-return-type] Return type does not match returned value: expected `str`, found `(Any & ~AlwaysFalsy & ~Secret[Unknown]) | str | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)`
+ src/prefect/runner/storage.py:859:16: warning[possibly-unbound-attribute] Attribute `startswith` on type `(Any & ~AlwaysFalsy & ~Secret[Unknown]) | (str & ~AlwaysFalsy) | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)` is possibly unbound
+ src/prefect/runner/storage.py:866:13: error[invalid-return-type] Return type does not match returned value: expected `str`, found `str | (Any & ~AlwaysFalsy & ~Secret[Unknown]) | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)`
+ src/prefect/runner/storage.py:867:20: warning[possibly-unbound-attribute] Attribute `startswith` on type `(Any & ~AlwaysFalsy & ~Secret[Unknown]) | (str & ~AlwaysFalsy) | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)` is possibly unbound
+ src/prefect/runner/storage.py:872:12: error[invalid-return-type] Return type does not match returned value: expected `str`, found `(Any & ~AlwaysFalsy & ~Secret[Unknown]) | (str & ~AlwaysFalsy) | (Secret[str] & ~AlwaysFalsy & ~Secret[Unknown]) | (@Todo(Type::Intersection.call()) & ~AlwaysFalsy)`
- Found 3013 diagnostics
+ Found 3019 diagnostics
```

Seems unrelated to `TypedDict`, only surfaces because we don't infer `Unknown` anymore. Potentially related to https://github.com/astral-sh/ty/issues/456.

```diff
dd-trace-py (https://github.com/DataDog/dd-trace-py)
+ ddtrace/llmobs/_experiment.py:506:28: warning[redundant-cast] Value is already of type `int`
- Found 6602 diagnostics
+ Found 6603 diagnostics
```

True positive

crates/ty_python_semantic/src/types/class.rs
sharkdp: Oops, yes. I missed this in my review. Thanks.

sharkdp (approved):
Thanks again!

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2331,49 +2331,176 @@ impl<'db> ClassLiteral<'db> {
...
+                    // Fallback overloads for unknown keys
+                    .chain(std::iter::once({
+                        Signature::new(
+                            Parameters::new([
+                                Parameter::positional_only(Some(Name::new_static("self")))
+                                    .with_annotated_type(instance_ty),
+                                Parameter::positional_only(Some(Name::new_static("key")))
+                                    .with_annotated_type(KnownClass::Str.to_instance(db)),
+                            ]),
+                            Some(Type::unknown()),
```
AlexWaygood: Wouldn't `Unknown | None` be better here, to express the fact that the unknown key might not be present (it might be a subtype of this typeddict, but it also might not?)?

```suggestion
                            Some(UnionType::from_elements(db, [Type::unknown(), Type::none(db)])),
```

sharkdp: Yes, that makes sense — changed. Let's see if that leads to new diagnostics.

crates/ty_python_semantic/resources/mdtest/typed_dict.md
```diff
@@ -450,19 +450,51 @@ class Person(TypedDict, total=False):
...
+
+    # We allow access to unknown keys (they could be set for a subtype of Person)
+    reveal_type(p.get("unknown"))  # revealed: Unknown | None
+    reveal_type(p.get("unknown", "default"))  # revealed: Unknown | Literal["default"]
+
+    # `pop()` only works on non-required fields
+    reveal_type(p.pop("extra"))  # revealed: str
+    reveal_type(p.pop("extra", "fallback"))  # revealed: str
+    # error: [invalid-argument-type] "Cannot pop required field 'name' from TypedDict `Person`"
+    reveal_type(p.pop("name"))  # revealed: Unknown
```
AlexWaygood, line 487: I feel like ideally we'd still infer `str` here (despite the fact that the operation is illegal, it seems pretty clear that the result of this operation will always be a `str`), but I can see that that might be difficult to do _consistently_ with the synthesized-overload approach. And consistently inferring `Unknown` here seems better than sometimes inferring `Unknown`, sometimes `str` (depending on exactly how the method is invoked on the typeddict object).

sharkdp: > I feel like ideally we'd still infer `str` here (despite the fact that the operation is illegal, it seems pretty clear that the result of this operation will always be a `str`)

Yes, interesting point. I agree that this would probably be the better return type, but as you said, it's not possible to achieve that with synthesized overloads, unless we manage to synthesize overloads that trigger specific `CallError`s when selected, or similar. Until then, `Unknown` is not ideal, but it's also not wrong. And if you're going to `# type: ignore` your illegal `.pop` operation, you might as well throw in a `cast` or a redeclaration :-)

PR #19882, opened 2025-08-12
[ty] render docstrings in hover

Gankra:
This PR has several components:

* Introduce a Docstring String wrapper type that has render_plaintext and render_markdown methods, to force docstring handlers to pick a rendering format
  * Implement [PEP-257](https://peps.python.org/pep-0257/) docstring trimming for it
  * The markdown rendering just renders the content in a plaintext codeblock for now (followup work)
* Introduce a `DefinitionsOrTargets` type representing the partial evaluation of `GotoTarget::get_definition_targets` to ideally stop at getting `ResolvedDefinitions`
  * Add `declaration_targets`, `definition_targets`, and `docstring` methods to `DefinitionsOrTargets` for the 3 usecases we have for this operation
  * `docstring` is of course the key addition here, it uses the same basic logic that `signature_help` was using: first check the goto-declaration for docstrings, then check the goto-definition for docstrings. 
* Refactor `signature_help` to use the new APIs instead of implementing it itself
  * Not fixed in this PR: an issue I found where `signature_help` will erroneously cache docs between functions that have the same type (hover docs don't have this bug)
* A handful of new tests and additions to tests to add docstrings in various places and see which get caught

Examples of it working with stdlib, third party, and local definitions:
Screenshot 2025-08-12 at 2 13 55 PM
Screenshot 2025-08-12 at 2 14 06 PM
Screenshot 2025-08-12 at 2 14 18 PM

Notably modules don't work yet (followup work):
Screenshot 2025-08-12 at 2 14 37 PM

Notably we don't show docs for an item if you hover its actual definition (followup work, but also, not the most important):
Screenshot 2025-08-12 at 2 16 54 PM

### DIFF
--- a/crates/ty_ide/src/docstring.rs
+++ b/crates/ty_ide/src/docstring.rs
@@ -12,6 +12,8 @@ use ruff_source_file::UniversalNewlines;
 use std::collections::HashMap;
 use std::sync::LazyLock;
 
+use crate::MarkupKind;
+
 // Static regex instances to avoid recompilation
 static GOOGLE_SECTION_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"(?i)^\s*(Args|Arguments|Parameters)\s*:\s*$")
@@ -45,6 +47,14 @@ impl Docstring {
     /// Create a new docstring from the raw string literal contents
     pub fn new(raw: String) -> Self {
         Docstring(raw)
+    }
+
+    /// Render the docstring to the given markup format
+    pub fn render(&self, kind: MarkupKind) -> String {
+        match kind {
+            MarkupKind::PlainText => self.render_plaintext(),
+            MarkupKind::Markdown => self.render_markdown(),
+        }
     }
 
     /// Render the docstring for plaintext display
@@ -121,7 +131,7 @@ fn documentation_trim(docs: &str) -> String {
     let mut lines = expanded.universal_newlines();
 
     // If the first line is non-blank then we need to include it *fully* trimmed
-    // As its indentation is ignored (effectively treated as having min-indent).
+    // As its indentation is ignored (effectively treated as having min_indent).
     if leading_blank_lines == 0 {
         if let Some(first_line) = lines.next() {
             output.push_str(first_line.as_str().trim_whitespace());
@@ -129,8 +139,10 @@ fn documentation_trim(docs: &str) -> String {
         }
     }
 
-    // For the rest of the lines remove the minimum indent (if possible) and trailing whitespace
-    // (all python whitespace is ascii so we can just remove that many bytes from the front).
+    // For the rest of the lines remove the minimum indent (if possible) and trailing whitespace.
+    //
+    // We computed min_indent by only counting python whitespace, and all python whitespace
+    // is ascii, so we can just remove that many bytes from the front.
     for line_obj in lines.skip(leading_blank_lines) {
         let line = line_obj.as_str();
         let trimmed_line = line[min_indent.min(line.len())..].trim_whitespace_end();
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -150,7 +150,11 @@ pub(crate) enum GotoTarget<'a> {
 /// The resolved definitions for a `GotoTarget`
 #[derive(Debug, Clone)]
 pub(crate) enum DefinitionsOrTargets<'db> {
+    /// We computed actual Definitions we can do followup queries on.
     Definitions(Vec<ResolvedDefinition<'db>>),
+    /// We directly computed a navigation.
+    ///
+    /// We can't get docs or usefully compute goto-definition for this.
     Targets(crate::NavigationTargets),
 }
 
--- a/crates/ty_ide/src/hover.rs
+++ b/crates/ty_ide/src/hover.rs
@@ -27,11 +27,13 @@ pub fn hover(db: &dyn Db, file: File, offset: TextSize) -> Option<RangedValue<Ho
             db,
             ty_python_semantic::ImportAliasResolution::ResolveAliases,
         )
-        .and_then(|definitions| definitions.docstring(db));
+        .and_then(|definitions| definitions.docstring(db))
+        .map(HoverContent::Docstring);
     tracing::debug!("Inferred type of covering node is {}", ty.display(db));
 
     // TODO: Render the symbol's signature instead of just its type.
-    let contents = vec![HoverContent::Type(ty, docs)];
+    let mut contents = vec![HoverContent::Type(ty)];
+    contents.extend(docs);
 
     Some(RangedValue {
         range: FileRange::new(file, goto_target.range()),
@@ -100,7 +102,8 @@ impl fmt::Display for DisplayHover<'_> {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum HoverContent<'db> {
-    Type(Type<'db>, Option<Docstring>),
+    Type(Type<'db>),
+    Docstring(Docstring),
 }
 
 impl<'db> HoverContent<'db> {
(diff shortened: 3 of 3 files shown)


crates/ty_ide/src/hover.rs
```diff
@@ -165,6 +182,261 @@ mod tests {
...
+                return 0
+
+        x = MyCla<CURSOR>ss(0)
+        "#,
+        );
+
+        assert_snapshot!(test.hover(), @r"
+        <class 'MyClass'>
+        ---------------------------------------------
+        This is such a great class!!
```
Gankra, line 332: This result is a disappointment but a pre-existing issue that we don't really distinguish a class from an invocation of the initializer (future work).

crates/ty_ide/src/hover.rs
```diff
@@ -264,9 +536,15 @@ mod tests {
...
+                """my cool test
 
-            test(a<CURSOR>= 123)
+                Args:
+                    ab: a nice little integer
+                """
+                return 0
+
+            test(a<CURSOR>b= 123)
             "#,
```
Gankra, line 541: This doesn't change the result at all but I changed this test in case one day our hover gets smart enough to see and render these docs.

crates/ty_ide/src/hover.rs
```diff
@@ -296,9 +575,13 @@ mod tests {
...
+                """The foo function"""
+                return 0
 
-            def bar(a, b): ...
+            def bar(a, b):
+                """The bar function"""
+                return 1
 
             if random.choice([True, False]):
                 a = foo
```
Gankra, line 580: Similarly this doesn't change anything, but I added docs just in case something got smarter

crates/ty_ide/src/hover.rs
```diff
@@ -339,7 +622,18 @@ mod tests {
...
         );
 
-        test.write_file("lib.py", "a = 10").unwrap();
+        test.write_file(
+            "lib.py",
+            r"
+        '''
+        The cool lib_py module!
+
+        Wow this module rocks.
```
Gankra, line 624: Future work, this should render eventually.

crates/ty_ide/src/hover.rs
```diff
@@ -362,6 +656,32 @@ mod tests {
...
+
+            lib
+            "#,
+        );
+
+        test.write_file(
+            "lib.py",
+            r"
+        '''
+        The cool lib_py module!
```
Gankra, line 666: This one's even worse and produces no result because apparently hovering an import doesn't yield the module (even though I fixed goto-def on it in a previous PR). I assume this is because hover currently insists it needs the type of the hovered item, not the def/decl.

crates/ty_ide/src/hover.rs
```diff
@@ -654,6 +974,12 @@ mod tests {
...
             r#"
             def foo(a: str | None, b):
+                '''
+                    My cool func
+                
+                    Args:
+                        a: hopefully a string, right?!   
+                '''
                 if a is not None:
                     print(a<CURSOR>)
```
Gankra, line 977: Another "maybe one day" change

crates/ty_ide/src/docstring.rs
```diff
@@ -35,21 +35,110 @@ static REST_PARAM_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
-    param_docs.extend(extract_google_style_params(docstring));
+    /// Extract parameter documentation from popular docstring formats.
+    /// Returns a map of parameter names to their documentation.
+    pub fn parameter_documentation(&self) -> HashMap<String, String> {
+        let mut param_docs = HashMap::new();
 
-    // NumPy-style docstrings
-    param_docs.extend(extract_numpy_style_params(docstring));
+        // Google-style docstrings
+        param_docs.extend(extract_google_style_params(&self.0));
```
MichaReiser, line 81: I know this is pre-existing code, so I'm fine deferring it but I wonder if it makes sense to short-circuit if we parsed parameters by at least one style because most code will only use one convention.

crates/ty_ide/src/docstring.rs
```diff
@@ -35,21 +35,110 @@ static REST_PARAM_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
+    // As its indentation is ignored (effectively treated as having min-indent).
+    if leading_blank_lines == 0 {
+        if let Some(first_line) = lines.next() {
+            output.push_str(first_line.as_str().trim_whitespace());
+            output.push('\n');
+        }
+    }
+
+    // For the rest of the lines remove the minimum indent (if possible) and trailing whitespace
+    // (all python whitespace is ascii so we can just remove that many bytes from the front).
```
MichaReiser:
```suggestion
    // (all python whitespace is ascii so we can just remove that many bytes from the front).
```

I found this comment slightly confusing because whitespace in strings can be unicode whitespace. It's only whitespace in indentations (and between expressions etc) that are all ASCII

crates/ty_ide/src/docstring.rs
```diff
@@ -35,21 +35,110 @@ static REST_PARAM_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
+        param_docs.extend(extract_rest_style_params(&self.0));
+
+        param_docs
+    }
+}
+
+/// Normalizes tabs and trims a docstring as specified in PEP-0257
+///
+/// See: <https://peps.python.org/pep-0257/#handling-docstring-indentation>
+fn documentation_trim(docs: &str) -> String {
```
MichaReiser, line 96: Yay, our second implementation of this algorithm (it makes sense for them to be different):

https://github.com/astral-sh/ruff/blob/9ae698fe30cf3526f0e7ae237b800b3ed19a819f/crates/ruff_python_formatter/src/string/docstring.rs#L27-L114

crates/ty_ide/src/goto.rs
```diff
@@ -146,6 +147,90 @@ pub(crate) enum GotoTarget<'a> {
...
+    ///
+    /// In this case we apply stub-mapping to try to find the "real" implementation
+    /// if the definition we have is found in a stub file.
+    pub(crate) fn definition_targets(
+        self,
+        db: &'db dyn crate::Db,
+    ) -> Option<crate::NavigationTargets> {
+        match self {
+            DefinitionsOrTargets::Definitions(definitions) => {
+                definitions_to_navigation_targets(db, Some(&StubMapper::new(db)), definitions)
```
MichaReiser, line 187: Nit: The stub mapper type still feels a bit overkill to me. It feels like it could easily be replaced by an enum with two variants (map, don't map) and a method (that takes `db` as an argument) that does the mapping. 

Maybe something for a follow up PR?

crates/ty_ide/src/goto.rs
```diff
@@ -146,6 +147,90 @@ pub(crate) enum GotoTarget<'a> {
     },
 }
 
+/// The resolved definitions for a `GotoTarget`
+#[derive(Debug, Clone)]
+pub(crate) enum DefinitionsOrTargets<'db> {
+    Definitions(Vec<ResolvedDefinition<'db>>),
+    Targets(crate::NavigationTargets),
```
MichaReiser: Could you add some documentation to these two variants. Without digging through `ResolvedDefinition` and `NavigationTargets`, it was hard to understand what each represents (and I did such a bad job to not document `NavigationTargets`)

crates/ty_ide/src/hover.rs
```diff
@@ -94,7 +100,7 @@ impl fmt::Display for DisplayHover<'_> {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum HoverContent<'db> {
-    Type(Type<'db>),
+    Type(Type<'db>, Option<Docstring>),
 }
```
MichaReiser: My idea behind making `HoverContent` an enum was that a `Hover` can be made up out of different building blocks where each building block knows how to render itself. A docstring could be such a building block and so is `Type`. Following that idea, I'd probably make `Docstring` its own enum variant. However, I'm not sure if what I had in mind back then does scale well and it heavily depends on how tightly the rendering is coupled between and maybe it's best if collapse `HoverContent` into `Hover`

crates/ty_ide/src/hover.rs
```diff
@@ -116,11 +122,22 @@ pub(crate) struct DisplayHoverContent<'a, 'db> {
...
+                    .fenced_code_block(ty.display(self.db), "python")
+                    .fmt(f)?;
+
+                if let Some(docstring) = docstring {
+                    self.kind.horizontal_line().fmt(f)?;
+
+                    match self.kind {
+                        MarkupKind::PlainText => docstring.render_plaintext().fmt(f)?,
+                        MarkupKind::Markdown => docstring.render_markdown().fmt(f)?,
+                    }
```
MichaReiser: Nit: Would it make sense for `docstring` to have a `render(MarkupKind)` method instead. It seems that would fit nicer together with `MarkupKind` and also allows to use the `fenced_code_block` helper

crates/ty_ide/src/goto.rs
Gankra: I agree, I have several followup refactors planned here.

crates/ty_ide/src/docstring.rs
Gankra: We actually have a test checked in for the mixed behaviour, so i guess @UnboundVariable thought it was something that we'll see in the wild?

crates/ty_ide/src/docstring.rs
Gankra: In retrospect I really should have checked because of course we do 😓

PR #19883, opened 2025-08-12
[ty] Add completion support for `import` and `from ... import` statements

BurntSushi:
This required adding what is effectively a second implementation of
module resolution in order to facilitate asking for "list modules"
instead of "resolve this module." While this does create some
duplication of some pretty hairy logic, there is also a substantial bit
of reuse. For example, `SearchPath` and `ModulePath` encapsulate a fair
bit of logic that we are able to reuse here.

Most of the work in this PR was in writing the tests and occasionally
fixing problems that they uncovered.

One thing this PR does not handle is caching. There are known cache
invalidation problems here. That is, if one adds a module to a search
path, completions may become stale. Similarly for deletions. I intend
to address this in another PR.

(This PR is best reviewed commit by commit.)

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/module.rs
+++ b/crates/ty_python_semantic/src/module_resolver/module.rs
@@ -59,7 +59,7 @@ impl<'db> Module<'db> {
     }
 
     /// Is this a module that we special-case somehow? If so, which one?
-    pub(crate) fn known(self, db: &'db dyn Database) -> Option<KnownModule> {
+    pub fn known(self, db: &'db dyn Database) -> Option<KnownModule> {
         match self {
             Module::File(module) => module.known(db),
             Module::Namespace(_) => None,
@@ -75,7 +75,7 @@ impl<'db> Module<'db> {
     ///
     /// It is guaranteed that if `None` is returned, then this is a namespace
     /// package. Otherwise, this is a regular package or file module.
-    pub(crate) fn search_path(self, db: &'db dyn Database) -> Option<&'db SearchPath> {
+    pub fn search_path(self, db: &'db dyn Database) -> Option<&'db SearchPath> {
         match self {
             Module::File(module) => Some(module.search_path(db)),
             Module::Namespace(_) => None,
@@ -214,7 +214,7 @@ fn all_submodule_names_for_package(db: &dyn Db, file: File) -> Option<Vec<Name>>
 }
 
 /// A module that resolves to a file (`lib.py` or `package/__init__.py`)
-#[salsa::tracked(debug, heap_size=ruff_memory_usage::heap_size)]
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
 pub struct FileModule<'db> {
     #[returns(ref)]
     pub(super) name: ModuleName,
@@ -229,7 +229,7 @@ pub struct FileModule<'db> {
 ///
 /// Namespace packages are special because there are
 /// multiple possible paths and they have no corresponding code file.
-#[salsa::tracked(debug, heap_size=ruff_memory_usage::heap_size)]
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
 pub struct NamespacePackage<'db> {
     #[returns(ref)]
     pub(super) name: ModuleName,
--- a/crates/ty_python_semantic/src/module_resolver/mod.rs
+++ b/crates/ty_python_semantic/src/module_resolver/mod.rs
@@ -3,7 +3,7 @@ use std::iter::FusedIterator;
 pub use list::list_modules;
 pub(crate) use module::KnownModule;
 pub use module::Module;
-pub use path::SearchPathValidationError;
+pub use path::{SearchPath, SearchPathValidationError};
 pub use resolver::SearchPaths;
 pub(crate) use resolver::file_to_module;
 pub use resolver::{resolve_module, resolve_real_module};
--- a/crates/ty_test/src/lib.rs
+++ b/crates/ty_test/src/lib.rs
@@ -18,8 +18,9 @@ use std::fmt::Write;
 use ty_python_semantic::pull_types::pull_types;
 use ty_python_semantic::types::check_types;
 use ty_python_semantic::{
-    Program, ProgramSettings, PythonEnvironment, PythonPlatform, PythonVersionSource,
-    PythonVersionWithSource, SearchPathSettings, SysPrefixPathOrigin,
+    Module, Program, ProgramSettings, PythonEnvironment, PythonPlatform, PythonVersionSource,
+    PythonVersionWithSource, SearchPath, SearchPathSettings, SysPrefixPathOrigin, list_modules,
+    resolve_module,
 };
 
 mod assertion;
@@ -68,13 +69,16 @@ pub fn run(
             Log::Filter(filter) => setup_logging_with_filter(filter),
         });
 
+        let failures = run_test(&mut db, relative_fixture_path, snapshot_path, &test);
+        let inconsistencies = run_module_resolution_consistency_test(&db);
+        let this_test_failed = failures.is_err() || inconsistencies.is_err();
+        any_failures = any_failures || this_test_failed;
+
+        if this_test_failed && output_format.is_cli() {
+            println!("\n{}\n", test.name().bold().underline());
+        }
+
         if let Err(failures) = run_test(&mut db, relative_fixture_path, snapshot_path, &test) {
-            any_failures = true;
-
-            if output_format.is_cli() {
-                println!("\n{}\n", test.name().bold().underline());
-            }
-
             let md_index = LineIndex::from_source_text(&source);
 
             for test_failures in failures {
@@ -100,19 +104,32 @@ pub fn run(
                     }
                 }
             }
-
-            let escaped_test_name = test.name().replace('\'', "\\'");
-
-            if output_format.is_cli() {
-                println!(
-                    "\nTo rerun this specific test, set the environment variable: {}='{escaped_test_name}'",
-                    EnvVars::MDTEST_TEST_FILTER,
-                );
-                println!(
-                    "{}='{escaped_test_name}' cargo test -p ty_python_semantic --test mdtest -- {test_name}",
-                    EnvVars::MDTEST_TEST_FILTER,
-                );
+        }
+        if let Err(inconsistencies) = run_module_resolution_consistency_test(&db) {
+            any_failures = true;
+            for inconsistency in inconsistencies {
+                match output_format {
+                    OutputFormat::Cli => {
+                        let info = relative_fixture_path.to_string().cyan();
+                        println!("  {info} {inconsistency}");
+                    }
+                    OutputFormat::GitHub => {
+                        println!("::error file={absolute_fixture_path}::{inconsistency}");
+                    }
+                }
             }
+        }
+
+        if this_test_failed && output_format.is_cli() {
+            let escaped_test_name = test.name().replace('\'', "\\'");
+            println!(
+                "\nTo rerun this specific test, set the environment variable: {}='{escaped_test_name}'",
+                EnvVars::MDTEST_TEST_FILTER,
+            );
+            println!(
+                "{}='{escaped_test_name}' cargo test -p ty_python_semantic --test mdtest -- {test_name}",
+                EnvVars::MDTEST_TEST_FILTER,
+            );
         }
     }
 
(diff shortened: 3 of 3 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/5cc6b303-f09e-4cc8-a59a-62558115f1be

crates/ty_ide/src/completion.rs
```diff
@@ -50,11 +61,11 @@ enum CompletionTargetTokens<'t> {
...
-        /// This is currently unused, but we should use this
-        /// eventually to remove completions that aren't a
-        /// prefix of what has already been typed. (We are
-        /// currently relying on the LSP client to do this.)
-        #[expect(dead_code)]
+        /// For right now, this is only used to determine
+        /// which module in an `import` statement to return
+        /// submodule completions for. But we could use it
+        /// for other things, like only returning completions
+        /// that with a prefix corresponding to this token.
```
MichaReiser: I feel like something's missing in this sentence

```suggestion
        /// for other things, like only returning completions
        /// that start with a prefix corresponding to this token.
```

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+use crate::program::Program;
+
+use super::module::{Module, ModuleKind};
+use super::path::{ModulePath, SearchPath, SystemOrVendoredPathRef};
+use super::resolver::{
+    ModuleResolveMode, ResolverContext, is_non_shadowable, resolve_file_module, search_paths,
+};
+
+/// List all available modules.
+#[salsa::tracked]
```
MichaReiser, line 16: I know that you intentionally "ignored" caching as part of this PR. However, I do think that caching at the `db` level is undesired and it may change your design (e.g. a global `Lister` isn't possible). 


The issue with caching `list_modules` is that any file addition or removal invalidates the entire cache. I think a better approach is to instead cache at the search path level. Whether we need a cache for the aggregated result is less clear to me.

I think another approach here could be that we have one query for each search path and all it returns is a listing of all entries it contains. A second query than does the aggregation. I think that would somewhat nicely map to your current model, where the entries returned by the search-path-specific-query map to the arguments to `add_path` (or `add_module`?). In the future, we could decide to run those queries in parallel. This might also give us better incrementality because:

* We avoid traversing the directories unnecessarily and invalidation is "as easy as" reading the `revision` of the search path's file root
* The aggregation query only runs if any search path returns a different list of entries. 
* `resolve_file_module` already uses `File`-level APIs, it only invalidates if the specific files chane.


I don't think it's necessary to address this concern as this PR, unless it invalidates some of the design. I'll leave that up to you

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+    /// module to this lister. The `file_type` should be the type
+    /// of `path` (file, directory or symlink).
+    ///
+    /// This may decide that the given path does not correspond to
+    /// a valid Python module. In which case, it is dropped and this
+    /// is a no-op.
+    fn add_path<'p>(
+        &mut self,
+        search_path: &SearchPath,
+        path: impl Into<SystemOrVendoredPathRef<'p>>,
```
MichaReiser: Is the `Into` worth it here, given that there are very few call sites (and it's a fairly chunky method)

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+///
+/// This is responsible for accumulating modules indexed by
+/// module name. It also handles precedence by implementing the
+/// rules that determine which module gets priority when there is
+/// otherwise ambiguity (e.g., `foo.py` versus `foo/__init__.py`
+/// in the same directory).
+struct Lister<'db> {
+    db: &'db dyn Db,
+    program: Program,
+    modules: BTreeMap<&'db ModuleName, Module<'db>>,
```
MichaReiser, line 50: Mainly curious because I think I can learn something here :). What's the motivation for using a `BTreeMap` here over a `FxHashMap`? Is it because `into_modules` returns a sorted list or are there other reasons?

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+
+use super::module::{Module, ModuleKind};
+use super::path::{ModulePath, SearchPath, SystemOrVendoredPathRef};
+use super::resolver::{
+    ModuleResolveMode, ResolverContext, is_non_shadowable, resolve_file_module, search_paths,
+};
+
+/// List all available modules.
+#[salsa::tracked]
+pub(crate) fn list_modules(db: &dyn Db) -> Vec<Module<'_>> {
```
MichaReiser: I wonder if we could have some form of integration test for `list_modules` that ensures that: for every module returned by list module, `resolve_module(module.name) == module`. Maybe that's something we could run at the end of each mdtest?

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+
+    /// Tests that if we have a `foo.py` and a `foo/__init__.py`, then the
+    /// latter takes precedence.
+    ///
+    /// This is somewhat difficult to test using the in-memory file system,
+    /// since it always returns directory entries in lexicographic order. This
+    /// in turn implies that `foo` will always appear before `foo.py`. But to
+    /// truly test this, we would like to also be correct in the case where
+    /// `foo.py` appears before `foo` (which can certainly happen in the real
+    /// world).
```
MichaReiser, line 476: You could extend `TestCaseBuilder` to allow customizing the underlying `System`. But I suspect that doesn't make it easier as you know rely on os behavior :)

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+    fn first_party_module() {
+        let TestCase { db, .. } = TestCaseBuilder::new()
+            .with_src_files(&[("foo.py", "")])
+            .build();
+
+        insta::assert_debug_snapshot!(
+            list_snapshot(&db),
+            @r#"
+        [
+            Module::File("foo", "first-party", "//src/foo.py", Module, None),
```
MichaReiser: An alternative to the inline tests here could have been to extend `ty_extensions.internal` with a `list_modules` method so that you can then write mdtests for it. But I'm not sure if other folks would agree with this :)

It just feels a bit unfortunate that we now have some mdtests testing module resolution and a separate set of tests that verify the very same behavior for list modules

crates/ty_python_semantic/src/module_resolver/module.rs
```diff
@@ -251,7 +254,18 @@ impl ModuleKind {
 }
 
 /// Enumeration of various core stdlib modules in which important types are located
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, strum_macros::EnumString, get_size2::GetSize)]
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    PartialOrd,
```
MichaReiser: What do we need `Ord` for in `KnownModule`? I'm asking because some of those entries map to multiple actual modules and it doesn't really have a natural ordering.

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+            // this ends up working okay. But if typeshed ever uses
+            // namespace packages, then this will need to be accounted
+            // for.
+            let is_dir =
+                file_type.is_definitely_directory() || module_path.is_directory(&self.context());
+            if is_dir {
+                if !search_path.is_standard_library() {
+                    self.add_module(
+                        &module_path,
+                        Module::namespace_package(self.db, module_name),
```
MichaReiser, line 158: I think we now need to make `Module` a salsa interned or we risk that the same module resolved with `resolve_module` and its counterpart from `list_modules` won't compare equal because the identity of a tracked struct not only depends on the tracked struct's values but also the query that creates it. That means, any tracked struct by `list_modules` won't compare equal to any tracked struct created by `resolve_module` even if their values are identical. 

You may need to run ty on some larger projects to see if the need to go through a global lock (because interning...) has any negative performance impact (see https://github.com/astral-sh/ruff/pull/19867)

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+            // When we had a namespace package and now try to
+            // insert a non-namespace package, the latter always
+            // takes precedent, even if it's in a lower priority
+            // search path.
+            (None, Some(_)) => {
+                entry.insert(module);
+            }
+            (Some(search_path_existing), Some(search_path_new)) => {
+                // All of the cases below require that we're
+                // in the same directory.
```
MichaReiser: The comment here explains what the condition right beneath it does without saying why the condition is here in the first place. I think I'd change it to something like: Merging across search paths is only necessary for namespace packages. For all other modules, entries from earlier search paths take precedence.

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+    /// with.
+    fn python_version(&self) -> PythonVersion {
+        self.program.python_version(self.db)
+    }
+
+    /// Returns true if and only if the given file extension corresponds
+    /// to a Python source or stub file. This also takes file system
+    /// case sensitivity into account.
+    fn is_python_extension(&self, ext: &str) -> bool {
+        if self.db.system().case_sensitivity().is_case_sensitive() {
```
MichaReiser: Checking whether the file system is case sensitive can be fairly expensive (as expensive as writing a file to disk!). The best case is that it's already cached. Either way, I'm not sure if this is a worthwhile performance optimization or if we should just always call `eq_ignore_ascii_case`

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+                m1.known(db),
+            );
+            let key2 = (
+                m2.name(db),
+                m2.search_path(db).map(|sp| sp.as_path()),
+                m2.kind(db),
+                m2.known(db),
+            );
+            key1.cmp(&key2)
+        });
```
MichaReiser: Shouldn't it be enough to sort by module name? I don't think we should ever return multiple modules with the same name?

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+    }
+
+    #[test]
+    #[cfg(target_family = "unix")]
+    fn symlink() -> anyhow::Result<()> {
+        use anyhow::Context;
+
+        let mut db = TestDb::new();
+
+        let temp_dir = tempfile::TempDir::with_prefix("PREFIX-SENTINEL")?;
```
MichaReiser, line 886: Nit: You can change the underlying system of a `TestDb` by using `db.use_system(other)`. See this test for an example https://github.com/astral-sh/ruff/blob/f34b65b7a00dfc61c5a54e4280b421c94151e677/crates/ty_python_semantic/src/module_resolver/resolver.rs#L2158-L2164

It's up to you if you think this simplifies anything or not.

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1688 @@
...
+            @r#"
+        [
+            Module::File("foo", "first-party", "//src/foo-stubs/__init__.pyi", Package, None),
+        ]
+        "#,
+        );
+    }
+
+    #[test]
+    fn stub_package_not_allowed_in_typeshed() {
```
MichaReiser, line 1657: That's a lot of tests :) Nice work

MichaReiser (approved):
This is so cool. Thank you for working on this. 

I think we should add an experimental LSP setting and use it to gate the completion for now (off by default). We can remove it again once we fix caching. See https://github.com/astral-sh/ruff/commit/7b6abfb030653f81a1d7cdb2b92e78947d4e28c6 on how to do that

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Nice catch. I flipped the `.into()` to the caller.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Yeah it's just for sorting. Nothing deeper. I just reached for the simplest thing.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: I think you mean threading down a parameter that would allow controlling the sort order of `MemoryFileSystem::read_directory` right? Yeah idk if it's worth it.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Not just for listing modules, for `resolve_module` as well. That's where these tests came from. i.e., I copied the approach to testing `resolve_module`. The `TestCaseBuilder` was already there for me to help write tests. This approach also lets us do some things that might be tricky in the context of mdtest, like setting up symlinks. (Although admittedly most tests I think could be written in the style you suggest, assuming the existence of an internal `list_modules` method.)

crates/ty_python_semantic/src/module_resolver/module.rs
BurntSushi: It just made it easy to sort `Vec<Module>` in order to give more sensible snapshots.

I can certainly use a different approach, but this was simplest and it didn't seem obviously wrong to me.

BurntSushi: I got rid of these impls and sorted on `known.as_str()` instead.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: I've made it an `interned` struct.

I'll try it on some larger repos.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Yes, thank you, that is much clearer.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Fine with me. I was thinking that if we always used `eq_ignore_ascii_case`, then we'd consider `foo.PY` as a module on a case sensitive file system. But maybe that's okay?

BurntSushi: There are also some uses of `is_case_sensitive` in the `resolve_module` implementation.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Yup, you're right. When I wrote this part of the test harness, it came before my `Lister` abstraction and I was returning multiple modules of the same name. So sorting by more stuff was helpful. But with `Lister` almost completely ruling that out by construction, it's probably not worth the extra sort criteria.

This also let me revert my `Ord` additions.

crates/ty_python_semantic/src/module_resolver/list.rs
```diff
@@ -0,0 +1,1699 @@
...
+
+        let mut db = TestDb::new();
+
+        let temp_dir = tempfile::TempDir::with_prefix("PREFIX-SENTINEL")?;
+        let root = temp_dir
+            .path()
+            .canonicalize()
+            .context("Failed to canonicalize temp dir")?;
+        let root = SystemPath::from_std_path(&root).unwrap();
+        db.use_system(OsSystem::new(root));
```
BurntSushi, line 892: @MichaReiser RE <https://github.com/astral-sh/ruff/pull/19883/files#r2272642630>

That's what I'm doing here? (This test originated from `resolver.rs`.)

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Is this what you mean? https://github.com/astral-sh/ruff/pull/19883/files#r2273636330

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Turns out that the extension always has to be `py` or `pyi`, even on case insensitive file systems (I confirmed this on macOS at least), so this can get even simpler!

crates/ty_python_semantic/src/module_resolver/list.rs
MichaReiser: Lol, right. Blind me. I think the only consideration here would be to add a `with_system` to the `TestBuilder` so that we could reuse some code around setting up the `ProgramSettings` but maybe that's not possible after all

BurntSushi:
I spoke with @MichaReiser offline and we agreed it'd probably be fine to land this without adding an experimental opt-in since I'm going to be working on caching next anyway. Plus, the bug related to caching can technically be worked around by restarting the LSP server (obviously not ideal). But, I'll wait to merge this until after the next release to give the biggest time window.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: Aye. I'll leave it as-is for now, since we only do this in a few places.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: I tried on our typical large repo test, and things seem okay:

```
$ time ty-main check > /dev/null
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
Checking ------------------------------------------------------------ ?????/????? files                                                                                                                                                 WARN A fatal error occurred while checking some files. Not all project files were analyzed. See the diagnostics list above for details.

real    21.397
user    2:34.59
sys     13.250
maxmem  20325 MB
faults  0

$ time ty-ag-import-module check > /dev/null
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
Checking ------------------------------------------------------------ ?????/????? files                                                                                                                                                 WARN A fatal error occurred while checking some files. Not all project files were analyzed. See the diagnostics list above for details.

real    21.490
user    2:35.97
sys     13.394
maxmem  20548 MB
faults  0
```

I tried it on several other projects and results were comparable.

crates/ty_python_semantic/src/module_resolver/list.rs
BurntSushi: This is finally done in #19887.

PR #19711, opened 2025-08-03
[ty] Add diagnostics for invalid `await` expressions

theammir:
## Summary

This PR adds a new lint, `invalid-await`, for all sorts of reasons why an object may not be `await`able, as discussed in astral-sh/ty#919.
Precisely, `__await__` is guarded against being missing, possibly unbound, or improperly defined (expects additional arguments or doesn't return an iterator).

Of course, diagnostics need to be fine-tuned.  If `__await__` cannot be called with no extra arguments, it indicates an error (or a quirk?) in the method signature, not at the call site. Without any doubt, such an object is not `Awaitable`, but I feel like talking about arguments for an *implicit* call is a bit leaky.
I didn't reference any actual diagnostic messages in the lint definition, because I want to hear feedback first.

Also, there's no mention of the actual required method signature for `__await__` anywhere in the docs. The only reference I had is the `typing` stub. I basically ended up linking `[Awaitable]` to ["must implement `__await__`"](https://docs.python.org/3/library/collections.abc.html#collections.abc.Awaitable), which is insufficient on its own.

## Test Plan

The following code was tested:
```python
import asyncio
import typing


class Awaitable:
    def __await__(self) -> typing.Generator[typing.Any, None, int]:
        yield None
        return 5


class NoDunderMethod:
    pass


class InvalidAwaitArgs:
    def __await__(self, value: int) -> int:
        return value


class InvalidAwaitReturn:
    def __await__(self) -> int:
        return 5


class InvalidAwaitReturnImplicit:
    def __await__(self):
        pass


async def main() -> None:
    result = await Awaitable()  # valid
    result = await NoDunderMethod()  # `__await__` is missing
    result = await InvalidAwaitReturn()  # `__await__` returns `int`, which is not a valid iterator 
    result = await InvalidAwaitArgs()  # `__await__` expects additional arguments and cannot be called implicitly
    result = await InvalidAwaitReturnImplicit()  # `__await__` returns `Unknown`, which is not a valid iterator


asyncio.run(main())
```

PR #19884, opened 2025-08-12
[ty] simplify return type of place_from_declarations

carljm:
## Summary

A [passing comment](https://github.com/astral-sh/ruff/pull/19711#issuecomment-3169312014) led me to explore why we didn't report a class attribute as possibly unbound if it was a method and defined in two different conditional branches.

I found that the reason was because of our handling of "conflicting declarations" in `place_from_declarations`. It returned a `Result` which would be `Err` in case of conflicting declarations.

But we only actually care about conflicting declarations when we are actually doing type inference on that scope and might emit a diagnostic about it. And in all cases (including that one), we want to otherwise proceed with the union of the declared types, as if there was no conflict.

In several cases we were failing to handle the union of declared types in the same way as a normal declared type if there was a declared-types conflict. The `Result` return type made this mistake really easy to make, as we'd match on e.g. `Ok(Place::Type(...))` and do one thing, then match on `Err(...)` and do another, even though really both of those cases should be handled the same.

This PR refactors `place_from_declarations` to instead return a struct which always represents the declared type we should use in the same way, as well as carrying the conflicting declared types, if any. This struct has a method to allow us to explicitly ignore the declared-types conflict (which is what we want in most cases), as well as a method to get the declared type and the conflict information, in the case where we want to emit a diagnostic on the conflict.

## Test Plan

Existing CI; added a test showing that we now understand a multiply-conditionally-defined method as possibly-unbound.

This does trigger issues on a couple new fuzzer seeds, but the issues are just new instances of an already-known (and rarely occurring) problem which I already plan to address in a future PR, so I think it's OK to land as-is.

I happened to build this initially on top of https://github.com/astral-sh/ruff/pull/19711, which adds invalid-await diagnostics, so I also updated some invalid-syntax tests to not await on an invalid type, since the purpose of those tests is to check the syntactic location of the `await`, not the validity of the awaited type.

### DIFF
--- a/crates/ty_python_semantic/src/place.rs
+++ b/crates/ty_python_semantic/src/place.rs
@@ -472,12 +472,56 @@ pub(crate) fn place_from_declarations<'db>(
     place_from_declarations_impl(db, declarations, RequiresExplicitReExport::No)
 }
 
-pub(crate) type DeclaredTypeAndConflictingTypes<'db> =
-    (TypeAndQualifiers<'db>, Box<indexmap::set::Slice<Type<'db>>>);
+type DeclaredTypeAndConflictingTypes<'db> = (
+    TypeAndQualifiers<'db>,
+    Option<Box<indexmap::set::Slice<Type<'db>>>>,
+);
 
 /// The result of looking up a declared type from declarations; see [`place_from_declarations`].
-pub(crate) type PlaceFromDeclarationsResult<'db> =
-    Result<PlaceAndQualifiers<'db>, DeclaredTypeAndConflictingTypes<'db>>;
+pub(crate) struct PlaceFromDeclarationsResult<'db> {
+    place_and_quals: PlaceAndQualifiers<'db>,
+    conflicting_types: Option<Box<indexmap::set::Slice<Type<'db>>>>,
+}
+
+impl<'db> PlaceFromDeclarationsResult<'db> {
+    fn conflict(
+        place_and_quals: PlaceAndQualifiers<'db>,
+        conflicting_types: Box<indexmap::set::Slice<Type<'db>>>,
+    ) -> Self {
+        PlaceFromDeclarationsResult {
+            place_and_quals,
+            conflicting_types: Some(conflicting_types),
+        }
+    }
+
+    pub(crate) fn ignore_conflicting_declarations(self) -> PlaceAndQualifiers<'db> {
+        self.place_and_quals
+    }
+
+    pub(crate) fn with_conflicting_declarations(
+        self,
+    ) -> (
+        PlaceAndQualifiers<'db>,
+        Option<Box<indexmap::set::Slice<Type<'db>>>>,
+    ) {
+        (self.place_and_quals, self.conflicting_types)
+    }
+}
+
+impl<'db> From<PlaceAndQualifiers<'db>> for PlaceFromDeclarationsResult<'db> {
+    fn from(place_and_quals: PlaceAndQualifiers<'db>) -> Self {
+        PlaceFromDeclarationsResult {
+            place_and_quals,
+            conflicting_types: None,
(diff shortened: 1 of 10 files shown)


MichaReiser:
Thank you for looking into this.

> But we only actually care about conflicting declarations when we are actually doing type inference on that scope and might emit a diagnostic about it. And in all cases (including that one), we want to otherwise proceed with the union of the declared types, as if there was no conflict.

To me, this doesn't sound much different to the `Type::try_*` methods that all return a `Result`. It seems that the main issue isn't that the return type is a `Result`, but that

1. We call `.ok()` on the result in some places when we shouldn't
2. We don't track the `PlaceAndQualifiers` for declarations with conflicts.

I don't feel strongly but I would prefer, if possible without much added complexity, to follow the same pattern that we use in our `Type::try_` methods instead of establishing another pattern. Which means, `place_from_declarations` would still return a `Result` but the type wrapped in the `Err` variant has a method to get the `PlaceAndQualifiers` for situations where we don't care about conflicting declarations.

carljm:
> this doesn't sound much different to the `Type::try_*` methods

It seems quite different to me, because `place_from_declarations` never "fails" -- it always succeeds, and always succeeds using exactly the same logic to determine its return type. It just in some cases also provides extra information that some callers may want to use to emit a diagnostic for the user. The key difference is that we _always_ (at literally every possibly callsite) want to handle the returned value in _exactly_ the same way, we _never_ want to distinguish the handling between Ok and Err cases (except that in one callsite we _additionally_ want to emit a diagnostic if conflicting-types were returned -- but we still handle the main return value identically.)

I think a better parallel is our `Place` type (the Bound and PossiblyUnbound cases), where we have a `.ignore_possibly_unbound()` method.

So I think this is following existing patterns in the codebase, just different ones (which I think are more similar to this case.)

> `place_from_declarations` would still return a `Result` but the type wrapped in the `Err` variant has a method to get the `PlaceAndQualifiers` for situations where we don't care about conflicting declarations.

I don't think this would address the problem I aimed to fix in this PR.  It isn't possible to add an `.ignore_conflicting_declarations()` method directly to `Result`, so it would remain the case that we would need separate paths in every callsite to handle the `Ok` and `Err` versions, introducing the risk that we fail to handle them the same. A match arm for `Ok(...)` would not match the `Err` variant, leading to bugs.

MichaReiser:
> It seems quite different to me, because place_from_declarations never "fails" -- it always succeeds, and always succeeds using exactly the same logic to determine its return type.

Thanks, I think this is the key distinction that wasn't clear to me

crates/ty_python_semantic/src/place.rs
```diff
@@ -472,12 +472,56 @@ pub(crate) fn place_from_declarations<'db>(
...
+
+    pub(crate) fn ignore_conflicting_declarations(self) -> PlaceAndQualifiers<'db> {
+        self.place_and_quals
+    }
+
+    pub(crate) fn with_conflicting_declarations(
+        self,
+    ) -> (
+        PlaceAndQualifiers<'db>,
+        Option<Box<indexmap::set::Slice<Type<'db>>>>,
```
MichaReiser, line 505: Nit: The `with` naming here suggests that this creates a new `PlaceFromDeclarationResult`. Should this be `into_conflicting_declarations` or similar?

MichaReiser (approved):

crates/ty_python_semantic/src/place.rs
carljm: Yeah, I wasn't sure of the best naming for this method, probably the most clear name would be `into_place_and_qualifiers_and_conflicting_declarations` but that's quite a mouthful :) On the other hand, there's only one callsite, so maybe a mouthful is fine.

Issue #17203, opened 2025-04-04
Multi-span diagnostics


ntBre:
Places to use multi-span diagnostics once they are added to ruff.

- [ ] https://github.com/astral-sh/ruff/pull/17121#discussion_r2022946393
- [ ] https://github.com/astral-sh/ruff/pull/17129#discussion_r2024242987
- [ ] https://github.com/astral-sh/ruff/pull/17138#discussion_r2026698681

Avasam:
- https://github.com/astral-sh/ruff/pull/17003#discussion_r2030543917 also seems to be on hold waiting for multi-span diagnostics

ntBre:
While reviewing #20178, I realized that [builtin-attribute-shadowing (A003)](https://docs.astral.sh/ruff/rules/builtin-attribute-shadowing/#builtin-attribute-shadowing-a003) might be another good candidate here. We currently emit diagnostics like this:

```
A003 Python builtin is shadowed by method `str` from line 14
  --> A003.py:17:31
   |
15 |         pass
16 |
17 |     def method_usage(self) -> str:
   |                               ^^^
18 |         pass
   |
```

underlining the _use_ of the shadowed name and pointing back to the location in the message. We could now add a secondary annotation where the shadowing takes place, on line 14 in this case.

11happy:
Linking the PR https://github.com/astral-sh/ruff/pull/22565
Thank you : )

PR #19886, opened 2025-08-12
[`pyupgrade`] Add sub-diagnostics showing type variable definitions (`UP046`)

ntBre:
Summary
--

Now that we've handled caching for the new diagnostics, I went looking for diagnostics that could use additional annotations or sub-diagnostics that weren't already tracked in #17203. UP046 (and UP040 and UP047 if we like this one) seemed like a reasonable candidate because we check for separate type variable definitions in the same file before emitting a diagnostic. In this case, I added an info-level sub-diagnostic pointing to this type variable definition. I'm not actually sure if this is super helpful information, so I wouldn't mind closing this if others don't think it's useful.

I also added a helper `DiagnosticGuard::info` method that I think will be useful for adding this kind of sub-diagnostic in general.

I spent a while debugging why the snapshots were showing a longer path in the sub-diagnostics before remembering that we overwrite the source file for the main diagnostic here:

https://github.com/astral-sh/ruff/blob/79c949f0f75f6582782f3d5fddc65089789801d5/crates/ruff_linter/src/test.rs#L280-L288

As shown below, the sub-diagnostics look as expected in the CLI. I think we _could_ delete this part of the test code, but it results in using the longer test file names (like our new sub-diagnostics) in every snapshot. So alternatively, we may want to apply the same transformation to the sub-diagnostics (and secondary annotations) too.

Test Plan
--

Existing UP046 tests with newly expanded snapshots, as well as some manual testing in the CLI:

image

Alternative Idea
--

I don't really love the `help:` and `info:` lines stacked so closely together, so I also tried this as a secondary annotation:

image

That's not implemented in the PR but could be a different route with a small change to `DiagnosticGuard::info` (and a rename of the method).

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -497,6 +497,16 @@ impl Diagnostic {
             other.expect_range().start(),
         ))
     }
+
+    /// Sort this diagnostic's sub-diagnostics by descending severity.
+    ///
+    /// This is particularly helpful to move `help`-level sub-diagnostics to the end, after any
+    /// longer, `info`-level diagnostics.
+    pub fn sort_sub_diagnostics(&mut self) {
+        Arc::make_mut(&mut self.inner)
+            .subs
+            .sort_by_key(|sub| std::cmp::Reverse(sub.inner.severity));
+    }
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash, get_size2::GetSize)]
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -254,6 +254,11 @@ impl Diagnostic {
             .find(|ann| ann.is_primary)
     }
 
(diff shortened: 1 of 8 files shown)


ntBre:
Ah, I guess we need to replace all the paths to make the tests work on Windows anyway.

Also, I'm starting to think the secondary annotation might be a lot better, especially for multiple type variables:

image

MichaReiser:
> I don't really love the help: and info: lines stacked so closely together, so I also tried this as a secondary annotation:

I think what we do in ty is that help texts and notes are always the last sub-diagnostics.

ntBre:
Ah, that does look better already. Thanks!

image

I just did this by sorting the sub-diagnostics in our `DiagnosticGuard` `Drop` impl. That seemed like the easiest fix since we push the `help` diagnostic right when creating the `Diagnostic` from a `Violation`, but we might want a better solution longer-term.

ntBre:
I talked with Alex on Discord, and he echoed my concern that this wasn't a very useful piece of information to show to users. I'll go ahead and close this for now and hope to use some of the more general helper code for a different diagnostic.

PR #19888, opened 2025-08-13
[ty] fix deferred name loading in PEP695 generic classes/functions

carljm:
## Summary

For PEP 695 generic functions and classes, there is an extra "type params scope" (a child of the outer scope, and wrapping the body scope) in which the type parameters are defined; class bases and function parameter/return annotations are resolved in that type-params scope.

This PR fixes some longstanding bugs in how we resolve name loads from inside these PEP 695 type parameter scopes, and also defers type inference of PEP 695 typevar bounds/constraints/default, so we can handle cycles without panicking.

We were previously treating these type-param scopes as lazy nested scopes, which is wrong. In fact they are eager nested scopes; the class `C` here inherits `int`, not `str`, and previously we got that wrong:

```py
Base = int

class C[T](Base): ...

Base = str
```

But certain syntactic positions within type param scopes (typevar bounds/constraints/defaults) are lazy at runtime, and we should use deferred name resolution for them. This also means they can have cycles; in order to handle that without panicking in type inference, we need to actually defer their type inference until after we have constructed the `TypeVarInstance`.

PEP 695 does specify that typevar bounds and constraints cannot be generic, and that typevar defaults can only reference prior typevars, not later ones. This reduces the scope of (valid from the type-system perspective) cycles somewhat, although cycles are still possible (e.g. `class C[T: list[C]]`). And this is a type-system-only restriction; from the runtime perspective an "invalid" case like `class C[T: T]` actually works fine.

I debated whether to implement the PEP 695 restrictions as a way to avoid some cycles up-front, but I ended up deciding against that; I'd rather model the runtime name-resolution semantics accurately, and implement the PEP 695 restrictions as a separate diagnostic on top. (This PR doesn't yet implement those diagnostics, thus some `# TODO: error` in the added tests.)

Introducing the possibility of cyclic typevars made typevar display potentially stack overflow. For now I've handled this by simply removing typevar details (bounds/constraints/default) from typevar display. This impacts display of two kinds of types. If you `reveal_type(T)` on an unbound `T` you now get just `typing.TypeVar` instead of `typing.TypeVar("T", ...)` where `...` is the bound/constraints/default. This matches pyright and mypy; pyrefly uses `type[TypeVar[T]]` which seems a bit confusing, but does include the name. (We could easily include the name without cycle issues, if there's a syntax we like for that.)

It also means that displaying a generic function type like `def f[T: int](x: T) -> T: ...` now displays as `f[T](x: T) -> T` instead of `f[T: int](x: T) -> T`. This matches pyright and pyrefly; mypy does include bound/constraints/defaults of typevars in function/callable type display. If we wanted to add this, we would either need to thread a visitor through all the type display code, or add a `decycle` type transformation that replaced recursive reoccurrence of a type with a marker.

## Test Plan

Added mdtests and modified existing tests to improve their correctness.

After this PR, there's only a single remaining py-fuzzer seed in the 0-500 range that panics! (Before this PR, there were 10; the fuzzer likes to generate cyclic PEP 695 syntax.)

## Ecosystem report

It's all just the changes to `TypeVar` display.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
@@ -86,7 +86,7 @@ error[call-non-callable]: Object of type `Literal[5]` is not callable
    |         ^^^^
    |
 info: Union variant `Literal[5]` is incompatible with this call site
-info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T: str](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
+info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
 info: rule `call-non-callable` is enabled by default
 
 ```
@@ -101,7 +101,7 @@ error[call-non-callable]: Object of type `PossiblyNotCallable` is not callable (
    |         ^^^^
    |
 info: Union variant `PossiblyNotCallable` is incompatible with this call site
-info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T: str](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
+info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
 info: rule `call-non-callable` is enabled by default
 
 ```
@@ -116,7 +116,7 @@ error[missing-argument]: No argument provided for required parameter `b` of func
    |         ^^^^
    |
 info: Union variant `def f3(a: int, b: int) -> int` is incompatible with this call site
-info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T: str](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
+info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
 info: rule `missing-argument` is enabled by default
 
 ```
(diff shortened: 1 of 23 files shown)


crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
```diff
@@ -86,7 +86,7 @@ error[call-non-callable]: Object of type `Literal[5]` is not callable
    |         ^^^^
    |
 info: Union variant `Literal[5]` is incompatible with this call site
-info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T: str](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
+info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
```
dcreager, line 89: This is the place where I'm most leery of not including bounds/constraints/defaults in the rendering of a typevar. Though pyright also does not include them here, so your argument still holds!

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -4520,7 +4520,9 @@ impl KnownClass {
                 }
 
                 let bound_or_constraint = match (bound, constraints) {
-                    (Some(bound), None) => Some(TypeVarBoundOrConstraints::UpperBound(*bound)),
+                    (Some(bound), None) => Some(LazyTypeVarBoundOrConstraints::Eager(
+                        TypeVarBoundOrConstraints::UpperBound(*bound),
+                    )),
```
dcreager: optional nit: These would all be mildly shorter if you added `From` impls to create eager bounds, constraints, and defaults.

```suggestion
                    (Some(bound), None) => Some(TypeVarBoundOrConstraints::UpperBound(*bound).into()),
```

dcreager (approved):
Looks good!

crates/ty_python_semantic/src/types.rs
```diff
@@ -6862,14 +6862,14 @@ pub struct TypeVarInstance<'db> {
...
+    /// The upper bound or constraint on the type of this TypeVar, if any
+    maybe_bound_or_constraints: Option<LazyTypeVarBoundOrConstraints<'db>>,
 
     /// The variance of the TypeVar
     variance: TypeVarVariance,
 
-    /// The default type for this TypeVar
-    default_ty: Option<Type<'db>>,
+    /// The default type for this TypeVar, if any
+    maybe_default: Option<TypeVarDefault<'db>>,
```
AlexWaygood: eh, not sure about the `maybe_` prefixes here -- I feel like that's already expressed well by the fact that the type of the field is an `Option`.

I guess I see more of an argument for giving the `bound_or_constraints` field a different name, since any TypeVar without an explicit upper bound or constraints implicitly has an upper bound of `object`. But if the aim here is to differentiate between explicit and implicit upper bounds then I'd call it `explicit_bound_or_constraints` rather than `maybe_bound_or_constraints`.

Edit: Oh, hmm, I see you've renamed these because you're adding methods that have the old name? In that case, I think I'd appreciate some docs on these fields that specify that the fields are internal and that you should usually use the methods instead

crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
carljm: Yes, agreed. Pyrefly and pyright both do not include typevar details here, mypy does.

We can add them back in future. I'm not opposed to that at all, it's just gonna require some extra cycle-handling machinery for type display that I don't think is top priority right now.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7054,6 +7156,24 @@ impl TypeVarVariance {
     }
 }
 
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, salsa::Update, get_size2::GetSize)]
+pub enum TypeVarDefault<'db> {
+    /// The default type is lazily evaluated.
+    Lazy,
+    /// The default type is eagerly specified.
+    Eager(Type<'db>),
+}
```
AlexWaygood: maybe

```suggestion
/// Whether a TypeVar is eagerly or lazily evaluated at runtime
#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, salsa::Update, get_size2::GetSize)]
pub enum TypeVarDefaultEvaluation<'db> {
    /// The default type is lazily evaluated.
    Lazy,
    /// The default type is eagerly specified.
    Eager(Type<'db>),
}
```

? And similar for your `LazyTypeVarBoundOrConstraints` enum?

crates/ty_python_semantic/src/types.rs
carljm: Yeah, I wanted to preserve the external API unchanged, and I struggled finding names for these internal fields. Would underscore-prefixing them be better? That both resolves the name clash and maybe communicates "internal-only" better, too.

crates/ty_python_semantic/src/types.rs
carljm: Hmm, yeah, I could go with that. Struggled with naming for this one, too.

crates/ty_python_semantic/src/types.rs
AlexWaygood: yeah, underscore-prefixing feels appealing! I think the main thing that would help would just be a 1-2 sentence doc-comment saying "this is internal, go use the method instead".

dcreager: I would be against underscore prefixing since that means "can be `dead_code`" to the compiler. I'd prefer an `_internal` suffix or similar instead

AlexWaygood: > I'd prefer an `_internal` suffix or similar instead

also works for me!

PR #19892, opened 2025-08-13
[ty] Reduce allocations in `tuple.rs`

AlexWaygood:
## Summary

The vast majority of real-world variable-length tuples are "pure homogeneous" tuples with no prefix and no suffix. We can optimize for the common case by using `Option<Box<[T]>>` for these fields rather than `Box<[T]>`, which in many cases allows us to avoid allocating.

An alternative approach here would be to use a different `Tuple` variant entirely for pure-homogeneous tuples that have no prefix and no suffix: rather than just `Tuple::FixedLength` and `Tuple::VariableLength`, we could have `Tuple::Heterogeneous`, `Tuple::Homogeneous` and `Tuple::Mixed`. I do like that the current design is quite expressive of the fact that a "pure homogeneous" tuple really has exactly the same semantics as a "mixed tuple" with a 0-length prefix and a 0-length suffix, however, so that's an argument for keeping homogeneous tuples and mixed tuples together in one `Tuple` variant.

## Test Plan

Existing tests pass. The mypy primer report is clean.

### DIFF
--- a/crates/ty_python_semantic/src/types/tuple.rs
+++ b/crates/ty_python_semantic/src/types/tuple.rs
@@ -185,7 +185,7 @@ impl<'db> TupleType<'db> {
         if let TupleSpec::Variable(tuple) = tuple {
             if tuple.variable.is_never() {
                 let tuple = TupleSpec::Fixed(FixedLengthTuple::from_elements(
-                    tuple.prefix.iter().chain(&tuple.suffix).copied(),
+                    tuple.prefix().iter().chain(tuple.suffix()).copied(),
                 ));
                 return Some(TupleType::new_internal::<_, TupleSpec<'db>>(db, tuple));
             }
@@ -209,9 +209,9 @@ impl<'db> TupleType<'db> {
     #[cfg(test)]
     pub(crate) fn mixed(
         db: &'db dyn Db,
-        prefix: impl IntoIterator<Item = Type<'db>>,
+        prefix: impl ExactSizeIterator<Item = Type<'db>>,
         variable: Type<'db>,
-        suffix: impl IntoIterator<Item = Type<'db>>,
+        suffix: impl ExactSizeIterator<Item = Type<'db>>,
     ) -> Option<Self> {
         TupleType::new(db, VariableLengthTuple::mixed(prefix, variable, suffix))
     }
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/types/tuple.rs
```diff
@@ -519,38 +527,57 @@ impl<'db> PySlice<'db> for FixedLengthTuple<Type<'db>> {
 ///
 /// Our tuple representation can hold instances of any Rust type. For tuples containing Python
 /// types, use [`TupleSpec`], which defines some additional type-specific methods.
+///
+/// Conceptually, the `prefix` and `suffix` fields could be `Box<[T]>`, but using an `Option` here
+/// allows us to avoid allocating a `Box<[T]>` for the common case of a tuple with no prefix or suffix.
```
MichaReiser, line 532: I don't think `Box<[T]>` allocates. I'm not sure what an empty box slice uses as start-pointer but they can use any value for as long as they don't dereference it when `length` is zero

AlexWaygood: Hmm, this is maybe poorly worded. My point is that are lots of places on `main` where we call `.collect()` on an iterator to create a `Box<[T]>` for the `prefix` or `suffix`, but we don't really need to if we know ahead of time that the iterator is 0-length (which is by far the most common case).

AlexWaygood:
No impact on Codspeed benchmarks `¯\_(ツ)_/¯`

crates/ty_python_semantic/src/types/tuple.rs
MichaReiser: I think this optimization (I'm not sure if it's something we'll see in benchmarks) wouldn't require an `Option<Box<[T]>>`, you could just call `Box::default()` if the iterator is empty. 

However, I'd expect Rust to already short-circuit for you if `sufix` and `prefix` are known to be empty because Rust can use specialization internally (they can have a custom `FromIter` implementation when `I` implements `ExactSizeIterator`).

So I suspect that Rust already does all of that for you

AlexWaygood: Makes sense. Well, it was just an experiment :-)

PR #19894, opened 2025-08-13
[ty] Improve `sys.version_info` special casing

AlexWaygood:
## Summary

Currently the `__getitem__` method we infer for `sys.version_info` is inconsistent with the types we infer for indexing/slicing/unpacking `sys.version_info`. This isn't really a big deal, but... it would be nice to be consistent :-)

The reason for the inconsistency is that we infer `sys.version_info` as inheriting from `tuple[int, int, int, @Todo(Type Alias), int]` due to [typeshed's definition](https://github.com/python/typeshed/blob/1e537d6216d4aef76a8c96ffac9da014ccfe7fc3/stdlib/sys/__init__.pyi#L328), but our special casing elsewhere treats it as a subtype of `tuple[Literal[3], Literal[<minor Python version>], int, Literal["alpha", "beta", ""candidate", "final"], int]`. We can resolve the inconsistency by adding some special-casing for `sys.version_info` to `ClassLiteral::explicit_bases()`.

In theory adding the special casing to `ClassLiteral::explicit_bases()` would let us remove any special casing for `sys.version_info` from `instance.rs`: we could just iterate through the MRO of `sys._version_info` until we find the tuple type in its MRO, and return the tuple spec of that tuple type. In practice, that (unsurpringly) leads to a big performance regression, however. 

I'm adding the "internal" label since I doubt this really changes any behaviour that users would realistically care about.

## Test Plan

Mdtests added

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -20,7 +20,7 @@ use crate::types::function::{DataclassTransformerParams, KnownFunction};
 use crate::types::generics::{GenericContext, Specialization, walk_specialization};
 use crate::types::infer::nearest_enclosing_class;
 use crate::types::signatures::{CallableSignature, Parameter, Parameters, Signature};
-use crate::types::tuple::TupleSpec;
+use crate::types::tuple::{TupleSpec, TupleType};
 use crate::types::{
     BareTypeAliasType, Binding, BoundSuperError, BoundSuperType, CallableType, DataclassParams,
     DeprecatedInstance, KnownInstanceType, StringLiteralType, TypeAliasType, TypeMapping,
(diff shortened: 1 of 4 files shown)


MichaReiser:
We need @carljm to fix the remaining recursive type / generic issues. The conformance test panic is irritating ;) (that's supposed to be pep talk, but I know, I'm not very good at it)

crates/ty_python_semantic/src/types/tuple.rs
```diff
@@ -1162,6 +1162,34 @@ impl<'db> Tuple<Type<'db>> {
             Tuple::Variable(_) => false,
         }
     }
+
+    /// Return the `TupleSpec` for the singleton `sys.version_info`
+    pub(crate) fn version_info_spec(db: &'db dyn Db) -> TupleSpec<'db> {
```
AlexWaygood, line 1167: It's fairly easy to add caching to this method if we revert https://github.com/astral-sh/ruff/commit/d76fd103aefd809440654a4e7a31920cd757eca6, but I still haven't been able to find any evidence yet that this would provide a significant speedup. I also haven't tried doing that as an isolated change on the codspeed runners, though; I've only experimented by running benchmarks locally.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1348,11 +1348,21 @@ impl<'db> ClassLiteral<'db> {
...
             semantic_index(db, self.file(db)).expect_single_definition(class_stmt);
 
-        class_stmt
-            .bases()
-            .iter()
-            .map(|base_node| definition_expression_type(db, class_definition, base_node))
-            .collect()
+        if self.is_known(db, KnownClass::VersionInfo) {
+            let tuple_type = TupleType::new(db, TupleSpec::version_info_spec(db))
+                .expect("sys.version_info tuple spec should always be a valid tuple");
```
MichaReiser, line 1353: Just something to be aware of. This has a risk of lock congestion if `explict_bases` is very hot because we keep interning the same `TupleType`. I don't think `explict_bases` is that hot that this is an issue here but I thought it's worth noting (and it's also something that we might just need to fix in Salsa)

carljm (approved):

crates/ty_python_semantic/src/types/class.rs
carljm: I think this is likely not an issue here because we only intern when we call `explicit_bases` on the `VersionInfo` class specifically. And this method is cached so if we do that frequently, all but the first time in a revision wouldn't reach here anyway.

PR #19644, opened 2025-07-30
Allow hiding the diagnostic severity in `ruff_db`

ntBre:
## Summary

This PR is a spin-off from https://github.com/astral-sh/ruff/pull/19415. It enables replacing the severity and lint name in a ty-style diagnostic:

```
error[unused-import]: `os` imported but unused
```

with the noqa code and optional fix availability icon for a Ruff diagnostic:

```
F401 [*] `os` imported but unused
F821 Undefined name `a`
```

or nothing at all for a Ruff syntax error:

```
SyntaxError: Expected one or more symbol names after import
```

Ruff adds the `SyntaxError` prefix to these messages manually.

Initially (d912458), I just passed a `hide_severity` flag through a bunch of calls to get it into `annotate-snippets`, but after looking at it again today, I think reusing the `None` severity/level gave a nicer result. As I note in a lengthy code comment, I think all of this code should be temporary and reverted when Ruff gets real severities, so hopefully it's okay if it feels a little hacky.

I think the main visible downside of this approach is that we can't style the asterisk in the fix availabilty icon in cyan, as in Ruff's current output. It's part of the message in this PR and any styling gets overwritten in `annotate-snippets`.

image

Hmm, I guess reusing `Level::None` also means the `F401` isn't red anymore. Maybe my initial approach was better after all. In any case, the rest of the PR should be basically the same, it just depends how we want to toggle the severity.

## Test Plan

New `ruff_db` tests. These snapshots should be compared to the two tests just above them (`hide_severity_output` vs `output` and `hide_severity_syntax_errors` against `syntax_errors`).

PR #19896, opened 2025-08-13
Add rule code to GitLab description

### LINKED ISSUE #19881: Change "description" key layout in gitlab output format
Hey! We are using Ruff in our project and have integrated it with Gitlab's codequality reports. Apart from that, we are using a few other tools using the same reports, and we have noticed that we struggle to find what kind of check is reported by Ruff in codequality report, because only the description is displayed there. 

Example: 
`Major - Missing docstring in public package in <file>`

This might be obvious for someone, but we are in the process of setting up a new repo and polishing rule configuration - removing rules that we do not like (we started from enabling all the possible rules). I noticed that pylint, on the other hand, displays it much better.

Example:
`Critical - E1120: No value for argument 'url' in function call in <file>`

The description from pylint contains the actual check name.

How does it look from the code? 

Right now, Ruff has it implemented like this: 
`let value = json!({
                "check_name": check_name,
                "description": description,
                "severity": "major",
                "fingerprint": format!("{:x}", message_fingerprint),
                "location": {
                    "path": path,
                    "positions": {
                        "begin": start_location,
                        "end": end_location,
                    },
                },
            });`

How I would imagine it:

`let value = json!({
                "check_name": check_name,
                "description": check_name + " " + description, # sorry for the syntax, don't know Rust at all
                "severity": "major",
                "fingerprint": format!("{:x}", message_fingerprint),
                "location": {
                    "path": path,
                    "positions": {
                        "begin": start_location,
                        "end": end_location,
                    },
                },
            });`

ntBre:
## Summary

Fixes #19881. While I was here, I also made a couple of related tweaks to the output format. First, we don't need to strip the `SyntaxError: ` prefix anymore since that's not added directly to the diagnostic message after #19644. Second, we can use `secondary_code_or_id` to fall back on the lint ID for syntax errors, which changes the `check_name` from `syntax-error` to `invalid-syntax`. And then the main change requested in the issue, prepending the `check_name` to the description.

## Test Plan

Existing tests and a new screenshot from GitLab:

image

### DIFF
--- a/crates/ruff/tests/snapshots/lint__output_format_gitlab.snap
+++ b/crates/ruff/tests/snapshots/lint__output_format_gitlab.snap
@@ -19,7 +19,7 @@ exit_code: 1
 [
   {
     "check_name": "F401",
-    "description": "`os` imported but unused",
+    "description": "F401: `os` imported but unused",
     "fingerprint": "4dbad37161e65c72",
     "location": {
       "path": "input.py",
@@ -38,7 +38,7 @@ exit_code: 1
   },
   {
     "check_name": "F821",
-    "description": "Undefined name `y`",
+    "description": "F821: Undefined name `y`",
     "fingerprint": "7af59862a085230",
     "location": {
       "path": "input.py",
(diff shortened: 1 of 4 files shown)


MichaReiser (approved):
Nice, thank you

PR #19898, opened 2025-08-13
[ty] resolve docstrings for modules

Gankra:
This also reintroduces the `ResolvedDefinition::Module` variant because reverse-engineering it in several places is a bit confusing. In an ideal world we wouldn't have `ResolvedDefinition::FileWithRange` as it kinda kills the ability to do richer analysis, so I want to chip away at its scope wherever I can (currently it's used to point at asname parts of import statements when doing `ImportAliasResolution::PreserveAliases`, and also keyword arguments).

This also makes a kind of odd change to allow a hover to *only* produce a docstring. This works around an oddity where hovering over a module name in an import fails to resolve to a `ty` even though hovering over uses of that imported name *does*.

The two fixed tests reflect the two interesting cases here.

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -7,7 +7,6 @@ use std::borrow::Cow;
 
 use crate::find_node::covering_node;
 use crate::stub_mapping::StubMapper;
-use ruff_db::files::FileRange;
 use ruff_db::parsed::ParsedModuleRef;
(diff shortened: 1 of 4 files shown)


Gankra:
Screenshot 2025-08-13 at 12 01 04 PM
Screenshot 2025-08-13 at 12 01 14 PM
Screenshot 2025-08-13 at 12 01 25 PM

MichaReiser (approved):

PR #19899, opened 2025-08-13
[ty] Synthesize read-only properties for all declared members on `NamedTuple` classes

AlexWaygood:
## Summary

Currently we treat declared attributes on `NamedTuple` classes just like declared attributes on any other class. But this is not correct! At runtime, the `NamedTuple` machinery synthesizes a read-only property for each attribute declared on a `NamedTuple` class. This PR implements that behaviour.

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/named_tuple.md
+++ b/crates/ty_python_semantic/resources/mdtest/named_tuple.md
@@ -163,7 +163,7 @@ class User(NamedTuple):
     nickname: str
 
 class SuperUser(User):
-    # this should be an error because it implies that the `id` attribute on
+    # TODO: this should be an error because it implies that the `id` attribute on
     # `SuperUser` is mutable, but the read-only `id` property from the superclass
     # has not been overridden in the class body
     id: int
--- a/crates/ty_python_semantic/resources/mdtest/named_tuple.md
+++ b/crates/ty_python_semantic/resources/mdtest/named_tuple.md
@@ -74,8 +74,16 @@ Person(3, "Eve", 99, "extra")
 # error: [invalid-argument-type]
 Person(id="3", name="Eve")
 
-# TODO: over-writing NamedTuple fields should be an error
+reveal_type(Person.id)  # revealed: property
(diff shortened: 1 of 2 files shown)


AlexWaygood:
The new diagnostic in the typing conformance suite is correct -- [this](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/namedtuples_usage.py#L40) is a place where we previously had a false negative

AlexWaygood:
There are some performance regressions of 1-2% on this PR due to the fact that we now need to eagerly check whether a class is an "exact `NamedTuple`" class (one that has `NamedTuple` directly in its class bases) on every attribute access on a class. https://github.com/astral-sh/ruff/pull/19901 (stacked on top of this PR) improves the situation there by adding caching to the mechanism we use for checking whether a class is a `NamedTuple` class.

AlexWaygood:
Only 1% regressions on Codspeed after rebasing on top of https://github.com/astral-sh/ruff/pull/19912.

crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -151,9 +159,42 @@ from typing import NamedTuple
...
     id: int
     name: str
+    age: int | None
+    nickname: str
 
 class SuperUser(User):
-    id: int  # this should be an error
+    # this should be an error because it implies that the `id` attribute on
+    # `SuperUser` is mutable, but the read-only `id` property from the superclass
+    # has not been overridden in the class body
```
carljm: So should there be a TODO here?

AlexWaygood: yeah, though it's a general TODO because declaring an attribute as mutable on a subclass when there's a property on the base class, without actually overriding the property from the base class in the class body of the subclass, should be just as much an error for a non-`NamedTuple` class as it should be for a `NamedTuple` class

PR #19214, opened 2025-07-08
[ty] Improved diagnostic for reassignments of `Final` symbols

sharkdp:
## Summary

Implement [this suggestion](https://github.com/astral-sh/ruff/pull/19178#discussion_r2192658146) by @AlexWaygood.

image

PR #19900, opened 2025-08-13
[`pyflakes`] Add secondary annotation showing previous definition (`F811`)

ntBre:
## Summary

This is a second attempt at a first use of a new diagnostic feature after #19886. I'll blame rustc for this one because it also has a similar diagnostic:

image

We end up with a very similar diagnostic:

image

## Test Plan

New snapshots and manual tests above

### DIFF
--- a/crates/ruff_db/src/diagnostic/render.rs
+++ b/crates/ruff_db/src/diagnostic/render.rs
@@ -264,7 +264,12 @@ impl<'a> ResolvedDiagnostic<'a> {
             .annotations
             .iter()
             .filter_map(|ann| {
-                let path = ann.span.file.path(resolver);
+                let path = ann
+                    .span
+                    .file
+                    .relative_path(resolver)
+                    .to_str()
+                    .unwrap_or_else(|| ann.span.file.path(resolver));
                 let diagnostic_source = ann.span.file.diagnostic_source(resolver);
                 ResolvedAnnotation::new(path, &diagnostic_source, ann, resolver)
             })
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -3349,7 +3349,8 @@ impl Drop for DiagnosticGuard<'_, '_> {
             return;
         }
 
-        if let Some(diagnostic) = self.diagnostic.take() {
+        if let Some(mut diagnostic) = self.diagnostic.take() {
+            diagnostic.sort_sub_diagnostics();
             self.context.diagnostics.borrow_mut().push(diagnostic);
         }
     }
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -28,9 +28,7 @@ use itertools::Itertools;
 use log::debug;
 use rustc_hash::{FxHashMap, FxHashSet};
 
-use ruff_db::diagnostic::{
-    Annotation, Diagnostic, IntoDiagnosticMessage, Span, SubDiagnostic, SubDiagnosticSeverity,
-};
+use ruff_db::diagnostic::{Annotation, Diagnostic, IntoDiagnosticMessage, Span};
 use ruff_diagnostics::{Applicability, Fix, IsolationLevel};
 use ruff_notebook::{CellOffsets, NotebookIndex};
 use ruff_python_ast::helpers::{collect_import_from_member, is_docstring_stmt, to_module_path};
@@ -3308,20 +3306,15 @@ impl DiagnosticGuard<'_, '_> {
         }
     }
 
-    /// Add an "info" sub-diagnostic with the given message and range.
-    ///
-    /// Note that this shadows `Diagnostic::info` from the `Deref` implementation because we'll
-    /// usually want to attach a range here.
-    pub(crate) fn info<'a>(
+    /// Add a secondary annotation with the given message and range.
+    pub(crate) fn secondary_annotation<'a>(
         &mut self,
         message: impl IntoDiagnosticMessage + 'a,
         range: impl Ranged,
     ) {
-        let mut sub = SubDiagnostic::new(SubDiagnosticSeverity::Info, message);
         let span = Span::from(self.context.source_file.clone()).with_range(range.range());
-        let ann = Annotation::primary(span);
-        sub.annotate(ann);
-        self.diagnostic.as_mut().unwrap().sub(sub);
+        let ann = Annotation::secondary(span).message(message);
+        self.diagnostic.as_mut().unwrap().annotate(ann);
     }
 }
 
@@ -3349,8 +3342,7 @@ impl Drop for DiagnosticGuard<'_, '_> {
             return;
         }
 
-        if let Some(mut diagnostic) = self.diagnostic.take() {
-            diagnostic.sort_sub_diagnostics();
+        if let Some(diagnostic) = self.diagnostic.take() {
             self.context.diagnostics.borrow_mut().push(diagnostic);
         }
     }
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -28,7 +28,9 @@ use itertools::Itertools;
 use log::debug;
 use rustc_hash::{FxHashMap, FxHashSet};
 
-use ruff_db::diagnostic::Diagnostic;
+use ruff_db::diagnostic::{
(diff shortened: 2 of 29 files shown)


AlexWaygood:
We have a similar diagnostic in ty for detecting when symbols declared as `Final` are overridden:

image

<br><br>

Although... now that I look at it, it's a bit weird that the "Symbol later reassigned here" annotation is rendered _before_ the "Symbol declared as `Final` here" annotation -- it doesn't really read cleanly as a pair of sentences anymore. Did that change recently with one of your diagnostics refactors? It doesn't look like that's how we originally had them 😆 you can see the screenshot of what they originally looked like in https://github.com/astral-sh/ruff/pull/19214#issue-3213382071

ntBre:
Oh interesting. I don't think that should have changed. Are those two screenshots from the same input? I think the rendering depends on exactly how many lines are between the two snippets.

AlexWaygood:
> Are those two screenshots from the same input? I think the rendering depends on exactly how many lines are between the two snippets.

Great point, they're not! Here's a screenshot on `main` using the same input as [the screenshot on the original PR](https://github.com/astral-sh/ruff/pull/19214#issue-3213382071) -- that still looks great:

image

<br><br>

I guess I should file a bug report at ty to improve that diagnostic if there are many lines between the two snippets (which is probably the common case!)

ntBre:
This might be the relevant code, but I haven't tested it:

https://github.com/astral-sh/ruff/blob/ab685c16a8c01bb33f054698d2f32bb023d07429/crates/ruff_db/src/diagnostic/render.rs#L372-L373

We sort the annotations by their start range earlier in this function but then sort the snippets to put the primary one first, which I think aligns with the behavior in your first screenshot.

MichaReiser:
Nice!

I'm not entirely sure I follow

> They appear to be using a secondary annotation. I kind of prefer a sub-diagnostic because it seems to ensure that the additional info is after the main diagnostic:

Or, maybe it's just that I dislike that part :) @BurntSushi will know better but I think we try to avoid having info/help text between sub diagnostics with snippets because it looks somewhat odd. 

Either way, I do prefer what we have in ty. It's more concise because it doesn't need the extra info text to explain the annotation; the explanation is right in the annotation.

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -3331,7 +3349,8 @@ impl Drop for DiagnosticGuard<'_, '_> {
             return;
         }
 
-        if let Some(diagnostic) = self.diagnostic.take() {
+        if let Some(mut diagnostic) = self.diagnostic.take() {
+            diagnostic.sort_sub_diagnostics();
```
MichaReiser: I'm not sure about this. I think it's important that a lint rule author is in full control about how a diagnostic is rendered. If they decide that a sub diagnostic with lower severity should come first, than so be it. I'm sure there's a reason for it.

ntBre:
I like the ty example too, and it will avoid needing to deal with the sub-diagnostic sorting, at least for now. I'll switch to a secondary annotation!

crates/ruff_linter/src/checkers/ast/mod.rs
ntBre: I know what you mean, I was just trying to work around the automatic `help` sub-diagnostic for every rule that has a fix as easily as possible. I'm assuming we don't want lint authors to have to control _that_, but maybe we can just expose the sub-diagnostic `Vec` or an API that lets them insert additional sub-diagnostics wherever they want. Or another level of `DiagnosticGuard` that pushes the `help`/`fix_title` last instead of immediately when the `Diagnostic` is created from a `Violation`.

MichaReiser: A more strimlined version of `report_diagnostic`, e.g. `report_custom` or similar, seems the most reasonable to me. Ideally, we'd just remove `fix` and `help`, but I believe those fields are used on the website or other output formats? If not, that would be the simplest: just remove `help` and `fix_title` and set them manually

ntBre:
How do y'all feel about the primary annotation (`bar redefined here`)? I kind of like adding one, like rustc and ty:

image

However, this has the downside of changing our `concise` diagnostic for F811. We use `Diagnostic::concise_message`, which concatenates the main diagnostic message and the primary annotation, if both are set. I'm sure we can work around that, but I didn't want to bother if we don't like the primary annotation anyway.

Adding the secondary annotation doesn't really make sense in the concise diagnostic:

```
F811_0.py:10:5: F811 Redefinition of unused `bar` from line 6: `bar` redefined here
```

However, that is exactly what rustc does:

```
$ rustc try.rs --error-format=short
try.rs:5:1: error[E0428]: the name `foo` is defined multiple times: `foo` redefined here
```

I guess ty must do the same too.

AlexWaygood:
I think ideally the concise diagnostic message would not include the text of the primary annotation in _this_ case (since it kinda duplicates the information already in the diagnostic summary). However, I don't think this is _awful_:

```
F811_0.py:10:5: F811 Redefinition of unused `bar` from line 6: `bar` redefined here
```

I'd say it's fine for now!

AlexWaygood:
(and yeah, I think ty does do similarly right now!)

MichaReiser:
A thought triggered by your message

> Adding the secondary annotation doesn't really make sense in the concise diagnostic:

It might be a nice addition for our testing framework to at least capture some concise diagnostics because, unlike you, I don't think many folks (myself included) verify the output manually as you did.

ntBre:
This actually did appear in a CLI snapshot I added in the past, in this case, so I got a bit lucky :) But yes, that sounds like a good idea in general!

ntBre:
We could potentially just dump a second rendering of the diagnostics in `concise` form in `print_messages`:

https://github.com/astral-sh/ruff/blob/22c284aefd8bb2a77a25e43eeb00023fcb129d1f/crates/ruff_linter/src/test.rs#L360-L363

I don't think that would be too bad to look at. It reminds me a bit of the parser tests that print the AST.

MichaReiser:
I guess we have some coverage for concise in the ecosystem checks. It only means that we'd rely on the issue being present in at least one of our ecosystem projects which might be a bit too fragile

crates/ruff_db/src/diagnostic/render.rs
```diff
@@ -264,7 +264,12 @@ impl<'a> ResolvedDiagnostic<'a> {
             .annotations
             .iter()
             .filter_map(|ann| {
-                let path = ann.span.file.path(resolver);
+                let path = ann
+                    .span
+                    .file
+                    .relative_path(resolver)
+                    .to_str()
+                    .unwrap_or_else(|| ann.span.file.path(resolver));
```
BurntSushi, line 272: Out of curiosity, what is the motivation for this change?

ntBre: Just to match the path in `from_diagnostic` above. This isn't used in this PR now that we're using a secondary annotation, so I can revert it!

I'll double-check if it's actually needed for full diagnostics too.

ntBre: Using `path` instead of `relative_path` changes some of our CLI snapshots for the main diagnostic, so I think we'll need this for sub-diagnostics too, once we use any sub-diagnostics with paths in Ruff.

```
    ────────────────────────────────────────────────────────────────────────────────
    -old snapshot
    +new results
    ────────────┬───────────────────────────────────────────────────────────────────
        0     0 │ success: false
        1     1 │ exit_code: 1
        2     2 │ ----- stdout -----
        3     3 │ F401 [*] `bar` imported but unused
        4       │- --> bar.py:2:8
              4 │+ --> /tmp/.tmpJqw1Qi/bar.py:2:8
        5     5 │   |
        6     6 │ 2 | import bar   # unused import
        7     7 │   |        ^^^
        8     8 │   |
        9     9 │ help: Remove unused import: `bar`
       10    10 │ 
       11    11 │ F401 [*] `foo` imported but unused
       12       │- --> foo.py:2:8
             12 │+ --> /tmp/.tmpJqw1Qi/foo.py:2:8
       13    13 │   |
       14    14 │ 2 | import foo   # unused import
       15    15 │   |        ^^^
       16    16 │   |
    ────────────┴───────────────────────────────────────────────────────────────────
```

MichaReiser (approved):

MichaReiser:
This is a nice find for a rule that can make use of the new features because it is one that's very commonly used.

crates/ruff_db/src/diagnostic/render.rs
ntBre: (I think I'll go ahead and leave this instead of reverting, since it seems like it will be useful in the future)

PR #19901, opened 2025-08-13
[ty] Add caching to `CodeGeneratorKind::matches()`

AlexWaygood:
## Summary

Stacked on top of https://github.com/astral-sh/ruff/pull/19899.

There are some small performance regressions in the Codspeed report for #19899 -- this appears to mitigate that!

## Test Plan

Existing tests


carljm (approved):
Were cycles observed here, or are you adding the cycle handling pre-emptively?

AlexWaygood:
Cycles were indeed observed! See e.g. https://github.com/astral-sh/ruff/actions/runs/16944090646/job/48020592511

AlexWaygood:
I split this out as a standalone change in https://github.com/astral-sh/ruff/pull/19912

PR #19903, opened 2025-08-13
[ty] Add some additional type safety to `CycleDetector`

dcreager:
This PR adds a type tag to the `CycleDetector` visitor (and its aliases).

There are some places where we implement e.g. an equivalence check by making a disjointness check. Both `is_equivalent_to` and `is_disjoint_from` use a `PairVisitor` to handle cycles, but they should not use the same viistor. I was finding it tedious to remember when it was appropriate to pass on a visitor and when not to. This adds a `PhantomData` type tag to ensure that we can't pass on one method's visitor to a different method.

For `has_relation` and `apply_type_mapping`, we have an existing type that we can use as the tag. For the other methods, I've added empty structs (`Normalized`, `IsDisjointFrom`, `IsEquivalentTo`) to use as tags.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -170,6 +170,15 @@ fn definition_expression_type<'db>(
     }
 }
 
+/// A type tag for the [`PairVisitor`] that is used in `is_disjoint_from` methods.
+pub(crate) struct IsDisjoint;
+
+/// A type tag for the [`PairVisitor`] that is used in `is_equivalent` methods.
+pub(crate) struct IsEquivalent;
+
+/// A type tag for the [`TypeTransformer`] that is used in `normalized` methods.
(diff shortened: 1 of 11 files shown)


carljm (approved):
This is great, thank you!

As someone who has added a lot of `visitor` arguments to methods, and will be adding a lot more, I think I'd really appreciate some type aliases here, so I can say e.g. `visitor: &NormalizeVisitor` instead of `visitor: &TypeTransformer<'db, Normalized>`. But not a blocker.

dcreager:
> As someone who has added a lot of `visitor` arguments to methods, and will be adding a lot more, I think I'd really appreciate some type aliases here, so I can say e.g. `visitor: &NormalizeVisitor` instead of `visitor: &TypeTransformer<'db, Normalized>`. But not a blocker.

Good idea! Done

PR #19905, opened 2025-08-13
[ty] support goto/hover on string annotations

Gankra:


crates/ty_ide/src/goto_type_definition.rs
```diff
@@ -226,24 +226,7 @@ mod tests {
...
-        894 |     str(bytes_or_buffer[, encoding[, errors]]) -> str
-            |
-        info: Source
-         --> main.py:2:22
-          |
-        2 |             a: str = "test"
-          |                      ^^^^^^
-          |
-        "#);
+        assert_snapshot!(test.goto_type_definition(), @"No goto target found");
```
Gankra: This is the main Problematic thing about this implementation -- we don't ask the type checker if this string actually was a string annotation.

Gankra:
Screenshot 2025-08-13 at 6 39 16 PM

Screenshot 2025-08-13 at 6 39 31 PM

Gankra:
I thought of a way to technically ask the inference engine "hey is this string an annotation?" (ask what the string infers to, and check if that's literally the string it is. if it's not, it must be a type annotation.)

This leaves the only real issue being "way to ask the type of a substring of an annotation". I think this is technically shippable but a bit dubious.

crates/ty_ide/src/goto.rs
```diff
@@ -634,6 +673,58 @@ impl GotoTarget<'_> {
...
             },
 
+            node @ AnyNodeRef::ExprStringLiteral(string_literal) => {
+                // If we encounter a string literal, try to figure out if it's actually
+                // a string annotation by asking the type checker if its type is
+                // a StringLiteral equal to itself
+                let model = SemanticModel::new(db, file);
+                let string_ty = string_literal.inferred_type(&model);
+                if let Type::StringLiteral(string) = string_ty {
+                    if string.value(db) == string_literal.value.to_str() {
```
MichaReiser, line 690: This check will return false for strings longer than 4096 bytes

Gankra:
bitrotted to hell, will restart this from scratch with months of experience under my belt

PR #19906, opened 2025-08-13
[ty] Remove py-fuzzer skips for seeds that are no longer slow

AlexWaygood:
## Summary

These used to take a very long time for ty to check, but it doesn't seem like that's the case anymore!

## Test Plan

- Ran `uvx --force-reinstall --from=./python/py-fuzzer fuzz --bin=ty 0-500`
- Cd'd into `python/py-fuzzer` and ran `uv run --with=mypy mypy`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,7 +38,8 @@ jobs:
       fuzz: ${{ steps.check_fuzzer.outputs.changed }}
       # Flag that is set to "true" when code related to ty changes.
       ty: ${{ steps.check_ty.outputs.changed }}
-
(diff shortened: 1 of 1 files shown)


AlexWaygood:
oh do we not run the fuzzer CI job on PRs that touch the fuzzer?

AlexWaygood:
> oh do we not run the fuzzer CI job on PRs that touch the fuzzer?

Fixed this. The run successfully completed in 1m25s 🎉 https://github.com/astral-sh/ruff/actions/runs/16951581527/job/48045472265?pr=19906

carljm (approved):

PR #19908, opened 2025-08-14
[ty] fix lazy snapshot sweeping in nested scopes

mtshiba:
## Summary

This PR closes astral-sh/ty#955.

## Test Plan

New test cases in `narrowing/conditionals/nested.md`.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/conditionals/nested.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/conditionals/nested.md
@@ -240,6 +240,21 @@ def f(x: str | None):
 
     # When there is a reassignment, any narrowing constraints on the place are invalidated in lazy scopes.
     x = None
+
+def f(x: str | None):
+    def _():
+        if x is not None:
+            def closure():
+                reveal_type(x)  # revealed: str | None
+    x = None
+
+def f(x: str | None):
+    class C:
+        def _():
+            if x is not None:
+                def closure():
+                    reveal_type(x)  # revealed: str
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/narrow/conditionals/nested.md
```diff
@@ -240,6 +240,21 @@ def f(x: str | None):
 
     # When there is a reassignment, any narrowing constraints on the place are invalidated in lazy scopes.
     x = None
+
+def f(x: str | None):
+    def _():
+        if x is not None:
+            def closure():
+                reveal_type(x)  # revealed: str | None
+    x = None
```
mtshiba, line 249: Upon closer analysis, it seems that whether or not narrowing is applied to `x` depends on whether the closure is exposed or just consumed in the scope (This PR does not take this into account and always assumes that the closure is exposed).

```python
def f(x: str | None):
    def _():
        if x is not None:
            def closure():
                reveal_type(x)  # revealed: str | None
            return closure  # expose a closure
    x = None

def f(x: str | None):
    def _():
        if x is not None:
            def closure():
                reveal_type(x)  # revealed: str
            return closure()  # just consume a closure
    x = None
```

To achieve this, an additional mechanism would be required to determine whether a closure is exposed (pyright does not seem to implement this). This would likely be a future work.

carljm: Yes, I think in general we consider escape analysis for closures to be out of scope. Perhaps we could tackle it someday, but it's not needed now.

carljm (approved):
Looks great, thank you!

PR #19780, opened 2025-08-06
[ty] Add `ty.inlayHints.variableTypes` server option

dhruvmanila:
## Summary

This PR adds a new `ty.inlayHints.variableTypes` server setting to configure ty to include / exclude inlay hints at variable position.

Currently, we only support inlay hints at this position so this option basically translates to enabling / disabling inlay hints for now :)

The VS Code extension PR is https://github.com/astral-sh/ty-vscode/pull/112.

closes: astral-sh/ty#472

## Test Plan

Add E2E tests.

PR #17214, opened 2025-04-04
[red-knot] Add inlay type hints

MatthewMckee4:
## Summary

Fixes #17205

## Test Plan

Add tests to red_knot_ide/src/inlay_hints.rs

https://github.com/user-attachments/assets/f3e654c5-d95b-4de3-91ae-602843c160d1

https://github.com/user-attachments/assets/fcc1861a-aa34-49c3-ad6f-fb375dc546d8

PR #19910, opened 2025-08-14
[ty] Default `ty.inlayHints.*` server settings to true

dhruvmanila:
## Summary

This PR changes the default of `ty.inlayHints.*` settings to `true`.

I somehow missed this in my initial PR.

This is marked as `internal` because it's not yet released.

### DIFF
--- a/crates/ty_server/tests/e2e/inlay_hints.rs
+++ b/crates/ty_server/tests/e2e/inlay_hints.rs
@@ -5,6 +5,60 @@ use ty_server::ClientOptions;
 
 use crate::TestServerBuilder;
 
+/// Tests that the default value of inlay hints settings is correct i.e., they're all enabled
+/// by default.
+#[test]
+fn default_inlay_hints() -> Result<()> {
+    let workspace_root = SystemPath::new("src");
+    let foo = SystemPath::new("src/foo.py");
+    let foo_content = "\
+x = 1
+
+def foo(a: int) -> int:
+    return a + 1
+
+foo(1)
+";
+
+    let mut server = TestServerBuilder::new()?
+        .with_initialization_options(ClientOptions::default())
+        .with_workspace(workspace_root, None)?
+        .with_file(foo, foo_content)?
+        .enable_inlay_hints(true)
+        .build()?
+        .wait_until_workspaces_are_initialized()?;
+
+    server.open_text_document(foo, &foo_content, 1);
+    let _ = server.await_notification::<PublishDiagnostics>()?;
+
+    let hints = server
+        .inlay_hints_request(foo, Range::new(Position::new(0, 0), Position::new(6, 0)))?
+        .unwrap();
+
+    insta::assert_json_snapshot!(hints, @r#"
+    [
+      {
+        "position": {
+          "line": 0,
+          "character": 1
+        },
+        "label": ": Literal[1]",
+        "kind": 1
+      },
+      {
+        "position": {
+          "line": 5,
+          "character": 4
+        },
+        "label": "a=",
+        "kind": 1
+      }
+    ]
+    "#);
+
+    Ok(())
+}
+
 /// Tests that disabling variable types inlay hints works correctly.
 #[test]
 fn variable_inlay_hints_disabled() -> Result<()> {


crates/ty_ide/src/inlay_hints.rs
```diff
@@ -86,6 +87,15 @@ pub struct InlayHintSettings {
...
 }
 
+impl Default for InlayHintSettings {
+    fn default() -> Self {
+        Self {
+            variable_types: true,
+            function_argument_names: true,
+        }
+    }
+}
```
MichaReiser, line 97: I'm not entirely sure about enabling all of them by default. E.g. pylance all inlays by default (which I'm also not sure if that's great). We probably want something in between where we enable the non noisy inlays by default but keep all others disabled. 

Having said that. I think defaulting those two settings to true makes sense.

I assume the new default implementation matches the defaults in the VS code settings?

dhruvmanila: Hmm, good point.

I think I have it backwards. Both of them are disabled by default in Pylance.

dhruvmanila: As discussed in our 1:1, we've decided to keep it enabled by default, and can change it in the future before GA based on user feedback.

AlexWaygood: Hmm, I'm a bit surprised by this change, and a bit unhappy with that resolution. We discussed this a fair bit as a team in the original PR adding inlay hints (https://github.com/astral-sh/ruff/pull/17214#issuecomment-2781460823, and there was also discussion on Discord as well IIRC), and I thought we came to a consensus that we should keep them disabled by default

dhruvmanila: Oh, I think I forgot about that discussion, thanks for the reminder. I'll revert this change back for now.

MichaReiser: @AlexWaygood editors have a global setting to opt in/out of inlay hints. I think it's confusing if enabling inlay hints requires both: enabling them globally in the editor and for the extension. In that sense, inlays are disabled by default unless you enable them in your editor.

This behavior also matches rust-analyzers behavior.

AlexWaygood: > This behavior also matches rust-analyzers behavior.

I find inlay hints far more useful when writing Rust than when writing Python, as I described in https://github.com/astral-sh/ruff/pull/17214#issuecomment-2781460823.

Does this behaviour match Pylance? I don't think I've ever toggled the inlay hints setting in VSCode either for the editor or for any specific extension, but I've always seen inlay hints from rust-analyzer when writing Rust, and I've never seen inlay hints from Pylance when writing Python.

MichaReiser: > Does this behaviour match Pylance? 

It does not. Pylance has them off by default. 

I still think we should keep inlays that we find useful and aren't noisy on by default. That's what users opt-in to by enabling inlay hints at the editor level and I rather change our defaults based on user feedback. Which I think we get more of if we enable them by default and it's something we can change very easily moving forward.

E.g. I find the argumet name inlay very useful. Maybe the variable type is less so. But we should discuss them independently

MichaReiser: I think it's a good call out that we should think about why we enable which inlays by default and we can create an issue to revisit this. In the meantime, I'd prefer to defer this discussion to later.

dhruvmanila: That sounds good, I can create an issue and proceed with the release as usual.

AlexWaygood: I'm afraid I'm still of the opinion that this is a change that I would definitely not want to see in my editor with what I understand the current state of this feature as being; that it would take me some time to figure out how to change the settings in VSCode (I can never find my way to the right setting on the rare occasion I need to change something); that this is going to be quite surprising to users coming from Pylance; and that this is going to negatively impact most users of the ty server.

dhruvmanila: Just for context, the variable type hints were enabled by default when it got added and it only got disabled by default in https://github.com/astral-sh/ruff/pull/19780 which I did not intend to make. That was the main motivation of this PR which is to turn it back on to return the behavior in the released ty version.

AlexWaygood: > Just for context, the variable type hints were enabled by default when it got added and it only got disabled by default in #19780 which I did not intend to make. That was the main motivation of this PR which is to turn it back on to return the behavior in the released ty version.

Ahhh, this I did not realise. I'm definitely fine with restoring the status quo, in that case, and continuing the discussion in an issue elsewhere.

PR #19912, opened 2025-08-14
[ty] Add caching to `CodeGeneratorKind::matches()`

AlexWaygood:
https://github.com/astral-sh/ruff/pull/19901 has been approved, but... not the PR that it's stacked on top of :-). This PR experiments to see whether #19901 makes sense as a standalone change.


AlexWaygood:
Speedups of 2-3% on the walltime benchmarks 🥳

PR #19913, opened 2025-08-14
[ty] Speedup project file discovery 

MichaReiser:
## Summary

The `ProjectFilesWalker` discovers all the files to check in a ty project by spawning multiple threads that walk the directory tree.
The walker's output is a `Vec` or `Set` of all the `File`s. The way this works today is that we first collect all file paths (multi threaded)
into a single `Vec`, and the main thread then interns the paths to their `File`'s. 

The problem with this approach is that `system_path_to_file` is fairly expensive when called for many files because it does a `stat` call internally. 

The ideal solution would move the `system_path_to_file` call to the walker threads. That would also eliminate the shared `Vec` that collects all paths. Unfortunately, this isn't possible because `&dyn Db` isn't `Send` nor does `&dyn Db` implement `Clone` (Salsa does have a `fork_db` feature but this isn't publicly exposed as of today). We also can't add a `fork` method to `ty_project::Db` that returns a cloned `Self` because the `Db` trait must be dyn compatible. 

~~The solution I've taken here is to move the stat files into the walker threads, which fixes most of the waterfall effect and is much simpler than exposing `fork_db` properly in salsa.~~

I implemented the ideal solution thanks to @ibraheemdev's suggestion to add a `dyn_clone`!

Fixes https://github.com/astral-sh/ty/issues/970

## Test Plan

Before 
Screenshot 2025-08-14 at 14 04 46

After

Screenshot 2025-08-14 at 14 04 59

### DIFF
--- a/crates/ruff_db/src/files.rs
+++ b/crates/ruff_db/src/files.rs
@@ -89,66 +89,30 @@ impl Files {
             .or_insert_with(|| {
                 let metadata = db.system().path_metadata(path);
 
-                self.create_system_entry(db, absolute, metadata)
+                tracing::trace!("Adding file '{absolute}'");
+
+                let durability = self
+                    .root(db, &absolute)
+                    .map_or(Durability::default(), |root| root.durability(db));
+
+                let builder = File::builder(FilePath::System(absolute))
+                    .durability(durability)
+                    .path_durability(Durability::HIGH);
+
+                let builder = match metadata {
+                    Ok(metadata) if metadata.file_type().is_file() => builder
+                        .permissions(metadata.permissions())
+                        .revision(metadata.revision()),
+                    Ok(metadata) if metadata.file_type().is_directory() => {
+                        builder.status(FileStatus::IsADirectory)
+                    }
+                    _ => builder
+                        .status(FileStatus::NotFound)
+                        .status_durability(Durability::MEDIUM.max(durability)),
+                };
+
+                builder.new(db)
             })
-    }
-
-    /// Looks up or inserts a [`File`] at the given path using the provided metadata.
-    ///
-    ///
-    /// Prefer using [`system_path_to_file`] unless its `stats` call when inserting a new file is a performance
-    /// concern because the metadata is already available or when computing the metadata on different threads.
-    ///
-    /// The metadata is only used when inserting a new file. It doesn't
-    /// update the metadata if a [`File`] for `path` already exists.
-    ///
-    /// This function differs from [`system_path_to_file`] in that it doesn't add a dependency
-    /// on `File.status` and always returns a `File`, even
-    /// if `path` doesn't point to a file (e.g. if `path` is a directory).
-    pub fn system_from_metadata(
-        &self,
-        db: &dyn Db,
-        path: &SystemPath,
-        metadata: crate::system::Metadata,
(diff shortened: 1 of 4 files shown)


crates/ruff_db/src/files.rs
```diff
@@ -87,33 +87,70 @@ impl Files {
...
+    ///
+    /// Prefer using [`system_path_to_file`] unless its `stats` call when inserting a new file is a performance
+    /// concern because the metadata is already available or when computing the metadata on different threads.
+    ///
+    /// The metadata is only used when inserting a new file. It doesn't
+    /// update the metadata if a [`File`] for `path` already exists.
+    ///
+    /// This function differs from [`system_path_to_file`] in that it doesn't add a dependency
+    /// on `File.status` and always returns a `File`, even
+    /// if `path` doesn't point to a file (e.g. if `path` is a directory).
```
BurntSushi: How come? If `metadata` lets you cheaply query the file type, can you still return a `Result` here?

I mention this because it seems like we are otherwise careful not to allow a `File` to be backed by a directory.

crates/ty_project/src/walk.rs
```diff
@@ -232,8 +237,12 @@ impl<'a> ProjectFilesWalker<'a> {
                                 }
                             }
 
-                            let mut paths = paths.lock().unwrap();
-                            paths.push(entry.into_path());
+                            // If this returns `Err`, then the file was deleted between now and when the walk callback was called.
+                            // We can ignore this.
+                            if let Ok(metadata) = system.path_metadata(entry.path()) {
```
BurntSushi: I was wondering why you can't use `ignore::DirEntry::file_type` here, and I think it's because you also need the file permissions? Because directory entries (on Unix) will include the `FileType` free of charge. But not permissions.

BurntSushi (approved):
Nice! This makes sense to me.

ibraheemdev (approved):
> We also can't add a fork method to ty_project::Db that returns a cloned Self because the Db trait must be dyn compatible.

The usual workaround for this is to have a `dyn_clone` method that returns a `Box<dyn Db>`, but the solution here also seems perfectly fine.

crates/ty_project/src/walk.rs
MichaReiser: Yeah, I thought about using the walk dir entry directly, but we, unfortunately, also need the permissions and, more importantly, the last modified time

MichaReiser:
> The usual workaround for this is to have a dyn_clone method that returns a Box<dyn Db>, but the solution here also seems perfectly fine.

Oh nice, that's exactly what I was looking for.

crates/ruff_db/src/files.rs
MichaReiser: Yeah we could. I had a versio where I did just that. 

I considered it to be fine because `Files::try_system` also doesn't return a `Result` but the file directly (even if the path maps to a directory). This where I consider the distinction between `system_path_to_file` which makes sure it adds a query dependency on status (so that your query re-runs if the status changes) and protects you from getting a `File` that points to a directory or doesn't exist. 

Either way, I'm going to revert all of this thanks to @ibraheemdev's suggestion

ibraheemdev (approved):

Issue #12600, opened 2024-07-31
Design display of fixes in `check` when there is no source


chriskrycho:
Spun off from #12595: in some cases, there may be a possible fix but no source, e.g. when passing no input at all (see [here](https://github.com/astral-sh/ruff/pull/12595/files/b824fd328a997e06038aaa8627e06303031f270e#diff-17ba716421a75d2bd60e7328e4b0f368736d61bab68142ac7642d1bfe57efe32) from an earlier version of the PR than the one that landed). Most likely, those fixes should not be emitted at all. However, this needs a bit of design: maybe it *is* legitimate in some cases; and/or perhaps the decision should happen upstream to avoid creating the fix at all?

MichaReiser:
https://github.com/astral-sh/ruff/pull/19919

Issue #12599, opened 2024-07-31
Require rules with fixes to include help message


chriskrycho:
While working on #12595, we noticed that some fixes do not have a corresponding help message. This means that when displaying those to the user, the fix can be displayed with no explanation or context.

An alternative design here might be to fall back to “Suggested fix” or something like that as a stopgap when there is no help message available.

Issue #12598, opened 2024-07-31
Design call to action messages for unsafe and display-only fixes displayed in `check`


chriskrycho:
[content will appear here shortly!]

MichaReiser:
https://github.com/astral-sh/ruff/pull/19919

PR #12595, opened 2024-07-31
Display fixes by default (#7352)

chriskrycho:
## Summary

Use the existing `printer::Flags::SHOW_FIX_DIFF` to enable printing fixable diffs in `check` mode automatically. Enable it whenever the output format is `Full`, but keep a distinct flag for it (a) because output format and whether to print diffs are really separate concerns, (b) because it makes it easy to make changes to what sets the flag to true if/as that is needful, and (c) because it minimally perturbs the rest of the code.

Fixes #7352.

## Details

This changes the result of running `ruff check` by adding the diff for possible fixes, and adds calls to action for those messages.

Note that this currently displays *only* safe fixes to users. There are a number of design questions around when and how to display them; see #12598.

<details><summary>a file with three unused imports</summary>

Input:

```python
import b
import a
import c
```

Output:

```
/Users/chris/Desktop/lol.py:1:8: F401 [*] `b` imported but unused
  |
1 | import b
  |        ^ F401
2 | import a
3 | import c
  |
  = help: Remove unused import: `b`

Suggested fix:
1   |-import b
2 1 | import a
3 2 | import c
4 3 |

    Run `ruff check --fix` to apply this fix.

/Users/chris/Desktop/lol.py:2:8: F401 [*] `a` imported but unused
  |
1 | import b
2 | import a
  |        ^ F401
3 | import c
  |
  = help: Remove unused import: `a`

Suggested fix:
1 1 | import b
2   |-import a
3 2 | import c
4 3 |

    Run `ruff check --fix` to apply this fix.

/Users/chris/Desktop/lol.py:3:8: F401 [*] `c` imported but unused
  |
1 | import b
2 | import a
3 | import c
  |        ^ F401
  |
  = help: Remove unused import: `c`

Suggested fix:
1 1 | import b
2 2 | import a
3   |-import c
4 3 |

    Run `ruff check --fix` to apply this fix.

Found 3 errors.
[*] 3 fixable with the `--fix` option.
```

</details>

<details><summary>The test file <code>crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI025_1.pyi</code></summary>

New output:

```
crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI025_1.pyi:2:40: F401 [*] `collections.abc.Set` imported but unused
  |
1 | def f():
2 |     from collections.abc import Set as AbstractSet  # Ok
  |                                        ^^^^^^^^^^^ F401
3 | 
4 | def f():
  |
[...]

PR #19919, opened 2025-08-14
Show fixes by default

### LINKED ISSUE #7352: CLI: Show fixes by default in `ruff check`
Requires #7350 for opt-out of display.

`ruff check` should display fixes that Ruff can apply. Currently this is only possible for the whole file with `--diff`.

Some fixes may modify content across a file, e.g. when renaming a binding. We should limit the display of fixes to roughly match the range of source context displayed (#7349). If we do not display the entire fix, we should include a notice of the number of additional lines that will be modified if the fix is applied. Naively, we may be able to implement this by only selecting a portion of the fix. In a more comprehensive implementation, we may need to add a new diagnostic range attribute for this display and address it in each problem case.

> Clippy has a nice example user experience of showing fixes:
> 
> 
> ```
> error[E0308]: mismatched types
> --> crates/ruff_python_semantic/src/model.rs:224:21
>     |
> 224 |             source: &self.stmt_id,
>     |                     ^^^^^^^^^^^^^ expected `Option<NodeId>`, found `&Option<NodeId>`
>     |
>     = note:   expected enum `std::option::Option<node::NodeId>`
>             found reference `&std::option::Option<node::NodeId>`
> help: consider removing the borrow
>     |
> 224 -             source: &self.stmt_id,
> 224 +             source: self.stmt_id,
>     |
> 
> For more information about this error, try `rustc --explain E0308`.
> ```

ntBre:
## Summary

This PR fixes #7352 by exposing the `show_fix_diff` option used in our snapshot tests in the CLI. As the issue suggests, we plan to make this the default output format in the future, so this is added to the `full` output format in preview for now.

This turned out to be pretty straightforward. I just used our existing `Applicability` settings to determine whether or not to print the diff. 

The snapshot differences are because we now set `Applicability::DisplayOnly` for our snapshot tests. This `Applicability` is also used to determine whether or not the fix icon (`[*]`) is rendered, so this is now shown for display-only fixes in our snapshots. This was already the case previously, but we were only setting `Applicability::Unsafe` in these tests and ignoring the `Applicability` when rendering fix diffs. CLI users can't enable display-only fixes, so this is only a test change for now, but this should work smoothly if we decide to expose a `--display-only-fixes` flag or similar in the future.

I also deleted the `PrinterFlags::SHOW_FIX_DIFF` flag. This was completely unused before, and it seemed less confusing just to delete it than to enable it in the right place and check it along with the `OutputFormat` and `preview`.

## Test Plan

I only added one CLI test for now. I'm kind of assuming that we have decent coverage of the cases where this shouldn't be firing, especially the `output_format` CLI test, which shows that this definitely doesn't affect non-preview `full` output. I'm happy to add more tests with different combinations of options, if we're worried about any in particular. I did try `--diff` and `--preview` and a few other combinations manually.

And here's a screenshot using our trusty UP049 example from the design discussion confirming that all the colors and other formatting still look as expected:

image

And one with an unsafe fix to see the footer:

image

## Related issues and PR
- https://github.com/astral-sh/ruff/issues/7352
- https://github.com/astral-sh/ruff/pull/12595
- https://github.com/astral-sh/ruff/issues/12598
- https://github.com/astral-sh/ruff/issues/12599
- https://github.com/astral-sh/ruff/issues/12600

I think we could probably close all of these issues now. I think we've either resolved or avoided most of them, and if we encounter them again with the new output format, it would probably make sense to open new ones anyway.

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -349,6 +349,13 @@ impl Diagnostic {
         self.fix().is_some()
     }
 
+    /// Returns `true` if the diagnostic is [`fixable`](Diagnostic::fixable) and applies at the
+    /// configured applicability level.
+    pub fn fix_applies(&self, config: &DisplayDiagnosticConfig) -> bool {
+        self.fix()
+            .is_some_and(|fix| fix.applies(config.fix_applicability))
+    }
+
     /// Returns the offset of the parent statement for this diagnostic if it exists.
     ///
     /// This is primarily used for checking noqa/secondary code suppressions.
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -351,7 +351,7 @@ impl Diagnostic {
 
     /// Returns `true` if the diagnostic is [`fixable`](Diagnostic::fixable) and applies at the
     /// configured applicability level.
-    pub fn fix_applies(&self, config: &DisplayDiagnosticConfig) -> bool {
+    pub fn is_fix_applicable(&self, config: &DisplayDiagnosticConfig) -> bool {
         self.fix()
             .is_some_and(|fix| fix.applies(config.fix_applicability))
     }
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -351,7 +351,7 @@ impl Diagnostic {
 
     /// Returns `true` if the diagnostic is [`fixable`](Diagnostic::fixable) and applies at the
     /// configured applicability level.
-    pub fn is_fix_applicable(&self, config: &DisplayDiagnosticConfig) -> bool {
+    pub fn has_applicable_fix(&self, config: &DisplayDiagnosticConfig) -> bool {
         self.fix()
             .is_some_and(|fix| fix.applies(config.fix_applicability))
     }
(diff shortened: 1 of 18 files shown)


zanieb (approved):

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -349,6 +349,13 @@ impl Diagnostic {
         self.fix().is_some()
     }
 
+    /// Returns `true` if the diagnostic is [`fixable`](Diagnostic::fixable) and applies at the
+    /// configured applicability level.
+    pub fn fix_applies(&self, config: &DisplayDiagnosticConfig) -> bool {
```
dhruvmanila: nit: `fix_applicable` / `is_fix_applicable`

zanieb: I was also thinking about this but landed on not saying something :) I think my best idea was `has_applicable_fix`

zanieb: (I think the `is_applicable` verbiage makes more sense on the fix type)

zanieb: (@ntBre un-resolving because I commented as you were resolving. Just for visibility — I don't have strong feelings)

ntBre: Ah `diagnostic.has_applicable_fix` does read nicely, thanks!

PR #19926, opened 2025-08-15
[ty] Remove incorrect type narrowing for `if type(x) is C[int]`

AlexWaygood:
## Summary

For this code, we currently apply incorrect type narrowing:

```py
class A[T]: ...
class B: ...

def f(x: A[int] | B):
    if type(x) is A[int]:
        reveal_type(x)  # revealed: (A[int] & A[Unknown]) | (B & A[Unknown])
    else:
        reveal_type(x)  # revealed: A[int] | B
```

`type(x)` will either be a subclass of `A` or a subclass of `B` at runtime, but `A[int]` isn't either of those things (it's not a class at all, it's a generic alias), so we shouldn't be doing any type narrowing at all if we see a pattern such as `if type(x) is A[int]`. (Well, we _could_ narrow the type to `Never`, since the condition will always resolve to `False`!)

Note that there is another pre-existing bug here that is not fixed by this PR:

```py
class A[T]: ...
class B: ...

def f(x: A[int] | B):
    if type(x) is A:
        reveal_type(x)  # revealed: Never
    else:
        reveal_type(x)  # revealed: A[int] | B
```

I'm not totally sure why we reveal `Never` for the first branch there. I tried quickly to fix it, but couldn't immediately figure it out, and it's a pre-existing issue so for now I just left it; it doesn't seem high-priority.

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/type.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/type.md
@@ -70,6 +70,56 @@ def _(x: A | B, y: A | C):
         reveal_type(y)  # revealed: A
 ```
 
+## No narrowing for `type(x) is C[int]`
+
+At runtime, `type(x)` will never return a generic alias object (only ever a class-literal object),
+so no narrowing can occur if `type(x)` is compared with a generic alias object.
+
+```toml
+[environment]
+python-version = "3.12"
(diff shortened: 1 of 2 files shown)


carljm (approved):
Thanks!

PR #19931, opened 2025-08-15
[ty] Implement partial stubs

Gankra:
Fixes #https://github.com/astral-sh/ty/issues/184

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/resolver.rs
+++ b/crates/ty_python_semantic/src/module_resolver/resolver.rs
@@ -684,7 +684,7 @@ fn resolve_name(db: &dyn Db, name: &ModuleName, mode: ModuleResolveMode) -> Opti
 
         if !search_path.is_standard_library() && resolver_state.mode.stubs_allowed() {
             match resolve_name_in_search_path(&resolver_state, &stub_name, search_path) {
-                Ok((package_kind, ResolvedName::FileModule(module))) => {
+                Ok((package_kind, _, ResolvedName::FileModule(module))) => {
                     if package_kind.is_root() && module.kind.is_module() {
                         tracing::trace!(
                             "Search path `{search_path}` contains a module \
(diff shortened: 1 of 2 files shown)


Gankra:
(I need to write some proper tests but this fixes `from authlib.integrations.requests_client import OAuth2Session`)

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -796,7 +809,7 @@ fn resolve_name_in_search_path(
     context: &ResolverContext,
     name: &RelaxedModuleName,
     search_path: &SearchPath,
-) -> Result<(PackageKind, ResolvedName), PackageKind> {
+) -> Result<(PackageKind, PyTyped, ResolvedName), (PackageKind, PyTyped)> {
```
Gankra, line 812: This is really gross and I considered factoring a proper type *but* literally only one function deals with this and immediately unpacks these so... eh?

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1006,6 +1030,32 @@ impl PackageKind {
...
+    /// Inherit py.typed info from the parent package
+    ///
+    /// > This marker applies recursively: if a top-level package includes it,
+    /// > all its sub-packages MUST support type checking as well.
+    ///
+    /// This implementation implies that once a `py.typed` is specified
+    /// all child packages inherit it, so they can never become Untyped.
+    /// However they can override whether that's Full or Partial by
+    /// redeclaring a `py.typed` file of their own.
+    fn inherit_parent(self, parent: Self) -> Self {
```
Gankra, line 1054: No idea if this is correct but i like it conceptually

carljm: The spec isn't totally clear here, it just says "This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well." It doesn't say anything about a sub-package going to `partial` where its parent was "full", or whatever. But I think this behavior is sensible.

carljm (commented):
Looks reasonable to me! But yeah, should add some tests before we land it.

Gankra:
Added tests

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1426,6 +1476,234 @@ mod tests {
         );
     }
 
+    /// Testing stub packages can be missing modules and we don't fallthrough
+    /// to the impl (when they don't declare py.typed partial)
+    #[test]
+    fn resolve_stub_module_missing_non_partial() {
```
MichaReiser: Could this be an mdtest similar to https://github.com/astral-sh/ruff/blob/1ba56b4bc65442c40d7d6f2873367380d37241a8/crates/ty_python_semantic/resources/mdtest/import/stub_packages.md#L35-L58

BurntSushi:
> Can you coordinate with @BurntSushi to make sure partial is also considered for completions (#19883)?

I added https://github.com/astral-sh/ty/issues/1042 to track this.

Gankra:
Oh sure I can make these mdtests.

PR #19321, opened 2025-07-14
[ty] improve lazy scope place lookup

mtshiba:
## Summary

This PR improves place (symbol) lookup in lazy scopes.

This PR has two main changes:
* Currently, we suppose the type of a nonlocal symbol as the union of the types of all reachable bindings and `Unknown` (#18750), but if the defined scope is private (e.g., function scope), we can exclude `Unknown` (`Unknown` is only necessary when the symbol is writable from the outside), so we will do so.
* Currently, we don't narrow nonlocal symbols (by constraints introduced in the outer scope) in lazy nested scopes, but if the symbol has never been modified and the defined scope is private, we can narrow it, so we will do so.

closes astral-sh/ty#659

## Test Plan

New tests in `narrow/conditionals/nested.md`.

PR #19932, opened 2025-08-15
[ty] more precise lazy scope place lookup

mtshiba:
## Summary

This is a follow-up to https://github.com/astral-sh/ruff/pull/19321.

The timing of sweeping lazy snapshots has been changed from the end of the scope to when the symbol is reassigned, which makes symbol lookup in lazy scopes more precise.

There's still work to be done to make the lookup more complete, but some of the TODOs can be eliminated with this simple change, so I'm submitting this PR for now.

Closes astral-sh/ty#559.

## Test Plan

Some TODOs in `public_types.md` now work properly.

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -400,28 +400,64 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
         }
     }
 
-    /// Any lazy snapshots of places that have been reassigned are no longer valid, so delete them.
-    fn sweep_lazy_snapshots(&mut self, popped_scope_id: FileScopeId) {
+    /// Any lazy snapshots of the place that have been reassigned are no longer valid, so delete them.
+    /// If there is no reassignment, the state of the symbol just before the closure definition is recorded as a lazy snapshot.
+    /// We have not yet tracked where the closure is actually referenced, but we can assume that the closure becomes referenceable after the closure is defined.
+    /// ```py
+    /// def outer() -> None:
+    ///     x = None
+    ///
+    ///     def inner2() -> None:
+    ///         # `inner` can be referenced before its definition,
+    ///         # but `inner2` must still be called after the definition of `inner` for this call to be valid.
+    ///         inner()
+    ///
+    ///         # In this scope, `x` may refer to `x = None` or `x = 1`.
+    ///         reveal_type(x)  # revealed: None | Literal[1]
+    ///
+    ///     # Reassignment of `x` after the definition of `inner2`.
+    ///     # Sweep lazy snapshots of `x` for `inner2`.
+    ///     x = 1
+    ///
+    ///     def inner() -> None:
+    ///         # In this scope, `x = None` appears as being shadowed by `x = 1`.
+    ///         reveal_type(x)  # revealed: Literal[1]
+    ///
+    ///     # No reassignment of `x` after the definition of `inner`, so we can safely record a lazy snapshot for `inner`.
+    ///     inner()
+    ///     inner2()
+    /// ```
+    fn sweep_lazy_snapshots(&mut self, bound_place: ScopedPlaceId) {
+        let current_scope = self.current_scope();
+        let current_place_table = &self.place_tables[current_scope];
+        let Some(bound_symbol) = bound_place.as_symbol() else {
+            return;
+        };
         // Retain only snapshots that are either eager
-        //     || (enclosing_scope != popped_scope && popped_scope is not a visible ancestor of enclosing_scope)
+        //     || (enclosing_scope != current_scope && current_scope is not a visible ancestor of enclosing_scope)
         //     || enclosing_place is not a symbol or not reassigned
         // <=> remove those that are lazy
-        //     && (enclosing_scope == popped_scope || popped_scope is a visible ancestor of enclosing_scope)
+        //     && (enclosing_scope == current_scope || current_scope is a visible ancestor of enclosing_scope)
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -6772,13 +6775,10 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                 ConstraintKey::NarrowingConstraint(constraint),
                             ));
                         }
-                        EnclosingSnapshotResult::FoundBindings(bindings) => {
-                            if place_expr.is_symbol()
-                                && !enclosing_scope_id.is_function_like(db)
-                                && !is_immediately_enclosing_scope
-                            {
-                                continue;
-                            }
```
mtshiba, line 6943: Not so related to the main point of this PR, but I realized that this branch is actually unnecessary. For the condition here to be met, the enclosing scope needs to be non-function-like, i.e., class scope or module scope, and in that case, this case cannot be reached in the first place.
This is because in module scopes, snapshots are not recorded, and in class scopes, constraints are recorded instead of bindings.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -473,7 +504,8 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
 
         if popped_scope.is_eager() {
             self.record_eager_snapshots(popped_scope_id);
-        } else {
+        } else if !popped_scope.kind().is_type_alias() {
+            // Don't record lazy snapshots in a type alias scope, as it is a deferred evaluation context.
```
mtshiba: This is just a shortcut for performance optimization.
`infer_name_load` only references the enclosing snapshot if `!self.is_deferred()` is true.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -750,6 +782,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
     /// Records that all remaining statements in the current block are unreachable.
     fn mark_unreachable(&mut self) {
         self.current_use_def_map_mut().mark_unreachable();
+        self.check_lazy_snapshots(self.current_scope());
```
mtshiba: We compare terminal (!= end-of-scope) bindings with snapshotted (captured) bindings to determine the completeness of the snapshot. We need to call `check_lazy_snapshots` even when the scope becomes unreachable.

crates/ty_python_semantic/src/semantic_index/builder.rs
carljm: This looked a little odd to me, and I realized we are using `DeferredExpressionState::Deferred` wrongly in this case. Pushed https://github.com/astral-sh/ruff/pull/20086 to fix, which means I think this shortcut should be removed in this PR. I assume this shouldn't be a major performance point, since PEP 695 type aliases aren't common in the ecosystem.

carljm (commented):
Thank you!

Took a quick look, made one comment -- I want to take a closer look at some of the changes in use-def map, will try to do that tomorrow.

crates/ty_python_semantic/src/semantic_index/use_def.rs
```diff
@@ -690,7 +768,7 @@ pub(crate) struct BindingWithConstraintsIterator<'map, 'db> {
     pub(crate) predicates: &'map Predicates<'db>,
```
carljm: There's a doc comment just above here on the `EnclosingSnapshots` type alias that I think is no longer accurate?

crates/ty_python_semantic/src/semantic_index/use_def/place_state.rs
```diff
@@ -228,6 +242,26 @@ impl Bindings {
...
+        self.live_bindings
+            .iter()
+            .find(|live_binding| live_binding.binding == id)
+            .map_or_else(
+                || {
+                    if self.live_bindings.last().is_some_and(|b| b.binding > id) {
+                        LiveBindingResult::Shadowed
+                    } else {
+                        LiveBindingResult::NotFound
+                    }
```
carljm: Making this distinction on the basis of binding id numbering is not right. It's using source-order incorrectly as a proxy for control flow, and it means we get inconsistent results on a case like this:

```py
def outer(flag: bool) -> None:
    if flag:
        x = 1
        return
    else:
        x = None

    def inner() -> None:
        reveal_type(x)  # revealed: None | Literal[2]
    inner()

    x = 2
```

The above assertion works, but if we just swap the order of the if/else clauses (which shouldn't change anything), now we get a different result:

```py
def outer(flag: bool) -> None:
    if flag:
        x = None
    else:
        x = 1
        return

    def inner() -> None:
        reveal_type(x)  # revealed: None | Literal[1, 2]
    inner()

    x = 2
```

Now we wrongly consider `x = 1` as a possibly-reachable binding, just because there's no live binding after it in source order in the snapshot bindings.

The question we really want to ask here is whether binding `id` is reachable from the point in control flow where the snapshot was taken.

carljm (changes requested):
Thanks for your work on this, and sorry it took me a while to complete review! I think this is an important issue to solve, and this definitely improves the behavior. That said, I'm not convinced that the approach taken is really correct. I commented inline on the specific problematic point. Currently this PR marks snapshots as "incomplete" and then later tries to figure out how to "merge" that incomplete snapshot with all-reachable-bindings. This makes it quite difficult to figure out which of the "all-reachable" bindings should be considered vs ignored.

I think a more robust approach here might be to actually update the snapshot with new bindings, as we encounter new bindings after the snapshot was created.

Issue #19153, opened 2025-07-05
PERF403 fix has two problems related to tuples


dscorbett:
### Summary

The fix for [`manual-dict-comprehension` (PERF403)](https://docs.astral.sh/ruff/rules/manual-dict-comprehension/) has two problems related to tuples.

Given a singleton tuple key, the fix changes behavior. [Example](https://play.ruff.rs/a242190c-ec87-4bfa-8423-fb3f87c4cf0a):
```console
$ cat >perf403_1.py <<'# EOF'
def f():
    v = {}
    for o, (x,) in ["ox"]:
        v[x,] = o
    return v
print(f())
# EOF

$ python perf403_1.py
{('x',): 'o'}

$ ruff --isolated check perf403_1.py --select PERF403 --preview --unsafe-fixes --fix
Found 1 error (1 fixed, 0 remaining).

$ cat perf403_1.py
def f():
    v = {x: o for o, (x,) in ["ox"]}
    return v
print(f())

$ python perf403_1.py
{'x': 'o'}
```

Given an unparenthesized tuple value, the fix introduces a syntax error. [Example](https://play.ruff.rs/54d03b7b-dd7c-4297-bacb-fa735245fd10):
```console
$ cat >perf403_2.py <<'# EOF'
def f():
    v = {}
    for (o, p), x in [("op", "x")]:
        v[x] = o, p
# EOF

$ ruff --isolated check perf403_2.py --select PERF403 --preview --unsafe-fixes --diff 2>&1 | grep error:
error: Fix introduced a syntax error. Reverting all changes.
```

### Version

ruff 0.12.2 (9bee8376a 2025-07-03)

liortct:
Hey,
I’d like to fix this issue and open a PR.

sudormrfbin:
Can this issue be closed now since #19934 is merged?

ntBre:
I think so, thank you!

PR #19934, opened 2025-08-16
[`perflint`] Handle tuples in dictionary comprehensions (`PERF403`)

liortct:
This pull request fixes the bug described in issue [#19153](https://github.com/astral-sh/ruff/issues/19153).

The issue occurred when `PERF403` incorrectly flagged cases involving tuple unpacking in a for loop. For example:

```python
def f():
    v = {}
    for (o, p), x in [("op", "x")]:
        v[x] = o, p
```

This code was wrongly suggested to be rewritten into a dictionary comprehension, which changes the semantics.

Changes in this PR:

Updated the `PERF403` rule to correctly handle tuple unpacking in loop targets.

Added regression tests to ensure this case (and similar ones) are no longer flagged incorrectly.

Why:
This ensures that `PERF403` only triggers when a dictionary comprehension is semantically equivalent to the original loop, preventing false positives.


liortct:
@ntBre 
Hello 👋 just a gentle reminder about this PR. I’d be happy to make any changes if needed. Thanks for reviewing when you get the chance!

ntBre:
Thank you! I'll try to get to it this week :)

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
```diff
@@ -354,19 +354,26 @@ fn convert_to_dict_comprehension(
...
-        expr.range()
+    let key_range = key.range();
+    let key_str = locator.slice(key_range);
+    // If key_str ends with a comma and lacks outer parentheses, add them.
+    let key_str = if key_str.ends_with(',') && !key_str.starts_with('(') {
+        format!("({key_str})")
     } else {
-        key.range()
+        key_str.to_string()
     };
```
ntBre: What do you think about something like this?


```suggestion
    let key_str = if let Expr::Tuple(ast::ExprTuple {
        elts,
        parenthesized,
        ..
    }) = key
    {
        if elts.len() != 1 {
            return None;
        };

        if *parenthesized {
            locator.slice(key).to_string()
        } else {
            format!("({})", locator.slice(key))
        }
    } else {
        locator.slice(key).to_string()
    };
```

That uses the same kind of `parenthesized` check as for `value_str`. I also used a small trick with `slice`: `Expr`s implement `Ranged` directly, so you can just pass them instead of having to call `key.range()`, for example. It also restores the `elts` check, although I don't really see why it shouldn't apply if there are multiple elements in the slice. That's also checked here, but I guess we might as well preserve it:

https://github.com/astral-sh/ruff/blob/59b078b1bf3a9482e08fe5a00a41076efbbc3974/crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs#L130-L139

It's a bit unfortunate to have to call `format!` and `to_string` everywhere, but I believe that is the typical way to add parentheses when needed. We _could_ use a `Cow` if we really want to avoid it, but I think it's probably okay.

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
```diff
@@ -354,19 +354,26 @@ fn convert_to_dict_comprehension(
...
+    let mut value_str = locator.slice(value.range()).to_string();
+
+    // If the value is a tuple without parentheses, add them
+    if let Expr::Tuple(ast::ExprTuple {
+        parenthesized: false,
+        ..
+    }) = value
+    {
+        value_str = format!("({value_str})");
+    }
```
ntBre: Similar to the case above, I think we can just do this instead of having a mutable variable:


```suggestion
    // If the value is a tuple without parentheses, add them
    let value_str = if let Expr::Tuple(ast::ExprTuple {
        parenthesized: false,
        ..
    }) = value
    {
        format!("({})", locator.slice(value))
    } else {
        locator.slice(value).to_string()
    };
```

crates/ruff_linter/resources/test/fixtures/perflint/PERF403.py
```diff
@@ -192,3 +192,17 @@ def issue_19005_3():
     c = {}
     for a[0], a[1] in ():
         c[a[0]] = a[1]
+
+
+def issue_19153_1():
+    v = {}
+    for o, (x,) in ["ox"]:
+        v[x,] = o
+    return v
```
ntBre, line 201: Your implementation already covers this (nice work!), but we may also want to add a test for the case where the key is already parenthesized just to make sure we don't add a second set:

```py
def issue_19153_3():
    v = {}
    for o, (x,) in ["ox"]:
        v[(x,)] = o
    return v
```

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
```diff
@@ -354,19 +354,26 @@ fn convert_to_dict_comprehension(
...
+
+    let elt_str = format!("{key_str}: {value_str}");
```
ntBre: This call already existed, but if we wanted to avoid one string allocation, we could fold this into the `format!` below:

```rust
let comprehension_str = format!("{{{key_str}: {value_str} {for_type} {target_str} in {iter_str}{if_str}}}");
```

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
liortct: Thanks for the comment, Fixed!

crates/ruff_linter/resources/test/fixtures/perflint/PERF403.py
liortct: You are absolutely right, I added this test case.

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
liortct: Fixed

crates/ruff_linter/src/rules/perflint/rules/manual_dict_comprehension.rs
liortct: Nice tip, thanks.

liortct:
@ntBre 
Thanks a lot for the review and comments!  
I’ve addressed all the issues, and the PR is ready for another review.  
If there’s anything else that needs to be adjusted, please let me know and I’ll update it promptly.  

Thanks again!

ntBre (approved):
Thank you!

PR #19936, opened 2025-08-16
[ty] Fix subtyping/assignability for `@property` protocol members

### LINKED ISSUE #1379: Properly support `@property` members on protocols
If a protocol `P` has an `@property` member `foo`, we currently say that any nominal type `N` will satisfy `P`'s interface if `N` has an attribute `foo`. This is obviously incorrect; we should also consider the type of the member `foo` on the protocol. For `@property` members with setters, we should also check whether the `foo` attribute can be set on `N` with the type the setter specifies.

Detailed tests for this are already in place at https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/protocols.md

### LINKED ISSUE #1380: Support `ClassVar` protocol members
Consider the following two protocols:

```py
from typing import Protocol, ClassVar

class P1(Protocol):
    x: int

class P2(Protocol):
    x: ClassVar[int]
```

In order for a type `N` to satisy `P1`'s interface, an `x` attribute needs to be readable with type `int` on instances of `N` and also writable with type `int` on instances of `N`. The `x` attribute does not need to be readable or writable on instances of `type[N]`.

In order for a type `N` to satsify `P2`'s interface, however, an `s` attribute needs to be readable with type `int` on instances of `N` _and_ instances of `type[N]`. It does not need to be writable on instances of `N`, but it _does_ need to be writable on instances of `type[N]`.

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1379, fixes https://github.com/astral-sh/ty/issues/1380


charliermarsh:
This merged as https://github.com/astral-sh/ruff/pull/25332.

PR #17896, opened 2025-05-06
Remove R from redirects, as it is not a valid rule

VanOvermeire:
## Summary

This PR should fix the following bug: https://github.com/astral-sh/ruff/issues/17890

## Test Plan

Tested against a local repository. main branch `ruff check --select R` works, with this redirect removed, `ruff check --select R` throws `error: invalid value 'R' for '--select <RULE_CODE>'`

All existing tests still pass. 

Not sure if it is worthwhile/necessary to add a unit test for this fix?

Issue #17890, opened 2025-05-06
R is valid prefix at runtime but not included in JSON schema


xixixao:
### Summary

**Problem**: `ruff check --select R` works, so `R` seems to be a valid rule selector.
But neither https://github.com/astral-sh/ruff/blob/main/ruff.schema.json nor https://github.com/SchemaStore/schemastore/edit/master/src/schemas/json/ruff.json include "R" in the list of valid select rules. This leads to `pyproject.toml` that includes "R" in `lint.select` items to error in IDEs.

**Fix**: I didn't open a PR because I don't know whether R is or isn't supposed to be valid.

**Workaround**: Use the more explicit `"RET"`, which is the only rule set that R enables.

### Version

ruff 0.11.8 (75effb8ed 2025-05-01)

xixixao:
In fact I just debugged it and "R" only enables "RET". Edited my first comment. Seems more likely that "R" at runtime is a bug (or perhaps a legacy behavior?).

MichaReiser:
Yeah, it may be a relict from before we had `RUF` rules. That seems even worse :)

VanOvermeire:
I was just exploring your tool and codebase when I saw this issue. Maybe I've misunderstood, but this seems like a one-line fix?

https://github.com/astral-sh/ruff/pull/17896

If so, apologies for stealing your bug @ntBre :D

bitloi:
@dhruvmanila Sorry to bother you, I would like to work on this issue. Is it possible?

ntBre:
@bitloi There's already a PR implementing the first step of this process in https://github.com/astral-sh/ruff/pull/19939. I think this issue is waiting on a decision from us, not on anyone to implement it, but thank you for your interest.

bitloi:
The PR was opened last year, but not finished yet. Is this normal?

Nish2005karsh:
I'd like to help move this forward. I see #17896 and #19939 are pending — is the team leaning toward (a) adding "R" to the schema, (b) removing "R" from runtime, or (c) warning users? Happy to implement whatever direction you decide.

PR #19939, opened 2025-08-16
Add warning for using R and S as selectors

GDYendell:
## Summary

Following on from discussion in https://github.com/astral-sh/ruff/pull/17896, this change adds a user deprecation warning for selectors that currently work, but are not included in the schema. I believe R and S are the only ones.

## Test Plan

I have tested this locally with 

```
ruff check --select R
```

However it does not display the warning. If I change to `println!` it does. Is there something I am not understanding about the `warn_user*` macros?

### DIFF
--- a/crates/ruff_linter/src/rule_selector.rs
+++ b/crates/ruff_linter/src/rule_selector.rs
@@ -10,9 +10,6 @@ use crate::codes::{RuleCodePrefix, RuleGroup};
 use crate::registry::{Linter, Rule, RuleNamespace};
 use crate::rule_redirects::get_redirect;
 use crate::settings::types::PreviewMode;
-use crate::warn_user_once;
-
-const DEPRECATED_SELECTORS: [&str; 2] = ["R", "U"];
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum RuleSelector {
@@ -68,12 +65,6 @@ impl FromStr for RuleSelector {
             "C" => Ok(Self::C),
             "T" => Ok(Self::T),
             _ => {
-                if DEPRECATED_SELECTORS.contains(&s) {
-                    warn_user_once!(
-                        "Using `{s}` as a selector is deprecated and will be removed in a future version."
-                    );
-                }
-
                 let (s, redirected_from) = match get_redirect(s) {
                     Some((from, target)) => (target, Some(from)),
                     None => (s, None),
--- a/crates/ruff_linter/src/rule_redirects.rs
+++ b/crates/ruff_linter/src/rule_redirects.rs
@@ -1,14 +1,30 @@
 use std::collections::HashMap;
 use std::sync::LazyLock;
 
+use crate::warn_user_once;
+
+const DEPRECATED_SELECTORS: [&str; 3] = ["R", "U", "IC"];
+
+fn check_for_deprecation(code: &str) {
+    if DEPRECATED_SELECTORS.contains(&code) {
+        warn_user_once!(
+            "Using `{code}` as a selector is deprecated and will be removed in a future version."
+        );
+    }
+}
+
 /// Returns the redirect target for the given code.
 pub(crate) fn get_redirect_target(code: &str) -> Option<&'static str> {
+    check_for_deprecation(code);
+
     REDIRECTS.get(code).copied()
 }
 
 /// Returns the code and the redirect target if the given code is a redirect.
 /// (The same code is returned to obtain it with a static lifetime).
 pub(crate) fn get_redirect(code: &str) -> Option<(&'static str, &'static str)> {
+    check_for_deprecation(code);
+
     REDIRECTS.get_key_value(code).map(|(k, v)| (*k, *v))
 }
 


crates/ruff_linter/src/rule_selector.rs
```diff
@@ -10,6 +10,9 @@ use crate::codes::{RuleCodePrefix, RuleGroup};
 use crate::registry::{Linter, Rule, RuleNamespace};
 use crate::rule_redirects::get_redirect;
 use crate::settings::types::PreviewMode;
+use crate::warn_user_once;
+
+const DEPRECATED_SELECTORS: [&str; 2] = ["R", "U"];
```
ntBre: Do you think it would make more sense to put this const and the warning in `rule_redirects.rs` next to the `REDIRECTS` hash map? Then the warning could be in `get_redirect` and possibly also `get_redirect_target`, which are where the redirects are looked up. That seems like a better fit to me than in the `FromStr` implementation.

I also mentioned the IC -> ICN redirect in my [comment](https://github.com/astral-sh/ruff/pull/17896/files#r2078434064) on the other PR. Should we add that here too? And there are several other `REDIRECTS` with TODO comments. I think it makes sense to try to cover them all in one PR. I think we can emit a warning for any redirect not present in the schema, as I also mention in the comment, based on the initial issue.

GDYendell: > Do you think it would make more sense to put this const and the warning in rule_redirects.rs next to the REDIRECTS hash map? Then the warning could be in get_redirect and possibly also get_redirect_target, which are where the redirects are looked up. That seems like a better fit to me than in the FromStr implementation.

Yep this makes much more sense!

> I also mentioned the IC -> ICN redirect in my [comment](https://github.com/astral-sh/ruff/pull/17896/files#r2078434064) on the other PR. Should we add that here too? And there are several other REDIRECTS with TODO comments. I think it makes sense to try to cover them all in one PR. I think we can emit a warning for any redirect not present in the schema, as I also mention in the comment, based on the initial issue.

Ah yes sorry, I was only checking the single letter redirects initially. So, any redirect that is not in the schema should be deprecated? That is quite a few of them!

MichaReiser: We might want to move this into `LintConfiguration`. It seems to be where we have all other deprecation warnings:

https://github.com/astral-sh/ruff/blob/e917d309f1d37887a9f5ea11698389caa6c76b11/crates/ruff_workspace/src/configuration.rs#L1044-L1052

ntBre: > So, any redirect that is not in the schema should be deprecated? That is quite a few of them!

I think this is true, that's how I was checking before, but maybe Micha can confirm.

MichaReiser: Do any of them appear in the schema? 

I'd focus on rules that are prefixes of other groups, because that can be confusing (e.g. `R` is a prefix of both `RET` and `RUF` but it only enables `RET`). I don't see much harm in supporting redirects that don't overlap with any other rule code and deprecating them does cause some churn for our users that I'd avoid unless we have good reason for it

ntBre: I just spot-checked a few more, and I don't think any of them appear in the schema, so it's fine with me to focus on the confusing ones. The original issue (https://github.com/astral-sh/ruff/issues/17890) mentioned that the mismatch with the schema was causing errors in IDEs (in the pyproject.toml?), which I think is where I got the schema idea.

MichaReiser: > mentioned that the mismatch with the schema was causing errors in IDEs (in the pyproject.toml?)

That's something we can solve separately by including those names in the schema (but mark them as deprecated)

GDYendell: > We might want to move this into LintConfiguration. It seems to be where we have all other deprecation warnings:

I think this would be too late. It is currently checking just before the redirect is applied during clap parsing
```
ruff_linter::rule_redirects::get_redirect (/home/gdy/development/ruff/crates/ruff_linter/src/rule_redirects.rs:26)
<ruff_linter::rule_selector::RuleSelector as core::str::traits::FromStr>::from_str (/home/gdy/development/ruff/crates/ruff_linter/src/rule_selector.rs:68)
...
clap_builder::derive::Parser::parse_from (/home/gdy/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.43/src/derive.rs:56)
ruff::main (/home/gdy/development/ruff/crates/ruff/src/main.rs:44)
```

after which "R" is converted to "RET". Unless it is possible to get the original command line arguments somehow? The map of redirects is empty here if I run with `check --select R`.

MichaReiser: Hmm I see. It would require validating the clap arguments (and probably options too?) to when we resolve the settings but that's obviously a much larger change. 

The main downside I see of the current solution is that it isn't obvious to users where they used the deprecated code, making it difficult for them to act on the error. 

Could we either:

* Add an argument to `get_redirect` which tells where the redirect is used (e.g. CLI, Config(Path)) so that we can display the source as part of the error message
* Change the return type to `Redirect::To(String)`/`Redirect::Deprecated(String)` depending on whether the redirect is deprecated and use custom errors at each call site

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #19939). The diff contains 35 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #19939). The diff contains 35 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

PR #19942, opened 2025-08-16
[`ruff`] Add `logging-eager-conversion` (`RUF065`)

### LINKED ISSUE #12734: Rule Request: redundant logfmt str ctor catch in G rules or add new rule
This is a really common antipattern I wish logfmt (G*) rules would catch:
```python
   log.warning("Cannot write to %s", str(profile_path))
```

is redundant and does an potentially expensive str construction which is what we are trying to avoid with these rules anyway. Instead, the `%s` ensure it will be formatted as a string anyway! As such, this can be simplified into the better code:
```python
   log.warning("Cannot write to %s", profile_path)
```
Which will be faster in both the case where it's printed and the case where it's not by either avoiding an unnecessary string copy or a unnecessary construction in the first place.

This rule could be added more generically to any string formatting in the new or old styles

GDYendell:
## Summary

Fixes #12734

I have started with simply checking if any arguments that are providing extra values to the log message are calls to `str` or `repr`, as suggested in the linked issue. There was a concern that this could cause false positives and the check should be more explicit. I am happy to look into that if I have some further examples to work with.

If this is the accepted solution then there are more cases to add to the test and it should possibly also do test for the same behavior via the `extra` keyword.

## Test Plan

I have added a new test case and python file to flake8_logging_format with examples of this anti-pattern.


crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
```diff
@@ -171,6 +171,24 @@ pub(crate) fn logging_call(checker: &Checker, call: &ast::ExprCall) {
         check_msg(checker, format_arg);
     }
 
+    // G005
+    let skip = match logging_call_type {
+        LoggingCallType::LevelCall(_) => 1,
+        LoggingCallType::LogCall => 2,
+    };
```
ntBre: Could we move this into the `is_rule_enabled` branch below? It shouldn't be too expensive but that narrows the scope at least too.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
```diff
@@ -333,6 +333,88 @@ impl Violation for LoggingFString {
     }
 }
 
+/// ## What it does
+/// Checks for pre-formatting of parameters to logging statements.
```
ntBre:
```suggestion
/// Checks for pre-formatting of arguments to `logging` calls.
```

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
```diff
@@ -333,6 +333,88 @@ impl Violation for LoggingFString {
...
+///
+/// ## Why is this bad?
+/// The `logging` module provides a mechanism for passing additional values to
+/// be logged using the `extra` keyword argument. This is more consistent, more
+/// efficient, and less error-prone than formatting the string directly.
+///
+/// Parameters to logging statements will be formatted as strings automatically, so it
+/// is unnecessary and less efficient to eagerly format the parameters before passing
+/// them in.
```
ntBre: tiny nit: I believe all of the instances of "parameters" here should be "arguments." I think I also prefer "`logging` calls" over to "logging statements," as I suggested above too.

```suggestion
/// Arguments to `logging` calls will be formatted as strings automatically, so it
/// is unnecessary and less efficient to eagerly format the arguments before passing
/// them in.
```

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
```diff
@@ -333,6 +333,88 @@ impl Violation for LoggingFString {
...
+///
+/// Additionally, the use of `extra` will ensure that the values are made
+/// available to all handlers, which can then be configured to log the values
+/// in a consistent manner.
```
ntBre: Hmm, it made sense to me at first to follow the G004 docs, but now I'm not sure if the `extra` discussion is very applicable. I think in this rule, the user is already using the `extra`/positional passing suggested by G004, so we may be able to omit any `extra` discussion. Does that sound right to you, or am I missing something?

crates/ruff_linter/src/codes.rs
```diff
@@ -961,6 +961,7 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Flake8LoggingFormat, "002") => (RuleGroup::Stable, rules::flake8_logging_format::violations::LoggingPercentFormat),
         (Flake8LoggingFormat, "003") => (RuleGroup::Stable, rules::flake8_logging_format::violations::LoggingStringConcat),
         (Flake8LoggingFormat, "004") => (RuleGroup::Stable, rules::flake8_logging_format::violations::LoggingFString),
+        (Flake8LoggingFormat, "005") => (RuleGroup::Stable, rules::flake8_logging_format::violations::LoggingPreFormat),
```
ntBre: New rules need to be added in `RuleGroup::Preview`.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
GDYendell: I was wondering about this and whether this should check should look at `extra`, but
given the user could be constructing a dict inline I think this would be excessive. I
would say this rule should not consider `extra` at all and we should remove any mention
of it here.

GDYendell:
Thanks for the review! I added a couple of fixup commits to address your comments.

> I'm also not totally sold on the name. What do you think about LoggingEagerConversion? Just an idea, I'm not sure I love that either.

I wasn't sure what to call it and chose the name in the hope that someone would come up with something better  :sweat_smile:

I prefer `LoggingEagerConversion`. It could also be `LoggingEagerFormat` or `LoggingEagerCast`?

> Finally, would you mind looking through the ecosystem results? The simple implementation here is very appealing, but we'll want to make sure there aren't any false positives that might warrant parsing and cross-referencing the formatting arguments, as Charlie mentioned on the issue.

Will do - this is a really cool CI helper! I have had a quick look and there are some casts of exceptions, which possibly do more the logging formatter does.

GDYendell:
I have had a look through the examples in the affected ecosystem packages and I don't see any issues. As a summary:
- Exceptions in except blocks - implements `__str__` and it is less work to just pass `exc_info=True` to the logging call with better results
- Path objects - implements `__str__`
- os.getpid() and other integers - implements `__str__`
- apache superset Slice.to_json() - not sure where the implementation is, by imagine this returns a str
- packaging Version - implements `__str__`

I didn't see any hits that aren't explicitly calling str or repr,  so I think those are all correct matches.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
ntBre: It looks like the first paragraph of `## Why is this bad?` still talks about `extra`. Should we remove that part too? I think it would be fine to lead with the second paragraph: `Arguments to logging calls...`

crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
```diff
@@ -171,6 +171,24 @@ pub(crate) fn logging_call(checker: &Checker, call: &ast::ExprCall) {
...
+    if checker.is_rule_enabled(Rule::LoggingPreFormat) {
+        let skip = match logging_call_type {
+            LoggingCallType::LevelCall(_) => 1,
+            LoggingCallType::LogCall => 2,
+        };
+        // Check if any positional args for log message values are calls to pre-format as a string
+        call.arguments.args.iter().skip(skip).for_each(|arg| {
+            if let Expr::Call(ast::ExprCall { func, .. }) = arg {
+                if let Expr::Name(ast::ExprName { id, .. }) = func.as_ref() {
+                    if id == "str" || id == "repr" {
```
ntBre: One thing I forgot to mention before, we should use `SemanticModel::match_builtin_expr` here. That will handle (rare) cases where `str` or `repr` has been overwritten:

https://github.com/astral-sh/ruff/blob/c6dcfe36d060f693d771e4ae3f6156527b67c816/crates/ruff_linter/src/rules/flake8_pyi/rules/str_or_repr_defined_in_stub.rs#L81-L83

We don't want to return in this case, but that's an example usage.

We could also add a small test for that behavior.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
GDYendell: Ah yeah it does... I am happy to remove that, but I do think all of these rules having the same opening paragraph is helpful for the reader. I find it odd that it suggests that `extra` is the primary way to pass additional values to logging statements with positional arguments as an alternative when - in my experience - the latter is much more common. It might be worth rewording this to suggest positional args so that they can continue to share the same opening paragraph and mention `extra` as the alternative for the rules that is relevant to.

What do you think?

ntBre: I can see what you mean, but I wouldn't really want to update the other rule docs in this PR. I guess I'm starting to see this rule as conceptually fairly different from G001-G004 anyway, so I don't mind if the docs are different. There's still a chance it will need to be a RUF rule too.

GDYendell:
> A case like this is currently flagged:
> ```python
> import logging
> try:
>     1/0
> except Exception as e:
>     logging.error("error: %s", e)
> ```

Is your first snippet here supposed to have `str(e)` rather than `e`, or is it actually flagging a case where it doesn't use `str` or `repr`?

> However, mixing %s and repr, does change the result:
> ```python
> try:
> ...     1/0
> ... except Exception as e:
> ...     logging.error("error: %s", repr(e))
> ...
> ERROR:root:error: ZeroDivisionError('division by zero')
> ```

Ah that is interesting, I hadn't appreciated that. This is a tricky one then because the user might prefer that output. They would want to use `logging.error("error: %r", e)` to get the same, so we need to change the suggestion depending on whether `str` or `repr` was used. If you are happy to leave that until another PR, for now maybe it would be useful to say something like "use `%s` to format with `str` or `%r` to format with `repr`".

The `%r` & `str()` case is an odd one - would that really be intentional? I am not sure what the suggestion should be because you can't get the same output with just a format string. It is relying on casting to a `str` and then formatting the `repr` of a `str`, which wraps it in quotes. This doesn't give the same output: 

```python
try:
   1/0
except Exception as e:
   logging.error("error: %r", str(e))

ERROR:root:error: 'division by zero'
```

```python
try:
   1/0
except Exception as e:
   logging.error("error: '%r'", e)

ERROR:root:error: 'ZeroDivisionError('division by zero')'
```

crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
GDYendell: Ah good one. I will update it.

ntBre:
> Is your first snippet here supposed to have `str(e)` rather than `e`, or is it actually flagging a case where it doesn't use `str` or `repr`?

Ah yes, you're right, sorry!

> Ah that is interesting, I hadn't appreciated that. This is a tricky one then because the user might prefer that output. They would want to use `logging.error("error: %r", e)` to get the same, so we need to change the suggestion depending on whether `str` or `repr` was used. If you are happy to leave that until another PR, for now maybe it would be useful to say something like "use `%s` to format with `str` or `%r` to format with `repr`".

I think I'd prefer to go ahead and give a tailored message for each case we know about, if it's not too much more work. I ran into a somewhat analogous case for another rule recently where we got reports about the confusing, generic help message. This looks like an example of similar parsing in another rule that doesn't look too bad:

https://github.com/astral-sh/ruff/blob/58efd19f11af5e24eed665e4381d5b9e0bdf130e/crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs#L388-L407

> The `%r` & `str()` case is an odd one - would that really be intentional? I am not sure what the suggestion should be because you can't get the same output with just a format string. It is relying on casting to a `str` and then formatting the `repr` of a `str`, which wraps it in quotes. This doesn't give the same output:

Oops, good catch again. Yeah, we might want to avoid a diagnostic in this case. It seems like kind of a roundabout way to wrap something in quotes, but I still think it's intentional. We could always expand the rule later, but I'd lean more towards definite true positives like `%s` paired with `str` for a first draft.

I also remembered that there's a vaguely similar rule for f-strings: [explicit-f-string-type-conversion (RUF010)](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/#explicit-f-string-type-conversion-ruf010). This replaces calls like `f"{repr(a)}"` with `f"{a!r}"`. Most notably, it applies to `str`, `repr`, and `ascii`. I think we could apply this rule to `ascii` and `%a` as well, but I think that's definitely the least commonly used of the three. Just another idea.

GDYendell:
OK I think I see how to do that. Just to confirm, we want to iterate over the placeholders in the format string with the corresponding values and only flag `%s` with `str()` and `%r` with `repr()` as issues, ignoring the two inverted cases for now?

ntBre:
> OK I think I see how to do that. Just to confirm, we want to iterate over the placeholders in the format string with the corresponding values and only flag `%s` with `str()` and `%r` with `repr()` as issues, ignoring the two inverted cases for now?

Yep, that's exactly what I had in mind!

GDYendell:
It now only flags %s with str() and %r with repr(). What do you think?

Is it possible to provide two ranges to the diagnostic so that it underlines both %s and str()?

ntBre:
> It now only flags %s with str() and %r with repr(). What do you think?

That looks great!

> Is it possible to provide two ranges to the diagnostic so that it underlines both %s and str()?

Yes, as of ~last week! The easiest way is  with `secondary_annotation` on the guard returned by `report_diagnostic`:

GDYendell:
Hmm. I am not sure this is possible because `Expr::StringLiteral` only has a `range` for the entire string. Unless we wanted the secondary diagnostic to underline the whole string, which I am not sure adds much:

```
G005 Unnecessary cast `str()` in logging values when formatting with `%s`
 --> G005.py:4:14
  |
3 | # %s + str()
4 | logging.info("Hello %s", str("World!"))
  |              ----------  ^^^^^^^^^^^^^
  |              |
  |              Logging message here
5 | logging.log(logging.INFO, "Hello %s", str("World!"))
```

Thoughts?

ntBre:
Yeah, I don't think it makes sense to underline the whole string. I guess I was assuming that `CFormatString` would have a range for each part, but it doesn't look like it does. No worries, and thanks for looking into it! A single diagnostic range seems fine.

Would you mind resolving the conflicts? Hopefully they're not too bad. I'll try to give this another review soon :)

GDYendell:
Squashed and rebased to resolve conflicts.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
```diff
@@ -339,6 +340,76 @@ impl Violation for LoggingFString {
...
+/// ```
+///
+/// ## Options
+/// - `lint.logger-objects`
+///
+/// ## References
+/// - [Python documentation: `logging`](https://docs.python.org/3/library/logging.html)
+/// - [Python documentation: Optimization](https://docs.python.org/3/howto/logging.html#optimization)
+#[derive(ViolationMetadata)]
+pub(crate) struct LoggingPreFormat {
```
ntBre: Sorry for the hassle, but let's go ahead and make this a Ruff rule. It still feels a bit different to me from the other `G` rules, and it's not implemented upstream anyway.

I also still like the name `LoggingEagerConversion`, but the alternatives you suggested (`LoggingEagerFormat` or `LoggingEagerCast`) also sound good if you strongly prefer one of those. I'd probably rank `Cast` slightly above `Format` though :)

crates/ruff_linter/src/rules/flake8_logging_format/snapshots/ruff_linter__rules__flake8_logging_format__tests__G005.py.snap
```diff
@@ -0,0 +1,83 @@
...
+G005 Unnecessary cast `repr()` in logging values when formatting with `%r`
+  --> G005.py:34:18
+   |
+33 | # %r + repr()
+34 | info("Hello %r", repr("World!"))
+   |                  ^^^^^^^^^^^^^^
+35 | log(logging.INFO, "Hello %r", repr("World!"))
+   |
+
+G005 Unnecessary cast `repr()` in logging values when formatting with `%r`
```
ntBre: I think it might sound a little better as:


```suggestion
G005 Unnecessary `repr()` cast in logging values when formatting with `%r`
```

and if we opt not to include `cast` in the lint name, it might make more sense to say `conversion` or whatever ends up in the lint name, just to be consistent.

I think we could possibly drop the `in logging values` part of the message too, but I don't feel too strongly either way. It might still be helpful in the `concise` output format, for example.

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
```diff
@@ -339,6 +340,76 @@ impl Violation for LoggingFString {
     }
 }
 
+/// ## What it does
+/// Checks for pre-formatting of arguments to `logging` calls.
```
ntBre: Similar to a comment below, it might be nice to make this consistent with the name of the lint. Something like


```suggestion
/// Checks for eager string conversion of arguments to `logging` calls.
```

crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
```diff
@@ -256,6 +260,52 @@ pub(crate) fn logging_call(checker: &Checker, call: &ast::ExprCall) {
...
+    if checker.is_rule_enabled(Rule::LoggingPreFormat) {
+        // Extract a format string from the logging statement msg argument
+        let Ok(format_string) = (match call.arguments.find_argument_value("msg", msg_pos) {
+            Some(Expr::StringLiteral(string_literal)) => {
+                CFormatString::from_str(string_literal.value.to_str())
+            }
+            _ => return,
+        }) else {
+            return;
+        };
```
ntBre: nit: apologies if I suggested this in a previous round of review, but this looks a bit confusing to me now. I think it would be nicer as a sequential check:



```suggestion
        let Some(Expr::StringLiteral(string_literal)) = call.arguments.find_argument_value("msg", msg_pos) else {
            return;
        }
        let Ok(format_string) = CFormatString::from_str(string_literal.value.to_str()) else {
            return;
        };
```

crates/ruff_linter/src/rules/flake8_logging_format/violations.rs
GDYendell: No problem!

I have had a go at this. I am not sure if it is OK to import from `flake8_logging_format` in `ruff` or if these functions should go somewhere common.

Is 103 the correct next number to use? It wasn't entirely clear because of the gaps.

ntBre: I'll take a closer look when I review again, but in principle it's fine to import helpers from `flake8_logging_format`. One day we might split up the crates more to help with compile times, but that's pretty much the only time it could be a problem, I think. And we don't have any immediate plans for that.

For the rule code, I think we can actually slot in at `RUF065`. There are two closed PRs referencing that number, so I think it should be safe to use. I meant to include that in my first message. Sorry for the trouble recoding it twice.

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -0,0 +1,137 @@
...
+        format!("Unnecessary `{call_arg}` conversion when formatting with `{format_str}`")
+    }
+}
+
+pub(crate) fn logging_eager_conversion(checker: &Checker, call: &ast::ExprCall) {
+    let Some((logging_call_type, _range)) = find_logging_call(checker, call) else {
+        return;
+    };
+
+    // RUF065
```
ntBre: nit: this is usually a doc comment on the function itself. At least that's how I search for rules: `grep '/// RULE'` 😆

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -0,0 +1,137 @@
...
+    if checker.is_rule_enabled(Rule::LoggingEagerConversion) {
```
ntBre: I think we should be able to remove this check. We check this before calling `logging_eager_conversion` in `expression.rs`.

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -0,0 +1,137 @@
...
+        };
+        let Ok(format_string) = CFormatString::from_str(string_literal.value.to_str()) else {
+            return;
+        };
+
+        // Iterate over % placeholders in format string and zip with logging statement arguments
+        let skip = match logging_call_type {
+            LoggingCallType::LevelCall(_) => 1,
+            LoggingCallType::LogCall => 2,
+        };
```
ntBre: I hadn't noticed this before, but is this just `msg_pos + 1`?

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -0,0 +1,137 @@
...
+use std::str::FromStr;
+
+use crate::rules::flake8_logging_format::rules::{LoggingCallType, find_logging_call};
+use ruff_python_literal::cformat::{CFormatPart, CFormatString, CFormatType};
+use ruff_python_literal::format::FormatConversion;
+use ruff_text_size::Ranged;
+
+use crate::Violation;
+use crate::checkers::ast::Checker;
+use crate::registry::Rule;
```
ntBre: I wasn't too worried about this, but since I found a few other nits, would you mind sorting these into `std`, "third-party," and first-party blocks:


```suggestion
use std::str::FromStr;

use ruff_macros::{ViolationMetadata, derive_message_formats};
use ruff_python_ast::{self as ast, Expr};
use ruff_python_literal::cformat::{CFormatPart, CFormatString, CFormatType};
use ruff_python_literal::format::FormatConversion;
use ruff_text_size::Ranged;

use crate::Violation;
use crate::checkers::ast::Checker;
use crate::registry::Rule;
use crate::rules::flake8_logging_format::rules::{LoggingCallType, find_logging_call};
```

Unfortunately this still isn't supported by stable rustfmt, so I try to do it manually if I notice it.

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
GDYendell: Huh, yeah you're right. I find this a lot clearer, though. Could I suggest we change the
`msg_pos` assignment to use this match statement and then use `msg_pos + 1` for the `.skip()`?

ntBre: Sure, I think that's a good idea. I always have to stop and think about the `usize::from(bool)` trick.

ntBre (approved):
Thanks again!

PR #19943, opened 2025-08-16
[ty] Detect illegal multiple inheritance with `NamedTuple`

AlexWaygood:
## Summary

This fails at runtime:

```py
from typing import NamedTuple

class Foo(NamedTuple, object): ...
```

but we don't currently detect that. This PR adds a new lint for detecting this mistake.

## Test Plan

Mdtest/snapshots


AlexWaygood:
(The new typing conformance suite diagnostic is correct!)

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -448,6 +449,39 @@ declare_lint! {
     }
 }
 
+declare_lint! {
+    /// ## What it does
+    /// Checks for invalidly defined `NamedTuple` classes.
+    ///
+    /// ## Why is this bad?
+    /// An invalidly defined `NamedTuple` class may lead to the type checker
+    /// inferring unexpected things. It may also lead to `TypeError`s at runtime.
```
sharkdp: Maybe

```suggestion
    /// An invalidly defined `NamedTuple` class may lead to the type checker
    /// drawing wrong conclusions. It may also lead to `TypeError`s at runtime.
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -448,6 +449,39 @@ declare_lint! {
...
+    /// >>> class Foo(NamedTuple, object): ...
+    /// ...
+    /// Traceback (most recent call last):
+    ///   File "<python-input-1>", line 1, in <module>
+    ///     class Foo(NamedTuple, object): ...
+    ///   File "/python3.13/typing.py", line 2998, in __new__
+    ///     raise TypeError(
+    ///         'can only inherit from a NamedTuple type and Generic')
+    /// TypeError: can only inherit from a NamedTuple type and Generic
+    /// ```
```
sharkdp:
```suggestion
    /// ```pycon
    /// >>> from typing import NamedTuple
    /// >>> class Foo(NamedTuple, object): ...
    /// TypeError: can only inherit from a NamedTuple type and Generic
    /// ```
```

crates/ty_python_semantic/src/types/diagnostic.rs
AlexWaygood:
```suggestion
    /// An invalidly defined `NamedTuple` class may lead to the type checker
    /// drawing incorrect conclusions. It may also lead to `TypeError`s at runtime.
```

PR #19944, opened 2025-08-16
[ty] Have `SemanticIndex::place_table()` and `SemanticIndex::use_def_map` return references

AlexWaygood:


crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -252,17 +250,17 @@ impl<'db> SemanticIndex<'db> {
     /// Use the Salsa cached [`place_table()`] query if you only need the
     /// place table for a single scope.
     #[track_caller]
-    pub(super) fn place_table(&self, scope_id: FileScopeId) -> Arc<PlaceTable> {
-        self.place_tables[scope_id].clone()
+    pub(super) fn place_table(&self, scope_id: FileScopeId) -> &Arc<PlaceTable> {
```
MichaReiser: Returning an `&Arc` seems a bit weird and I suspect we only do it for the `place_table` query. I suggest returning a `&PlaceTable` here and a) have a `place_table_arc` method instead that returns a cloned `Arc` OR just copy paste this one liner into the `place_table` query. The same for `use_def_map`.

PR #19945, opened 2025-08-16
[ty] Detect `NamedTuple` classes where fields without default values follow fields with default values

AlexWaygood:
(Stacked on top of https://github.com/astral-sh/ruff/pull/19943 and https://github.com/astral-sh/ruff/pull/19944; review them first.)

## Summary

This raises a `TypeError` at runtime, and the typing spec states that type checkers must detect this in order to be compliant:

```pytb
>>> from typing import NamedTuple
>>> class Foo(NamedTuple):
...     x: int = 0
...     y: str
...     
    class Foo(NamedTuple):
        x: int = 0
        y: str
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 3007, in __new__
    raise TypeError(f"Non-default namedtuple field {field_name} "
    ...<2 lines>...
                    f"{', '.join(default_names)}")
TypeError: Non-default namedtuple field y cannot follow default field x
```

Closes https://github.com/astral-sh/ty/issues/545. After this PR has landed, the only significant remaining piece of work for `NamedTuple` support is the functional syntax for defining a `NamedTuple` class. I don't think it's essential that we get that landed before the beta, and I'd also prefer to hold off until the in-progress work on `NewType` has landed, since I think some of the refactors being done for that will make supporting functional syntax for `NamedTuple`s easier.

## Test Plan

Mdtests/snapshots.

### DIFF
--- a/crates/ty/docs/rules.md
+++ b/crates/ty/docs/rules.md
@@ -36,7 +36,7 @@ def test(): -> "int":
 <small>
 Default level: [`error`](../rules.md#rule-levels "This lint has a default level of 'error'.") ·
 [Related issues](https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20call-non-callable) ·
-[View source](https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L103)
+[View source](https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L104)
 </small>
 
 **What it does**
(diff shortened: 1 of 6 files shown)


AlexWaygood:
The new diagnostic on the typing conformance suite (https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/namedtuples_define_class.py#L52-L59) is correct!

AlexWaygood:
Hmm, I just realised that there's a considerable overlap here with #19825. The only differences really are:
- Dataclasses are more complicated due to the `KW_ONLY` sentinel, the `kw_only=True` class decorator argument, and the `kw_only=True` field argument.
- I've been a bit fancier here with trying to display subdiagnostics to point to the previous field that had a default value. We could probably do similar with the dataclasses diagnostic, though it's not essential.

#19825 probably deserves to be merged first given it's been open longer! Whichever one gets merged first, I'm happy to look into combining the implementations to reduce the extent to which logic is duplicated between two locations.

carljm (approved):
Nice!

carljm:
I just looked at #19825 and I don't think it's quite ready yet, so I think you can go ahead with this; we can revisit to combine the implementations once they are both landed.

PR #19946, opened 2025-08-17
[ty] Solve unbound typevar to Unknown & UpperBound

hauntsaninja:
Was just playing around with intersections. I think this is a thing that makes sense and has also come up in real life.

It looks like it is an improvement on some test cases, mostly because ty doesn't yet bind Self. Need to look into the scary cycle error and see if some of the new diagnostics should be suppressed as duplicates

### DIFF
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -32,9 +32,9 @@ use crate::types::variance::VarianceInferable;
 use crate::types::visitor::{TypeCollector, TypeVisitor, walk_type_with_recursion_guard};
 use crate::types::{
     ApplyTypeMappingVisitor, BindingContext, BoundTypeVarInstance, CallableType, CallableTypes,
-    ClassLiteral, FindLegacyTypeVarsVisitor, IntersectionBuilder, IntersectionType, KnownClass,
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Great, this also fixes false negatives like:
```
from typing import *

class Base: ...

BaseT = TypeVar("BaseT", bound=Base)

def f(x: Callable[[], BaseT]) -> BaseT:
    return x()

class Unrelated: ...

def make_unrelated() -> Unrelated:
    return Unrelated()

f(Unrelated)  # error
f(make_unrelated)  # error
```

carljm:
We have much better cycle handling now; would be cool to try reviving this.

hauntsaninja:
Did a basic rebase / test update.
Unfortunately it looks like we stack overflow on two tests and hang on two more.
The tests that do run look to mostly have positive changes. We both infer more and issue new diagnostics (like in the test case in my comment above). Errors are a little double reported now, but that can be fixed.

carljm:
Hmm, that's a little surprising, especially the hangs. I'll take a look tomorrow. Thanks for giving it a shot.

hauntsaninja:
Seeing what looks like a lot of recursion involving `recursive_type_normalized_impl`

PR #19947, opened 2025-08-17
[ty] Add a Todo-type branch for `type[P]` where `P` is a protocol class

AlexWaygood:
## Summary

"Meta-protocols" (the type implied by `type[P]` where `P` is a protocol class) are an interesting and under-specified concept. We'll have to support them in some form eventually, because they're used in real-world code, including in typeshed -- but we don't yet, and this causes false positives. They're also far from the highest priority thing we need to fix in our protocol implementation currently, and the missing support for them is complicating matters (adding lots of new false positives) in other protocol work I'm doing. For now, therefore, this PR just adds some Todo-type branches to get rid of the false positives in user code.

Fixes https://github.com/astral-sh/ty/issues/582. It obviously fixes it in a pretty hacky way, but we have https://github.com/astral-sh/ty/issues/903 also open, and that issue tracks proper support for meta-protocols.

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -2114,7 +2114,8 @@ static_assert(not is_assignable_to(type[Bar], type[Foo]))  # error: [static-asse
 static_assert(not is_assignable_to(TypeOf[Bar], type[Foo]))  # error: [static-assert-error]
 
 class Baz:
-    y: str = "foo"
+    x: int
+    y: ClassVar[str] = "foo"
     def method(self) -> bytes:
         return b"foo"
 


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -2069,6 +2071,61 @@ def f(value: Iterator):
...
+static_assert(not is_assignable_to(type[Bar], type[Foo]))  # error: [static-assert-error]
+static_assert(not is_assignable_to(TypeOf[Bar], type[Foo]))  # error: [static-assert-error]
+
+class Baz:
+    y: str = "foo"
+    def method(self) -> bytes:
+        return b"foo"
+
+static_assert(is_assignable_to(type[Baz], type[Foo]))
+static_assert(is_assignable_to(TypeOf[Baz], type[Foo]))
```
AlexWaygood, line 2123: these two assertions don't pass on `main`, leading to false positives on user code.

AlexWaygood:
The typing conformance diff shows two false positives going away: lines 30 and 35 [here](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/protocols_class_objects.py#L30-L35).

AlexWaygood:
That looks like roughly the mypy_primer report I'd expect here! Not totally sure what's up with the new hits on `mitmproxy`, but it looks very positive overall. Many diagnostics related to `dataclasses.fields()` are going away.

crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -2069,6 +2071,61 @@ def f(value: Iterator):
...
+# TODO: these should pass
+static_assert(not is_assignable_to(type[Bar], type[Foo]))  # error: [static-assert-error]
+static_assert(not is_assignable_to(TypeOf[Bar], type[Foo]))  # error: [static-assert-error]
+
+class Baz:
+    y: str = "foo"
+    def method(self) -> bytes:
+        return b"foo"
+
+static_assert(is_assignable_to(type[Baz], type[Foo]))
```
carljm, line 2122: The theory here is that this should disregard `x: int` on the protocol `Foo` because it describes an instance attribute? Should we mention that explicitly? It's not obvious that this should be the case, given that today we consider `x: int` to describe a class-or-instance attribute, in general.

crates/ty_python_semantic/src/types/subclass_of.rs
```diff
@@ -91,7 +91,7 @@ impl<'db> SubclassOfType<'db> {
                 TypeVarVariance::Invariant => {
                     // We need to materialize this to `type[T]` but that isn't representable so
                     // we instead use a type variable with an upper bound of `type`.
-                    Type::NonInferableTypeVar(BoundTypeVarInstance::new(
+                    Type::TypeVar(BoundTypeVarInstance::new(
```
carljm: I think this deserves a TODO comment, because I don't think either the previous code nor this new code is really correct here. I think the inferrability should probably depend on some outer context?

What real-world cases triggered the need for this change?

AlexWaygood: What on earth... I didn't mean to make any change here to do with `TypeVar` materialization! I think this was a mistake introduced when resolving merge conflicts. It's a bit scary that no tests failed as a result of this -- cc. @dcreager; looks like we might be missing some coverage here?

Sorry for the error!! Thanks v much for the catch -- I'll change it back

AlexWaygood:
```suggestion
                    Type::NonInferableTypeVar(BoundTypeVarInstance::new(
```

crates/ty_python_semantic/resources/mdtest/protocols.md
AlexWaygood: > The theory here is that this should disregard `x: int` on the protocol `Foo` because it describes an instance attribute?

Right. If we have a class `Spam` that looks like this:


```py
class Spam:
    def __init__(self):
        self.x: int = 42
```

then we would consider instances of class `Spam` to inhabit the type `Foo` as well as the type `Spam`, and `x` would obviously not be available on the class object `Spam` here (nor would we consider it available on the class object `Spam` here). It follows that therefore we cannot assume or enforce that `x` is available on the meta-type of any type `T` just because `T` is a subtype of `Foo`.

I'll add some more prose explaining this.

AlexWaygood: Ah, but on second look, there is a bug in the test here. `type[Baz]` should not be assignable to `type[Foo]` because it fails the third test I gave in my prose intro to this snippet: instantiating `Baz` would not create an object that would inhabit `Foo`.

Thanks for the catch!

PR #19949, opened 2025-08-17
[ty] Remove unused code

AlexWaygood:
## Summary

Most functions in `types.rs` do not need to be `pub`. Them being `pub` prevents Clippy from warning about unused code, and also seems to suppress other useful lints such as `trivially_copy_pass_by_ref`.

## Test Plan

`cargo build`

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -3,7 +3,6 @@ use itertools::{Either, Itertools};
 use ruff_db::parsed::parsed_module;
 
 use std::borrow::Cow;
-use std::slice::Iter;
 
 use bitflags::bitflags;
 use call::{CallDunderError, CallError, CallErrorKind};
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -9147,10 +9090,6 @@ impl<'db> UnionType<'db> {
         Self::from_elements(db, self.elements(db).iter().filter(filter_fn))
     }
 
-    pub fn iter(&self, db: &'db dyn Db) -> Iter<'_, Type<'db>> {
-        self.elements(db).iter()
-    }
```
AlexWaygood, line 9152: I don't think we really need the indirection here, and removing this method allowed Clippy to point out that there was a callsite in `arguments.rs` where a `union.elements(db).to_vec()` call would be more efficient than calling `union.elements(db).iter().copied().collect()`

MichaReiser (approved):
I do except that more methods need to be made public to implement LSP and linter features but I'm fine promoting the methods once that need arises

PR #19756, opened 2025-08-05
[ty] Keep track of type qualifiers in stub declarations without right-hand side

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/937

## Test Plan

Regression test

PR #19950, opened 2025-08-17
[ty] Fix protocol interface inference for stub protocols and subprotocols

AlexWaygood:
## Summary

This PR fixes two bugs in our inference of protocol interfaces.

Firstly, we currently correctly record whether a protocol member is marked as a `ClassVar`... but only if the protocol originates from a `.py` file! If the protocol originated from a `.pyi` file, we were accidentally throwing this information away -- this is due to the fact that we iterate over both declarations and bindings to infer a protocol interface, and in a stub file `x: ClassVar[int]` is considered both a declaration _and_ a binding, but the bindings iterator doesn't return any information about the qualifiers, and the information from the bindings iterator was overwriting any qualifiers information we'd obtained from the declarations iterator.

Secondly, we currently get subprotocols very wrong. For a situation like this, we infer `SubProto` as having a method member `method` that returns `int`, not `bool`, because we're incorrectly overwriting the information regarding the subprotocol with information about the parent protocol:

```py
from typing import Protocol

class BaseProto(Protocol):
    def method(self) -> int: ...

class SubProto(BaseProto, Protocol):
    def method(self) -> bool: ...
```

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -397,7 +397,7 @@ To see the kinds and types of the protocol members, you can use the debugging ai
 
 ```py
 from ty_extensions import reveal_protocol_interface
-from typing import SupportsIndex, SupportsAbs
+from typing import SupportsIndex, SupportsAbs, ClassVar
 
 # error: [revealed-type] "Revealed protocol interface: `{"method_member": MethodMember(`(self) -> bytes`), "x": AttributeMember(`int`), "y": PropertyMember { getter: `def y(self) -> str` }, "z": PropertyMember { getter: `def z(self) -> int`, setter: `def z(self, z: int) -> None` }}`"
 reveal_protocol_interface(Foo)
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -623,13 +641,14 @@ class HasXWithDefault(Protocol):
 class FooWithZero:
     x: int = 0
 
+static_assert(is_subtype_of(FooWithZero, HasXWithDefault))
+static_assert(is_assignable_to(FooWithZero, HasXWithDefault))
+static_assert(is_subtype_of(Foo, HasXWithDefault))
+static_assert(is_assignable_to(Foo, HasXWithDefault))
```
sharkdp, line 670: The polarity of these tests has changed. They previously asserted that `Foo` was not a a subtype of `HasXWithDefault`. Indeed, `Foo` does not have an `x` attribute with a default, so if the current tests reflect the intended behavior, can we add a comment why this is correct (is the fact that the protocol member has a default value irrelevant?)?

crates/ty_python_semantic/src/types/protocol_class.rs
```diff
@@ -533,69 +547,69 @@ fn cached_protocol_interface<'db>(
...
-                            if callable.is_function_like(db) =>
-                        {
-                            ProtocolMemberKind::Method(callable)
-                        }
-                        (Type::FunctionLiteral(function), BoundOnClass::Yes) => {
-                            ProtocolMemberKind::Method(function.into_callable_type(db))
-                        }
-                        _ => ProtocolMemberKind::Other(ty),
-                    };
+        for (symbol_id, declarations) in use_def_map.all_end_of_scope_symbol_declarations() {
```
sharkdp, line 570: We do this sort of thing a lot in ty's code base: iterating over both bindings and declarations separately, hoping that there is a 1:1 relationship. This is not always the case. For example:
```py
class D: # maybe a protocol, maybe a dataclass with fields, …
    if condition:
        attr: int
    else:
        attr = ""
```

This is unlikely to be a problem for protocols in particular, I think. And it's certainly not something specific to this PR. But I think it would be very beneficial in a lot of places if we could somehow iterate over all *definitions* and then get the declared type (for `attr: int`), the inferred type (for `attr = ""`), or both (for `attr: int = 0`).

Here, it would eliminate the mutation of the entry. And the bug would probably not have happened in the first place if the iterator would yield either a `DeclaredType { ty }`, an `InferredType { ty }`, or `Both { declared_ty, inferred_ty }`. See https://github.com/astral-sh/ruff/pull/19756 for a somewhat related problem that I fixed a few weeks ago.

sharkdp (approved):
Thank you!

crates/ty_python_semantic/resources/mdtest/protocols.md
AlexWaygood: Yeah, I keep on changing my mind about this, and need to make a decision 😆 thanks for calling it out.

As the conformance-suite diff shows, the conformance suite tests include an example where a class with implicit instance attributes is considered a subtype of a protocol where the member is bound in the class body. I think you can make that sound, but only if you say that the instance attribute is _not_ present on the meta-protocol (`type[HasXWithDefault]`), which would obviously imply that we'd do attribute lookup differently for `type[]` Protocol types than we would for `type[]` nominal types.

I was actually confused about why the conformance suite diff changed, though, and you made me look again and realise that this PR was fixing _another_ bug. We were inferring this protocol as having an `x: Literal[0]` member rather than an `x: int` member:

```py
class P(Protocol):
    x: int = 0
```

I'll add a comment by this test, and add another test that explicitly asserts that we infer the correct interface for members that have default values.

crates/ty_python_semantic/src/types/protocol_class.rs
carljm: Created https://github.com/astral-sh/ty/issues/1051 from this comment

PR #19963, opened 2025-08-18
[ty] Short-circuit inlayhints request if disabled in settings

MichaReiser:
## Summary

Not only is it unnecessary to visit the entire AST if all inlay hints are disabled, it also makes it more difficult to
narrow down issues because there's no easy way to disable all inlays (other than disabling them at the editor level).

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -85,6 +85,7 @@ pub struct InlayHintSettings {
     /// foo("x="1)
     /// ```
     pub call_argument_names: bool,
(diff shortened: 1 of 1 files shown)


crates/ty_ide/src/inlay_hints.rs
```diff
@@ -87,6 +87,12 @@ pub struct InlayHintSettings {
     pub call_argument_names: bool,
```
dhruvmanila, line 87: Can we add a short comment somewhere here stating that if more options are added, it should be considered for `any_enabled`?

Issue #19966, opened 2025-08-18
Changed error formatting between 0.12.8.and 0.12.9


MeggyCal:
### Question

Hi, thank you for maintaining this useful tool! I am a packager in openSUSE, with an update to 0.12.9 we are observing a test failure in `pytest-examples` (the tests ran fine with `ruff` 0.12.8).

It looks like `ruff` stopped printing pointers to the place where the error occurred. The patch to `pytest-examples` would be trivial, but I just wanted to make sure that it was intentional (nothing in the release highlights sounded like this change to me).

The log:
```
[   14s] =================================== FAILURES ===================================
[   14s] _______________________________ test_ruff_offset _______________________________
[   14s] 
[   14s]     def test_ruff_offset():
[   14s]         code = 'print(x)\n'
[   14s]         example = CodeExample.create(code)
[   14s]         with pytest.raises(FormatError, match='testing.md:1:7: F821 Undefined name'):
[   14s] >           ruff_check(example, ExamplesConfig())
[   14s] 
[   14s] tests/test_lint.py:28: 
[   14s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[   14s] 
[   14s] example = CodeExample(source='print(x)\n', path=PosixPath('testing.md'), start_line=0, end_line=1, start_index=0, end_index=9, prefix='', indent=0, group=None, test_id=None)
[   14s] config = ExamplesConfig(line_length=88, quotes='either', magic_trailing_comma=True, target_version='py37', upgrade=False, isort=False, ruff_line_length=None, ruff_select=None, ruff_ignore=None, white_space_dot=False)
[   14s] 
[   14s]     def ruff_check(
[   14s]         example: CodeExample,
[   14s]         config: ExamplesConfig,
[   14s]         *,
[   14s]         extra_ruff_args: tuple[str, ...] = (),
[   14s]     ) -> str:
[   14s]         ruff = find_ruff_bin()
[   14s]         args = ruff, 'check', '-', *config.ruff_config(), *extra_ruff_args
[   14s]     
[   14s]         p = Popen(args, stdin=PIPE, stdout=PIPE, stderr=PIPE, universal_newlines=True)
[   14s]         stdout, stderr = p.communicate(example.source, timeout=10)
[   14s]         if p.returncode == 1 and stdout:
[   14s]     
[...]

PR #19415, opened 2025-07-17
Move full diagnostic rendering to `ruff_db`

ntBre:
## Summary

This PR switches the `full` output format in Ruff over to use the rendering code
in `ruff_db`. As proposed in the design doc, this involves a lot of
changes to the snapshot output.

I also had to comment out this assertion with a TODO to replace it after https://github.com/astral-sh/ruff/issues/19688 because many of Ruff's "file-level" annotations aren't actually file-level. They just happen to occur at the start of the file, especially in tests with very short snippets.

https://github.com/astral-sh/ruff/blob/529d81daca30e0af93a18d7406bd2c419008e8f4/crates/ruff_annotate_snippets/src/renderer/display_list.rs#L1204-L1208

I broke up the snapshot commits at the end into several blocks, but I don't think it's enough to help with review. The first few (notebooks, syntax errors, and test rules) are small enough to look at, but I couldn't really think of other categories beyond that. I'm happy to break those up or pick out specific examples beyond what I have below, if that would help.

The minimal code changes are in this [range](https://github.com/astral-sh/ruff/pull/19415/files/abd28f1e776e442379186e9f24c9aa6290f64163), with the snapshot commits following. Moving the `FullRenderer` and updating the `EmitterFlags` aren't strictly necessary either. I even dropped the renderer commit this morning but figured it made sense to keep it since we have the `full` module for tests. I don't feel strongly either way.

## Test Plan

I did actually click through all 1700 snapshots individually instead of
accepting them all at once, although I moved through them quickly. There are a
few main categories:

### Lint diagnostics

```diff
-unused.py:8:19: F401 [*] `pathlib` imported but unused
+F401 [*] `pathlib` imported but unused
+  --> unused.py:8:19
    |
  7 | # Unused, _not_ marked as required (due to the alias).
  8 | import pathlib as non_alias
-   |                   ^^^^^^^^^ F401
+   |                   ^^^^^^^^^
  9 |
 10 | # Unused, marked as required.
    |
-   = help: Remove unused import: `pathlib`
+help: Remove unused import: `pathlib`
```

- The filename and line numbers are moved to the second line
- The second noqa code next to the underline is removed

### Syntax errors

These are much like the above.

```diff
-    -:1:16: invalid-syntax: Expected one or more symbol names after import
+    invalid-syntax: Expected one or more symbol names after import
+     --> -:1:16
       |
     1 | from foo import
       |                ^
```

[...]

PR #19968, opened 2025-08-18
Add `full` output format changes to the changelog

ntBre:
Summary
--

I thought this might warrant a small blog-style writeup, especially since we already got a question about it (#19966), but I'm happy to switch back to a one-liner under `### Other changes` if preferred.

I'll copy whatever we add here to the release notes too.

Do we need a note at the top about the late addition?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Add rule code to error description in GitLab output ([#19896](https://github.com/astral-sh/ruff/pull/19896))
 
 - Improve rendering of the `full` output format ([#19415](https://github.com/astral-sh/ruff/pull/19415))
+
     Below is an example diff for [`F401`](https://docs.astral.sh/ruff/rules/unused-import/):
 
     ```diff


CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
 
 ## 0.12.9
 
+### New default diagnostic format
```
MichaReiser: I think this headline is a bit confusing because it suggests that we now use an entirely different output format, and that users have the option to switch back to the old format. However, this is not the case. 

I think I'd phrase it something like: *Improvements to rendering of `full` output format. 


I'm also a bit torn on the blog-like write-up. The prominence of the change now suggests that this should have been part of a minor release. On the other hand, I do think it's useful for users to understand the change. 


It's a bit unfortunate that https://github.com/astral-sh/ruff/pull/19900 landed right after the release. It would have allowed us to use it as an example and explain the entire change as part of an improved rendering for F811. 

What we could try to make it less sound like a breaking change (which it isn't, because we don't consider the exact output of full to be under semver), is to set the focus on what this will enable in the future (multifile diagnostics too). As far as the format changes go, the only change I would mention is that the file name is now rendered on a separate line.

ntBre: Updated the headline! (locally)

I know what you mean about the write-up. I think uv includes more text in some of their release notes, but for Ruff this is exclusive to the minor releases, so this might give the wrong idea. I could shorten the section substantially and/or move it to a sub-bullet in `### Other changes` if that would help. It would certainly be less prominent:

### Other changes

- ...
- Improve rendering of the `full` output format ([#19415](https://github.com/astral-sh/ruff/pull/19415))
	    Ruff now uses the same default diagnostic format as ty. Below is an example diff for [`F401`](https://docs.astral.sh/ruff/rules/unused-import/):
    
    ```diff
    -unused.py:8:19: F401 [*] `pathlib` imported but unused
    +F401 [*] `pathlib` imported but unused
    +  --> unused.py:8:19
        |
      7 | # Unused, _not_ marked as required (due to the alias).
      8 | import pathlib as non_alias
    -   |                   ^^^^^^^^^ F401
    +   |                   ^^^^^^^^^
      9 |
     10 | # Unused, marked as required.
        |
    -   = help: Remove unused import: `pathlib`
    +help: Remove unused import: `pathlib`
    ```
	For now, the primary change is in the header, but this new representation will allow us to make further additions to Ruff's diagnostics, such as adding sub-diagnostics and multiple annotations to the same snippet.

<hr>

Or would it be more helpful to omit the diff and only show the new F811 diagnostic? And agreed that it would have been nice to include both, I somehow didn't quite realize that this was going out in the last release and without the F811 PR.

MichaReiser: Moving it under Other makes sense to me. It makes it less prominent (while still prominent enough). I would omit the mention of ty, most users won't be familiar with it and it's unclear why it even matters for them.

Issue #4188, opened 2023-05-02
[FEATURE] Only allow relative imports


Kludex:
I've created https://github.com/Kludex/only-relative-import/ (only lints for now), but it's more comfortable to have it here. Would it be possible?

This is the opposite of `only-absolute-imports`.

robertschweizer:
This is also offered by https://github.com/MarcoGorelli/absolufy-imports using the `--never` option.

I would love to see this in ruff!

RobertCraigie:
we'd also love to have this supported in Ruff! are there any updates here?

PR #16092, opened 2025-02-11
[`flake8-builtins`] Remove `builtins-` prefix from option names

InSyncWithFoo:
## Summary

Resolves #15368.

The following options have been renamed:

* `builtins-allowed-modules` &rarr; `allowed-modules`
* `builtins-ignorelist` &rarr; `ignorelist`
* `builtins-strict-checking` &rarr; `strict-checking`

To preserve compatibility, the old names are kept as Serde aliases.

## Test Plan

`cargo nextest run` and `cargo insta test`.

PR #19973, opened 2025-08-18
[relative-imports] Add setting to require relative imports

genevieve-me:
## Summary

Adds an 'always relative' import style setting to flake8-tidy-imports, closing #4188 .
    
The PR changes the flake8-tidy-imports TID252 rule and its ban-relative-imports setting to allow forcing imports to be always absolute, relative, or absolute for parents.
    
Since the rule used to just configure the maximum level of relative import (either never allowing relative or allowing at the same level), I renamed the rule from `ban-relative-imports` to `relative-import-style`. I also made some internal name changes, e.g., relative import `Strictness` became relative import `ImportStyle` (perhaps a slightly redundant name), and diagnostics were updated.
    
The implementation relies on the categorize function used for the isort rule to determine if absolute imports are first party and therefore could be relative.
    
NOTE: I realize now that this implementation is wrong, which is why this PR is marked as draft.
First party doesn't mean 'can be relative,'  as users can configure separate packages to be first party with `isort.known-first-party` but they can't be imported relatively. I'll have to take a closer look at the ImportType options and see which one applies to same-package imports that could become `from .bar import baz` instead of `from foo.bar import baz`.

### Settings

I also presume that the way this PR naively renames the setting is a big no-no for breaking configs, especially since TID252 is not in preview. I actually couldn't find any examples of renaming settings in the changelogs: presumably there should be some sort of deprecation period.

**Getting feedback on the settings is the main reason I'm opening the PR** even in its incomplete state. The renaming seems necessary, because:

* The initial feedback on the feature request said "We could probably add support for this as a configurable setting in TID252"
* The existing setting can cleanly cover all three cases, so adding a second one and having two overlapping settings for this rule would seem odd
* However, setting `ban-relative-imports` to force relative imports would be confusing. Even something like `ban-relative-imports = 'none'`, while in keeping with the current rules, doesn't make it clear that absolute imports are actually being banned.

## Test Plan

Right now, I updated the existing tests, but I still need to add new tests where absolute imports trigger diagnostics if appropriate (they are in the same package namespace and could be relative).

I also want to look into the performance. It feels redundant to re-categorize all imports when they're already being categorized elsewhere for the isort rule, but we can't assume that the isort rule is enabled and we can get the information from there. Maybe we could categorize them all at a higher level and pass that information to the relevant places, but that could also add unnecessary complexity?

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_tidy_imports/rules/relative_imports.rs
+++ b/crates/ruff_linter/src/rules/flake8_tidy_imports/rules/relative_imports.rs
@@ -7,9 +7,11 @@ use ruff_python_codegen::Generator;
 use ruff_python_stdlib::identifiers::is_identifier;
 
 use crate::checkers::ast::Checker;
+use crate::preview::is_full_path_match_source_strategy_enabled;
 use crate::{Edit, Fix, FixAvailability, Violation};
 
-use crate::rules::flake8_tidy_imports::settings::Strictness;
+use crate::rules::flake8_tidy_imports::settings::ImportStyle;
+use crate::rules::isort;
 
 /// ## What it does
 /// Checks for relative imports.
@@ -42,12 +44,12 @@ use crate::rules::flake8_tidy_imports::settings::Strictness;
 /// ```
 ///
 /// ## Options
-/// - `lint.flake8-tidy-imports.ban-relative-imports`
+/// - `lint.flake8-tidy-imports.relative-import-style`
 ///
 /// [PEP 8]: https://peps.python.org/pep-0008/#imports
 #[derive(ViolationMetadata)]
 pub(crate) struct RelativeImports {
-    strictness: Strictness,
+    strictness: ImportStyle,
 }
 
 impl Violation for RelativeImports {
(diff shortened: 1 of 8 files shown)


ntBre:
Apologies for the delay here. I just wanted to say thanks for your work on this, and that I do intend to take a closer look at some point!

crates/ruff_linter/src/rules/flake8_tidy_imports/rules/relative_imports.rs
```diff
@@ -42,12 +44,12 @@ use crate::rules::flake8_tidy_imports::settings::Strictness;
 /// ```
 ///
 /// ## Options
-/// - `lint.flake8-tidy-imports.ban-relative-imports`
+/// - `lint.flake8-tidy-imports.relative-import-style`
 ///
 /// [PEP 8]: https://peps.python.org/pep-0008/#imports
 #[derive(ViolationMetadata)]
 pub(crate) struct RelativeImports {
-    strictness: Strictness,
```
ntBre, line 50: Hmm, I actually kind of like the "strictness" name. What do you think about using this as the setting name? `strictness = "require"` sounds nice to me, as Charlie mentioned on the issue.

ntBre (commented):
Thanks again for your work on this and apologies again for the delay. The code changes look reasonable to me, but I had a couple of thoughts about adding the option.

As you noted, we'd definitely need to make this a preview change and also deprecate the existing option name, if we change it. #16092 is an example where we've renamed an option in the past.[^1]

I was a bit worried about the isort code because it has been a bit unreliable, with a lot of [issues](https://github.com/astral-sh/ruff/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aclosed%20isort%20third-party) around the first- vs third-party categorization. But I think we can instead reuse smaller parts of the isort code, like [`match_sources`](https://github.com/astral-sh/ruff/blob/a722df6a7309c6de2eaac1ba55abe8e741ea7e17/crates/ruff_linter/src/rules/isort/categorize.rs#L169) instead of the full `categorize` call. (Assuming there aren't any weird edge cases where checking the existence of local paths isn't correct, I somewhat expect that there are).

I'd be curious to get @dylwil3 and @amyreese's thoughts on this too, but assuming we can reliably detect viable relative imports and isolate this to a preview change, I'm supportive of adding this feature.

[^1]: I think we can probably get away with just adding a third `Strictness` variant that we only respect in preview (and emit a warning if it's used outside preview) to allow landing this in a patch release. Then we could open a separate `breaking` PR changing the name, to be included in the next minor release.

amyreese:
I would personally love for *something* to enforce/convert modules using relative imports, because nothing bugs me more than using relative imports everywhere, and then VSCode's auto-import code actions creating absolute imports, even from modules that are already using relative imports, and only noticing once tests fail or a release gets flagged for breaking something. 😅

genevieve-me:
Hey Brent, thanks for taking a look and for the guidance on the preview approach. I'll be able to take a closer look over the weekend.

PR #19975, opened 2025-08-18
[ty] Ask the LSP client to watch all project search paths

BurntSushi:
This change rejiggers how we register globs for file watching with the
LSP client. Previously, we registered a few globs like `**/*.py`,
`**/pyproject.toml` and more. There were two problems with this
approach.

Firstly, it only watches files within the project root. Search paths may
be outside the project root. Such as virtualenv directory.

Secondly, there is variation on how tools interact with virtual
environments. In the case of uv, depending on its link mode, we might
not get any file change notifications after running `uv add foo` or
`uv remove foo`.

To remedy this, we instead just list for file change notifications on
all files for all search paths. This simplifies the globs we use, but
does potentially increase the number of notifications we'll get.
However, given the somewhat simplistic interface supported by the LSP
protocol, I think this is unavoidable (unless we used our own file
watcher, which has its own considerably downsides). Moreover, this is
seemingly consistent with how `ty check --watch` works.

This also required moving file watcher registration to *after*
workspaces are initialized, or else we don't know what the right search
paths are.

This change is in service of #19883, which in order for cache
invalidation to work right, the LSP client needs to send notifications
whenever a dependency is added or removed. This change should make that
possible.

I tried this patch with #19883 in addition to my work to activate Salsa
caching, and everything seems to work as I'd expect. That is,
completions no longer show stale results after a dependency is added or
removed.

### DIFF
--- a/crates/ty_server/src/server/main_loop.rs
+++ b/crates/ty_server/src/server/main_loop.rs
@@ -5,10 +5,8 @@ use crate::session::{ClientOptions, SuspendedWorkspaceDiagnosticRequest};
 use anyhow::anyhow;
 use crossbeam::select;
 use lsp_server::Message;
-use lsp_types::notification::{DidChangeWatchedFiles, Notification};
-use lsp_types::{
-    ConfigurationParams, DidChangeWatchedFilesRegistrationOptions, FileSystemWatcher, Url,
-};
+use lsp_types::notification::Notification;
+use lsp_types::{ConfigurationParams, Url};
 use serde_json::Value;
 
 pub(crate) type ConnectionSender = crossbeam::channel::Sender<Message>;
@@ -154,6 +152,10 @@ impl Server {
                     Action::InitializeWorkspaces(workspaces_with_options) => {
                         self.session
                             .initialize_workspaces(workspaces_with_options, &client);
+                        // We do this here after workspaces have been initialized
+                        // so that the file watcher globs can take project search
+                        // paths into account.
+                        // self.try_register_file_watcher(&client);
                     }
                 },
             }
@@ -195,7 +197,6 @@ impl Server {
 
     fn initialize(&mut self, client: &Client) {
         self.request_workspace_configurations(client);
-        self.try_register_file_watcher(client);
     }
 
     /// Requests workspace configurations from the client for all the workspaces in the session.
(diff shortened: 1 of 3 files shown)


crates/ty_server/src/server/main_loop.rs
```diff
@@ -154,6 +156,10 @@ impl Server {
                     Action::InitializeWorkspaces(workspaces_with_options) => {
                         self.session
                             .initialize_workspaces(workspaces_with_options, &client);
+                        // We do this here after workspaces have been initialized
+                        // so that the file watcher globs can take project search
+                        // paths into account.
+                        self.try_register_file_watcher(&client);
```
dhruvmanila: I think this needs to be moved inside `self.session.initialize_workspace` because we call that method directly (instead of taking this branch) for clients that doesn't support `workspace/configuration`. Refer to the other reference call of `self.session.initialize_workspace`.

The `try_register_file_watcher` should also be moved on the `Session` then. At that point, I think it would be useful to use the `Session::register_dynamic_capabilities` method to include the logic from `try_register_file_watcher` by checking whether the client supports it and then adding the `Unregistration` and `Registration` accordingly to the vector.

crates/ty_server/src/server/main_loop.rs
```diff
@@ -284,60 +289,104 @@ impl Server {
     }
 
     /// Try to register the file watcher provided by the client if the client supports it.
+    ///
+    /// Note that this should be called *after* workspaces/projects have been initialized.
+    /// This is required because the globs we use for registering file watching take
+    /// project search paths into account.
     fn try_register_file_watcher(&mut self, client: &Client) {
```
dhruvmanila: I think it would be useful to move this on `Session` and re-use the `register_dynamic_capabilities` method. Refer to my other comment.

crates/ty_server/src/server/main_loop.rs
```diff
@@ -284,60 +289,104 @@ impl Server {
...
         static FILE_WATCHER_REGISTRATION_ID: &str = "ty/workspace/didChangeWatchedFiles";
 
+        fn make_watcher(glob: &str) -> FileSystemWatcher {
```
dhruvmanila: nit: this seems to only be used once, should we inline the logic unless I've missed any other usage?

crates/ty_server/src/server/main_loop.rs
```diff
@@ -284,60 +289,104 @@ impl Server {
...
+        #[allow(clippy::if_not_else)] // no! it reads better this way ---AG
+        let watchers = if !self
+            .session
+            .client_capabilities()
+            .supports_relative_file_watcher()
+        {
+            tracing::warn!(
+                "Client does not support watching relative globs, \
+                 falling back to watching only things in the project root",
+            );
```
dhruvmanila: What's the reason that this message is at the warning level? I think it should be fine at the info level because I don't think the user is in control of this and they cannot act on this warning.

crates/ty_server/src/server/main_loop.rs
```diff
@@ -284,60 +289,104 @@ impl Server {
...
         };
 
+        // Unregister all extant file watchers.
+        // If we don't do this step, then we will
+        // wind up registering multiple watchers
+        // with the same glob patterns, and at least
+        // with the VS Code client, this will result
+        // in duplicate notifications for the same
+        // change event.
+        client.send_request::<lsp_types::request::UnregisterCapability>(
```
dhruvmanila: Yeah, I forgot that we didn't unregister the file watcher before registering it. I think we can re-use the `Session::register_dynamic_capabilities`. The spec doesn't mention whether the server can unregister a capability which wasn't registered before but I've avoided that in the above mentioned method. I think we should follow that.

dhruvmanila (approved):
This looks great!

The only thing that would be required in this PR is to move the `try_register_file_watcher` to be called inside `initialize_workspaces` and optionally re-use the `register_dynamic_capabilities` for this.

crates/ty_server/src/server/main_loop.rs
```diff
@@ -284,60 +289,104 @@ impl Server {
...
-                            kind: None,
-                        },
-                        FileSystemWatcher {
-                            glob_pattern: lsp_types::GlobPattern::String("**/*.ipynb".into()),
-                            kind: None,
-                        },
-                    ],
+        // We also want to watch everything in the search paths as well.
+        // But this seems to require "relative" watcher support. I had
+        // trouble getting this working without using a base uri.
```
MichaReiser: Can you expand on what issues you ran into and what you used as base uri so that a future reader knows what to test if they want to fix file watching for clients not supporting relative paths (I'm not suggesting we should fix this now, I just wouldn't know what I'd have to test)

MichaReiser (approved):
Nice, thank you for fixing this

crates/ty_server/src/server/main_loop.rs
BurntSushi: Got it. That makes sense!

crates/ty_server/src/server/main_loop.rs
BurntSushi: Just copying what was already there. Specifically:

https://github.com/astral-sh/ruff/blob/e5c091b850a95601c86d7db718953a91de1e1b48/crates/ty_server/src/server/main_loop.rs#L290-L293

Should I change that to an info-level message as well?

crates/ty_server/src/server/main_loop.rs
BurntSushi: No, you're right, it's only used once. But I'd prefer to keep it. `make_relative_watcher` is only used once too. IMO it makes the code a touch easier to read as it pulls out some mildly tedious construction of LSP glob patterns and lets the reader focus on the surrounding logic. The "legacy" branch becomes straight-forwardly simple and the relative glob branch lets you focus on how the search paths are collected and deduplicated.

crates/ty_server/src/server/main_loop.rs
MichaReiser: My reasoning for using warn for `"Client does not support file system watching"` was that it can lead to a degraded experience. In hindsight, I think it would have been great to phrase the message like: `Your LSP client doesn't support file watching: You may see stale results when files change outside the editor`. 

I'd suggest that we keep the message at warning severity if it impacts the user experience and include a summary of what behavior they have to expect

BurntSushi: Sounds good. Done.

crates/ty_server/src/server/main_loop.rs
BurntSushi: I expanded on this a bit:

```rust
        // We also want to watch everything in the search paths as
        // well. But this seems to require "relative" watcher support.
        // I had trouble getting this working without using a base uri.
        //
        // Specifically, I tried this for each search path:
        //
        //     make_watcher(&format!("{path}/**"))
        //
        // But while this seemed to work for the project root, it
        // simply wouldn't result in any file notifications for changes
        // to files outside of the project root.
```

crates/ty_server/src/server/main_loop.rs
BurntSushi: This is done. Much cleaner!

PR #19976, opened 2025-08-18
[ty] improve goto/hover for definitions

Gankra:
By computing the actual Definition for, well, definitions, we unlock a bunch of richer machinery in the goto/hover subsystems for free. 

Fixes https://github.com/astral-sh/ty/issues/1001
Fixes https://github.com/astral-sh/ty/issues/1004

### DIFF
--- a/crates/ty_ide/src/goto_references.rs
+++ b/crates/ty_ide/src/goto_references.rs
@@ -405,9 +405,7 @@ except ValueError as err:
 ",
         );
 
-        // Note: Currently only finds the declaration, not the usages
-        // This is because semantic analysis for except handler variables isn't fully implemented
-        assert_snapshot!(test.references(), @r###"
+        assert_snapshot!(test.references(), @r"
         info[references]: Reference 1
          --> main.py:4:29
           |
@@ -418,7 +416,37 @@ except ValueError as err:
         5 |     print(f'Error: {err}')
         6 |     return err
           |
-        "###);
+
+        info[references]: Reference 2
+         --> main.py:5:21
+          |
+        3 |     x = 1 / 0
+        4 | except ZeroDivisionError as err:
+        5 |     print(f'Error: {err}')
+          |                     ^^^
+        6 |     return err
+          |
+
+        info[references]: Reference 3
+         --> main.py:6:12
+          |
+        4 | except ZeroDivisionError as err:
+        5 |     print(f'Error: {err}')
+        6 |     return err
+          |            ^^^
+        7 |
+        8 | try:
+          |
+
+        info[references]: Reference 4
+          --> main.py:11:31
+           |
+         9 |     y = 2 / 0
+        10 | except ValueError as err:
+        11 |     print(f'Different error: {err}')
+           |                               ^^^
+           |
+        ");
     }
 
     #[test]
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -11,6 +11,7 @@ use ruff_db::parsed::ParsedModuleRef;
 use ruff_python_ast::{self as ast, AnyNodeRef};
 use ruff_python_parser::TokenKind;
 use ruff_text_size::{Ranged, TextRange, TextSize};
+use ty_python_semantic::HasDefinition;
 use ty_python_semantic::ImportAliasResolution;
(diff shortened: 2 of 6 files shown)


Gankra:
mea culpa: this does not fix https://github.com/astral-sh/ty/issues/1004 *for vendored typeshed definitions* (possibly stdlib-specific?). It seems like file_to_module doesn't understand opened-in-your-ide vendored typeshed files? Something to look into more in a followup.

crates/ty_ide/src/goto_references.rs
```diff
@@ -405,9 +405,7 @@ except ValueError as err:
 ",
         );
 
-        // Note: Currently only finds the declaration, not the usages
-        // This is because semantic analysis for except handler variables isn't fully implemented
-        assert_snapshot!(test.references(), @r###"
+        assert_snapshot!(test.references(), @r"
         info[references]: Reference 1
```
Gankra, line 409: Love when a fix suddenly enriches an existing test case :D

crates/ty_ide/src/goto_references.rs
```diff
@@ -418,7 +416,37 @@ except ValueError as err:
...
+        7 |
+        8 | try:
+          |
+
+        info[references]: Reference 4
+          --> main.py:11:31
+           |
+         9 |     y = 2 / 0
+        10 | except ValueError as err:
+        11 |     print(f'Different error: {err}')
```
AlexWaygood, line 446: Hmm, is this one correct? It looks like a reference to a different variable with the same name to me

AlexWaygood: I doubt your PR introduced a bug here, just wondering if we should have a TODO comment or open an issue!

carljm: Same variable, different definitions. Except blocks don't create subscopes, they just overwrite whatever was previously in that variable. So I think this is the expected behavior.

MichaReiser: This also matches pycharm's behavior

AlexWaygood: > Except blocks don't create subscopes, they just overwrite whatever was previously in that variable.

You know that I know that ;) I'm not sure it's relevant, though.

Determining the set of references to a given symbol is important for code actions such as being able to right-click on a symbol in an IDE and rename it. If I right clicked on `err` here in the `except ValueError as err` branch, I'm not sure I'd expect the `err` symbol in the second `except` branch to be renamed as well? It has a wholly independent definition from the `err` symbol in the first branch

```py
try:
    foo
except ValueError as err:
    print(err)
except TypeError as err:
    print(err)
```

AlexWaygood: Obviously not a critical debate to have right now if our current behaviour matches other mainstream language servers, though

MichaReiser: That does make sense to me but I think it might get tricky if there are any `err` usages after the `except` block. It then is important that they all have the same name or renaming only some of them changes program semantics. 

This sort of lookahead is probably just complicated enough that renaming all instances is a reasonable default (it reduces the risk of breaking anything or, at least, can avoid any heavy machinary to figure out if it would)

AlexWaygood: > That does make sense to me but I think it might get tricky if there are any `err` usages after the `except` block.

For `except` definitions specifically, that can't happen because although the `except` branch doesn't introduce a new scope, the interpreter inserts an implicit `del` of the variable bound by the `except` statement at the end of the block:

```pycon
>>> try:
...     1 / 0
... except ZeroDivisionError as e:
...     print(e)
...     
division by zero
>>> print(e)
    print(e)
          ^
NameError: name 'e' is not defined
```

but yes, I see your point!

carljm: I don't think we should treat `except` defined variables differently due to their implicit `del` behavior; that _is_ going down the route of sort of treating them as a separate sub-scope. And in general, I think Micha's comment accurately describes why we shouldn't / can't try to treat separate definitions of the same variable in the same scope as separate variables for purposes of goto-references, rename, etc.

AlexWaygood: makes sense -- thanks!

PR #19977, opened 2025-08-18
Replace `pyflakes` hand-crafted tests with inline snapshots

dylwil3:
This PR makes an incremental improvement on the maintainability of our `pyflakes` test fixtures by replacing hundreds of hand-written tests of the form

```rust
assert_eq!(actual,expected)
```

with inline snapshot tests.

There are two commits - the first keeps the same format, pretty much, as the original tests. It's compact but a little hard to review the snapshots out of context. The second has snippets of our diagnostic output. Let me know if you'd like me to experiment with yet another view.

### DIFF
--- a/crates/ruff_linter/src/rules/pyflakes/mod.rs
+++ b/crates/ruff_linter/src/rules/pyflakes/mod.rs
@@ -737,7 +737,7 @@ mod tests {
 
     /// A re-implementation of the Pyflakes test runner.
     /// Note that all tests marked with `#[ignore]` should be considered TODOs.
-    fn flakes(contents: &str, expected: &[Rule]) {
+    fn flakes(contents: &str) -> Vec<&str> {
         let contents = dedent(contents);
         let source_type = PySourceType::default();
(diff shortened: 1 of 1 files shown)


MichaReiser:
I haven't looked at the details but I'd prefer if we can avoid a 11k file. IDEs struggle with files that large (rustfmt as well), I do, and LLMs do as well :)

dylwil3:
> I haven't looked at the details but I'd prefer if we can avoid a 11k file. IDEs struggle with files that large (rustfmt as well), I do, and LLMs do as well :)

Ok I reverted to the more concise output. I guess the alternative would be to move these tests to separate files and use ordinary, non-inline snapshots. It's already quite a large file, so maybe that's a good idea?

ntBre:
I'd probably go for a bunch of new snapshot files over the inline snapshots. But otherwise I liked the version with `test_snippet` and `assert_diagnostics`!

MichaReiser:
@dylwil3 is this something you still plan on doing or should we close the PR?

PR #19978, opened 2025-08-18
[ty] Look for `site-packages` directories in `<sys.prefix>/lib64/` as well as `<sys.prefix>/lib/` on non-Windows systems

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1043; fixes https://github.com/astral-sh/ty/issues/257.

According to [the documentation of `sys.platlibdir`](https://docs.python.org/3/library/sys.html#sys.platlibdir) in the standard library, Python modules should never be installed into the `lib64/pythonX.Y/site-packages` directory; that directory should only be used for C extensions (which can never be submodules of a Python package, only top-level modules). Empirically, however, some installers appear to put Python files into the `lib64/pythonX.Y/site-packages` directory rather than the `lib/pythonX.Y/site-packages` directory, in some situations. I'm not sure why they're doing this -- but we'll need to start looking at the `lib64` directory at some point in the future anyway to solve https://github.com/astral-sh/ty/issues/487, so we may as well start looking at it now.

The practical implication of this is that a virtual environment can often have two `site-packages` directories on Unix, and so can a system environment. This isn't too big of a change to our overall model, since there were already situations where we accumulated `site-packages` directories from several different environments and added them all as search paths, but it requires a little bit of a refactor to `site_packages.rs`.

## Test Plan

I added a CLI integration test.

I looked at whether it would be possible to add a unit test for this as an mdtest or in `site_packages.rs`... but concluded that both would be too much pain for too little gain here. The mdtest framework abstracts away the path to the `site-packages` directory by using "magic" `<path-to-site-packages>` path segments; this allows us to write platform-independent tests for `site-packages` import resolution, but means that we cannot add a unit test for something like this.

It would be possible to expand the testing framework in `site_packages.rs` to allow for us to add a unit test there, but it would be a big diff; I don't think it's worth it for a single test.

### DIFF
--- a/crates/ty_python_semantic/src/site_packages.rs
+++ b/crates/ty_python_semantic/src/site_packages.rs
@@ -117,12 +117,6 @@ impl<const N: usize> From<[SystemPathBuf; N]> for SitePackagesPaths {
     }
 }
 
-impl FromIterator<SystemPathBuf> for SitePackagesPaths {
-    fn from_iter<T: IntoIterator<Item = SystemPathBuf>>(iter: T) -> Self {
-        Self(IndexSet::from_iter(iter))
-    }
-}
-
 impl IntoIterator for SitePackagesPaths {
     type Item = SystemPathBuf;
     type IntoIter = indexmap::set::IntoIter<SystemPathBuf>;
@@ -289,7 +283,7 @@ impl PartialEq<&str> for UnixLibDir {
 
 impl PartialEq<UnixLibDir> for &str {
     fn eq(&self, other: &UnixLibDir) -> bool {
-        *self == other.as_str()
+        other == self
     }
 }
 


AlexWaygood:
I'm guessing this also fixes https://github.com/astral-sh/ty/issues/257, but haven't verified that

EDIT: @geofft kindly verified that poetry is indeed putting those packages into the `lib64/pythonX.Y/site-packages` directory, so it almost certainly seems to be the same issue

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -110,6 +111,12 @@ impl SitePackagesPaths {
     }
 }
 
+impl<const N: usize> From<[SystemPathBuf; N]> for SitePackagesPaths {
+    fn from(paths: [SystemPathBuf; N]) -> Self {
+        Self(IndexSet::from(paths))
+    }
+}
```
MichaReiser: Nit: Coudl we use the `from_iter` implementation below?

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -240,6 +247,52 @@ impl PythonEnvironment {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+impl AsRef<SystemPath> for UnixLibDir {
+    fn as_ref(&self) -> &SystemPath {
+        SystemPath::new(self.as_str())
+    }
+}
```
MichaReiser, line 276: I think I'd prefer using `as_path()` over `as_ref` as it is converting to a more specific type and not a less specific (e.g. String -> str). 

Especially considering that this type also implements `as_str`

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -240,6 +247,52 @@ impl PythonEnvironment {
...
+
+impl PartialEq<&str> for UnixLibDir {
+    fn eq(&self, other: &&str) -> bool {
+        self.as_str() == *other
+    }
+}
```
MichaReiser, line 282: Does this work
```suggestion
impl PartialEq<str> for UnixLibDir {
    fn eq(&self, other: &str) -> bool {
        self.as_str() == other
    }
}
```

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -240,6 +247,52 @@ impl PythonEnvironment {
...
+
+impl PartialEq<UnixLibDir> for &str {
+    fn eq(&self, other: &UnixLibDir) -> bool {
+        *self == other.as_str()
```
MichaReiser: Nit: You could do
```suggestion
        other == self 
```

MichaReiser (approved):
Nice, thank you

crates/ty_python_semantic/src/site_packages.rs
AlexWaygood: if I do that then it makes the places where I'm using the `impl` less ergonomic; I have to do awkward dereferences

crates/ty_python_semantic/src/site_packages.rs
AlexWaygood: hmm, but we already implement `AsRef<SystemPath>` for `str` -- isn't that also converting to a more specific type?

I'd prefer to leave things as they are because it means I can pass `UnixLibDir`s directly into `SystemPath::join()`, which requires the argument passed in to implement `AsRef<SystemPath>`. There are ways round this, but they all make the code less ergonomic

crates/ty_python_semantic/src/site_packages.rs
AlexWaygood: The `FromIterator` implementation was unused, so I got rid of that one and kept the new one that's used

PR #19984, opened 2025-08-19
[ty] Enable virtual terminal on Windows

sharkdp:
## Summary

I believe this fixes https://github.com/astral-sh/ty/issues/1045

## Test Plan

I have not tested this.


crates/ty/src/lib.rs
```diff
@@ -62,6 +62,9 @@ pub(crate) fn version() -> Result<()> {
 }
 
 fn run_check(args: CheckCommand) -> anyhow::Result<ExitStatus> {
+    #[cfg(windows)]
+    colored::control::set_virtual_terminal(true).ok();
```
sharkdp: `set_virtual_terminal` [will always return `Ok(())`](https://github.com/colored-rs/colored/blob/68761c1dfe306c870aa94af085c4686bce8d5fbd/src/control.rs#L14-L15).

MichaReiser: Maybe add a comment similar to ruff:

```
    // Enabled ANSI colors on Windows 10.
    #[cfg(windows)]
    assert!(colored::control::set_virtual_terminal(true).is_ok());
```

MichaReiser:
Thank you.

crates/ty/src/lib.rs
sharkdp: Oh, oops. I even searched for `set_virtual_terminal` and was surprised we didn't use it anywhere. Forgot to disable my `crates/ty*` filter ...

Made it consistent with what we have in ruff.

PR #19985, opened 2025-08-19
[ty] Add precise inference for unpacking a TypeVar if the TypeVar has an upper bound with a precise tuple spec

AlexWaygood:
## Summary

If the upper bound of a TypeVar can be unpacked, then so can the TypeVar, because no solutions of the TypeVar exist that would not be assignable to the TypeVar's upper bound.

This is similar in spirit to https://github.com/astral-sh/ruff/pull/19981!

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
@@ -455,3 +455,43 @@ def overloaded_outer(t: T | None = None) -> None:
     if t is not None:
         inner(t)
 ```
+
+## Unpacking a TypeVar
+
+We can infer precise heterogeneous types from the result of an unpacking operation applied to a
+TypeVar if the TypeVar's upper bound is a type with a precise tuple spec:
+
+```py
+from dataclasses import dataclass
+from typing import NamedTuple, Final, TypeVar, Generic
+
+T = TypeVar("T", bound=tuple[int, str])
+
+def f(x: T) -> None:
+    a, b = x
+    reveal_type(a)  # revealed: int
+    reveal_type(b)  # revealed: str
+
+@dataclass
+class Team(Generic[T]):
+    employees: list[T]
+
+def x(team: Team[T]):
+    age, name = team.employees[0]
+    reveal_type(age)  # revealed: int
+    reveal_type(name)  # revealed: str
+
+class Age(int): ...
+class Name(str): ...
+
+class Employee(NamedTuple):
+    age: Age
+    name: Name
+
+EMPLOYEES: Final = (Employee(name=Name("alice"), age=Age(42)),)
+team = Team(employees=list(EMPLOYEES))
+reveal_type(team.employees)  # revealed: list[Employee]
+age, name = team.employees[0]
+reveal_type(age)  # revealed: Age
+reveal_type(name)  # revealed: Name
+```
--- a/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
@@ -454,3 +454,41 @@ def overloaded_outer[T](t: T | None = None) -> None:
     if t is not None:
         inner(t)
 ```
+
+## Unpacking a TypeVar
+
+We can infer precise heterogeneous types from the result of an unpacking operation applied to a
+TypeVar if the TypeVar's upper bound is a type with a precise tuple spec:
+
+```py
+from dataclasses import dataclass
+from typing import NamedTuple, Final
+
+def f[T: tuple[int, str]](x: T) -> None:
+    a, b = x
+    reveal_type(a)  # revealed: int
+    reveal_type(b)  # revealed: str
+
+@dataclass
+class Team[T: tuple[int, str]]:
+    employees: list[T]
+
+def x[T: tuple[int, str]](team: Team[T]):
+    age, name = team.employees[0]
+    reveal_type(age)  # revealed: int
+    reveal_type(name)  # revealed: str
+
+class Age(int): ...
+class Name(str): ...
+
+class Employee(NamedTuple):
+    age: Age
+    name: Name
+
+EMPLOYEES: Final = (Employee(name=Name("alice"), age=Age(42)),)
+team = Team(employees=list(EMPLOYEES))
+reveal_type(team.employees)  # revealed: list[Employee]
+age, name = team.employees[0]
+reveal_type(age)  # revealed: Age
+reveal_type(name)  # revealed: Name
+```
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -4869,6 +4869,14 @@ impl<'db> Type<'db> {
             Type::TypeAlias(alias) => {
                 return alias.value_type(db).try_iterate_with_mode(db, mode);
             }
+            Type::NonInferableTypeVar(tvar) => match tvar.typevar(db).bound_or_constraints(db) {
+                Some(TypeVarBoundOrConstraints::UpperBound(bound)) => {
+                    return bound.try_iterate_with_mode(db, mode);
+                }
+                // TODO: could we create a "union of tuple specs"...?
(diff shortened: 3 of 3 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -4869,6 +4869,14 @@ impl<'db> Type<'db> {
...
                 return alias.value_type(db).try_iterate_with_mode(db, mode);
             }
+            Type::NonInferableTypeVar(tvar) => match tvar.typevar(db).bound_or_constraints(db) {
+                Some(TypeVarBoundOrConstraints::UpperBound(bound)) => {
+                    return bound.try_iterate_with_mode(db, mode);
+                }
+                // TODO: could we create a "union of tuple specs"...?
+                // (Same question applies to the `Type::Union()` branch lower down)
+                Some(TypeVarBoundOrConstraints::Constraints(_)) | None => {}
+            },
```
AlexWaygood, line 4879: Possibly I should also be adding a branch for `Type::TypeVar` here too...? I couldn't immediately find any tests that would fail without that branch, though

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
```diff
@@ -454,3 +454,41 @@ def overloaded_outer[T](t: T | None = None) -> None:
...
+class Employee(NamedTuple):
+    age: Age
+    name: Name
+
+EMPLOYEES: Final = (Employee(name=Name("alice"), age=Age(42)),)
+team = Team(employees=list(EMPLOYEES))
+reveal_type(team.employees)  # revealed: list[Employee]
+age, name = team.employees[0]
+reveal_type(age)  # revealed: Age
+reveal_type(name)  # revealed: Name
```
AlexWaygood: I enjoyed writing this test. So cool to see the number of features that are coming together here 😃

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -455,3 +455,43 @@ def overloaded_outer(t: T | None = None) -> None:
...
+```py
+from dataclasses import dataclass
+from typing import NamedTuple, Final, TypeVar, Generic
+
+T = TypeVar("T", bound=tuple[int, str])
+
+def f(x: T) -> None:
+    a, b = x
+    reveal_type(a)  # revealed: int
+    reveal_type(b)  # revealed: str
```
carljm: As written, this is a useless use of a typevar (since it appears only once in the signature), which we might at some point want to emit a diagnostic on? This annotation should just be replaced with a direct `tuple[int, str]` instead.

I think we should either have a TODO comment about that, or (maybe less distracting) change this signature so the function e.g. also returns a `T`.

```suggestion
def f(x: T) -> T:
    a, b = x
    reveal_type(a)  # revealed: int
    reveal_type(b)  # revealed: str
    return x
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -455,3 +455,43 @@ def overloaded_outer(t: T | None = None) -> None:
     if t is not None:
         inner(t)
 ```
+
+## Unpacking a TypeVar
+
+We can infer precise heterogeneous types from the result of an unpacking operation applied to a
+TypeVar if the TypeVar's upper bound is a type with a precise tuple spec:
```
dcreager: very nit: I've only been using the camel-case spelling to refer to `TypeVar` the Python entity. In prose like this I've been using "type variable" or "typevar" (lowercase)

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -455,3 +455,43 @@ def overloaded_outer(t: T | None = None) -> None:
...
+def f(x: T) -> None:
+    a, b = x
+    reveal_type(a)  # revealed: int
+    reveal_type(b)  # revealed: str
+
+@dataclass
+class Team(Generic[T]):
+    employees: list[T]
+
+def x(team: Team[T]):
```
carljm: Same as above, this function probably should be written `def x(team: Team[tuple[int, str]]):` instead, unless there's a second use of `T` somewhere in the signature.

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
```diff
@@ -454,3 +454,41 @@ def overloaded_outer[T](t: T | None = None) -> None:
...
+    reveal_type(b)  # revealed: str
+
+@dataclass
+class Team[T: tuple[int, str]]:
+    employees: list[T]
+
+def x[T: tuple[int, str]](team: Team[T]):
+    age, name = team.employees[0]
+    reveal_type(age)  # revealed: int
+    reveal_type(name)  # revealed: str
```
carljm: Same comment as above on these two functions

crates/ty_python_semantic/src/types.rs
dcreager: I don't think you should — if anything I think you want a `debug_assert` that it's not possible to iterate a typevar in an inferable position, like we have here for instance member access:

https://github.com/astral-sh/ruff/blob/c82e255ca8570a683fcd0cde1e62bd1086837c12/crates/ty_python_semantic/src/types.rs#L2791-L2799

dcreager (approved):
Love it

crates/ty_python_semantic/src/types.rs
carljm: In general, I think if we encounter a `Type::TypeVar` type in any position where we'd end up trying to iterate or unpack it, it means we have a bug elsewhere. So I think if anything, here and probably various other places, if we wanted to add a branch for `Type::TypeVar` it should just contain an `unreachable!(...)` (or maybe to avoid end-user crashes, a `debug_assert!(false, ...)`) to help us catch such bugs early. CC @dcreager to correct me if I'm wrong.

carljm (approved):
Nice!

crates/ty_python_semantic/src/types.rs
AlexWaygood: Great -- thanks both!

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
AlexWaygood:
```suggestion
## Unpacking a TypeVar

We can infer precise heterogeneous types from the result of an unpacking operation applied to a
type variable if the type variable's upper bound is a type with a precise tuple spec:
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -455,3 +455,43 @@ def overloaded_outer(t: T | None = None) -> None:
...
+    age, name = team.employees[0]
+    reveal_type(age)  # revealed: int
+    reveal_type(name)  # revealed: str
```
AlexWaygood:
```suggestion
def x(team: Team[T]) -> Team[T]:
    age, name = team.employees[0]
    reveal_type(age)  # revealed: int
    reveal_type(name)  # revealed: str
    return team
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
AlexWaygood:
```suggestion
def f[T: tuple[int, str]](x: T) -> T:
    a, b = x
    reveal_type(a)  # revealed: int
    reveal_type(b)  # revealed: str
    return x

@dataclass
class Team[T: tuple[int, str]]:
    employees: list[T]

def x[T: tuple[int, str]](team: Team[T]) -> Team[T]:
    age, name = team.employees[0]
    reveal_type(age)  # revealed: int
    reveal_type(name)  # revealed: str
    return team
```

crates/ty_python_semantic/src/types.rs
carljm: Oops, hadn't seen Doug's comment yet when I wrote and submitted mine! Glad we agree, anyway 😆

PR #19986, opened 2025-08-19
[ty] apply `KW_ONLY` sentinel only to local fields

PrettyWood:
## Summary

This PR fixes how `KW_ONLY` is applied in dataclasses. Previously, the sentinel leaked into subclasses and incorrectly marked their fields as keyword-only; now it only affects fields declared in the same class.

```py
from dataclasses import dataclass, KW_ONLY

@dataclass
class D:
    x: int
    _: KW_ONLY
    y: str

@dataclass
class E(D):
    z: bytes

# This should work: x=1 (positional), z=b"foo" (positional), y="foo" (keyword-only)
E(1, b"foo", y="foo")

reveal_type(E.__init__)  # revealed: (self: E, x: int, z: bytes, *, y: str) -> None
```

## Test Plan

mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
@@ -988,6 +988,28 @@ class D:  # error: [duplicate-kw-only]
         z: float
 ```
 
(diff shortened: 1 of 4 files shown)


carljm (approved):
This is excellent, thank you!

PR #19990, opened 2025-08-19
[ty] distinguish base conda from child conda

Gankra:
This is a port of the logic in https://github.com/astral-sh/uv/pull/7691

The basic idea is we use CONDA_DEFAULT_ENV as a signal for whether CONDA_PREFIX is just the ambient system conda install, or the user has explicitly activated a custom one. If the former, then the conda is treated like a system install (having lowest priority). If the latter, the conda is treated like an activated venv (having priority over everything but an Actual activated venv).

Fixes https://github.com/astral-sh/ty/issues/611

### DIFF
--- a/crates/ty/tests/cli/python_environment.rs
+++ b/crates/ty/tests/cli/python_environment.rs
@@ -857,59 +857,165 @@ fn defaults_to_a_new_python_version() -> anyhow::Result<()> {
 /// The `site-packages` directory is used by ty for external import.
 /// Ty does the following checks to discover the `site-packages` directory in the order:
 /// 1) If `VIRTUAL_ENV` environment variable is set
-/// 2) If `CONDA_PREFIX` environment variable is set
+/// 2) If `CONDA_PREFIX` environment variable is set (and .filename != `CONDA_DEFAULT_ENV`)
 /// 3) If a `.venv` directory exists at the project root
+/// 4) If `CONDA_PREFIX` environment variable is set (and .filename == `CONDA_DEFAULT_ENV`)
 ///
-/// This test is aiming at validating the logic around `CONDA_PREFIX`.
+/// This test (and the next one) is aiming at validating the logic around these cases.
 ///
-/// A conda-like environment file structure is used
-/// We test by first not setting the `CONDA_PREFIX` and expect a fail.
-/// Then we test by setting `CONDA_PREFIX` to `conda-env` and expect a pass.
+/// To do this we create a program that has these 4 imports:
+///
+/// ```python
+/// from package1 import ActiveVenv
+/// from package1 import ChildConda
+/// from package1 import WorkingVenv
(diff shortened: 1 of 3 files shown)


crates/ty/tests/cli/python_environment.rs
```diff
@@ -857,73 +857,540 @@ fn defaults_to_a_new_python_version() -> anyhow::Result<()> {
...
+///
+/// We then create 4 different copies of package1. Each copy defines all of these
+/// classes... except the one that describes it. Therefore we know we got e.g.
+/// the working venv if we get a diagnostic like this:
+///
+/// ```text
+/// Unresolved import
+/// 4 | from package1 import WorkingVenv
+///   |                      ^^^^^^^^^^^
+/// ```
```
Gankra, line 883: ngl this is so beautifully evil i might cry if y'all reject this

Gankra:
cc @zanieb

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -539,6 +546,62 @@ System stdlib will not be used for module definitions.",
...
+    fn from_prefix_path(system: &dyn System, path: &SystemPath) -> Self {
+        // If we cannot read `CONDA_DEFAULT_ENV`, there's no way to know if the base environment
+        let Ok(default_env) = system.env_var(EnvVars::CONDA_DEFAULT_ENV) else {
+            return CondaEnvironmentKind::Child;
+        };
+
+        // These are the expected names for the base environment
+        if default_env != "base" && default_env != "root" {
+            return CondaEnvironmentKind::Child;
+        }
```
zanieb, line 573: N.B. this isn't discussed in your commentary but we only allow these names for base environments. I don't know if people use other default environment names, but there's never been a complaint in uv.

crates/ty/tests/cli/python_environment.rs
MichaReiser: This is fun (and clever). It took me a moment to realise how this all works. An alternative (I don't have a strong preference) which results in fewer diagnostics is to use something like this:

Each virtual environment defines a `package1` that defines a single symbol, environment:

```py
from typing import Final

environment: Final = "active_venv"
```

The main file then becomes

```py
from ty_extensions import static_assert
from package1 import environment

static_assert(environment == "active_venv", "Expected `package1` to resolve to the active environment")
```

MichaReiser (approved):
Thank you. Sorry for commenting on your evil but beautiful test setup :)

crates/ty_python_semantic/src/site_packages.rs
MichaReiser: This seems straightforward to change if necessary

Gankra:
Appreciate the suggestion but the evil lives on :)

(it avoids us having to change the body of main between subtests)

PR #19991, opened 2025-08-19
[ty] Fix server hang

MichaReiser:
## Summary

This PR updates salsa to pull in https://github.com/salsa-rs/salsa/pull/981

This fixes astral-sh/ty#1024


MichaReiser:
Hmm, not what I expected to see but also not entirely surprising, given that we now need to collect more information.

MichaReiser:
Okay, so pulling in the tracked struct fix results in a memory decrease, which is great. But pulling in the incrementality fix shows a regression on the incremental benchmark (which isn't entirely surprising but still a little surprising)

PR #19994, opened 2025-08-19
[ty] Fix namespace packages that behave like partial stubs

Gankra:
In implementing partial stubs I had observed that this continue in the namespace package code seemed erroneous since the same continue for partial stubs didn't work. Unfortunately I wasn't confident enough to push on that hunch. Fortunately I remembered that hunch to make this an easy fix.

The issue with the continue is that it bails out of the current search-path without testing any .py files. This breaks when for example `google` and `google-stubs`/`types-google` are both in the same site-packages dir -- failing to find a module in `types-google` has us completely skip over `google`!

Fixes https://github.com/astral-sh/ty/issues/520

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/partial_stub_packages.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/partial_stub_packages.md
@@ -54,9 +54,9 @@ partial
 ```pyi
 ```
 
-`/packages/foo-stubs/both.py`:
+`/packages/foo-stubs/both.pyi`:
 
-```py
+```pyi
 class Both:
     both: str
     other: int
@@ -107,9 +107,9 @@ extra-paths = ["/packages"]
 ```pyi
 ```
 
-`/packages/foo-stubs/both.py`:
+`/packages/foo-stubs/both.pyi`:
 
-```py
+```pyi
 class Both:
     both: str
     other: int
@@ -165,9 +165,9 @@ extra-paths = ["/packages"]
 ```pyi
 ```
 
-`/packages/foo-stubs/both.py`:
+`/packages/foo-stubs/both.pyi`:
 
-```py
+```pyi
 class Both:
     both: str
     other: int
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/import/partial_stub_packages.md
```diff
@@ -54,9 +54,9 @@ partial
 ```pyi
 ```
 
-`/packages/foo-stubs/both.py`:
+`/packages/foo-stubs/both.pyi`:
 
-```py
+```pyi
```
Gankra, line 59: (All these changes are a driveby fix that doesn't change the results but I realized was wrong and should be fixed)

crates/ty_python_semantic/resources/mdtest/import/partial_stub_packages.md
```diff
@@ -423,3 +423,48 @@ reveal_type(Both().both)  # revealed: str
 reveal_type(Impl().impl)  # revealed: Unknown
 reveal_type(Fake().fake)  # revealed: Unknown
 ```
+
+## Namespace stub with missing module
+
+Namespace stubs are always partial.
```
MichaReiser, line 429: Do you have a reference that backs this claim? This is somewhat surprising to me and I remember that I struggled to find good documentation for what's supposed to happen in some of those non obvious cases

Gankra: https://peps.python.org/pep-0561/#partial-stub-packages

> Type checkers should treat namespace packages within stub-packages as incomplete since multiple distributions may populate them. Regular packages within namespace packages in stub-package distributions are considered complete unless a py.typed with partial\n is included.

MichaReiser: It would be great to add this link to the test

PR #19996, opened 2025-08-19
[ty] Experimental persistent caching

ibraheemdev:
## Summary

Adds persistent caching experimentally under the `TY_PERSIST` environment variable. This currently has many limitations and is not ready for public usage:
- Performance is quite bad and significantly slower than a cold run.
- The size of the file grows unbounded across runs, because we persist stale values.
- The cache file is determined by the environment variable instead of using a `.ty` director with a key for the given project configuration. I intentionally avoided this for now.

Limitations aside, I still want to land this to make it easier to iterate on in-tree.


crates/ty_project/src/db.rs
```diff
@@ -64,18 +66,28 @@ impl ProjectDatabase {
             system: Arc::new(system),
         };
 
+        // Load the database from the persistent cache.
+        //
         // TODO: Use the `program_settings` to compute the key for the database's persistent
         //   cache and load the cache if it exists.
```
ibraheemdev: @MichaReiser I'm not actually sure if we need a cache key, because the `Program` input is persisted, and its fields are updated (bumping the revision) if the settings have changed? A local `.ty_cache` file/directory might be enough.

crates/ty_project/src/metadata.rs
```diff
@@ -35,7 +34,6 @@ pub struct ProjectMetadata {
     /// knowing which files to watch for changes.
     ///
     /// The path ordering doesn't imply precedence.
-    #[cfg_attr(test, serde(skip_serializing_if = "Vec::is_empty"))]
```
ibraheemdev, line 38: `bincode` doesn't support the `serde(skip_serializing_if)` or `serde(untagged)` attributes.

crates/ty_python_semantic/src/semantic_index/scope.rs
```diff
@@ -14,7 +14,13 @@ use crate::{
...
 
 /// A cross-module identifier of a scope that can be used as a salsa query parameter.
-#[salsa::tracked(debug, heap_size=ruff_memory_usage::heap_size)]
+///
+/// This type is an interned struct as opposed to a tracked struct as it is the input of
+/// the `infer_scope_types` query, which is the root query of the persistent cache.
+///
+/// Additionally, this type disables garbage collection, as garbage collectable interned values are volatile,
+/// and would cause `infer_scope_types` to have a dependency on every single `ScopeId` in the file.
+#[salsa::interned(persist, revisions=usize::MAX, debug, heap_size=ruff_memory_usage::heap_size)]
```
ibraheemdev: This is a notable change, and might have an impact on performance.

crates/ty_python_semantic/src/types.rs
```diff
@@ -304,67 +304,88 @@ fn class_lookup_cycle_initial<'db>(
...
+//     Clone,
+//     Debug,
+//     PartialEq,
+//     Eq,
+//     Hash,
+//     get_size2::GetSize,
+//     serde::Serialize,
+//     serde::Deserialize,
+// )]
+// pub struct TodoType(pub &'static str);
```
ibraheemdev: I'm not sure exactly what to do about this, we can't deserialize into a `&'static str` (without leaking). Making this store a `String`/`Cow` is really hard because `Type` is `Copy`. Could we use an enum here?

crates/ty_python_semantic/src/types.rs
```diff
@@ -385,7 +406,7 @@ pub(crate) use todo_type;
 /// # Ordering
 /// Ordering is based on the property instance's salsa-assigned id and not on its values.
 /// The id may change between runs, or when the property instance was garbage collected and recreated.
-#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[salsa::interned(persist, revisions = usize::MAX, debug, heap_size=ruff_memory_usage::heap_size)]
```
ibraheemdev: I disabled garbage collection for all interned values because it significantly reduces the amount of data we need to persist, but I'm not sure if that change should be made in this PR.

carljm (commented):
This looks generally fine to me, assuming CI is cleared up and there are no observed regressions. Ideally I'd prefer @MichaReiser to take a look, if he has time.

If there is a noticeable performance regression for non-persisting cases, I'd be hesitant to pay that cost to land an experimental feature that we haven't even confirmed can be a net win, and I'd hope we can find a way to eliminate that regression, even if it's via conditional compilation.

crates/ty_project/src/db.rs
MichaReiser: A not uncommon use case is to check a project with different Python versions: Each run then uses a different `ProgramSettings`. Ideally, the caches would be separate

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -868,30 +880,51 @@ pub enum DiagnosticTag {
 /// be in kebab case, e.g. `no-foo` (all lower case).
 ///
 /// Rules use kebab case, e.g. `no-foo`.
-#[derive(Debug, Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash, get_size2::GetSize)]
-pub struct LintName(&'static str);
```
MichaReiser, line 872: It's not the end of the world but it seems unfortunate if we have to change such fundamental types for an experimental feature. 

The alternative here would be to retrieve the lint registry from the database (or a thread local) during deserialization and to use it to resolve the lint metadata. 

https://github.com/astral-sh/ruff/blob/99d0ac60b417ee623bb9257a6502a7c21c444b28/crates/ty_python_semantic/src/db.rs#L14

crates/ruff_source_file/src/lib.rs
```diff
@@ -165,6 +165,7 @@ impl SourceFileBuilder {
 /// Cloning a [`SourceFile`] is cheap, because it only requires bumping a reference count.
 #[derive(Clone, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "get-size", derive(get_size2::GetSize))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
MichaReiser: I'd prefer if we don't add serialize implementation for those types. We could simply panic in the `UnifiedFile` implementation whenever we encounter a ruff file.

The reason is that it seems unnecessary to maintain this complexity when Ruff doesn't even use salsa's caching and serializing the line index is a little useles (it's very cheap to build it from the source)

crates/ty_project/src/db.rs
```diff
@@ -126,6 +138,59 @@ impl ProjectDatabase {
...
+    /// Deserialize the database from the persistent cache.
+    pub fn deserialize(&mut self, path: &str, metadata: &ProjectMetadata) -> anyhow::Result<()> {
+        // Read from the persistent cache.
+        let contents = match std::fs::read(path) {
+            Ok(contents) => contents,
+            Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(()),
+            Err(err) => return Err(err.into()),
+        };
+
+        let _guard = ValueSourceGuard::new(ValueSource::Cli, false);
```
MichaReiser: Does this mean that all deserialized options now have a source CLI even if they were originally loaded from a settings file?

crates/ty_project/src/db.rs
```diff
@@ -126,6 +138,59 @@ impl ProjectDatabase {
...
+            self.files.seed_root(root, self);
+        }
+
+        for project in Project::load_all(self) {
+            if project.root(self) == metadata.root() {
+                project.resolve_settings(self)?;
+                self.project = Some(project);
+                break;
+            }
+        }
```
MichaReiser: What happens if the user moved the `metadata` file (e.g. if they moved it a directory up)? Do we replace the metadata somewhere (by calling update project), or do we use a stale configuration?

I think what would be ideal is if we wouldn't serialize the project settings at all and always load them instead. But that would require some form of hashing to know when they've changed (including which fields) which is out of scope for now

crates/ty_project/src/lib.rs
```diff
@@ -81,8 +82,11 @@ pub struct Project {
     /// We box the metadata here because it's a fairly large type and
     /// reducing the size of `Project` helps reduce the size of the
     /// salsa allocated table for `Project`.
-    #[returns(deref)]
-    pub settings: Box<Settings>,
+    ///
+    /// This field is unintialized temporarily after deserialization,
+    /// until `Project::resolve_settings` is called.
```
MichaReiser: and the reason for the `OnceLock` is so that we can change the field without bumping the salsa revision?

crates/ty_python_semantic/src/python_platform.rs
```diff
@@ -16,7 +16,6 @@ pub enum PythonPlatform {
     /// We use a string (instead of individual enum variants), as the set of possible platforms
     /// may change over time. See <https://docs.python.org/3/library/sys.html#sys.platform> for
     /// some known platform identifiers.
-    #[cfg_attr(feature = "serde", serde(untagged))]
```
MichaReiser, line 19: `PythonPlatform` is used in the user-facing options struct: I don't see any schema changes, but we can't remove it here if it does change the schema (or we have to use another type in the user-facing options)

crates/ty_python_semantic/src/semantic_index/place.rs
```diff
@@ -133,7 +133,18 @@ impl std::fmt::Display for PlaceExprRef<'_> {
...
+    Debug,
+    Copy,
+    Clone,
+    Eq,
+    PartialEq,
+    Hash,
+    get_size2::GetSize,
+    salsa::Update,
+    serde::Serialize,
+    serde::Deserialize,
```
MichaReiser, line 146: Do all the `serialize` / `deserialize` impls notably impact the build time?

crates/ty_python_semantic/src/semantic_index/scope.rs
MichaReiser: Hmm, this seems very unfortunate as it now introduces a global lock into semantic index building (which I think is the first?). 

Can you run some benchmarks on large projects to see how it impacts performance? It seems a very unfortunate change for an experimental feature where it isn't even clear if we'll use it long term.

crates/ty_python_semantic/src/types.rs
MichaReiser: We could do the same as what you did with `AstNodeRef` where we use an `Option<&'static str>` and set it to `None` when deserializing. This could lead to issues downstream because types now suddenly compare equal that didn't before but I wonder if that should even be the case, given that the release implementation's equal implementation never compares the string.  

An enum is an option too

I don't think we should spend too much time working around this because the ultimate goal is to remove this type entirely


CC: @sharkdp

crates/ty_python_semantic/src/types.rs
MichaReiser: I think this should be a separate PR and requires some reasoning why it's fine to do this

MichaReiser (commented):
Thank you.

I need to take a closer look at the project serialization/deserialization. It's not entirely clear to me how that works. 

Overall, I'd prefer if this PR made fewer changes to the overall code base, given that this is an experimental feature that we might decide not to ship. This is mainly about changing `LintName` to use a `Cow` and disabling interning. For the latter, I think the solution is to gate the more questionable changes behind a feature flag. This allows us to pull this change in without regressing the production experience and gives us some more time to a) figure out a solution for those problem b) decide that its what we want, given that it enables persistent caching.

crates/ty_python_semantic/src/types.rs
sharkdp: > We could do the same as what you did with `AstNodeRef` where we use an `Option<&'static str>` and set it to `None` when deserializing. This could lead to issues downstream because types now suddenly compare equal that didn't before but I wonder if that should even be the case, given that the release implementation's equal implementation never compares the string.

Hmm.. yes. I think we should be fine if all `TodoType`s compare equal. Maybe that will lead to some observable differences (different todo messages showing up in `reveal_type` assertions?), but those differences should never be behavioral. All of these types are completely dynamic types and should behave the same way. And if it's not consistent between debug and release right now, maybe that's something we should fix anyway (always have them compare equal, because there's no other way in release mode).

Instead of `Option<&'static str>` you could maybe also deserialize these messages as `"<deserialized>"`, making it clear that the todo-messages are not persisted?

> I don't think we should spend too much time working around this because the ultimate goal is to remove this type entirely

I agree.

crates/ty_project/src/db.rs
ibraheemdev: The only place I see the options being created is [with `ValueSource::Cli` here](https://github.com/astral-sh/ruff/blob/main/crates/ty/src/args.rs#L184). Are they created elsewhere with a different source?

crates/ty_project/src/db.rs
ibraheemdev: This was incorrect, I updated the check to be `project.metadata(db) == metadata`. If the metadata has changed, i.e. there is no project with matching metadata, a new `Project` input will be created.

crates/ty_project/src/lib.rs
ibraheemdev: Yes, the settings should not change if the metadata has not changed.

crates/ty_python_semantic/src/python_platform.rs
ibraheemdev: Should there be failing tests if the schema changes? This does change the representation.

MichaReiser: Yes, there should be a failing test

crates/ty_project/src/db.rs
MichaReiser: https://github.com/astral-sh/ruff/blob/f34b65b7a00dfc61c5a54e4280b421c94151e677/crates/ty_project/src/metadata.rs#L166-L177

ibraheemdev: Right, should we serialize the source at the top-level (in `Options`)?

crates/ty_python_semantic/src/python_platform.rs
ibraheemdev: I rewrote the serde implementations in a way that bincode accepts, but there should be no change in behavior.

crates/ty_project/src/metadata.rs
ibraheemdev: @MichaReiser is this one public facing?

crates/ty_project/src/db.rs
MichaReiser: We can't. The options are merged together from multiple sources. We'd need to preserve them per value, which might be tricky

ibraheemdev: Is there a reason that `RangedValue` ignores the source when serializing/deserializing in the first place?

crates/ty_python_semantic/src/semantic_index/scope.rs
ibraheemdev: I'm actually seeing a 7% speedup locally?

```bash
$ hyperfine "./target/release/ty-new check --quiet --python ./prefect/.venv" "./target/release/ty-old check --quiet --python ./prefect/.venv" --ignore-failure --warmup 10
Benchmark 1: ./target/release/ty-new check --quiet --python ./prefect/.venv
  Time (mean ± σ):     674.4 ms ±  21.7 ms    [User: 7657.4 ms, System: 1588.2 ms]
  Range (min … max):   644.4 ms … 721.5 ms    10 runs
 
Benchmark 2: ./target/release/ty-old check --quiet --python ./prefect/.venv
  Time (mean ± σ):     722.3 ms ±  13.0 ms    [User: 7699.8 ms, System: 1694.5 ms]
  Range (min … max):   703.2 ms … 744.5 ms    10 runs
 
Summary
  ./target/release/ty-new check --quiet --python ./prefect/.venv ran
    1.07 ± 0.04 times faster than ./target/release/ty-old check --quiet --python ./prefect/.venv
```

Codspeed [reports a similar speedup on the multithreaded benchmark](https://codspeed.io/astral-sh/ruff/runs/compare/68a622dc2992c5c139cf3e1e..68a6249e7162a5372439a629).

crates/ty_project/src/db.rs
MichaReiser: It only ignores it when serializing. The deserialization uses the serde_toml's spanned feature to get the span and we get the source from the thread local:

https://github.com/astral-sh/ruff/blob/f34b65b7a00dfc61c5a54e4280b421c94151e677/crates/ty_project/src/metadata/value.rs#L281-L308

We don't serialize the span and source because the range is only for better diagnostics. It isn't actually part of the data itself. Now, the good news is that we only use serialization for tests. So we would consider serializing the ranges (or only if a specific thread local is set) and deserialize the ranges from the data if present (over using the thread local). I guess, we could always serialize/deserialize the source if the thread local isn't set

ibraheemdev: I updated the code to use the new `ProjectMetadata` when resolving the settings (and program settings), so the new value sources are propagated.

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -881,6 +888,64 @@ impl LintName {
...
+        where
+            D: serde::Deserializer<'de>,
+        {
+            let name: &str = serde::Deserialize::deserialize(deserializer)?;
+
+            LINT_REGISTRY.with_borrow(|registry| {
+                let registry = registry
+                    .expect("must set the `LintRegistryGuard` when deserializing a `LintName`");
+
+                registry(name).ok_or(serde::de::Error::custom("invalid `LintName`"))
```
MichaReiser: Should we include the name in the error message or does serde do this for us?

crates/ruff_db/src/files/file_root.rs
```diff
@@ -35,9 +35,19 @@ impl FileRoot {
...
         self.kind_at_time_of_creation(db).durability()
     }
+
+    /// Loads all existing [`FileRoot`]s in the database.
+    pub fn load_all(db: &dyn Db) -> Vec<FileRoot> {
+        FileRoot::ingredient(db)
+            .entries(db.zalsa())
+            .map(|(key, _)| salsa::plumbing::FromId::from_id(key.key_index()))
+            .collect()
+    }
```
MichaReiser: We'll want to have some form of garbage collection here to aovid creating file roots for paths that no longer exists but we can defer that to later.

crates/ruff_db/src/files.rs
```diff
@@ -259,6 +265,29 @@ impl Files {
             root.set_revision(db).to(FileRevision::now());
         }
     }
+
+    /// Seed the files with an existing [`File`] instance.
+    pub fn seed(&self, file: File, db: &dyn Db) {
+        match file.path(db) {
+            FilePath::System(path) => {
+                self.inner.system_by_path.insert(path.clone(), file);
```
MichaReiser: It might be worth asserting here that no such file existed and panic otherwise

crates/ruff_db/src/files.rs
```diff
@@ -414,6 +443,14 @@ impl File {
         }
     }
 
+    /// Loads all existing [`File`]s in the database.
+    pub fn load_all(db: &dyn Db) -> Vec<File> {
+        File::ingredient(db)
+            .entries(db.zalsa())
+            .map(|(key, _)| salsa::plumbing::FromId::from_id(key.key_index()))
```
MichaReiser: The `key.key_index` feels a bit magical and we shouldn't have to use an API from `salsa::plumbing`. Could `entries` return a struct that provides an `as_struct()` method (or similar) to encapsulate this logic (I'm fine if this is done as a follow up, it just feels a bit off)

crates/ty_project/src/metadata/value.rs
```diff
@@ -57,14 +57,14 @@ thread_local! {
     static VALUE_SOURCE: RefCell<Option<(ValueSource, bool)>> = const { RefCell::new(None) };
 }
 
-/// Guard to safely change the [`VALUE_SOURCE`] for the current thread.
+/// Guard to safely change the [`ValueSource`] for the current thread.
 #[must_use]
-pub(super) struct ValueSourceGuard {
+pub struct ValueSourceGuard {
```
MichaReiser, line 62: Can we leave this `pub(super)`, now that we load the settings from the "fresh" metadata?

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+    /// Deserialize the database from the persistent cache.
+    pub fn deserialize(&mut self, path: &str, metadata: &ProjectMetadata) -> anyhow::Result<()> {
+        // Read from the persistent cache.
+        let contents = match std::fs::read(path) {
+            Ok(contents) => contents,
+            Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(()),
+            Err(err) => return Err(err.into()),
+        };
+
+        let _value_source = ValueSourceGuard::new(ValueSource::Cli, false);
```
MichaReiser: Can we add a comment explaining why it's okay to use `ValueSource::Cli` here

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+        let _lint_registry = LintRegistryGuard::new(|name| {
+            default_lints_registry()
+                .get(name)
+                .map(|lint| lint.name())
+                .ok()
+        });
+
```
MichaReiser: Can we use `DEFAULT_LINT_REGISTRY` instead?  It avoids building the lint registry everytime we lookup a rule.

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+        );
+
+        <dyn salsa::Database>::deserialize(self, decoder.as_deserializer())?;
+
+        // Sync any deserialized inputs.
+        for file in File::load_all(self) {
+            file.sync(self);
+            self.files.seed(file, self);
+        }
+
```
MichaReiser: It may help if we run this multi threaded, but that's something we can leave for a separate PR

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+    /// Persist the current state of the database to disk.
+    pub fn persist(&mut self, path: &str) -> anyhow::Result<()> {
+        let contents = bincode::serde::encode_to_vec(
+            <dyn salsa::Database>::as_serialize(self),
+            bincode::config::standard(),
+        )?;
+
+        std::fs::write(path, contents)?;
+
```
MichaReiser: Nit: Maybe use `encode_into_std_write`

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+                .ok()
+        });
+
+        // Deserialize the database.
+        let mut decoder = bincode::serde::BorrowedSerdeDecoder::from_slice(
+            &contents,
+            bincode::config::standard(),
+            (),
+        );
+
```
MichaReiser, line 166: Nit: Maybe use `decode_from_std_read`

Edit: Actually, we can't do that. We need to use `system.read_to_string` and use `system.as_writeable().write_file` to write the changes back

crates/ty_project/src/files.rs
```diff
@@ -20,7 +20,7 @@ use crate::{IOErrorDiagnostic, Project};
 /// so no query can be depending on the contents of the indexed files before that. All subsequent mutations to
 /// the indexed files must go through `IndexedMut`, which uses the Salsa setter `project.set_file_set` to
 /// ensure that Salsa always knows when the set of indexed files have changed.
-#[derive(Debug, get_size2::GetSize)]
+#[derive(Debug, get_size2::GetSize, serde::Serialize, serde::Deserialize)]
```
MichaReiser: I think we should skip serializing `IndexedFiles` because we need to reload all files anyway. Unless that isn't easily possible because it then always results in a change?

crates/ty_project/src/lib.rs
```diff
@@ -81,8 +82,11 @@ pub struct Project {
     /// We box the metadata here because it's a fairly large type and
     /// reducing the size of `Project` helps reduce the size of the
     /// salsa allocated table for `Project`.
-    #[returns(deref)]
-    pub settings: Box<Settings>,
+    ///
+    /// This field is uninitialized temporarily after deserialization,
+    /// until `Project::resolve_settings` is called.
```
MichaReiser: Let's extend this documentation to also mention why it uses a `OnceLock` instead of an `Option`. Same for the other fields

crates/ty_project/src/db.rs
```diff
@@ -66,18 +68,27 @@ impl ProjectDatabase {
             system: Arc::new(system),
         };
 
+        // Load the database from the persistent cache.
+        //
         // TODO: Use the `program_settings` to compute the key for the database's persistent
-        //   cache and load the cache if it exists.
-        //   we may want to have a dedicated method for this?
+        //       cache, to support multiple persistent caches for different configurations.
+        if let Ok(path) = std::env::var(EnvVars::TY_PERSIST) {
```
MichaReiser:
```suggestion
        if let Ok(path) = system.env_var(EnvVars::TY_PERSIST) {
```

To avoid your local `TY_PERSIST` environment variable leaking into tests

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
         )
     }
 
+    /// Deserialize the database from the persistent cache.
+    pub fn deserialize(&mut self, path: &str, metadata: &ProjectMetadata) -> anyhow::Result<()> {
```
MichaReiser:
```suggestion
    fn deserialize(&mut self, path: &str, metadata: &ProjectMetadata) -> anyhow::Result<()> {
```

crates/ty_project/src/db.rs
```diff
@@ -126,6 +137,64 @@ impl ProjectDatabase {
...
+            file.sync(self);
+            self.files.seed(file, self);
+        }
+
+        for root in FileRoot::load_all(self) {
+            self.files.seed_root(root, self);
+        }
+
+        for project in Project::load_all(self) {
+            if project.resolve_settings(metadata, self)? {
```
MichaReiser: Nit: `resolve_settings` seems to be only used by deserialization but its name doesn't suggest that. Maybe `init_after_deserialization`?

crates/ty_project/src/metadata.rs
MichaReiser: No, not yet :) We only use serialization in tests (that's why `cfg_attr(test)`)

crates/ty_python_semantic/src/semantic_index/scope.rs
```diff
@@ -14,7 +14,13 @@ use crate::{
 };
 
 /// A cross-module identifier of a scope that can be used as a salsa query parameter.
-#[salsa::tracked(debug, heap_size=ruff_memory_usage::heap_size)]
+///
+/// This type is an interned struct as opposed to a tracked struct as it is the input of
+/// the `infer_scope_types` query, which is the root query of the persistent cache.
+///
+/// Additionally, this type disables garbage collection, as garbage collectable interned values are volatile,
+/// and would cause `infer_scope_types` to have a dependency on every single `ScopeId` in the file.
```
MichaReiser, line 22: Let's add a `TODO` to consider reverting to a tracked struct once supported by the persistent caching backend.

crates/ty_static/src/env_vars.rs
```diff
@@ -39,6 +40,12 @@ impl EnvVars {
     /// when necessary, e.g. to watch for file system changes or a dedicated UI thread.
     pub const TY_MAX_PARALLELISM: &'static str = "TY_MAX_PARALLELISM";
 
+    /// Used to enable persistent caching at the specified path.
+    ///
+    /// This improves performance and memory usage by caching to disk across ty runs.
```
MichaReiser, line 45: Let's add a note that this is highly experimental and currently makes ty slower.

MichaReiser (approved):
Thank you. This looks great. I've a few smaller comments that would be great to address before landing but nothing major. 

The only thing we should do before landing is to add a CLI test that exercises the `persist`, `load` to prevent us from accidentially breaking this feature. It would be nice to have more extensive tests that verify that e.g. changes to the files that need checking or settings propagate as expected but this is something we can add when we fully commit to persistent caching.

See the `cli` module for how to add CLI tests 

https://github.com/astral-sh/ruff/blob/4daf59e5e7f4613eeb5b16616d6f1021297fc12e/crates/ty/tests/cli/main.rs#L18

crates/ty_python_semantic/src/semantic_index/scope.rs
MichaReiser: It's funny that this is the case and i don't fully understand it. Especially considering that it doubles the metadata size for `ScopeId`

main:

`ScopeId`                                          metadata=27.84MB  fields=11.93MB  count=994370

this branch:

`ScopeId`                                          metadata=51.71MB  fields=11.93MB  count=994346

crates/ty_project/src/db.rs
ibraheemdev: I'm also not sure if it would be faster to deserialize streaming. serde_json at least has performance issues with this and recommends reading into a temporary buffer first.

crates/ty_project/src/metadata/value.rs
ibraheemdev: We still need it to set the dummy `ValueSource::Cli` values.

crates/ty_python_semantic/src/semantic_index/scope.rs
ibraheemdev: Hmm, I'm not sure how this would work without persisting the `semantic_index` query.

crates/ty_project/src/files.rs
ibraheemdev: Hmm.. updating the `file_set` field is problematic. The `semantic_index` query has a dependency on `file_set` (for example, `report_semantic_error` checks if the file is in the set), and so `file_set` ends up in the flattened dependencies of `infer_scope_types`.

I think we need to make  `should_check_file` a tracked function so that the `infer_scope_types` queries have finer-grained dependencies on specific files instead of the entire set. I'm going to leave this as a TODO for now as it also requires Salsa changes (to avoid flattening transitive queries if they're marked as persistable).

ibraheemdev:
I think I've addressed most of the comments, but this still needs tests. I also want to do some benchmark to see why making `ScopeId` an interned struct is a performance improvement.

carljm:
What should we do with this PR? It's quite conflicted by now, seems like if we decide to revisit this we might just do this PR from scratch?

ibraheemdev:
Going to close this because it would likely be easier to start from scratch at this point.

PR #19997, opened 2025-08-19
[ty] Add constraint set implementation

dcreager:
This PR adds an implementation of constraint sets.

An individual constraint restricts the specialization of a single typevar to be within a particular lower and upper bound: the typevar can only specialize to types that are a supertype of the lower bound, and a subtype of the upper bound. (Note that lower and upper bounds are fully static; we take the bottom and top materializations of the bounds to remove any gradual forms if needed.) Either bound can be “closed” (where the bound is a valid specialization), or “open” (where it is not).

You can then build up more complex constraint sets using union, intersection, and negation operations. We use a disjunctive normal form (DNF) representation, just like we do for types: a _constraint set_ is the union of zero or more _clauses_, each of which is the intersection of zero or more individual constraints. Note that the constraint set that contains no clauses is never satisfiable (`⋃ {} = 0`); and the constraint set that contains a single clause, which contains no constraints, is always satisfiable (`⋃ {⋂ {}} = 1`).

One thing to note is that this PR does not change the logic of the actual assignability checks, and in particular, we still aren't ever trying to create an "individual constraint" that constrains a typevar. Technically we're still operating only on `bool`s, since we only ever instantiate `C::always_satisfiable` (i.e., `true`) and `C::unsatisfiable` (i.e., `false`) in the `has_relation_to` methods. So if you thought that #19838 introduced an unnecessarily complex stand-in for `bool`, well here you go, this one is worse! (But still seemingly not yielding a performance regression!) The next PR in this series, #20093, is where we will actually create some non-trivial constraint sets and use them in anger.

That said, the PR does go ahead and update the assignability checks to use the new `ConstraintSet` type instead of `bool`. That part is fairly straightforward since we had already updated the assignability checks to use the `Constraints` trait; we just have to actively choose a different impl type. (For the `is_whatever` variants, which still return a `bool`, we have to convert the constraint set, but the explicit `is_always_satisfiable` calls serve as nice documentation of our intent.)


dcreager:
I'm excited to see how much slower this is than `bool`... :grimacing:

dcreager:
This is ready for review! #20093 is the real proof that this representation works well. In some ways, this PR is just a setup for that, even though we're introducing a pretty complex new data structure here.

crates/ty_python_semantic/src/types.rs
```diff
@@ -9617,6 +9621,19 @@ pub(super) fn walk_intersection_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>
...
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<Type<'db>>,
+    {
+        elements
+            .into_iter()
+            .fold(IntersectionBuilder::new(db), |builder, element| {
+                builder.add_positive(element.into())
+            })
+            .build()
```
```suggestion
        IntersectionBuilder::new(db).positive_elements(elements).build()
```

there might be some other existing uses of `IntersectionBuilder::positive_elements()` that could be switched over to use this new method, too

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -12,21 +12,37 @@
...
-//! work on any impl of that trait. But the only impl we have right now is `bool`, which means that
-//! we are still not tracking the full detail as promised in the description above. (`bool` is a
-//! perfectly fine impl, but it can generate false positives when you have to break down a
-//! particular assignability check into subchecks: each subcheck might say "yes", but technically
-//! under conflicting constraints, which a single `bool` can't track.) Soon we will add a proper
-//! constraint set implementation, and the `bool` impl of the trait (and possibly the trait itself)
-//! will go away.
+//! An individual constraint restricts the specialization of a single typevar to be within a
+//! particular lower and upper bound: the typevar can only specialize to types that are a supertype
+//! of the lower bound, and a subtype of the upper bound. (Note that lower and upper bounds are
```
```suggestion
//! particular lower and upper bound: the typevar can only specialize to a type that is a supertype
//! of the lower bound, and a subtype of the upper bound. (Note that lower and upper bounds are
```

or

```suggestion
//! particular lower and upper bound: the typevar can only specialize to types that are supertypes
//! of the lower bound, and subtypes of the upper bound. (Note that lower and upper bounds are
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -12,21 +12,37 @@
...
+//! fully static; we take the bottom and top materializations of the bounds to remove any gradual
+//! forms if needed.) Either bound can be "closed" (where the bound is a valid specialization), or
+//! "open" (where it is not).
```
AlexWaygood, line 22: Would it be possible to give Python examples here of the difference between closed and open bounds?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -12,21 +12,37 @@
...
+//!
+//! You can then build up more complex constraint sets using union, intersection, and negation
+//! operations. We use a distributed normal form (DNF) representation, just like we do for types: a
+//! [constraint set][ConstraintSet] is the union of zero or more [clauses][ConstraintClause], each
+//! of which is the intersection of zero or more [individual constraints][AtomicConstraint]. Note
+//! that the constraint set that contains no clauses is never satisfiable (`⋃ {} = 0`); and the
+//! constraint set that contains a single clause, which contains no constraints, is always
+//! satisfiable (`⋃ {⋂ {}} = 1`).
```
```suggestion
//! constraint set that contains a single clause, where that clause contains no constraints,
//! is always satisfiable (`⋃ {⋂ {}} = 1`).
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+/// [constraints][AtomicConstraint].
+///
+/// This is called a "set of constraint sets", and denoted _𝒮_, in [[POPL2015][]].
+///
+/// [POPL2015]: https://doi.org/10.1145/2676726.2676991
+///
+/// ### Invariants
+///
+/// - The clauses are simplified as much as possible — there are no two clauses in the set that can
+///   be simplified into a single clause.
```
```suggestion
/// ### Invariants
///
/// - The clauses are simplified as much as possible — there are no two clauses in the set that can
///   be simplified into a single clause.
///
/// [POPL2015]: https://doi.org/10.1145/2676726.2676991
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+#[derive(Clone, Debug)]
+pub(crate) struct ConstraintSet<'db> {
+    clauses: SmallVec<[ConstraintClause<'db>; 2]>,
```
AlexWaygood, line 231: Why 2 here? Is it worth adding a comment?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+    fn union_clause(&mut self, db: &'db dyn Db, mut clause: ConstraintClause<'db>) {
+        // Naively, we would just append the new clause to the set's list of clauses. But that
+        // doesn't ensure that the clauses are simplified with respect to each other. So instead,
+        // we iterate through the list of existing clauses, and try to simplify the new clause
+        // against each one in turn. (We can assume that the existing clauses are already
+        // simplified with respect to each other, since we can assume that the invariant holds upon
+        // entry to this method.)
+        let prev = std::mem::take(&mut self.clauses);
+        let mut existing_clauses = prev.into_iter();
+        while let Some(existing) = existing_clauses.next() {
```
```suggestion
        let mut existing_clauses = std::mem::take(&mut self.clauses).into_iter();
        for existing in existing_clauses.by_ref() {
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+                        self.clauses.push(ConstraintClause::always());
+                        return;
+                    }
+                    Satisfiable::Constrained(clause) => self.union_clause(db, clause),
+                }
+            }
+        }
+    }
+
+    #[allow(dead_code)]
```
```suggestion
    #[expect(dead_code)]
```

what's the reason for having the method at all -- just for debugging? Is it worth adding a comment?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+/// The intersection of zero or more atomic constraints.
+///
+/// This is called a "constraint set", and denoted _C_, in [[POPL2015][]].
+///
+/// [POPL2015]: https://doi.org/10.1145/2676726.2676991
+///
+/// ### Invariants
+///
+/// - No two constraints in the clause will constrain the same typevar.
+/// - The constraints are sorted by typevar.
```
```suggestion
/// ### Invariants
///
/// - No two constraints in the clause will constrain the same typevar.
/// - The constraints are sorted by typevar.
///
/// [POPL2015]: https://doi.org/10.1145/2676726.2676991
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+    fn is_always(&self) -> bool {
+        self.constraints.is_empty()
+    }
+
+    /// Returns the intersection of this clause and an atomic constraint.
+    fn intersect_constraint(
+        &mut self,
+        db: &'db dyn Db,
+        constraint: AtomicConstraint<'db>,
+    ) -> Satisfiable<()> {
```
AlexWaygood: It's a bit surprising to me that this method mutates internal state but then returns a non-`Self` type -- it's not a pure function but it's also not an in-place mutation exactly. It feels like it might be worth calling out in the doc-comment that it mutates `self` internally before returning a `Satisfiable` instance (and stating why it does that)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+        }
+        if result.is_always() {
+            Satisfiable::Always
+        } else {
+            Satisfiable::Constrained(result)
+        }
+    }
+
+    /// Tries to simplify the union of two clauses into a single clause, if possible.
+    fn simplify_clauses(self, db: &'db dyn Db, other: Self) -> Simplifiable<Self> {
```
AlexWaygood, line 523: The comments for this method are great. It would be amazing if we could maybe have some Python examples as well as all the maths, though

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+            }
+        }
+        true
+    }
+
+    /// If the union of two clauses is simpler than either of the individual clauses, returns the
+    /// union. This happens when (a) they mention the same set of typevars, (b) the union of the
+    /// constraints for exactly one typevar simplifies to a single constraint, and (c) the
+    /// constraints for all other typevars are identical. Otherwise returns `None`.
+    fn simplifies_via_distribution(&self, db: &'db dyn Db, other: &Self) -> Option<Self> {
```
AlexWaygood, line 662: Here again, I feel like an example might be really helpful to illustrate how this relates to type-checking Python code

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+
+        DisplayConstraintClause { clause: self, db }
+    }
+}
+
+/// A constraint on a single typevar, providing lower and upper bounds for the types that it can
+/// specialize to. The lower and upper bounds can each be either _closed_ (the bound itself is
+/// included) or _open_ (the bound itself is not included).
+///
+/// We render constraints using an interval notation: `[s, t]`, `(s, t)`, `[s, t)`, or `(s, t]`,
```
AlexWaygood: "Render" as in "render in `Display` implementations"?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+/// We will also sometimes render them graphically: `s┠──┨t`, `s╟──╢t`, `s┠──╢t`, or `s╟──┨t`,
+/// where a solid bar indicates a closed bound, and a double bar indicates an open  bound.
+///
+/// ### Invariants
+///
+/// - The bounds must be fully static.
+/// - The bounds must actually constrain the typevar. If the typevar can be specialized to any
+///   type, or if there is no valid type that it can be specialized to, then we don't create an
+///   `AtomicConstraint` for the typevar.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```
AlexWaygood: should we maybe enforce that users have to explicitly `.clone()` instances of this struct if they want to copy the data? It has several fields; it might be cheaper to take references where possible

```suggestion
#[derive(Clone, Debug, Eq, PartialEq)]
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+    typevar: BoundTypeVarInstance<'db>,
+    lower: ConstraintBound<'db>,
+    upper: ConstraintBound<'db>,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ConstraintBound<'db> {
+    Open(Type<'db>),
+    Closed(Type<'db>),
+}
```
AlexWaygood, line 800: How did you consider the tradeoffs of an enum here vs a struct that wrapped an enum, e.g.

```rs
#[derive(Clone, Debug, Eq, PartialEq)]
pub(crate) enum ConstraintBound<'db> {
    value: Type<'db>,
    kind: ConstraintBoundKind,
}

#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum ConstraintBoundKind {
    Open,
    Closed,
}
```

^That kind of representation might remove the need for the `ConstraintBound::bound_type()` method, for example -- the information would be right there in the `value` field

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+            ConstraintBound::Closed(bound) => bound,
+        }
+    }
+
+    fn is_open(self) -> bool {
+        matches!(self, ConstraintBound::Open(_))
+    }
+
+    /// Returns the minimum of two upper bounds. (This produces the upper bound of the
+    /// [intersection][AtomicConstraint::intersect] of two constraints.)
```
AlexWaygood, line 822: Would it be possible to include a short Python snippet in this doc-comment that includes two typevars with upper bounds and explains what `.min_upper()` would return in that case? And similar for some of the other methods on this enum?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+            }
+            _ => ConstraintBound::Closed(result_bound),
+        }
+    }
+}
+
+impl<'db> AtomicConstraint<'db> {
+    /// Returns a new atomic constraint.
+    ///
+    /// Panics of `lower` and `upper` are not both fully static.
```
```suggestion
    /// Panics if `lower` and `upper` are not both fully static.
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+    fn new(
+        db: &'db dyn Db,
+        typevar: BoundTypeVarInstance<'db>,
+        lower: ConstraintBound<'db>,
+        upper: ConstraintBound<'db>,
+    ) -> Satisfiable<Self> {
+        let lower_type = lower.bound_type();
+        let upper_type = upper.bound_type();
+        debug_assert!(lower_type == lower_type.bottom_materialization(db));
+        debug_assert!(upper_type == upper_type.top_materialization(db));
```
```suggestion
        debug_assert_eq!(lower_type, lower_type.bottom_materialization(db));
        debug_assert_eq!(upper_type, upper_type.top_materialization(db));
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+            self.lower.max_lower(db, other.lower),
+            self.upper.min_upper(db, other.upper),
+        )
+    }
+
+    /// Returns the union of this atomic constraint and another.
+    ///
+    /// Panics if the two constraints have different typevars.
+    fn union(self, db: &'db dyn Db, other: Self) -> Simplifiable<Self> {
+        debug_assert!(self.typevar == other.typevar);
```
```suggestion
        debug_assert_eq!(self.typevar, other.typevar);
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+                ConstraintBound::Closed(Type::object(db)),
+            ),
+        );
+        result
+    }
+
+    /// Returns whether `self` has tighter bounds than `other` — that is, if the intersection of
+    /// `self` and `other` is `self`.
+    fn subsumes(self, db: &'db dyn Db, other: Self) -> bool {
+        debug_assert!(self.typevar == other.typevar);
```
```suggestion
        debug_assert_eq!(self.typevar, other.typevar);
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+        self.intersect(db, other) == Satisfiable::Constrained(self)
+    }
+
+    /// Returns the intersection of this atomic constraint and another.
+    ///
+    /// Panics if the two constraints have different typevars.
+    fn intersect(self, db: &'db dyn Db, other: Self) -> Satisfiable<Self> {
+        debug_assert!(self.typevar == other.typevar);
```
```suggestion
        debug_assert_eq!(self.typevar, other.typevar);
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -318,6 +313,31 @@ impl Display for DisplayRepresentation<'_> {
     }
 }
 
+impl<'db> BoundTypeVarInstance<'db> {
+    #[allow(dead_code)]
```
```suggestion
    #[expect(dead_code)]
```

Could we add a comment to say why we're keeping the method around at all for now, rather than removing it?

AlexWaygood (commented):
This looks cool! I haven't done a deep review of the code for correctness -- this is mainly a docs review :-)

AlexWaygood:
Should this PR still have "WIP" in its title? 😄

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum Simplifiable<T> {
+    NeverSatisfiable,
+    AlwaysSatisfiable,
+    Simplified(T),
+    NotSimplified(T, T),
+}
+
+impl<T> Simplifiable<T> {
+    fn from_one(constraint: Satisfiable<T>) -> Self {
```
AlexWaygood, line 1156:
```suggestion
    const fn from_one(constraint: Satisfiable<T>) -> Self {
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +169,876 @@ where
...
+    fn bound_type(self) -> Type<'db> {
+        match self {
+            ConstraintBound::Open(bound) => bound,
+            ConstraintBound::Closed(bound) => bound,
+        }
+    }
+
+    fn is_open(self) -> bool {
+        matches!(self, ConstraintBound::Open(_))
+    }
```
```suggestion
    const fn flip(self) -> Self {
        match self {
            ConstraintBound::Open(bound) => ConstraintBound::Closed(bound),
            ConstraintBound::Closed(bound) => ConstraintBound::Open(bound),
        }
    }

    const fn bound_type(self) -> Type<'db> {
        match self {
            ConstraintBound::Open(bound) => bound,
            ConstraintBound::Closed(bound) => bound,
        }
    }

    const fn is_open(self) -> bool {
        matches!(self, ConstraintBound::Open(_))
    }
```

dcreager:
> Should this PR still have "WIP" in its title? 😄

Nope! Removed

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Yep, I use `printf` debugging a lot, and I find it useful for all of our non-trivial types to have `display` methods or `Display` impls always available, even if we're not currently using them in anything user-facing.

I'm happy to add a comment to that effect at each place where there's a `dead_code` annotation, though I'd also like to suggest this as a team norm if there aren't any objections!

crates/ty_python_semantic/src/types/constraints.rs
dcreager: It was originally clippy that suggested this, but that was back when the bounds were just `Type`s and not `ConstraintBound`s. That might make this type big enough that the calculus changes. Will investigate

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Turns out this one can't be `const` because of a custom `Drop` impl somewhere

crates/ty_python_semantic/src/types/constraints.rs
dcreager: I tried that just now on your recommendation, and it does remove the need for the `bound_type` method, but there are some match patterns that become more complex. e.g.

```rust
match (self, other) {
    (ConstraintBound::Closed(other_bound), ConstraintBound::Open(open_bound))
    | (ConstraintBound::Open(open_bound), ConstraintBound::Closed(other_bound))
      if ... =>
```

would become


```rust
match (self, other) {
    (Self { kind: ConstraintBoundKind::Closed, bound_type: other_bound },
     Self { kind: ConstraintBoundKind::Open, bound_type: open_bound })
    | (Self { kind: ConstraintBoundKind::Open, bound_type: open_bound },
       Self { kind: ConstraintBoundKind::Closed, bound_type: other_bound })
      if ... =>
```

I don't think the removal of the `bound_type` method is worth the extra complexity in that kind of match.

crates/ty_python_semantic/src/types/constraints.rs

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: Sounds good, thanks for trying it!

dcreager: (it could also become

```rust
match (self, other) {
    (closed @ Self { kind: ConstraintBoundKind::Closed, .. },
     other @ Self { kind: ConstraintBoundKind::Open, .. })
    | (open @ Self { kind: ConstraintBoundKind::Open, .. },
       closed @ Self { kind: ConstraintBoundKind::Closed, .. })
      if ... =>
```

and then you'd use `open.bound_type` instead of `open_bound` in the body of the arm, but I still find that more complex than the single enum version)

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: I think debuggability is really important! I also think an internal team norm probably isn't sufficient here, though -- without a comment about why we're keeping the dead code, an external contributor coming across this code is probably going to wonder why we're leaving it around, and might even file a PR "cleaning it up"!

As an alternative to adding an `#[expect(dead_code)]` or `#[allow(dead_code)]`, what about adding a new "magic function" to `ty_extensions` that causes us to print a diagnostic with some of this `Display` information embedded in the diagnostic? That has two advantages:
- It means we directly test the `Display` implementation!
- It means you don't have to recompile ty if you want to change one of your `printf` or `dbg` calls -- changing a Python or markdown file you're invoking ty on doesn't require any recompilation!

I did something like this for debugging protocol interfaces recently, and I've found it incredibly useful: https://github.com/astral-sh/ruff/blob/b3c400528944c9e9a77dc41e60cad56b19742683/crates/ty_python_semantic/resources/mdtest/protocols.md?plain=1#L416-L429

dcreager: > what about adding a new "magic function"

Good idea, I'll give that a shot!

dcreager: I very much like where this is headed, but it is also looking like it will be very invasive on the current PR diff. So I'm going to pull that part out into a separate PR. For _this_ PR, I'll go with `expect` + a comment as per your first suggestion.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Added explanatory comment here and at the other `SmallVec` below

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: SGTM!

crates/ty_python_semantic/src/types/display.rs
dcreager: Updated per above

crates/ty_python_semantic/src/types/constraints.rs
dcreager: No, I meant in the doc comments. Updated this to be more clear

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Described this better in the doc comment. (Note that in this case the `Satisfiable` being returned is just `()` — I'm using this as a flag to indicate whether the updated-in-place result is never, always, or sometimes satisfiable.)

crates/ty_python_semantic/src/types/constraints.rs
dcreager: It's a bit difficult because Python doesn't have a direct way to construct open bounds (they primarily come up via negation), but I've tried my best. lmkwyt

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Done

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -12,21 +12,37 @@
...
-//! will go away.
+//! An individual constraint restricts the specialization of a single typevar to be within a
+//! particular lower and upper bound: the typevar can only specialize to a type that is a supertype
+//! of the lower bound, and a subtype of the upper bound. (Note that lower and upper bounds are
+//! fully static; we take the bottom and top materializations of the bounds to remove any gradual
+//! forms if needed.) Either bound can be "closed" (where the bound is a valid specialization), or
+//! "open" (where it is not).
+//!
+//! You can then build up more complex constraint sets using union, intersection, and negation
+//! operations. We use a distributed normal form (DNF) representation, just like we do for types: a
```
carljm: Isn't DNF "disjunctive normal form"? I haven't heard of "distributed normal form", and it seems like Google hasn't either.

```suggestion
//! operations. We use a disjunctive normal form (DNF) representation, just like we do for types: a
```

(Note: may want to fix this also in the PR description, which will become the commit message.)

crates/ty_python_semantic/src/types/constraints.rs
dcreager: This is a helper method for `simplify_clauses`, so I think the Python examples I added there cover this.  I did add a comment here directing the reader to `simplify_clauses` for more details.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: For this one, I'm having a hard time coming up with a Python example that would actually be illustrative. For the _types_ of the bounds, I could use Python to show that we use intersection to combine them. But just as important for this method is that it's determining whether the combined bound should be open or closed, and I'm wracking my brain trying to find easy Python snippets for lower bounds and open bounds.

AlexWaygood: no worries, we can leave it for now!

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: thank you!!

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: the Python examples are great, thank you!

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +208,953 @@ where
...
+        let mut existing_clauses = std::mem::take(&mut self.clauses).into_iter();
+        for existing in existing_clauses.by_ref() {
+            // Try to simplify the new clause against an existing clause.
+            match existing.simplify_clauses(db, clause) {
+                Simplifiable::NeverSatisfiable => {
+                    // If two clauses cancel out to 0, that does NOT cause the entire set to become
+                    // 0.  We need to keep whatever clauses have already been added to the result,
+                    // and also need to copy over any later clauses that we hadn't processed yet.
+                    self.clauses.extend(existing_clauses);
+                    return;
```
carljm, line 289: Correctness here depends on the invariant that a single new clause can only ever simplify-to-never with one existing clause (i.e. it can't cancel out two different existing clauses.) How do we know that to be the case here? Below with the `Simplified` case, in contrast, we explicitly handle the possibility that the new clause may simplify with a later clause.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +208,953 @@ where
...
+    fn always_satisfiable(_db: &'db dyn Db) -> Self {
+        Self::singleton(ConstraintClause::always())
+    }
+
+    fn is_never_satisfied(&self, _db: &'db dyn Db) -> bool {
+        self.clauses.is_empty()
+    }
+
+    fn is_always_satisfied(&self, _db: &'db dyn Db) -> bool {
+        self.clauses.len() == 1 && self.clauses[0].constraints.is_empty()
```
```suggestion
        self.clauses.len() == 1 && self.clauses[0].is_always()
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +208,953 @@ where
...
+        //
+        // class C2[V: int]:
+        //     # `T ≤ int`
+        //     x: A[V]
+        //
+        // class Saturation[U, V: int]:
+        //     # `1 ∪ (T ≤ int)`
+        //     # simplifies via saturation to
+        //     # `T ≤ int`
+        //     x: A[U] | A[V]
```
carljm, line 543: I could be missing something here: I can see how we can abstractly say that these constraints apply here, but concretely I don't think this code would ever result in us creating a `ConstraintSet` at all? There is no assignment here (where we'd create a `ConstraintSet` ephemerally in `has_relation_to_impl`, just in order to check if its `always_satisfiable`), nor is there a call to a generic function or constructor, where we'd create a `ConstraintSet` across multiple assignability checks (for each argument) and then solve it in order to generate a specialization.

I think to the extent that there is value in having Python examples (I'm not convinced that it's useful in code at this level of abstraction), it should ideally be examples where we would actually have to exercise the code in question in order to arrive at a correct type-checking answer in the Python example. I'm not quite seeing that in these examples; they are more like re-stating the set theory with a different syntax.

That said, I also don't think we should spend more time right now on improving these examples, so I'm fine leaving them as-is; this is more of a thought for future.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +208,953 @@ where
...
+        //     # simplifies via subsumption to
+        //     # `(T ≤ int ∩ U ≤ str)`
+        //     x: A[X1, Y1, Z2] | A[X2, Y2, Z2]
+        // ```
+        if self.subsumes_via_intersection(db, &other) {
+            return Simplifiable::Simplified(other);
+        }
+        if other.subsumes_via_intersection(db, &self) {
+            return Simplifiable::Simplified(self);
+        }
```
carljm, line 585: We do this via two merged iterations, but I think it can be easily done with a single iteration and a tri-valued return?

Maybe doesn't matter in practice, depends how hot this ends up being in practice, and how many multi-constraint clauses we see.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -182,3 +208,953 @@ where
...
+        typevar: BoundTypeVarInstance<'db>,
+        lower: ConstraintBound<'db>,
+        upper: ConstraintBound<'db>,
+    ) -> Satisfiable<Self> {
+        let lower_type = lower.bound_type();
+        let upper_type = upper.bound_type();
+        debug_assert_eq!(lower_type, lower_type.bottom_materialization(db));
+        debug_assert_eq!(upper_type, upper_type.top_materialization(db));
+
+        // If `lower > upper`, then the constraint cannot be satisfied, since there is no type that
```
carljm: Nit: I think you mean "lower is not a subtype of upper" here? That doesn't imply that upper is a subtype of lower (or `upper < lower`), since types are only partially ordered.

carljm (approved):
This is really clean and elegant!

Two thoughts, neither of them blocking this PR:

1. There will be a lot of `has_relation_to` checks where we collect constraints but never evaluate them for anything other than always-satisfied or never-satisfied. Will there be opportunities to improve performance on those checks if we know up-front that all we care about is always, or all we care about is never? It seems like it could potentially allow us to short-circuit a lot of work. (Something to explore in a future PR, not now.)
2.  My recursion spidey-sense tingles a bit about the fact that we use a `ConstraintSet` in evaluating `has_relation_to`, and building a `ConstraintSet` involves a lot of `is_subtype_of` checks on upper and lower bound types. Is there potential for stack overflow here? Do we need anything additional to prevent that? Can we get into a situation where evaluating a subtype relation causes us to build a constraint set that requires evaluating the original subtype relation? If so, our CycleDetector on `has_relation_to` wouldn't help, because it would be separate `is_subtype_of` checks.
3. Possibly related to (2): the spec says typevar bounds/constraints cannot be generic, but there's been recent discussion of lifting that requirement, and it sounds like Pyrefly will experiment with that. It seems to me that we're well-positioned for that as well (you'd just end up adding constraints on the nested typevar, too), but maybe something to consider.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Haha whoops yes it should be "disjunctive" not "distributed" :grimacing:

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Any existing clauses must already be simplified relative to each other. So I think that for a new clause to cancel out more than one existing clause, it would have to do it in multiple steps, in a confluent way. So the new clause would "partially" simplify against the first existing clause that we encounter (i.e. simplify a bit but not all the way to 0). (That would trigger the `Simplified` branch below, where we carry the simplified result over to check against later existing clauses.) Then that partially simplified clause would simplify the "rest of the way" to 0 when we encounter the second (relevant) existing clause. And the "confluent" part means that it would need to happen regardless of the order that the two existing clauses appear in the original result.

I have not done a proof that :point_up: holds, but that's my intuition for why it should™ work.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Added a TODO to consider this

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Yes, good catch! Done

dcreager (commented):
> There will be a lot of `has_relation_to` checks where we collect constraints but never evaluate them for anything other than always-satisfied or never-satisfied. Will there be opportunities to improve performance on those checks if we know up-front that all we care about is always, or all we care about is never? It seems like it could potentially allow us to short-circuit a lot of work. (Something to explore in a future PR, not now.)

I don't think this would give correct results. This is related to my comment from last week https://github.com/astral-sh/ruff/pull/19838#discussion_r2289637481, and you can see it in the draft of #20093. In that PR I've moved around the non-inferrable typevar match arms in `has_relation_to`, because we no longer have to be careful about doing some typevar checks before we handle the connectives, and others after. We can rely on how we combine the constraints from the recursive calls to let partially satisfiable recursive constraint sets either (a) "build up" towards 1, or (b) "cancel out" towards 0. Doing so requires having the full constraint sets available, so that we can look at their structure to see what they do when unioned or intersected together. Doing that on `bool` loses that detail, leading to wrong answers.

> My recursion spidey-sense tingles a bit about the fact that we use a `ConstraintSet` in evaluating `has_relation_to`, and building a `ConstraintSet` involves a lot of `is_subtype_of` checks on upper and lower bound types. Is there potential for stack overflow here? Do we need anything additional to prevent that? Can we get into a situation where evaluating a subtype relation causes us to build a constraint set that requires evaluating the original subtype relation? If so, our CycleDetector on `has_relation_to` wouldn't help, because it would be separate `is_subtype_of` checks.

If the bounds of a constraint don't contain any typevars (a "concrete" type), then I think we're okay, since calculating subtyping of two concrete types can only produce `true`, `false`, and combinations of those. (If there are no typevars in the type, then there's nothing to create an `AtomicConstraint` for.) And so we never hit any of the new logic for combining and simplifying constraints.

If there are bounds that _do_ contain typevars, we do have to worry about this — and the way POPL15 etc solve this is by introducing an ordering on typevars, and saying that typevar bounds can only reference other typevars that are smaller according to that ordering. That ensures that you don't get cycles in the "bounds graph". I figure we'll just use Salsa IDs as our ordering when we get to that part.

> Possibly related to (2): the spec says typevar bounds/constraints cannot be generic, but there's been recent discussion of lifting that requirement, and it sounds like Pyrefly will experiment with that. It seems to me that we're well-positioned for that as well (you'd just end up adding constraints on the nested typevar, too), but maybe something to consider.

I think we will already have to support typevars that have _constraints_ involving other typevars, to handle things like calling a generic function (and inferring its specialization) from inside another (such that the constraints of the calling function are needed to figure out the valid specializations of the called function). So at that point it should be no problem to have typevar _bounds_ mention other typevars, since that would just translate into a constraint that can already contain other typevars. (Modulo the bit above about using an artificial ordering to keep the bounds graph acyclic.)

PR #19999, opened 2025-08-20
[ty] Avoid unnecessary argument type expansion

dhruvmanila:
## Summary

Part of: https://github.com/astral-sh/ty/issues/868

This PR adds a heuristic to avoid argument type expansion if it's going to eventually lead to no matching overload.

This is done by checking whether the non-expandable argument types are assignable to the corresponding annotated parameter type. If one of them is not assignable to all of the remaining overloads, then argument type expansion isn't going to help.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/overloads.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/overloads.md
@@ -620,6 +620,81 @@ def _(ab: A | B, ac: A | C, cd: C | D):
     reveal_type(f(*(cd,)))  # revealed: Unknown
 ```
 
+### Performance
+
+Argument type expansion could lead to exponential growth of the number of argument lists that needs
+to be evaluated, so ty deploys some heuristics to prevent this from happening.
+
+#### Avoid argument type expansion
+
+Heuristic: If an argument type that cannot be expanded and cannot be assighned to any of the
+remaining overloads before argument type expansion, then even with argument type expansion, it won't
+lead to a successful evaluation of the call.
+
+`overloaded.pyi`:
+
+```pyi
+from typing import overload
+
+class A: ...
+class B: ...
+class C: ...
+
+@overload
+def f() -> None: ...
+@overload
+def f(**kwargs: int) -> C: ...
+@overload
+def f(x: A, /, **kwargs: int) -> A: ...
+@overload
+def f(x: B, /, **kwargs: int) -> B: ...
+```
+
+```py
+from overloaded import A, B, C, f
+
+def _(a=1):
+    reveal_type(f(a1=a, a2=a, a3=a))  # revealed: C
+    reveal_type(f(A(), a1=a, a2=a, a3=a))  # revealed: A
+    reveal_type(f(B(), a1=a, a2=a, a3=a))  # revealed: B
+
+    # error: [no-matching-overload]
+    # revealed: Unknown
+    reveal_type(f(
+        C(),
+        a1=a,
+        a2=a,
+        a3=a,
+        a4=a,
+        a5=a,
+        a6=a,
+        a7=a,
+        a8=a,
+        a9=a,
+        a10=a,
+        a11=a,
+        a12=a,
+        a13=a,
+        a14=a,
+        a15=a,
+        a16=a,
+        a17=a,
+        a18=a,
+        a19=a,
+        a20=a,
+        a21=a,
+        a22=a,
+        a23=a,
+        a24=a,
+        a25=a,
+        a26=a,
+        a27=a,
+        a28=a,
+        a29=a,
+        a30=a,
+    ))
+```
+
 ## Filtering based on `Any` / `Unknown`
 
 This is the step 5 of the overload call evaluation algorithm which specifies that:
--- a/crates/ty_python_semantic/src/types/call/arguments.rs
+++ b/crates/ty_python_semantic/src/types/call/arguments.rs
@@ -5,7 +5,7 @@ use ruff_python_ast as ast;
 
 use crate::Db;
 use crate::types::KnownClass;
-use crate::types::enums::enum_member_literals;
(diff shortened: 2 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/call/overloads.md
```diff
@@ -620,6 +620,127 @@ def _(ab: A | B, ac: A | C, cd: C | D):
...
+    # Here, the arity check filters out the first and second overload, type checking fails on the
+    # remaining overloads, so ty moves on to argument type expansion. But, the first argument (`C`)
+    # isn't assignable to any of the remaining overloads (3 and 4), so there's no point in expanding
+    # the other 30 arguments of type `Unknown | Literal[1]` which would result in allocating a
+    # vector containing 2**30 argument lists after expanding all of the arguments.
+    reveal_type(
+        # error: [no-matching-overload]
+        # revealed: Unknown
+        f(
+            C(),
```
dhruvmanila, line 679: Considering the same example, but pass in `A()` as the first argument and use `a=None` instead of `a=1` for the parameter above, this would still continue with the expansion. The reason being that `A()` matches parameter of at least one overload which means we cannot skip argument type expansion.

I'm trying to think through what kind of heuristic to have that would avoid the expansion in this case as well. This would still lead to no-matching-overload because even though `Unknown` is assignable to `int`, the spec says that _all_ argument lists resulting from an expansion should evaluate successfully and there's no such expansion as expanding `Unknown | None` leads to argument lists where at least one would have `None` which isn't assignable to `int`.

Regardless, I think it'd be best to set a higher limit on the expansion.

dhruvmanila:
EDIT: Nevermind, it _is_ because ty doesn't support `**kwargs` and the new logic should avoid checking those arguments to skip argument type expansion for now.

> ```diff
> rotki (https://github.com/rotki/rotki)
> + rotkehlchen/tests/db/test_db.py:377:13: error[no-matching-overload] No overload of bound method `set_dynamic_cache` matches arguments
> + rotkehlchen/tests/db/test_db.py:378:20: error[no-matching-overload] No overload of bound method `get_dynamic_cache` matches arguments
> + rotkehlchen/tests/db/test_db.py:418:20: error[no-matching-overload] No overload of bound method `get_dynamic_cache` matches arguments
> - Found 1588 diagnostics
> + Found 1591 diagnostics
> ```

I wasn't expecting any mypy-primer diff but it seems that we have new diagnostics :)

I think this might be due to the fact that ty doesn't support `**kwargs` yet but I'm looking into it a bit closely. The main reason I think that is because when I remove the `**kwargs` in the call, the diagnostic disappear.

https://github.com/user-attachments/assets/6cecc561-3252-468d-8d4b-551d7f537268

crates/ty_python_semantic/src/types/call/arguments.rs
```diff
@@ -208,6 +208,27 @@ impl<'a, 'db> FromIterator<(Argument<'a>, Option<Type<'db>>)> for CallArguments<
...
+                        .all_elements()
+                        .any(|element| is_type_expandable(db, *element)),
+                    Tuple::Variable(_) => false,
+                })
+                || enum_metadata(db, class.class_literal(db).0).is_some()
+        }
+        Type::Union(_) => true,
+        _ => false,
+    }
+}
```
MichaReiser, line 230: Should we add a note to `expand_type` to update `is_type_expandable` (maybe rename to `is_expandable_type` for symmetry)  if necessary

crates/ty_python_semantic/resources/mdtest/call/overloads.md
carljm: Yes, I think there will be cases where we can't avoid actually trying all the expansions and see if any of them work -- best we can do there is optimize the expansions as best we can (e.g. iterator instead of eagerly materialized?) and then set a reasonable limit.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1327,6 +1328,48 @@ impl<'db> CallableBinding<'db> {
         // for evaluating the expanded argument lists.
         snapshotter.restore(self, pre_evaluation_snapshot);
 
+        // At this point, there's at least one argument that can be expanded.
```
carljm, line 1331: Would it make sense to first explicitly check if there are any expandable arguments (using the new method you added), then do this check, then actually expand the arguments? This way even if the heuristic applies, we've already generated the expansion, just haven't used the expansions yet.

But maybe generating the expansions is very cheap relative to actually trying the call with new argument types, so this doesn't matter?

And I guess we'd slow down the fast path slightly if we first check all arguments for expandability, then separately expand them?

dhruvmanila: Yeah, I don't think this would really matter in practice mainly because there are only a few differences between `expand_type` and `is_type_expandable` which are (1) collecting the enum members from the metadata (both method generates the metadata) (2) performing multi-cartesian product of tuple elements and (3) allocation. I guess it wouldn't hurt to avoid allocating when it's easy to do so. Let me try it, I think I'll make this change.

dhruvmanila: Ah right, so the main reason I did it at this location is so that the bindings state is the one before the type checking step, so that we only skip the overloads that have been filtered by the arity check.

So, we'd either have to pay the cost of either
- Allocation for the first expansion, taking the bindings snapshot. We'd skip the snapshot if there are no expansion in this case.
- Always take the binding snapshot, check whether expansion needs to happen using the logic in this PR, skip allocating for the first expansion if there's no need

I think going with the latter sounds reasonable i.e., instead of always allocating the first expansion, we'd always allocate the bindings snapshot.

dhruvmanila: Hmm, but then there's complication on restoring the snapshots at the correct places. I think I'll leave this as is for now.

PR #20002, opened 2025-08-20
[ty] correctly ignore field specifiers when not specified

leandrobbraga:
This commit corrects the type checker's behavior when handling `dataclass_transform` decorators that don't explicitly specify `field_specifiers`. According to [PEP 681 (Data Class Transforms)](https://peps.python.org/pep-0681/#dataclass-transform-parameters), when `field_specifiers` is not provided, it defaults to an empty tuple, meaning no field specifiers are supported and `dataclasses.field`/`dataclasses.Field` calls should be ignored.

Fixes https://github.com/astral-sh/ty/issues/980


carljm:
Going to go ahead and merge this, it's a small PR, there's nothing here that we can't easily change again.

Thank you for the fix!

sharkdp:
> @sharkdp are you okay with this temporary fix?

Yes, of course. Thank you for this contribution!

PR #20007, opened 2025-08-20
`Option::unwrap` is now const

ntBre:
Summary
--

I noticed while working on #20006 that we had a custom `unwrap` function for `Option`. This has been const on stable since 1.83 ([docs](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap), [release notes](https://blog.rust-lang.org/2024/11/28/Rust-1.83.0/)), so I think it's safe to use now. I grepped a bit for related todos and found this one for `AsciiCharSet` but no others.

Test Plan
--

Existing tests


MichaReiser (approved):
Nice

PR #20008, opened 2025-08-20
[ty] add docstrings to completions based on type

Gankra:
This is a fairly simple but effective way to add docstrings to like 95% of completions from initial experimentation.

Fixes https://github.com/astral-sh/ty/issues/1036

Although ironically this approach *does not* work specifically for `print` and I haven't looked into why.

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -8,9 +8,11 @@ use ruff_text_size::{Ranged, TextRange, TextSize};
 use ty_python_semantic::{Completion, NameKind, SemanticModel};
 
 use crate::Db;
+use crate::docstring::Docstring;
 use crate::find_node::covering_node;
+use crate::goto::DefinitionsOrTargets;
 
-pub fn completion(db: &dyn Db, file: File, offset: TextSize) -> Vec<Completion<'_>> {
+pub fn completion(db: &dyn Db, file: File, offset: TextSize) -> Vec<DetailedCompletion<'_>> {
     let parsed = parsed_module(db, file).load(db);
 
     let Some(target_token) = CompletionTargetTokens::find(&parsed, offset) else {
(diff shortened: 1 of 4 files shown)


Gankra:
https://github.com/user-attachments/assets/d335e367-5114-45cf-8732-627f7d783934

crates/ty_wasm/src/lib.rs
```diff
@@ -421,7 +421,7 @@ impl Workspace {
             .into_iter()
             .map(|completion| Completion {
                 kind: completion.kind(&self.db).map(CompletionKind::from),
-                name: completion.name.into(),
+                name: completion.inner.name.into(),
```
MichaReiser, line 424: It would be nice to also hook up documentation in the playground. You have to add a new `documentation` field to `Completion`, then propagate the field here

https://github.com/astral-sh/ruff/blob/8c0743df97dc7afa1deab7506b322551f2e392bb/playground/ty/src/Editor/Editor.tsx#L321

crates/ty_ide/src/completion.rs
```diff
@@ -40,6 +42,27 @@ pub fn completion(db: &dyn Db, file: File, offset: TextSize) -> Vec<Completion<'
...
+                inner: completion,
+                documentation,
+            }
+        })
+        .collect()
+}
+
+pub struct DetailedCompletion<'db> {
+    pub inner: Completion<'db>,
+    pub documentation: Option<Docstring>,
```
MichaReiser, line 59: Should we make this a field on `Completion` instead that defaults to `None`: Given that `completion` is the only method constructing the completions

MichaReiser (approved):
Nice

crates/ty_ide/src/completion.rs
Gankra: Currently that doesn't work out because Docstring is defined in ty_ide and Completion is defined in ty_python_semantic. The crate boundaries have gotten quite wobbly 😅

crates/ty_wasm/src/lib.rs
Gankra: I think done? Never tested the playground.

PR #20009, opened 2025-08-20
[`flake8-use-pathlib`] Add autofix for `PTH211`

chirizxc:
## Test Plan


crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_symlink.rs
```diff
@@ -0,0 +1,143 @@
...
+        let src_code = locator.slice(src.range());
+        let dst_code = locator.slice(dst.range());
+
+        let target_is_directory = call
+            .arguments
+            .find_argument_value("target_is_directory", 2)
+            .and_then(|expr| {
+                let code = locator.slice(expr.range());
+                (!code.eq("False")).then(|| format!(", target_is_directory={code}"))
+            })
```
dylwil3: Can you add a test where there is some whitespace around `False`?

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_symlink.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    // `dir_fd` is not supported by pathlib, so check if there are non-default values.
+    // Signature as of Python 3.13 (https://docs.python.org/3/library/os.html#os.symlink)
+    // ```text
+    //            0    1    2                             3
+    // os.symlink(src, dst, target_is_directory=False, *, dir_fd=None)
+    // ```
+    if is_keyword_only_argument_non_default(&call.arguments, "dir_fd") {
+        return;
+    }
```
dylwil3, line 81: It looks like we don't offer a fix when `dir_fd = None`, though - was that intentional? (It's probably okay not to)

chirizxc: In other autofixes, we have the same `if is_keyword_only_argument_non_default(&call.arguments, <...>)`

chirizxc: [os_rmdir.rs](https://github.com/astral-sh/ruff/blob/main/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_rmdir.rs), [os_rmdir.rs](https://github.com/astral-sh/ruff/blob/main/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_rmdir.rs) and other files

chirizxc:
done

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_symlink.rs
dylwil3: It looks like you added a test with whitespace around `True`. I was hoping for a test like this:

```python
os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory=     False    )
```

to see whether your expression range that you're slicing includes the whitespace or not. 

In fact: I think it'd probably be better to just replace this code snippet with something like:

```rust
            .and_then(|expr| {
                let code = locator.slice(expr.range());
                expr.as_boolean_literal_expr()
                    .is_some_and(|bl| bl.value == true)
                    .then(|| format!(", target_is_directory={code}"))
            })
```

chirizxc: изображение
with these spaces also works without replacing it with this code

chirizxc: изображение
In this case, there are also invalid fixes for the variant with `target_is_directory=True`

dylwil3:
Adding a minor edit - it looks like you can pass an arbitrary type to `target_is_directory` without CPython throwing an error. (I think it evaluates the truthiness of the value). I've skipped the fix in these cases instead of just throwing out the value, since otherwise I think we may change program behavior. If CI passes I'll merge

chirizxc:
For some reason, a lot of `Safe fixes` are missing from the snapshots

chirizxc:
изображение
i mean this

dylwil3:
> For some reason, a lot of `Safe fixes` are missing from the snapshots

I swear I checked the ecosystem results and everything... maybe they were not updated even when they said they were? Oy. Thank you for the quick followup PR!

PR #20013, opened 2025-08-20
[ty] fix GotoTargets for keyword args in nested function calls

Gankra:
While implementing similar logic for initializers I noticed that this code appeared to be walking the ancestors in the wrong direction, and so if you have nested function calls it would always grab the outermost one instead of the closest-ancestor.

The four copies of the test are because there's something really evil in our caching that can't seem to be demonstrated in our cursor testing framework, which I'm filing a followup for.

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -599,7 +599,6 @@ impl GotoTarget<'_> {
                     // Find the containing call expression from the ancestor chain
                     let call_expression = covering_node
                         .ancestors()
-                        .rev()
                         .find_map(ruff_python_ast::AnyNodeRef::expr_call)?;
                     Some(GotoTarget::KeywordArgument {
                         keyword,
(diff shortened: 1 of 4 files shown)


crates/ty_ide/src/goto.rs
```diff
@@ -599,6 +599,7 @@ impl GotoTarget<'_> {
                     // Find the containing call expression from the ancestor chain
                     let call_expression = covering_node
                         .ancestors()
+                        .rev()
```
MichaReiser: I think we should fix this in `ancestors` instead. At least one of the remaining references are equally wrong and we can call `rev` for the remaining usages where they want to walk the ancestors top to bottom (but bottom up is more intuitive to me)

MichaReiser (approved):

crates/ty_ide/src/goto.rs
Gankra: Yeah good call, it does seem like a footgun

PR #20014, opened 2025-08-21
[ty] Make initializer calls GotoTargets

Gankra:
This introduces `GotoTarget::Call` that represents the kind of ambiguous/overloaded click of a callable-being-called:

```py
x = mymodule.MyClass(1, 2)
             ^^^^^^^
```

This is equivalent to `GotoTarget::Expression` for the same span but enriched
with information about the actual callable implementation.

That is, if you click on `MyClass` in `MyClass()` it is *both* a
reference to the class and to the initializer of the class. Therefore
it would be ideal for goto-* and docstrings to be some intelligent
merging of both the class and the initializer.

In particular the callable-implementation (initializer) is prioritized over the callable-itself (class) so when showing docstrings we will preferentially show the docs of the initializer if it exists, and then fallback to the docs of the class.

For goto-definition/goto-declaration we will yield both the class and the initializer, requiring you to pick which you want (this is perhaps needlessly pedantic but...).

Fixes https://github.com/astral-sh/ty/issues/898
Fixes https://github.com/astral-sh/ty/issues/1010

### DIFF
--- a/crates/ty_ide/src/find_node.rs
+++ b/crates/ty_ide/src/find_node.rs
@@ -56,7 +56,6 @@ pub(crate) fn covering_node(root: AnyNodeRef, range: TextRange) -> CoveringNode 
 }
 
 /// The node with a minimal range that fully contains the search range.
-#[derive(Clone)]
 pub(crate) struct CoveringNode<'a> {
     /// The covering node, along with all of its ancestors up to the
     /// root. The root is always the first element and the covering
--- a/crates/ty_ide/src/find_node.rs
+++ b/crates/ty_ide/src/find_node.rs
@@ -56,6 +56,7 @@ pub(crate) fn covering_node(root: AnyNodeRef, range: TextRange) -> CoveringNode 
 }
 
 /// The node with a minimal range that fully contains the search range.
+#[derive(Clone)]
 pub(crate) struct CoveringNode<'a> {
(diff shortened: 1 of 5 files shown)


MichaReiser (approved):

MichaReiser:
> For goto-definition/goto-declaration we will yield both the class and the initializer, requiring you to pick which you want (this is perhaps needlessly pedantic but...).

Pyright jumps to the definition of the class. I'm leaning towards aligning the behavior for now (although it is somewhat unexpected, so maybe what you have is best)

Gankra:
Something I've vaguely been considering is another variant of DefinitionsOrTargets that has two lists of definitions: one is the "right" answer and one is "fallback". So docstrings would consider both but goto-def would be opinionated and just pick one?

Gankra:
Surprisingly a lot of reports of different IDEs actually making the `(` of the call separately clickable as the ctor call!

MichaReiser:
> Surprisingly a lot of reports of different IDEs actually making the `(` of the call separately clickable as the ctor call!

This sounds nice

Gankra:
Ok I've rebased this on top of #20148 and minimized its intelligence, making this a reasonably safe change. I'm going to land this now (feedback is the best way to iterate on exact UX of this I think).

PR #19769, opened 2025-08-05
[`pyupgrade`] Avoid reporting `__future__` features as unnecessary when they are used (`UP010`)

IDrokin117:
## Summary
Resolves #19561

Fixes the [unnecessary-future-import (UP010)](https://docs.astral.sh/ruff/rules/unnecessary-future-import/) rule to correctly identify when imported __future__ modules are actually used in the code, preventing false positives.

I assume there is no way to check usage in `analyze::statements`, because we don't have any usage bindings for imports. To determine unused imports, we have to fully scan the file to create bindings and then check usage, similar to [unused-import (F401)](https://docs.astral.sh/ruff/rules/unused-import/#unused-import-f401). So, `Rule::UnnecessaryFutureImport` was moved from the `analyze::statements` to the `analyze::deferred_scopes` stage. This caused the need to change the logic of future import handling to a bindings-based approach.

Also, the diagnostic report was changed.
Before
```
  |
1 | from __future__ import nested_scopes, generators
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP010
```
after
```
  |
1 | from __future__ import nested_scopes, generators
  |                        ^^^^^^^^^^^^^ UP010
```

I believe this is the correct way, because `generators` may be used, but `nested_scopes` is not.

### Special case
I've found out about some specific case.
```python
from __future__ import nested_scopes

nested_scopes = 1
```
Here we can treat `nested_scopes` as an unused import because the variable `nested_scopes` shadows it and we can safely remove the future import (my fix does it).

But [F401](https://docs.astral.sh/ruff/rules/unused-import/#unused-import-f401) not triggered for such case ([sandbox](https://play.ruff.rs/296d9c7e-0f02-4659-b0c0-78cc21f3de76))
```
from foo import print_function

print_function = 1
```
In my mind, `print_function` here is an unused import and should be deleted (my IDE highlight it). What do you think?

## Test Plan

Added test cases and snapshots:
- Split test file into separate _0 and _1 files for appropriate checks.
- Added test cases to verify fixes when future module are used.

PR #20020, opened 2025-08-21
[`pyupgrade`] Gated behind a preview of separate diagnostics (`UP010`)

IDrokin117:
## Summary

Second part of https://github.com/astral-sh/ruff/pull/19769

Separate diagnostic generation for each unused import and range updating accordingly. Gated behind a preview

## Test Plan

Added preview snapshots


IDrokin117:
@ntBre Could you please review PR?

ntBre:
Yes, sorry for the delay! We've been preparing for the 0.13 release this past week. I'll be catching up on reviews once that goes out.

MichaReiser:
What's the motivation for splitting the diagnostics? This seems unnecessarily verbose to me. Could we instead use the new sub-diagnostics to highlight the individual ranges?

ntBre:
What I liked about the split diagnostics in https://github.com/astral-sh/ruff/pull/19769#pullrequestreview-3111691918 was mostly just underlining the affected imports instead of the whole import statement, which I think could also be handled by sub-diagnostics.

As one supporting example, we also emit one diagnostic per import for [unused-import (F401)](https://docs.astral.sh/ruff/rules/unused-import/#unused-import-f401). Although we could revisit that as well now that we have sub-diagnostics.

I don't feel too strongly either way, though.

[playground](https://play.ruff.rs/bf87eac4-7c5d-45f5-8183-6b1830a84ed1)

IDrokin117:
@ntBre What should I do here? It seems diagnostics work per import as for F401, doesn't it? Should I use sub-diagnostics instead? What piece of code can I take as a "sub-diagnostic" example?

ntBre:
I think we just need to reach agreement on how many diagnostics to emit. If you want to try sub-diagnostics to see how they look, here is a previous comment I used on another PR:

The easiest way is  with `secondary_annotation` on the guard returned by `report_diagnostic`:

<hr>

That's not technically a sub-diagnostic, but we've only added a nice helper for secondary annotations rather than `SubDiagnostic`s proper. The helper code would be quite similar to the annotation version, but I'm happy to help with that if we go that route.

IDrokin117:
@ntBre @MichaReiser I've rewritten the diagnostic using secondary_annotation. Please check it and let me know how it looks now.  
I've also handled the edge case when there is only one alias in the import statement, so there is no need to underline that specific alias separately.

crates/ruff_linter/src/rules/pyupgrade/snapshots/ruff_linter__rules__pyupgrade__tests__UP010_0.py__preview.snap
```diff
@@ -0,0 +1,208 @@
+---
+source: crates/ruff_linter/src/rules/pyupgrade/mod.rs
+---
+UP010 [*] Unnecessary `__future__` imports `generators`, `nested_scopes` for target Python version
+ --> UP010_0.py:1:1
+  |
+1 | from __future__ import nested_scopes, generators
+  | ^^^^^^^^^^^^^^^^^^^^^^^-------------^^----------
```
ntBre, line 8: I think if we do go this route, we should narrow the primary diagnostic range to just `__future__` (or maybe `from __future__` or `from __future__ import`) to keep the two ranges from overlapping:


```suggestion
1 | from __future__ import nested_scopes, generators
  |      ^^^^^^^^^^        -------------  ----------
```

I think I'd still prefer either one diagnostic per import, or just leaving the diagnostic as it is on stable, over this, though. But I'm curious to hear what Micha thinks. Thank you for exploring this!

MichaReiser: I'd have to double-check but narrowing the range changes the places where suppression comments are allowed. 

E.g, the following currently works but I suspect won't work if we narrow the range

```
from __future__ import (
	nested_scopes, # noqa <RULE>
 	generators
)
```

MichaReiser: Hmm yeah, using multiple annotated ranges and highlighting the entire import statement doesn't help improve readability.

I only found this usage in cargo:

```
warning: methods `unused1` and `unused2` are never used
  --> src/function/maybe_changed_after.rs:41:19
   |
32 | impl VerifyResult {
   | ----------------- methods in this implementation
...
41 |     pub(crate) fn unused1(&self) {}
   |                   ^^^^^^^
42 |
43 |     pub(crate) fn unused2(&self) {}
   |                   ^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

They use the `unused1` as the primary range and mark the `impl` block as a secondary annotation. But doing the same in Ruff requires figuring out how to make our suppression system understand that this diagnostic represents two separate violations (the current assumption is that diagnostics only represent a single error.

ntBre: Ah, would narrowing the noqa range actually be a benefit of separate diagnostics here? Something like this doesn't currently [work](https://play.ruff.rs/1fa9e588-8306-4480-80d9-f720350e0506):

```py
from __future__ import ( 
    nested_scopes,  # noqa: UP010
    generators,
) 

generators
```

You have to put the noqa [here](https://play.ruff.rs/1d712d9e-8afa-4082-8872-c70379080045):

```py
from __future__ import ( # noqa: UP010
    nested_scopes,  
    generators,
) 

generators
```

at least that's the only place I found that works.

But yeah, I think secondary annotations are probably better for highlighting additional context that helps to explain the main diagnostic but likely falls outside the normal snippet context window, like in your Rust example.

IDrokin117: It appears that the previous implementation, which uses separate diagnostics, functions well for `# noqa: UP010` per import object. Should I revert to that approach?

MichaReiser: I'm leaning towards leaving it as is. I don't think that any of the tried approaches greatly improve the rendered diagnostics to justify a breaking change. But maybe Brent thinks differently? I'd otherwise close this PR and suggest that we revisit the rendering once we have figured out how to better handle those cases with our diagnostic system. But thank you a lot for trying out all those different approaches. It was very helpful to drive the decision-making and show where there's some need to improve our diagnostic system.

ntBre: Yeah, I think I agree with Micha. Thank you for working on this, I also agree that it was very helpful to see the different options!

PR #20021, opened 2025-08-21
[ty] Fix incorrect docstring in call signature completion

MichaReiser:
## Summary

This PR fixes https://github.com/astral-sh/ty/issues/1071

The core issue is that `CallableType` is a salsa interned but `Signature` (which `CallableType` stores) ignores the `Definition` in its `Eq` and `Hash` implementation.

This PR tries to simplest fix by removing the custom `Eq` and `Hash` implementation. The main downside of this fix is that it can increase memory usage because `CallableType`s that are equal except for their `Definition` are now interned separately.

The alternative is to remove `Definition` from `CallableType` and instead, call `bindings` directly on the callee (call_expression.func). However, this would require 
addressing the TODO 

here https://github.com/astral-sh/ruff/blob/39ee71c2a57bd6c51482430ba8bfe3728b4ca443/crates/ty_python_semantic/src/types.rs#L4582-L4586 

This might probably be worth addressing anyway, but is the more involved fix. That's why I opted for removing the custom `Eq` implementation.

We already "ignore" the definition during normalization, thank's to Alex's work in https://github.com/astral-sh/ruff/pull/19615

## Test Plan

https://github.com/user-attachments/assets/248d1cb1-12fd-4441-adab-b7e0866d23eb

### DIFF
--- a/crates/ty_python_semantic/src/types/signatures.rs
+++ b/crates/ty_python_semantic/src/types/signatures.rs
@@ -234,7 +234,7 @@ impl<'db> VarianceInferable<'db> for &CallableSignature<'db> {
 }
 
 /// The signature of one of the overloads of a callable.
-#[derive(Clone, Debug, salsa::Update, get_size2::GetSize)]
+#[derive(Clone, Debug, salsa::Update, get_size2::GetSize, PartialEq, Eq, Hash)]
 pub struct Signature<'db> {
     /// The generic context for this overload, if it is generic.
     pub(crate) generic_context: Option<GenericContext<'db>>,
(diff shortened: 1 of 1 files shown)


MichaReiser:
I'll give @carljm a chance to chime in, as he reasoned that we should remove `Definition` from `CallableType`.

carljm:
> I'll give @carljm a chance to chime in, as he reasoned that we should remove `Definition` from `CallableType`.

I do think that's what we should do eventually, but it doesn't have to be done right now -- this fix improves the behavior in the short term and I don't think makes anything more difficult for removing the definition in future, so I'm in favor of it.

(Also it removes code, and all else equal I'm always in favor of that 😆 )

Gankra:
I'm gonna land this since it's a important/pervasive bugfix, makes it easier to test/debug other function handling features.

PR #20022, opened 2025-08-21
[ty] Improve diagnostics for bad calls to functions

AlexWaygood:
## Summary

It was annoyingly hard to debug https://github.com/astral-sh/ty/issues/1076. It wasn't clear whether ty was resolving the type definition to a location in pandas or pandas-stubs; and it wasn't clear what ty thought the signature of the function _was_ (the diagnostic only said that the user had got the signature wrong).

This PR adds subdiagnostics to our `missing-argument`, `too-many-positional-arguments` and `unknown-argument` error codes, so that these will be easier to debug in the future -- both for users, and for us!

I've taken care not to add the subdiagnostics if it's a union of callables being called. We currently emit a separate diagnostic for each element of the union; having a sub-diagnostic for each element would probably be too verbose for it to be worth it, in my opinion.

## Test Plan

Snapshots.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/diagnostics/missing_argument.md
+++ b/crates/ty_python_semantic/resources/mdtest/diagnostics/missing_argument.md
@@ -1,0 +1,37 @@
+# "Missing argument" diagnostics
+
+<!-- snapshot-diagnostics -->
+
+If a non-union callable is called with a required parameter missing, we add a subdiagnostic showing
+where the parameter was defined. We don't do this for unions as we currently emit a separate
+diagnostic for each element of the union; having a sub-diagnostic for each element would probably be
+too verbose for it to be worth it.
+
+`module.py`:
+
+```py
+def f(a, b=42): ...
+def g(a, b): ...
+
+class Foo:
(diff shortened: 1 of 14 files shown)


AlexWaygood:
Not sure why GitHub is repeatedly failing to checkout this branch on Windows, but it doesn't seem related to the changes being made here

AlexWaygood:
oh I put a Bad Windows Character in a generated snapshot path, didn't I?

carljm (approved):
Nice!

Not sure I agree about unions -- in every case shown in these snapshot tests (and I think probably in the majority of real-world cases, too), there's still just one diagnostic because only one element of the union failed the call, and the union context is just an additional info line or two. In all of those cases, I think the extra subdiagnostic (regarding the one union element that failed the call) would be an improvement and useful in debugging, it wouldn't be too much info. It's possible that there are cases where many union elements fail the call; those might be cases where it makes sense to skip the sub-diagnostic?

But it looks like that would be an additional feature here, not just removing a restriction; I don't think we need to spend more time on it now.

PR #20023, opened 2025-08-21
[ty] Add precise tuple specs when iterating over other kinds of types

AlexWaygood:
Curious if this has any ecosystem impact


AlexWaygood:
> ```diff
> - scipy/linalg/tests/test_lapack.py:171:41: warning[possibly-unresolved-reference] Name `ref` used when possibly not defined
> ```

Okay, this one's pretty cool -- ty is now able to infer that all possibilities have been exhaustively covered here, meaning the variable cannot possibly be undefined by the time the code reaches the assertion at the bottom: https://github.com/scipy/scipy/blob/99c0ef6af161a4d8157cae5276a20c30b7677c6f/scipy/linalg/tests/test_lapack.py#L147-L171

carljm:
> ty is now able to infer that all possibilities have been exhaustively covered here

This is a good example of the sort of issue that led us to disable `possibly-unresolved-reference` by default. Maybe at some point we'll be able to reconsider that!

PR #20024, opened 2025-08-21
[`flake8-bugbear`] Include certain guaranteed-mutable expressions: tuples, generators, and assignment expressions (`B006`)

### LINKED ISSUE #20004: B006 doesn’t detect mutable defaults in tuples, generators, and assignment expressions
### Summary

[mutable-argument-default (B006)](https://docs.astral.sh/ruff/rules/mutable-argument-default/) doesn’t detect certain guaranteed-mutable expressions which should be simple to support: tuples containing mutable expressions, generator expressions, and assignment expressions containing mutable expressions. [Example](https://play.ruff.rs/280e1f9b-99c8-4c79-8d27-0e9bbff72f34):
```console
$ cat >b006.py <<'# EOF'
def f(x=([],), y=(y for y in "y"), z=(z := [])):
    pass
# EOF

$ ruff --isolated check b006.py --select B006
All checks passed!
```

### Version

IDrokin117:
## Summary
Resolves #20004

The implementation now supports guaranteed-mutable expressions in the following cases:
- Tuple literals with mutable elements (supporting deep nesting)
- Generator expressions
- Named expressions (walrus operator) containing mutable components

Preserves original formatting for assignment value:

```python
# Test case
def f5(x=([1, ])):
    print(x)
```
```python
# Fix before
def f5(x=(None)):
    if x is None:
        x = [1]
    print(x)
```
```python
# Fix after 
def f5(x=None):
    if x is None:
        x = ([1, ])
    print(x)
```
The expansion of detected expressions and the new fixes gated behind previews.

## Test Plan
- Added B006_9.py with a bunch of test cases
- Generated snapshots

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_bugbear/snapshots/ruff_linter__rules__flake8_bugbear__tests__preview__B006_B006_9.py.snap
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/snapshots/ruff_linter__rules__flake8_bugbear__tests__preview__B006_B006_9.py.snap
@@ -72,7 +72,7 @@ help: Replace with `None`; initialize within function
    - def f4(x=(z := [1, ])):
 13 + def f4(x=None):
 14 +     if x is None:
-15 +         x = (z := [1])
+15 +         x = (z := [1, ])
 16 |     print(x)
 17 | 
 18 | 
@@ -92,7 +92,7 @@ help: Replace with `None`; initialize within function
    - def f5(x=([1, ])):
 17 + def f5(x=None):
 18 +     if x is None:
-19 +         x = [1]
+19 +         x = ([1, ])
 20 |     print(x)
 21 | 
 22 | 


crates/ruff_python_codegen/src/generator.rs
```diff
@@ -111,6 +111,10 @@ impl<'a> Generator<'a> {
         self.unparse_expr(expr, 0);
         self.generate()
     }
+    pub fn expr_parenthesized(mut self, expr: &Expr) -> String {
+        self.unparse_expr(expr, precedence::MAX);
+        self.generate()
+    }
```
IDrokin117: I've added this method to be able to generate an expression with parentheses. I didn't find a way to ensure it is parenthesized. Using `expr` to generate source code for Tuple, I got
```
tuple_expr = "([1], 2)"
.expr(tuple_expr) # -> "[1], 2", but expected ([1], 2)
```
I am not sure about this, would be appreciated for more sophisticated solution.

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
```diff
@@ -122,32 +122,48 @@ pub(crate) fn mutable_argument_default(checker: &Checker, function_def: &ast::St
...
-    parameter: &Parameter,
+    parameter: &ParameterWithDefault,
     default: &Expr,
     semantic: &SemanticModel,
     locator: &Locator,
     stylist: &Stylist,
     indexer: &Indexer,
     generator: Generator,
+    comment_ranges: &CommentRanges,
+    source: &str,
```
IDrokin117: Would it be reasonable to eliminate all these parameters and pass only the `checker` object? Since this is a private function that will not be utilized elsewhere, using `checker` directly might be more convenient.

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
```diff
@@ -167,14 +188,14 @@ fn move_initialization(
...
+    let _ = write!(&mut content, "if {} is None:", parameter.parameter.name());
     content.push_str(stylist.line_ending().as_str());
     content.push_str(stylist.indentation());
     let _ = write!(
         &mut content,
         "{} = {}",
-        parameter.name(),
-        generator.expr(default)
+        parameter.parameter.name(),
+        generator.expr_parenthesized(default)
```
dylwil3: Is there some reason why the following would be a bad idea? It has the virtue of preserving whatever idiosyncracies were used in the original formatting, as well as type annotations. (I realize this is sort of orthogonal to the PR, but it looks like you had to use `expr_parenthesized` anyway to avoid a syntax error, and this is an alternative approach which feels like an improvement in any case?)

Do you want to try it and see if anything horrible happens in the ecosystem check?

```diff
diff --git i/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs w/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
index 6a753f94a..ba2b2261b 100644
--- i/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
+++ w/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
@@ -5,7 +5,7 @@ use ruff_python_ast::helpers::is_docstring_stmt;
 use ruff_python_ast::name::QualifiedName;
 use ruff_python_ast::parenthesize::parenthesized_range;
 use ruff_python_ast::{self as ast, Expr, ParameterWithDefault};
-use ruff_python_codegen::{Generator, Stylist};
+use ruff_python_codegen::Stylist;
 use ruff_python_index::Indexer;
 use ruff_python_semantic::SemanticModel;
 use ruff_python_semantic::analyze::function_type::is_stub;
@@ -128,7 +128,6 @@ pub(crate) fn mutable_argument_default(checker: &Checker, function_def: &ast::St
                 checker.locator(),
                 checker.stylist(),
                 checker.indexer(),
-                checker.generator(),
                 checker.comment_ranges(),
                 checker.source(),
             ) {
@@ -161,7 +160,6 @@ fn move_initialization(
     locator: &Locator,
     stylist: &Stylist,
     indexer: &Indexer,
-    generator: Generator,
     comment_ranges: &CommentRanges,
     source: &str,
 ) -> Option<Fix> {
@@ -191,12 +189,7 @@ fn move_initialization(
     let _ = write!(&mut content, "if {} is None:", parameter.parameter.name());
     content.push_str(stylist.line_ending().as_str());
     content.push_str(stylist.indentation());
-    let _ = write!(
-        &mut content,
-        "{} = {}",
-        parameter.parameter.name(),
-        generator.expr_parenthesized(default)
-    );
+    let _ = write!(&mut content, "{}", locator.slice(parameter));
     content.push_str(stylist.line_ending().as_str());
 
     // Determine the indentation depth of the function body.

````

File: crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs (hunk 2 above)
Comment (dylwil3) [reply crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs]: That seems fine to me - go for it!

File: crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs (hunk 3 above)
Comment (dylwil3) [reply crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs]: I guess the formatting of assignments is often a little different than for parameter defaults, so you could also keep it closer to the original fix but use the locator to slice out the `default` instead of generating it?

Comment (IDrokin117) [reply crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs]: @dylwil3 I've experimented with your proposal and got some results:
1. Using `write!(&mut content, "{}", locator.slice(parameter));` preserves formatting, but, as you mention, variable assignment requires spaces around `=` unlike parameter.
2. Using `locator.slice(default)` will lead to a problem that `generator.expr_parenthesized(default)` tries to solve: `default` doesn't grab parenthesizes. 
```rust
let _ = write!(
    &mut content,
    "{} = {}",
    parameter.parameter.name(),
    locator.slice(default)
);
```
I used `locator.slice` to grab annotation from parameter and place it to variable assignment. What you think?

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
IDrokin117: Done

IDrokin117:
@dylwil3 Could you please review the PR again?

dylwil3:
> @dylwil3 Could you please review the PR again?

Yes! Sorry for the delay - I will not be able to get to this till the end of the week, but I have not forgotten!

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
dylwil3: I pushed a commit to grab the `parenthesized_range` instead which seems to work!

PR #20027, opened 2025-08-21
[`pyupgrade`] Enable rule triggering for stub files  (`UP043`)

### LINKED ISSUE #20011: `unnecessary-default-type-args (UP043)` isn't triggering on stubs
### Summary

`UP043` doesn't trigger at all on typeshed. I guess because it doesn't ignore th version in stubs? (relates to https://github.com/astral-sh/ruff/issues/19566)
```
(typeshed) PS ...\typeshed> uv run ruff check --select=UP043
All checks passed!
```

### Version

ruff 0.12.9

IDrokin117:
## Summary
Resolves #20011

Implemented alternative triggering condition for rule [`UP043`](https://docs.astral.sh/ruff/rules/unnecessary-default-type-args/) based on requirements outlined in [issue #20011](https://github.com/astral-sh/ruff/issues/20011)
## Test Plan
Created .pyi file to ensure triggering the rule

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP043.pyi
+++ b/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP043.pyi
@@ -1,0 +1,27 @@
+# https://github.com/python/typeshed/blob/main/stubs/networkx/networkx/algorithms/community/louvain.pyi
+from _typeshed import Incomplete
+from collections.abc import Generator
+
+from networkx.classes.graph import Graph, _Node
+from networkx.utils.backends import _dispatchable
+from numpy.random import RandomState
+
+__all__ = ["louvain_communities", "louvain_partitions"]
+
+@_dispatchable
+def louvain_communities(
+        G: Graph[_Node],
+        weight: str | None = "weight",
+        resolution: float | None = 1,
+        threshold: float | None = 1e-07,
+        max_level: int | None = None,
+        seed: int | RandomState | None = None,
+) -> list[set[Incomplete]]: ...
+@_dispatchable
+def louvain_partitions(
+        G: Graph[_Node],
+        weight: str | None = "weight",
+        resolution: float | None = 1,
+        threshold: float | None = 1e-07,
+        seed: int | RandomState | None = None,
+) -> Generator[list[set[Incomplete]], None, None]: ...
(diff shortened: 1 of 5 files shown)


dylwil3:
This looks right to me but since there's (necessarily) such a huge diff in the ecosystem it would be great if e.g. @AlexWaygood could confirm it's all good (and shouldn't be in preview)?

IDrokin117:
@AlexWaygood Could you please review PR?

crates/ruff_linter/resources/test/fixtures/pyupgrade/UP043.pyi
ntBre, line 1: Thanks for identifying one of the files from the issue! However, I'd probably lean toward a stripped-down example over including a file verbatim from another project since the specifics of the code aren't really that important. We could even just duplicate `UP043.py` and call it `UP043.pyi`, I think.

crates/ruff_linter/src/rules/pyupgrade/rules/unnecessary_default_type_args.rs
```diff
@@ -7,7 +7,7 @@ use crate::{AlwaysFixableViolation, Applicability, Edit, Fix};
 
 /// ## What it does
 /// Checks for unnecessary default type arguments for `Generator` and
-/// `AsyncGenerator` on Python 3.13+.
+/// `AsyncGenerator` on Python 3.13+ or stubs.
```
ntBre:
```suggestion
/// `AsyncGenerator` on Python 3.13+ or in stubs.
```

AlexWaygood:
Yup, it makes sense for this to always be enabled for stubs regardless of Python version.

I agree with Brent that this should probably be gated behind preview initially. The change was requested by a typeshed maintainer, and I'm also a typeshed maintainer, so I wouldn't worry too much about the impact on typeshed (it's not going to take _us_ by surprise! 😄) -- but there are other authors of stubs out there, and it could take them by surprise to have such a big expansion of the rule go directly into a stable rule without a preview-only period first.

AlexWaygood:
(Sorry for the delayed response -- been a bit busy recently with a house move and other things!)

IDrokin117:
@AlexWaygood Thanks!  
@ntBre I've gated changes behind preview.

PR #20028, opened 2025-08-21
[ty] Shrink size of `AstNodeRef`

ibraheemdev:
## Summary

Removes the `module_ptr` field from `AstNodeRef` in release mode, and change `NodeIndex` to a `NonZeroU32` to reduce the size of `Option<AstNodeRef<_>>` fields.

I believe CI runs in debug mode, so this won't show up in the memory report, but this reduces memory by ~2% in release mode.


crates/ty_python_semantic/src/ast_node_ref.rs
```diff
@@ -119,20 +119,11 @@ unsafe impl<T> salsa::Update for AstNodeRef<T> {
...
-            false
-        } else {
-            *old_ref = new_value;
-            true
-        }
+        // We could store the module pointer and check equality of both the module pointer
+        // and node index, but any change to the AST is likely to invalidate most node indices
+        // anyways, so we save on memory by not storing the module pointer and assuming all
+        // nodes have changed.
+        *old_ref = new_value;
```
ibraheemdev: I'm not completely sure if this is worth the memory usage improvements, cc @MichaReiser.

crates/ruff_python_ast/src/node_index.rs
```diff
@@ -40,30 +40,30 @@ impl AtomicNodeIndex {
...
 impl From<u32> for NodeIndex {
     fn from(value: u32) -> Self {
-        NodeIndex(value)
-    }
-}
-
-impl From<u32> for AtomicNodeIndex {
-    fn from(value: u32) -> Self {
-        AtomicNodeIndex(AtomicU32::from(value))
+        NodeIndex(NonZeroU32::new(value + 1).unwrap())
```
MichaReiser: I was a bit confused by the `Debug` output because it wasn't clear to me that `..` means the same as dummy. I think something like `AtomicNodeIndex::dummy` would be more clear (and also avoids the line break)

crates/ruff_python_ast/src/node_index.rs
```diff
@@ -40,30 +40,30 @@ impl AtomicNodeIndex {
...
     }
 
     pub fn as_u32(self) -> u32 {
-        self.0
+        self.0.get() - 1
     }
 }
 
 impl From<u32> for NodeIndex {
     fn from(value: u32) -> Self {
```
MichaReiser: Should we add a `debug_assertion` here that `value + 1 != dummy`

crates/ty_python_semantic/src/ast_node_ref.rs
MichaReiser: It's probably fine because `AstNodeRef` fields are all tracked separately (it's very likely that many indices change when you make a change in code) and
any query depending on an `AstNodeRef` field always depends on the AST too. 

Could we now derive `Update?

ibraheemdev: > Could we now derive `Update`

I don't think so, the `Update` derive checks for equality of the fields, which would be incorrect here.

ibraheemdev:
Here's the diff from on a release build of `prefect`. The effect on `Definition` and `Expression` is quite significant.
```diff
 Found 29228 diagnostics
 =======SALSA STRUCTS=======
-`Definition`                                       metadata=19.83MB  fields=39.66MB  count=495746
-`Expression`                                       metadata=12.75MB  fields=14.87MB  count=265605
+`Definition`                                       metadata=19.83MB  fields=27.76MB  count=495746
+`Expression`                                       metadata=12.75MB  fields=6.37MB   count=265605
 `StringLiteralType`                                metadata=3.24MB   fields=5.88MB   count=57943
 `IntersectionType`                                 metadata=1.99MB   fields=4.60MB   count=35545
 `File`                                             metadata=1.69MB   fields=4.52MB   count=26449
@@ -24,12 +24,12 @@
 `FileModule`                                       metadata=0.11MB   fields=0.29MB   count=1992
 `GenericAlias`                                     metadata=0.91MB   fields=0.26MB   count=16249
 `ModuleNameIngredient`                             metadata=0.24MB   fields=0.23MB   count=4344
-`Unpack`                                           metadata=0.17MB   fields=0.20MB   count=4223
 `ModuleLiteralType`                                metadata=0.44MB   fields=0.17MB   count=8470
 `TypeVarInstance`                                  metadata=0.12MB   fields=0.16MB   count=2176
 `StarImportPlaceholderPredicate`                   metadata=0.22MB   fields=0.16MB   count=7889
 `BoundMethodType`                                  metadata=0.36MB   fields=0.16MB   count=6485
 `GenericContext`                                   metadata=0.10MB   fields=0.15MB   count=1822
+`Unpack`                                           metadata=0.17MB   fields=0.14MB   count=4223
 `BoundTypeVarInstance`                             metadata=0.25MB   fields=0.07MB   count=4513
 `PatternPredicate`                                 metadata=0.02MB   fields=0.07MB   count=768
 `Project`                                          metadata=0.00MB   fields=0.06MB   count=1
@@ -51,7 +51,7 @@
 `module_type_symbols::interned_arguments`          metadata=0.00MB   fields=0.00MB   count=1
 =======SALSA QUERIES=======
 `semantic_index -> ty_python_semantic::semantic_index::SemanticIndex`
-    metadata=21.75MB  fields=409.71MB count=6117
+    metadata=21.75MB  fields=407.52MB count=6117
 `parsed_module -> ruff_db::parsed::ParsedModule`
     metadata=0.47MB   fields=284.40MB count=6126
 `use_def_map -> alloc::sync::Arc<ty_python_semantic::semantic_index::use_def::UseDefMap>`
@@ -157,8 +157,8 @@
 `TypeVarInstance < 'db >::lazy_constraints_ -> core::option::Option<ty_python_semantic::types::TypeVarBoundOrConstraints>`
     metadata=0.00MB   fields=0.00MB   count=3
 =======SALSA SUMMARY=======
-TOTAL MEMORY USAGE: 1420.38MB
+TOTAL MEMORY USAGE: 1397.73MB
     struct metadata = 71.31MB
-    struct fields = 93.10MB
+    struct fields = 72.63MB
     memo metadata = 213.48MB
-    memo fields = 1042.49MB
+    memo fields = 1040.30MB
```

PR #20030, opened 2025-08-21
[ty] Use Salsa to cache workspace symbols

BurntSushi:
This PR pretty much does what astral-sh/ty#998 suggests: it turns the
symbol gathering at the file level into a Salsa query while being
careful not to make the query string a dependency of that query. This
lets Salsa cache the symbols on a per-file basis, while we do the
filtering in a separate aggregation step.

This PR also includes a (small) bonus optimization using a regex for
query filtering. :-)

Fixes astral-sh/ty#998

### DIFF
--- a/crates/ty_ide/src/symbols.rs
+++ b/crates/ty_ide/src/symbols.rs
@@ -456,60 +456,60 @@ impl SourceOrderVisitor<'_> for SymbolVisitor {
 
             Stmt::Assign(assign) => {
                 // Include assignments only when we're in global or class scope
-                if !self.in_function {
-                    for target in &assign.targets {
-                        if let Expr::Name(name) = target {
-                            let kind = if Self::is_constant_name(name.id.as_str()) {
-                                SymbolKind::Constant
-                            } else if self
-                                .iter_symbol_stack()
-                                .any(|s| s.kind == SymbolKind::Class)
-                            {
-                                SymbolKind::Field
-                            } else {
-                                SymbolKind::Variable
-                            };
+                if self.in_function {
+                    return;
+                }
+                for target in &assign.targets {
+                    let Expr::Name(name) = target else { continue };
+                    let kind = if Self::is_constant_name(name.id.as_str()) {
+                        SymbolKind::Constant
+                    } else if self
+                        .iter_symbol_stack()
+                        .any(|s| s.kind == SymbolKind::Class)
+                    {
+                        SymbolKind::Field
+                    } else {
+                        SymbolKind::Variable
+                    };
 
-                            let symbol = SymbolTree {
-                                parent: None,
-                                name: name.id.to_string(),
-                                kind,
-                                name_range: name.range(),
-                                full_range: stmt.range(),
-                            };
-
-                            self.add_symbol(symbol);
-                        }
-                    }
+                    let symbol = SymbolTree {
+                        parent: None,
+                        name: name.id.to_string(),
+                        kind,
+                        name_range: name.range(),
+                        full_range: stmt.range(),
+                    };
+                    self.add_symbol(symbol);
                 }
             }
 
             Stmt::AnnAssign(ann_assign) => {
                 // Include assignments only when we're in global or class scope
-                if !self.in_function {
-                    if let Expr::Name(name) = &*ann_assign.target {
-                        let kind = if Self::is_constant_name(name.id.as_str()) {
-                            SymbolKind::Constant
-                        } else if self
-                            .iter_symbol_stack()
-                            .any(|s| s.kind == SymbolKind::Class)
-                        {
-                            SymbolKind::Field
-                        } else {
-                            SymbolKind::Variable
-                        };
-
-                        let symbol = SymbolTree {
-                            parent: None,
-                            name: name.id.to_string(),
-                            kind,
-                            name_range: name.range(),
-                            full_range: stmt.range(),
-                        };
-
-                        self.add_symbol(symbol);
-                    }
+                if self.in_function {
+                    return;
                 }
+                let Expr::Name(name) = &*ann_assign.target else {
+                    return;
+                };
+                let kind = if Self::is_constant_name(name.id.as_str()) {
+                    SymbolKind::Constant
+                } else if self
+                    .iter_symbol_stack()
+                    .any(|s| s.kind == SymbolKind::Class)
+                {
+                    SymbolKind::Field
+                } else {
+                    SymbolKind::Variable
+                };
+
+                let symbol = SymbolTree {
+                    parent: None,
+                    name: name.id.to_string(),
+                    kind,
+                    name_range: name.range(),
+                    full_range: stmt.range(),
+                };
+                self.add_symbol(symbol);
             }
 
             _ => {


crates/ty_ide/src/symbols.rs
```diff
@@ -16,7 +18,55 @@ pub struct SymbolsOptions {
...
+        // In theory regex compilation could fail if the pattern string
+        // was long enough to exceed the default regex compilation size
+        // limit. But this length would be approaching ~10MB or so.
+        if let Ok(re) = Regex::new(&pattern) {
+            return QueryPattern {
+                re,
+                original: literal_query_string.to_string(),
+            };
+        }
+        todo!()
```
dhruvmanila: Is this a leftover? Or, should this instead of `unreachable` / `panic`?

crates/ty_ide/src/symbols.rs
```diff
@@ -81,48 +131,58 @@ pub(crate) fn symbols_for_file(
         "Cannot use hierarchical mode with a query string"
     );
 
+    let ingredient = SymbolsOptionsIngredient::new(db, options.hierarchical, options.global_only);
+    let mut symbols = symbols_for_file_inner(db, file, ingredient);
+    if let Some(ref query) = options.query_string {
+        symbols.retain(|symbol| query.is_match(symbol));
+    }
```
dhruvmanila: Should we special case an empty string here so that we don't pay the cost of trying to match it when it's going to match every symbol? My main motivation is that the spec mentions:

> Clients may send an empty string here to request all symbols.

crates/ty_server/src/server/api/requests/workspace_symbols.rs
```diff
@@ -29,7 +29,13 @@ impl BackgroundRequestHandler for WorkspaceSymbolRequestHandler {
         // Iterate through all projects in the session
         for db in snapshot.projects() {
             // Get workspace symbols matching the query
+            let start = std::time::Instant::now();
             let workspace_symbol_infos = workspace_symbols(db, query);
+            tracing::debug!(
+                "Workspace symbols request returned {len} suggestions in {elapsed:?}",
+                len = workspace_symbol_infos.len(),
+                elapsed = std::time::Instant::now().duration_since(start)
+            );
```
dhruvmanila: nit: I think the server does log the time it takes for the entire request to run, so we might want to re-word this to not mention "request" to avoid any confusion? Maybe something like: "Found {len} workspace symbols in {elapsed:?}"

crates/ty_ide/src/symbols.rs
```diff
@@ -81,48 +131,58 @@ pub(crate) fn symbols_for_file(
...
+    symbols
+}
+
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+struct SymbolsOptionsIngredient<'db> {
+    hierarchical: bool,
```
MichaReiser: I'm not sure if it's worth it but I'm wondering if we want to remove the `hierarchical` flag here. I suspect that the most costly part is iterating over all files, parsing them, and not necessarily the filtering/post processing once we found a matching file. 

Also, having both `hierarchical` and `global_only` also means that we have up to 4 cached results for each file.

Could we instead change `symbols_for_file_inner` to return `IndexVec<SymbolIndex, SymbolInfo>` where:

```rust
struc SymbolInfo {
	parent: Option<SymbolIndex>,
	/// The name of the symbol
  pub name: String,
  /// The kind of symbol (function, class, variable, etc.)
  pub kind: SymbolKind,
  /// The range of the symbol name
  pub name_range: TextRange,
  /// The full range of the symbol (including body)
  pub full_range: TextRange,
}
```

and we do the transformation to a hierarchical representation post filtering.

This would then allow us to remove this struct altogether and simply have two queries:

```rust
#[salsa::tracked(returns(ref))]
fn global_symbols_for_file(db: &dyn Db, file: File);

#[salsa::tracked(returns(ref))]
fn symbols_for_file(db: &dyn Db, file: File);
```

The advantage of using two different queries is that it removes the need of interning the arguments because the function has exactly one argument

crates/ty_ide/src/symbols.rs
```diff
@@ -81,48 +131,58 @@ pub(crate) fn symbols_for_file(
...
+    global_only: bool,
+}
+
+#[salsa::tracked]
```
MichaReiser: We should use `returns(ref)` or `returns(deref)` here to avoid cloning the returned vector

crates/ty_ide/src/symbols.rs
```diff
@@ -81,48 +131,58 @@ pub(crate) fn symbols_for_file(
...
 
+    let ingredient = SymbolsOptionsIngredient::new(db, options.hierarchical, options.global_only);
+    let mut symbols = symbols_for_file_inner(db, file, ingredient);
+    if let Some(ref query) = options.query_string {
+        symbols.retain(|symbol| query.is_match(symbol));
+    }
+    symbols
+}
+
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
```
MichaReiser: You don't need to intern the `Options` struct. In fact, `symbols_for_file_inner` now requires two level of interning:

1. Your manual interning of `SymbolsOptionsIngredient`
2. Salsa creates a hidden interned struct `Arguments { file: File, options: SymbolsOptionsIngredient }` because the query has more than 1 argument (besides `db`). 

You can simply remove the `interned` here and use salsa's "behind the scene" interning

This is the code in salsa that decides if an interner is required:

First, Salsa decides the "function type" by looking at how many arguments a function has

https://github.com/MichaReiser/salsa/blob/f39e1947af0b00f63afac32e27449029c1b1499c/components/salsa-macros/src/tracked_fn.rs#L335-L344

Based on that, it decides if interning is required or not

https://github.com/MichaReiser/salsa/blob/f39e1947af0b00f63afac32e27449029c1b1499c/components/salsa-macros/src/tracked_fn.rs#L164-L167

Ideally, interning wouldn't be required for constant functions but it currently does :(

crates/ty_server/src/server/api/requests/workspace_symbols.rs
```diff
@@ -29,7 +29,13 @@ impl BackgroundRequestHandler for WorkspaceSymbolRequestHandler {
         // Iterate through all projects in the session
         for db in snapshot.projects() {
             // Get workspace symbols matching the query
+            let start = std::time::Instant::now();
```
MichaReiser, line 32: This is, sort of, a separate optimization but parallelization would help a great deal with workspace symbols:

```rust
    let files = project.files(db);
    let results = std::sync::Mutex::new(Vec::new());
    {
        let db = db.dyn_clone();
        let files = &files;
        let options = &options;
        let results = &results;

        rayon::scope(move |s| {
            // For each file, extract symbols and add them to results
            for file in files.iter() {
                let db = db.dyn_clone();
                s.spawn(move |_| {
                    let file_symbols = symbols_for_file(&*db, *file, options);

                    for symbol in file_symbols {
                        results.lock().unwrap().push(WorkspaceSymbolInfo {
                            symbol,
                            file: *file,
                        });
                    }
                });
            }
        });
    }

    results.into_inner().unwrap()
```

This makes workspace symbols complete on home assistant and without caching and with a debug build in under 2 seconds and it is instant with caching

```
9:16:47.935070000 DEBUG request{id=8 method="workspace/symbol"}: Workspace symbols request returned 133750 suggestions in 631.400375ms
2025-08-22 09:16:47.968469000 DEBUG request{id=9 method="workspace/symbol"}: Workspace symbols request returned 101056 suggestions in 412.80475ms
2025-08-22 09:16:49.367840000 DEBUG request{id=10 method="workspace/symbol"}: Workspace symbols request returned 42840 suggestions in 33.498042ms
2025-08-22 09:16:52.971051000 DEBUG request{id=11 method="workspace/symbol"}: Workspace symbols request returned 1874 suggestions in 57.296833ms
2025-08-22 09:16:55.151914000 DEBUG request{id=12 method="workspace/symbol"}: Workspace symbols request returned 4162 suggestions in 48.12ms
2025-08-22 09:16:55.993286000 DEBUG request{id=13 method="workspace/symbol"}: Workspace symbols request returned 7559 suggestions in 57.637541ms
```

crates/ty_ide/src/symbols.rs
BurntSushi: Whoops, forgot to circle back around to this!

I'd enable a Clippy lint for checking `todo!()`, but it looks like we're meaningfully using it in at least one place. I added a comment near there to add `todo = "warn"` once `todo!()` is gone.

crates/ty_ide/src/symbols.rs
BurntSushi: Sounds reasonable! Done.

crates/ty_server/src/server/api/requests/workspace_symbols.rs
BurntSushi: Makes sense!

crates/ty_ide/src/symbols.rs
BurntSushi: Nice. This required a few other touchups, and we do ultimately still clone the `SymbolInfo`, but at least now we only clone the ones that have passed the query string filter.

crates/ty_ide/src/symbols.rs
BurntSushi: Got it, makes sense. Done!

crates/ty_server/src/server/api/requests/workspace_symbols.rs
BurntSushi: Nice! Applied!

crates/ty_ide/src/symbols.rs
MichaReiser: Uff, this turned out to be more complicated than I thought. I'm sorry

BurntSushi: Phew, this is done. I think the API is now better for it too.

BurntSushi:
> Do we have any numbers on the difference between a few requests like (a) no query (b) with small and large query string?

After the initial cold run, requests generally take a few dozen milliseconds. There isn't a ton of variation with respect to query string length (which is what I'd expect I think).

The slowest part of the process here is when we return _a lot_ of symbols and VS Code needs to update its drop-down UI. There's a fairly sizeable delay there (on the order of a couple seconds). But to VS Code's credit, if you keep typing, it's pretty good about stopping what it was doing and moving on to the next request.

It overall feels pretty good. It'd be nice if VS Code was a little snappier, but I think the only thing we could do there is return fewer symbols. And I'm not sure if that's a good idea.

crates/ty_ide/src/symbols.rs
```diff
@@ -1,41 +1,248 @@
...
+            // all parents are ordered before their children.
+            children_ids[parent_id].push(id);
+        }
+
+        let mut symbols = IndexVec::new();
+        let mut children: Vec<SymbolId> = vec![];
+        let mut last_end: usize = 0;
+        for (tree, child_symbol_ids) in self.symbols.iter().zip(children_ids) {
+            let start = last_end;
+            let end = start + child_symbol_ids.len();
```
MichaReiser, line 133: I'm a bit surprised this works. Is it guaranteed that we add all symbols (children) in breadth-first and not depth-first order? Or am I overlooking something here that makes this possible?

We use a similar representation for `Scope` in the `SemanticIndex` where each `Scope` stores a `Range<FileScopeId>`. However, that range doesn't capture the `Scope's` children; it's the `Scope'`s descendants. Finding the `Scope'`s children requires iterating over the descendants and skipping all those with a different `parent`.

It's not fully clear to me why we don't have the descendants problem here. Maybe it's worth adding a comment why `start + child_symbol_ids.len()` works.

https://github.com/astral-sh/ruff/blob/4ac2b2c22283c9f8eb7f7fd945ab1c5f08098ab4/crates/ty_python_semantic/src/semantic_index.rs#L612-L615

BurntSushi: Yeah I can add a comment. It works because we guarantee all parents are added before their children in the visitor. I think I wrote a comment about that somewhere, but I should make it more prominent.

All this code is doing is taking the `Map<SymbolId, Vec<SymbolId>>` representation created above and flattening it.

Issue #14445, opened 2024-11-19
[Linter panic] when fixing W391 in Jupyter notebook


beskep:
- command: `ruff check .\src\test.ipynb --fix --diff`
(No panic when `--isolated` included)

- ruff version: 0.7.4

- pyproject.toml
```toml
[project]
requires-python = ">=3.13"
dependencies = ["ruff>=0.7.4"]

[tool.ruff]
preview = true

[tool.ruff.lint]
select = ["W"]
```

- ipynb file: [test.ipynb.zip](https://github.com/user-attachments/files/17807695/test.ipynb.zip)

dylwil3:
Thanks so much for this!

This can be reproduced with a notebook that just consists of three empty cells. We are somehow counting empty cells as newlines and then erroring when we try to "delete" them.

dylwil3:
Actually this bug has a wider blast radius.

If you make a notebook with three cells like this:

```python
a = [1]
a.append(2)
# new cell
a.append(3)
# new cell
a.append(4)
```
and apply the unsafe fix for `FURB118` you also get a panic.

dhruvmanila:
I'm guessing that cell deletion is creating a problem when `update_cell_offsets` and `update_cell_contents` are being called.

dhruvmanila:
The root cause is because the edit spans across the cell boundary which is an invariant we like to follow. In the case of `FURB113`, the diagnostics is between 3 cells and thus the edit is as well.

Refer to how the cell offsets are being updated (incorrectly):
```rs
// Original cell offsets
[crates/ruff_notebook/src/notebook.rs:229:9] &self.cell_offsets = CellOffsets(
    [
        0,
        20,
        32,
        44,
    ],
)

// The edit
[crates/ruff_linter/src/fix/mod.rs:97:13] edit = Edit {
    range: 8..43,
    content: Some(
        "x.extend((2, 3, 4))",
    ),
}

// Source map created for the edit
[crates/ruff_notebook/src/notebook.rs:230:9] source_map = SourceMap(
    [
        SourceMarker {
            source: 8,
            dest: 8,
        },
        SourceMarker {
            source: 43,
            dest: 27,
        },
    ]
)

// After the above source map is applied to update the cell offsets
[crates/ruff_notebook/src/notebook.rs:263:9] &self.cell_offsets = CellOffsets(
    [
        0,
        20,
        32,
        28,
    ],
)
```

I don't this rule existed when I implemented this.

[...]

PR #20036, opened 2025-08-21
Improve diff rendering for notebooks

ntBre:
## Summary

As noted in a code TODO, our `Diff` rendering code previously didn't have any
special handling for notebooks. This was particularly obvious when the diffs
were rendered right next to the corresponding diagnostic because the diagnostic
used cell-based line numbers, while the diff was still using line numbers from
the concatenated source. This PR updates the diff rendering to handle notebooks
too.

The main improvements shown in the example below are:

- Line numbers are now remapped to be relative to their cell
- Context lines from other cells are suppressed

```
error[unused-import][*]: `math` imported but unused                             
 --> notebook.ipynb:cell 2:2:8                                                  
  |                                                                             
1 | # cell 2                                                                    
2 | import math                                                                 
  |        ^^^^                                                                 
3 |                                                                             
4 | print('hello world')                                                        
  |                                                                             
help: Remove unused import: `math`                                              
                                                                                
ℹ Safe fix                                                                      
1 1 | # cell 2                                                                  
2   |-import math                                                               
3 2 |                                                                           
4 3 | print('hello world')                                                      
```

I tried a few different approaches here before finally just splitting the notebook into separate text ranges by cell and diffing each one separately. It seems to work and passes all of our tests, but I don't know if it's actually enforced anywhere that a single edit doesn't span cells. Such an edit would silently be dropped right now since it would fail the `contains_range` check. I also feel like I may have overlooked an existing way to partition a file into cells like this.

## Test Plan

Existing notebook tests, plus a new one in `ruff_db`

### DIFF
--- a/crates/ruff_source_file/src/lib.rs
+++ b/crates/ruff_source_file/src/lib.rs
@@ -105,12 +105,6 @@ impl<'src, 'index> SourceCode<'src, 'index> {
     pub fn line_count(&self) -> usize {
         self.index.line_count()
     }
-
-    /// Returns the [byte offsets](TextSize) for every line
-    #[inline]
-    pub fn line_starts(&self) -> &[TextSize] {
-        self.index.line_starts()
-    }
 }
 
 impl PartialEq<Self> for SourceCode<'_, '_> {


dhruvmanila:
> It seems to work and passes all of our tests, but I don't know if it's actually enforced anywhere that a single edit doesn't span cells.

It is enforced, just not in a good way ;)

Related issue: https://github.com/astral-sh/ruff/issues/14445

https://github.com/astral-sh/ruff/blob/7a44ea680e55f10af6e566eb5d9928244f79edd1/crates/ruff_notebook/src/notebook.rs#L278-L284

crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -106,20 +108,30 @@ impl std::fmt::Display for Diff<'_> {
...
-
-        for edit in self.fix.edits() {
-            output.push_str(source_code.slice(TextRange::new(last_end, edit.start())));
-            output.push_str(edit.content().unwrap_or_default());
-            last_end = edit.end();
+        // Partition the source code into ranges for each cell. If `self.notebook_index` is `None`,
+        // indicating a regular script file, all the lines will be in one "cell" under the `None`
+        // key.
+        let mut cells: BTreeMap<Option<OneIndexed>, TextRange> = BTreeMap::default();
+        for line in source_code.line_starts() {
```
MichaReiser: A few thoughts here:

* Iterating over all lines seems especially unfortunate if this isn't a notebook. Can't we just push a single entry if the notebook index is absent that spans the entire source code range instead of iterating over the lines?
* Are there cases where the notebook cells don't appear in source order? I'm asking because we don't need a `BTreeMap` if it's guaranteed that we iterate over the cells in order anyway
* This data structure seems very similar to `Notebook::cell_offsets`.

crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -133,59 +145,81 @@ impl std::fmt::Display for Diff<'_> {
         // tests, which is the only place these are currently used.
         writeln!(f, "ℹ {}", fmt_styled(message, self.stylesheet.separator))?;
 
-        let (largest_old, largest_new) = diff
-            .ops()
-            .last()
-            .map(|op| (op.old_range().start, op.new_range().start))
-            .unwrap_or_default();
+        for range in cells.values() {
```
MichaReiser: I wonder if we could reduce the overhead for non notebooks by extracing the shared logic for rendering fixes into a separate method and:

* We call that method directly if source isn't a notebook (in which case we can skip all/most of the operations?)
* For notebooks, do whatever mapping is necessary, call that method for each cell

crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__unused_variable.snap
```diff
@@ -49,12 +44,10 @@ F841 [*] Local variable `bar1` is assigned to but never used
 help: Remove assignment to unused variable `bar1`
 
 ℹ Unsafe fix
-2 2 |     foo1 = %matplotlib --list
-3 3 |     foo2: list[str] = %matplotlib --list
-4 4 | def f():
-5   |-    bar1 = !pwd
-  5 |+    !pwd
-6 6 |     bar2: str = !pwd
+1 1 | def f():
```
MichaReiser, line 50: I think we have to show the cell number somewhere because I don't think it's guaranteed that the diagnostic cell is the same as the one for fixes.

I'm also curious on how the rendering looks if a diagnostic has fixes across multiple cells. We should add a test for this. 

The UP049 example that you used to evaluate the fixes rendering could be a good example here too, just extend it so that the fixes span multiple cells.

ntBre: Oh that's a good point, in all of these cases the cell number matched. What do you think about the header lines I removed in this commit: 80f9b55621ba0c42e66a659dc47c0c66c2e80229? I was trying to match the horizontal line we draw for different sections, but I ended up ripping them out because I thought they were too noisy, especially for the single-cell cases in our tests. Another option I considered was something like the `:::` in our diagnostics:

```
error[unused-import]: `os` imported but unused
 --> notebook.ipynb:cell 1:2:8                
  |                                           
1 | # cell 1                                  
2 | import os                                 
  |        ^^                                 
  |                                           
 ::: notebook.ipynb:cell 2:2:8                
  |                                           
1 | # cell 2                                  
2 | import math                               
  |        ---- second cell                   
3 |                                           
4 | print('hello world')                      
  |                                           
help: Remove unused import: `os`              
```

I'll work on a test for these. Unfortunately I don't think we can use UP049 because the PEP-695 generics are restricted to a single class or function body, which I don't think can span cells. I'm sure I can find another rule, though.

MichaReiser: I like `::: notebook.ipynb:cell 2:2:8` except that I would strip everything except the cell number because it feels redundant

```
--> notebook.ipynb:cell 1:2:8                
  |                                           
1 | # cell 1                                  
2 | import os                                 
  |        ^^                                 
  |                                           
 ::: cell 2
  |                                           
1 | # cell 2                                  
2 | import math                               
  |        ---- second cell                   
3 |                                           
4 | print('hello world')                      
  |                                           
help: Remove unused import: `os`              
```

crates/ruff_db/src/diagnostic/render/full.rs
ntBre: Ah thanks, `Notebook::cell_offsets` does seem like what I need. I looked into adding a `FileResolver` method to get a `CellOffsets`, but it only works, or at least works easily, for ty files. Ruff's `FileResolver` only has access to a `NotebookIndex`, not the `Notebook` itself. Should I try to pass that down through `Diagnostics`, and `EmitterContext`?

I don't think we want to clone a whole `Notebook`, so I'm thinking of adding a new struct wrapping only the index and the cell offsets to store in the (renamed) `Diagnostics::notebook_indexes` field.

https://github.com/astral-sh/ruff/blob/0b6ce1c788bceea13afeac63ec25894a5edd4435/crates/ruff/src/diagnostics.rs#L338-L342

I guess we could also considering moving the `CellOffsets` into the `NotebookIndex`, making it the wrapper struct. They're mostly accessed via `Notebook::cell_offsets` anyway.

But yes, we definitely don't need to iterate the lines for a normal file.

ntBre: Another idea, I think we can reconstruct something more efficient than looping over all the lines by only looking at the unique entries in `NotebookIndex::row_to_cell`. That's probably the easiest thing to do.

ntBre:
I've updated the output formatting, and I believe also resolved Micha's other suggestions. As I mentioned [here](https://github.com/astral-sh/ruff/pull/20036#discussion_r2294063712) it seems a bit tricky to get access to a `Notebook` to call `Notebook::cell_offsets` directly, but I think I've reconstructed something similar here.

I didn't extract any code as suggested [here](https://github.com/astral-sh/ruff/pull/20036#discussion_r2292995749), but I think most of the overhead for non-notebooks, and for notebooks too really, should be resolved by getting rid of the `BTreeMap` and the loop over all of the source lines.

crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -654,6 +698,107 @@ print()
...
+        error[unused-import][*]: `os` imported but unused
+         --> notebook.ipynb:cell 1:2:8
+          |
+        1 | # cell 1
+        2 | import os
+          |        ^^
+          |
+        help: Remove unused import: `os`
+
+        ℹ Safe fix
```
MichaReiser, line 716: Not something we have to change as part of this PR. But I only just realized that our fix infrastructure always assumes that all changes belong to the file from the primary span. 

How would the rendered diagnostic look like if we had a sub-diagnostic with a span pointing to a different file? I don't think it's something we have to solve now but it might be worth documenting if the rendered output looks confusing.

ntBre: Opened an issue to track this here: https://github.com/astral-sh/ruff/issues/20080.

PR #20040, opened 2025-08-22
[WIP] Add search paths info to unresolved import diagnostics

Renkai:
try to solve https://github.com/astral-sh/ty/issues/457

### DIFF
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -72,6 +72,7 @@ use crate::place::{
     module_type_implicit_global_declaration, module_type_implicit_global_symbol, place,
     place_from_bindings, place_from_declarations, typing_extensions_symbol,
 };
+
 use crate::semantic_index::ast_ids::node_key::ExpressionNodeKey;
 use crate::semantic_index::ast_ids::{HasScopedUseId, ScopedUseId};
 use crate::semantic_index::definition::{
@@ -4960,11 +4961,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         let Some(builder) = self.context.report_lint(&UNRESOLVED_IMPORT, range) else {
             return;
         };
+
         let mut diagnostic = builder.into_diagnostic(format_args!(
             "Cannot resolve imported module `{}{}`",
             ".".repeat(level as usize),
             module.unwrap_or_default()
         ));
+
         if level == 0 {
             if let Some(module_name) = module.and_then(ModuleName::new) {
                 let program = Program::get(self.db());
@@ -4988,6 +4991,27 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                     }
                 }
             }
+
+            // Add search paths information to the diagnostic
+            let mut search_paths: Vec<String> = Vec::new();
+            let mut path_index = 1;
+
+            // Get all system paths (this includes first-party and site-packages)
+            for path in crate::module_resolver::system_module_search_paths(self.db()) {
+                search_paths.push(format!("  {}. {}", path_index, path));
+                path_index += 1;
+            }
+
+            // Note about vendored stdlib (since it's not included in system paths)
+            search_paths.push(format!(
+                "  {}. <vendored stdlib> (built-in Python standard library)",
+                path_index
+            ));
+
+            diagnostic.info(format_args!(
+                "Searched in the following paths:\n{}",
+                search_paths.join("\n")
+            ));
 
             diagnostic.info(
                 "make sure your Python environment is properly configured: \


Renkai:
This solution called `search_paths` again to show the paths, not sure if carry the original results here is a better way, that way seems need changes in many places.

carljm:
Thanks for working on this!

@AlexWaygood can you take the lead on review here?

AlexWaygood:
Sure

crates/ty_python_semantic/src/module_resolver/path.rs
```diff
@@ -700,6 +700,25 @@ impl SearchPath {
...
+    #[must_use]
+    pub(crate) fn describe_kind(&self) -> &'static str {
+        match *self.0 {
+            SearchPathInner::Extra(_) => {
+                "extra search path specified on the CLI or in your config file"
+            }
+            SearchPathInner::FirstParty(_) => "first-party code",
+            SearchPathInner::StandardLibraryCustom(_) => {
+                "custom stdlib stubs specified on the CLI or in your config file"
+            }
```
AlexWaygood, line 715: Ideally I feel like we'd be able to tell the user whether it was a CLI option or a config-file setting that caused us to look at an extra search path (or a custom stdlib search paths) during module resolution. But that would involve a fair bit of refactoring probably; I think it's best to leave it out of this PR.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -4989,6 +4994,23 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    .map(|(index, path)| {
+                        format!("  {}. {} ({})", index + 1, path, path.describe_kind())
+                    })
+                    .collect();
+
+            if !search_paths_list.is_empty() {
+                diagnostic.info(format_args!(
+                    "Searched in the following paths during module resolution:\n{}",
+                    search_paths_list.join("\n")
+                ));
```
AlexWaygood: @BurntSushi -- thoughts on having a multiline diagnostic like this? Would it be better to render each line as a separate `info` diagnostic, or is this okay?

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -64,14 +64,17 @@ use super::string_annotation::{
...
+    KnownModule, ModuleResolveMode, file_to_module, resolve_module, search_paths,
+};
 use crate::node_key::NodeKey;
 use crate::place::{
     Boundness, ConsideredDefinitions, LookupError, Place, PlaceAndQualifiers,
     builtins_module_scope, builtins_symbol, explicit_global_symbol, global_symbol,
     module_type_implicit_global_declaration, module_type_implicit_global_symbol, place,
     place_from_bindings, place_from_declarations, typing_extensions_symbol,
 };
+
```
BurntSushi: Why insert this extra newline? (And also below.)

I guess the ones below are whatever, but this one should probably be removed to keep `crate` imports together in the same block.

crates/ty_python_semantic/src/types/infer.rs
BurntSushi: Good catch. Yes, I think each item should be its own info sub-diagnostic, like what we do for overloads:

https://github.com/astral-sh/ruff/blob/e7237652a9048e5fa0e6ec002855d89f9554833b/crates/ty_python_semantic/src/types/call/bind.rs#L1809-L1819

crates/ty_python_semantic/src/types/infer.rs
Renkai: Maybe some of my local formatter did this, I just removed it. Is it possible to include your role `keep crate imports together in the same block` in `cargo fmt` or something similar?

crates/ty_python_semantic/src/types/infer.rs
Renkai: updated

crates/ty_python_semantic/src/types/infer.rs
BurntSushi: Do you mean [`imports_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#imports_granularity)? That isn't a stable rustfmt option.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -4989,6 +4991,26 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 }
             }
 
+            // Add search paths information to the diagnostic
+            // Use the same search paths function that is used in actual module resolution
+            let search_paths: Vec<_> =
+                search_paths(self.db(), ModuleResolveMode::StubsAllowed).collect();
```
BurntSushi: Why collect all the search paths into memory first? It seems a little gratuitous. You could do `let search_paths = search_paths(..).peekable();` and then `if search_paths.peek().is_some() {` and `for (index, path) in search_paths.enumerate() {` below.

Renkai: updated by peekable. It seems requires more review to trigger next CI test. But the test looks fine on my computer now.

BurntSushi:
Thank you!!!

PR #20041, opened 2025-08-22
[ty] Limit argument expansion size for overload call evaluation

dhruvmanila:

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/overloads.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/overloads.md
@@ -831,6 +831,76 @@ def _(foo: Foo, ab: A | B, a=1):
     )
 ```
 
+### Optimization: Limit expansion size
+
+To prevent combinatorial explosion, ty limits the number of argument lists created by expanding a
+single argument.
+
+`overloaded.pyi`:
+
+```pyi
+from typing import overload
+
+class A: ...
+class B: ...
+class C: ...
+
+@overload
+def f() -> None: ...
+@overload
+def f(**kwargs: int) -> C: ...
+@overload
+def f(x: A, /, **kwargs: int) -> A: ...
+@overload
+def f(x: B, /, **kwargs: int) -> B: ...
+```
+
+```py
+from overloaded import A, B, f
+from typing_extensions import reveal_type
+
+
+def _(ab: A | B, a=None):
+    reveal_type(
+        # error: [no-matching-overload]
+        # revealed: Unknown
+        f(
+            A(),
+            a1=a,
+            a2=a,
+            a3=a,
+            a4=a,
+            a5=a,
+            a6=a,
+            a7=a,
+            a8=a,
+            a9=a,
+            a10=a,
+            a11=a,
+            a12=a,
+            a13=a,
+            a14=a,
+            a15=a,
+            a16=a,
+            a17=a,
+            a18=a,
+            a19=a,
+            a20=a,
+            a21=a,
+            a22=a,
+            a23=a,
+            a24=a,
+            a25=a,
+            a26=a,
+            a27=a,
+            a28=a,
+            a29=a,
+            a30=a,
+        )
+    )
+```
+
+
 ## Filtering based on `Any` / `Unknown`
 
 This is the step 5 of the overload call evaluation algorithm which specifies that:
--- a/crates/ty_python_semantic/src/types/call/arguments.rs
+++ b/crates/ty_python_semantic/src/types/call/arguments.rs
@@ -131,6 +131,9 @@ impl<'a, 'db> CallArguments<'a, 'db> {
         &self,
         db: &'db dyn Db,
     ) -> impl Iterator<Item = Vec<CallArguments<'a, 'db>>> + '_ {
+        /// Maximum number of argument lists that can be generated in a single expansion step.
+        static MAX_EXPANSIONS: usize = 512;
+
         /// Represents the state of the expansion process.
         ///
         /// This is useful to avoid cloning the initial types vector if none of the types can be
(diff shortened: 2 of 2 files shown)


crates/ty_python_semantic/src/types/call/arguments.rs
```diff
@@ -172,7 +175,16 @@ impl<'a, 'db> CallArguments<'a, 'db> {
                 index += 1;
             };
 
-            let mut expanded_arguments = Vec::with_capacity(expanded_types.len() * previous.len());
+            let expansion_size = expanded_types.len() * previous.len();
+            if expansion_size > MAX_EXPANSIONS {
+                tracing::debug!(
+                    "Skipping argument type expansion as it would exceed the \
```
MichaReiser: Sorry for commenting on a draft PR. Is there some information that we could include to make it easier for users to narrow down where this is happening (e.g. the full signature of the called function?, or some information about the call site?)

dhruvmanila: I think we could expand the diagnostic of `no-matching-overload` to include this information as a note for users to notice this.

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Argument_type_expans…_-_Optimization___Limit_…_(cd61048adbc17331).snap
```diff
@@ -0,0 +1,183 @@
...
+35 | |             a26=a,
+36 | |             a27=a,
+37 | |             a28=a,
+38 | |             a29=a,
+39 | |             a30=a,
+40 | |         )
+   | |_________^
+41 |       )
+   |
+info: Limit of argument type expansions reached
```
dhruvmanila: I'm not exactly sure how best to convey this to the user in a way that they can understand. It seems that this kind of information would require additional context, possibly even linking to the overload call evaluation section of the typing spec. If anyone has any suggestions, I'd appreciate that.

crates/ty_python_semantic/resources/mdtest/call/overloads.md
```diff
@@ -831,6 +831,76 @@ def _(foo: Foo, ab: A | B, a=1):
...
+def f(x: A, /, **kwargs: int) -> A: ...
+@overload
+def f(x: B, /, **kwargs: int) -> B: ...
+```
+
+```py
+from overloaded import A, B, f
+from typing_extensions import reveal_type
+
+def _(ab: A | B, a=None):
```
carljm: A couple nits: 

1. `ab` parameter seems unused here
2. I think it's pretty implicit / unclear that we infer `Unknown | None` for `a` due to lack of annotation, plus default type -- but this is crucial to the test, because that's the union we repeatedly try to expand. Why not explicitly annotate `a: int | None` and leave out the default?

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Argument_type_expans…_-_Optimization___Limit_…_(cd61048adbc17331).snap
carljm: I think the piece of info that might be most useful to the user would be which argument types we were trying to expand. Not sure how hard that would be to add.

I also think this limit should ideally be high enough that almost no user should ever see it in realistic code, so we maybe don't need to spend much time perfecting the diagnostic; the key thing is not to have runaway execution time / memory.

carljm:
> proving to be difficult to implement mainly because we **need** to maintain the previous expansion to generate the next expansion

I think it is good to move on from this issue now, if the limit takes care of the runaway problem on the fuzzer code sample, and we don't have reports of this in real code. But for future, I'm curious -- why is the above hard to implement? It sounds like a custom-implemented iterator that stores some state, which doesn't seem particularly unusual?

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Argument_type_expans…_-_Optimization___Limit_…_(cd61048adbc17331).snap
dhruvmanila: > I think the piece of info that might be most useful to the user would be which argument types we were trying to expand. Not sure how hard that would be to add.

This isn't too difficult i.e., I can add the argument index value at which point the expansion exceeded the maximum limit.

> I also think this limit should ideally be high enough that almost no user should ever see it in realistic code, so we maybe don't need to spend much time perfecting the diagnostic; the key thing is not to have runaway execution time / memory.

Yes, agreed.

crates/ty_python_semantic/resources/mdtest/call/overloads.md
dhruvmanila: Thanks! I think that makes sense. I'll also update the parameter type for other functions to have an explicit annotation and not depend on the implicit behavior of having the default value without an annotation.

dhruvmanila:
> But for future, I'm curious -- why is the above hard to implement? It sounds like a custom-implemented iterator that stores some state, which doesn't seem particularly unusual?

I'll try to explain it using this example: `(bool, int | str, A | B | C)`, assuming that we're at the final argument in the expansion process, one of the data that we'd need to maintain is all the argument lists that resulted in expanding all the previous arguments i.e., 1 and 2 in this example. This means that when the expansion is at `A | B | C` argument, the data would contain `(True, int, ...), (False, int, ...), (True, str, ...), (False, str, ...)` (4) argument lists. Imagine this at a larger scale where the previous iteration could generate 200 argument lists which needs to be kept between iteration.

One way to resolve this could be that we only maintain the expanded types and generate the combinations for each iteration manually i.e., in the above example, for the first argument, we'd generate 2 combinations by expanding `bool` to `True` and `False` and we'll store the expanded list. Then, for the second argument, we'll expand `int | str` and generate 4 combinations. In this implementation, we'd save memory but each iteration would have to generate the combinations every time.

Does this make sense?

But, now that I think this out loud, maybe the second solution isn't too bad.

PR #20044, opened 2025-08-22
Fix incorrect lsp inlay hint type

MatthewMckee4:
## Summary

Previously we just communicated that the inlay hint type was "TYPE", so now we give the correct type

### DIFF
--- a/crates/ty_server/src/server/api/requests/inlay_hints.rs
+++ b/crates/ty_server/src/server/api/requests/inlay_hints.rs
@@ -9,7 +9,7 @@ use crate::session::client::Client;
 use lsp_types::request::InlayHintRequest;
 use lsp_types::{InlayHintParams, Url};
 use ruff_db::source::{line_index, source_text};
-use ty_ide::inlay_hints;
+use ty_ide::{InlayHintContent, inlay_hints};
 use ty_project::ProjectDatabase;
 
 pub(crate) struct InlayHintRequestHandler;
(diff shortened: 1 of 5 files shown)


crates/ty_server/src/server/api/requests/inlay_hints.rs
```diff
@@ -70,3 +70,10 @@ impl BackgroundDocumentRequestHandler for InlayHintRequestHandler {
 }
 
 impl RetriableRequestHandler for InlayHintRequestHandler {}
+
+fn inlay_hint_type(inlay_hint_content: &InlayHintContent) -> lsp_types::InlayHintKind {
```
MichaReiser: Nice! Would you mind to also wire it up in wasm and the playground?

https://github.com/astral-sh/ruff/blob/859475f017c3295ba2dbac144dcefdb2a2318250/crates/ty_wasm/src/lib.rs#L980-L987

and here

https://github.com/astral-sh/ruff/blob/859475f017c3295ba2dbac144dcefdb2a2318250/playground/ty/src/Editor/Editor.tsx#L410-L416

MichaReiser (commented):
Nice!

crates/ty_server/src/server/api/requests/inlay_hints.rs
MatthewMckee4: yeah

MatthewMckee4:
Not really sure how the ide handles these kinds differently

MichaReiser (approved):
Thank you

PR #20046, opened 2025-08-22
[ty] Disallow std::env and io methods in most ty crates

MichaReiser:
## Summary

We use the `System` abstraction in ty to abstract away the host/system on which ty runs. 
This has a few benefits:

* Tests can run in full isolation using a memory system (that uses an in-memory file system)
* The LSP has a custom implementation where `read_to_string` returns the content as seen by the editor (e.g. unsaved changes) instead of always returning the content as it is stored on disk
* We don't require any file system polyfills for wasm in the browser

However, it does require extra care that we don't accidentially use `std::fs` or `std::env` (etc.) methods in ty's code base (which is very easy).

This PR sets up clippy and disallowes the most common methods and points users towards the corresponding `System` methods instead. 

The setup is a bit awkward because clippy doesn't support inheriting configurations. That means, a crate can only override the entire workspace conifguration or not at all. 
The approach taken in this PR is:

* Configure the disallowed methods at the workspace level
* Allow `disallowed_methods` at the workspace level
* Enable the lint at the crate level using the warn attribute (in code)

The obvious downside of this is that it won't work if we ever want to disallow other methods... but that's something we can figure out once we reach that point.

## Test Plan

Clippy found a place where we incorrectly used `std::fs::read_to_string`

### DIFF
--- a/crates/ty_python_semantic/src/site_packages.rs
+++ b/crates/ty_python_semantic/src/site_packages.rs
@@ -211,7 +211,7 @@ impl PythonEnvironment {
         match VirtualEnvironment::new(path, system) {
             Ok(venv) => Ok(Self::Virtual(venv)),
             // If there's not a `pyvenv.cfg` marker, attempt to inspect as a system environment
-            Err(SitePackagesDiscoveryError::NoPyvenvCfgFile(path, _))
+            Err(SitePackagesDiscoveryError::NoPyvenvCfgFile(path, _, _))
                 if !path.origin.must_be_virtual_env() =>
             {
                 Ok(Self::System(SystemEnvironment::new(path)))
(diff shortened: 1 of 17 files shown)


crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -833,16 +839,26 @@ impl SystemEnvironment {
 #[derive(Debug)]
 pub enum SitePackagesDiscoveryError {
     /// `site-packages` discovery failed because the provided path couldn't be canonicalized.
-    CanonicalizationError(SystemPathBuf, SysPrefixPathOrigin, io::Error),
+    CanonicalizationError(
+        SystemPathBuf,
+        SysPrefixPathOrigin,
+        io::Error,
+        Box<dyn System>,
```
MichaReiser, line 846: This is a bit annoying. I don't love it but the only alternative that I can think of is to a) use a thread local or b) have a global `System` somewhere (which breaks test isolation)

MichaReiser:
Feel free to hit merge (or not merge ;))

crates/ty_python_semantic/src/site_packages.rs
AlexWaygood: Huh, yeah, I was initially surprised at these changes but I also don't see a better solution

AlexWaygood (approved):

carljm (approved):

PR #19783, opened 2025-08-06
[`pyflakes`/`pylint`] Skip `__class__` checks in method definitions (…

mikeleppane:
### Problem
Rules [F841](https://docs.astral.sh/ruff/rules/unused-variable/) (`unused-variable`) and [PLE0117](https://docs.astral.sh/ruff/rules/nonlocal-without-binding/) (`nonlocal-without-binding`) incorrectly flagged usage within method definitions, causing false positives.
Root Cause: Both rules failed to recognize that it is a special variable automatically available in Python method definitions. In methods, it is implicitly bound by the Python runtime, making assignments to it legitimate and declarations valid. `__class__``__class__``nonlocal __class__`

### Solution
- Added special handling for when detected within method definitions: `__class__`
- **F841**: Skip unused variable warnings for assignments in methods `__class__`
- **PLE0117**: Allow statements in methods without requiring explicit outer bindings `nonlocal __class__`

```python
class C:
    def set_class(self, cls):
        __class__ = cls  # No longer warns about unused variable
        
    def get_class(self):
        nonlocal __class__  # No longer warns about missing binding
        return __class__
```

## Test Plan

```bash
cargo test
```

PR #19424, opened 2025-07-18
[`pyupgrade`]  Apply `UP008` only when the `__class__` cell exists

IDrokin117:
## Summary

Resolves #19357 

Skip UP008 diagnostic for `builtins.super(P, self)` calls when `__class__` is not referenced locally, preventing incorrect fixes.  

**Note:** I haven't found concrete information about which cases `__class__` will be loaded into the scope. Let me know if anyone has references, it would be useful to enhance the implementation. I did a lot of tests to determine when `__class__` is loaded. Considered sources:
1.  [Python doc super](https://docs.python.org/3/library/functions.html#super)
2.  [Python doc classes](https://docs.python.org/3/tutorial/classes.html)
3. [pep-3135](https://peps.python.org/pep-3135/#specification)

As I understand it, Python will inject at runtime into local scope a `__class__` variable if it detects references to `super` or `__class__`. This allows calling `super()` and passing appropriate parameters. However, the compiler doesn't do the same for `builtins.super`, so we need to somehow introduce `__class__` into the local scope.

I figured out `__class__` will be in scope with valid value when two conditions are met:
1. `super` or `__class__` names have been loaded within function scope
4. `__class__` is not overridden.

I think my solution isn't elegant, so I would be appreciate a detailed review.

## Test Plan

Added 19 test cases, updated snapshots.

Issue #19357, opened 2025-07-15
UP008 should only apply when the `__class__` cell exists


dscorbett:
### Summary

[`super-call-with-parameters` (UP008)](https://docs.astral.sh/ruff/rules/super-call-with-parameters/) should only apply when `super` is a global variable or `__class__` is a nonlocal variable. Expressions that are merely equal to `super` aren’t good enough. Otherwise, the zero-argument `super` call doesn’t work. One solution is for the autofix to rewrite the `super`-equivalent expression to `super`, although that won’t work in all contexts. [Example](https://play.ruff.rs/852c338a-1d97-4fa4-bbbd-6f768ef373b8):
```console
$ cat >up008.py <<'# EOF'
import builtins
class A:
    def f(self):
        print("!")
class B(A):
    def f(self):
        builtins.super(B, self).f()
B().f()
# EOF

$ python up008.py
!

$ ruff --isolated check up008.py --select UP008 --unsafe-fixes --fix
Found 1 error (1 fixed, 0 remaining).

$ cat up008.py
import builtins
class A:
    def f(self):
        print("!")
class B(A):
    def f(self):
        builtins.super().f()
B().f()

$ python up008.py 2>&1 | tail -n 1
RuntimeError: super(): __class__ cell not found
```

Note that the expression that UP008 rewrites can still use `builtins.super` (or other expressions that equal `super` but are not literally `super`) but `super` or `__class__` just needs to be mentioned somewhere in the method in order for the `__class__` cell, which the zero-argument `super` uses, to be created. [True positive example](https://play.ruff.rs/4a54ab21-24a0-448a-9331-4b416e84ab75) with `super`:
```console
$ cat >up008_super.py <<'# EOF'
import builtins
super = None  # It still works even if `super` is shadowed!
class A:
    def f(self):
        print("!")
class B(A):
    def f(self):
        if False: super
        builtins.super(B, self).f()
B().f()
# EOF

$ ruff --isolated check up008_super.py --select UP008 --unsafe-fixes --fix
Found 1 error (1 fixed, 0 remaining).

$ python up008_super.py
!
```

[True positive example](https://play.ruff.rs/fd2bbb26-b678-46aa-9e2f-0d2e76862eb9) with `__class__`:
```console
$ cat >up008_class.py <<'# EOF'
import builtins
class A:
    def f(self):
        print("!")
class B(A):
    def f(self):
        if False: __class__
        builtins.super(B, self).f()
B().f()
# EOF

[...]

Issue #18442, opened 2025-06-03
F841 and PLE0117 don’t understand `__class__` bindings


dscorbett:
### Summary

By default, `__class__` within a method definition is a free variable and its corresponding cell variable is always bound. [`unused-variable` (F841)](https://docs.astral.sh/ruff/rules/unused-variable/) and [`nonlocal-without-binding` (PLE0117)](https://docs.astral.sh/ruff/rules/nonlocal-without-binding/) have false positives because they don’t understand this special case. This is probably a general problem with how Ruff resolves bindings which could affect other rules.

```console
$ cat >f841_ple0117.py <<'# EOF'
class A:
    x = 1


class B:
    x = 2


class C(A, B):
    def set_class(self, cls):
        nonlocal __class__
        if cls not in type(self).__mro__:
            raise ValueError(f"Invalid superclass: {cls}")
        __class__ = cls

    def get_class(self):
        return __class__

    def __getattribute__(self, attr):
        if attr in vars(type(self)):
            return object.__getattribute__(self, attr)
        if attr in vars(__class__):
            mro = type(self).__mro__
            return getattr(super(mro[mro.index(__class__) - 1], self), attr)
        return getattr(super(), attr)


c = C()
print(c.x)
c.set_class(B)
print(c.x)
# EOF

$ python f841_ple0117.py
1
2

$ ruff --isolated check --select F841,PLE0117 f841_ple0117.py --output-format concise -q
f841_ple0117.py:11:18: PLE0117 Nonlocal name `__class__` found without binding
f841_ple0117.py:14:9: F841 Local variable `__class__` is assigned to but never used

$ ruff --isolated check --select F841,PLE0117 f841_ple0117.py -s --unsafe-fixes --fix

$ python f841_ple0117.py
1
1
```

That PLE0117 diagnostic is a false positive because `__class__` does have a binding; it just isn’t explicit in the source code. That F841 diagnostic is a false positive because `__class__` is not a local variable.

### Version

ruff 0.11.12 (aee3af0f7 2025-05-29)

ntBre:
Wow, another interesting one! I assumed that `__class__` would also be bound in the class definition itself, but this is not the case:

```pycon
>>> class C:
...     print(__class__)
...
Traceback (most recent call last):
  File "<python-input-6>", line 1, in <module>
    class C:
        print(__class__)
  File "<python-input-6>", line 2, in C
    print(__class__)
          ^^^^^^^^^
[...]

PR #20048, opened 2025-08-22
Add a `ScopeKind` for the `__class__` cell

ntBre:
Summary
--

This PR aims to resolve (or help to resolve) #18442 and #19357 by encoding the CPython semantics around the `__class__` cell in our semantic model. Namely,

> __class__ is an implicit closure reference created by the compiler if any methods in a class body refer to either __class__ or super.

from the Python [docs](https://docs.python.org/3/reference/datamodel.html#creating-the-class-object).

As noted in the code comment by @AlexWaygood, we don't fully model this behavior, opting always to create the `__class__` cell binding in a new `ScopeKind::ClassCell` around each method definition, without checking if any method in the class body actually refers to `__class__` or `super`.

As such, this PR may only help with #18442 and not #19357.

Test Plan
--

Existing tests, plus (TODO) tests from #19783 and #19424, which tackled the issues above on a per-rule basis.


crates/ruff_python_semantic/src/binding.rs
```diff
@@ -446,11 +446,17 @@ impl Ranged for Binding<'_> {
...
 /// bindings because bindings must be separated by whitespace (and have an assignment).
 #[newtype_index]
 pub struct BindingId;
 
+impl BindingId {
+    pub const fn dunder_class_id() -> Self {
+        BindingId::from_u32(u32::MAX)
+    }
+}
+
```
AlexWaygood: oops, I think this is unused and can go (leftover from an earlier version of the PR!)

```suggestion
```

ntBre:
I pulled in the tests from #19783 (and added @mikeleppane as a co-author, thank you!) and the issue with those rules is now resolved, with one additional change to the semantic model to reuse the cell binding for `__class__` in `add_binding` (c1604dd1c4).

I also tried the tests from #19424, but I think these still need some rule-specific code, so I decided to leave that as a separate PR. The changes here should still help, but they aren't sufficient on their own.

Finally, I applied our `class_variables_visible` change in a couple more places. Those were some of the last references to `"__class__"` that I found from grepping.

crates/ruff_python_semantic/src/scope.rs
```diff
@@ -169,6 +169,7 @@ bitflags! {
 #[derive(Debug, is_macro::Is)]
 pub enum ScopeKind<'a> {
     Class(&'a ast::StmtClassDef),
+    ClassCell,
```
AlexWaygood: a couple of nits:
- I think maybe `DunderClassCell` might be a better name for the variant? `ClassCell` might imply that it's something to do with classes, which it _is_, but we're really using "class" here to refer to the dunder-class variable rather than the class itself
- I wonder if we should move the long comment we added in `crates/ruff_linter/src/checkers/ast/mod.rs` to be a doc-comment above this enum variant, and add a new comment where the long comment currently is that just says to see the doc-comment for this enum variant to find out what this `DunderClassCell` thing is. That way I'll see the long comment if I'm hovering over the variant in VSCode. I think I'd also expect to see some docs here if I clicked go-to definition in my IDE.

crates/ruff_python_semantic/src/model.rs
```diff
@@ -442,7 +431,11 @@ impl<'a> SemanticModel<'a> {
             // Allow class variables to be visible for an additional scope level
             // when a type scope is seen — this covers the type scope present between
             // function and class definitions and their parent class scope.
-            class_variables_visible = scope.kind.is_type() && index == 0;
+            //
+            // Also allow an additional level beyond that to cover the implicit
+            // `__class__` closure created around methods and enclosing the type scope.
+            class_variables_visible =
+                (scope.kind.is_type() && index == 0) || (scope.kind.is_class_cell() && index == 1);
```
AlexWaygood: I guess we could also write this as

```suggestion
            class_variables_visible =
                matches!((scope.kind, index), (ScopeKind::Type, 0) | (ScopeKind::ClassCell, 1));
```

I'm happy with either! (If you do update this, it would be good to update the other occurence in this file too)

crates/ruff_python_semantic/src/binding.rs
```diff
@@ -672,6 +672,24 @@ pub enum BindingKind<'a> {
...
+    ///     def f():
+    ///         print(__class__)  # allowed
+    ///
+    ///     def g():
+    ///         nonlocal __class__  # also allowed because the scope is *not* the function scope
+    /// ```
+    ///
+    /// See <https://docs.python.org/3/reference/datamodel.html#creating-the-class-object> for more
+    /// details.
+    ClassCell,
```
AlexWaygood: I guess I'd prefer `DunderClassCell` as the name for this variant too

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -2421,6 +2484,15 @@ impl<'a> Checker<'a> {
...
                 .unwrap_or(self.semantic.scope_id)
+        }
+        // Similarly, `__class__` references bind to the nearest enclosing implicit closure added
+        // around each method definition.
+        else if name == "__class__" {
+            self.semantic
+                .scopes
+                .ancestor_ids(self.semantic.scope_id)
+                .find_or_last(|scope_id| self.semantic.scopes[*scope_id].kind.is_class_cell())
+                .unwrap_or(self.semantic.scope_id)
```
AlexWaygood: I know that if you remove this branch of code, we get a lot more F841 errors in the snapshot you added saying that various `__class__` definitions are assigned to but never used. But I actually feel like most of those diagnostics are desirable? I know that there's an example in https://github.com/astral-sh/ruff/issues/18442#issue-3114385820 of where assigning to `__class__` in one method will change the value of `__class__` when it's accessed in a very specific way from another method. But that honestly seems like extremely far-fetched code that neither a human nor an LLM would ever write. To me a diagnostic telling me that the assignment to `__class__` in `F.method()` here is unused seems useful -- for all common ways of accessing `__class__`, mutations made in one method will not affect the value as seen by other methods:

```pycon
>>> class F:
...     def method(self):
...         __class__ = 42
...     def method2(self):
...         print(__class__)
...         
>>> f = F()
>>> f.method()
>>> f.method2()
<class '__main__.F'>
```

IMO, practicality beats purity here -- if there's a bug in F841 regarding `__class__`, it's only that it refers to it as a "local variable". We could consider making this edit to the error message if the variable name is `__class__` specifically:

```diff
- f841_ple0117.py:14:9: F841 Local variable `__class__` is assigned to but never used
+ f841_ple0117.py:14:9: F841 Variable `__class__` is assigned to but never used
```

But even _that_ feels just as likely to cause confusion as it is to clarify things -- honestly, I'd be inclined to just leave F841 as it is (but add more tests for how it behaves with `__class__`, and ideally add some comments to those tests!)

crates/ruff_python_semantic/src/model.rs
ntBre: Yeah that seems a little nicer, thanks!

crates/ruff_linter/src/checkers/ast/mod.rs
ntBre: That makes sense to me! I didn't really look closely at the `__getattribute__` stuff in the issue until you mentioned it, that does seem pretty uncommon! I can comment there after we merge this and close the issue.

Are there any other tests you would add, or are the ones from #19783 enough once I add some comments?

AlexWaygood: Those tests LGTM! We can always add some more if more issues come up in the future. The big user-facing improvement here to me is that we no longer think this is a syntax error, which it obviously would be for any non-`__class__` variable:

```py
class F:
    def method(self):
        nonlocal __class__
```

To me that seems like a somewhat serious bug on `main`, because obviously syntax errors are unsuppressable!

And I think the fact that modelling it as a separate scope (emulating the semantics at runtime precisely) means that we can remove special-casing for `__class__` elsewhere is another great reason to make this change.

ntBre: I wrote a comment on this case to explain why we don't flag it:

```py
class A:
    __class__ = 1
```

I said that it was a normal class variable, not the special `__class__` cell, but maybe that's not the case:

```pycon
>>> class A:
...     __class__ = 1
...
>>> A.__class__
<class 'type'>
```

Should we be flagging that as well? That might be moving beyond the scope of this PR, though.

AlexWaygood: I regret to inform you that there's sometimes a ~~cursed~~ "good" reason to assign to `__class__` in a class body:

https://github.com/python/typing_extensions/blob/d372913fd71f8f01008c20276fbfe01519ddf1df/src/typing_extensions.py#L1910-L1911

So... no, we shouldn't be flagging that 😆

AlexWaygood: Specifically:

```pycon
>>> class Foo:
...     __class__ = bool
...     
>>> Foo().__class__
<class 'bool'>
>>> isinstance(Foo(), bool)
True
```

AlexWaygood: I think it's some crazy descriptor magic that means that if you access it on the class object itself, it doesn't pay attention to the value you set in the class body, but if you access it on the instance it does

ntBre: That's good news to me! I think this PR is good to go then :laughing: Well, maybe "normal class variable" isn't exactly right still...

ntBre: I guess there's an implicit "for the purposes of F841" in an F841 test file, so it seems close enough.

crates/ruff_linter/resources/test/fixtures/pyflakes/F841_0.py
```diff
@@ -151,3 +151,32 @@ def f():
         pass
     except Exception as _:
         pass
+
+
+# OK -- `__class__` in this case is not the special `__class__` cell, it's just
+# a normal class variable, so we don't emit a diagnostic.
```
AlexWaygood:
```suggestion
# OK -- `__class__` in this case is not the special `__class__` cell,
# so we don't emit a diagnostic. (It has its own special semantics --
# see https://github.com/astral-sh/ruff/pull/20048#discussion_r2298338048 --
# but those aren't relevant here.)
```

crates/ruff_linter/resources/test/fixtures/pyflakes/F841_0.py
```diff
@@ -151,3 +151,32 @@ def f():
...
+class A:
+    __class__ = 1
+
+
+# The following three cases should technically be allowed because `__class__`
+# is a nonlocal variable in the method scope. However, `__class__` isn't easily
+# accessible in other methods, so setting it like this is still likely to be an
+# error, and we still emit a diagnostic.
```
AlexWaygood:
```suggestion
# error, and we still emit a diagnostic. See https://github.com/astral-sh/ruff/pull/20048#discussion_r2296252395
```

dscorbett:
This comment is not true:
```
# The following three cases should technically be allowed because `__class__`
# is a nonlocal variable in the method scope. However, `__class__` isn't easily
# accessible in other methods, so setting it like this is still likely to be an
# error, and we still emit a diagnostic. See https://github.com/astral-sh/ruff/pull/20048#discussion_r2296252395
class A:
    def set_class(self, cls):
        __class__ = cls
```

`__class__` is a local variable there because it is assigned. It would only be nonlocal with a `nonlocal __class__` statement.

ntBre:
Ah I see, thanks. For some reason I thought this should work even without the explicit `nonlocal`, but I think this resolves my confusion. This case is now fully resolved, not triggering PLE0117 or F841:

```py
class A:
    def set_class(self, cls):
        nonlocal __class__
        __class__ = cls
```

and these cases are properly flagged because the local `__class__` variable _is_ unused:

```py
class A:
    class B:
        def set_class(self, cls):
            __class__ = cls


class A:
    def foo():
        class B:
            print(__class__)
            def set_class(self, cls):
                __class__ = cls
```

I'll update the first test case with the actual `nonlocal` statement and also fix the comment.

dscorbett:
In the following example, `__class__` is a local variable, which breaks the zero-argument `super`.
```python
class C:
    def __str__(self):
        if False:
            __class__ = C
        return super().__str__()
print(str(C()))
```
With this PR, will that be handled properly?

AlexWaygood:
> With this PR, will that be handled properly?

No, that issue will still be outstanding after this PR has merged. (That doesn't mean that we won't necessarily fix it in the future. But while it's a valid bug report, it's probably low priority for us to fix it since it seems unlikely that a human or LLM would ever write that in real code :-)

ntBre:
I think https://github.com/astral-sh/ruff/pull/19424 will still be necessary to fix that, if that's the same issue. But the local variable should shadow the nonlocal binding from the cell after this PR, so I think it makes it slightly easier to resolve the other issue too, at least.

PR #20049, opened 2025-08-22
[flake8-use-pathlib] Fix `PTH211` autofix

chirizxc:


chirizxc:
@dylwil3 this [commit](https://github.com/astral-sh/ruff/pull/20009/commits/994f983d9643bbc3f3cc7bc03a56c396bab4998a) caused the misbehavior

chirizxc:
I think we can check if `target_is_directory` is actually a boolean before, as in this PR

Issue #17412, opened 2025-04-15
[syntax-errors] Remaining errors raised by the compiler


ntBre:
## Summary

A number of semantic or compile-time syntax errors have already been implemented
in #11934, but there are still many remaining ones. The purpose of this issue is
to track progress on this remainder. The initial source for most of these is
[this comment], which reports fuzzer results comparing Python's compiler and
ruff, but others may be included over time.

The next section lists the syntax errors that still need to be implemented, and
the sections after that include some more information on getting started with an
implementation.

### Errors mapping to current rules

These might be the easiest to start with because the existing rules have working
implementations and corresponding tests. However, they could possibly require
new context methods (see below), which could complicate things.

- `break` outside loop (F701)
- `continue` not properly in loop (F702)
- `return` with value in async generator (B901)
  - partial overlap with B901, which also applies to sync generators. This is
    only a syntax error in the async case
- `yield from` in async function (PLE1700)
- future feature is not defined (F407)
- import `*` only allowed at module level (F406)
- multiple starred expressions in assignment (F622)
- no binding for nonlocal (PLE0117)

### New errors

- generator expression must be parenthesized
``` python
"/".join(
    x for x in range(3), # this comma causes the issue
)
```
- can't use starred expression here
``` python
x = *value
```
- name is parameter and global
``` python
def f(a): global a
```
- name is parameter and nonlocal
``` python
def f(a): nonlocal a
```
- nonlocal declaration not allowed module level
``` python
nonlocal x
```
- not a chance (easter egg)
``` python
from __future__ import braces
```
- source code string cannot contain null bytes
```shell
printf '# \x00' > example.py
```
This mostly seems like an issue in comments

### Encoding issues

These syntax errors are all related to #6791 and should only be implemented if
we decide to respect these alternate encodings in general.

These can also be a bit tricky to put into a file, so most of these have shell
code snippets for generating the input rather than Python code directly.

- ascii codec can't decode byte `...` in position `...`: ordinal not in `range(128)`

```shell
printf '# -*- coding: ascii -*-\n\xc2' > example.py
```

- charmap codec can't decode byte ...

```shell
printf '# -*- coding: cp1252 -*-\n\x8d' > example.py
```

- utf7 codec can't decode byte 0xc3

```shell
printf '# -*- coding: utf-7 -*-\n\xc3' > example.py
```

This one reports an `E902` error, but only if it's enabled. Otherwise we only
print a warning about invalid UTF-8.

[...]

PR #20051, opened 2025-08-22
[syntax-errors] Detect `yield from` inside async function

11happy:
## Summary

This PR implements https://docs.astral.sh/ruff/rules/yield-from-in-async-function/  as a syntax semantic error

## Test Plan

I have written a simple inline test as directed in [https://github.com/astral-sh/ruff/issues/17412](https://github.com/astral-sh/ruff/issues/17412)

### DIFF
--- a/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__yield_from_in_async_function.py.snap
+++ b/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__yield_from_in_async_function.py.snap
@@ -1,0 +1,16 @@
+---
+source: crates/ruff_linter/src/linter.rs
+---
+PLE1700 `yield from` statement in async function; use `async for` instead
+ --> resources/test/fixtures/syntax_errors/yield_from_in_async_function.py:1:16
+  |
+1 | async def f(): yield from x # error
+  |                ^^^^^^^^^^^^
+  |
+
+PLE1700 `yield from` statement in async function; use `async for` instead
+ --> resources/test/fixtures/syntax_errors/yield_from_in_async_function.py:1:16
+  |
+1 | async def f(): yield from x # error
+  |                ^^^^^^^^^^^^
+  |
--- a/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__yield_from_in_async_function.py.snap
+++ b/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__yield_from_in_async_function.py.snap
@@ -7,10 +7,3 @@ PLE1700 `yield from` statement in async function; use `async for` instead
 1 | async def f(): yield from x # error
   |                ^^^^^^^^^^^^
   |
-
-PLE1700 `yield from` statement in async function; use `async for` instead
- --> resources/test/fixtures/syntax_errors/yield_from_in_async_function.py:1:16
-  |
-1 | async def f(): yield from x # error
-  |                ^^^^^^^^^^^^
-  |
--- a/crates/ruff_linter/resources/test/fixtures/syntax_errors/yield_from_in_async_function.py
+++ b/crates/ruff_linter/resources/test/fixtures/syntax_errors/yield_from_in_async_function.py
@@ -1,0 +1,1 @@
+async def f(): yield from x # error
--- a/crates/ruff_linter/src/checkers/ast/analyze/expression.rs
+++ b/crates/ruff_linter/src/checkers/ast/analyze/expression.rs
@@ -1319,12 +1319,9 @@ pub(crate) fn expression(expr: &Expr, checker: &Checker) {
                 pylint::rules::yield_in_init(checker, expr);
             }
         }
-        Expr::YieldFrom(yield_from) => {
+        Expr::YieldFrom(_) => {
             if checker.is_rule_enabled(Rule::YieldInInit) {
                 pylint::rules::yield_in_init(checker, expr);
-            }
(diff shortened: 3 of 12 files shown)


ntBre:
Thanks for working on this! It looks like there are a few unrelated test failures in the CI results. The inline tests also produce a couple of files that have to be checked in. Let me know if you need help resolving either of those issues!

ntBre:
(labeling this as `ty` since it shouldn't cause any external behavior to change in Ruff but will be a new syntax error detected in ty)

11happy:
Have updates to fix the CI, however this test is failing
```
thread 'valid_syntax' panicked at crates/ruff_python_parser/tests/fixtures.rs:119:9:
"/home/happy/ruff/crates/ruff_python_parser/resources/valid/expressions/arguments.py": Expected no semantic syntax errors for a valid program:
   |
29 | # Yield expression
30 | call((yield x))
31 | call((yield from x))
   |       ^^^^^^^^^^^^ Syntax Error: `yield from` statement in async function; use `async for` instead
32 |
33 | # Named expression
```
file https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_parser/resources/valid/expressions/arguments.py

11happy:
I have a couple of doubts :

first: while reviewing snapshots I found this 
```
        241 │+1 | x: *int = 1
        242 │+2 | x: yield a = 1
        243 │+3 | x: yield from b = 1
        244 │+  |    ^^^^^^^^^^^^ Syntax Error: `yield from` statement in async function; use `async for` instead
        245 │+4 | x: y := int = 1
```
Here this `yield from`  syntax error should not be there as its not in a async scope its a module level scope ?
for this I looked at the implementation of `in_async_context`  & I am not perfectly sure how's here the `in_async_context` is returning true, Am I missing something here?
```
fn in_async_context(&self) -> bool {
        for scope_info in self.scope_stack.iter().rev() {
            let scope = &self.scopes[scope_info.file_scope_id];
            match scope.kind() {
                ScopeKind::Class | ScopeKind::Lambda => return false,
                ScopeKind::Function => {
                    return scope.node().expect_function(self.module).is_async;
                }
                ScopeKind::Comprehension
                | ScopeKind::Module
                | ScopeKind::TypeAlias
                | ScopeKind::TypeParams => {}
            }
        }
        false
    }
```

second: should I commit all the updated snapshots ?

Thank you

11happy:
Hii @ntBre,
I’d really appreciate your help in moving this PR forward when you get a chance.
Thank you : )

ntBre:
I think the test failures are running into a limitation of our inline parser error test suite:

https://github.com/astral-sh/ruff/blob/32cb86351bfadda29a335d6d3aa6bc52585f4bdf/crates/ruff_python_parser/tests/fixtures.rs#L530-L532

You can see there that we've hard-coded `true` in `in_async_context`. We do have some handling for scopes, as you can see in the function just above:

https://github.com/astral-sh/ruff/blob/32cb86351bfadda29a335d6d3aa6bc52585f4bdf/crates/ruff_python_parser/tests/fixtures.rs#L534-L541

So I think you'll also need to add a real implementation of `in_async_context` to avoid the  unrelated test failures. Sorry about that, I guess this is the first rule that needs it. I'm happy to help with this if you run into trouble.

Along these lines, we usually add two other kinds of tests for these errors since they depend on ruff's and ty's implementations of `SemanticSyntaxContext`. You can see an example in  Ruff:

https://github.com/astral-sh/ruff/blob/32cb86351bfadda29a335d6d3aa6bc52585f4bdf/crates/ruff_linter/src/linter.rs#L1185-L1190

and in ty:

https://github.com/astral-sh/ruff/blob/32cb86351bfadda29a335d6d3aa6bc52585f4bdf/crates/ty_python_semantic/resources/mdtest/diagnostics/semantic_syntax_errors.md?plain=1#L17-L24

> second: should I commit all the updated snapshots ?

Yes, once you get a set of snapshot failures that look correct, you'll need to commit the updated snapshots to pass the tests in CI.

11happy:
@ntBre thank you for guiding this PR!

I have updated it accordingly , could you please take a look.
Thank you

ntBre:
I'll try to take a look soon! It looks like the representation of  AST nodes changed slightly last week (#20028), you might need to rebase/merge main and update your snapshots once more to get CI passing. There's also one small clippy suggestion.

11happy:
have rebased it, CI should pass now

11happy:
Updated the snapshot & fixed the clippy suggestion as per new CI results.

crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__yield_from_in_async_function.py.snap
ntBre, line 1: It looks like we're getting duplicate diagnostics. I think there should be some existing PLE1700 code that we need to delete now that it's being detected and reported as a syntax error:

https://github.com/astral-sh/ruff/blob/9b1b58a4513eae2c3cf206eb5fc3aea19380883c/crates/ruff_linter/src/rules/pylint/rules/yield_from_in_async_function.rs#L40-L42

We should be able to delete this function and any callers.

11happy: done

11happy:
Hi maintainers, please let me know if any further input or changes are required from my end for this PR.

Thank you

ntBre (approved):
Thank you!

PR #20052, opened 2025-08-22
[ty] Refactor inlay hints structure to use separate parts

MatthewMckee4:
## Summary

Our internal inlay hints structure (`ty_ide::InlayHint`) now more closely resembles `lsp_types::InlayHint`.

This mainly allows us to convert to `lsp_types::InlayHint` with less hassle, but it also allows us to manage the different parts of the inlay hint better, which in the future will allow us to implement features like goto on the type part of the type inlay hint.

It also really isn't important to store a specific `Type` instance in the `InlayHintContent`. So we remove this and use `InlayHintLabel` instead which just shows the representation of the type (along with other information). 

We see a similar structure used in rust-analyzer too.

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{fmt, vec};
 
 use crate::Db;
 use ruff_db::files::File;
@@ -18,11 +18,10 @@ pub struct InlayHint {
 
 impl InlayHint {
     fn type_hint(position: TextSize, ty: Type, db: &dyn Db) -> Self {
-        let mut label_parts = Vec::new();
-
-        label_parts.push(": ".into());
-
-        label_parts.push(InlayHintLabelPart::new(ty.display(db).to_string()));
+        let label_parts = vec![
+            InlayHintLabelPart::new(": ".into()),
+            InlayHintLabelPart::new(ty.display(db).to_string()),
+        ];
 
         Self {
             position,
@@ -32,11 +31,10 @@ impl InlayHint {
     }
 
     fn call_argument_name(position: TextSize, name: &str) -> Self {
-        let mut label_parts = Vec::new();
-
-        label_parts.push(InlayHintLabelPart::new(name.into()));
-
-        label_parts.push("=".into());
+        let label_parts = vec![
+            InlayHintLabelPart::new(name.into()),
+            InlayHintLabelPart::new("=".into()),
+        ];
 
         Self {
             position,
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -15,7 +15,7 @@ pub struct InlayHint {
 }
 
 impl InlayHint {
-    pub fn type_hint<'db>(position: TextSize, ty: Type, db: &'db dyn Db) -> Self {
+    pub fn type_hint(position: TextSize, ty: Type, db: &dyn Db) -> Self {
         Self {
             position,
             kind: InlayHintKind::Type,
@@ -58,8 +58,8 @@ impl From<String> for InlayHintLabel {
     }
 }
 
-pub fn inlay_hints<'db>(
-    db: &'db dyn Db,
+pub fn inlay_hints(
+    db: &dyn Db,
     file: File,
     range: TextRange,
     settings: &InlayHintSettings,
@@ -138,7 +138,7 @@ impl<'a, 'db> InlayHintVisitor<'a, 'db> {
         self.hints.push(InlayHint::type_hint(position, ty, self.db));
     }
 
-    fn add_call_argument_name(&mut self, position: TextSize, name: String) {
+    fn add_call_argument_name(&mut self, position: TextSize, name: &str) {
         if !self.settings.call_argument_names {
             return;
         }
@@ -148,7 +148,7 @@ impl<'a, 'db> InlayHintVisitor<'a, 'db> {
         }
 
         self.hints
-            .push(InlayHint::call_argument_name(position, &name));
+            .push(InlayHint::call_argument_name(position, name));
     }
 }
 
@@ -214,10 +214,7 @@ impl SourceOrderVisitor<'_> for InlayHintVisitor<'_, '_> {
 
                 for (index, arg_or_keyword) in call.arguments.arguments_source_order().enumerate() {
                     if let Some(name) = argument_names.get(&index) {
-                        self.add_call_argument_name(
-                            arg_or_keyword.range().start(),
-                            name.to_string(),
-                        );
+                        self.add_call_argument_name(arg_or_keyword.range().start(), name);
                     }
                     self.visit_expr(arg_or_keyword.value());
                 }
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -1,9 +1,12 @@
+use std::fmt;
+
 use crate::Db;
 use ruff_db::files::File;
 use ruff_db::parsed::parsed_module;
 use ruff_python_ast::visitor::source_order::{self, SourceOrderVisitor, TraversalSignal};
 use ruff_python_ast::{AnyNodeRef, Expr, Stmt};
 use ruff_text_size::{Ranged, TextRange, TextSize};
+use smallvec::{SmallVec, smallvec};
 use ty_python_semantic::types::{Type, inlay_hint_function_argument_details};
 use ty_python_semantic::{HasType, SemanticModel};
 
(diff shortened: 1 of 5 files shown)


crates/ty_server/tests/e2e/inlay_hints.rs
```diff
@@ -42,15 +42,23 @@ foo(1)
           "line": 0,
           "character": 1
         },
-        "label": ": Literal[1]",
+        "label": [
```
MatthewMckee4, line 45: These change because we now use `InlayHintLabel::LabelParts` instead of `InlayHintLabel::String`

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,140 @@
...
 }
 
-impl<'db> InlayHint<'db> {
-    pub const fn display(&self, db: &'db dyn Db) -> DisplayInlayHint<'_, 'db> {
-        self.content.display(db)
+impl InlayHint {
+    pub fn type_hint(position: TextSize, ty: Type, db: &dyn Db) -> Self {
+        let mut label = InlayHintLabel::simple(ty.display(db).to_string(), None);
+
+        label.prepend_str(": ");
```
MatthewMckee4: This is roughly how rust analyzer does it and we need to separate these for the goto targets

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,140 @@
...
-    CallArgumentName(String),
+#[derive(Debug, Clone)]
+pub enum InlayHintKind {
+    Type,
+    CallArgumentName,
+}
+
+#[derive(Debug, Clone)]
+pub struct InlayHintLabel {
+    pub parts: SmallVec<[InlayHintLabelPart; 1]>,
```
MatthewMckee4: I think i should change this to 2. @MichaReiser would you agree?

MatthewMckee4: ive done it, i think it makes sense for us.

MatthewMckee4: ah okay in this situation we only use 1 of the `InlayHintLabelPart`, but when we add the goto, we will use 2.

MichaReiser:
Could you extend the summary? I think I've an idea why we may want this but I'd rather be sure (and documenting this will also help a future reader who comes across this PR and wonders why it works the way it does)

crates/ty_ide/src/inlay_hints.rs
MichaReiser: I think it makes sense today because all our usages have exactly two parts. But it's hard to tell, depending on how you plan to use this going forward. 

A too-large number has the downside that `InlayHintLabel` becomes larger overall, wasting memory when we only need to store 1 or more than two elements. This is especially problematic if we create many `InlayHintLabel`s. I don't think this is a big concern here.

I'm not even sure if it's worth using a `SmallVec` here. We'll need to serialize all that data later on, we might as well just store a Vec

MatthewMckee4: Currently because we don't use the `target` variable in `InlayHintLabel`. This means we just have one `part` so we will only use one of the elements in the small vec.

I'll change to vec for now anyway.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
-    pub content: InlayHintContent<'db>,
+    pub kind: InlayHintKind,
+    pub label: InlayHintLabel,
 }
 
-impl<'db> InlayHint<'db> {
-    pub const fn display(&self, db: &'db dyn Db) -> DisplayInlayHint<'_, 'db> {
-        self.content.display(db)
+impl InlayHint {
+    pub fn type_hint(position: TextSize, ty: Type, db: &dyn Db) -> Self {
```
dhruvmanila: I think it'd be useful to add basic documentation for all the `pub` items in this module.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
+    CallArgumentName,
+}
+
+#[derive(Debug, Clone)]
+pub struct InlayHintLabel {
+    pub parts: Vec<InlayHintLabelPart>,
+}
+
+impl InlayHintLabel {
+    pub fn simple(s: impl Into<String>, target: Option<crate::NavigationTarget>) -> InlayHintLabel {
```
dhruvmanila: I think `simple` sounds confusing, what does simple mean in the context of `InlayHintLabel`? Based on my other suggestion to introduce `with_target` for `InlayHintLabelPart`, we could rename this to `single(part: InlayHintLabelPart)` and let the caller construct the `InlayHintLabelPart` or just a `new(parts: Vec<InlayHintLabelPart>`) as `InlayHintLabelPart` is a public type.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
+#[derive(Debug, Clone)]
+pub enum InlayHintKind {
+    Type,
+    CallArgumentName,
+}
+
+#[derive(Debug, Clone)]
+pub struct InlayHintLabel {
+    pub parts: Vec<InlayHintLabelPart>,
+}
```
dhruvmanila: Are you thinking of expanding `InlayHintLabel` with additional fields in the future?

Regardless, I think the `parts` field should be private and we should expose this information via a method that returns `&[InlayHintLabelPart]` instead.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
+                text: s.into(),
+                target,
+            }],
+        }
+    }
+
+    pub fn prepend_str(&mut self, s: &str) {
+        match &mut *self.parts {
+            [InlayHintLabelPart { text, target: None }, ..] => text.insert_str(0, s),
+            _ => self.parts.insert(
```
dhruvmanila: At first, I thought that the catch-all branch is for the empty vector (`[]`) but I don't think that's true, the branch that isn't covered here is an `InlayHintLabelPart` where `target` is `Some`. Can you expand on why does the string needs to be added only when navigation target is not present? Based on the method name, I would expect this to add the string regardless of whether target is present or not.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
+                    target: None,
+                },
+            ),
+        }
+    }
+
+    pub fn append_str(&mut self, s: &str) {
+        match &mut *self.parts {
+            [.., InlayHintLabelPart { text, target: None }] => text.push_str(s),
+            _ => self.parts.push(InlayHintLabelPart {
```
dhruvmanila: Same question as above, why should this append string only when navigation target is not present?

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
 }
 
-pub fn inlay_hints<'db>(
-    db: &'db dyn Db,
+#[derive(Default, Debug, Clone)]
+pub struct InlayHintLabelPart {
+    pub text: String,
+
+    pub target: Option<crate::NavigationTarget>,
+}
```
dhruvmanila: I would recommend to keep the field private and expose methods to provide the information.

Additionally, I think we should have a `From<&str>` / `From<String>` / `new(text: String)` implementation for `InlayHintLabelPart` and use builder pattern for the target like `with_target` because it's optional.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -1,62 +1,139 @@
...
+        }
     }
 }
 
-pub struct DisplayInlayHint<'a, 'db> {
-    db: &'db dyn Db,
-    hint: &'a InlayHintContent<'db>,
+pub struct InlayHintDisplay<'a> {
+    pub inlay_hint: &'a InlayHint,
 }
```
dhruvmanila: Let's keep the `inlay_hint` field private, I don't think we need to expose it publicly.

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: I'll try to make more things private

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: going to leave the `with_target` method for now as we don't use it yet.

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: Ive just removed these methods, i think its better for us to just be more explicit with what the specific parts are anyway.

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: I have removed this too.

dhruvmanila:
Feel free to re-request for review once it's ready.

MatthewMckee4:
could you re run that cargo test, looks like a flake

crates/ty_ide/src/inlay_hints.rs
dhruvmanila: Is this because that would allow us to have the navigation target that is relevant for that specific part of the label? For example, in `: Type`, it's just the "Type" part that's possible to do go to definition.

dhruvmanila (approved):
Thanks!

dhruvmanila:
> could you re run that cargo test, looks like a flake

I'm assuming that this was fixed?

PR #20053, opened 2025-08-22
[ty] Remove duplicate global lint registry

ibraheemdev:
## Summary

Looks like an oversight at some point that led to two identical globals, the one in `ty_project` just calls `ty_python_semantic::register_lints`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2935,6 +2935,7 @@ dependencies = [
  "tracing-subscriber",
  "ty",
  "ty_project",
+ "ty_python_semantic",
  "ty_static",
  "url",
 ]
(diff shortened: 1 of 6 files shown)


carljm (approved):
Thanks!

MichaReiser:
That was sort of intentional to already have an extension point where we register lints that aren't from ty_python_semantic. Similar to how each crate has their own db. For example, the project crate will need to register all rules from the linter crate

carljm:
> That was sort of intentional to already have an extension point where we register lints that aren't from ty_python_semantic. Similar to how each crate has their own db. For example, the project crate will need to register all rules from the linter crate

Oops. I would have left this one for you to review if you weren't out :)

That said, I think there's an argument for not leaving around confusing unused things -- we can easily add this back in when we actually have a use for it.

PR #20055, opened 2025-08-23
[ty] don't eagerly unpack aliases in user-authored unions

carljm:
## Summary

Add a subtly different test case for recursive PEP 695 type aliases, which does require that we relax our union simplification, so we don't eagerly unpack aliases from user-provided union annotations.

## Test Plan

Added mdtest.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -239,11 +239,25 @@ def h(x: Intersection[A, B]):
 
 Adapted from <https://github.com/pypa/packaging/blob/main/src/packaging/_parser.py>:
 
+#### With old-style union
+
 ```py
 from typing import Union
 
 type MarkerAtom = Union[int, list["MarkerAtom"]]
 type MarkerList = list[Union["MarkerList", MarkerAtom, str]]
+
+def f(marker_list: MarkerList):
+    reveal_type(marker_list)  # revealed: list[MarkerList | MarkerAtom | str]
+    for item in marker_list:
+        reveal_type(item)  # revealed: list[MarkerList | MarkerAtom | str] | int | list[MarkerAtom] | str
+```
+
+#### With new-style union
+
+```py
+type MarkerAtom = int | list["MarkerAtom"]
+type MarkerList = list["MarkerList" | MarkerAtom | str]
 
 def f(marker_list: MarkerList):
     reveal_type(marker_list)  # revealed: list[MarkerList | MarkerAtom | str]
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -204,13 +204,17 @@ def f(x: OptNestedInt) -> None:
 ### Invalid self-referential
 
 ```py
-# TODO emit a diagnostic here
+# TODO emit a diagnostic on these two lines
 type IntOr = int | IntOr
+type OrInt = OrInt | int
 
-def f(x: IntOr):
+def f(x: IntOr, y: OrInt):
     reveal_type(x)  # revealed: int
+    reveal_type(y)  # revealed: int
     if not isinstance(x, int):
         reveal_type(x)  # revealed: Never
+    if not isinstance(y, int):
+        reveal_type(y)  # revealed: Never
 ```
 
 ### Mutually recursive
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -208,3 +208,19 @@ from ty_extensions import Intersection
...
+```py
+from typing import Union
+
+type MarkerAtom = Union[int, list["MarkerAtom"]]
+type MarkerList = list[Union["MarkerList", MarkerAtom, str]]
+
+def f(marker_list: MarkerList):
+    reveal_type(marker_list)  # revealed: list[MarkerList | MarkerAtom | str]
+    for item in marker_list:
+        reveal_type(item)  # revealed: list[MarkerList | MarkerAtom | str] | int | list[MarkerAtom] | str
```
AlexWaygood: This sort-of feels like a bit of a confusing test to me, because it's not really something anybody would ever write. If you're able to use PEP-695 syntax in your Python code, then you'd surely also use PEP-604 syntax. And if you're using PEP-695 syntax, you'd also surely make use of the fact that the r.h.s. of a PEP-695 alias is lazily evaluated (no need to quote any forward references).

```suggestion
type MarkerAtom = int | list[MarkerAtom]
type MarkerList = list[MarkerList | MarkerAtom | str]

def f(marker_list: MarkerList):
    reveal_type(marker_list)  # revealed: list[MarkerList | MarkerAtom | str]
    for item in marker_list:
        reveal_type(item)  # revealed: list[MarkerList | MarkerAtom | str] | int | list[MarkerAtom] | str
```

But it seems like we still panic on ^that rewritten test on this branch. (It looks like it's the switch to PEP-604 unions that makes us panic; we still run fine on that snippet if all I do is unquote the forward references.)

I'm okay with landing this test as-is (or with just the forward references unquoted) since it definitely demonstrates us making progress towards the goal of full support for recursive type aliases. But it might be worth adding some prose that says why it is the way it is?

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -204,19 +204,39 @@ enum ReduceResult<'db> {
...
+            simplification_strategy: SimplificationStrategy::Full,
         }
     }
 
+    pub(crate) fn minimal_simplify(mut self) -> Self {
+        self.simplification_strategy = SimplificationStrategy::Minimal;
+        self
+    }
+
+    fn should_simplify_full(&self) -> bool {
```
AlexWaygood: don't think it really makes much of a difference, but I'd always do this where it's possible

```suggestion
    const fn should_simplify_full(&self) -> bool {
```

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -10492,7 +10492,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
             }
             SpecialFormType::Union => match arguments_slice {
                 ast::Expr::Tuple(t) => {
-                    let union_ty = UnionType::from_elements(
+                    let union_ty = UnionType::from_elements_minimal_simplify(
                         db,
                         t.iter().map(|elt| self.infer_type_expression(elt)),
                     );
```
AlexWaygood: I think you also need to add this handling to `Optional` -- this test case still causes us to panic on your branch:

```py
from typing import Optional, Union

type MarkerAtom = Optional[list["MarkerAtom"]]
type MarkerList = list[Optional[Union["MarkerList", MarkerAtom, str]]]

def f(marker_list: MarkerList):
    reveal_type(marker_list)
    for item in marker_list:
        reveal_type(item)
```

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -204,19 +204,39 @@ enum ReduceResult<'db> {
 // representations, so that we can make large unions of literals fast in all operations.
 const MAX_UNION_LITERALS: usize = 200;
 
+enum SimplificationStrategy {
```
AlexWaygood:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
enum SimplificationStrategy {
```

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
carljm: Ugh, silly oversight -- I just always need to write tests for all forms of the syntax. Good catch! No, I'll fix this in this PR. But I want to include both tests, even the "weird" one -- this is fixing things that will be needed for other kinds of type aliases, too, but I want it tested for all combinations, even the weird ones.

PR #20056, opened 2025-08-23
[`flake8-simplify`] Fix incorrect fix for positive `maxsplit` without separator (`SIM905`)

### LINKED ISSUE #20033: SIM905 fix counts splits wrong for positive `maxsplit` and unspecified `sep`
### Summary

The fix for [`split-static-string` (SIM905)](https://docs.astral.sh/ruff/rules/split-static-string/) with positive `maxsplit` and unspecified `sep` counts the splits on individual white space characters. This can change the program’s behavior. It should count the splits on white space sequences. [Example](https://play.ruff.rs/273894ac-88a5-4c19-8285-68adcf6f77c2):
```console
$ cat >sim905.py <<'# EOF'
print("a  b".split(maxsplit=1))
# EOF

$ python sim905.py
['a', 'b']

$ ruff --isolated check sim905.py --select SIM905 --preview --fix

$ python sim905.py
['a', ' b']
```

### Version

RazerM:
## Summary

Resolves #20033

## Test Plan

unit tests added to the new split function, existing snapshot test updated.


crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
```diff
@@ -352,3 +306,115 @@ const fn py_unicode_is_whitespace(ch: char) -> bool {
...
+                self.remaining = self.remaining[end..].trim_start_matches(py_unicode_is_whitespace);
+                self.splits += 1;
+
+                Some(s)
+            }
+            Direction::Right => {
+                let start = self
+                    .remaining
+                    .rfind(py_unicode_is_whitespace)
+                    .map(|pos| pos + 1)
```
ntBre: I don't think this is safe to do. `rfind` returns the byte offset of the first character of the `match`. I tried a case like this:

```rust
const fn py_unicode_is_whitespace(ch: char) -> bool {
    matches!(ch, '\u{3000}')
}

fn main() {
    let text = "\u{3000}next";
    let start = text.rfind(py_unicode_is_whitespace).unwrap();
    dbg!(&text[start + 1..]);
}
```

in the Rust [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1b8ef1c7637312c09da7ed90f5bb40f5), and it panics with a `byte index is not a char boundary error`. This would be the case for any of the multi-byte whitespace characters in our real `py_unicode_is_whitespace` function.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
```diff
@@ -180,65 +180,19 @@ fn split_default(
...
-            } else {
-                let mut items: Vec<&str> = string_val
-                    .trim_end_matches(py_unicode_is_whitespace)
-                    .rsplitn(max_split + 1, py_unicode_is_whitespace)
-                    .filter(|s| !s.is_empty())
-                    .collect();
-                items.reverse();
-                items
-            };
+            let list_items = split_whitespace_with_maxsplit(string_val, max_split, direction);
```
ntBre: I'm kind of on the fence here because the iterator is cool, but I'd also rather not rewrite the `Ordering::Equal` case, which was already working and pretty simple to begin with. Would something like `string_val.split(py_unicode_is_whitespace).filter(|s| !s.is_empty()).take(max_split + 1)` and then manually joining the rest of the iterator as the remainder work? I think `itertools` should provide a `join` method we can use on an iterator of `&str` items. It's a bit wasteful because that will allocate a `String`, but it seems much simpler if it works.

Then we could also restore the TODO about using [`std::str::Split::remainder`](https://doc.rust-lang.org/std/str/struct.Split.html#method.remainder) in the future, which would be the ideal solution instead of joining them ourselves.

ntBre (commented):
Thanks for jumping on this!

I'm kind of leaning away from defining our own iterator, if we can avoid it.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
RazerM: We can't join the remaining parts because the character that was split is lost.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
RazerM: Right, I need to increment by `char::len_utf8`.

RazerM:
> I'm kind of leaning away from defining our own iterator, if we can avoid it.

Depending how you feel about it we could just remove the fix, since it wasn't spotted in #19851 by me or in review. The iterator is a bit verbose but I think the logic is ok.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
RazerM: > I'd also rather not rewrite the Ordering::Equal case, which was already working and pretty simple to begin with.

I almost left it as is, but then I thought for maxsplit=0 the iterator looks trivially correct: initial trim, then yields at most 1 item if non-empty.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
```diff
@@ -352,3 +306,118 @@ const fn py_unicode_is_whitespace(ch: char) -> bool {
...
+                    .map(|(pos, c)| pos + c.len_utf8())
+                    .unwrap_or(0);
+
+                let s = &self.remaining[start..];
+                self.remaining = self.remaining[..start].trim_end_matches(py_unicode_is_whitespace);
+                self.splits += 1;
+
+                Some(s)
+            }
+        }
```
ntBre: What do you think about something like this? I think this accomplishes the same thing and we can avoid having to do any index arithmetic, if I understand correctly.


```suggestion
        self.splits += 1;
        match self.direction {
            Direction::Left => match self.remaining.split_once(py_unicode_is_whitespace) {
                Some((s, remaining)) => {
                    self.remaining = remaining.trim_start_matches(py_unicode_is_whitespace);
                    Some(s)
                }
                None => Some(std::mem::take(&mut self.remaining)),
            },
            Direction::Right => match self.remaining.rsplit_once(py_unicode_is_whitespace) {
                Some((remaining, s)) => {
                    self.remaining = remaining.trim_end_matches(py_unicode_is_whitespace);
                    Some(s)
                }
                None => Some(std::mem::take(&mut self.remaining)),
            },
        }
```

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
ntBre: Ah you're right, I see now!

ntBre (commented):
Thanks! I think you're right, I was being overly scared of the iterator. This looks good to me overall, just one idea for simplifying a bit by using `(r)split_once`. It seemed to work well for me locally, but let me know if I'm still missing something.

crates/ruff_linter/src/rules/flake8_simplify/rules/split_static_string.rs
RazerM: `split_once` doesn't work as it only splits one matching character instead of all consecutive whitespace

ntBre: That's why I kept the trim calls afterward. I believe I tried this patch locally, and it passed all of the tests in the PR. But I could still be wrong, I'd welcome a test case with a counterexample!

RazerM: ah, sorry for not looking more closely, I'll take a deeper look

RazerM: Works well indeed, and is much more elegant than the index twiddling. I've rebased to fix the conflicts and included your suggestion

ntBre:
I took care of the merge conflicts. It looks like `Direction` mostly just got renamed to `Method` in #20459.

PR #20077, opened 2025-08-25
[`fastapi`] Fix false positive for paths with spaces around parameters (`FAST003`)

### LINKED ISSUE #20060: FAST003 false positive for parameters with spaces like `/{ x }`
### Summary

[`fast-api-unused-path-parameter` (FAST003)](https://docs.astral.sh/ruff/rules/fast-api-unused-path-parameter/) trims spaces from parameters in a path, treating e.g. `/{ x }` as `/{x}`, but this leads to false positives because [FastAPI does not recognize spaces around parameters](https://github.com/encode/starlette/blob/0.47.3/starlette/routing.py#L121). FAST003 should only apply when FastAPI would detect a parameter. [The following example](https://play.ruff.rs/f520e4b9-c4ee-4627-9e9d-56bfd6e5aed2) shows that FastAPI treats `/{ x }` literally and that the function parameter introduced by FAST003’s fix is detected as a query parameter, not as part of the path.

```console
$ cat >fast003.py <<'# EOF'
from fastapi import FastAPI
app = FastAPI()
@app.get("/{ x }")
async def f():
    return {"x": x}
# EOF

$ pip install -q 'fastapi[standard]'

$ fastapi dev fast003.py >/dev/null 2>&1

$ curl http://127.0.0.1:8000/foo
{"detail":"Not Found"}

$ curl 'http://127.0.0.1:8000/%7B%20x%20%7D'
Internal Server Error

$ ruff --isolated check fast003.py --select FAST003 --unsafe-fixes --fix

$ fastapi dev fast003.py >/dev/null 2>&1

$ curl http://127.0.0.1:8000/foo
{"detail":"Not Found"}

$ curl 'http://127.0.0.1:8000/%7B%20x%20%7D'
{"detail":[{"type":"missing","loc":["query","x"],"msg":"Field required","input":null}]}

$ curl 'http://127.0.0.1:8000/%7B%20x%20%7D?x=1'
{"x":"1"}
```

### Version

danparizher:
## Summary

Fixes #20060

### DIFF
--- a/crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
+++ b/crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
@@ -457,6 +457,9 @@ fn parameter_alias<'a>(parameter: &'a Parameter, semantic: &SemanticModel) -> Op
 ///
 /// The iterator yields tuples of the parameter name and the range of the parameter in the input,
 /// inclusive of curly braces.
+///
+/// FastAPI only recognizes path parameters when there are no leading or trailing spaces around
+/// the parameter name. For example, `/{x}` is a valid parameter, but `/{ x }` is treated literally.
 #[derive(Debug)]
 struct PathParamIterator<'a> {
     input: &'a str,
@@ -483,10 +486,18 @@ impl<'a> Iterator for PathParamIterator<'a> {
                     // We ignore text after a colon, since those are path converters
                     // See also: https://fastapi.tiangolo.com/tutorial/path-params/?h=path#path-convertor
                     let param_name_end = param_content.find(':').unwrap_or(param_content.len());
-                    let param_name = &param_content[..param_name_end].trim();
+                    let param_name = &param_content[..param_name_end];
 
-                    #[expect(clippy::range_plus_one)]
-                    return Some((param_name, start..end + 1));
+                    if param_name.starts_with(' ') || param_name.ends_with(' ') {
+                        continue;
+                    }
+
+                    let trimmed_param = param_name.trim();
+
+                    if !trimmed_param.is_empty() && is_identifier(trimmed_param) {
+                        #[expect(clippy::range_plus_one)]
+                        return Some((trimmed_param, start..end + 1));
+                    }
                 }
             }
         }
--- a/crates/ruff_linter/resources/test/fixtures/fastapi/FAST003.py
+++ b/crates/ruff_linter/resources/test/fixtures/fastapi/FAST003.py
@@ -213,3 +213,17 @@ async def get_id_pydantic_full(
 async def get_id_pydantic_short(params: Annotated[PydanticParams, Depends()]): ...
 @app.get("/{my_id}")
 async def get_id_init_not_annotated(params = Depends(InitParams)): ...
+
+@app.get("/things/{ thing_id }")
+async def read_thing(query: str):
+    return {"query": query}
+
+
+@app.get("/things/{ thing_id : path }")
+async def read_thing(query: str):
+    return {"query": query}
+
+
+@app.get("/things/{ thing_id : str }")
+async def read_thing(query: str):
+    return {"query": query}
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
```diff
@@ -483,10 +486,18 @@ impl<'a> Iterator for PathParamIterator<'a> {
...
                     // See also: https://fastapi.tiangolo.com/tutorial/path-params/?h=path#path-convertor
                     let param_name_end = param_content.find(':').unwrap_or(param_content.len());
-                    let param_name = &param_content[..param_name_end].trim();
+                    let param_name = &param_content[..param_name_end];
+
+                    if param_name.starts_with(' ') || param_name.ends_with(' ') {
+                        continue;
+                    }
+
+                    let trimmed_param = param_name.trim();
```
dylwil3: I don't think this is correct - trim removes _any_ unicode whitespace, not just ordinary spaces. Why not just remove `.trim` and leave everything else the same?

crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
```diff
@@ -483,10 +486,18 @@ impl<'a> Iterator for PathParamIterator<'a> {
...
 
-                    #[expect(clippy::range_plus_one)]
-                    return Some((param_name, start..end + 1));
+                    if !trimmed_param.is_empty() && is_identifier(trimmed_param) {
```
dylwil3: I think this check already occurs elsewhere https://github.com/astral-sh/ruff/blob/376e3ff39519d3c0563dd8f7a1d0daf9db52e98a/crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs#L167-L172

dscorbett:
I recommend making FAST003 only apply to [parameters as recognized by FastAPI](https://github.com/encode/starlette/blob/0.47.3/starlette/routing.py#L121), instead of just checking for leading and trailing ASCII spaces.

dylwil3:
> I recommend making FAST003 only apply to [parameters as recognized by FastAPI](https://github.com/encode/starlette/blob/0.47.3/starlette/routing.py#L121), instead of just checking for leading and trailing ASCII spaces.

I think if we avoid the trim then this is basically true (modulo reserved keywords) since we check whether we have an identifier (as pointed out above). So I don't think we need to port the regex over.

dscorbett:
I don’t think it is true, even ignoring keywords, but I can open another issue later since it’s not directly related to this one.

crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
```diff
@@ -483,10 +486,16 @@ impl<'a> Iterator for PathParamIterator<'a> {
                     // We ignore text after a colon, since those are path converters
                     // See also: https://fastapi.tiangolo.com/tutorial/path-params/?h=path#path-convertor
                     let param_name_end = param_content.find(':').unwrap_or(param_content.len());
-                    let param_name = &param_content[..param_name_end].trim();
+                    let param_name = &param_content[..param_name_end];
+
+                    if param_name.starts_with(' ') || param_name.ends_with(' ') {
+                        continue;
+                    }
```
dylwil3: Looks like we forgot to remove this part

crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
```diff
@@ -483,10 +486,16 @@ impl<'a> Iterator for PathParamIterator<'a> {
...
 
-                    #[expect(clippy::range_plus_one)]
-                    return Some((param_name, start..end + 1));
+                    if !param_name.is_empty() {
+                        #[expect(clippy::range_plus_one)]
+                        return Some((param_name, start..end + 1));
+                    }
```
dylwil3: Do we need this?

PR #20078, opened 2025-08-25
[ty] Update mypy_primer, add egglog-python project

sharkdp:
Now that https://github.com/astral-sh/ruff/pull/20263 is merged, we can update mypy_primer and add the new `egglog-python` project to `good.txt`. The ecosystem-analyzer run shows that we now add 1,356 diagnostics (where we had over 5,000 previously, due to the unsupported project layout).

### DIFF
--- a/crates/ty_python_semantic/resources/primer/good.txt
+++ b/crates/ty_python_semantic/resources/primer/good.txt
@@ -39,6 +39,7 @@ django-stubs
 downforeveryone
 dragonchain
 dulwich
+egglog-python
 flake8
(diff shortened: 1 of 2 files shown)


sharkdp:
The large number of diagnostics on the new project (especially `unresolved-reference`) seems to be caused by a non working project setup. In particular, `from egglog import *` imports don't seem to work for examples and tests?

sharkdp:
With `--extra-search-path python`, the number of diagnostics is reduced from 4531 to 1345.

PR #20079, opened 2025-08-25
[ty] Completely ignore typeshed's stub for `Any`

AlexWaygood:
## Summary

`Any` is a class (on most versions of Python), but it is not a normal class. Its many differences to normal classes include:
- It cannot be instantiated
- It cannot be used as a metaclass
- It cannot be used in `isinstance()` checks
- `Any` in a type expression does not mean "any object whose `__class__` is either the class `Any` or a class that has `Any` in its MRO".

These many differences to normal classes result in us having lots of special-case branches just for `Any`. It seems cleaner here if we just override typeshed's stubs completely for `Any`, similar to what we do for `NamedTuple`. Excluding the change I've made here to disallow `Any` as the second argument to `isinstance()` (which is new functionality), this PR is a net reduction in code, and improves our understanding of the semantics of the symbol `typing.Any` itself at runtime.

## Test Plan

mdtests

### DIFF
--- a/crates/ty_python_semantic/src/types/narrow.rs
+++ b/crates/ty_python_semantic/src/types/narrow.rs
@@ -183,15 +183,7 @@ impl ClassInfoConstraintFunction {
 
         match classinfo {
             Type::TypeAlias(alias) => self.generate_constraint(db, alias.value_type(db)),
-            Type::ClassLiteral(class_literal) => {
-                // At runtime (on Python 3.11+), this will return `True` for classes that actually
-                // do inherit `typing.Any` and `False` otherwise. We could accurately model that?
-                if class_literal.is_known(db, KnownClass::Any) {
-                    None
-                } else {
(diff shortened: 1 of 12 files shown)


crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -183,15 +183,7 @@ impl ClassInfoConstraintFunction {
 
         match classinfo {
             Type::TypeAlias(alias) => self.generate_constraint(db, alias.value_type(db)),
-            Type::ClassLiteral(class_literal) => {
-                // At runtime (on Python 3.11+), this will return `True` for classes that actually
-                // do inherit `typing.Any` and `False` otherwise. We could accurately model that?
```
AlexWaygood, line 188: this comment was only accurate for `issubclass(x, Any)`, since `isinstance(x, Any)` always raises an exception

PR #20084, opened 2025-08-25
[ty] Add support for PEP 800

AlexWaygood:
## Summary

This PR adds support for [PEP 800](https://peps.python.org/pep-0800/). Since ty's implementation of disjointness was the original inspiration for the PEP, nearly everything was already implemented; this PR mainly consists of renaming internal concepts to match up with the terms used by the PEP, and hooking things up so that we recognize the effect the new `@disjoint_base` decorator has.

## Test Plan

Mdtests and snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -1526,13 +1526,13 @@ impl<'db> ClassLiteral<'db> {
 
     /// Return `Some()` if this class is known to be a [`DisjointBase`], or `None` if it is not.
     pub(super) fn as_disjoint_base(self, db: &'db dyn Db) -> Option<DisjointBase<'db>> {
-        if let Some(known_class) = self.known(db) {
-            known_class
-                .is_disjoint_base()
-                .then_some(DisjointBase::hard_coded(self))
-        } else if self
-            .known_function_decorators(db)
-            .contains(&KnownFunction::DisjointBase)
+        // TODO: Typeshed cannot add `@disjoint_base` to its `tuple` definition without breaking pyright.
+        // See <https://github.com/microsoft/pyright/issues/10836>.
+        // This should be fixed soon; we can remove this workaround then.
+        if self.is_known(db, KnownClass::Tuple)
+            || self
+                .known_function_decorators(db)
+                .contains(&KnownFunction::DisjointBase)
         {
             Some(DisjointBase::due_to_decorator(self))
         } else if SlotsKind::from(db, self) == SlotsKind::NotEmpty {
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> No ecosystem changes detected ✅

A bit of a bummer -- I thought this would improve our understanding of disjointness a fair bit now that typeshed [has added `@disjoint_base`](https://github.com/python/typeshed/pull/14599) across all applicable stdlib classes. I expected to see at least some ecosystem hits.

> Memory usage changes were detected when running on open source projects

Interestingly exactly the same increase is being reported over in #20085 -- not sure if this job is getting flaky again...?

crates/ty_python_semantic/resources/mdtest/snapshots/instance_layout_conf…_-_Tests_for_ty's_`inst…_-_Built-ins_with_impli…_(f5857d64ce69ca1d).snap
```diff
@@ -12,160 +12,202 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/instance_layout_conflict
...
+   |
+info: Two classes cannot coexist in a class's MRO if their instances have incompatible memory layouts
+  --> src/mdtest_snippet.py:33:5
+   |
+32 | class I(  # error: [instance-layout-conflict]
+33 |     G,
+   |     - `G` instances have a distinct memory layout because of the way `G` is implemented in a C extension
+34 |     H
+   |     - `H` instances have a distinct memory layout because of the way `H` is implemented in a C extension
+35 | ): ...
```
AlexWaygood, line 199: of course, unlike the classes that we hardcode as being disjoint bases, we don't _know_ that two arbitrary classes decorated as being disjoint bases are marked as such due to the way they're implemented in a C extension. You could add that decorator to a class for whatever reason you like.

It holds true for all of typeshed's disjoint bases, though, and it's _probably_ a good guess in general...?

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -3624,13 +3656,12 @@ impl KnownClass {
         }
     }
 
-    /// Return `true` if this class is a [`SolidBase`]
-    const fn is_solid_base(self) -> bool {
+    /// Return `true` if this class is a [`DisjointBase`]
+    const fn is_disjoint_base(self) -> bool {
```
carljm: In principle we could eliminate this method and always look at the typeshed fallback class (and its decorator), right? This is just a small performance optimization (maybe) at the cost of preserving some special-casing that we technically shouldn't need anymore?

carljm (approved):
Heh, I wondered after the typeshed PR if you were just gonna go ahead and do this. Looks good!

AlexWaygood:
> Heh, I wondered after the typeshed PR if you were just gonna go ahead and do this

Ah, I'm an open book I see!

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: Exactly. I'll try getting rid of it and see what Codspeed says!

AlexWaygood: heh, Codspeed says that removing the method is no regression at all! Which means that, excluding tests, this PR now removes lines of code on net 🚀

PR #20085, opened 2025-08-25
[ty] Add support for PEP 750 t-strings 

dylwil3:
This PR attempts to add support for inferring `string.templatelib.Template` for t-string literals.

First PR of this kind so apologies in advance if I've made a mess of things! Also this is low priority so feel free to ignore. 

Diff is relatively small but feel free to review commit-by-commit if it's more convenient.

Also tested LSP:
demo

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/t_strings.md
+++ b/crates/ty_python_semantic/resources/mdtest/t_strings.md
@@ -1,5 +1,7 @@
 # Template strings
 
+(NB: `black` does not support Python 3.14 at the time of this writing)
+
 <!-- blacken-docs:off -->
 
 ```toml


crates/ty_python_semantic/src/types/class.rs
```diff
@@ -4128,6 +4136,7 @@ impl KnownClass {
             Self::NamedTupleFallback => "NamedTupleFallback",
             Self::NamedTupleLike => "NamedTupleLike",
             Self::TypedDictFallback => "TypedDictFallback",
+            Self::Template => "Template",
```
dylwil3, line 4061: Not sure if you'd like this to be the name or if I should use the full `string.templatelib.Template` here. A point of confusion is that (very unfortunately) `string.Template` exists.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -5177,7 +5191,13 @@ mod tests {
...
     #[test]
     fn known_class_roundtrip_from_str() {
-        let db = setup_db();
+        let mut db = setup_db();
+        Program::get(&db)
+            .set_python_version_with_source(&mut db)
+            .to(PythonVersionWithSource {
+                version: PythonVersion::latest_preview(),
+                source: PythonVersionSource::default(),
+            });
```
dylwil3, line 5117: Let me know if you'd like me to handle this differently, and/or add an assertion to remind us to modify this once the latest supported version is Python 3.14

crates/ty_python_semantic/resources/mdtest/t_strings.md
```diff
@@ -0,0 +1,18 @@
+# Template strings
+
+<!-- blacken-docs:off -->
```
AlexWaygood, line 5: what's the reason for switching off the formatter here?

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -3653,7 +3656,8 @@ impl KnownClass {
             | Self::Super
             | Self::GenericAlias
             | Self::Deque
-            | Self::Bytes => true,
+            | Self::Bytes
+            | Self::Template => true,
```
AlexWaygood: `Template` is [marked as `@final` in typeshed](https://github.com/python/typeshed/blob/91e2ed0953592795fd8c29e3005a1315bf652ffc/stdlib/string/templatelib.pyi#L7-L8), so I don't think we also need to consider it a disjoint base (see the comment a couple of lines below this one) -- it'll just result in us emitting two diagnostics if you tried to inherit from it, rather than one diagnostic. One seems enough :-)

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: this is correct -- it's just meant to be the `__name__` here, not the fullname!

AlexWaygood (approved):
Looks great!

crates/ty_python_semantic/resources/mdtest/t_strings.md
dylwil3: `black` does not yet support Python 3.14 so it errors here

AlexWaygood: ahhh, makes sense. Can you add a comment?

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: ...and I just deleted this method entirely in https://github.com/astral-sh/ruff/pull/20084, so the question is now moot 😆 sorry for giving you merge conflicts!

dylwil3: no worries!

PR #20089, opened 2025-08-25
Fix incorrect D413 links in docstrings convention FAQ

Avasam:
## Summary

D413 in this section was incorrectly linking to D410.

I haven't checked if this issue happens anywhere else in the docs.

## Test Plan

Look at docs


dhruvmanila (approved):
Thanks!

Issue #8451, opened 2023-11-03
Implement flake8-trio rules


charliermarsh:

qdegraaf:
`TRIO118` sounds like it is not needed in Ruff, similar to `TRIO106`

The `TRIO2XX` rules are partially covered by the port of [flake8_async](https://github.com/astral-sh/ruff/tree/75c9be099f878c381c8734ecd7c1217c44d2b612/crates/ruff_linter/src/rules/flake8_async/rules). 

The recommended fixes in `flake8-trio` are TRIO specific though. How do we want to combine those two?

jakkdl:
whoa, this is super cool!! As the primary author of flake8-trio I'm super honored to be included in ruff :heart: 
Feel free to ping me about any questions of functionality or implementation details
@Zac-HD might also have opinions or suggestions about stuff

jakkdl:
> The recommended fixes in `flake8-trio` are TRIO specific though. How do we want to combine those two?

There's some `anyio` support as well in flake8-trio and for a while we considered renaming the library. If `flake8-trio` sees `import anyio` and no `import trio` in the code, or [--anyio](https://github.com/Zac-HD/flake8-trio#--anyio) is passed it will suggest autofixes from the anyio library instead of trio.

jakkdl:
> TRIO118 sounds like it is not needed in Ruff, similar to TRIO106

TRIO106 definitely isn't needed by itself, and solely a consequence of implementation at the time. TRIO118 is only partially a weakness in implementation, the part about multi-backend programs still holds though.

> The TRIO2XX rules are partially covered by the port of [flake8_async](https://github.com/astral-sh/ruff/tree/75c9be099f878c381c8734ecd7c1217c44d2b612/crates/ruff_linter/src/rules/flake8_async/rules).

Yeah there's overlap there, though I gave the implementation in flake8-trio quite a bit more love and the suggestions/error messages are quite a bit more detailed.



TRIO105 is mostly redundant as of the recently released trio 0.23 which included types into the main package, and type checkers will complain about async functions not being awaited. I suppose it could be useful for somebody not using type-checking, but enabling this check, but I think that's a pretty slim proportion and the rule requires constantly keeping up with the API of anyio/trio.

Zac-HD:
🥳 I'm delighted to see this going into `ruff`!

[...]

PR #20091, opened 2025-08-26
[flake8-async] Implement `blocking-http-call-httpx` (ASYNC212)

amyreese:
## Summary

Adds new rule to find and report use of `httpx.Client` in synchronous
functions.

See issue #8451

## Test Plan

New snapshots for `ASYNC212.py` with `cargo insta test`.


amyreese:
Still needs some work (and learning) to catch usage of `httpx.Client` methods, not just instantiation.

crates/ruff_linter/resources/test/fixtures/flake8_async/ASYNC212.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+async def foo(client: Optional[httpx.Client]):
+    client.request()  # ASYNC212
+    client.anything()  # Ok
+
+
+async def foo():
+    client: httpx.Client = ...
+    client.request()  # ASYNC212
```
amyreese, line 61: This one still isn't caught, presumably because the TypeChecker bits only capture annotations at the function level?

crates/ruff_linter/src/rules/flake8_async/rules/blocking_http_call_httpx.rs
```diff
@@ -0,0 +1,120 @@
...
+///     async with httpx.AsyncClient() as client:
+///         response = await client.get(...)
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct BlockingHttpCallHttpxInAsyncFunction;
+
+impl Violation for BlockingHttpCallHttpxInAsyncFunction {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Async functions should not use blocking httpx clients".to_string()
```
ntBre: You don't have to do this, but if you want to include the blocking call and/or the client in the message, you can include their names on the struct. Something like:

```rust
#[derive(ViolationMetadata)]
pub(crate) struct BlockingHttpCallHttpxInAsyncFunction {
    call: String,
    client: String,
}

impl Violation for BlockingHttpCallHttpxInAsyncFunction {
    #[derive_message_formats]
    fn message(&self) -> String {
        format!(
            "Blocking sync HTTP call {call} on httpx object {client}, use httpx.AsyncClient.",
            call = self.call, 
            client = self.client,
        )
    }
}
```

Nice work on the upstream contribution, I saw it when checking the upstream rule docs this morning :)

We also don't have to match the upstream phrasing, just using that as an idea to let you know this was possible.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_http_call_httpx.rs
```diff
@@ -0,0 +1,120 @@
...
+
+struct HttpxClientChecker;
+
+impl TypeChecker for HttpxClientChecker {
+    fn match_annotation(
+        annotation: &ruff_python_ast::Expr,
+        semantic: &ruff_python_semantic::SemanticModel,
+    ) -> bool {
+        let mut found = false;
+        traverse_union_and_optional(
```
ntBre, line 73: I'm not sure we need this `traverse_union_and_optional` call, but to be honest I haven't written one of these `TypeChecker`s myself. Would something more like our `PathlibPathChecker` work:

https://github.com/astral-sh/ruff/blob/147f7a7f28c2c23ae5c749b7b57cadfc4a776ec3/crates/ruff_python_semantic/src/analyze/typing.rs#L914-L917

or did you run into cases where we weren't handling unions correctly? (That might be an issue with our other type checkers if so)

crates/ruff_linter/src/rules/flake8_async/rules/blocking_http_call_httpx.rs
```diff
@@ -0,0 +1,120 @@
...
+            .is_some_and(|qualified_name| matches!(qualified_name.segments(), ["httpx", "Client"]))
+    }
+}
+
+/// ASYNC212
+pub(crate) fn blocking_http_call_httpx(checker: &Checker, call: &ExprCall) {
+    let semantic = checker.semantic();
+    if semantic.in_async_context() {
+        if let Some(ast::ExprAttribute { value, attr, .. }) = call.func.as_attribute_expr() {
+            if let Some(name) = value.as_name_expr() {
```
ntBre: This looks great! My one tiny nit is to make this a sequence of early returns:


```suggestion
    if !semantic.in_async_context() {
        return;
    }
    
    let Some(ast::ExprAttribute { value, attr, .. }) = call.func.as_attribute_expr() else {
        return;
    }
    
    let Some(name) = value.as_name_expr() else {
        return;
    }
```

That's a bit more verbose here, but I think that's our general style.

crates/ruff_linter/resources/test/fixtures/flake8_async/ASYNC212.py
ntBre: I think this _should_ work. I played around a bit with the PTH210 rule, which uses the `PathlibPathChecker`, and it will detect both module-level and function-level bindings like this:

```py
def foo():
    path: Path = ...
    path.with_suffix(".")
```

I'm wondering if the union traversal might be disrupting this check.

crates/ruff_linter/src/rules/flake8_async/snapshots/ruff_linter__rules__flake8_async__tests__ASYNC212_ASYNC212.py.snap
```diff
@@ -0,0 +1,149 @@
...
+   |     ^^^^^^^^^^^^^
+38 |
+39 |     client.anything()  # Ok
+   |
+
+ASYNC212 Async functions should not use blocking httpx clients
+  --> ASYNC212.py:50:5
+   |
+49 | async def foo(client: httpx.Client | None):
+50 |     client.request()  # ASYNC212
```
ntBre, line 137: Ah okay, this doesn't work for PTH210. I understand the `traverse_union` call now. I'd probably lean toward not traversing unions and matching our other type inference calls, but I don't feel too strongly if you think this is an important use case. We could also consider making a more general change to our `TypeChecker` code if this would always be helpful.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_http_call_httpx.rs
ntBre: I saw below that you do need this check. If I'm following the `traverse_union_options` code correctly, I think it's actually not "traversing" the top-level expression:

https://github.com/astral-sh/ruff/blob/147f7a7f28c2c23ae5c749b7b57cadfc4a776ec3/crates/ruff_python_semantic/src/analyze/typing.rs#L528-L531

which would explain the basic annotation check not working. You might just need to call your closure on `annotation` itself before passing it to `traverse_union_and_optional` to cover both cases.

crates/ruff_linter/src/rules/flake8_async/snapshots/ruff_linter__rules__flake8_async__tests__ASYNC212_ASYNC212.py.snap
amyreese: I was doing this mostly to match behavior with the upstream ASYNC212 implementation, which explicitly includes these unions/optionals in its test cases.

ntBre: That makes sense! If we can handle both this and the regular annotation case, it's great to handle both. We can punt on trying to update any other `TypeChecker`s for now too.

ntBre (approved):
Nice, thanks!

Would you mind updating the title to:

[`flake8-async`] Implement `blocking-http-call-httpx` (`ASYNC212`)

(just wrapping the linter name and rule code in backticks) 

We pull the PR titles for changelog entries, and that's how we usually stylize them.

Other than that, feel free to merge whenever you're ready!

PR #20093, opened 2025-08-26
[ty] WIP: Check typevar assignability using constraint sets

dcreager:
This stacks on top of #19997, and uses the new constraint set implementation to check assignability for typevars.

Not ready for review yet.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
@@ -45,31 +45,27 @@ def even_given_unsatisfiable_constraints():
 
 ## Type variables
 
-The interesting case is typevars. The other typing relationships (TODO: will) all "punt" on the
-question when considering a typevar, by translating the desired relationship into a constraint set.
+The interesting case is typevars. The other typing relationships all "punt" on the question when
+considering a typevar, by translating the desired relationship into a constraint set.
 
 ```py
 from typing import Any
 from ty_extensions import is_assignable_to, is_subtype_of
 
 def assignability[T]():
-    # TODO: revealed: ty_extensions.ConstraintSet[T@assignability ≤ bool]
-    # revealed: ty_extensions.ConstraintSet[never]
+    # revealed: ty_extensions.ConstraintSet[(T@assignability ≤ bool)]
     reveal_type(is_assignable_to(T, bool))
-    # TODO: revealed: ty_extensions.ConstraintSet[T@assignability ≤ int]
-    # revealed: ty_extensions.ConstraintSet[never]
+    # revealed: ty_extensions.ConstraintSet[(T@assignability ≤ int)]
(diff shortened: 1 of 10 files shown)


dcreager:
The timeout in `koda-validate` comes from a constrained typevar with 9 constraints. There's some quadratic blowup that either wasn't there before, or now has enough extra work happening in each combination that we time out.

crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md
```diff
@@ -129,78 +129,270 @@ specialization. Thus, the typevar is a subtype of itself and of `object`, but no
...
+    reveal_type(is_assignable_to(T, T))
+    static_assert(is_assignable_to(T, T))
+
+    # revealed: ty_extensions.ConstraintSet[always]
+    reveal_type(is_assignable_to(T, object))
+    static_assert(is_assignable_to(T, object))
+
+    # revealed: ty_extensions.ConstraintSet[some valid specializations: (T@unbounded_unconstrained ≤ Super)]
+    reveal_type(is_assignable_to(T, Super))
+    static_assert(not is_assignable_to(T, Super))
```
dcreager: This is an example where the new reveal output provides more detail about our calculations. We determine the constraints under which the assignability check would hold. In this case, we're just moving the check into the constraint itself — `T` is assignable to `Super` when it is specialized to a type that is a subtype of `Super`. We also show that this constraint set does not cover all of the valid specializations of `T`, and so the assignability check fails.

crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md
```diff
@@ -129,78 +129,270 @@ specialization. Thus, the typevar is a subtype of itself and of `object`, but no
...
-
-    reveal_type(is_subtype_of(T, Any))  # revealed: ty_extensions.ConstraintSet[never]
-    reveal_type(is_subtype_of(Any, T))  # revealed: ty_extensions.ConstraintSet[never]
-    reveal_type(is_subtype_of(T, Super))  # revealed: ty_extensions.ConstraintSet[never]
-    reveal_type(is_subtype_of(Super, T))  # revealed: ty_extensions.ConstraintSet[never]
-    reveal_type(is_subtype_of(T, Sub))  # revealed: ty_extensions.ConstraintSet[never]
-    reveal_type(is_subtype_of(Sub, T))  # revealed: ty_extensions.ConstraintSet[never]
+    # revealed: ty_extensions.ConstraintSet[all valid specializations: (T@bounded ≤ Super)]
+    reveal_type(is_assignable_to(T, Super))
+    static_assert(is_assignable_to(T, Super))
```
dcreager: And here is an example where we infer some constraints for the assignabiltiy check, and where those constraints _do_ cover all valid specializations of `T`. And so the overall check does succeed.

dcreager:
Superceded by quite a lot of more recent work

PR #19029, opened 2025-06-29
[ty] Implement protocol property check

mtshiba:
## Summary

This is a follow-up to #18847 and implements checks on protocol property members.

Since the logic implemented in `TypeInferenceBuilder::validate_attribute_assignment` was required to check setters, I extracted the judgment part as a `Type` method. As a side effect, the diagnostics emission part and the judgment part are now separated, which I think has improved readability a little. Additionally, some error messages have been consolidated to make them more consistent.

## Test Plan

Some of the TODOs in `protocols.md` now work properly.

PR #20095, opened 2025-08-26
[ty] Add more tests for protocols

AlexWaygood:
## Summary

This PR adds more tests for protocols. It's split out from other PRs that provide functional changes to make those PRs easier to review.

Included here are:
- Bugfixes to tests that aren't actually testing the thing they're meant to be testing
- Regression tests for things that caused stack overflows in early versions of #19029 and #19866
- New equivalence tests for protocols with property members

## Test Plan

this pr is all tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1392,10 +1392,16 @@ static_assert(is_subtype_of(XClassVar, HasXProperty))
 static_assert(is_assignable_to(XClassVar, HasXProperty))
 
 class XFinal:
-    x: Final = 42
+    x: Final[int] = 42
 
 static_assert(is_subtype_of(XFinal, HasXProperty))
 static_assert(is_assignable_to(XFinal, HasXProperty))
+
+class XImplicitFinal:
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -1451,9 +1457,8 @@ static_assert(is_assignable_to(XReadWriteProperty, HasXProperty))
 class XSub:
     x: MyInt
 
-# TODO: should pass
-static_assert(not is_subtype_of(XSub, HasXProperty))  # error: [static-assert-error]
-static_assert(not is_assignable_to(XSub, HasXProperty))  # error: [static-assert-error]
+static_assert(not is_subtype_of(XSub, XReadWriteProperty))
+static_assert(not is_assignable_to(XSub, XReadWriteProperty))
```
AlexWaygood, line 1461: this was a bug in the test: we check assignability of `XSub` to `HasXProperty` a few lines above. Here we were meant to be asserting assignability of `XSub` vs `XReadWriteProperty`, but there was a copy/paste error in the test

carljm (approved):

PR #20098, opened 2025-08-26
[ty] Reverse TDD node ordering

sharkdp:

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
+++ b/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
@@ -424,9 +424,9 @@ impl ReachabilityConstraintsBuilder {
         }
     }
 
-    /// Returns whether `a` or `b` has a "larger" atom. TDDs are ordered such that interior nodes
-    /// can only have edges to "larger" nodes. Terminals are considered to have a larger atom than
-    /// any internal node, since they are leaf nodes.
+    /// Implements an ordering of constraint IDs based on their atoms. TDDs are ordered such that
+    /// interior nodes can only have edges to "larger" nodes (with a "smaller" atom). Terminals
+    /// are considered to have a larger atom than any internal node, since they are leaf nodes.
     fn cmp_atoms(
         &self,
         a: ScopedReachabilityConstraintId,
@@ -439,7 +439,7 @@ impl ReachabilityConstraintsBuilder {
         } else if b.is_terminal() {
             Ordering::Less
         } else {
-            self.interiors[a].atom.cmp(&self.interiors[b].atom)
+            self.interiors[b].atom.cmp(&self.interiors[a].atom)
         }
     }
 


crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
```diff
@@ -424,9 +424,9 @@ impl ReachabilityConstraintsBuilder {
         }
     }
 
-    /// Returns whether `a` or `b` has a "larger" atom. TDDs are ordered such that interior nodes
-    /// can only have edges to "larger" nodes. Terminals are considered to have a larger atom than
-    /// any internal node, since they are leaf nodes.
+    /// Implements an ordering of constraint IDs based on their atoms. TDDs are ordered such that
+    /// interior nodes can only have edges to "larger" nodes (with a "smaller" atom). Terminals
+    /// are considered to have a larger atom than any internal node, since they are leaf nodes.
```
carljm: I think this comment is now quite confusing.

Would it be equivalent here if we compared interiors in the "forward" direction, swapped the `Greater` and `Less` cases below, and then reversed the direction in which we use this ordering? It seems like that would make it possible to describe the behavior of this method more straightforwardly.

sharkdp: The problem is that the terminal nodes need to be larger than anything else. If we swap the `Greater` and `Less` cases below instead, that would also invert the logic for terminal nodes (which violates invariants and leads to panics). But maybe I haven't understood what you meant by this:

> and then reversed the direction in which we use this ordering

carljm: What I meant was that in each of the two places we call `.cmp_atoms` we would also swap the match arms for `Greater` and `Less`, so the invariant becomes that interior nodes can only have edges to "smaller" nodes and terminals are considered "smaller" than any internal node. Then I think we get the same effect but eliminate the confusing idea that a "larger" node has a "smaller" atom.

This change passes mdtests for me locally:

```diff
diff --git a/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs b/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
index 9042e919be..2545413f0e 100644
--- a/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
+++ b/crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
@@ -435,9 +435,9 @@ impl ReachabilityConstraintsBuilder {
         if a == b || (a.is_terminal() && b.is_terminal()) {
             Ordering::Equal
         } else if a.is_terminal() {
-            Ordering::Greater
-        } else if b.is_terminal() {
             Ordering::Less
+        } else if b.is_terminal() {
+            Ordering::Greater
         } else {
             self.interiors[a].atom.cmp(&self.interiors[b].atom)
         }
@@ -549,7 +549,7 @@ impl ReachabilityConstraintsBuilder {
                 };
                 (a_node.atom, if_true, if_ambiguous, if_false)
             }
-            Ordering::Less => {
+            Ordering::Greater => {
                 let a_node = self.interiors[a];
                 let if_true = self.add_or_constraint(a_node.if_true, b);
                 let if_false = self.add_or_constraint(a_node.if_false, b);
@@ -560,7 +560,7 @@ impl ReachabilityConstraintsBuilder {
                 };
                 (a_node.atom, if_true, if_ambiguous, if_false)
             }
-            Ordering::Greater => {
+            Ordering::Less => {
                 let b_node = self.interiors[b];
                 let if_true = self.add_or_constraint(a, b_node.if_true);
                 let if_false = self.add_or_constraint(a, b_node.if_false);
@@ -615,7 +615,7 @@ impl ReachabilityConstraintsBuilder {
                 };
                 (a_node.atom, if_true, if_ambiguous, if_false)
             }
-            Ordering::Less => {
+            Ordering::Greater => {
                 let a_node = self.interiors[a];
                 let if_true = self.add_and_constraint(a_node.if_true, b);
                 let if_false = self.add_and_constraint(a_node.if_false, b);
@@ -626,7 +626,7 @@ impl ReachabilityConstraintsBuilder {
                 };
                 (a_node.atom, if_true, if_ambiguous, if_false)
             }
-            Ordering::Greater => {
+            Ordering::Less => {
                 let b_node = self.interiors[b];
                 let if_true = self.add_and_constraint(a, b_node.if_true);
```

carljm: But I don't think this is worth a lot of bother. If there's an issue with that code change and we need to leave the code the same, and accept that "smaller atom" means "larger node", then I think there's still something confusing/wrong that we should fix in the docs here -- it currently says "Terminals are considered to have a larger atom than any internal node" but I think that's backwards, it should say "Terminals are considered to be larger than any internal node." Because being a larger node implies "smaller atom" under the current "reversed" semantics.

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
```diff
@@ -439,7 +439,9 @@ impl ReachabilityConstraintsBuilder {
         } else if b.is_terminal() {
             Ordering::Less
         } else {
-            self.interiors[a].atom.cmp(&self.interiors[b].atom)
+            // See https://github.com/astral-sh/ruff/pull/20098 for an explanation of why this
+            // ordering is reversed.
+            self.interiors[b].atom.cmp(&self.interiors[a].atom)
```
dcreager: nit: What do you think about this instead?

```suggestion
            // See https://github.com/astral-sh/ruff/pull/20098 for an explanation of why this
            // ordering is reversed.
            self.interiors[a].atom.cmp(&self.interiors[b].atom).reverse()
```

To me that stands out more clearly that we are doing a reversal, rather than having to notice that `a` and `b` are reversed compared to how they appear in the parameter list.

(I'm also completely fine leaving this as-is if you prefer it that way)

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
dcreager: How about something like

```
Implements the ordering that determines which level a TDD node appears at.

Each interior node checks the value of a single variable (for us, a `Predicate`).
TDDs are ordered such that every path from the root of the graph to the leaves
must check each variable at most once, and must check each variable in the same
order.

We can choose any ordering that we want, as long as it's consistent — with the
caveat that terminal nodes must always be last in the ordering, since they are
the leaf nodes of the graph.

We currently compare interior nodes by looking at the Salsa IDs of each variable's
`Predicate`, since this is already available and easy to compare. We also
_reverse_ the comparison of those Salsa IDs. The Salsa IDs are assigned roughly
sequentially while traversing the source code. Reversing the comparison means
`Predicate`s that appear later in the source will tend to be placed "higher"
(closer to the root) in the TDD graph. We have [found empirically](https://github.com/astral-sh/ruff/pull/20098)
that this leads to smaller TDD graphs, since there are often repeated
combinations of `Predicate`s from earlier in the file.
```

(You'd have to reflow that to appease the pre-commit gremlin)

sharkdp: > How about something like

That is great, thank you!

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
sharkdp: I changed the `.reverse()` call to `b.cmp(a)` because I was worried about it being an additional operation, since this function is called *a lot*. But it's probably negligible compared to the hashmap operations and everything else that we do in `add_{and,or}_constraint`. And maybe it's a no-op after optimization anyway.

Changed back to `.reverse()` now.

PR #20099, opened 2025-08-26
[`flake8-use-pathlib`] Delete unused `Rule::OsSymlink`

chirizxc:


ntBre (approved):
Thanks for following up!

Issue #20088, opened 2025-08-25
PTH100 unsafely converts `os.path.abspath` to `pathlib.Path.resolve`


dscorbett:
### Summary

The fix for [`os-path-abspath` (PTH100)](https://docs.astral.sh/ruff/rules/os-path-abspath/) converts `os.path.abspath` to `pathlib.Path.resolve`, which does not do the same thing. Sometimes, as in [upstream flake8-use-pathlib issue #8](https://gitlab.com/RoPP/flake8-use-pathlib/-/issues/8), `pathlib.Path.absolute` is the preferred replacement, but there is no exact equivalent; see https://github.com/python/cpython/issues/69200. The documentation for PTH100 should explain this and the fix should be marked unsafe.

| Function | Removes `..`? | Resolves symlinks? |
| --- | --- | --- |
| `abspath` | ✅ | ❌ |
| `resolve` | ✅ | ✅ |
| `absolute` | ❌ | ❌ |

[Example](https://play.ruff.rs/3a148d9d-e0a8-4282-a692-cd8de79492f2):
```console
$ cat >pth100.py <<'# EOF'
import os
open("pth100_src", "w").close()
os.symlink("pth100_src", "pth100_dst")
os.mkdir("pth100_dir")
print(str(os.path.abspath("pth100_dir/../dst")).removeprefix(f"{os.getcwd()}/"))
# EOF

$ rm -fr pth100_{dir,dst,src}

$ python pth100.py
dst

$ ruff --isolated check pth100.py --select PTH100 --preview --fix
Found 1 error (1 fixed, 0 remaining).

$ cat pth100.py
import os
import pathlib
open("pth100_src", "w").close()
os.symlink("pth100_src", "pth100_dst")
os.mkdir("pth100_dir")
print(str(pathlib.Path("pth100_dir/../dst").resolve()).removeprefix(f"{os.getcwd()}/"))

$ rm -fr pth100_{dir,dst,src}

$ python pth100.py
src
```

### Version

ruff 0.12.10 (c68ff8d90 2025-08-21)

chirizxc:
I'm fix it

ntBre:
[...]

PR #20100, opened 2025-08-26
[`flake8-use-pathlib`] Make `PTH100` fix unsafe because it can change behavior

chirizxc:
… behavior

## Test Plan

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
@@ -1,9 +1,15 @@
-use crate::checkers::ast::Checker;
-use crate::preview::is_fix_os_path_abspath_enabled;
-use crate::rules::flake8_use_pathlib::helpers::check_os_pathlib_single_arg_calls;
-use crate::{FixAvailability, Violation};
+use ruff_diagnostics::{Applicability, Edit, Fix};
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::ExprCall;
+use ruff_text_size::Ranged;
+
+use crate::checkers::ast::Checker;
+use crate::importer::ImportRequest;
+use crate::preview::is_fix_os_path_abspath_enabled;
+use crate::rules::flake8_use_pathlib::helpers::{
+    has_unknown_keywords_or_starred_expr, is_pathlib_path_call,
+};
+use crate::{FixAvailability, Violation};
 
 /// ## What it does
 /// Checks for uses of `os.path.abspath`.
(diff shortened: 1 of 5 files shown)


crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
```diff
@@ -63,12 +73,45 @@ pub(crate) fn os_path_abspath(checker: &Checker, call: &ExprCall, segments: &[&s
...
+            format!("{arg_code}.resolve()")
+        } else {
+            format!("{binding}({arg_code}).resolve()")
+        };
+
+        Ok(Fix::applicable_edits(
+            Edit::range_replacement(replacement, range),
+            [import_edit],
+            Applicability::Unsafe,
+        ))
```
ntBre: Since it's always unsafe, we can just use `Fix::unsafe_edits` instead of `applicable_edits`.


```suggestion
        Ok(Fix::unsafe_edits(
            Edit::range_replacement(replacement, range),
            [import_edit],
        ))
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
```diff
@@ -36,6 +42,10 @@ use ruff_python_ast::ExprCall;
 /// ## Fix Safety
 /// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
 ///
+/// `Path.resolve()` resolves symlinks, while `os.path.abspath()` does not.
+/// If preserving symlinks is important, you may need to use `Path.absolute()`.
+/// The suggested autofix is marked as unsafe because it can change behavior.
```
ntBre: I don't think it's so important to mention the comment issue now that the rule is always unsafe. I'd probably say something like:

> This rule's fix is always marked as unsafe because `Path.resolve()` resolves symlinks, while `os.path.abspath()` does not. If resolving symlinks is important, you may need to use `Path.absolute()`. However, `Path.absolute()` also does not remove any `..` components in a path, unlike `os.path.abspath()` and `Path.resolve()`, so if that specific combination of behaviors is required, there's no existing `pathlib` alternative. See CPython issue [#69200](https://github.com/python/cpython/issues/69200).

That's a bit wordy, so it could be nice to include @dscorbett's nice table from the issue instead, up to you.

ntBre: I can't comment on the line below, but as I mentioned on the issue, we should also update the `Correspondence between os and pathlib` link below (https://github.com/astral-sh/ruff/issues/20088#issuecomment-3225030199). The new link is https://docs.python.org/3/library/pathlib.html#corresponding-tools

ntBre (commented):
Thanks! Just a couple of small suggestions.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
chirizxc: It seems to me that if a fix is going to change behavior in any way, it's better to leave it as a unsafe

chirizxc:
Should I change links in this PR in the same way in all `PTH` rules?

chirizxc:
I also see that there are many more functions in the table for which there are no rules and fixes, can we start adding them?

ntBre:
> Should I change links in this PR in the same way in all `PTH` rules?

Sure, but let's do that in a follow-up documentation PR. I think this one is good to go now.

> I also see that there are many more functions in the table for which there are no rules and fixes, can we start adding them?

I think we've implemented all of the rules in the upstream `flake8-use-pathlib`, and it's not a high priority to add more rules right now, so I don't think we need to exhaustively cover the table. You could open an issue to track interest in any of the ones we're missing, though, if you want!

PR #20101, opened 2025-08-26
Use new diff rendering format in tests

ntBre:
## Summary

I spun this off from #19919 to separate the rendering code change and snapshot updates from the (much smaller) changes to expose this in the CLI. I grouped all of the `ruff_linter` snapshot changes in the final commit in an effort to make this easier to review. The code changes are in [this range](https://github.com/astral-sh/ruff/pull/20101/files/619395eb417d2030e31fbb0e487a72dac58902db).

I went through all of the snapshots, albeit fairly quickly, and they all looked correct to me. In the last few commits I was trying to resolve an existing issue in the alignment of the line number separator:

https://github.com/astral-sh/ruff/blob/73720c73be981df6f71ce837a67ca1167da0265e/crates/ruff_linter/src/rules/flake8_comprehensions/snapshots/ruff_linter__rules__flake8_comprehensions__tests__C409_C409.py.snap#L87-L89

In the snapshot above on `main`, you can see that a double-digit line number at the end of the context lines for a snippet was causing a misalignment with the other separators. That's now resolved. The one downside is that this can lead to a mismatch with the diagnostic above:

```
C409 [*] Unnecessary list literal passed to `tuple()` (rewrite as a tuple literal)
 --> C409.py:4:6
  |
2 |   t2 = tuple([1, 2])
3 |   t3 = tuple((1, 2))
4 |   t4 = tuple([
  |  ______^
5 | |     1,
6 | |     2
7 | | ])
  | |__^
8 |   t5 = tuple(
9 |       (1, 2)
  |
help: Rewrite as a tuple literal
1  | t1 = tuple([])
2  | t2 = tuple([1, 2])
3  | t3 = tuple((1, 2))
   - t4 = tuple([
4  + t4 = (
5  |     1,
6  |     2
   - ])
7  + )
8  | t5 = tuple(
9  |     (1, 2)
10 | )
note: This is an unsafe fix and may remove comments or change runtime behavior
```

But I don't think we can avoid that without really reworking this rendering to make the diagnostic and diff rendering aware of each other. Anyway, this should only happen in relatively rare cases where the diagnostic is near a digit boundary and also near a context boundary. Most of our diagnostics line up nicely.

Another potential downside of the new rendering format is its handling of long stretches of `+` or `-` lines:

```
help: Replace with `Literal[...] | None`
21 |     ...
22 |
23 |
   - def func6(arg1: Literal[
   -     "hello",
   -     None  # Comment 1
   -     , "world"
   -     ]):
24 + def func6(arg1: Literal["hello", "world"] | None):
25 |     ...
26 |
27 |
note: This is an unsafe fix and may remove comments or change runtime behavior
```

To me it just seems a little hard to tell what's going on with just a long streak of `-`-prefixed lines. I saw an even more exaggerated example at some point, but I think this is also fairly rare. Most of the snapshots seem more like the examples we looked at on Discord with plenty of `|` lines and pairs of `+` and `-` lines. 

## Test Plan

Existing tests plus one new test in `ruff_db` to isolate a line separator alignment issue


dhruvmanila (approved):
This looks pretty straightforward change, thanks for taking the time to go through the snapshot changes!

crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -235,31 +225,35 @@ impl std::fmt::Display for Diff<'_> {
...
-            },
+        match self.fix.applicability() {
+            Applicability::Safe => {}
+            Applicability::Unsafe => {
+                writeln!(
+                    f,
+                    "{note}: {msg}",
+                    note = fmt_styled("note", self.stylesheet.warning),
+                    msg = fmt_styled(
+                        "This is an unsafe fix and may remove comments or change runtime behavior",
```
zanieb: I might just say "change behavior"? I guess we could keep "may remove comments" for now, but I'm wary to encode that into the meaning of "unsafe" (I think it's sort of vague these days). The silly part is that we should _know_ if we've dropped a comment, e.g., by inspecting the diff? That's separate work though :)

zanieb (approved):

crates/ruff_db/src/diagnostic/render/full.rs
ntBre: Sounds good! Those are the two main criteria in our docs, but it does feel a bit too specific to include in each diagnostic message.

PR #20103, opened 2025-08-26
[`flake8-use-pathlib`] Update links to the table showing the correspondence between `os` and `pathlib`

chirizxc:
## Summary

Part of https://github.com/astral-sh/ruff/pull/20100 | https://github.com/astral-sh/ruff/pull/20100#issuecomment-3225349156


chirizxc:
Ecosystem seems to be showing the wrong results for the previews, there isn't even a code change in this PR, just documentation

ntBre (approved):
Awesome, thanks!

The ecosystem check may have grabbed the binary from before your other PR merged. I might just kick that off again, but this should be good to go anyway.

PR #20105, opened 2025-08-26
Deduplicate input paths

### LINKED ISSUE #19395: Formatting an overlapping set of files can lead to errors
### Summary

This is probably on us, but I thought it would be good to bring it to your attention at least for doc purposes.

We were running `ruff format <dir1> <dir2>` on some generated code as part of our build pipeline. From time to time, the format step will blow up with syntax errors in some of the source files, i.e. as if the source python code is not valid python. After splitting the command into two separate format commands (one for dir 1 and one for dir 2), the issue got resolved (this was guesswork-type of fix).

So we took a harder look and we saw that the dir2 is a subdir of dir1 and we are passing it because it is in .gitignore. So I went and looked at the ruff format source code and noticed it is doing a formatting of all given arguments in parallel: https://github.com/astral-sh/ruff/blob/main/crates/ruff/src/commands/format.rs#L108

Is it possible there is a race condition here where formatting on dir may mess up the formatting of of another, overlapping dir?

### Version

ruff 0.9.9

### LINKED ISSUE #20035: `ruff format` can double count formatted files
### Summary

```
❯ cd /tmp
❯ rm -rf example
❯ mkdir example

❯ echo "x   = 1" > example/foo.py
❯ uvx ruff format example example/foo.py
2 files reformatted

❯ echo "x   = 1" > example/foo.py
❯ uvx ruff format example
1 file reformatted
```

`foo.py` appears to be counted twice when there are two ways it is selected.

Here are some logs

```
[2025-08-21][18:02:39][ruff::resolve][DEBUG] Using Ruff default settings
[2025-08-21][18:02:39][ignore::gitignore][DEBUG] opened gitignore file: /Users/zb/.gitignore
[2025-08-21][18:02:39][ruff_workspace::resolver][DEBUG] Included path via `include`: "/private/tmp/example/foo.py"
[2025-08-21][18:02:39][ruff::commands::format][DEBUG] format_path; path=/private/tmp/example/foo.py
[2025-08-21][18:02:39][ruff::commands::format][DEBUG] format_path; path=/private/tmp/example/foo.py
[2025-08-21][18:02:39][tracing::span][DEBUG] Printer::print;
[2025-08-21][18:02:39][tracing::span][DEBUG] Printer::print;
[2025-08-21][18:02:39][ruff::commands::format][DEBUG] Formatted 2 files in 2.87ms
```

### Version

0.12.10

TaKO8Ki:
## Summary

Fixes #20035, fixes #19395

This is for deduplicating input paths to avoid processing the same file multiple times.

This is my first contribution, so I'm sorry if I miss something. Please tell me if this is needed for this feature.

## Test Plan

I just added a test `find_python_files_deduplicated` in https://github.com/TaKO8Ki/ruff/blob/eee1020e322e693bf977d91bf3edd03b45420254/crates/ruff_workspace/src/resolver.rs#L1017
. This pull request adds changes to `WalkPythonFilesState::finish`, which is used in `python_files_in_path`, so they affect some commands such as `analyze`, `format`, `check` and so on. I will add snapshot tests for them if necessary.

I’ve already confirmed that the same thing happens with ruff check as well.

```
$ echo "x   = 1" > example/foo.py
$ uvx ruff check example example/foo.py
I002 [*] Missing required import: `from __future__ import annotations`
--> /path/to/example/foo.py:1:1
help: Insert required import: `from __future__ import annotations`

I002 [*] Missing required import: `from __future__ import annotations`
--> /path/to/example/foo.py:1:1
help: Insert required import: `from __future__ import annotations`

Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

### DIFF
--- a/crates/ruff_workspace/src/resolver.rs
+++ b/crates/ruff_workspace/src/resolver.rs
@@ -1023,32 +1023,6 @@ mod tests {
         Ok(())
     }
 
-    #[test]
-    fn find_python_files_deduplicated() -> Result<()> {
-        // Initialize the filesystem:
-        //   root
-        //   ├── file1.py
-        let tmp_dir = TempDir::new()?;
-        let root = tmp_dir.path();
-        let file1 = root.join("file1.py");
-        File::create(&file1)?;
-
-        let (paths, _) = python_files_in_path(
-            &[root.to_path_buf(), file1.clone()],
-            &PyprojectConfig::new(PyprojectDiscoveryStrategy::Fixed, Settings::default(), None),
-            &NoOpTransformer,
-        )?;
-        let paths = paths
-            .into_iter()
-            .flatten()
-            .map(ResolvedFile::into_path)
-            .sorted()
-            .collect::<Vec<_>>();
-        assert_eq!(paths, [file1]);
-
-        Ok(())
-    }
-
     fn make_exclusion(file_pattern: FilePattern) -> GlobSet {
         let mut builder = globset::GlobSetBuilder::new();
         file_pattern.add_to(&mut builder).unwrap();


TaKO8Ki:
I need to handle this cause `Explicitly pass test.py, should be linted regardless of it being excluded by lint.exclude`, so I will.

TaKO8Ki:
@ntBre Thank you for reviewing another pull request of mine. Could you approve the workflow and review this pull request if possible?

ntBre:
No problem! I'll get to this soon, it's on my todo list :) I'll kick off the workflow now

TaKO8Ki:
@ntBre Thank you. I fixed clippy errors on CI.

crates/ruff_workspace/src/resolver.rs
```diff
@@ -523,10 +523,50 @@ impl<'config> WalkPythonFilesState<'config> {
...
     }
 }
 
+/// Deduplicate files by path, prioritizing Root files over Nested files.
+///
+/// This ensures that when the same file is found both as a directly specified
+/// input (Root) and discovered through directory traversal (Nested), the Root
+/// version takes precedence. This behavior is important for explicit exclusion
+/// handling, where explicitly passed files should override directory-based
+/// discovery rules.
```
ntBre: Did you run into this situation when testing this out? This is my first time looking at this code, but it seems like we will have already applied exclusion rules in `python_files_in_path` here:

https://github.com/astral-sh/ruff/blob/e42006ece8c53f5e4196849af5c0b3d6c6fadfd2/crates/ruff_workspace/src/resolver.rs#L470-L476

and in `PythonFilesVisitorBuilder`:

https://github.com/astral-sh/ruff/blob/e42006ece8c53f5e4196849af5c0b3d6c6fadfd2/crates/ruff_workspace/src/resolver.rs#L574-L581

which will both run before `WalkPythonFileState::finish`.

It makes sense to me to favor `Root` nodes over `Nested` ones when deduplicating, but I'm not sure if it actually affects exclusions. It might be good to add a test for that case, if it does.

crates/ruff_workspace/src/resolver.rs
```diff
@@ -1003,6 +1043,32 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn find_python_files_deduplicated() -> Result<()> {
```
ntBre: I think I would kind of prefer a CLI test or two. Maybe one in [`lint.rs`](https://github.com/astral-sh/ruff/blob/e42006ece8c53f5e4196849af5c0b3d6c6fadfd2/crates/ruff/tests/lint.rs#L5863) and one in [`format.rs`](https://github.com/astral-sh/ruff/blob/e42006ece8c53f5e4196849af5c0b3d6c6fadfd2/crates/ruff/tests/format.rs#L2329) to make sure it works for both cases?

I guess both of the bug reports were about formatting, but the linter shows the same behavior in my local testing.

crates/ruff_workspace/src/resolver.rs
```diff
@@ -523,10 +523,50 @@ impl<'config> WalkPythonFilesState<'config> {
...
 }
 
+/// Deduplicate files by path, prioritizing Root files over Nested files.
+///
+/// This ensures that when the same file is found both as a directly specified
+/// input (Root) and discovered through directory traversal (Nested), the Root
+/// version takes precedence. This behavior is important for explicit exclusion
+/// handling, where explicitly passed files should override directory-based
+/// discovery rules.
+fn deduplicate_files(files: &ResolvedFiles) -> ResolvedFiles {
```
ntBre: I think if we take an owned `ResolvedFiles` here we could avoid some clones. That makes sense anyway since we're "consuming" the input `files` and returning a new version.

crates/ruff_workspace/src/resolver.rs
```diff
@@ -523,10 +523,50 @@ impl<'config> WalkPythonFilesState<'config> {
...
+    let mut seen_paths = FxHashSet::default();
+    let mut deduplicated_files = Vec::new();
+
+    for file_result in files {
```
ntBre: Would it be simpler just to sort the `files` and then deduplicate them? Something like this:

```rust
fn deduplicate_files(mut files: ResolvedFiles) -> ResolvedFiles {
    files.sort();
    files.dedup_by_key(|result| result.map(|file| file.path()));
    files
```

If we derive `PartialOrd` on `ResolvedFile`, I think this will automatically sort `Root` files before `Nested` files, and then the dedup call will take the first element with a given path.

I think this would also deduplicate the `Error`s, which could be a good thing or a bad thing. I'm not really sure.

Another idea, which seems like it could be cheaper overall, would be to filter out duplicates as we're walking the file system. Did you give that a try? It seems a bit more intuitive to me and avoids having to sort or filter anything at the end. I think if `PythonFilesVisitor::local_files` were an `FxHashMap` of path -> `Result<ResolvedFile>`, it might be pretty straightforward.

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: Yes. They don't deduplicate input paths in this situation.

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: Yes. You're right. As I said pull request description, ruff check show the same result. 

> I’ve already confirmed that the same thing happens with ruff check as well.

So I will add CLI tests too.

TaKO8Ki:
@ntBre Thank you for the review. I have addressed your comments.

crates/ruff_workspace/src/resolver.rs
MichaReiser: I'm not entirely sure I follow why this logic is important. If the paths are identical, why does it matter that we return the root path first? 

Can you add an example why this is important? If it isn't important, should we use a `Set` instead of a `Vec` to avoid this collecting step altogether? Or could we do better and deduplicate the input paths instead?

TaKO8Ki: It's important and related to: https://github.com/astral-sh/ruff/blob/9edbeb44dd5869a32200cbb442221041ce4ace1a/crates/ruff/tests/lint.rs#L252

Root has to be prioritized because `ResolvedFile::Root` means “explicitly passed on the CLI,” and excludes only apply to non‑root entries unless --force-exclude is set.

Dropping the root entry means that the explicitly passed path may be unintentionally ignored, since it is treated as nested and can be excluded despite being requested.

Concretely, with `lint.exclude = ["foo.py"]` and `ruff check . foo.`py, we must keep Root(foo.py) and drop Nested(foo.py) so foo.py is linted as the user requested.

TaKO8Ki: I have added this explanation to the comment.

crates/ruff_workspace/src/resolver.rs
```diff
@@ -763,7 +755,14 @@ impl PartialOrd for ResolvedFile {
 
 impl Ord for ResolvedFile {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.path().cmp(other.path())
+        match self.path().cmp(other.path()) {
```
ntBre: I'm pretty sure this is equivalent to the implementation that would be derived, unless I'm missing some subtlety here. Could we replace the manual `Ord` and `PartialOrd` implementations with `derive(Ord, PartialOrd)`?

crates/ruff_workspace/src/resolver.rs
```diff
@@ -523,48 +523,40 @@ impl<'config> WalkPythonFilesState<'config> {
...
+/// we must keep `Root(foo.py)` and drop `Nested(foo.py)` so `foo.py` is
+/// linted as the user requested.
+fn deduplicate_files(mut files: ResolvedFiles) -> ResolvedFiles {
+    // Sort by path; for identical paths, prefer Root over Nested; place errors after files
+    files.sort_by(|a, b| match (a, b) {
+        (Ok(a_file), Ok(b_file)) => a_file.cmp(b_file),
+        (Ok(_), Err(_)) => Ordering::Less,
+        (Err(_), Ok(_)) => Ordering::Greater,
+        (Err(_), Err(_)) => Ordering::Equal,
+    });
```
ntBre, line 552: I think this is also nearly the same as the derived `PartialOrd` implementation for `Result`, except that we consider any errors equal. It seems okay to me just to call `files.sort()` and use the default implementation.

ntBre (commented):
Thanks, I just had a couple more potential simplification suggestions. I'm also still interested in deduplicating the files as we traverse the file system instead of sorting and filtering at the end, as I mentioned in https://github.com/astral-sh/ruff/pull/20105#discussion_r2308327043, but if Micha is happy with this then I am too! This is probably the easier approach if it's not too expensive.

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: I also think it's ideal, but we cannot use `files.sort()` here since `ignore::Error` does not implement `Ord`.

```
type ResolvedFiles = Vec<Result<ResolvedFile, ignore::Error>>;
```

```
the trait bound `ignore::Error: std::cmp::Ord` is not satisfied
the trait `std::cmp::Ord` is implemented for `std::result::Result<T, E>`
required for `std::result::Result<resolver::ResolvedFile, ignore::Error>` to implement `std::cmp::Ord`
```

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: I will try using FxHashMap for local_files and confirm if it does not cause any side effects.

crates/ruff_workspace/src/resolver.rs
ntBre: Ahh okay, thanks!

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: I have added `derive(Ord, PartialOrd)`.

TaKO8Ki:
@ntBre 

> I'm also still interested in deduplicating the files as we traverse the file system instead of sorting and filtering at the end, as I mentioned in https://github.com/astral-sh/ruff/pull/20105#discussion_r2308327043, but if Micha is happy with this then I am too! This is probably the easier approach if it's not too expensive.

I tested the approach and confirmed it seems feasible to implement. However, changing the type of `local_files` would directly affect `python_files_in_path`, which is used by various commands, so the impact area is fairly large. It’s possible to contain the change within `PythonFilesVisitorBuilder`, but doing so would require modifying the current implementation to discard the `ignore::Error` that’s currently passed to `python_files_in_path`. Given the scope, this looks like a significant change—would it be alright if I handle it in a separate follow-up pull request?

crates/ruff_workspace/src/resolver.rs
```diff
@@ -1003,6 +1023,32 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn find_python_files_deduplicated() -> Result<()> {
```
ntBre: I think we could probably drop this test now that we have CLI tests covering the same code.

crates/ruff/tests/lint.rs
```diff
@@ -271,6 +271,40 @@ OTHER = "OTHER"
     Ok(())
 }
 
+/// Regression test for <https://github.com/astral-sh/ruff/issues/20035>
+#[test]
+fn deduplicate_directory_and_explicit_file() -> Result<()> {
```
ntBre, line 276: Could we add an exclusion to one of these to show the `Root` vs `Nested` behavior?

crates/ruff_workspace/src/resolver.rs
TaKO8Ki: I've removed it.

crates/ruff/tests/lint.rs
TaKO8Ki: I've added exclude option to the test case for lint.

ntBre (approved):
Thank you! I tried out my hash map idea locally, and I think I agree with you that this is a nicer way to go.

I think we could get around the `local_files` type issues (https://github.com/astral-sh/ruff/pull/20105#issuecomment-3299500183) by using some kind of wrapper type (either wrapping a `Vec` or converting to a `Vec` at the end). I was playing with something like this locally:

```rust
struct LocalFiles(Vec<Result<ResolvedFile, ignore::Error>>);
```

and doing the deduplication in `LocalFiles::push`, but the validation is still tricky. I think what you have is good for now.

ntBre:
I updated the summary to close https://github.com/astral-sh/ruff/issues/19395 too!

PR #20107, opened 2025-08-27
[ty] support PEP 613 type aliases

carljm:
## Summary

Add support for PEP 613 type aliases (defined by annotating with `typing.TypeAlias`), including recursive ones.

## Test Plan

Added mdtests and updated existing tests.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -11785,7 +11785,7 @@ pub(super) fn determine_upper_bound<'db>(
 // Make sure that the `Type` enum does not grow unexpectedly.
 #[cfg(not(debug_assertions))]
 #[cfg(target_pointer_width = "64")]
-static_assertions::assert_eq_size!(Type, [u8; 16]);
+static_assertions::assert_eq_size!(Type, [u8; 24]);
 
 #[cfg(test)]
 pub(crate) mod tests {
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6257,7 +6257,9 @@ impl<'db> Type<'db> {
             }),
 
             Type::KnownInstance(known_instance) => match known_instance {
-                KnownInstanceType::TypeAliasType(alias) => Ok(Type::TypeAlias(*alias)),
+                KnownInstanceType::TypeAliasType(alias) | KnownInstanceType::TypeAlias(alias) => {
+                    Ok(Type::TypeAlias(*alias))
+                }
                 KnownInstanceType::TypeVar(typevar) => {
                     let index = semantic_index(db, scope_id.file(db));
                     Ok(bind_typevar(
@@ -6353,7 +6355,7 @@ impl<'db> Type<'db> {
                     )
                     .unwrap_or(*self))
                 }
-                SpecialFormType::TypeAlias => Ok(Type::Dynamic(DynamicType::TodoTypeAlias)),
+                SpecialFormType::TypeAlias => Ok(*self),
                 SpecialFormType::TypedDict => Err(InvalidTypeExpressionError {
                     invalid_expressions: smallvec::smallvec_inline![
                         InvalidTypeExpression::TypedDict
@@ -7490,6 +7492,10 @@ pub enum KnownInstanceType<'db> {
     /// A single instance of `typing.TypeAliasType` (PEP 695 type alias)
     TypeAliasType(TypeAliasType<'db>),
 
+    /// A single instance of a PEP 613 type alias (in other words, an arbitrary type form at
+    /// runtime.)
+    TypeAlias(TypeAliasType<'db>),
+
     /// A single instance of `warnings.deprecated` or `typing_extensions.deprecated`
     Deprecated(DeprecatedInstance<'db>),
 
@@ -7514,7 +7520,7 @@ fn walk_known_instance_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
         KnownInstanceType::TypeVar(typevar) => {
             visitor.visit_type_var_type(db, typevar);
         }
-        KnownInstanceType::TypeAliasType(type_alias) => {
+        KnownInstanceType::TypeAliasType(type_alias) | KnownInstanceType::TypeAlias(type_alias) => {
             visitor.visit_type_alias_type(db, type_alias);
         }
         KnownInstanceType::Deprecated(_) | KnownInstanceType::ConstraintSet(_) => {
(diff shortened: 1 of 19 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -11702,7 +11785,7 @@ pub(super) fn determine_upper_bound<'db>(
 // Make sure that the `Type` enum does not grow unexpectedly.
 #[cfg(not(debug_assertions))]
 #[cfg(target_pointer_width = "64")]
-static_assertions::assert_eq_size!(Type, [u8; 16]);
+static_assertions::assert_eq_size!(Type, [u8; 24]);
```
AlexWaygood, line 11788: I don't understand why the old assertion no longer passes after rebasing this PR on `main`. (It seemed to pass on earlier versions of this PR, IIRC.) This PR adds new `Type` variants, but the wrapped data is all Salsa-interned, and (from local experimentation -- I tried adding some more `static_assertions::assert_eq_size!()` calls on various wrapped structs and ran `cargo check -p ty_python_semantic --release`) doesn't appear particularly large

crates/ty_python_semantic/src/types.rs
```diff
@@ -11785,7 +11785,7 @@ pub(super) fn determine_upper_bound<'db>(
...
```
MichaReiser, line 11788: hmm...

AlexWaygood: see https://github.com/astral-sh/ruff/pull/20107#discussion_r2435798449

crates/ty_python_semantic/src/types.rs
MichaReiser: I'm not sure what changed but I remember that @ibraheemdev ran into this as well. 

it seems that `SubclassOfType` is 16 bytes and so is `KnownInstanceType`. But `KnownInstanceType` was 12 bytes before.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7490,6 +7492,10 @@ pub enum KnownInstanceType<'db> {
     /// A single instance of `typing.TypeAliasType` (PEP 695 type alias)
     TypeAliasType(TypeAliasType<'db>),
 
+    /// A single instance of a PEP 613 type alias (in other words, an arbitrary type form at
+    /// runtime.)
+    TypeAlias(TypeAliasType<'db>),
```
MichaReiser, line 7497: `TypeAliasType` could be a `salsa::Supertype`

AlexWaygood:
hah, I got excited about the Codspeed speedup there for a minute before I remembered that that's the project that now panics in mypy_primer 🤣

crates/ty_python_semantic/src/types.rs
AlexWaygood: I tried it locally, it doesn't help. `Type` is still 192 bits on a release build, not 128

AlexWaygood:
Notes on the ecosystem diagnostics:
- The bokeh ones in `src/bokeh/transform.py` are really weird, need to look into them more
- Lots of true positives from `int()` calls due to the fact that we didn't understand typeshed's type alias it used for `int.__new__` before
- Lots of ddtrace hits due to the fact that we now understand the return type of `sys.exc_info()` properly... the typeshed signature there is accurate (it says this function returns `tuple[type[BaseException] | None, BaseException | None, TracebackType | None]`, but we could possibly special-case this function to model that it will always return `tuple[type[BaseException], BaseException, TracebackType]` if you're inside an `except:` block (this is where the hits are coming from)
- The new dulwich diagnostics in `dulwich/ignore.py` (and I think some of the others in that repo too) are false positives that stem from a combination of https://github.com/astral-sh/ty/issues/1100 and some `@Todo` types that we infer when we see a subscription of an object with an intersection type. Here's a minimal repro on this branch:

   ```py
   from typing import TypeVar

   AnyStr = TypeVar("AnyStr", str, bytes)
	
   def escape(x: AnyStr) -> AnyStr:
       return x
	
   def f(x: bytes):
       if x != b"foo":
           reveal_type(x)  # bytes & ~Literal[b"foo"]
           reveal_type(x[1:])  # @Todo(Subscript expressions on intersections)
           reveal_type(escape(x[1:]))  # str
   ```
- Some diagnostics on `graphql` that look like this which seem like they're technically-speaking true positives (type aliases being used in runtime contexts), but don't have very friendly error messages right now:
   ```
   Invalid class base with type `Unknown | typing.TypeAlias` 
   ```

   Edit: or, _are_ these true positives? I guess typeshed does this kind of thing, e.g. in https://github.com/python/typeshed/blob/11c7821a79a8ab7e1982f3ab506db16f1c4a22a9/stdlib/sys/__init__.pyi#L104... I'm not totally sure whether typeshed is doing the Right Thing there though (hmm, I think it was me who wrote that typeshed type alias 🙈)
- Also some `graphql` diagnostics on e.g. `src/graphql/language/printer.py` that definitely look like false positives, and should be looked into
- We seem to infer the type of `np.float32` as being `object` on this branch? Seems probably incorrect. Is causing _many_ false-positive diagnostics on jax/pandas/sklearn/scipy -- there are many diagnostics like this:
   ```
   Object of type `object` is not callable 
   ```
- This diagnostic on `kornia` requires special-casing by ty because it's true that you can do `isinstance(x, typing.List)` at runtime, but that's not expressible in typeshed's stubs (and typeshed doesn't try to capture it):

   ```
   Argument to function `isinstance` is incorrect: Expected `type | tuple[Unknown, ...]`, found `typing.List`
   ```

   This doesn't really come up that much, especially now that these legacy `typing`-module aliases are all deprecated; we could add that special-casing as a followup. There are some similar diagnostics in `pip` too, though, regarding `typing.Callable`, and in `prefect` regarding `typing.Dict`.
- The win32 errors on `com/win32comext/shell/demos/walk_shell_folders.py` look weird. (Lots of diagnostics like that across the pywin32 project.) Not sure why it thinks that `shell.error` isn't a subclass of `Exception`, since in typeshed's stubs for pywin32 [`error`](https://github.com/python/typeshed/blob/11c7821a79a8ab7e1982f3ab506db16f1c4a22a9/stubs/pywin32/win32comext/shell/shell.pyi#L7) is a type alias for `com_error`, which [definitely inherits from `Exception` in typeshed's stubs](https://github.com/python/typeshed/blob/main/stubs/pywin32/win32/lib/pywintypes.pyi#L17). Maybe this is another example of a type-alias-in-runtime-context issue?
- This one on Sphinx is an issue where we need to use type context to avoid promoting literals when solving the type parameters for a `set` literal (cc. @ibraheemdev):

   ```
   [error] invalid-argument-type - [:177:50] - Argument to bound method `__init__` is incorrect: Expected `AbstractSet[Literal["", "env", "epub", "gettext", "html", "applehelp", "devhelp"]]`, found `frozenset[Unknown | str]`
   ```

carljm:
Closing in favor of #21394

PR #20109, opened 2025-08-27
[`ruff`] Fix false negative for empty f-strings in `deque` calls (`RUF037`)

### LINKED ISSUE #20050: RUF037 false negative on `deque(f"{""}")`
### Summary

[`unnecessary-empty-iterable-within-deque-call` (RUF037)](https://docs.astral.sh/ruff/rules/unnecessary-empty-iterable-within-deque-call/) checks whether an f-string is empty in the sense of being the empty literal `f""`, whereas it should should check whether it evaluates to an empty iterable. That is, it should be implemented in terms of `is_empty_f_string` instead of `is_empty_literal`. Example:
```console
$ cat >ruf037.py <<'# EOF'
from collections import deque
deque(f"{""}")
# EOF

$ ruff --isolated check ruf037.py --select RUF037 --preview
All checks passed!
```

### Version

danparizher:
## Summary

Fixes #20050


crates/ruff_linter/src/rules/ruff/rules/unnecessary_literal_within_deque_call.rs
```diff
@@ -102,7 +103,13 @@ pub(crate) fn unnecessary_literal_within_deque_call(checker: &Checker, deque: &a
         }
         Expr::StringLiteral(string) => string.value.is_empty(),
         Expr::BytesLiteral(bytes) => bytes.value.is_empty(),
-        Expr::FString(fstring) => fstring.value.is_empty_literal(),
+        Expr::FString(_) => {
```
ntBre: Did you try an approach using [`is_empty_f_string`](https://github.com/astral-sh/ruff/blob/3927b0c9318f4b9b3c71127de5030aefb8390ed5/crates/ruff_python_ast/src/helpers.rs#L1409) as suggested on the issue? That seems a bit cleaner if it would accomplish the same thing.

danparizher: Didn't see that, thanks!

ntBre (approved):
Thanks!

PR #20111, opened 2025-08-27
[ty] Infer slightly more precise types for comprehensions

AlexWaygood:

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -75,7 +75,7 @@ a: tuple[()] = (1, 2)
 # error: [invalid-assignment] "Object of type `tuple[Literal["foo"]]` is not assignable to `tuple[int]`"
 b: tuple[int] = ("foo",)
 
-# error: [invalid-assignment] "Object of type `tuple[list[Unknown], Literal["foo"]]` is not assignable to `tuple[str | int, str]`"
+# error: [invalid-assignment]
 c: tuple[str | int, str] = ([], "foo")
 ```
 
--- a/crates/ty_python_semantic/resources/mdtest/directives/assert_never.md
+++ b/crates/ty_python_semantic/resources/mdtest/directives/assert_never.md
@@ -34,12 +34,6 @@ def _():
     assert_never(None)  # error: [type-assertion-failure]
 
 def _():
-    assert_never([])  # error: [type-assertion-failure]
-
-def _():
-    assert_never({})  # error: [type-assertion-failure]
-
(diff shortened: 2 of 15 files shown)


AlexWaygood:
The conformance test suite diff is good -- we're now correctly emitting an error on [this line](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/aliases_implicit.py#L109)

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -75,7 +75,7 @@ a: tuple[()] = (1, 2)
 # error: [invalid-assignment] "Object of type `tuple[Literal["foo"]]` is not assignable to `tuple[int]`"
 b: tuple[int] = ("foo",)
 
-# error: [invalid-assignment] "Object of type `tuple[list[Unknown], Literal["foo"]]` is not assignable to `tuple[str | int, str]`"
+# error: [invalid-assignment]
```
AlexWaygood, line 78: asserting the full error message here is now slightly more complicated because Todo types are rendered differently in release builds, and our mdtest regex that usually deals with that didn't seem able to handle a Todo type nested inside two pairs of square brackets. It didn't seem essential for us to assert the full error message here, so I just simplified the test

crates/ty_python_semantic/resources/mdtest/directives/assert_never.md
```diff
@@ -33,12 +33,6 @@ def _():
 def _():
     assert_never(None)  # error: [type-assertion-failure]
 
-def _():
-    assert_never([])  # error: [type-assertion-failure]
-
-def _():
-    assert_never({})  # error: [type-assertion-failure]
-
```
AlexWaygood, line 41: `<!-- snapshot-diagnostics -->` is enabled in this section, and it gets complicated to have snapshots for diagnostics that talk about Todo types because of the fact that Todo types are rendered differently in release builds. It seemed like we already had good test coverage of `assert_never()` diagnostics in this section even without these lines, so I just removed them

AlexWaygood:
I haven't done a detailed analysis of the primer report -- I can do if we want, but from a brief skim through it looks like basically what we'd expect. We're getting more diagnostics because we now understand that dict comprehensions create dicts and list comprehensions create lists, and dicts and lists (unlike `Unknown`) do not have all arbitrary attributes. It doesn't look like we're running into any false positives that would be fixed by bidirectional inference, for example -- specializing these containers with dynamic types works around that issue.

sharkdp (approved):
This makes sense to me!

PR #19579, opened 2025-07-27
[ty] Evaluate reachability of non-definitely-bound to Ambiguous

Glyphack:
## Summary

closes https://github.com/astral-sh/ty/issues/692

If the expression (or any child expressions) is not definitely bound the reachability constraint evaluation is determined as ambiguous.

This fixes the infinite cycles panic in the following code:

```py
from typing import Literal

class Toggle:
    def __init__(self: "Toggle"):
        if not self.x:
            self.x: Literal[True] = True
```

Credit of this solution is for David.

## Test Plan

- Added a test case with too many cycle iterations panic.
- Previous tests.

PR #20114, opened 2025-08-27
[ty] Preserve qualifiers when accessing attributes on unions/intersections

sharkdp:
## Summary

Properly preserve type qualifiers when accessing attributes on unions and intersections. This is a prerequisite for https://github.com/astral-sh/ruff/pull/19579.

Also fix a completely wrong implementation of `map_with_boundness_and_qualifiers`. It now closely follows `map_with_boundness` (just above).

## Test Plan

I thought about it, but didn't find any easy way to test this. This only affected `Type::member`. Things like validation of attribute writes (where type qualifiers like `ClassVar` and `Final` are important) were already handling things correctly.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/callable.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/callable.md
@@ -398,7 +398,7 @@ def f_okay(c: Callable[[], None]):
         c.__qualname__ = "my_callable"
 
         result = getattr_static(c, "__qualname__")
-        reveal_type(result)  # revealed: Never
+        reveal_type(result)  # revealed: property
         if isinstance(result, property) and result.fset:
             c.__qualname__ = "my_callable"  # okay
 ```
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -3309,19 +3309,14 @@ impl<'db> Type<'db> {
         let name_str = name.as_str();
 
         match self {
-            Type::Union(union) => union
-                .map_with_boundness(db, |elem| {
-                    elem.member_lookup_with_policy(db, name_str.into(), policy)
-                        .place
-                })
-                .into(),
+            Type::Union(union) => union.map_with_boundness_and_qualifiers(db, |elem| {
+                elem.member_lookup_with_policy(db, name_str.into(), policy)
+            }),
 
             Type::Intersection(intersection) => intersection
-                .map_with_boundness(db, |elem| {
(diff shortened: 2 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/annotations/callable.md
```diff
@@ -398,7 +398,7 @@ def f_okay(c: Callable[[], None]):
         c.__qualname__ = "my_callable"
 
         result = getattr_static(c, "__qualname__")
-        reveal_type(result)  # revealed: Never
+        reveal_type(result)  # revealed: property
```
sharkdp, line 401: Slightly surprised that there was no TODO here. This looks fine?

crates/ty_python_semantic/src/types.rs
```diff
@@ -9743,21 +9738,20 @@ impl<'db> IntersectionType<'db> {
         let mut builder = IntersectionBuilder::new(db);
         let mut qualifiers = TypeQualifiers::empty();
 
-        let mut any_unbound = false;
-        let mut any_possibly_unbound = false;
+        let mut all_unbound = true;
+        let mut any_definitely_bound = false;
```
sharkdp, line 9742: The implementation here now follows `map_with_boundness` just above. It was wrong before (see test change in `callable.md`).

crates/ty_python_semantic/resources/mdtest/annotations/callable.md
AlexWaygood: yes, this looks fine!

crates/ty_python_semantic/src/types.rs
```diff
@@ -9743,21 +9738,20 @@ impl<'db> IntersectionType<'db> {
...
-                    any_unbound = true;
-                }
+                Place::Unbound => {}
                 Place::Type(ty_member, member_boundness) => {
-                    if member_boundness == Boundness::PossiblyUnbound {
-                        any_possibly_unbound = true;
+                    all_unbound = false;
+                    if member_boundness == Boundness::Bound {
+                        any_definitely_bound = true;
                     }
```
AlexWaygood, line 9755:
```suggestion
                    any_definitely_bound |= member_boundness == Boundness::Bound;
```

carljm (approved):
LGTM, thank you!

crates/ty_python_semantic/src/types.rs
sharkdp: I think I prefer what we have in terms of readability. Not sure if it compiles to worse code (probably not?), but this is not a hot code path anyway.

Issue #15705, opened 2025-01-24
allowed-unused-imports doesn't work with relative imports


dhduvall:
### Description

If an "unused" import is done with a relative path, `allowed-unused-imports` can't be configured to to allow it:
```shell
$ uv init --lib --name mod
$ cat >> pyproject.toml <<EOF

[tool.ruff]
lint.pyflakes.allowed-unused-imports = [
  "mod.foo.bar",
]
EOF
$ echo "from .foo import bar" > src/mod/junk.py
$ echo "bar = None" > src/mod/foo.py
$ uvx ruff check src
src/mod/junk.py:1:18: F401 [*] `.foo.bar` imported but unused
  |
1 | from .foo import bar
  |                  ^^^ F401
  |
  = help: Remove unused import: `.foo.bar`

Found 1 error.
[*] 1 fixable with the `--fix` option.
$ uvx ruff version
ruff 0.9.3 (90589372d 2025-01-23)
```
I was unable to find a spelling of `mod.foo.bar` that would work—none of `.foo.bar`, `foo.bar`, `foo` work.
If `junk.py` contains `from mod.foo import bar` (i.e., absolute path for the import), `ruff` allows the import as expected.

TaKO8Ki:
@ntBre I have confirmed that my PR #20115 does not solve this issue, so I will work on this.

Issue #4656, opened 2023-05-25
F401 unused import false positive


sbrugman:
False negative
```
import importlib # this package is unused, but not detected
import importlib.util 

package = "ruff"
is_present = importlib.util.find_spec(package)
```
When `from importlib import util` is used, correct behaviour is observed
```
import importlib # this package is unused, and detected
from importlib import util 

package = "ruff"
is_present = util.find_spec(package)
```

charliermarsh:
This is consistent with Pyflakes and I think it's intentional, but submodule imports are really complex, I'm not certain whether it's safe in the general case to mark `import a` as unused in:

```py
import a
import a.b

a.b.c()
```

It might be safe? I believe we had a PR long ago that addressed this: #58.

(Here's a good example of why / how submodules can be difficult, though it's not exactly the same as above: https://github.com/asottile/scratch/wiki/Why-doesn't-flake8-mark-this-as-an-unused-import%3F.)

sbrugman:
Thanks for sharing the resource.

For what it's worth PyCharm does flag `import a` as unused import in the example above.

Can we think of examples where we would like to treat that example differently from this:

```python
import b
import a.b

a.b.c()
```
(The case in the linked resource above is the inverse)

konstin:
I'm struggling to find a case where `import a.b` is semantically different from `import a; import a.b` given that `import a.b` also runs `a/__init__.py` (https://docs.python.org/3/reference/import.html#the-meta-path, fourth paragraph)

charliermarsh:
Yeah, it might be the case that if we have `import a; import a.b` and then `a.b` is used but not _just_ `a`, then we can mark `import a` as unused.

On the other hand, per Anthony's linked note, if we have `import a; import a.b`, and then `a.z.x` is used, it seems like we can't actually treat `import a.b` as unused, since it _could_ modify `a`.

Similarly, if you have `import a; import a.b`, and then you do `print(a); print(a.b.c)`, I don't think we can treat either import as unused.

pkulev:
> On the other hand, per Anthony's linked note, if we have import a; import a.b, and then a.z.x is used, it seems like we can't actually treat import a.b as unused, since it could modify a.

Oh, indeed. I think that if something happens on import time, like filling some global variables from other modules (handlers registration or some `__init_subclass__` registration magic) it must be visible, and `# noqa: import for side-effects` is a good indicator of such things. I would prefer throwing a bunch of noqa's for such rare cases, than considering it normal.

PR #20115, opened 2025-08-27
[`pyflakes`] Fix `allowed-unused-imports` matching for top-level modules (`F401`)

### LINKED ISSUE #19664: `allowed-unused-imports` doesn't work for top-level modules
### Summary

```console
$ ruff version
ruff 0.12.7

$ ruff check --isolated --config 'lint.pyflakes.allowed-unused-imports = ["os"]' <(echo 'import os')
/dev/fd/63:1:8: F401 [*] `os` imported but unused
  |
1 | import os
  |        ^^ F401
  |
  = help: Remove unused import: `os`

Found 1 error.

$ ruff check --isolated --config 'lint.pyflakes.allowed-unused-imports = ["os"]' <(echo 'import os.path')
/dev/fd/63:1:8: F401 [*] `os.path` imported but unused
  |
1 | import os.path
  |        ^^^^^^^ F401
  |
  = help: Remove unused import: `os.path`

Found 1 error.

$ ruff check --isolated --config 'lint.pyflakes.allowed-unused-imports = ["os.path"]' <(echo 'import os.path')
All checks passed!

```

### Version

ruff 0.12.7

TaKO8Ki:
## Summary

Fixes #19664

Fix allowed unused imports matching for top-level modules.

I've simply replaced `from_dotted_name` with `user_defined`. Since QualifiedName for imports is created in crates/ruff_python_semantic/src/imports.rs, I guess it's acceptable to use `user_defined` here. Please tell me if there is better way.

https://github.com/astral-sh/ruff/blob/0c5089ed9e180da7bc76733ffe1a1d132e9142b0/crates/ruff_python_semantic/src/imports.rs#L62

## Test Plan

I've added a snapshot test `f401_allowed_unused_imports_top_level_module`.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
@@ -4,7 +4,9 @@ Test: allowed-unused-imports-top-level-module
 
 # No errors
 
-import hvplot.pandas
-import hvplot.pandas.plots
-from hvplot.pandas import scatter_matrix
-from hvplot.pandas.plots import scatter_matrix
+def f():
+    import hvplot
+    import hvplot.pandas
+    import hvplot.pandas.plots
+    from hvplot.pandas import scatter_matrix
+    from hvplot.pandas.plots import scatter_matrix
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
@@ -1,0 +1,10 @@
+"""
+Test: allowed-unused-imports-top-level-module
+"""
+
+# No errors
+
+import hvplot.pandas
+import hvplot.pandas.plots
+from hvplot.pandas import scatter_matrix
+from hvplot.pandas.plots import scatter_matrix
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
@@ -6,7 +6,11 @@ Test: allowed-unused-imports-top-level-module
 
 def f():
     import hvplot
+def f():
     import hvplot.pandas
+def f():
     import hvplot.pandas.plots
+def f():
     from hvplot.pandas import scatter_matrix
+def f():
     from hvplot.pandas.plots import scatter_matrix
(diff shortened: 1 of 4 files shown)


crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
```diff
@@ -0,0 +1,10 @@
+"""
+Test: allowed-unused-imports-top-level-module
+"""
+
+# No errors
+
+import hvplot.pandas
```
ntBre: Should we add a case that's just `import hvplot`? I see by comparing to the [playground](https://play.ruff.rs/aafcb007-2f0f-45f9-b75e-a5d46732e7a0) that your change is having the intended effect, but that might be a nice one to throw in since it's closest to the original report.

Based on another [quirk](https://github.com/astral-sh/ruff/issues/4656) of F401, and as you can also see on the playground, only a couple of these actually trigger F401 on `main`. We might have to wrap each test case in a function so that F401 actually tries to trigger in each case:

```py
def f(): import hvplot
def f(): import hvplot.pandas
# ...
```

TaKO8Ki:
@ntBre 

Thank you for the review. I have added a simple test case `import hvplot` and wrapped all test cases in a function.

> Do you want to test if this fixes https://github.com/astral-sh/ruff/issues/15705 as well? There's no need to extend the PR if not, I just thought it sounded related and had a chance of fixing it already.

I want to try it and I guess it takes some time to fix it, so can I work on it in a different pull request?

crates/ruff_linter/resources/test/fixtures/pyflakes/F401_35.py
```diff
@@ -0,0 +1,12 @@
...
+"""
+
+# No errors
+
+def f():
+    import hvplot
+    import hvplot.pandas
+    import hvplot.pandas.plots
+    from hvplot.pandas import scatter_matrix
+    from hvplot.pandas.plots import scatter_matrix
```
ntBre: Sorry, I actually meant to wrap each one in a separate function:


```suggestion
def f():
    import hvplot
def f():
    import hvplot.pandas
def f():
    import hvplot.pandas.plots
def f():
    from hvplot.pandas import scatter_matrix
def f():
    from hvplot.pandas.plots import scatter_matrix
```

Otherwise each import interferes with the others, like in the first example reported in #4656. Wrapping each import in its own function ensures that the `allowed-unused-import` setting allows all of these cases.

TaKO8Ki: I see. I have added separate functions.

PR #20117, opened 2025-08-27
Move GitLab output rendering to `ruff_db`

ntBre:
## Summary

This PR is a first step toward adding a GitLab output format to ty. It converts the `GitlabEmitter` from `ruff_linter` to a `GitlabRenderer` in `ruff_db` and updates its implementation to handle non-Ruff files and diagnostics without primary spans. I tried to break up the changes here so that they're easy to review commit-by-commit, or at least in groups of commits:
- preparatory changes in-place in `ruff_linter`
- moving the code over with no implementation changes mixed in
- tidying up the code now in `ruff_db`

This wasn't strictly necessary, but I also added some `Serialize` structs instead of calling `json!` to make it a little clearer that we weren't modifying the schema.

I plan to follow this up with a separate PR exposing this output format in the ty CLI, which should be quite straightforward.

## Test Plan

Existing tests, especially the two that show up in the diff as renamed nearly without changes


crates/ruff_db/src/diagnostic/render/gitlab.rs
```diff
@@ -0,0 +1,204 @@
...
+    project_path.hash(&mut hasher);
+
+    hasher.finish()
+}
+
+/// Convert an absolute path to be relative to the specified project root.
+fn relativize_path_to<P: AsRef<Path>, R: AsRef<Path>>(path: P, project_root: R) -> String {
+    format!(
+        "{}",
+        pathdiff::diff_paths(&path, project_root)
```
ntBre, line 172: I suspect that we could just use our `render::relativize_path` helper here and drop this dependency,  but I kept it around for now. We don't have any existing tests covering the `CI_PROJECT_DIR` case.

crates/ruff_db/src/diagnostic/render/gitlab.rs
```diff
@@ -0,0 +1,75 @@
...
+        diagnostics: &[Diagnostic],
+    ) -> std::fmt::Result {
+        write!(
+            f,
+            "{:#}",
+            serde_json::json!(SerializedMessages {
+                diagnostics,
+                resolver: self.resolver,
+                project_dir: std::env::var("CI_PROJECT_DIR").ok().as_deref(),
+            })
```
BurntSushi, line 43: How come `serde_json::json!` here instead of `serde_json::to_string`? I'd expect the latter to be faster since it doesn't have to construct an intermediate `serde_json::Value`, but that's just a wild guess.

BurntSushi: I guess it's a shame we can't just stream this right to a `std::fmt::Write` without going through some intermediate value first.

crates/ruff_db/src/diagnostic/render/gitlab.rs
BurntSushi: Dropping a dependency sounds great. :-)

crates/ruff_db/src/diagnostic/render/gitlab.rs
ntBre: `json!` hides the `unwrap` call in the macro and we used it in the other JSON formats, but those aren't very good reasons :grimacing: 

But yeah, the `Emitter`s in `ruff_linter` take `std::io::Write`s instead of `std::fmt::Write`s, making this a little easier. We added an adapter in the `junit` PR, but I haven't used it any of the other formats yet:

https://github.com/astral-sh/ruff/blob/ef0586d0da0f04b73e0f53b375497c99eeb6ef18/crates/ruff_db/src/diagnostic/render/junit.rs#L150-L154

ntBre:
Thanks for the review!

> Possible idea for next time: if you have 3 logical groupings of commits, it might make sense to just squash that into 3 commits. Then you can write a commit message for that grouping and it should hopefully make review easier. (And avoid needing to write out the links for the individual groups of commits. Which I assume will go bad if you rebase.)

Will do, sorry about that. I thought each commit would be helpful initially, but then they got a bit too small. I should have squashed after identifying the groups. It's definitely a pain to maintain the links!

crates/ruff_db/src/diagnostic/render/gitlab.rs
ntBre: I switched it over to `to_string` for now!

crates/ruff_db/src/diagnostic/render/gitlab.rs
ntBre: I think I'll open a follow-up issue here. I'm _pretty_ sure that we can drop this entirely and just use the CWD-relative path, but I'm not 100% sure and don't want to break GitLab users.

It's a [one-function](https://docs.rs/pathdiff/0.2.3/src/pathdiff/lib.rs.html#43-86) dependency, so we could also vendor it. The behavior seems a bit different from our `relativize_path` if the `project_root` isn't a prefix, if we really need that.

Issue #20112, opened 2025-08-27
PTH119 and PTH120 should be marked unsafe


dscorbett:
### Summary

The fixes for [`os-path-basename` (PTH119)](https://docs.astral.sh/ruff/rules/os-path-basename/) and [`os-path-dirname` (PTH120)](https://docs.astral.sh/ruff/rules/os-path-dirname/) should be marked unsafe because the replacements are [not always equivalent](https://docs.python.org/3/library/pathlib.html#comparison-to-the-os-and-os-path-modules). In particular, `pathlib` normalizes paths by collapsing adjacent slashes and removing trailing slashes, but `os.path.basename` and `os.path.dirname` don’t. [Example](https://play.ruff.rs/f67d4912-aa4a-4266-a449-a3ecfb10ef0c):
```console
$ cat >pth119_pth120.py <<'# EOF'
import os
path = "a//b/c/"
print(os.path.basename(path))
print(os.path.dirname(path))
# EOF

$ python pth119_pth120.py

a//b/c

$ ruff --isolated check pth119_pth120.py --select PTH119,PTH120 --preview --fix
Found 2 errors (2 fixed, 0 remaining).

$ cat pth119_pth120.py
import os
import pathlib
path = "a//b/c/"
print(pathlib.Path(path).name)
print(pathlib.Path(path).parent)

$ python pth119_pth120.py
c
a/b
```

### Version

ruff 0.12.10 (c68ff8d90 2025-08-21)

PR #20118, opened 2025-08-27
[`flake8-use-pathlib`] Make `PTH119` and `PTH120` fixes unsafe because they can change behavior

chirizxc:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/20112

## Test Plan

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
@@ -53,7 +53,6 @@ pub(crate) fn is_pure_path_subclass_with_preview(
 /// We check functions that take only 1 argument,  this does not apply to functions
 /// with `dir_fd` argument, because `dir_fd` is not supported by pathlib,
 /// so check if it's set to non-default values
-#[allow(clippy::too_many_arguments)]
 pub(crate) fn check_os_pathlib_single_arg_calls(
     checker: &Checker,
     call: &ExprCall,


dscorbett:
Another normalization that could be worth mentioning is that `pathlib` eliminates `"."` components, as in `"a/./b"` → `"a/b"`.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_dirname.rs
```diff
@@ -29,7 +35,16 @@ use ruff_python_ast::ExprCall;
 /// ```
 ///
 /// ## Fix Safety
-/// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
+/// This rule's fix is always marked as unsafe because the replacement is not always semantically
+/// equivalent to the original code. In particular, `pathlib` performs path normalization,
+/// which can alter the result compared to `os.path.dirname`. For example:
```
dylwil3:
```suggestion
/// which can alter the result compared to `os.path.dirname`. For example, this normalization:
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_basename.rs
```diff
@@ -62,12 +77,44 @@ pub(crate) fn os_path_basename(checker: &Checker, call: &ExprCall, segments: &[&
...
+
+    let mut diagnostic = checker.report_diagnostic(OsPathBasename, call.func.range());
+
+    if has_unknown_keywords_or_starred_expr(&call.arguments, &["p"]) {
+        return;
+    }
+
+    if !is_fix_os_path_basename_enabled(checker.settings()) {
+        return;
+    }
```
dylwil3: Can we move this shared code somewhere? Could this logic be moved to the body of `check_os_pathlib_single_arg_calls`?

chirizxc: yes

chirizxc:
@dylwil3 idk if we could use `#[allow(clippy::too_many_arguments)]`, but this way we don't duplicate the fn`s for safe and unsafe fixes

dylwil3:
> @dylwil3 idk if we could use `#[allow(clippy::too_many_arguments)]`, but this way we don't duplicate the fn`s for safe and unsafe fixes

Clippy doesn't seem to complain locally if I remove the `allow` - we'll see if CI passes 😄

chirizxc:
hmm, it's weird, i had something like 7>6, but if CI passes, it's okay!

PR #20121, opened 2025-08-27
[ty] Variant of #20076 (simpler materialization of specialized generics)

JelleZijlstra:
- **[ty] Add Top[] and Bottom[] special forms, replacing top_materialization_of() function**
- **tests for invalid use**
- **Type relations for top and bottom**
- **all tests pass**
- **clippy**
- **renames, more tests**
- **more tests, found a bug**
- **reformat**
- **one approach**
- **more tests and fix a bug**
- **fix**
- **feedback**
- **simpler version that does not work**


crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -286,16 +287,30 @@ impl<'db> ClassBase<'db> {
         specialization: Option<Specialization<'db>>,
     ) -> Self {
         if let Some(specialization) = specialization {
-            self.apply_type_mapping_impl(
+            let new_self = self.apply_type_mapping_impl(
```
carljm, line 290: Nice find!

carljm (approved):
This seems like a win to me. The materialization kind is part of the specialization; it seems reasonable that `apply_specialization` methods would be responsible for handling it.

I think if we wanted to invest in better error-resilience here (if we find ourselves forgetting to apply the materialization in more cases), the approach would be to extract a Rust type that holds only the types part of a `Specialization`, and have `TypeMappling::Specialization` hold only that Rust type, so you'd have to extract that and the materialization from a `Specialization` in order to wrap it in a `TypeMapping`. This would make it impossible to ignore the materialization part accidentally.

PR #20124, opened 2025-08-28
[ty] add cycle detection for find_legacy_typevars

carljm:
## Summary

Add cycle detection to the `find_legacy_typevars` type method.

## Test Plan

Added mdtest that stack overflowed without this.


AlexWaygood (approved):

dcreager (approved):
This is great, thank you!

PR #20125, opened 2025-08-28
[ty] add support for cyclic legacy generic protocols

carljm:
## Summary

Just add the necessary Salsa cycle handling.

## Test Plan

Added mdtest.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -2339,6 +2339,28 @@ class Bar(Protocol[S]):
 z: S | Bar[S]
 ```
 
+### Recursive legacy generic protocol
+
+```py
+from typing import Generic, TypeVar, Protocol
+
+T = TypeVar("T")
+
+class P(Protocol[T]):
+    attr: "P[T] | T"
+
+class A(Generic[T]):
+    attr: T
+
+class B(A[P[int]]):
+    pass
+
+def f(b: B):
+    reveal_type(b)  # revealed: B
+    reveal_type(b.attr)  # revealed: P[int]
+    reveal_type(b.attr.attr)  # revealed: P[int] | int
+```
+
 ### Recursive generic protocols with property members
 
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -2452,6 +2452,30 @@ from ty_extensions import reveal_protocol_interface
 reveal_protocol_interface(Foo)
 ```
 
+## Cycles
```
AlexWaygood: Rather than creating a new `## Cycles` section, could you put this new test case in the existing `## Recursive protocols` section here?

https://github.com/astral-sh/ruff/blob/18eaa659c1ea9a03bee798e161d2f2db454e154f/crates/ty_python_semantic/resources/mdtest/protocols.md?plain=1#L2045

crates/ty_python_semantic/src/types.rs
```diff
@@ -6576,6 +6576,24 @@ impl<'db> VarianceInferable<'db> for Type<'db> {
...
+) -> salsa::CycleRecoveryAction<Type<'db>> {
+    salsa::CycleRecoveryAction::Iterate
+}
+
+fn apply_specialization_cycle_initial<'db>(
+    _db: &'db dyn Db,
+    _self: Type<'db>,
+    _specialization: Specialization<'db>,
+) -> Type<'db> {
+    Type::Never
```
AlexWaygood, line 6594: is `Never` the best starting point here, or would we finish cycle recovery faster if we returned `self`?

carljm: Hmm, it's a good question. My instinct is that returning `self` would only help us resolve faster if `self` is actually the right answer (that is, the specialization shouldn't change anything), and I think that wouldn't be common (not even sure if we would hit the cycle in the first place if that were the case.) I don't think it's going to make much difference either way, I'd rather not devote more time right now to exploring it, and I'm more comfortable with `Never` as a starting point.

PR #20126, opened 2025-08-28
[ty] implement `typing.NewType`

oconnor663:
https://github.com/astral-sh/ty/issues/742

This is an initial draft of a working implementation of `NewType`. New test cases are in `new_types.md`, in particular:

```py
from typing_extensions import NewType

Foo = NewType("Foo", int)
Bar = NewType("Bar", Foo)

Foo(42)
Foo(Foo(42))       # allowed: `Foo` is a subtype of `int`.
Foo(Bar(Foo(42)))  # allowed: `Bar` is a subtype of `int`.
Foo(True)          # allowed: `bool` is a subtype of `int`.
Foo("fourty-two")  # error: [invalid-argument-type]

def f(_: int): ...
f(42)
f(Foo(42))
f(Bar(Foo(42)))

def g(_: Foo): ...
g(42)  # error: [invalid-argument-type]
g(Foo(42))
g(Bar(Foo(42)))

def h(_: Bar): ...
h(42)       # error: [invalid-argument-type]
h(Foo(42))  # error: [invalid-argument-type]
h(Bar(Foo(42)))
```

@AlexWaygood suggested a substantially different approach involving a new `Type::NewTypeInstance` variant rather than what this PR currently does, which is adding a new `NominalInstanceType::NewType` variant. I haven't yet tried that, and I don't have much intuition of my own about which is better, but I had this close to working, and I wanted to get some early feedback on it before I tried to do an overhaul.

That said, my baby daughter was born this evening (happy, healthy :heart:),  so it will be a few days or weeks before I get back to this.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -731,20 +731,20 @@ impl<'db> Type<'db> {
 
     fn is_none(&self, db: &'db dyn Db) -> bool {
         self.into_nominal_instance()
-            .is_some_and(|instance| instance.class(db).is_known(db, KnownClass::NoneType))
+            .and_then(|instance| instance.class_if_not_newtype(db))
+            .is_some_and(|class| class.is_known(db, KnownClass::NoneType))
     }
 
     fn is_bool(&self, db: &'db dyn Db) -> bool {
         self.into_nominal_instance()
-            .is_some_and(|instance| instance.class(db).is_known(db, KnownClass::Bool))
+            .and_then(|instance| instance.class_if_not_newtype(db))
+            .is_some_and(|class| class.is_known(db, KnownClass::Bool))
     }
 
     pub(crate) fn is_notimplemented(&self, db: &'db dyn Db) -> bool {
-        self.into_nominal_instance().is_some_and(|instance| {
-            instance
-                .class(db)
-                .is_known(db, KnownClass::NotImplementedType)
-        })
+        self.into_nominal_instance()
+            .and_then(|instance| instance.class_if_not_newtype(db))
+            .is_some_and(|class| class.is_known(db, KnownClass::NotImplementedType))
     }
 
     pub(crate) fn is_object(&self, db: &'db dyn Db) -> bool {
@@ -1932,8 +1932,13 @@ impl<'db> Type<'db> {
                     return C::unsatisfiable(db);
                 }
 
-                let class_literal = instance.class(db).class_literal(db).0;
-                C::from_bool(db, is_single_member_enum(db, class_literal))
+                C::from_bool(
+                    db,
+                    instance.class_if_not_newtype(db).is_some_and(|class| {
+                        let class_literal = class.class_literal(db).0;
+                        is_single_member_enum(db, class_literal)
+                    }),
+                )
             }
             _ => C::unsatisfiable(db),
         }
@@ -2232,7 +2237,7 @@ impl<'db> Type<'db> {
             // (<https://github.com/rust-lang/rust/issues/129967>)
             (Type::ProtocolInstance(protocol), nominal @ Type::NominalInstance(n))
             | (nominal @ Type::NominalInstance(n), Type::ProtocolInstance(protocol))
-                if n.class(db).is_final(db) =>
+                if n.class_ignoring_newtype(db).is_final(db) =>
             {
                 visitor.visit((self, other), || {
                     any_protocol_members_absent_or_disjoint(db, protocol, nominal, visitor)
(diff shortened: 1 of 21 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -6817,6 +6900,9 @@ impl<'db> KnownInstanceType<'db> {
                         field.default_type(self.db).display(self.db).fmt(f)?;
                         f.write_str("]")
                     }
+                    KnownInstanceType::NewType(declaration) => {
+                        f.write_str(declaration.name(self.db))
```
AlexWaygood, line 6904: I think this should be something like

```suggestion
                        write!(f, "<NewType pseudo-class '{}'>, declaration.name(self.db))
```

So that the display of a `NewType` pseudo-class clearly differentiates it from an "instance of that pseudo-class". Otherwise the output of `reveal_type` for something like this is quite confusing:

```py
from typing import NewType, reveal_type

X = NewType("X", int)
reveal_type(X)         # We reveal `X` here
reveal_type(X(42))     # W also reveal `X` here currently, but the object returned by `X(42)`
                       # inhabits a very different type in our model to the object `X` itself!
```

AlexWaygood:
the typing conformance diff all looks good! New diagnostics being emitted on lines that have `# E:` on them

AlexWaygood:
It's a big ecosystem diff but all the ones I've looked at so far are true positives, which is sort-of shocking -- I guess lots of people out there are just misusing `NewType`?!

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -886,6 +887,27 @@ declare_lint! {
     }
 }
 
+declare_lint! {
+    /// ## What it does
+    /// Checks for the creation of invalid `NewType`s
+    ///
+    /// ## Why is this bad?
+    /// There are several requirements that you must follow when creating a `NewType`.
```
AlexWaygood, line 895: I think if I were a user I would immediately have the question here "...but _why_ must I follow these requirements?" 😆 I think we should either explain a bit more clearly here why the rules around `NewType` are the way they are, or link to some docs elsewhere that give that explanation

crates/ty_python_semantic/src/types.rs
```diff
@@ -7185,6 +7271,158 @@ impl<'db> FieldInstance<'db> {
...
+/// x = Foo(42)
+/// ```
+///
+/// The revealed types there are:
+/// - `NewType`: `Type::ClassLiteral(ClassLiteral)` with `KnownClass::NewType`.
+/// - `Foo`: `Type::KnownInstance(KnownInstanceType::NewType(NewTypeInstance { .. }))`
+/// - `x`: `Type::NominalInstance(...(NominalInstanceInner::NewType(NewTypeInstance { .. }))`
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
+pub struct NewTypeInstance<'db> {
```
AlexWaygood, line 7290: I think it would be better to simply name this struct `NewType`, since it can both appear in the type `Type::KnownInstance(KnownInstanceType::NewType(..))` and in the type `Type::NominalInstance(NominalInstanceType { inner: NominalInstanceInner::NewType(..) })`.

The first of those is a type which means "a set of Python objects with exactly length one, where the sole inhabitant of the set is an object returned by a specific all to `NewType()`". The second of those is a type which means "a set of Python objects with unknown size, where all inhabitants of the set are said to be 'instances' of a specific `NewType` pseudo-class". The first is therefore similar conceptually to a `Type::ClassLiteral` type, and the second is similar conceptually to a "normal" nominal-instance type, but using the word "Instance" in the name of this struct implies that it can only ever be used as the wrapped data in instance types, which isn't true.

I think there's also enough code here that it's worth moving the struct into a `newtypes.rs` submodule -- `types.rs` is already huge, and I think we'd like to make it smaller.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7185,6 +7271,158 @@ impl<'db> FieldInstance<'db> {
...
+impl<'db> NewTypeInstance<'db> {
+    // Walk the `NewTypeBase` chain to find the underlying `ClassType`.
+    fn base_class_type(self, db: &'db dyn Db) -> ClassType<'db> {
+        let mut base = self.base(db);
+        loop {
+            match base {
+                NewTypeBase::ClassType(class) => return class,
+                NewTypeBase::NewType(newtype) => base = newtype.base(db),
+            }
+        }
```
AlexWaygood, line 7314: could shave off a few lines here by making it recursive?

```suggestion
        match self.base(db) {
            NewTypeBase::ClassType(class) => return class,
            NewTypeBase::NewType(newtype) => newtype.base_class_type(db),
        }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7185,6 +7271,158 @@ impl<'db> FieldInstance<'db> {
...
+                // Either `self` is `other`, or it's a (transitive) NewType wrapper of it.
+                return true;
+            }
+            match candidate_newtype.base(db) {
+                NewTypeBase::NewType(base_newtype) => candidate_newtype = base_newtype,
+                // Classes can't inherit from NewTypes, so if we reach the base `ClassType` without
+                // seeing `other`, then we can't be a subtype of it.
+                NewTypeBase::ClassType(_) => return false,
+            }
+        }
```
AlexWaygood, line 7381: similarly here, I'd be tempted to use recursion, unless you think this will be faster? We generally use recursion for this kind of thing elsewhere

```suggestion
        if self == other {
            // Either `self` is `other`, or it's a (transitive) NewType wrapper of it.
            return true;
        }
        match self.base(db) {
            NewTypeBase::NewType(base_newtype) => base_new_type.is_subtype_of(db, other),
            // Classes can't inherit from NewTypes, so if we reach the base `ClassType` without
            // seeing `other`, then we can't be a subtype of it.
            NewTypeBase::ClassType(_) => false,
        }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7185,6 +7271,158 @@ impl<'db> FieldInstance<'db> {
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, get_size2::GetSize)]
+pub enum NewTypeBase<'db> {
+    ClassType(ClassType<'db>),
+    NewType(NewTypeInstance<'db>),
+}
+
+impl<'db> NewTypeBase<'db> {
+    fn to_type(self, db: &'db dyn Db) -> Type<'db> {
```
AlexWaygood, line 7418:
```suggestion
    fn to_instance(self, db: &'db dyn Db) -> Type<'db> {
```

to make it clear that this method creates the type that represents "the set containing all possible instances of this newtypebase", not the type "the singleton set that only contains the runtime object representing this newtypebase itself"

AlexWaygood (commented):
Nice! I'm afraid I still suspect that having a `Type::NewTypeInstance()` variant will probably be cleaner and more robust here, but it looks like a lot of the logic here would probably be reusable if we decide to go with that approach

oconnor663:
Closing in favor of https://github.com/astral-sh/ruff/pull/21157.

PR #20128, opened 2025-08-28
[ty] No reachability analysis for implicit instance attributes

sharkdp:


crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
```diff
@@ -820,6 +820,8 @@ impl ReachabilityConstraints {
     }
 
     fn analyze_single(db: &dyn Db, predicate: &Predicate) -> Truthiness {
+        let _span = tracing::trace_span!("analyze_single", ?predicate).entered();
```
sharkdp, line 823: I have added and removed this so many times... it's obviously useful, so I guess there's no harm in leaving it.

sharkdp:
> Branch-dependent attributes are much more common in stub files than in runtime code. I wonder if it would make sense to add a diagnostic that _enforces_ that you should never have any implicit instance attributes in a stub file. I.e., since we will now have different behaviour for the two different snippets […]

I don't see any reason why someone would declare the type of instance attributes the complicated way *in a stub*, if there is a much easier way to just declare it on the body of the class. The implicit syntax wouldn't enable any use cases that wouldn't be covered by a simple `attribute: type` declaration on the body, I think? So I would argue that this inconsistency is nothing we need to immediately worry about, especially if there is a Ruff lint that would "catch" it.

AlexWaygood:
> I don't see any reason why someone would declare the type of instance attributes the complicated way _in a stub_, if there is a much easier way to just declare it on the body of the class.

there's no reason, no! But, folks do strange things sometimes -- I'd wager money that there's _a_ stub file out there _somewhere_ that does something like that 😆

PR #20130, opened 2025-08-28
[ty] Unpack variadic argument type in specialization

dhruvmanila:
## Summary

This PR fixes various TODOs around overload call when a variadic argument is used.

The reason this bug existed is because the specialization wouldn't account for unpacking the type of the variadic argument.

This is fixed by expanding `MatchedArgument` to contain the type of that argument _only_ when it is a variadic argument. The reason is that there's a split for when the argument type is inferred -- the non-variadic arguments are inferred using `infer_argument_types` _after_ parameter matching while the variadic argument type is inferred _during_ the parameter matching. And, the `MatchedArgument` is populated _during_ parameter matching which means the unpacking would need to happen during parameter matching.

This split seems a bit inconsistent but I don't want to spend a lot of time on trying to merge them such that all argument type inference happens in a single place. I might look into it while adding support for `**kwargs`.

## Test Plan

Update existing tests by resolving the todos.

The ecosystem changes looks correct to me except for the `slice` call but it seems that it's unrelated to this PR as we infer `slice[Any, Any, Any]` for a `slice(1, 2, 3)` call on `main` as well ([playground](https://play.ty.dev/9eacce00-c7d5-4dd5-a932-4265cb2bb4f6)).

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -1975,7 +1975,6 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
         }
         let matched_argument = &mut self.argument_matches[argument_index];
         matched_argument.parameters.push(parameter_index);
-        tracing::debug!("argument_type: {:?}", argument_type);
         matched_argument.types.push(argument_type);
         matched_argument.matched = true;
         self.parameter_matched[parameter_index] = true;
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1969,6 +1975,8 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
         }
         let matched_argument = &mut self.argument_matches[argument_index];
         matched_argument.parameters.push(parameter_index);
+        tracing::debug!("argument_type: {:?}", argument_type);
```
carljm: Do we want to leave this in?
```suggestion
```

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2304,6 +2334,20 @@ pub struct MatchedArgument {
     /// elements must have been successfully matched. (That means that this can be `false` while
     /// the `parameters` field is non-empty.)
     pub matched: bool,
+
+    /// The types of a variadic argument when it's unpacked. For example, given a `*args` whose
+    /// type is `tuple[A, B, C]`, this would be a 3 element vector containing the 3 types.
```
carljm: It looks to me like the length of this vector is always the same as the length of the `parameters` vector -- that is, these are the types assigned to each matched parameter. This isn't necessarily the same as the number of types in the argument type (which might not be a fixed-length iterable). I think this comment could be clearer about this?

carljm (approved):
Looks great, thank you!

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: Nope, thanks for catching that!

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: Thanks, I've expanded the comment to include this and some additional notes around the fact that this is only populated for a variadic argument.

dhruvmanila: I've also renamed the field from `types` to `unpacked_variadic_types`

dcreager:
> This split seems a bit inconsistent but I don't want to spend a lot of time on trying to merge them such that all argument type inference happens in a single place.

I like how you found a simpler solution for this — I was worried this would be a much more invasive change!

Note that doing argument inference in a single place will require being able to infer the type of a particular expression as both a value form and a type form. Right now we assume that each expression is inferred to a single type, and so we have to _choose_ whether it should be a value form type or a type form type. That's why we're currently doing the argument type inference in two places — for normal arguments, we have to match the argument against a parameter, to see if the parameter is annotated as a `TypeForm` (which we support synthetically but not via the actual `TypeForm` Python syntax yet), which tells us whether to call `infer_expression` or `infer_type_expression` for that argument. For variadic arguments, we need to know the type of the iterable before we perform parameter matching, which is why we perform that inference early. (That means that we don't currently support splatting an argument into a `TypeForm` parameter!)

So the ideal end state for this, I think, would be to update `TypeInferenceBuilder` to let us track the value-form and type-form inferred types separately. At the moment we only allow ourselves to infer a single type for an expression, and panic if we try to infer more than one. With this change, we would panic if we try to infer more than one value-form type, or more than one type-form type, but would allow ourselves to infer one of each without panicking.

PR #20133, opened 2025-08-28
[ty] Benchmarks for problematic implicit instance attributes cases

sharkdp:

### DIFF
--- a/crates/ruff_benchmark/benches/ty.rs
+++ b/crates/ruff_benchmark/benches/ty.rs
@@ -450,9 +450,6 @@ fn benchmark_complex_constrained_attributes_2(criterion: &mut Criterion) {
                     r#"
                     class C:
                         def f(self: "C"):
-                            self.a = ""
-                            self.b = ""
-
                             if isinstance(self.a, str):
                                 return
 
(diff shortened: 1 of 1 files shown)


crates/ruff_benchmark/benches/ty.rs
```diff
@@ -450,9 +450,6 @@ fn benchmark_complex_constrained_attributes_2(criterion: &mut Criterion) {
                     r#"
                     class C:
                         def f(self: "C"):
-                            self.a = ""
-                            self.b = ""
```
sharkdp, line 454: Apparently something changed here, because previously it was able to reproduce this with the attributes at the beginning of the method. Now it only seems to reproduce if they are defined at the end, after all `return`s.

crates/ruff_benchmark/benches/ty.rs
```diff
@@ -466,6 +463,56 @@ fn benchmark_complex_constrained_attributes_2(criterion: &mut Criterion) {
                                 return
                             if isinstance(self.b, str):
                                 return
+                            if isinstance(self.b, str):
+                                return
+                            if isinstance(self.b, str):
+                                return
```
sharkdp, line 469: We reduced this so much that it wasn't actually showing a problematic runtime anymore. So I added two more clauses.

PR #20140, opened 2025-08-28
[ty] Add more specific diagnostics for unsupported comparison diagnostics

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/921

We now emit an extra info diagnostic explaining why the rich comparison is not supported.

## Test Plan

Add tests to `crates/ty_python_semantic/resources/mdtest/comparison/instances/rich_comparison.md`


MatthewMckee4:
in sympy The type of `Poly().__eq__` is `(Unknown, Unknown, /) -> Unknown` (due to the use of a decorator which changes the signature of the instance method). So that is why we get those primer hits.

This is quite unfortunate

MatthewMckee4:
I'm unsure about the other mypy primer hits here

carljm:
Thanks for the PR! 

Regarding the sympy diagnostics: an annotation of `Unknown` should never result in an error, if there is any possible annotation it could be replaced with that would not be an error. I haven't looked at the PR code changes in detail yet, but if we are emitting lots of new errors due to this signature full of `Unknown`, that suggests this PR is doing something incorrect (maybe using a subtype check where it should be using an assignability check? Not sure)

I think it's clear that we can't land this without fixing that -- let me know if you need help figuring out the cause

MatthewMckee4:
The problem I was trying to show was that the type was `(Unknown, Unknown, /) -> Unknown` and not `(Unknown, /) -> Unknown`

carljm:
Oh, sorry, I missed that! That makes sense. I think it's a case of https://github.com/astral-sh/ty/issues/491 then. The fact that this comes up with decorated methods where the decorator is annotated to return a `Callable` does increase the priority on doing something different there.

In that case maybe we can land this with those new diagnostics -- the same issue already affects any normal method, this just makes it show up in a few more cases with dunder comparison methods.

crates/ty_python_semantic/resources/mdtest/comparison/instances/rich_comparison.md
```diff
@@ -389,3 +391,145 @@ reveal_type(A() == A())  # revealed: Literal[True]
...
+reveal_type(A() > A())  # revealed: Unknown
+# error: [unsupported-operator]
+reveal_type(A() <= A())  # revealed: Unknown
+# error: [unsupported-operator]
+reveal_type(A() >= A())  # revealed: Unknown
+
+# Fine
+A() == 1
+A() != 1
+```
```
carljm, line 453: This appears to be an exact copy of the previous test, it doesn't seem to show "possibly not callable"

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -7968,6 +7968,37 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                                            }
+                                            CallErrorKind::BindingError => {
+                                                diagnostic.info(format_args!(
+                                                    "Operator '{}' on object of type '{error_left_ty_display}' cannot be used with object of type '{error_right_ty_display}'",
+                                                    error.op,
+                                                ));
+                                            }
+                                            CallErrorKind::PossiblyNotCallable => {
+                                                diagnostic.info(format_args!(
+                                                    "Operator '{}' is possibly not callable on object of type '{error_left_ty_display}'",
```
carljm: I think when we are talking about something (possibly) not being callable or bound, then we need to be speaking in terms of the dunder method, not the operator. It doesn't make sense to say "operator '==' is not callable", it makes sense to say "method `__eq__` is not callable"

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -8522,36 +8553,50 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 db,
                 op.dunder(),
                 &mut CallArguments::positional([right]),
-                policy,
+                MemberLookupPolicy::MRO_NO_OBJECT_FALLBACK,
```
carljm: This change looks a bit suspicious -- it seems like we now ignore the passed-in `policy`, except that we consider it below in some code that now looks wrong (since we are now always using "no-object-fallback" policy here).

I think we need to determine whether this method needs to accept a passed-in policy (in which case we should ensure that policy is correct at the call sites and we should not override it here) or not (in which case we should just remove the argument).

crates/ty_python_semantic/resources/mdtest/comparison/instances/rich_comparison.md
MatthewMckee4: I have removed this now

crates/ty_python_semantic/src/types/infer.rs
MatthewMckee4: yeah makes sense, I have updated the diagnostic now

carljm:
It looks like this is now panicking on an ecosystem project?

MatthewMckee4:
There are a few places that call expect, expecting eq to never error, I would suspect that to bean issue, on things like 'in' checks.

Seems I need to update some of that code.

I can't check the actual error for a few days

MatthewMckee4:
Yeah, that's exactly what it is
```
fatal[panic] Panicked at crates/ty_python_semantic/src/types/infer/builder.rs:9495:18 when checking `/tmp/mypy_primer/projects/operator/ops/framework.py`: `infer_binary_type_comparison should never return None for `CmpOp::Eq`: CompareUnsupportedError { op: Eq, left_ty: NominalInstance(NominalInstanceType(NonTuple(NonGeneric(ClassLiteral { name: Name("Handle"), body_scope: ScopeId { [salsa id]: Id(14407), file: File(System("/tmp/mypy_primer/projects/operator/ops/framework.py")), file_scope_id: FileScopeId(7) }, known: None, deprecated: None, type_check_only: false, dataclass_params: None, dataclass_transformer_params: None })))), right_ty: NominalInstance(NominalInstanceType(NonTuple(NonGeneric(ClassLiteral { name: Name("NoneType"), body_scope: ScopeId { [salsa id]: Id(21148), file: File(Vendored(VendoredPathBuf("stdlib/types.pyi"))), file_scope_id: FileScopeId(222) }, known: Some(NoneType), deprecated: None, type_check_only: false, dataclass_params: None, dataclass_transformer_params: None })))), reason: Some("Operator '==' from dunder method `__eq__` on object of type 'Handle' cannot be used with object of type 'None'") }`

```

carljm:
Closing this to clean up the PR queue since it's quite old -- feel free to reopen if you resurrect it!

PR #20141, opened 2025-08-28
[ty] Improve disambiguation of types via fully qualified names

AlexWaygood:
## Summary

Some small followups to #19850:
- Disambiguate classes (where appropriate) in `invalid-return-type` as well as `invalid-assignment`
- Extend the disambiguation to work with the `Display` of generic aliases

## Test Plan

Snapshot added


carljm (approved):
Thank you!

PR #20142, opened 2025-08-28
[ty] Enforce that an attribute on a class `X` must be callable in order to satisfy a member on a protocol `P`

AlexWaygood:
## Summary

Small, incremental progress towards checking the types of method members.

## Test Plan

Added an mdtest

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1704,7 +1704,11 @@ class NotSubtype:
     def m(self, x: int) -> int:
         return 42
 
+class DefinitelyNotSubtype:
+    m = None
+
 static_assert(is_subtype_of(NominalSubtype, P))
+static_assert(not is_subtype_of(DefinitelyNotSubtype, P))
 
 # TODO: should pass
 static_assert(not is_subtype_of(NotSubtype, P))  # error: [static-assert-error]
--- a/crates/ty_python_semantic/src/types/protocol_class.rs
+++ b/crates/ty_python_semantic/src/types/protocol_class.rs
@@ -560,7 +560,8 @@ impl<'a, 'db> ProtocolMember<'a, 'db> {
                 db,
                 matches!(
                     other.to_meta_type(db).member(db, self.name).place,
-                    Place::Type(_, Boundness::Bound)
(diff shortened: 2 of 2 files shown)


AlexWaygood:
> ```diff
> freqtrade (https://github.com/freqtrade/freqtrade)
> + freqtrade/data/btanalysis/bt_fileutils.py:377:22: error[invalid-argument-type] Argument to bound method `drop` is incorrect: Expected `Hashable | Index[Any]`, found `list[@Todo(list literal element type)]`
> + freqtrade/freqai/RL/BaseReinforcementLearningModel.py:371:13: error[no-matching-overload] No overload of bound method `drop` matches arguments
> + freqtrade/optimize/backtesting.py:479:32: error[invalid-argument-type] Argument to bound method `drop` is incorrect: Expected `Hashable | Index[Any]`, found `Unknown | list[@Todo(list literal element type)]`
> - Found 384 diagnostics
> + Found 387 diagnostics
> ```

For these diagnostics, it looks like our eager simplification of unions is coming up against the fact that `Hashable` fundamentally doesn't make sense as a type in Python. (I wish pandas-stubs wouldn't use it quite so much!)

The relevant overload for `pandas.DataFrame.drop()` (inherited from its base class `NDFrame`) is here: https://github.com/pandas-dev/pandas-stubs/blob/e959fc933b024f7af6ad6efb5e8f07e8772ede61/pandas-stubs/core/generic.pyi#L341-L352. We can see that it is annotated with `Hashable | Sequence[Hashable] | Index[Any]`, which I think ty then immediately simplifies to `Hashable | Index[Any]`, since `Sequence[Hashable]` is a subtype of `Hashable`. The only trouble is, `list[Unknown]` and `list[str]` are both assignable to `Sequence[Hashable]` but not to `Hashable` (because `list` overrides `__hash__` with `None` in typeshed's stubs to indicate that it is not hashable). That then leads us to reject this call [here](https://github.com/freqtrade/freqtrade/blob/1cf1d9e3d7af4c18b09e2d6e83b4da5eadfe004c/freqtrade/data/btanalysis/bt_fileutils.py#L377), even though it's clearly meant to be allowed.

I guess to fix this we could either simplify unions less or special-case `Hashable` in some way? Not sure.

carljm (approved):
Ugh, Hashable is really a problem :/

Issue #20134, opened 2025-08-28
PTH101, PTH104, PTH105, and PTH121 ignore extra arguments


dscorbett:
### Summary

The fixes for [`os-chmod` (PTH101)](https://docs.astral.sh/ruff/rules/os-chmod/), [`os-rename` (PTH104)](https://docs.astral.sh/ruff/rules/os-rename/), [`os-replace` (PTH105)](https://docs.astral.sh/ruff/rules/os-replace/), and [`os-path-samefile` (PTH121)](https://docs.astral.sh/ruff/rules/os-path-samefile/) do not check for extra arguments. This can change a program’s behavior. The fix for PTH101 should preserve the `follow_symlinks` argument if specified. All four fixes should be suppressed when the arguments don’t match the signature. [Example](https://play.ruff.rs/cea45bb6-0a5e-45ac-b24b-b07a5b82f4bf):

```console
$ cat >pth1.py <<'# EOF'
import os

# PTH101
os.close(os.open("pth1_file", os.O_CREAT | os.O_WRONLY, 0o640))
try: os.unlink("pth1_link")
except FileNotFoundError: pass
os.symlink("pth1_file", "pth1_link")
os.chmod("pth1_link", 0o600, follow_symlinks=False)
print(oct(os.stat("pth1_file").st_mode))
print(oct(os.lstat("pth1_link").st_mode))

# PTH101
try: os.chmod("pth1_file", 0o700, None, True, 1, *[1], **{"x": 1}, foo=1)
except TypeError as e: print(e)

# PTH104
try: os.rename("pth1_file", "pth1_file1", None, None, 1, *[1], **{"x": 1}, foo=1)
except TypeError as e: print(e)

# PTH105
try: os.replace("pth1_file1", "pth1_file", None, None, 1, *[1], **{"x": 1}, foo=1)
except TypeError as e: print(e)

# PTH121
try: os.path.samefile("pth1_file", "pth1_link", 1, *[1], **{"x": 1}, foo=1)
except TypeError as e: print(e)

os.unlink("pth1_file")
os.unlink("pth1_link")
# EOF

$ python pth1.py
0o100640
0o120600
chmod() takes at most 4 arguments (8 given)
rename() takes at most 4 arguments (8 given)
replace() takes at most 4 arguments (8 given)
samefile() got an unexpected keyword argument 'x'

$ ruff --isolated check pth1.py --select PTH101,PTH104,PTH105,PTH121 --preview --fix
[...]

PR #20143, opened 2025-08-28
[`flake8-use-pathlib`] Fix `PTH101`, `PTH104`, `PTH105`, `PTH121` fixes

chirizxc:
## Summary
Fixes https://github.com/astral-sh/ruff/issues/20134

## Test Plan

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_use_pathlib/PTH211.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_use_pathlib/PTH211.py
@@ -20,3 +20,4 @@ os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory=False)
 os.symlink(src="usr/bin/python", dst="tmp/python", dir_fd=None)
 
 os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory=     True    )
+os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory="nonboolean")
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__PTH211_PTH211.py.snap
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__PTH211_PTH211.py.snap
@@ -84,5 +84,15 @@ PTH211 `os.symlink` should be replaced by `Path.symlink_to`
 21 |
 22 | os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory=     True    )
    | ^^^^^^^^^^
+23 | os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory="nonboolean")
+   |
+help: Replace with `Path(...).symlink_to(...)`
+
+PTH211 `os.symlink` should be replaced by `Path.symlink_to`
+  --> PTH211.py:23:1
+   |
+22 | os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory=     True    )
+23 | os.symlink("usr/bin/python",  dst="tmp/python", target_is_directory="nonboolean")
+   | ^^^^^^^^^^
    |
 help: Replace with `Path(...).symlink_to(...)`


chirizxc:
I don't know how to add a co-author to a PR, maybe only maintainers can do that?

chirizxc:
i use this cases:
```python
import os
from pathlib import Path

# os.chmod(mode=0o600, follow_symlinks=True, path="pth1_link")
os.chmod(mode=0o600, follow_symlinks=True, path="pth1_link")
Path("pth1_link").chmod(mode=0o600)

# os.chmod("pth1_link", mode=0o600, follow_symlinks=      False                    )
os.chmod("pth1_link", mode=0o600, follow_symlinks=      False                    )
Path("pth1_link").chmod(mode=0o600, follow_symlinks=False)

# os.chmod(path="pth1_link", mode=0o600, follow_symlinks=True)
os.chmod(path="pth1_link", mode=0o600, follow_symlinks=True)
Path("pth1_link").chmod(mode=0o600)

# os.chmod("pth1_link", follow_symlinks=True, mode=0o600)
os.chmod("pth1_link", follow_symlinks=True, mode=0o600)
Path("pth1_link").chmod(mode=0o600)

# os.chmod("pth1_link", follow_symlinks=False, mode=0o600, dir_fd=None)
os.chmod("pth1_link", follow_symlinks=False, mode=0o600, dir_fd=None)
Path("pth1_link").chmod(follow_symlinks=False, mode=0o600)

# os.chmod("pth1_link", 0o600)
os.chmod("pth1_link", 0o600)
Path("pth1_link").chmod(0o600)

# os.chmod("pth1_link", 0o600, dir_fd=None)
os.chmod("pth1_link", 0o600, dir_fd=None)
Path("pth1_link").chmod(0o600)

# os.chmod("pth1_link", 0o600, follow_symlinks=False)
os.chmod("pth1_link", 0o600, follow_symlinks=False)
Path("pth1_link").chmod(0o600, follow_symlinks=False)

# os.chmod("pth1_link", 0o600, dir_fd=None, follow_symlinks=False)
os.chmod("pth1_link", 0o600, dir_fd=None, follow_symlinks=False)
Path("pth1_link").chmod(0o600, follow_symlinks=False)

# Only diagnostic
os.chmod("pth1_link", 0o600, follow_symlinks="False")
os.chmod("pth1_file", 0o700, None, True, 1, *[1], **{"x": 1}, foo=1)
os.chmod()
os.chmod("pth1_link", follow_symlinks=True)
os.chmod("pth1_link", dir_fd=None)
```

ntBre:
> I don't know how to add a co-author to a PR, maybe only maintainers can do that?

You can add a co-author in git with a special line like this in a commit message:

```
Co-authored-by: NAME <NAME@EXAMPLE.COM>
```

But I can also add that when merging, no worries!

I'll try to give this a review soon.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -198,3 +195,10 @@ pub(crate) fn has_unknown_keywords_or_starred_expr(
         None => true,
     })
 }
+
+pub(crate) fn is_optional_bool_literal(args: &Arguments, name: &str, pos: usize) -> bool {
+    match args.find_argument_value(name, pos) {
+        Some(expr) => expr.as_boolean_literal_expr().is_some(),
+        None => true,
+    }
```
ntBre: I think this can use `is_none_or`:


```suggestion
    args.find_argument_value(name, pos).is_none_or(|expr| expr.is_boolean_literal_expr())
```

and assuming that `is_boolean_literal_expr` exists. I think all of the `Expr` variants should have `as_` and `is_` methods, but I could be wrong.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -198,3 +195,10 @@ pub(crate) fn has_unknown_keywords_or_starred_expr(
         None => true,
     })
 }
+
+pub(crate) fn is_optional_bool_literal(args: &Arguments, name: &str, pos: usize) -> bool {
```
ntBre: It looks like we're often calling this either before or after a separate call to `args.find_argument_value(name, pos)`. Would it make sense just to pass in the `Expr` from that? Alternatively, maybe this could return `Some(&Expr)` or `Some(&ExprBooleanLiteral)` instead of a bool, just to avoid calling `find_argument_value` multiple times.

I don't think it's too big of a deal, just an idea.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
```diff
@@ -73,22 +78,92 @@ pub(crate) fn os_chmod(checker: &Checker, call: &ExprCall, segments: &[&str]) {
...
+    let mut diagnostic = checker.report_diagnostic(OsChmod, call.func.range());
+
+    // Needs a minimum two arguments for fix
+    if call.arguments.len() < 2 {
+        return;
+    }
+
+    if !is_fix_os_chmod_enabled(checker.settings()) {
+        return;
+    }
```
ntBre: tiny nit: I'd probably check this before the number of arguments since that check is also for the fix.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
```diff
@@ -73,22 +78,92 @@ pub(crate) fn os_chmod(checker: &Checker, call: &ExprCall, segments: &[&str]) {
...
+                    None
+                } else {
+                    Some(locator.slice(expr.range()).to_string())
+                }
+            }
+            ArgOrKeyword::Keyword(kw) => match kw.arg.as_deref() {
+                Some("mode") => Some(format!("mode={}", locator.slice(&kw.value))),
+                Some("follow_symlinks") => match kw.value.as_boolean_literal_expr() {
+                    Some(bl) if !bl.value => Some("follow_symlinks=False".to_string()),
+                    Some(_) => None,
```
ntBre: This will drop the `follow_symlink` kwarg if it's `True` right? I think we should probably just preserve it, even if it matches the default `True` value.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
```diff
@@ -73,22 +78,92 @@ pub(crate) fn os_chmod(checker: &Checker, call: &ExprCall, segments: &[&str]) {
...
+                    None => Some(format!("follow_symlinks={}", locator.slice(&kw.value))),
+                },
+                _ => None,
```
ntBre: I'm always hesitant to add a call to `unreachable!`, but just for my understanding, this _should_ be unreachable right? At most I'd say to add a `debug_assert`, but `has_unknown_keywords_or_starred_expr` above should verify all of the kwargs, if I'm following correctly.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_makedirs.rs
```diff
@@ -98,6 +98,10 @@ pub(crate) fn os_makedirs(checker: &Checker, call: &ExprCall, segments: &[&str])
         return;
     }
 
+    if !is_optional_bool_literal(&call.arguments, "exist_ok", 2) {
```
ntBre: Thinking more about this, do we really need to check for a boolean literal? It seems like any boolean or even truthy value would be fine.

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_full_name.py.snap
```diff
@@ -824,5 +825,113 @@ PTH103 `os.makedirs()` should be replaced by `Path.mkdir(parents=True)`
...
+3   + import pathlib
+4   | 
+5   | p = "/foo"
+6   | q = "bar"
+--------------------------------------------------------------------------------
+128 | os.makedirs("name", unknown_kwarg=True)
+129 | 
+130 | # https://github.com/astral-sh/ruff/issues/20134
+    - os.chmod("pth1_link", mode=0o600, follow_symlinks=      False    )
+131 + pathlib.Path("pth1_link").chmod(mode=0o600, follow_symlinks=False)
```
ntBre: This is very closely related to my other comment about this kwarg fix, but I think we should just use the whole `follow_symlinks=...` slice from the source code if possible. I don't think we really have to reformat it like this, unless I'm missing something.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
chirizxc: yea, but in other fixes we do the same thing

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
chirizxc: yes, `has_unknown_keywords_or_starred_expr` is already checking kwargs

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_full_name.py.snap
chirizxc: I think we're already doing this in the other fix:

https://github.com/astral-sh/ruff/pull/20049/files#diff-78c4eb312beda28752a62632b77bee91f7c01d6ad4c57961069d4115e299ec83R155

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
chirizxc: https://github.com/astral-sh/ruff/pull/20049/files#diff-78c4eb312beda28752a62632b77bee91f7c01d6ad4c57961069d4115e299ec83L73

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_makedirs.rs
chirizxc: I think it is still necessary to be sure of the fix, although it is strange enough if a string ("True") is passed there, even if you fix such a variant it will turn out to be: garbage on input => garbage on output

chirizxc: I think for such variants it is better not to offer autofix, because broke the source code and it will be the same broken after fixing

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
ntBre: The logic in the other rule looks a bit different to me. It's okay to drop the extra whitespace, I forgot that that probably happens even if we slice the existing keyword range. I just don't think we need to check for a boolean literal or drop the user's explicit `follow_symlinks=True`. The only time we need to check for a default argument like that is so that we know it's safe to drop the argument _if_ the `pathlib` version doesn't support it, at least that's my understanding. In general we should preserve as much of the user's code as we can.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_makedirs.rs
ntBre: If the code was broken initially because the keyword value had the wrong type, I think it's fine to preserve that. We don't check for `name` having the right type, for example. That's beyond the scope of the rule and really even Ruff. This version won't even fix a case like:

```py
import os

exist_ok = True
os.makedirs("foo", exist_ok=exist_ok)
```

It just seems like an unnecessary restriction on the fix when we can just as easily forward whatever argument they passed. The only exception would be if `pathlib` has different behavior, but they both seem to accept truthy values just fine:

```pycon
>>> from pathlib import Path
>>> Path("foo").mkdir(exist_ok=True)
>>> Path("foo").mkdir(exist_ok=[1])
>>> import os
>>> os.makedirs("foo", exist_ok=[1])
>>>
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
chirizxc: I think I know what you mean, thanks

chirizxc:
I think this good now

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
```diff
@@ -73,22 +78,85 @@ pub(crate) fn os_chmod(checker: &Checker, call: &ExprCall, segments: &[&str]) {
...
+                    None
+                } else {
+                    Some(locator.slice(expr.range()).to_string())
+                }
+            }
+            ArgOrKeyword::Keyword(kw) => match kw.arg.as_deref() {
+                Some("mode") => Some(format!("mode={}", locator.slice(&kw.value))),
+                Some("follow_symlinks") => {
+                    Some(format!("follow_symlinks={}", locator.slice(&kw.value)))
+                }
```
ntBre: Could we use the trick like in the other PR where we slice out the whole `kw`?


```suggestion
                Some("mode" | "follow_symlinks") => locator.slice(kw),
```

then we can avoid the `to_string` call for positional args too.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_symlink.rs
```diff
@@ -104,12 +104,8 @@ pub(crate) fn os_symlink(checker: &Checker, call: &ExprCall, segments: &[&str])
         return;
     };
 
-    let target_is_directory_arg = call.arguments.find_argument_value("target_is_directory", 2);
-
-    if let Some(expr) = &target_is_directory_arg {
-        if expr.as_boolean_literal_expr().is_none() {
-            return;
-        }
+    if !is_optional_bool_literal(&call.arguments, "target_is_directory", 2) {
```
ntBre: I think it should be okay to omit this check too, by the same reasoning as the thread at https://github.com/astral-sh/ruff/pull/20143#discussion_r2341568706. It looks like this is the last call of `is_optional_bool_literal`, so we may also be able to delete the helper function entirely.

ntBre (commented):
Thanks, this is looking good! Two more small things (and a conflict from the other PTH PR, it looks like) and then I think this is good to go.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_chmod.rs
chirizxc:
```
error[E0308]: `match` arms have incompatible types                                                                                                   
   --> crates\ruff_linter\src\rules\flake8_use_pathlib\rules\os_chmod.rs:132:22
    |
130 |               ArgOrKeyword::Keyword(kw) => match kw.arg.as_deref() {
    |  __________________________________________-
131 | |                 Some("mode" | "follow_symlinks") => locator.slice(kw),
    | |                                                     ----------------- this is found to be of type `&str`
132 | |                 _ => None,
    | |                      ^^^^ expected `&str`, found `Option<_>`
133 | |             },
    | |_____________- `match` arms have incompatible types
    |
    = note: expected reference `&str`
                    found enum `std::option::Option<_>
```

ntBre: Oh oops, I forgot to wrap it in `Some`, sorry! I think we can still remove `to_string` as long as we remove it from the `ArgOrKeyword::Arg` arm too.

chirizxc:
I couldn't find another e-mail on github so I used `<danparizher@users.noreply.github.com>`

PR #19758, opened 2025-08-05
[ty] Fix attribute access on `TypedDict`s

sharkdp:
## Summary

This PR fixes a few inaccuracies in attribute access on `TypedDict`s. It also changes the return type of `type(person)` to `type[dict[str, object]]` if `person: Person` is an inhabitant of a `TypedDict`  `Person`. We still use `type[Person]` as the *meta type* of Person, however (see reasoning [here](https://github.com/astral-sh/ruff/pull/19733#discussion_r2253297926)).

## Test Plan

Updated Markdown tests.

PR #19733, opened 2025-08-04
[ty] New `Type` variant for `TypedDict`

sharkdp:
## Summary

This PR adds a new `Type::TypedDict` variant. Before this PR, we treated `TypedDict`-based types as dynamic Todo-types, and I originally planned to make this change a no-op. And we do in fact still treat that new variant similar to a dynamic type when it comes to type properties such as assignability and subtyping. But then I somehow tricked myself into implementing some of the things correctly, so here we are. The two main behavioral changes are: (1) we now also detect generic `TypedDict`s, which removes a few false positives in the ecosystem, and (2) we now support *attribute* access (not key-based indexing!) on these types, i.e. we infer proper types for something like `MyTypedDict.__required_keys__`. Nothing exciting yet, but gets the infrastructure into place.

Note that with this PR, the type of (the type) `MyTypedDict` itself is still represented as a `Type::ClassLiteral` or `Type::GenericAlias` (in case `MyTypedDict` is generic). Only inhabitants of `MyTypedDict` (instances of `dict` at runtime) are represented by `Type::TypedDict`. We may want to revisit this decision in the future, if this turns out to be too error-prone. Right now, we need to use `.is_typed_dict(db)` in all the right places to distinguish between actual (generic) classes and `TypedDict`s. But so far, it seemed unnecessary to add additional `Type` variants for these as well.

~~Note: this PR crucially depends on https://github.com/salsa-rs/salsa/pull/954 by @MichaReiser. I'm not sure if we want to merge this before it has landed in salsa.~~ (this is now merged, updated the commit ref to the latest commit on salsa main)

part of https://github.com/astral-sh/ty/issues/154

## Ecosystem impact

The new diagnostics on `cloud-init` look like true positives to me.

## Test Plan

Updated and new Markdown tests

PR #20146, opened 2025-08-29
[ty] minor TypedDict fixes

carljm:
## Summary

In `is_disjoint_from_impl`, we should unpack type aliases before we check `TypedDict`. This change probably doesn't have any visible effect until we have a more discriminating implementation of disjointness for `TypedDict`, but making the change now can avoid some confusion/bugs in future.

In `type_ordering.rs`, we should order `TypedDict` near more similar types, and leave Union/Intersection together at the end of the list. This is not necessary for correctness, but it's more consistent and it could have saved me some confusion trying to figure out why I was only getting an unreachable panic when my code example included a `TypedDict` type.

## Test Plan

None besides existing tests.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1986,11 +1986,6 @@ impl<'db> Type<'db> {
 
             (Type::Dynamic(_), _) | (_, Type::Dynamic(_)) => C::unsatisfiable(db),
 
-            (Type::TypedDict(_), _) | (_, Type::TypedDict(_)) => {
-                // TODO: Implement disjointness for TypedDict
-                C::unsatisfiable(db)
-            }
-
             (Type::TypeAlias(alias), _) => {
                 let self_alias_ty = alias.value_type(db);
                 visitor.visit((self_alias_ty, other), || {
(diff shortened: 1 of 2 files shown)


AlexWaygood:
> In `to_meta_type` -- all inhabitants of a `TypedDict` type must be `dict` instances at runtime (not instances of a subclass of `dict`). So the meta type of a `TypedDict` type should be the class `dict`.

Previous discussion on this:
- https://github.com/astral-sh/ruff/pull/19733#discussion_r2252598425
- https://github.com/astral-sh/ruff/pull/19758

If we want to go back on the decisions made there, we no longer need the `Type::dunder_class()` method added in #19758; we can go back to using `Type::to_meta_type()` for the type of `td.__class__` and `type(td)` (where `td` inhabits a `TypedDict` type)

sharkdp:
> In `to_meta_type` -- all inhabitants of a `TypedDict` type must be `dict` instances at runtime (not instances of a subclass of `dict`). So the meta type of a `TypedDict` type should be the class `dict`.

I explained [here](https://github.com/astral-sh/ruff/pull/19733#discussion_r2252598425) why I think it's necessary to include information about the actual `TypedDict` type in its meta type. Also note that we do already infer `<class 'dict[str, object]'>` for `type(some_typed_dict)` (see this [MD test](https://github.com/astral-sh/ruff/blob/04dc223710e9dfeb265603b9dcc04c08fd7a3601/crates/ty_python_semantic/resources/mdtest/typed_dict.md?plain=1#L470-L483)).

With the changes in https://github.com/astral-sh/ruff/pull/19758, `type(obj)` is not necessarily equivalent with calling `Type::to_meta_type` on the type of `obj`. See also [this doc comment](https://github.com/astral-sh/ruff/blob/04dc223710e9dfeb265603b9dcc04c08fd7a3601/crates/ty_python_semantic/src/types.rs#L5869-L5870).

(oops, Alex beat me to it)

carljm:
Makes sense, thanks for the clarification and sorry for the noise! I've updated the PR to exclude the meta-type change, and instead add a comment to avoid someone making the same mistake as me in future.

PR #20147, opened 2025-08-29
[ty] ensure union normalization really normalizes

carljm:
## Summary

Now that we have `Type::TypeAlias`, which can wrap a union, and the possibility of unions including non-unpacked type aliases (which is necessary to support recursive type aliases), we can no longer assume in `UnionType::normalized_impl` that normalizing each element of an existing union will result in a set of elements that we can order and then place raw into `UnionType` to create a normalized union. It's now possible for those elements to themselves include union types (unpacked from an alias). So instead, we need to feed those elements into the full `UnionBuilder` (with alias-unpacking turned on) to flatten/normalize them, and then order them.

## Test Plan

Added mdtest.


crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -120,6 +120,36 @@ def f(x: IntOrStr, y: str | bytes):
...
+    pass
+
+class D:
+    pass
+
+class E:
+    pass
+
+class G[T]:
+    pass
```
AlexWaygood: nit: Black formats these more concisely if we use `...`, since we pass `--pyi` to blacken-docs

```suggestion
class A: ...
class B: ...
class C: ...
class D: ...
class E: ...
class G[T]: ...
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7675,7 +7673,18 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
                 TypeVarBoundOrConstraints::UpperBound(bound.normalized_impl(db, visitor))
             }
             TypeVarBoundOrConstraints::Constraints(constraints) => {
-                TypeVarBoundOrConstraints::Constraints(constraints.normalized_impl(db, visitor))
+                // Constraints are a non-normalized union by design (it's not really a union at
+                // all, we are just using a union to store the types). Normalize the types but not
+                // the containing union.
```
AlexWaygood, line 7678: And I don't think we _should_ be using a union here, really (though it's a separate issue) -- I'd wager it's the cause of this panic (though I haven't checked that theory)

https://github.com/astral-sh/ruff/blob/0d7ed32494df58f62ed6b48a3491f258ca7fd086/crates/ty_python_semantic/resources/primer/bad.txt#L16-L17

crates/ty_python_semantic/src/types.rs
```diff
@@ -7675,7 +7673,18 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
...
+                TypeVarBoundOrConstraints::Constraints(UnionType::new(
+                    db,
+                    constraints
+                        .elements(db)
+                        .iter()
+                        .map(|ty| ty.normalized_impl(db, visitor))
+                        .collect::<Vec<_>>()
+                        .into_boxed_slice(),
```
```suggestion
                        .collect::<Box<_>>(),
```

AlexWaygood (approved):

crates/ty_python_semantic/src/types.rs
carljm: could be related to constraints -- or it might be fixed by this PR. I'll check

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -120,6 +120,25 @@ def f(x: IntOrStr, y: str | bytes):
...
 
+## Multiple layers of union aliases
+
+```py
+class A: ...
+class B: ...
+class C: ...
+class D: ...
+class E: ...
+class G[T]: ...
```
carljm:
```suggestion
```

crates/ty_python_semantic/src/types.rs
AlexWaygood: the pandas-stubs commit that made us start panicking sure did add a TypeVar with a lot of constraints, soo... https://github.com/pandas-dev/pandas-stubs/commit/bf1221eb7ea0e582c30fe233d1f4f5713fce376b

carljm: It is fixed by this PR, but I think actually that could be the case even if it was caused by a constrained type var

PR #20148, opened 2025-08-29
[ty] add more lsp tests for overloads

Gankra:
I decided to split out the addition of these tests from other PRs so that it's easier to follow changes to the LSP's function call handling. I'm not particularly concerned with whether the results produced by these tests are "good" or "bad" in this PR, I'm just establishing a baseline.

### DIFF
--- a/crates/ty_ide/src/goto_definition.rs
+++ b/crates/ty_ide/src/goto_definition.rs
@@ -802,6 +802,318 @@ my_func(my_other_func(a<CURSOR>b=5, y=2), 0)
         }
     }
 
+    #[test]
+    fn goto_definition_overload_type_disambiguated1() {
+        let test = CursorTest::builder()
+            .source(
+                "main.py",
+                "
+from mymodule import ab
+
+a<CURSOR>b(1)
+",
+            )
+            .source(
+                "mymodule.py",
+                r#"
+def ab(a):
+    """the real implementation!"""
+"#,
+            )
+            .source(
+                "mymodule.pyi",
+                r#"
+from typing import overload
+
+@overload
+def ab(a: int): ...
+
+@overload
+def ab(a: str): ...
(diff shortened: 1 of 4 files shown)


Gankra:
(Each of the 4 files received a copy of the same 6 tests)

crates/ty_ide/src/goto_definition.rs
```diff
@@ -802,6 +802,318 @@ my_func(my_other_func(a<CURSOR>b=5, y=2), 0)
...
+a<CURSOR>b(1)
+",
+            )
+            .source(
+                "mymodule.py",
+                r#"
+def ab(a):
+    """the real implementation!"""
+"#,
+            )
```
dhruvmanila, line 822: Should this have the overloads defined so that the test asserts that goto definition goes to the correct overload? I'm fine with it not being present, I mainly want to understand the reasoning behind the test cases in this file as they all have defined only the implementation.

Gankra: Hmm interesting, would you expect a .pyi *and* a .py to declare overloads? (they're purely type annotations right? there isn't an implementation to go to for an overload? or did I get myself mixed up)

dhruvmanila: Hmm, yeah that's a good point. It's not required for the overload definitions to be present in both runtime and stub files as they're mainly for typing purposes (as you've stated as well). So, I think my main question would be do we want to test the goto definition functionality by defining the overloads in the runtime file?

Gankra: I think ultimately goto-definition just has boring answers here, but it's worth making sure stub mapping doesn't get confused by the situation (that said it's possible we want like, 3 different variations of every one of these tests but... for now let's not).

PR #20150, opened 2025-08-29
[ty] Better error message for attempting to assign to a read-only property

AlexWaygood:
## Summary

The error message we give if somebody attempts to assign to a read-only property is currently not very user-friendly: https://play.ty.dev/599eb199-504a-4fff-8ee9-585880838b5e. But this is a common error that is worth special casing so that we present a better error message. https://github.com/astral-sh/ruff/pull/19936 (my work-in-progress PR for `@property` protocol members) also threatens to make this error message more prominent in some situations, so it seems worth cleaning it up now.

## Test Plan

Mdtests


sharkdp (approved):
Very nice — thank you!

PR #20152, opened 2025-08-29
Less confidently mark f-strings as empty when inferring truthiness

### LINKED ISSUE #19935: `is_empty_f_string` causes false positives in PT015
### Summary

[`is_empty_f_string`](https://github.com/astral-sh/ruff/blob/0.12.9/crates/ruff_python_ast/src/helpers.rs#L1409) categorizes some f-string interpolations as empty strings which are non-empty strings, causing false positives in [`pytest-assert-always-false` (PT015)](https://docs.astral.sh/ruff/rules/pytest-assert-always-false/). [Example](https://play.ruff.rs/8d4dd448-d037-46ff-9a4b-e73fca1e7153):
```console
$ cat >pt015.py <<'# EOF'
assert f"{b""}"
assert f"{""=}"
assert f"{""!a}"
assert f"{""!r}"
assert f"{"":1}"
# EOF

$ python pt015.py; echo $?
0

$ ruff --isolated check pt015.py --select PT015 --output-format concise -q
pt015.py:1:1: PT015 Assertion always fails, replace with `pytest.fail()`
pt015.py:2:1: PT015 Assertion always fails, replace with `pytest.fail()`
pt015.py:3:1: PT015 Assertion always fails, replace with `pytest.fail()`
pt015.py:4:1: PT015 Assertion always fails, replace with `pytest.fail()`
pt015.py:5:1: PT015 Assertion always fails, replace with `pytest.fail()`

$ sed 's/assert \(.*\)/print(\1 + f": {len(\1)}")/' pt015.py | python
b'': 3
""='': 5
'': 2
'': 2
 : 1
```

### Version

dylwil3:
When computing the boolean value of an f-string, we over-eagerly interpreted some f-string interpolations as empty. In this PR we now mark the truthiness of f-strings involving format specs, debug text, and bytes literals as "unknown".

This will probably result in some false negatives, which may be further refined (for example - there are probably many cases where `is_not_empty_f_string` should be modified to return `true`), but for now at least we should have fewer false positives.

Affected rules (may not be an exhaustive list):

- [unnecessary-empty-iterable-within-deque-call (RUF037)](https://docs.astral.sh/ruff/rules/unnecessary-empty-iterable-within-deque-call/#unnecessary-empty-iterable-within-deque-call-ruf037)
- [falsy-dict-get-fallback (RUF056)](https://docs.astral.sh/ruff/rules/falsy-dict-get-fallback/#falsy-dict-get-fallback-ruf056)
- [pytest-assert-always-false (PT015)](https://docs.astral.sh/ruff/rules/pytest-assert-always-false/#pytest-assert-always-false-pt015)
- [expr-or-not-expr (SIM221)](https://docs.astral.sh/ruff/rules/expr-or-not-expr/#expr-or-not-expr-sim221)
- [expr-or-true (SIM222)](https://docs.astral.sh/ruff/rules/expr-or-true/#expr-or-true-sim222)
- [expr-and-false (SIM223)](https://docs.astral.sh/ruff/rules/expr-and-false/#expr-and-false-sim223)

Closes #19935

### DIFF
--- a/crates/ruff_linter/src/rules/ruff/snapshots/ruff_linter__rules__ruff__tests__RUF037_RUF037.py.snap
+++ b/crates/ruff_linter/src/rules/ruff/snapshots/ruff_linter__rules__ruff__tests__RUF037_RUF037.py.snap
@@ -396,6 +396,8 @@ RUF037 [*] Unnecessary empty iterable within a deque call
 111 | # https://github.com/astral-sh/ruff/issues/20050
 112 | deque(f"{""}")  # RUF037
     | ^^^^^^^^^^^^^^
+113 |
+114 | deque(f"{b""}")
     |
 help: Replace with `deque()`
 109 | deque(t"{""}") # OK
@@ -403,3 +405,6 @@ help: Replace with `deque()`
 111 | # https://github.com/astral-sh/ruff/issues/20050
     - deque(f"{""}")  # RUF037
 112 + deque()  # RUF037
+113 | 
+114 | deque(f"{b""}")
+115 | deque(f"{""=}")


ntBre:
Thanks! I'll take a look at this after planning. Do you think we can still use `is_empty_f_string` for RUF037 (#20109), or will we need a slightly different version? I just suggested using that yesterday :laughing:

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1410,18 +1410,22 @@ pub fn is_empty_f_string(expr: &ast::ExprFString) -> bool {
     fn inner(expr: &Expr) -> bool {
         match expr {
             Expr::StringLiteral(ast::ExprStringLiteral { value, .. }) => value.is_empty(),
-            Expr::BytesLiteral(ast::ExprBytesLiteral { value, .. }) => value.is_empty(),
+            // Confusingly, `bool(f"{b""}") == True` even though
```
ntBre, line 1413: This did seem confusing initially, but this cleared it up a bit for me:

```pycon
>>> f"{b""}"
"b''"
>>> str(b'')
"b''"
```

So this is actually correct for RUF037 too:

```pycon
>>> from collections import deque
>>> deque(f"")
deque([])
>>> deque(f"{b""}")
deque(['b', "'", "'"])
```

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1410,18 +1410,22 @@ pub fn is_empty_f_string(expr: &ast::ExprFString) -> bool {
...
+            // `bool(b"") == False`
+            Expr::BytesLiteral(_) => false,
             Expr::FString(ast::ExprFString { value, .. }) => {
                 value
                     .elements()
                     .all(|f_string_element| match f_string_element {
```
ntBre: Should we factor out this closure now that it's more complicated? It looks the same as below.

ntBre (approved):
Thanks! I tested all of these cases with `deque`, and I think this is a fix for RUF037 too!

```pycon
>>> deque(f"{b""}")
deque(['b', "'", "'"])
>>> deque(f"{""=}")
deque(['"', '"', '=', "'", "'"])
>>> deque(f"{""!a}")
deque(["'", "'"])
>>> deque(f"{""!r}")
deque(["'", "'"])
>>> deque(f"{"":1}")
deque([' '])
```

PR #19636, opened 2025-07-30
[ty] Remove special casing for tuple addition

AlexWaygood:

PR #20153, opened 2025-08-29
[ty] [WIP] Retry parameter matching for argument type expansion

dhruvmanila:
## Summary

This is very much WIP !!

There are multiple issues with the solution that I want to address.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/arguments.rs
+++ b/crates/ty_python_semantic/src/types/call/arguments.rs
@@ -202,10 +202,22 @@ impl<'a, 'db> CallArguments<'a, 'db> {
                     for subtype in &expanded_types {
                         let mut new_expanded_types = pre_expanded_types.to_vec();
                         new_expanded_types[index] = Some(*subtype);
-                        expanded_arguments.push(CallArguments::new(
-                            self.arguments.clone(),
-                            new_expanded_types,
-                        ));
+
+                        // Update the arguments list to handle variadic argument expansion
+                        let mut new_arguments = self.arguments.clone();
+                        if let Argument::Variadic(_) = self.arguments[index] {
+                            // If the expanded type is a tuple, update the TupleLength
+                            if let Some(expanded_type) = new_expanded_types[index] {
+                                let length = expanded_type
+                                    .try_iterate(db)
+                                    .map(|tuple| tuple.len())
+                                    .unwrap_or(TupleLength::unknown());
+                                new_arguments[index] = Argument::Variadic(length);
+                            }
+                        }
+
+                        expanded_arguments
+                            .push(CallArguments::new(new_arguments, new_expanded_types));
                     }
                 }
 
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -131,7 +131,14 @@ impl<'db> Bindings<'db> {
         argument_types: &CallArguments<'_, 'db>,
     ) -> Result<Self, CallError<'db>> {
         for element in &mut self.elements {
-            element.check_types(db, argument_types);
+            if let Some((global_argument_forms, global_conflicting_forms)) =
+                element.check_types(db, argument_types)
+            {
+                // If this element returned global forms (indicating successful argument type expansion),
+                // update the Bindings with these forms
+                self.argument_forms = global_argument_forms.into();
+                self.conflicting_forms = global_conflicting_forms.into();
+            }
         }
 
         self.evaluate_known_cases(db);
@@ -1261,7 +1268,11 @@ impl<'db> CallableBinding<'db> {
         }
     }
 
-    fn check_types(&mut self, db: &'db dyn Db, argument_types: &CallArguments<'_, 'db>) {
+    fn check_types(
+        &mut self,
+        db: &'db dyn Db,
+        argument_types: &CallArguments<'_, 'db>,
+    ) -> Option<(Vec<Option<ParameterForm>>, Vec<bool>)> {
         // If this callable is a bound method, prepend the self instance onto the arguments list
         // before checking.
         let argument_types = argument_types.with_self(self.bound_type);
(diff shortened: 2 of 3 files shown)


dhruvmanila:
(This isn't completely ready but I'd appreciate some initial feedback on the approach.)

crates/ty_python_semantic/src/types/call/arguments.rs
```diff
@@ -202,10 +202,25 @@ impl<'a, 'db> CallArguments<'a, 'db> {
...
+                            // If the argument corresponding to this type is variadic, we need to
+                            // update the tuple length because expanding could change the length.
+                            // For example, in `tuple[int] | tuple[int, int]`, the length of the
+                            // first type is 1, while the length of the second type is 2.
+                            if let Some(expanded_type) = new_expanded_types[index] {
+                                let length = expanded_type
+                                    .try_iterate(db)
+                                    .map(|tuple| tuple.len())
+                                    .unwrap_or(TupleLength::unknown());
+                                new_arguments[index] = Argument::Variadic(length);
```
dcreager, line 218: :+1: Yep this looks like a good approach! It should pull out the more precise length of each union element during expansion

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1902,10 +1909,61 @@ enum MatchingOverloadIndex {
     Multiple(Vec<usize>),
 }
 
+#[derive(Default)]
+struct ArgumentForms {
```
dcreager: I like the cleanup from introducing this new type, too

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1398,13 +1389,30 @@ impl<'db> CallableBinding<'db> {
...
+                // The spec mentions that each expanded argument list should re-evaluate from step
+                // 2 which is the type checking step but we're re-evaluating from step 1. The tldr
+                // is that it allows ty to match the correct overload in case a variadic argument
+                // would expand into different number of arguments with each expansion. Refer to
+                // https://github.com/astral-sh/ty/issues/735 for more details.
+                for overload in &mut self.overloads {
+                    // Clear the state of all overloads before re-evaluating from step 1
+                    overload.reset();
+                    overload.match_parameters(db, expanded_arguments, &mut argument_forms);
+                }
```
dcreager: :100:

carljm: I don't know if we need to reflect this in our comments, but I think the spec is not wrong here -- the difference is that in the spec "step 1" means only "eliminate impossible overloads due to arity mismatch" whereas our step 1 is bigger than that, it also includes "match arguments to parameters", which in the spec is implicitly part of step 2, not part of step 1. This is why we need to "repeat step 1" and the spec does not. It may be clearer to say that we don't separate step 1 and 2 in the same way the spec does.

dhruvmanila: That's a good point, I didn't realize this subtle difference, I'll update the comment to reflect this.

dhruvmanila:
> ```diff
> openlibrary (https://github.com/internetarchive/openlibrary)
> + openlibrary/coverstore/archive.py:422:13: error[no-matching-overload] No overload of function `__new__` matches arguments
> - Found 711 diagnostics
> + Found 712 diagnostics
> ```

I'm not exactly sure what's wrong here. I've created a playground snippet which breaks the union into it's elements which is what the type expansion would retry with (https://play.ty.dev/1990e010-195f-4800-9e31-b77894e37dd3). It's only the `c` variable which has a fixed length of 2 because the outermost type is a `tuple` containing two elements.

carljm:
@dcreager I'm assuming this is in your review queue unless you request review from me.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2281,22 +2342,6 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
-            // arbitrary number of `Unknown`s at the end, so that if the expanded value has a
-            // smaller arity than the unexpanded value, we still have enough values to assign to
-            // the already matched parameters.
-            let argument_types = match argument_types.as_ref() {
-                Tuple::Fixed(_) => {
-                    Cow::Owned(argument_types.concat(self.db, &Tuple::homogeneous(Type::unknown())))
-                }
-                Tuple::Variable(_) => argument_types,
-            };
-
```
dcreager, line 2313: :heart: Love that we can get rid of this workaround!

crates/ty_python_semantic/src/types/tuple.rs
```diff
@@ -1016,6 +1016,7 @@ impl<'db> Tuple<Type<'db>> {
     }
 
     /// Concatenates another tuple to the end of this tuple, returning a new tuple.
+    #[expect(dead_code)]
```
dcreager: I don't think there's anything we're using this for anymore, now that we're no longer special-casing `__add__` for tuples (#19636). So I'd say we can go ahead and remove it; we can always resurrect it from the git history if we need to.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -106,13 +105,11 @@ impl<'db> Bindings<'db> {
...
-        let mut argument_forms = vec![None; arguments.len()];
-        let mut conflicting_forms = vec![false; arguments.len()];
+        let mut argument_forms = ArgumentForms::new(arguments.len());
         for binding in &mut self.elements {
-            binding.match_parameters(db, arguments, &mut argument_forms, &mut conflicting_forms);
+            binding.match_parameters(db, arguments, &mut argument_forms);
         }
-        self.argument_forms = argument_forms.into();
-        self.conflicting_forms = conflicting_forms.into();
+        (self.argument_forms, self.conflicting_forms) = argument_forms.into_boxed_slice();
```
dcreager: You might consider storing the new `ArgumentForms` in `Bindings` instead of separating it out into two separate slices. That way it would be stored consistently throughout this file. That might require adding an accessor method for external callers to get access to the `argument_forms`. And you'd replace `ArgumentForms::into_boxed_slice` with `ArgumentForms::shrink_to_fit`.

dhruvmanila: That sounds reasonable, thanks!

PR #20154, opened 2025-08-29
[`pyflakes`] Fix false positives for `__annotate__` (Py3.14+) and `__warningregistry__` (`F821`)

### LINKED ISSUE #19970: F821 false positives for `__annotate__` and `__warningregistry__`
### Summary

There are two documented module-level dunder attributes that [`undefined-name` (F821)](https://docs.astral.sh/ruff/rules/undefined-name/) doesn’t recognize: [`__annotate__`](https://docs.python.org/3.14/reference/datamodel.html#other-writable-attributes-on-module-objects) and [`__warningregistry__`](https://docs.python.org/3.14/library/warnings.html#warnings.warn_explicit). They are like `__file__` and `__cached__` in that not every module has them and they are not commonly used, but if someone uses them it was probably intentional, so F821 shouldn’t flag them. [Example](https://play.ruff.rs/1ba60c54-4666-4cd7-abfb-e02fc526c6ff):
```console
$ cat >f821.py <<'# EOF'
x: str = str(b"!")
print(__warningregistry__)
print(__annotate__(1))
# EOF

$ python3.14 -b f821.py 2>/dev/null
{'version': 1, ('str() on a bytes instance', <class 'BytesWarning'>, 1): True}
{'x': <class 'str'>}

$ ruff --isolated check f821.py --select F821 --target-version py314 --preview --output-format concise -q
f821.py:2:7: F821 Undefined name `__warningregistry__`
f821.py:3:7: F821 Undefined name `__annotate__`
```

### Version

danparizher:
## Summary

Fixes #19970

### DIFF
--- a/crates/ruff_linter/src/rules/pyflakes/mod.rs
+++ b/crates/ruff_linter/src/rules/pyflakes/mod.rs
@@ -921,6 +921,19 @@ mod tests {
     }
 
     #[test]
+    fn module_warningregistry() {
+        // Using __warningregistry__ should not be considered undefined.
+        flakes("__warningregistry__", &[]);
+    }
+
+    #[test]
+    fn module_annotate_py314_plus() {
+        // __annotate__ is available starting in Python 3.14; ensure no F821 when enabled.
+        // Note: The helper uses default target version from settings; ensure CI targets latest.
+        flakes("__annotate__", &[]);
+    }
+
+    #[test]
     fn magic_globals_file() {
         // Use of the C{__file__} magic global should not emit an undefined name
         // warning.
--- a/crates/ruff_python_stdlib/src/builtins.rs
+++ b/crates/ruff_python_stdlib/src/builtins.rs
@@ -18,8 +18,10 @@ const IPYTHON_BUILTINS: &[&str] = &["__IPYTHON__", "display", "get_ipython"];
 pub const MAGIC_GLOBALS: &[&str] = &[
     "WindowsError",
     "__annotations__",
+    "__annotate__",
     "__builtins__",
     "__cached__",
+    "__warningregistry__",
     "__file__",
 ];
 
--- a/crates/ty_python_semantic/src/place.rs
+++ b/crates/ty_python_semantic/src/place.rs
@@ -1379,9 +1379,9 @@ mod implicit_globals {
         if name == "__file__" {
             Place::bound(KnownClass::Str.to_instance(db)).into()
         } else if name == "__warningregistry__" {
-            // Present when warnings machinery is engaged; treat as defined to avoid F821.
-            // Exact type is not important for undefined-name; use `Any` for resilience.
-            Place::bound(Type::any()).into()
+            // Created lazily by the warnings machinery; may be absent.
+            // Model as possibly-unbound to avoid false negatives.
+            Place::Type(Type::any(), super::Boundness::PossiblyUnbound).into()
         } else if name == "__annotate__" {
             // Added in Python 3.14; only treat as implicit when target version is >= 3.14.
             if Program::get(db).python_version(db) >= PythonVersion::PY314 {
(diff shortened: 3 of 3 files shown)


crates/ty_python_semantic/src/place.rs
ntBre, line 1: This change looks reasonable to me, but we'll want someone from the ty side to have a look too, maybe @AlexWaygood? I'm guessing they'd like you to have an mdtest for this as well. Possibly in [`moduletype_attrs.md`](https://github.com/astral-sh/ruff/blob/c3f1f0b9f1ed8e522a6d22c7d1670997c19183de/crates/ty_python_semantic/resources/mdtest/scopes/moduletype_attrs.md?plain=1#L1), but I'm not really sure.

crates/ruff_python_stdlib/src/builtins.rs
```diff
@@ -18,8 +18,10 @@ const IPYTHON_BUILTINS: &[&str] = &["__IPYTHON__", "display", "get_ipython"];
 pub const MAGIC_GLOBALS: &[&str] = &[
     "WindowsError",
     "__annotations__",
+    "__annotate__",
     "__builtins__",
     "__cached__",
+    "__warningregistry__",
```
ntBre, line 23: I think it's correct to put `__warningregistry__` here because I don't think it's available in `builtins`, but I think we should create a separate `PY314_PLUS_BUILTINS` down below like the other version-specific groups for `__annotate__` since it was introduced in 3.14.

It appears not to _really_ matter since both `MAGIC_GLOBALS` and `python_builtins` are only called once each in the same place. But we may also need to update `is_python_builtin` since its comment says it's supposed to be kept in sync with `python_builtins`.

crates/ruff_linter/src/rules/pyflakes/mod.rs
```diff
@@ -920,6 +920,19 @@ mod tests {
...
+        // Using __warningregistry__ should not be considered undefined.
+        flakes("__warningregistry__", &[]);
+    }
+
+    #[test]
+    fn module_annotate_py314_plus() {
+        // __annotate__ is available starting in Python 3.14; ensure no F821 when enabled.
+        // Note: The helper uses default target version from settings; ensure CI targets latest.
+        flakes("__annotate__", &[]);
+    }
```
ntBre: This doesn't seem to be testing what the test name suggests. As the comment says, the `flakes` test runner uses the `latest` version, which is still 3.13, so we'll need to write some custom test code here. We've been talking about switching this whole module over to the more common `test_path` and `assert_diagnostics` tests like most other rules, so I'd probably suggest just doing something like that here. One example of that:

https://github.com/astral-sh/ruff/blob/c3f1f0b9f1ed8e522a6d22c7d1670997c19183de/crates/ruff_linter/src/rules/ruff/mod.rs#L116-L125

We should also test that `__annotate__` is not available before 3.14. I think this test should actually be failing given the version being used.

danparizher:
Thanks for the detailed review!

I went ahead and version-gated magic globals by introducing `python_magic_globals` (with `__annotate__` behind `PY314_PLUS_MAGIC_GLOBALS`) and updated the linter to bind them per target version; I left `is_python_builtin` as-is since `__annotate__` isn’t a builtin.

I also reworked the pyflakes tests to explicitly check `__annotate__` on 3.13 vs 3.14, and added/adjusted ty mdtests to model `__warningregistry__` as possibly unbound and to validate the `__annotate__` gating. All tests pass locally.

ntBre (commented):
Thanks! This looks good to me on the Ruff side.

MichaReiser:
@AlexWaygood would you mind taking a look at how the changes impact ty?

AlexWaygood:
The primer report LGTM too!

PR #20155, opened 2025-08-29
[ty] Add GitLab output format

ntBre:
## Summary

This wires up the GitLab output format moved into `ruff_db` in https://github.com/astral-sh/ruff/pull/20117 to the ty CLI.

While I was here, I made one unrelated change to the CLI docs. Clap was rendering the escapes around the `\[default\]` brackets for the `full` output, so I just switched those to parentheses:

```
--output-format <OUTPUT_FORMAT>
    The format to use for printing diagnostic messages

    Possible values:
    - full:    Print diagnostics verbosely, with context and helpful hints \[default\]
    - concise: Print diagnostics concisely, one per line
    - gitlab:  Print diagnostics in the JSON format expected by GitLab Code Quality reports
```

## Test Plan

New CLI test, and a manual test with `--config 'terminal.output-format = "gitlab"'` to make sure this works  as a configuration option too. I also tried piping the output through jq to make sure it's at least valid JSON

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -1444,7 +1444,7 @@ pub enum DiagnosticFormat {
     Junit,
     /// Print diagnostics in the JSON format used by GitLab [Code Quality] reports.
     ///
-    /// [Code Quality]: https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool
+    /// [Code Quality]: https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format
     #[cfg(feature = "serde")]
     Gitlab,
 }
--- a/crates/ty_project/src/metadata/options.rs
+++ b/crates/ty_project/src/metadata/options.rs
@@ -1057,7 +1057,7 @@ pub enum OutputFormat {
     Concise,
     /// Print diagnostics in the JSON format expected by GitLab [Code Quality] reports.
     ///
-    /// [Code Quality]: https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool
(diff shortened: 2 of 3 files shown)


crates/ty/tests/cli/main.rs
```diff
@@ -618,6 +618,71 @@ fn concise_diagnostics() -> anyhow::Result<()> {
...
+    let _s = settings.bind_to_scope();
+
+    assert_cmd_snapshot!(case.command().arg("--output-format=gitlab").arg("--warn").arg("unresolved-reference"), @r#"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+    [
+      {
+        "check_name": "unresolved-reference",
+        "description": "unresolved-reference: Name `x` used when not defined",
```
ntBre, line 642: One minor note, I know this looks redundant, but GitLab doesn't render the `check_name` anywhere, so we manually prepend this to the description now (https://github.com/astral-sh/ruff/issues/19881).

crates/ty/src/lib.rs
```diff
@@ -319,37 +319,48 @@ impl MainLoop {
...
-                                "{}",
-                                "All checks passed!".green().bold()
-                            )?;
+                            if is_human_readable {
+                                writeln!(
+                                    self.printer.stream_for_success_summary(),
+                                    "{}",
+                                    "All checks passed!".green().bold()
+                                )?;
+                            }
```
dhruvmanila, line 331: I'm assuming that we do the same thing for Ruff as well?

ntBre: Yep, pretty similar. Ruff has some `Flags` for the `Printer` that toggle showing these messages.

https://github.com/astral-sh/ruff/blob/bbfcf6e111648052783bdb68ec71706fc69b714f/crates/ruff/src/printer.rs#L85-L93

PR #15550, opened 2025-01-17
[red-knot] Add support for `typing.ClassVar`

sharkdp:
## Summary

Add support for `typing.ClassVar`, i.e. emit a diagnostic in this scenario:
```py
from typing import ClassVar

class C:
    x: ClassVar[int] = 1

c = C()
c.x = 3  # error: "Cannot assign to pure class variable `x` from an instance of type `C`"
```

## Test Plan

- New tests for the `typing.ClassVar` qualifier
- Fixed one TODO in `attributes.md`

PR #20156, opened 2025-08-29
[ty] Use `invalid-assignment` error code for invalid assignments to `ClassVar`s

AlexWaygood:
## Summary

This error is about assigning to attributes rather than reading attributes, so I think `invalid-assignment` makes more sense than `invalid-attribute-access`

## Test Plan

existing mdtests updated


sharkdp:
The `invalid-attribute-access` diagnostic was sort-of intended for this specific purpose (https://github.com/astral-sh/ruff/pull/15550). I'm not necessarily opposed to this change, but I also don't think the previous diagnostic code was that bad (an access can be a read or a write, in my opinion). If we're going to keep this, can you please update the description of the `invalid-attribute-access` diagnostic, which is now outdated? Or remove it all-together?

PR #20158, opened 2025-08-29
Revert "[ty] Use `invalid-assignment` error code for invalid assignments to `ClassVar`s"

AlexWaygood:
Reverts astral-sh/ruff#20156. As @sharkdp noted in his post-merge review, there were several issues with that PR that I didn't spot before merging — but I'm out for four days now, and would rather not leave things in an inconsistent state for that long. I'll revisit this on Wednesday.


sharkdp (approved):
Thanks

PR #20166, opened 2025-08-30
[syntax-errors]: import from * only allowed at module scope (F406)

11happy:
## Summary

This PR implements F406 https://docs.astral.sh/ruff/rules/undefined-local-with-nested-import-star-usage/ as a semantic syntax error

## Test Plan

### DIFF
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -25,7 +25,6 @@ use std::cell::RefCell;
 use std::path::Path;
 
 use itertools::Itertools;
-use libcst_native::Name;
 use log::debug;
 use rustc_hash::{FxHashMap, FxHashSet};
 
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -25,6 +25,7 @@ use std::cell::RefCell;
 use std::path::Path;
 
 use itertools::Itertools;
+use libcst_native::Name;
 use log::debug;
 use rustc_hash::{FxHashMap, FxHashSet};
 
@@ -69,7 +70,8 @@ use crate::package::PackageRoot;
 use crate::preview::is_undefined_export_in_dunder_init_enabled;
 use crate::registry::Rule;
 use crate::rules::pyflakes::rules::{
-    LateFutureImport, ReturnOutsideFunction, YieldOutsideFunction,
+    LateFutureImport, ReturnOutsideFunction, UndefinedLocalWithNestedImportStarUsage,
+    YieldOutsideFunction,
 };
 use crate::rules::pylint::rules::{
     AwaitOutsideAsync, LoadBeforeGlobalDeclaration, YieldFromInAsyncFunction,
@@ -657,6 +659,14 @@ impl SemanticSyntaxContext for Checker<'_> {
             SemanticSyntaxErrorKind::YieldOutsideFunction(kind) => {
                 if self.is_rule_enabled(Rule::YieldOutsideFunction) {
                     self.report_diagnostic(YieldOutsideFunction::new(kind), error.range);
+                }
+            }
+            SemanticSyntaxErrorKind::UndefinedLocalWithNestedImportStarUsage(name) => {
+                if self.is_rule_enabled(Rule::UndefinedLocalWithNestedImportStarUsage) {
+                    self.report_diagnostic(
+                        UndefinedLocalWithNestedImportStarUsage { name },
+                        error.range,
+                    );
                 }
             }
             SemanticSyntaxErrorKind::ReturnOutsideFunction => {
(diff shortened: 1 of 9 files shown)


11happy:
I have fixed the md file test which was causing CI to fail.

Thank you

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -58,10 +57,30 @@ impl SemanticSyntaxChecker {
...
+                ..
+            }) => {
                 if self.seen_futures_boundary && matches!(module.as_deref(), Some("__future__")) {
                     Self::add_error(ctx, SemanticSyntaxErrorKind::LateFutureImport, *range);
                 }
+                for alias in names {
+                    if alias.name.as_str() == "*" && !ctx.in_module_scope() {
+                        // test_err import_from_star
+                        // def f():
+                        //     from module import *
```
dhruvmanila: Let's add a few more test cases like including it in class scope, module scope, etc. to check both the error and non-error case. The non-error case would need to go under `test_ok` (instead of `test_err`).

11happy:
have resolved the merge conflict as well.
Thank you

11happy:
Hii @ntBre, would appreciate your feedback on this when you get a chance. 
Thank you : )

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -659,6 +660,14 @@ impl SemanticSyntaxContext for Checker<'_> {
                     self.report_diagnostic(YieldOutsideFunction::new(kind), error.range);
                 }
             }
+            SemanticSyntaxErrorKind::UndefinedLocalWithNestedImportStarUsage(name) => {
```
ntBre: I know this is the name of the lint rule, but I think it would make sense to rename the syntax error to something like `LocalImportStar` or `NonModuleImportStar`. For the purpose of the syntax error, we don't actually care about the `undefined-local` part, it's still a syntax error even with no local variables:

```pycon
>>> def foo():
...     from math import *
...
  File "<python-input-7>", line 2
    from math import *
                     ^
SyntaxError: import * only allowed at module level
```

crates/ruff_python_parser/src/semantic_errors.rs
ntBre: I see you added a module-scoped test, would you mind throwing in a class scope like Dhruv also suggested? It can be part of the same `test_err` block as the function:

```rust
// test_err import_from_star
// def f():
//     from module import *
// class C:
//     from module import *
```

I think that will cover all of the scopes where you can actually have an import statement.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -58,10 +57,33 @@ impl SemanticSyntaxChecker {
...
+
+                        // test_ok import_from_star
+                        // from module import *
+                        Self::add_error(
+                            ctx,
+                            SemanticSyntaxErrorKind::UndefinedLocalWithNestedImportStarUsage(
+                                module.as_deref().unwrap_or_default().to_string(),
```
ntBre: I guess we don't have any tests covering this, but this looks very different from the code used in the lint rule:

https://github.com/astral-sh/ruff/blob/f507ff8cd036fd2d7ab8ad13f1a207555bc7596d/crates/ruff_linter/src/checkers/ast/analyze/statement.rs#L815-L820

Should we be using `format_import_from` again here?

For an example test, the `level` argument for this would be 3:

```py
from ...math import *
```

I think without the helper function we would format that as just `math`.

crates/ruff_linter/src/checkers/ast/mod.rs

crates/ruff_python_parser/src/semantic_errors.rs
11happy: yes , I checked without helper function it was just math, I have used helper function likewise.

crates/ruff_python_parser/src/semantic_errors.rs

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -58,10 +58,38 @@ impl SemanticSyntaxChecker {
...
                 }
+                for alias in names {
+                    if alias.name.as_str() == "*" && !ctx.in_module_scope() {
+                        // test_err import_from_star
+                        // def f1():
+                        //     from module import *
+                        // class C:
+                        //     from module import *
+                        // def f2():
+                        //     from ..module import *
```
ntBre, line 79: Would you mind adding one more test? `from module import *, *` is already a separate syntax error, but I'm wondering how many `NonModuleImportStar` errors we should emit in this case. We don't emit any `F406`s because the `Star import must be the only import` error is a parse error, but the interaction between the two syntax errors will be interesting in the future when we make `F406` a syntax error.

With the `break` on line 90, I think we'll only emit one `NonModuleImportStar` and one parse error, which seems reasonable to me.

11happy: done

PR #20167, opened 2025-08-30
Feat/ignore deprecated rules by unless selected by exact code

### LINKED ISSUE #18349: Warn about error being related to a deprecated rule
### Summary

My case - was trying out Ruff and enabled `UP`. Going through the log of found errors, I've found that [UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/) from my point of view seems unreasonable, checking the documentation I've found that it's actually deprecated and will be removed soon.

A suggestion - `ruff check` to warn that found error is related to a deprecated rule, so it will be possible to find deprecated rules just by using `ruff`, without checking documentation. This would improve rule deprecation discoverability.

(I'm hope I'm not missing some option that already does that, but couldn't find anything related in https://docs.astral.sh/ruff/settings/)

```python
a = 25
print(isinstance(25, (str, int)))
# ruff check L:\1_test.py --select UP --target-version py311
# UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`
#   |
# 2 | a = 25
# 3 | print(isinstance(25, (str, int)))
#   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^ UP038
#   |
#   = help: Convert to `X | Y`
# 
# Found 1 error.
# No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```

LoicRiegel:
## Summary

Closes #18349

After this change:
- All deprecated rules are deselected by default
- They are only selected if the user specifically selects them by code, e.g. `--select UP038`
- Thus, `--select ALL --select UP --select UP0` won't select the deprecated rule UP038
- Documented the change in version policy. From now on, deprecating a rule should increase the minor version

## Test Plan

Integration tests in "integration_tests.rs"

Also tested with a temporary test package:
```
~> ../../ruff/target/debug/ruff.exe check --select UP038
warning: Rule `UP038` is deprecated and will be removed in a future release.
warning: Detected debug build without --no-cache.
UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`
 --> main.py:2:11
  |
1 | def main():
2 |     print(isinstance(25, (str, int)))
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
help: Convert to `X | Y`

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).

~> ../../ruff/target/debug/ruff.exe check --select UP03
warning: Detected debug build without --no-cache.
All checks passed!

~> ../../ruff/target/debug/ruff.exe check --select UP0
warning: Detected debug build without --no-cache.
All checks passed!

~> ../../ruff/target/debug/ruff.exe check --select UP
warning: Detected debug build without --no-cache.
All checks passed!

~> ../../ruff/target/debug/ruff.exe check --select ALL
# warnings and errors, but because of other errors, UP038 was deselected
```

### DIFF
--- a/crates/ruff/tests/integration_test.rs
+++ b/crates/ruff/tests/integration_test.rs
@@ -1469,8 +1469,8 @@ fn redirect_prefix() {
 }
 
 #[test]
-fn deprecated_direct() {
-    // Selection of a deprecated rule without preview enabled should still work
+fn deprecated_selected_exact() {
+    // A deprecated rule should be included if selected by exact code
     // but a warning should be displayed
     let mut cmd = RuffCheck::default().args(["--select", "RUF920"]).build();
     assert_cmd_snapshot!(cmd, @r"
@@ -1488,7 +1488,7 @@ fn deprecated_direct() {
 }
 
 #[test]
-fn deprecated_multiple_direct() {
+fn deprecated_selected_exact_multiple() {
     let mut cmd = RuffCheck::default()
         .args(["--select", "RUF920", "--select", "RUF921"])
         .build();
@@ -1511,28 +1511,21 @@ fn deprecated_multiple_direct() {
 }
 
 #[test]
-fn deprecated_indirect() {
-    // `RUF92` includes deprecated rules but should not warn
-    // since it is not a "direct" selection
+fn deprecated_selected_prefix() {
+    // A deprecated rule should not be included if selected by rule prefix
     let mut cmd = RuffCheck::default().args(["--select", "RUF92"]).build();
     assert_cmd_snapshot!(cmd, @r"
-    success: false
-    exit_code: 1
+    success: true
+    exit_code: 0
     ----- stdout -----
-    RUF920 Hey this is a deprecated test rule.
-    --> -:1:1
-
-    RUF921 Hey this is another deprecated test rule.
-    --> -:1:1
-
-    Found 2 errors.
+    All checks passed!
 
     ----- stderr -----
     ");
 }
 
 #[test]
-fn deprecated_direct_preview_enabled() {
+fn deprecated_selected_exact_preview_enabled() {
     // Direct selection of a deprecated rule in preview should fail
     let mut cmd = RuffCheck::default()
         .args(["--select", "RUF920", "--preview"])
@@ -1549,23 +1542,7 @@ fn deprecated_direct_preview_enabled() {
 }
 
 #[test]
-fn deprecated_indirect_preview_enabled() {
-    // `RUF920` is deprecated and should be off by default in preview.
-    let mut cmd = RuffCheck::default()
-        .args(["--select", "RUF92", "--preview"])
-        .build();
-    assert_cmd_snapshot!(cmd, @r"
-    success: true
-    exit_code: 0
-    ----- stdout -----
-    All checks passed!
-
-    ----- stderr -----
-    ");
-}
-
-#[test]
-fn deprecated_multiple_direct_preview_enabled() {
+fn deprecated_selected_exact_multiple_preview_enabled() {
     // Direct selection of the deprecated rules in preview should fail with
     // a message listing all of the rule codes
     let mut cmd = RuffCheck::default()
@@ -1581,6 +1558,22 @@ fn deprecated_multiple_direct_preview_enabled() {
       Cause: Selection of deprecated rules is not allowed when preview is enabled. Remove selection of:
     	- RUF920
     	- RUF921
+    ");
+}
+
+#[test]
+fn deprecated_selected_prefix_preview_enabled() {
+    // `RUF920` is deprecated and should be off by default in preview.
+    let mut cmd = RuffCheck::default()
+        .args(["--select", "RUF92", "--preview"])
+        .build();
+    assert_cmd_snapshot!(cmd, @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    All checks passed!
+
+    ----- stderr -----
     ");
 }
 
@@ -2155,16 +2148,10 @@ extend-safe-fixes = ["RUF9"]
     RUF903 Hey this is a stable test rule with a display only fix.
     --> -:1:1
 
-    RUF920 Hey this is a deprecated test rule.
-    --> -:1:1
-
-    RUF921 Hey this is another deprecated test rule.
-    --> -:1:1
-
     RUF950 Hey this is a test rule that was redirected from another.
     --> -:1:1
 
-    Found 7 errors.
+    Found 5 errors.
     [*] 1 fixable with the `--fix` option (1 hidden fix can be enabled with the `--unsafe-fixes` option).
 
     ----- stderr -----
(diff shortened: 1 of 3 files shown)


crates/ruff/tests/integration_test.rs
```diff
@@ -1469,8 +1469,8 @@ fn redirect_prefix() {
 }
 
 #[test]
-fn deprecated_direct() {
-    // Selection of a deprecated rule without preview enabled should still work
+fn deprecated_selected_exact() {
+    // A deprecated rule should be included if selected by exact code
```
ntBre: This is a pretty small nit, but I'd lean toward not changing any names or comments unless they're really outdated. I think this case, and `deprecated_multiple_direct` at least were okay with the old names.

crates/ruff/tests/integration_test.rs
```diff
@@ -1511,28 +1511,21 @@ fn deprecated_multiple_direct() {
 }
 
 #[test]
-fn deprecated_indirect() {
-    // `RUF92` includes deprecated rules but should not warn
-    // since it is not a "direct" selection
+fn deprecated_selected_prefix() {
+    // A deprecated rule should not be included if selected by rule prefix
```
ntBre: Similar to the above, I think the name here was okay. But the new comment makes sense to me given the changed test result.

crates/ruff/tests/integration_test.rs
```diff
@@ -1549,38 +1542,38 @@ fn deprecated_direct_preview_enabled() {
 }
 
 #[test]
-fn deprecated_indirect_preview_enabled() {
-    // `RUF920` is deprecated and should be off by default in preview.
+fn deprecated_selected_exact_multiple_preview_enabled() {
```
ntBre: I'd kind of prefer to preserve the existing tests and add new ones rather than modify these, unless they're now redundant with other tests. Is that the case?

ntBre:
I also labeled this breaking, as Micha mentioned [here](https://github.com/astral-sh/ruff/issues/18349#issuecomment-2929741873). But we're preparing for a minor release next week, so we can merge this into the [release branch](https://github.com/astral-sh/ruff/pull/20194) instead of `main`. GitHub warned me that changing the base branch might remove review comments, though, so I held off for now.

LoicRiegel:
Thank you for the review, I will make the changes now.
Do you want me to keep the commit history clean, or will you squash everything anyway?

ntBre:
Thank you! We'll squash everything when merging, so it's actually a bit easier to review if you push additional commits, in my opinion

LoicRiegel:
I just pushed three commits where
- I reverted the tests to the way they were
- Wrote comments for all tests concerned by this change so they are more precise and all consistent
- I swapped two tests to make the order consistent, I hope that's okay

Tell me if everything's alright now :)

crates/ruff/tests/integration_test.rs
```diff
@@ -1470,7 +1470,7 @@ fn redirect_prefix() {
 
 #[test]
 fn deprecated_direct() {
-    // Selection of a deprecated rule without preview enabled should still work
+    // A deprecated rule selected by exact code should be included
```
ntBre: I think the `without preview` part is still accurate right? I do slightly prefer your wording otherwise, but I'd probably still leave this comment alone just to minimize the diff.

crates/ruff/tests/integration_test.rs
```diff
@@ -1512,28 +1514,22 @@ fn deprecated_multiple_direct() {
...
 
     ----- stderr -----
     ");
 }
 
 #[test]
 fn deprecated_direct_preview_enabled() {
-    // Direct selection of a deprecated rule in preview should fail
+    // A deprecated rule selected by exact code should be included in preview
+    // but a warning should be displayed
```
ntBre: Yeah looks like the preview behavior is still preserved, which I think is desired. I'd say just to revert this comment change.

crates/ruff/tests/integration_test.rs
```diff
@@ -1549,38 +1545,37 @@ fn deprecated_direct_preview_enabled() {
 }
 
 #[test]
-fn deprecated_indirect_preview_enabled() {
-    // `RUF920` is deprecated and should be off by default in preview.
+fn deprecated_multiple_direct_preview_enabled() {
```
ntBre: The old test still seems relevant to me. It shows that indirect selection in preview does _not_ produce the failure that a direct selection in preview does. I think we can revert the changes to this test.

crates/ruff/tests/integration_test.rs
```diff
@@ -1549,38 +1545,37 @@ fn deprecated_direct_preview_enabled() {
...
+    	- RUF920
+    	- RUF921
     ");
 }
-
 #[test]
-fn deprecated_multiple_direct_preview_enabled() {
-    // Direct selection of the deprecated rules in preview should fail with
-    // a message listing all of the rule codes
+fn deprecated_indirect_preview_enabled() {
```
ntBre: Ohh, I see. It looks like you swapped the order of these two tests. I think the existing tests were okay, we can probably just revert this too.

ntBre (commented):
Thanks! Just a few more test suggestions to reduce the diff a bit more. I have a better understanding of the tests now and think they mostly still look relevant in their original form.

crates/ruff/tests/integration_test.rs
LoicRiegel: Is minimizing the diff such a big goal?
In my experience, it's better to allow (even encourage) small improvements like this. This helps keep the code clean and up-to-date.
If the change is too big (like renaming a variable which is used a lot across the project), I'd want it to be a proper PR/commit but in my case, the diff is small.

That said, I understand it's important in this project, so I'll comply, no problem:)

ntBre: It's not such a big deal, I just lean toward preserving the original code/comments if it's not a really clear improvement. This is just a slight phrasing improvement, so I don't think it's worth adding another step to checking the git-blame in the future, for example. In other cases I'd probably agree with you :)

crates/ruff/tests/integration_test.rs
LoicRiegel: reverted

crates/ruff/tests/integration_test.rs
LoicRiegel: reverted

crates/ruff/tests/integration_test.rs
LoicRiegel: reverted

crates/ruff/tests/integration_test.rs
LoicRiegel: reverted

LoicRiegel:
> Thanks! Just a few more test suggestions to reduce the diff a bit more. I have a better understanding of the tests now and think they mostly still look relevant in their original form.

Done :)
Now the diff is as small as it can be :)

PR #20169, opened 2025-08-30
[`flake8-use-pathlib`] Add fix for `PTH123`

chirizxc:
## Test Plan

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
@@ -1,0 +1,187 @@
+use ruff_diagnostics::{Applicability, Edit, Fix};
+use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::{Expr, ExprBooleanLiteral, ExprCall};
+use ruff_text_size::Ranged;
+
+use crate::checkers::ast::Checker;
+use crate::importer::ImportRequest;
+use crate::preview::is_fix_builtin_open_enabled;
+use crate::rules::flake8_use_pathlib::helpers::{
+    has_unknown_keywords_or_starred_expr, is_argument_non_default, is_file_descriptor,
+    is_pathlib_path_call,
+};
+use crate::{FixAvailability, Violation};
+
+/// ## What it does
+/// Checks for uses of the `open()` builtin.
+///
+/// ## Why is this bad?
+/// `pathlib` offers a high-level API for path manipulation. When possible,
+/// using `Path` object methods such as `Path.open()` can improve readability
+/// over the `open` builtin.
+///
+/// ## Examples
+/// ```python
+/// with open("f1.py", "wb") as fp:
+///     ...
+/// ```
+///
+/// Use instead:
+/// ```python
+/// from pathlib import Path
+///
+/// with Path("f1.py").open("wb") as fp:
+///     ...
+/// ```
+///
+/// ## Known issues
+/// While using `pathlib` can improve the readability and type safety of your code,
+/// it can be less performant than working directly with strings,
+/// especially on older versions of Python.
+///
+/// ## Fix Safety
+/// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
+///
+/// ## References
+/// - [Python documentation: `Path.open`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open)
+/// - [Python documentation: `open`](https://docs.python.org/3/library/functions.html#open)
+/// - [PEP 428 – The pathlib module – object-oriented filesystem paths](https://peps.python.org/pep-0428/)
+/// - [Correspondence between `os` and `pathlib`](https://docs.python.org/3/library/pathlib.html#corresponding-tools)
+/// - [Why you should be using pathlib](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/)
+/// - [No really, pathlib is great](https://treyhunner.com/2019/01/no-really-pathlib-is-great/)
+#[derive(ViolationMetadata)]
+pub(crate) struct BuiltinOpen;
+
+impl Violation for BuiltinOpen {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "`open()` should be replaced by `Path.open()`".to_string()
+    }
+
+    fn fix_title(&self) -> Option<String> {
+        Some("Replace with `Path.open()`".to_string())
+    }
+}
+
+// PTH123
+pub(crate) fn builtin_open(checker: &Checker, call: &ExprCall, segments: &[&str]) {
+    // `closefd` and `opener` are not supported by pathlib, so check if they
+    // are set to non-default values.
+    // https://github.com/astral-sh/ruff/issues/7620
+    // Signature as of Python 3.11 (https://docs.python.org/3/library/functions.html#open):
+    // ```text
+    // builtins.open(
+    //   file,          0
+    //   mode='r',      1
+    //   buffering=-1,  2
+    //   encoding=None, 3
+    //   errors=None,   4
+    //   newline=None,  5
+    //   closefd=True,  6 <= not supported
+    //   opener=None    7 <= not supported
+    // )
+    // ```
+    // For `pathlib` (https://docs.python.org/3/library/pathlib.html#pathlib.Path.open):
+    // ```text
+    // Path.open(mode='r', buffering=-1, encoding=None, errors=None, newline=None)
+    // ```
+    if call
+        .arguments
+        .find_argument_value("closefd", 6)
+        .is_some_and(|expr| {
+            !matches!(
+                expr,
+                Expr::BooleanLiteral(ExprBooleanLiteral { value: true, .. })
+            )
+        })
+        || is_argument_non_default(&call.arguments, "opener", 7)
+        || call
+            .arguments
+            .find_argument_value("file", 0)
+            .is_some_and(|expr| is_file_descriptor(expr, checker.semantic()))
+    {
+        return;
+    }
+
+    if !matches!(segments, ["" | "builtins", "open"]) {
+        return;
+    }
+
+    let range = call.range();
+    let mut diagnostic = checker.report_diagnostic(BuiltinOpen, call.func.range());
+
+    let Some(file) = call.arguments.find_argument_value("file", 0) else {
+        return;
+    };
+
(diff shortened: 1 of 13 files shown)


chirizxc:
I'm not sure which tests should be included in the tests files, but i tested in this file:

The first line with `#noqa` so that they don't turn during the fix, the second line is needed to observe how the code is fixed, and the third line is what I expect to see after the fix, and then I compare lines 2 and 3 
```python
# ruff: noqa: SIM115, PLW1514, FBT003
import builtins
from pathlib import Path

_file = "file.txt"
_x = ("r+", -1)
r_plus = "r+"

builtins.open(file=_file)  # noqa: PTH123
builtins.open(file=_file)
Path(_file).open()

open(mode="wb", file=_file)  # noqa: PTH123
open(mode="wb", file=_file)
Path(_file).open(mode="wb")

open(mode="r+", buffering=-1, file=_file, encoding="utf-8")  # noqa: PTH123
open(mode="r+", buffering=-1, file=_file, encoding="utf-8")
Path(_file).open(mode="r+", buffering=-1, encoding="utf-8")

open(_file, "r+", -1, None, None, None)  # noqa: PTH123
open(_file, "r+", -1, None, None, None)
Path(_file).open("r+", -1, None, None, None)

open(_file, "r+", -1, None, None, None, closefd=True, opener=None)  # noqa: PTH123
open(_file, "r+", -1, None, None, None, closefd=True, opener=None)
Path(_file).open("r+", -1, None, None, None)

open(_file, mode="r+", buffering=-1, encoding=None, errors=None, newline=None)  # noqa: PTH123
open(_file, mode="r+", buffering=-1, encoding=None, errors=None, newline=None)
Path(_file).open(mode="r+", buffering=-1, encoding=None, errors=None, newline=None)


open(buffering=-      1, file=_file, encoding=         "utf-8")  # noqa: PTH123
open(buffering=-      1, file=_file, encoding=         "utf-8")
Path(_file).open(buffering=-1, encoding="utf-8")

open(_file, "r+", - 1, None, None, None, True, None)  # noqa: PTH123
open(_file, "r+", - 1, None, None, None, True, None)
Path(_file).open("r+", -1, None, None, None)

open(_file, "r+ ", -  1)  # noqa: PTH123
open(_file, "r+ ", -  1)
Path(_file).open("r+ ", -1)

open(_file, f"  {r_plus}      ", -  1)  # noqa: PTH123
open(_file, f"  {r_plus}      ", -  1)
Path(_file).open(f"  {r_plus}      ", -1)

# Only diagnostic
open()
open(_file, *_x)
open(_file, "r+", unknown=True)
open(_file, "r+", closefd=False)
open(_file, "r+", None, None, None, None, None, None, None)
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -0,0 +1,207 @@
...
+        || file_arg.is_some_and(|expr| is_file_descriptor(expr, checker.semantic()))
+    {
+        return;
+    }
+
+    if !matches!(segments, ["" | "builtins", "open"]) {
+        return;
+    }
+
+    let range = call.range();
```
ntBre: Could we move this closer to where it's used? I could be missing another usage, but I only see this used right at the end of the function since we use `call.func.range()` in the `report_diagnostic` call.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -0,0 +1,207 @@
...
+        )?;
+
+        let locator = checker.locator();
+        let file_code = locator.slice(file.range());
+
+        let args = |i: usize, arg: ArgOrKeyword| {
+            let clean_expr = |expr: &Expr, s: &str| match expr {
+                Expr::StringLiteral(_) | Expr::TString(_) | Expr::FString(_) => s.to_string(),
+                _ => s.replace(char::is_whitespace, ""),
+            };
```
ntBre: Which test cases does this affect? As I mentioned on your other PTH PR, we shouldn't alter the user's code unless it's necessary. This doesn't look necessary to me. If we drop whitespace from using `locator.slice(expr)`, that's okay, but this kind of intentional reformatting seems gratuitous.

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_import_from.py.snap
```diff
@@ -650,3 +701,143 @@ help: Replace with `Path(...).rename(...)`
...
+8  | q = "bar"
+--------------------------------------------------------------------------------
+72 | 
+73 |     with open(file, "r+", -1, None, None, None) as _: ...
+74 | 
+   -     with open(
+   -             file, mode="r+", buffering=-1,
+   -             encoding="utf-8", errors=None, newline=None,
+   -     ) as _: ...
+75 +     with pathlib.Path(file).open(mode="r+", buffering=-1, encoding="utf-8", errors=None, newline=None) as _: ...
```
ntBre: I remember we discussed this before, and it wasn't a good fit for another rule, but could we use [`remove_argument`](https://github.com/astral-sh/ruff/blob/59c8fda3f8f3bf2cc5c1ae34e7ca9dbea4d0278f/crates/ruff_linter/src/fix/edits.rs#L207) here? This really looks like all we need to do is remove the `file` argument and replace `open` with `pathlib.Path(file).open`. Then we could leave more of the formatting untouched again.

It's not a huge deal, but it might lead to simpler code too if `remove_argument` can work.

chirizxc: it seems with `remove_argument` there will be more calls since we will be removing 1, 6, 7 arguments, would it be easier to discard them just by index, don't really know the best way to do it

ntBre: Oh right, I forgot we had to discard other default arguments. This seems fine again then!

chirizxc:
https://github.com/astral-sh/ruff/pull/20169/files#diff-c2de436c83da6c97764f1d9381108ba93899ba45b32492dc04cfb7d1a84db760R569

i don't think that's the way to fix it either

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -108,17 +108,24 @@ pub(crate) fn builtin_open(checker: &Checker, call: &ExprCall, segments: &[&str]
...
         return;
-    };
+    }
 
-    if !is_fix_builtin_open_enabled(checker.settings()) {
+    if checker
+        .semantic()
+        .current_expression_grandparent()
+        .is_some_and(|expr| matches!(expr, Expr::Call(_) | Expr::Attribute(_)))
+    {
```
ntBre: What is this check for? I thought the fix it removed looked okay, but I could be missing something.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -145,37 +152,25 @@ pub(crate) fn builtin_open(checker: &Checker, call: &ExprCall, segments: &[&str]
...
-                },
             }
+            ArgOrKeyword::Keyword(kw) => match kw.arg.as_deref() {
+                Some(arg)
+                    if matches!(
+                        arg,
+                        "mode" | "buffering" | "encoding" | "errors" | "newline"
+                    ) =>
+                {
+                    Some(format!("{arg}={}", locator.slice(&kw.value)))
```
ntBre: Could we just slice out the whole `kw` instead of `format!` here? Then we could remove the other `to_string` call too. I think `locator.slice(kw)` should work and preserve both the `arg` name and `value`.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
chirizxc: https://github.com/astral-sh/ruff/pull/20169#issuecomment-3282552111

chirizxc: there after `open()` comes `.close()` it seems if after `open()` there is some call it is better not to suggest fix

ntBre: I saw the other comment, but I don't think the GitHub link is working properly. It keeps showing me this snippet:

image

I think it's fine to call `open().close()` as long as both functions return the same type, which I think is true? I tried checking the types with [ty](https://play.ty.dev/0b9e6089-f92a-4775-8cc4-6697d8f80298), but the number of overloads makes it a bit difficult to parse. I'd just assume it's okay based on the `pathlib` [docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open):

> Open the file pointed to by the path, like the built-in `open()` function does

I don't think checking the grandparent expression type is that robust anyway, there are plenty of other erroneous ways to use the return type if it's not the same. I think all of the PTH fixes rely on this assumption.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let locator = checker.locator();
+        let file_code = locator.slice(file.range());
+
+        let args = |i: usize, arg: ArgOrKeyword| match arg {
+            ArgOrKeyword::Arg(expr) => {
+                if expr.range() == file.range() || i == 6 || i == 7 {
+                    None
+                } else {
+                    Some(locator.slice(expr.range()).to_string())
```
ntBre:
```suggestion
                    Some(locator.slice(expr.range()))
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/builtin_open.rs
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+            ArgOrKeyword::Keyword(kw) => match kw.arg.as_deref() {
+                Some("mode" | "buffering" | "encoding" | "errors" | "newline") => {
+                    Some(locator.slice(kw).to_string())
```
ntBre:
```suggestion
                    Some(locator.slice(kw))
```

ntBre (approved):
Thank you!

chirizxc:
I guess I forgot about that, thanks for the change

PR #20172, opened 2025-08-31
[`airflow`] Move `airflow.operators.postgres_operator.Mapping` from `AIR302` to `AIR301`

Lee-W:
### Why
Removal should be grouped into the same category. It doesn't matter whether it's from a provider or not (and the only case we used to have was not anyway).
`ProviderReplacement` is used to indicate that we have a replacement and we might need to install an extra Python package to cater to it.

### What
Move `airflow.operators.postgres_operator.Mapping` from AIR302 to AIR301 and get rid of `ProviderReplace::None`

## Test Plan

Update the test fixtures accordingly in the first commit and reorganize them in the second commit


ntBre (approved):
Nice!

PR #20173, opened 2025-08-31
docs(AIR002): improve AIR002 error message

Lee-W:

### DIFF
--- a/crates/ruff_linter/src/rules/airflow/rules/dag_schedule_argument.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/dag_schedule_argument.rs
@@ -46,7 +46,7 @@ pub(crate) struct AirflowDagNoScheduleArgument;
 impl Violation for AirflowDagNoScheduleArgument {
     #[derive_message_formats]
     fn message(&self) -> String {
-        "DAG should have an explicit `schedule` argument".to_string()
+        "`DAG` or `@dag` should have an explicit `schedule` argument".to_string()
     }
 }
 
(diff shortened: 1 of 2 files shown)


ntBre (approved):
Thank you!

A bit unrelated to this PR, but we're getting ready for the 0.13 release early next week. We're planning to stabilize AIR002, AIR301, AIR302, AIR311, and AIR312. Does that sound good to you?

The only reason to hold off is if you plan to make substantial changes to any of these rules, in which case you might prefer to keep the whole rule in preview.

I was going to ping you on the stabilization PR itself but wanted to give you a heads-up!

I'll review https://github.com/astral-sh/ruff/pull/20202 and https://github.com/astral-sh/ruff/pull/20172, so they can be part of the stable release too.

Lee-W:
> Thank you!
> 
> A bit unrelated to this PR, but we're getting ready for the 0.13 release early next week. We're planning to stabilize AIR002, AIR301, AIR302, AIR311, and AIR312. Does that sound good to you?
> 
> The only reason to hold off is if you plan to make substantial changes to any of these rules, in which case you might prefer to keep the whole rule in preview.
> 
> I was going to ping you on the stabilization PR itself but wanted to give you a heads-up!
> 
> I'll review #20202 and #20172, so they can be part of the stable release too.

We might extend some more rules, but I don't expect there will be a removal for the time being. would it be ok in this case?

ntBre:
By "extend" do you mean changes like the two PRs I linked? If so, I think that would be okay. Larger changes _can_ be okay, but we'll likely have to preview-gate them in line with our [versioning guidelines](https://docs.astral.sh/ruff/versioning/#version-changes).

Lee-W:
Yes, something like these 2 PRs. 

The following 2 issues are the ones we might extend AIR301 but not yet have the bandwidth to actually do it (the second one is easier, I might be able to create a PR tomorrow)

https://github.com/apache/airflow/issues/41641
https://github.com/apache/airflow/issues/55208

ntBre:
I think that should be okay then, thank you!

Lee-W:
Thanks!

PR #20177, opened 2025-08-31
[`flake8-bandit`] Fix truthiness: dict-only `**` displays not truthy for `shell` (`S602`, `S604`, `S609`)

### LINKED ISSUE #19927: S602, S604, and S609 treat `{**{}}` as truthy
### Summary

As the value of a `shell` argument, [`subprocess-popen-with-shell-equals-true` (S602)](https://docs.astral.sh/ruff/rules/subprocess-popen-with-shell-equals-true/), [`call-with-shell-equals-true` (S604)](https://docs.astral.sh/ruff/rules/call-with-shell-equals-true/), and [`unix-command-wildcard-injection` (S609)](https://docs.astral.sh/ruff/rules/unix-command-wildcard-injection/) treat a dictionary display as truthy when its elements are all double-starred non-name expressions, but a non-name expression can evaluate to an empty mapping, so the `shell` value might actually be falsey. These rules should check the truthiness of the double-starred expressions, or they should treat them as having unknown truthiness. [Example](https://play.ruff.rs/d23da6b9-9d36-4c44-a8e9-4018621ee7f4):
```
$ cat >s60.py <<'# EOF'
import subprocess
subprocess.Popen(["chmod", "777", "*.py"], shell={**{}})
dict(shell={**{}})
# EOF

$ ruff --isolated check s60.py --select S602,S604,S609 --output-format concise -q
s60.py:2:1: S602 `subprocess` call with truthy `shell` identified, security issue
s60.py:2:18: S609 Possible wildcard injection in call due to `*` usage
s60.py:3:1: S604 Function call with truthy `shell` parameter identified, security issue
```

### Version

danparizher:
## Summary
Fixes #19927


dscorbett:
With this change, `{**{}}` is considered falsey; what about `{**{**{}}}`?

danparizher:
> With this change, `{**{}}` is considered falsey; what about `{**{**{}}}`?

Good call, I had nested dict displays as truthy. I just pushed a change to have them checked recursively.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1294,17 +1353,47 @@ impl Truthiness {
                     return Self::Falsey;
                 }
 
-                if dict.items.iter().all(|item| {
-                    matches!(
-                        item,
-                        DictItem {
-                            key: None,
-                            value: Expr::Name(..)
+                // If the dict consists only of double-starred items (e.g., {**x, **y}),
```
ntBre, line 1297: Thanks for diving deep here, but I think I'd kind of lean towards just considering any starred mapping to be of unknown truthiness (so just removing the `Expr::Name` requirement from `matches!`). I don't think nested, unpacked, empty dict literals are very common, so it seems like a nicer trade-off between implementation complexity and accuracy. That's also what we do for lists, sets, and tuples just above this. `Expr::is_starred_expr` for those elements is analogous to `matches!(item, DictItem { key: None, ... })` for dicts.

crates/ruff_linter/resources/test/fixtures/flake8_bandit/S602.py
```diff
@@ -18,3 +18,7 @@
 Popen(var_string, shell=True)
 Popen([var_string], shell=True)
 Popen([var_string, ""], shell=True)
+
+# Check dict display with only double-starred expressions can be falsey.
+Popen("true", shell={**{}})
+Popen("true", shell={**{**{}}})
```
ntBre, line 24: It might also be nice to add a test case for a more common version of this problem like `{**self.shell_defaults, **self.fetch_shell_config(self.username)}` mentioned [here](https://github.com/astral-sh/ruff/issues/19927#issuecomment-3194809639). But the code fix should be the same anyway.

danparizher:
The CI failures seem to be unrelated to my changes, wondering if the one for `cargo fuzz build` was introduced elsewhere

ntBre:
Yeah no worries about the fuzz build. There was a breaking libcst release that we need to handle.

PR #20178, opened 2025-08-31
[`flake8-bultins`] Detect class-scope builtin shadowing in decorators, default args, and attribute initializers (`A003`)

### LINKED ISSUE #20171: A003 misses some attributes referenced within the class scope
### Summary

[`builtin-attribute-shadowing` (A003)](https://docs.astral.sh/ruff/rules/builtin-attribute-shadowing/) is only emitted for attributes that “are referenced from within the class scope”, but that seems to only include references in annotations. There are false negatives for attributes used within the class scope in decorators, in argument defaults, and in other attributes’ definitions, as in `property`, `id`, and `bin` respectively in [the example below](https://play.ruff.rs/d9ec5da2-a26d-4ac1-a40d-e20d3c8f446e).
```console
$ cat >a003.py <<'# EOF'
class C:
    @staticmethod
    def property(f): return f
    id = 1
    @[property][0]
    def f(self, x=[id]): return x
    bin = 2
    foo = [bin]
print(C().f(), C().foo)
# EOF

$ python a003.py
[1] [2]

$ ruff --isolated check a003.py --select A003
All checks passed!

$ flake8 --select A a003.py
a003.py:3:5: A003 class attribute "property" is shadowing a Python builtin
a003.py:4:5: A003 class attribute "id" is shadowing a Python builtin
a003.py:7:5: A003 class attribute "bin" is shadowing a Python builtin

```

### Version

ruff 0.12.11 (c2bc15bc1 2025-08-28)

danparizher:
## Summary
Fix #20171

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_builtins/mod.rs
+++ b/crates/ruff_linter/src/rules/flake8_builtins/mod.rs
@@ -14,6 +14,7 @@ mod tests {
     use crate::registry::Rule;
     use crate::rules::flake8_builtins;
     use crate::settings::LinterSettings;
+    use crate::settings::types::PreviewMode;
     use crate::test::{test_path, test_resource_path};
     use ruff_python_ast::PythonVersion;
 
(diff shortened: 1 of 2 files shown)


danparizher:
We can have this PR expanded to cover #20179 too if it makes sense, since they are in the same scope. Let me know!

crates/ruff_linter/src/rules/flake8_builtins/rules/builtin_attribute_shadowing.rs
```diff
@@ -123,16 +124,27 @@ pub(crate) fn builtin_attribute_shadowing(
             //     def repeat(value: int, times: int) -> list[int]:
             //         return [value] * times
             // ```
+            // In stable, only consider references from within the class scope.
+            // In preview, also consider references whose first non-type parent scope is the class
+            // scope (e.g., decorators, default args, and attribute initializers).
+            let consider_reference = |reference_scope_id: ScopeId| {
+                if reference_scope_id == scope_id {
+                    return true;
+                }
```
ntBre: I think this is the part that needs to be preview-gated right? The check currently inside of the `preview` check is the code we had before.

There shouldn't be any stable changes in the ecosystem report. We may also need to add a preview test runner.

PR #20190, opened 2025-09-01
[ty] Support `__init_subclass__`

sharkdp:
## Summary

`__init_subclass__` is implicitly a classmethod.

closes https://github.com/astral-sh/ty/issues/1106

## Test Plan

Regression test

### DIFF
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -721,6 +721,13 @@ impl<'db> FunctionType<'db> {
         self.literal(db).has_known_decorator(db, decorator)
     }
 
+    /// Returns true if this method is decorated with `@classmethod`, or if it is implicitly a
+    /// classmethod.
+    pub(crate) fn is_classmethod(self, db: &'db dyn Db) -> bool {
+        self.has_known_decorator(db, FunctionDecorators::CLASSMETHOD)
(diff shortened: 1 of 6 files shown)


crates/ty_python_semantic/src/types/function.rs
```diff
@@ -721,6 +721,13 @@ impl<'db> FunctionType<'db> {
         self.literal(db).has_known_decorator(db, decorator)
     }
 
+    /// Returns true if this method is decorated with `@classmethod`, or if it is implicitly a
+    /// classmethod.
+    pub(crate) fn is_classmethod(self, db: &'db dyn Db) -> bool {
+        self.has_known_decorator(db, FunctionDecorators::CLASSMETHOD)
+            || self.name(db) == "__init_subclass__"
+    }
```
dhruvmanila, line 729: Should `__new__` be included as well given that it's an implicit class method?

sharkdp: `__new__` takes `cls` as its first argument, but it's a static method, not a classmethod (https://docs.python.org/3/reference/datamodel.html#object.__new__). `cls` needs to be passed explicitly.

sharkdp: But `__class_getitem__` could probably go here? Will check separately.

PR #20192, opened 2025-09-01
[ty] `__class_getitem__` is a classmethod

sharkdp:
## Summary

`__class_getitem__` is [implicitly a classmethod](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__).

## Test Plan

Added regression test.


AlexWaygood:
Possibly `__mro_entries__` too? I might be misremembering though

dhruvmanila (approved):

sharkdp:
> Possibly `__mro_entries__` too? I might be misremembering though

Looks like that is a normal (instance) method.

Issue #19837, opened 2025-08-08
FLY002 fix causes syntax error with double quotes in the strings


MeGaGiGaGon:
### Summary

When the joined items contain both a variable and a string with a double quote, the fix for [static-join-to-f-string (FLY002)](https://docs.astral.sh/ruff/rules/static-join-to-f-string/#static-join-to-f-string-fly002) will cause a syntax error [playground](https://play.ruff.rs/425c7b15-4eaf-47b8-9f5d-2bf428c665b6):
```powershell
PS ~>echo @'
"".join((foo, '"'))
'@ | uvx ruff check --select FLY002 - --fix --unsafe-fixes 2>&1 | rg "syntax error"
error: Fix introduced a syntax error. Reverting all changes.
```

### Version

ruff 0.12.8 (f51a228f0 2025-08-07)

PR #20197, opened 2025-09-01
[`flynt`] Use triple quotes for joined raw strings with newlines (`FLY002`)

### LINKED ISSUE #19887: FLY002 autofix with raw strings causes a syntax error
### Summary

I get an error with the autofix for FLY002 when one (or more) of the strings in the join are a raw string:

```bash
$ cat ruff_bug.py
'\n'.join([r'line1','line2'])

$ uvx ruff --version
ruff 0.12.8
$ uvx ruff check ruff_bug.py --no-cache --isolated --fix --unsafe-fixes --select FLY002
Installed 1 package in 1ms

    https://github.com/astral-sh/ruff/issues/new?title=%5BFix%20error%5D

...quoting the contents of `ruff_bug.py`, the rule codes FLY002, along with the `pyproject.toml` settings and executed command, we'd be very appreciative!

ruff_bug.py:1:1: FLY002 Consider f-string instead of string join
  |
1 | '\n'.join([r'line1','line2'])
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ FLY002
  |
  = help: Replace with f-string

Found 1 error.
[*] 1 fixable with the --fix option.
```

### Version

ruff 0.12.8

TaKO8Ki:
## Summary

Fixes #19887

- flynt(FLY002): When joining only string constants, upgrade raw single-quoted strings to raw triple-quoted if the resulting
content contains a newline.
- Choose a safe triple-quote delimiter by switching to the opposite quote style if the preferred triple appears inside the
content.
- Update FLY002 snapshot to include the `\n'.join([r'line1','line2'])` case.

## Test Plan

I've added one test case to FLY002.py.

### DIFF
--- a/crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
+++ b/crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
@@ -91,7 +91,7 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
         let mut flags = flags?;
 
         // If the result is a raw string and contains a newline, use triple quotes.
-        if flags.prefix().is_raw() && (content.contains('\n') || content.contains('\r')) {
+        if flags.prefix().is_raw() && content.contains(['\n', '\r']) {
             flags = flags.with_triple_quotes(ruff_python_ast::str::TripleQuotes::Yes);
 
             // Prefer a delimiter that doesn't occur in the content; if both occur, bail.


crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -72,24 +72,47 @@ fn is_static_length(elts: &[Expr]) -> bool {
...
+            .join(joiner);
+
+        let mut flags = flags?;
+
+        // If the result is a raw string and contains a newline, use triple quotes.
+        if flags.prefix().is_raw() && (content.contains('\n') || content.contains('\r')) {
+            flags = flags.with_triple_quotes(ruff_python_ast::str::TripleQuotes::Yes);
+
+            // Ensure we pick a safe triple-quote delimiter that doesn't occur in the content.
+            let quote = flags.quote_style();
```
ntBre: I think we could use a combination of [`StringFlags::quote_str`](https://github.com/astral-sh/ruff/blob/59c8fda3f8f3bf2cc5c1ae34e7ca9dbea4d0278f/crates/ruff_python_ast/src/nodes.rs#L746) and [`Quote::opposite`](https://github.com/astral-sh/ruff/blob/59c8fda3f8f3bf2cc5c1ae34e7ca9dbea4d0278f/crates/ruff_python_ast/src/str.rs#L37) to simplify this section a bit.

Should we add a test case (or a few test cases) for this behavior?

I also think we may need to check for both `'''` and `"""` in every string. We could end up with a pathological case like this:

```py
'\n'.join([r"raw string", '<""">', "<'''>"])
```

which we can't join into a single string with either delimiter. I think if we have a raw string, and the joined string contains our attempted triple quotes, we'll just have to bail out without a fix.

CPython escapes some of the quotes in this case:

```pycon
>>> '\n'.join([r"raw string", '<""">', "<'''>"])
'raw string\n<""">\n<\'\'\'>'
```

I think we could also do that with our [`UnicodeEscape`](https://github.com/astral-sh/ruff/blob/d412f49fb2ec5af376cba15fd5c3348a9d0c20e0/crates/ruff_python_literal/src/escape.rs#L50) type, but I'd be fine with, or even prefer, avoiding the fix in that case.

TaKO8Ki: Thank you for the suggestion. I have addressed your comment.

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -72,24 +72,42 @@ fn is_static_length(elts: &[Expr]) -> bool {
...
+                } else {
+                    None
+                }
+            })
+            .join(joiner);
+
+        let mut flags = flags?;
+
+        // If the result is a raw string and contains a newline, use triple quotes.
+        if flags.prefix().is_raw() && (content.contains('\n') || content.contains('\r')) {
```
ntBre: nit: I think we can combine the `contains` checks:


```suggestion
        if flags.prefix().is_raw() && content.contains(['\n', '\r']) {
```

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -72,24 +72,42 @@ fn is_static_length(elts: &[Expr]) -> bool {
...
+
+            // Prefer a delimiter that doesn't occur in the content; if both occur, bail.
+            if content.contains(flags.quote_str()) {
+                flags = flags.with_quote_style(flags.quote_style().opposite());
+                if content.contains(flags.quote_str()) {
+                    // Both "'''" and "\"\"\"" are present in content; avoid emitting
+                    // an invalid raw triple-quoted literal (or escaping). Bail on the fix.
+                    return None;
+                }
+            }
```
ntBre, line 105: I think something like this would help with https://github.com/astral-sh/ruff/issues/19837 too, but it looks sufficiently different that we should leave it for a separate PR. I think we'd need this kind of check in one of the `helpers` below.

TaKO8Ki: Ok. I will try it in a different pull request:)

PR #20199, opened 2025-09-01
[ty] chore: `ContextManagerError` message update

Y3K:
## Summary

Fixes https://github.com/astral-sh/ty/issues/940

## Test Plan

```python
class Bound:
    def __enter__(self): ...

    def __exit__(self, *args): ...


class EnterUnbound:
    def __exit__(self): ...


class ExitUnbound:
    def __enter__(self): ...


class Unbound: ...


def bound(x: Bound):
    with x: ...


def enter_unbound(x: EnterUnbound):
    with x: ...


def exit_unbound(x: ExitUnbound):
    with x: ...


def unbound(x: Unbound):
    with x: ...


def union_enter_unbound(x: Bound | EnterUnbound):
    with x: ...


def union_exit_unbound(x: Bound | ExitUnbound):
    with x: ...


def union_unbound(x: Bound | Unbound):
    with x: ...


def multiple(x: Bound | EnterUnbound | ExitUnbound):
    with x: ...
```

```bash
$ cargo run --bin ty -- check --project ../pydummy

error[invalid-context-manager]: Object of type `EnterUnbound` cannot be used with `with` because it does not implement `__enter__`, and it does not correctly implement `__exit__`
  --> ../pydummy/main.py:23:10
   |
22 | def enter_unbound(x: EnterUnbound):
23 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `ExitUnbound` cannot be used with `with` because it does not implement `__exit__`
  --> ../pydummy/main.py:27:10
   |
26 | def exit_unbound(x: ExitUnbound):
27 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `Unbound` cannot be used with `with` because it does not implement `__enter__` and `__exit__`
  --> ../pydummy/main.py:31:10
   |
30 | def unbound(x: Unbound):
31 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `Bound | EnterUnbound` cannot be used with `with` because the method `__enter__` of `EnterUnbound` is possibly unbound
  --> ../pydummy/main.py:35:10
   |
34 | def union_enter_unbound(x: Bound | EnterUnbound):
35 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `Bound | ExitUnbound` cannot be used with `with` because the method `__exit__` of `ExitUnbound` is possibly unbound
  --> ../pydummy/main.py:39:10
   |
38 | def union_exit_unbound(x: Bound | ExitUnbound):
39 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `Bound | Unbound` cannot be used with `with` because the methods `__enter__` and `__exit__` of `Unbound` are possibly unbound
  --> ../pydummy/main.py:43:10
   |
42 | def union_unbound(x: Bound | Unbound):
43 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default

error[invalid-context-manager]: Object of type `Bound | EnterUnbound | ExitUnbound` cannot be used with `with` because the method `__enter__` of `EnterUnbound` is possibly unbound, and the method `__exit__` of `ExitUnbound` is possibly unbound
  --> ../pydummy/main.py:47:10
   |
46 | def multiple(x: Bound | EnterUnbound | ExitUnbound):
47 |     with x: ...
   |          ^
   |
info: rule `invalid-context-manager` is enabled by default
```

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -7899,26 +7899,23 @@ impl<'db> ContextManagerError<'db> {
         let db = context.db();
 
         // Helper function to analyze union members for specific method issues
-        let analyze_union_method_issues = |method_name: &str| -> Vec<Type<'db>> {
-            if let Type::Union(union) = context_expression_type {
-                let mut missing_method = Vec::new();
+        let analyze_union_method_issues = |union: UnionType, method_name: &str| -> Vec<Type<'db>> {
+            let mut missing_method = Vec::new();
 
-                for element in union.elements(db) {
-                    let member_lookup = element.member_lookup_with_policy(
-                        db,
-                        method_name.into(),
-                        MemberLookupPolicy::default() | MemberLookupPolicy::NO_INSTANCE_FALLBACK,
-                    );
+            for element in union.elements(db) {
+                let member_lookup = element.member_lookup_with_policy(
+                    db,
+                    method_name.into(),
+                    MemberLookupPolicy::default() | MemberLookupPolicy::NO_INSTANCE_FALLBACK,
+                );
 
-                    match member_lookup.place {
-                        Place::Unbound => missing_method.push(*element),
-                    }
+                match member_lookup.place {
+                    Place::Unbound => missing_method.push(*element),
+                    Place::Type(_, _) => (),
                 }
-
-                missing_method
-            } else {
-                Vec::new()
             }
+
+            missing_method
         };
 
         let format_call_dunder_error = |call_dunder_error: &CallDunderError<'db>, name: &str| {
@@ -7926,15 +7923,15 @@ impl<'db> ContextManagerError<'db> {
                 CallDunderError::MethodNotAvailable => format!("it does not implement `{name}`"),
                 CallDunderError::PossiblyUnbound(_) => {
                     // For unions, provide more specific information about which members lack the method
-                    if let Type::Union(_) = context_expression_type {
-                        let missing_method = analyze_union_method_issues(name);
+                    if let Type::Union(union) = context_expression_type {
+                        let missing_method = analyze_union_method_issues(union, name);
                         if !missing_method.is_empty() {
                             let missing_types: Vec<_> = missing_method
                                 .iter()
                                 .map(|t| t.display(db).to_string())
                                 .collect();
                             return format!(
-                                "the method `{name}` of `{}` {} possibly unbound",
+                                "the method `{name}` of `{}` {} unbound",
                                 missing_types.join("` and `"),
                                 if missing_types.len() == 1 {
                                     "is"
@@ -7947,6 +7944,9 @@ impl<'db> ContextManagerError<'db> {
 
                     format!("the method `{name}` is possibly unbound")
                 }
+                // TODO: Use more specific error messages for the different error cases.
+                //  E.g. hint toward the union variant that doesn't correctly implement enter,
+                //  distinguish between a not callable `__enter__` attribute and a wrong signature.
                 CallDunderError::CallError(_, _) => {
                     format!("it does not correctly implement `{name}`")
                 }
@@ -7960,9 +7960,9 @@ impl<'db> ContextManagerError<'db> {
             match (error_a, error_b) {
                 (CallDunderError::PossiblyUnbound(_), CallDunderError::PossiblyUnbound(_)) => {
                     // For unions, provide more specific information about which members lack the methods
-                    if let Type::Union(_) = context_expression_type {
-                        let missing_enter = analyze_union_method_issues(name_a);
-                        let missing_exit = analyze_union_method_issues(name_b);
+                    if let Type::Union(union) = context_expression_type {
+                        let missing_enter = analyze_union_method_issues(union, name_a);
+                        let missing_exit = analyze_union_method_issues(union, name_b);
 
                         // Find union members that are missing both methods
                         let missing_both: Vec<_> = missing_enter
@@ -7973,7 +7973,7 @@ impl<'db> ContextManagerError<'db> {
 
                         if !missing_both.is_empty() {
                             return format!(
-                                "the methods `{name_a}` and `{name_b}` of `{}` are possibly unbound",
+                                "the methods `{name_a}` and `{name_b}` of `{}` are unbound",
                                 missing_both.join("` and `")
                             );
                         }
@@ -7988,9 +7988,9 @@ impl<'db> ContextManagerError<'db> {
                 }
                 (_, _) => {
                     // For mixed error cases with unions, provide detailed information about which members have issues
-                    if let Type::Union(_) = context_expression_type {
-                        let missing_enter = analyze_union_method_issues(name_a);
-                        let missing_exit = analyze_union_method_issues(name_b);
+                    if let Type::Union(union) = context_expression_type {
+                        let missing_enter = analyze_union_method_issues(union, name_a);
+                        let missing_exit = analyze_union_method_issues(union, name_b);
 
                         let mut parts = Vec::new();
 
@@ -8000,7 +8000,7 @@ impl<'db> ContextManagerError<'db> {
                                 .map(|t| t.display(db).to_string())
                                 .collect();
                             parts.push(format!(
-                                "the method `{name_a}` of `{}` {} possibly unbound",
+                                "the method `{name_a}` of `{}` {} unbound",
                                 missing_types.join("` and `"),
                                 if missing_types.len() == 1 {
                                     "is"
@@ -8016,7 +8016,7 @@ impl<'db> ContextManagerError<'db> {
                                 .map(|t| t.display(db).to_string())
                                 .collect();
                             parts.push(format!(
-                                "the method `{name_b}` of `{}` {} possibly unbound",
+                                "the method `{name_b}` of `{}` {} unbound",
                                 missing_types.join("` and `"),
                                 if missing_types.len() == 1 {
                                     "is"


crates/ty_python_semantic/src/types.rs
```diff
@@ -7896,15 +7896,59 @@ impl<'db> ContextManagerError<'db> {
...
+                        MemberLookupPolicy::default() | MemberLookupPolicy::NO_INSTANCE_FALLBACK,
+                    );
+
+                    match member_lookup.place {
+                        Place::Unbound => missing_method.push(*element),
+                        Place::Type(_, _) => has_method.push(*element),
+                    }
+                }
+
+                (missing_method, has_method)
```
carljm: It seems we never use the second element of this return type, and I'm not sure why we ever would use it. Maybe just remove it?

crates/ty_python_semantic/src/types.rs
```diff
@@ -7896,15 +7896,59 @@ impl<'db> ContextManagerError<'db> {
             EvaluationMode::Sync => ("__enter__", "__exit__"),
         };
 
+        let db = context.db();
+
+        // Helper function to analyze union members for specific method issues
+        let analyze_union_method_issues = |method_name: &str| -> (Vec<Type<'db>>, Vec<Type<'db>>) {
+            if let Type::Union(union) = context_expression_type {
```
carljm: Maybe have this method take a `UnionType` (the struct wrapped by `Type::Union`) instead of a `Type`, since it's only useful for union types, and only called for union types? Then we wouldn't need the "empty return" case below. Every call site below is already guarded by `if let Type::Union(_) = context_expression_type {` which could just as easily be `if let Type::Union(union) = context_expression_type {`.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7896,15 +7896,59 @@ impl<'db> ContextManagerError<'db> {
...
+                for element in union.elements(db) {
+                    let member_lookup = element.member_lookup_with_policy(
+                        db,
+                        method_name.into(),
+                        MemberLookupPolicy::default() | MemberLookupPolicy::NO_INSTANCE_FALLBACK,
+                    );
+
+                    match member_lookup.place {
+                        Place::Unbound => missing_method.push(*element),
+                        Place::Type(_, _) => has_method.push(*element),
```
carljm: We could distinguish bound vs possibly-unbound here, because we shouldn't give the wishy-washy "possibly unbound" below if we know it's definitely not bound on a type.

We should only report "possibly unbound" on a type like this:

```py
class PossiblyUnbound:
    if random.randint(0, 1):
        def __enter__(self): pass
```

But this kind of type should be uncommon, so I also think it's fine to only give the extra details for union elements where it's definitely unbound.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7896,15 +7896,59 @@ impl<'db> ContextManagerError<'db> {
...
+                    // For unions, provide more specific information about which members lack the method
+                    if let Type::Union(_) = context_expression_type {
+                        let (missing_method, _) = analyze_union_method_issues(name);
+                        if !missing_method.is_empty() {
+                            let missing_types: Vec<_> = missing_method
+                                .iter()
+                                .map(|t| t.display(db).to_string())
+                                .collect();
+                            return format!(
+                                "the method `{name}` of `{}` {} possibly unbound",
```
carljm: At this point we've verified that for these specific elements of the union, the method is definitely not bound.

```suggestion
                                "the method `{name}` of `{}` {} unbound",
```

(If we did also record the union elements where the method is possibly-unbound, then we'd need to report both cases here, but we should still be clear about which union element is which.)

crates/ty_python_semantic/src/types.rs
```diff
@@ -7917,6 +7961,25 @@ impl<'db> ContextManagerError<'db> {
...
+                        // Find union members that are missing both methods
+                        let missing_both: Vec<_> = missing_enter
+                            .iter()
+                            .filter(|t| missing_exit.contains(t))
+                            .map(|t| t.display(db).to_string())
+                            .collect();
+
+                        if !missing_both.is_empty() {
+                            return format!(
+                                "the methods `{name_a}` and `{name_b}` of `{}` are possibly unbound",
```
```suggestion
                                "the methods `{name_a}` and `{name_b}` of `{}` are unbound",
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7917,6 +7961,25 @@ impl<'db> ContextManagerError<'db> {
                                          name_b: &str| {
             match (error_a, error_b) {
                 (CallDunderError::PossiblyUnbound(_), CallDunderError::PossiblyUnbound(_)) => {
+                    // For unions, provide more specific information about which members lack the methods
+                    if let Type::Union(_) = context_expression_type {
+                        let (missing_enter, _) = analyze_union_method_issues(name_a);
+                        let (missing_exit, _) = analyze_union_method_issues(name_b);
+
+                        // Find union members that are missing both methods
```
carljm, line 7967: I think in this case we'd want to report on any union element that is missing _either_ method. We could potentially reach this case when there is no union element missing both methods.

One thing that could simplify the code here (and in other cases in this PR) is that we don't have to squeeze all the information into a single sentence for a single diagnostic message. It would require a bit more restructuring in this `report_diagnostic` method, but we can report multiple additional info sub-diagnostics. So e.g. each union element that is missing a method could just be its own `info` sub-diagnostic, and then we wouldn't have to worry so much about joining with " and ", is vs are, etc.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7925,16 +7988,60 @@ impl<'db> ContextManagerError<'db> {
...
+
+                        let mut parts = Vec::new();
+
+                        if !missing_enter.is_empty() {
+                            let missing_types: Vec<_> = missing_enter
+                                .iter()
+                                .map(|t| t.display(db).to_string())
+                                .collect();
+                            parts.push(format!(
+                                "the method `{name_a}` of `{}` {} possibly unbound",
```
```suggestion
                                "the method `{name_a}` of `{}` {} unbound",
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7925,16 +7988,60 @@ impl<'db> ContextManagerError<'db> {
...
+                            ));
+                        }
+
+                        if !missing_exit.is_empty() {
+                            let missing_types: Vec<_> = missing_exit
+                                .iter()
+                                .map(|t| t.display(db).to_string())
+                                .collect();
+                            parts.push(format!(
+                                "the method `{name_b}` of `{}` {} possibly unbound",
```
```suggestion
                                "the method `{name_b}` of `{}` {} unbound",
```

crates/ty_python_semantic/resources/mdtest/with/async.md
```diff
@@ -84,6 +84,117 @@ async def main():
         ...
 ```
 
+## Union context managers with specific member issues
+
```
carljm, line 88: We can snapshot diagnostics for these tests, which is an easier way to see the full resulting diagnostic, instead of putting full error messages directly in the mdtest. This is usually preferred for tests focused on diagnostic rendering, and is especially useful when the diagnostic is not all crammed into a single sentence, as I suggest below.
```suggestion

<!-- snapshot-diagnostics -->

```

Then you use `cargo insta test -p ty_python_semantic` to generate new snapshots to review with `cargo insta review` and then commit.

crates/ty_python_semantic/resources/mdtest/with/async.md
```diff
@@ -84,6 +84,117 @@ async def main():
...
+            return "foo"
+
+        async def __aexit__(self, exc_type, exc_value, traceback): ...
+
+    class EnterUnbound:
+        async def __aexit__(self): ...
+
+    context_expr = Bound() if flag else EnterUnbound()
+
+    # error: [invalid-context-manager] "Object of type `Bound | EnterUnbound` cannot be used with `async with` because the method `__aenter__` of `EnterUnbound` is possibly unbound"
```
carljm, line 104: E.g. if we snapshot diagnostics, then we can just do this here:
```suggestion
    # error: [invalid-context-manager]
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7896,15 +7896,59 @@ impl<'db> ContextManagerError<'db> {
...
+                                }
+                            );
+                        }
+                    }
+
                     format!("the method `{name}` is possibly unbound")
                 }
-                // TODO: Use more specific error messages for the different error cases.
-                //  E.g. hint toward the union variant that doesn't correctly implement enter,
-                //  distinguish between a not callable `__enter__` attribute and a wrong signature.
```
carljm: I don't think this PR really addresses most of this TODO. We still don't surface the full underlying call error for wrongly-implemented dunders, like we do for e.g. `IterationError::report_diagnostic`. We don't need to do all of that in this PR, but I think we should leave this TODO in place.

carljm (changes requested):
Thanks for the pull request! This is definitely an improvement.

My idea when I filed the issue was that we would bubble up the information about which types had the dunder unbound or possibly-unbound as part of the `CallDunderError`, but looking at it now I see this would probably require special-casing `Type::Union` in `try_call_dunder` (because `member_lookup_with_policy` squashes union information into a single `PlaceAndQualifiers`). I still think this might make sense, but I also think it's OK to just re-create the information once we know there's a problem.

Y3K:
Thank you for your review @carljm!

I'm quite rookie with Rust and `ty` therefore the changes requested still are a bit confussing to me.

I'll read each suggestion carefully and work on a new approach, will ping you once it's ready.

Thanks!

carljm:
@Y3K happy to offer clarification on any comments. I think for purposes of this PR you can ignore the comment about "bubbling up the information through `CallDunderError`", and you can ignore one other comment I'll mention inline. The other ones I think don't require an entirely "new approach" to address, just some updates to the current approach.

crates/ty_python_semantic/src/types.rs
carljm: I think for this PR you can ignore this comment; it could be a later improvement if it turns out to be useful.

Y3K:
Thank you again, @carljm !

I don't want to waste a lot of your time, but I do accept any extra help you want and can provide.

Will address all your feedback within this PR 🙌🏼

crates/ty_python_semantic/src/types.rs
Y3K: Updated the function and it's usage (not pushed yet).

crates/ty_python_semantic/src/types.rs
Y3K: It is so clear now that you mention it! Sorry for such a dumb mistake.

I've updated the code (not pushed yet).

crates/ty_python_semantic/src/types.rs
Y3K: Marking as `Resolved` for now 🙆🏼‍♂️

crates/ty_python_semantic/src/types.rs
Y3K: Updated 👌🏼 (not pushed yet)

crates/ty_python_semantic/src/types.rs
Y3K: ♻️ ditto

crates/ty_python_semantic/src/types.rs
Y3K: ♻️ ditto

crates/ty_python_semantic/src/types.rs
Y3K: ♻️ ditto

crates/ty_python_semantic/src/types.rs
Y3K: Added it back (not pushed)

crates/ty_python_semantic/resources/mdtest/with/async.md
Y3K: Sorry for the rookie question.

Can I get a really brief example of how snapshots work?

e.g. should be snapshotted in main or post changes? What if the message changes do I need to update snaps? Are all the snaps run when I do `cargo nextest run`?

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

I'll add a comment to the linked ticket that this is useful as a resource if someone picks that task up again.

PR #13965, opened 2024-10-28
[pyflakes] Fix check of unused imports.

gpilikin:
## Summary
Fixed incorrect detection of unused imports with submodules. They were always marked as used even though they are not.

For example:
``` 
import multiprocessing.pool
import multiprocessing.process  # marked as used

tp = multiprocessing.pool.ThreadPool()
...
```

The solution:
- Added checking all possible attribute combinations for similar imports and marking them as used ("foo.bar", "foo" etc),  if there are none, the root module itself should be marked as used.
- Added checks on the use of all imports with submodules.

## Test Plan

cargo test

Closes: https://github.com/astral-sh/ruff/issues/60

PR #5011, opened 2023-06-11
Consider submodule imports to detect unused imports

dhruvmanila:
## Summary

Take the below code as an example:
```python
import foo.bar
import foo.some
```

Here, a single binding is created (`foo`) for both imports and the binding created for `foo.bar` will shadow the one created for `foo.some`. Now, if the binding `foo` is unused, then it means all of the bindings shadowing this is unused as well.

There's a conflict which I've commented down below. This is resulting in 2 test failures.

## Test Plan

`cargo test`

fixes: #60
fixes: #4655

PR #5010, opened 2023-06-10
Consider submodule imports to detect unused imports

dhruvmanila:
## Summary

TODO

## Test Plan

TODO

fixes: #60
fixes: #4655

PR #20200, opened 2025-09-01
[`pyflakes`] Handle some common submodule import situations for `unused-import` (`F401`)

dylwil3:
# Summary

The PR under review attempts to make progress towards the age-old problem of submodule imports, specifically with regards to their treatment by the rule [`unused-import` (`F401`)](https://docs.astral.sh/ruff/rules/unused-import/).

Some related issues:
- https://github.com/astral-sh/ruff/issues/60
- https://github.com/astral-sh/ruff/issues/4656

Prior art:
- https://github.com/astral-sh/ruff/pull/13965
- https://github.com/astral-sh/ruff/pull/5010
- https://github.com/astral-sh/ruff/pull/5011
- https://github.com/astral-sh/ruff/pull/666

One distinguishing feature of this PR is that we have attempted to be less ambitious in the hopes that this will improve the chances of landing nonzero progress into `main`.

## What the People Want

Users expect to see the following behavior in these common situations:

```python
# Example 1
import a    # ok
import a.b  # F401: unused-import

a.foo()
```

```python
# Example 2
import a    # F401: unused-import
import a.b  # ok

a.b.bar()
```

```python
# Example 3
import a    # ok
import a.b  # ok

a.foo()
a.b.bar()
```

The following situations are unintuitive to users, but are valid Python and we are probably forced to proceed as indicated:

```python
# Example 4
import a.b  # ok

a.foo()
```

```python
# Example 5
import a.b  # F401: unused-import
import a.c  # ok

a.x.baz()
```

The goal of this PR is to modify the implementation of `unused-import` to match this expectation.

## Modeling the Python in Users' Brains

Given the expected behavior above, how might a typical user be modeling Python in their minds? One possibility is as follows. Upon seeing the statement:

```python
import a.b
```

the user thinks:

> \[User\]: We have made available the symbol `a`, which is a module, but we are only allowed to access members of this module that have the form `a.b.*`.

This happens to be incorrect. Rather than _restrict_ what is available to access on `a`, the opposite happens:

> \[Python\]: We have executed the equivalent of the statement `import a`, and, in particular, can access any top-level member of `a` made available by this import. We _then_ import the submodule `b`, making possibly even more members available.

Similarly, when this user sees:

```python
import a
import a.b

a.foo()
a.b.bar()
```

they think:

> [User] The line `a.foo()` is only allowed because of the first import. So that is definitely needed. The second import may or may not be needed depending on whether `a/__init__.py` imports `b`.

when really the opposite is true:

> [Python] The first import is totally redundant and is not used. The call `a.foo()` references the symbol `a` that is being loaded in on the _second_ line.

These views are actually incompatible, so it is not possible to model the user's mind without running afoul of Examples 4 and 5 above.

## The Worst of Both Worlds

To model both the behavior that the user expects _and_ the unintuitive behavior that Python allows, we adopt an approach I'll call "Thanks I Hate It (TIHI)". It is the worst of both worlds, and it alters how we think about accessing members of a module. Again we will consider the example:

```python
import a
import a.b

a.foo()
a.b.bar()
```

> \[TIHI\]: Following Python, the import statements both make available the symbol `a` and all of it's top-level members. The second import statement also makes available the members of the form `a.b.*`. However, the attribute call `a.foo` will access `a` _via the first import statement_ and the call `a.b.bar` will access `a.b` via the second.

In general the TIHI model will resolve an attribute load of the form `a.a1...an` (for a module `a`) by:

1. Finding the maximum prefix match amongst all available
   import statements `import a.b1...bm`,
1. Among these, finding those of minimal path-length, and
1. Adding a reference to the binding created by the nearest
   of these.

## What we actually do

Rather than force Ruff's semantic model to be incorrect by actually implementing TIHI, we hack TIHI into the logic of `unused-import` itself. Doing this is somewhat awkward due to the nature of the implementation of `unused-import`.

Let me briefly review the current implementation and then explain what we do differently.

### Current implementation
After having the semantic model visit the entire module, we look at the current _live_ import bindings that have no references. There are various other filters and logic then applied, but our essential starting bucket for this rule consists of that: bindings to import statements that are live at the end of the file and have no references.

In particular, if the module consists solely of:

```python
import a
import a.b
```

then our starting "bucket" where we might emit a lint does not include `import a` since the symbol `a` is shadowed by the statement `import a.b`.

### This PR

From now on we will refer to the current behavior as the "stable" behavior. In this PR we introduce preview behavior under very specific assumptions designed to limit the scope of this PR to handle the most common cases where users get tripped up. If any of these assumptions are not met then we revert to the stable behavior.

To begin, instead of only considering import bindings that are live at the end of the module, we also consider all bindings shadowed by these. We assume that all of these bindings are also import bindings - specifically simple imports and submodule imports without aliases, not `from` imports. If not - revert to stable behavior. Next, we have to decide which of these statements have been "used" in the sense of the "TIHI" model above. So we iterate through the references to the last binding and decide which import statement is _intuitively_ being referenced, according to the matching logic we described earlier.

As a technical note here, a call like `a.b.foo()` will create a reference to the symbol `a` - so we will not see the full attribute access. We must therefore crawl up the AST and collect the qualified name of the attribute access. We assume that all references arise from a `Name` node or from appearance in a binding to `__all__` - if we find some other kind of reference, we revert to stable behavior.

Having marked each import statement as used or unused, we collect the unused ones and then proceed as in the stable behavior.

## Questions

> What if the submodule import has a side-effect and so is used? For example, if `a/b.py` contains `import a.x` but `a/__init__.py` does not. Then `a.x.baz` really _does_ need `import a.b`.

We already don't model side-effects, even for simple module imports like `import a`. So the user would be required to suppress the lint for such an anti-pattern.

> Shouldn't you do ____ to avoid allocating and performance regressions?

Maybe! The benchmarks did not show a regression, but if we want to pre-emptively do that I'm happy to. Some places where it may make sense are:

- We currently allocate a vector of size 1 in a common situation where we "bail" to the stable behavior when iterating over bindings. This is to make the opaque type of various iterators the same. But we could avoid that.
- We could use something like `SmallVec` when collecting candidate unused import bindings shadowing a given one, since there will often not be very many
- We could use a bitmask to mark used bindings (for a given symbol) as long as we are only dealing with < 64 of them and fallback to `Vec<bool>` otherwise
- ... probably other things I'm not thinking of!

### DIFF
--- a/crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
+++ b/crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
@@ -726,11 +726,11 @@ fn unused_imports_from_binding<'a, 'b>(
     for ref_id in binding.references() {
         let resolved_reference = semantic.reference(ref_id);
         if !marked_dunder_all && resolved_reference.in_dunder_all_definition() {
-            let first = binding
-                .as_any_import()
-                .expect("binding to be import binding since current function called after restricting to these in `unused_imports_in_scope`")
-                .qualified_name()
-                .segments()[0];
+            let first = *binding
+                                .as_any_import()
+                                .expect("binding to be import binding since current function called after restricting to these in `unused_imports_in_scope`")
+                                .qualified_name()
+                                .segments().first().expect("import binding to have nonempty qualified name");
             mark_uses_of_qualified_name(&mut marked, &QualifiedName::user_defined(first));
             marked_dunder_all = true;
             continue;


crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__f401_import_submodules_in_function_scope.snap
dylwil3, line 1: Note that we have restricted our analyses to the present scope, so this snapshot is possibly surprising. The main reason to restrict to one scope at a time is that the unused import rule already acts on the scope level. So the current behavior is already to emit two diagnostics even for something like:

```
import a

def foo():
    import a
```

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+    let left = qname
+        .segments()
+        .iter()
+        .zip(prototype.segments())
+        .take_while(|(x, y)| x == y)
+        .count();
+    (left, std::cmp::Reverse(qname.segments().len()))
+}
+
+fn best_match<'a, 'b, 'c>(
```
dylwil3: I imagine someone smarter than me could avoid so many lifetimes here... Presumably this is a code smell at this point

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+                BindingKind::Import(_)
+                    | BindingKind::FromImport(_)
+                    | BindingKind::SubmoduleImport(_)
+            )
+        })
+        .filter(|(_, bdg)| !bdg.is_global() && !bdg.is_nonlocal() && !bdg.is_explicit_export())
+        .flat_map(|(id, bdg)| {
+            if is_refined_submodule_import_match_enabled(settings)
+                // Only apply the new logic when all shadowed bindings
+                // are un-aliased imports and submodule imports to avoid
```
MichaReiser: What's the reason for excluding aliased imports?

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+    fn from_binding_id(semantic: &'a SemanticModel<'b>, id: BindingId, scope: &'a Scope) -> Self {
+        let unused: Vec<_> = scope
+            .shadowed_bindings(id)
+            .map(|id| semantic.binding(id))
+            .collect();
+
+        let num_unused = &unused.len();
+
+        Self {
+            bindings: unused,
```
MichaReiser: Nit: You can avoid the temporary variable by flipping the field order
```suggestion
        Self {
            used: vec![false; unused.len()],
            bindings: unused,
        }
```

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+        })
+}
+
+#[derive(Debug)]
+struct MarkedBindings<'a, 'b> {
+    bindings: Vec<&'a Binding<'b>>,
+    used: Vec<bool>,
+}
+
+impl<'a, 'b> MarkedBindings<'a, 'b> {
```
MichaReiser: Nit: I found the name `unused` slightly mislieading. At this point, we don't know yet which bindings are or aren't used. 


```suggestion
        let bindings: Vec<_> = scope
```

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+
+/// Returns a `Vec` of bindings to unused import statements that
+/// are shadowed by a given binding.
+///
+/// Beginning with the collection of all bindings shadowed by
+/// the given one, we iterate over references to the module.
+/// Associated to each reference, we attempt to build a [`QualifiedName`]
+/// corresponding to an iterated attribute access (e.g. `a.b.foo`).
+/// We then determine the closest matching import statement to that
+/// qualified name, and mark it as used.
```
MichaReiser: You explained it in the PR summary but a future reader might not have the context. 

It wasn't immediately clear why we need to collect all bindings and why they are shadowed and what the current binding is. 

I really like the example, maybe we can move the example more to the top, then explain that `id` points to the binding of `a` created by `import a.b.c` and that said import shadows the bindings introduced by `import a` and `import a.b` (which we need to check if they're used)

This might also be a good place to document that, as of today (18 Sept 2025), Ruff's semantic index only creates a single binding for `a`

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+            } else {
+                vec![bdg]
+            }
+        })
+}
+
+#[derive(Debug)]
+struct MarkedBindings<'a, 'b> {
+    bindings: Vec<&'a Binding<'b>>,
+    used: Vec<bool>,
```
MichaReiser: Nit: I would move this definition after `unused_imports_from_binding`. `unused_imports_from_binding` seems more important (and provides way more context) to someone reading the code. The `MarkedBindings` in between is sort of distracting from the most important. (I think it's called newspaper style, obviously, something you can't do in Python)

Obviously, this very much comes down to personal preference. So feel free to disregard.

https://kentcdodds.com/blog/newspaper-code-structure

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+
+    let binding = semantic.binding(id);
+
+    // ensure we only do this once
+    let mut marked_dunder_all = false;
+
+    for ref_id in binding.references() {
+        let resolved_reference = semantic.reference(ref_id);
+        if !marked_dunder_all && resolved_reference.in_dunder_all_definition() {
+            let first = binding
```
MichaReiser: The `expect` message here is helpful in the sense that it makes it easier to find where the program is panicking if this condition isn't satisfied and debug symbols are off. However, it isn't very helpful for a future reader because it doesn't clarify **why** this invariant should hold true. 

The rust documentation provides some guidance on how to phrase your message instead: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style


`bindings ith shadowed bindings that aren't imports should be skipped by `unused_imports_in_scope``

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+                .as_any_import()
+                .expect("Binding to be an import binding")
```
MichaReiser: Nit: Maybe make this a method on `qualified_name` instead, e.g. `name.root()` (and can we guarantee that it's always set)

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+                .qualified_name()
+                .segments()[0];
+            mark_uses_of_qualified_name(&mut marked, &QualifiedName::user_defined(first));
+            marked_dunder_all = true;
+            continue;
+        }
+        let Some(expr_id) = resolved_reference.expression_id() else {
```
MichaReiser, line 738: This makes sense to me. If you have, it might help future readers to add an example here

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+        if *is_used {
+            continue;
+        }
+
+        if binding
+            .as_any_import()
+            .is_some_and(|imp| imp.qualified_name() == bname)
+        {
+            *is_used = true;
+        }
```
MichaReiser, line 852: I found the `bimp` and `bname` naming here more confusing than useful. I suggest being more explicit and naming the variables `best_import`, `best_name`.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+    for (binding, is_used) in marked.iter_mut() {
+        if *is_used {
+            continue;
+        }
+
+        if binding
+            .as_any_import()
+            .is_some_and(|imp| imp.qualified_name() == bname)
+        {
+            *is_used = true;
```
MichaReiser, line 851: Can there be multiple bindings for the same name or could be abort the loop when we found the binding? 

It also feels a bit unnecessary that we find the best matching binding in `best_match` first and then have to loop through all bindings again to mark said binding as used. Could `best_match` return the index or even better, the best matching binding and the mutable `used` variable?

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+    let qname = import.qualified_name();
+    let left = qname
+        .segments()
+        .iter()
+        .zip(prototype.segments())
+        .take_while(|(x, y)| x == y)
+        .count();
+    (left, std::cmp::Reverse(qname.segments().len()))
+}
+
```
MichaReiser, line 879: Rename this to `bindings`. We iterate over all bindings, ignoring if they have been used or not

This is also something that you could make a method on `MarkedBindings`
```suggestion
    unused: &'a [&'b Binding<'c>],
```

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+        }
+
+        if binding
+            .as_any_import()
+            .is_some_and(|imp| imp.qualified_name() == bname)
+        {
+            *is_used = true;
+        }
+    }
+}
```
MichaReiser, line 854: I'd find some documentation on how the ranking works useful. 

What I understand is that it takes the common segments and counts how many common segments there are. 

It's not entirely clear to me why we need to include the `qname.segments().len()`. 

```py
import a.b.c
import a.b.d.e

a.b
```

Is my understanding correct that this would prefer `a.b.d.e` over `a.b.c`? What's the motivation for this? Why not use the first/last? 


```py
import a.b.c
import a.b.d.e

a.b.x.f
```

This would return `(2, 3)` for `a.b.c` and `(2, 4)` for `a.b.d.e` where both aren't valid matches. Is this handled somewhere else?

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
MichaReiser: I don't think you need `'a` because it isn't used in the return type. Try omitting it

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+        .count();
+    (left, std::cmp::Reverse(qname.segments().len()))
+}
+
+fn best_match<'a, 'b, 'c>(
+    unused: &'a Vec<&'b Binding<'c>>,
+    prototype: &'a QualifiedName,
+) -> Option<&'b Binding<'c>> {
+    unused
+        .iter()
```
MichaReiser, line 889:
```suggestion
        .copied()
```

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+            )
+        })
+        .filter(|(_, bdg)| !bdg.is_global() && !bdg.is_nonlocal() && !bdg.is_explicit_export())
+        .flat_map(|(id, bdg)| {
+            if is_refined_submodule_import_match_enabled(settings)
+                // Only apply the new logic when all shadowed bindings
+                // are un-aliased imports and submodule imports to avoid
+                // complexity, false positives, and intersection with
+                // `redefined-while-unused` (`F811`).
+                && scope.shadowed_bindings(id).all(|shadow| {
```
MichaReiser: I'm not sure if it's worth it but we could consider skipping the new logic if `id` has now shadowed bindings except itself (which should allow us to skip some unnecessayr work)

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+
+    while let Some(node_id) = semantic.parent_expression_id(expr_id) {
+        let Some(expr) = semantic.expression(node_id) else {
+            break;
+        };
+        let Some(expr_attr) = expr.as_attribute_expr() else {
+            break;
+        };
+        builder.push(expr_attr.attr.as_str());
+        expr_id = node_id;
```
MichaReiser, line 813: I assume this is the same as `UnqualifiedName::from_expr`. The main reason we don't use it here is that we first need to find the top-level expression by following all ids

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,247 @@ fn fix_by_reexporting<'a>(
...
+    }
+    Some(builder.build())
+}
+
```
MichaReiser, line 817: I think one bug in the current implementation is that we iterate over all bindings, without taking into account whether they're visible at the reference's point

```py
import a

a.b

import a.b
```


In preview, this reports one error that `import a` is unused which is incorrect because `import a.b` only comes after

MichaReiser (commented):
This is a masive improvement with very low overhead. Well done.

I've a couple of nit comments but I also found a correctness bug that we need to look into. 

I haven't reviewed all tests yet.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: I'll add some documentation, but also answer here.

First, it's the opposite of what you suggest - we pick the _shortest_ path that matches. So in your example

```python
import a.b.c
import a.b.d.e

a.b
```

we would consider `a.b.c` as used, and in your next example:

```python
import a.b.c
import a.b.d.e

a.b.x.f
```

it would consider `a.b.c` as used. Also note that while it seems like...

> where both aren't valid matches

these _are_ all valid matches because of the strange behavior of Python imports described in the PR summary.

The motivation for marking the shortest as used as opposed to last one is because otherwise you wouldn't get the correct answer even in the most common situations:

```python
import a
import a.b

a.foo()
```

here both `a.b` and `a` match `a.foo` at a prefix of length `1`, so they're tied. Among those we pick the _shortest_ since we want to mark `import a` as used.

Flipping the order gives a counterexample to the heuristic of choosing the _first_ one, as well.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: We loop through because we may have repeated imports like

```python
import a
import a

a.foo()
```

so we have to mark all of them as used otherwise they will be left over and trigger this lint. But it's the job of [redefined-while-unused (F811)](https://docs.astral.sh/ruff/rules/redefined-while-unused/#redefined-while-unused-f811) to catch this situation, not `F401`

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: `copied` is not a method on `Vec`. Do you want me to use `into_iter` first or something?

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Yes, we are already walking upwards anyway so I figured we could collect the name on the way instead of backtracking once we get to the top.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Hmm, let me play around with this and also add some more tests - thanks for catching it!

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Just to avoid complexity - I didn't think it through very hard. It's possible we could handle them. I just worry vaguely about things like

```python
import a as b
import b

b.foo()
```

but maybe I shouldn't.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
ntBre: Probably `iter().copied()` to avoid the `map(|v| &**v)` at the end

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Added via

```rust
            if is_refined_submodule_import_match_enabled(settings)
                // No need to apply refined logic if there is only a single binding
                && scope.shadowed_bindings(id).nth(1).is_some()
```

is there a more idomatic way to do that? (I want to check that the iterator has at least two elements, but I don't want to use, say, `count` because there's no need to keep iterating once you've found two elements).

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: In fact I don't _think_ the `else` branch is possible... marking it as `unreachable` does not cause any tests to panic. But I couldn't find an actual justification so I took the coward's way out... I'll try to dig a little and see if it's possible

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Couldn't find a clean way to handle this "correctly" with the new logic, so instead I detect this situation now and revert to the stable behavior in that case. This means that we no longer give the ideal diagnostic for

```python
import a
a.foo()
import a.b
```

We used to mark the last import as unused, but now there's no lint error.

Note this is all related to a somewhat fishy choice of the semantic model: we copy over all references to an import binding to the _new_ binding when we have a second import statement. So, in the above example, at the end of the file we think that `a.foo()` is a reference for `import a.b`, which is a bit weird. It may be worth investigating the history of that decision and whether it could be modified to make the present logic simpler. But I think that'd be out of scope for now.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: Reworded/expanded

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: > (and can we guarantee that it's always set)

Can you say more about what you had in mind here?

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,284 @@ fn fix_by_reexporting<'a>(
...
+                    // Bail if one of the shadowed bindings is
+                    // used before the last live binding. This is
+                    // to avoid situations like this:
+                    //
+                    // ```
+                    // import a
+                    // a.b
+                    // import a.b
+                    // ```
+                    if i>0 && shadowed_binding.is_used() {
```
MichaReiser: Did you break cargo fmt or does rustfmt not add spaces around operators (`i > 0`)?

Yeah, it looks like you broke rustfmt 😆 

Adding an empty line in front of the `matches` makes it recover for some lines but it still fails for the `matches` call. I'm not sure what's wrong with your match call that it breaks formatting but I suspect it might help rustfmt (and readers) if you extract parts of the if condition into standalone expressions (just a tiny bit)

Oh, I think rustfmt struggles with the leading comments between the different expressions in the condition. Extracting some variables (or functions) is probably your best solution

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,284 @@ fn fix_by_reexporting<'a>(
...
+        .take_while(|(x, y)| x == y)
+        .count();
+    (left, std::cmp::Reverse(qname.segments().len()))
+}
+
+/// Returns the import binding that shares the longest prefix
+/// with the `prototype` and is of minimal length amongst these.
+///
+/// See also [`rank_matches`].
+fn best_match<'b, 'c>(
```
MichaReiser: Nit: Using `'b` and `'c` without `'a` is a bit weird. I suggest renaming them to `'a` and `'b` (a reader otherwise wonders why `b` and `c`)

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
MichaReiser: There's a hidden `unwrap` call when you access `segments[0]`. There's an implicit assumption that `segments` is never empty. 

```
binding
                .as_any_import()
                .expect("binding to be import binding since current function called after restricting to these in `unused_imports_in_scope`")
                .qualified_name()
                .segments()[0];
```

I haven't checked all code paths but I think the assumption is true for all `QualifiedName`. If so, then having a method on `QualifiedName` that returns the first segment (and documents why doing so without checking the length of segments is okay) seems generally useful (e.g. `first_segment`, `root_name`, what not :))

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
MichaReiser: Oh I see. This would make a great inline comment

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -582,3 +614,284 @@ fn fix_by_reexporting<'a>(
...
+/// name associated to the reference in the last line is `"a.b.foo"`.
+/// The closest matches are `import a` and `import a.b`, respectively,
+/// leaving `import a.b.c` unused.
+///
+/// For a precise definition of "closest match" see [`best_match`]
+/// and [`rank_matches`].
+///
+/// Note: if any reference comes from something other than
+/// a `Name` or a dunder all expression, then we return just
+/// the original binding, thus reverting the stable behavior.
```
MichaReiser, line 713: This documentation with the example is great!

crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__f401_import_submodules_in_function_scope.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: crates/ruff_linter/src/rules/pyflakes/mod.rs
+---
+F401 [*] `a` imported but unused
+ --> f401_preview_submodule.py:2:8
+  |
+2 | import a
```
MichaReiser: I found this a bit surprising but it's the same behavior as on main. 

Which makes me wonder if we should have used the assertion helper that compares preview with stable behavior.

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
dylwil3: There's one place in Ruff where it's not clear to me that this invariant is satisfied:

https://github.com/astral-sh/ruff/blob/2af8c53110227598cb4758a777667dbe18bc7f15/crates/ruff_python_ast/src/helpers.rs#L908-L918

So for now I've added an explicit `expect` here, similar to what's done elsewhere like 

https://github.com/astral-sh/ruff/blob/2af8c53110227598cb4758a777667dbe18bc7f15/crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs#L494-L498

crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__f401_import_submodules_in_function_scope.snap
dylwil3: Right now our assertion helper only works for fixtures passed as file paths, whereas here I've written the source code inline. So for now I've lazily added comments to the fixtures indicating expected behavior 😄

PR #20201, opened 2025-09-02
[`flake8-comprehensions`] Skip `C417` when lambda contains `yield`/`yield from` (`C417`)

### LINKED ISSUE #20198: C417 fix introduces a syntax error for `yield` in lambda expression
### Summary

The fix for [`unnecessary-map` (C417)](https://docs.astral.sh/ruff/rules/unnecessary-map/) introduces a syntax error when the lambda expression contains a yield expression. The fix, and possibly the rule, should be suppressed in that case. [Example](https://play.ruff.rs/dffb7894-a98f-4643-9efd-67ad02563071):
```console
$ cat >c417.py <<'# EOF'
base = 0
gens = map(lambda x: (yield base + x), [1, 2, 3])
for gen in gens:
    print(*[*gen])
    base += 20
# EOF

$ python c417.py
1
22
43

$ ruff --isolated check c417.py --select C417 --unsafe-fixes --fix

$ python c417.py 2>&1 | tail -n 3
    gens = ((yield base + x) for x in [1, 2, 3])
             ^^^^^^^^^^^^^^
SyntaxError: 'yield' inside generator expression
```

### Version

ruff 0.12.11 (c2bc15bc1 2025-08-28)

danparizher:
## Summary

Fixes #20198

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_comprehensions/C417.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_comprehensions/C417.py
@@ -75,3 +75,11 @@ list(map(lambda x, y: x, [(1, 2), (3, 4)]))
 _ = t"{set(map(lambda x: x % 2 == 0, nums))}"
 _ = t"{dict(map(lambda v: (v, v**2), nums))}"
 
+
+# See https://github.com/astral-sh/ruff/issues/20198
+# No error: lambda contains `yield`, so map() should not be rewritten
+base = 0
+gens = map(lambda x: (yield base + x), [1, 2, 3])
+for gen in gens:
+    _ = [*gen]
+    base += 20
--- a/crates/ruff_linter/src/rules/flake8_comprehensions/rules/unnecessary_map.rs
+++ b/crates/ruff_linter/src/rules/flake8_comprehensions/rules/unnecessary_map.rs
@@ -122,6 +122,13 @@ pub(crate) fn unnecessary_map(checker: &Checker, call: &ast::ExprCall) {
         }
     };
 
+    // If the lambda body contains a `yield` or `yield from`, rewriting `map(lambda ...)` to a
+    // generator or comprehension changes the semantics or is outright invalid syntax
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/resources/test/fixtures/flake8_comprehensions/C417.py
```diff
@@ -75,3 +75,11 @@ def func(arg1: int, arg2: int = 4):
...
 _ = t"{dict(map(lambda v: (v, v**2), nums))}"
 
+
+# See https://github.com/astral-sh/ruff/issues/20198
+# No error: lambda contains `yield`, so map() should not be rewritten
+base = 0
+gens = map(lambda x: (yield base + x), [1, 2, 3])
+for gen in gens:
+    _ = [*gen]
+    base += 20
```
ntBre: nit: We might want to strip this down a bit like the other test cases:


```suggestion
map(lambda x: (yield x), [1, 2, 3])
```

crates/ruff_linter/src/rules/flake8_comprehensions/rules/unnecessary_map.rs
```diff
@@ -122,6 +122,13 @@ pub(crate) fn unnecessary_map(checker: &Checker, call: &ast::ExprCall) {
         }
     };
 
+    // If the lambda body contains a `yield` or `yield from`, rewriting `map(lambda ...)` to a
+    // generator or comprehension changes the semantics or is outright invalid syntax
```
ntBre: Do you have a case in mind where it only changes the semantics? I thought it would always cause a syntax error.

ntBre (commented):
Thank you! Just one nit about the test case. Nice use of `any_over_expr`!

crates/ruff_linter/src/rules/flake8_comprehensions/rules/unnecessary_map.rs
danparizher: You're right, I'll update the comment to make it more accurate.

PR #20207, opened 2025-09-02
[ty] Add naive implementation of completions for unimported symbols

BurntSushi:
This PR takes a first step toward implementing auto-import (see #535).

There are two main reasons why this is considered naive:

1. Once a helping of dependencies have been added to an environment,
   completions can become quite slow. Into the hundreds of milliseconds.
   I've intentionally not spent any time on performance here. We're
   just using a per-file Salsa caching mechanism and searching is done
   exhaustively.
2. When a completion for an unimported symbol is selected, the necessary
   import is not yet added. (This is what I plan to focus on next.)

For those two reasons, these completions will only appear when users opt
into them via the `ty.experimental.autoImport` setting.

This PR is best reviewed commit by commit.


BurntSushi:
Demo:

https://github.com/user-attachments/assets/f960feb7-3cf0-4089-b7f8-7bceb3fda3e9

AlexWaygood:
> 2\. When a completion for an unimported symbol is selected, the necessary
>         import is not yet added. (This is what I plan to focus on next.)

We already have quite good infrastructure for this for some Ruff autofixes. It might be worth seeing if that can be factored out and put somewhere where it could be reused by ty. (It also might be completely impossible to reuse it, but it could be worth a look.) It's found here: https://github.com/astral-sh/ruff/tree/main/crates/ruff_linter/src/importer

BurntSushi:
@AlexWaygood Nice yeah! I was planning to look for that. Thanks for finding it. :-)

crates/ty_ide/src/all_symbols.rs
```diff
@@ -0,0 +1,275 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AllSymbolInfo {
+    /// The symbol information
+    pub symbol: SymbolInfo<'static>,
+    /// The file containing the symbol
+    pub file: File,
+}
+
+#[cfg(test)]
+mod tests {
```
dhruvmanila, line 67: I think it's fine but I was wondering if we want to restrict the test cases to avoid the builtins because any changes to typeshed would require updating the snapshots here if it's related. And, I think this also allows us to get coverage over the builtins symbols which seems useful.

crates/ty_server/src/session/options.rs
```diff
@@ -149,6 +155,7 @@ impl GlobalOptions {
             .experimental
             .map(|experimental| ExperimentalSettings {
                 rename: experimental.rename.unwrap_or(true),
```
dhruvmanila: Oh shoot, the `unwrap_or` here was meant to be `false`. Can you make this change? Luckily, no one should be affected because there was only one setting in the experimental group.

crates/ty_python_semantic/src/module_resolver/module.rs
dhruvmanila, line 1: It might be useful to have @AlexWaygood to take a look at the changes here once he is back.

crates/ty_ide/src/symbols.rs
```diff
@@ -282,6 +283,27 @@ impl SymbolKind {
...
+            SymbolKind::Constant => CompletionKind::Constant,
+            SymbolKind::Property => CompletionKind::Property,
+            SymbolKind::Field => CompletionKind::Field,
+            SymbolKind::Constructor => CompletionKind::Constructor,
+            SymbolKind::Parameter => CompletionKind::Variable,
+            SymbolKind::TypeParameter => CompletionKind::TypeParameter,
+            // Not quite sure what to do with this one. I guess
+            // in theory the import should be "resolved" to its
+            // underlying kind, but that seems expensive.
+            SymbolKind::Import => return None,
```
dhruvmanila, line 304: Yeah, I think it's fine to not do this now. Pylance does provide the icon and the documentation as well.

dhruvmanila (approved):
This looks great!

If it's not already planned, it would be great if you could update the ty documentation for editor settings to include the new setting over at https://github.com/astral-sh/ty/blob/59bf06d2675cc6add7b4ba6f603f70edebcf9499/docs/reference/editor-settings.md#L4 and update the VS Code extension to include the setting ([reference change](https://github.com/astral-sh/ty-vscode/commit/e325761ea4f547b03fec7d9d26c9c7dae8851321)).

crates/ty_server/src/session/options.rs
BurntSushi: Yup, done!

crates/ty_python_semantic/src/module_resolver/module.rs
BurntSushi: Aye. I'd love that. I'll probably merge before then, but I'd be happy to address feedback post-merge.

(I think one known deficiency is that namespace packages aren't really supported. Which matches the status quo. I think they're going to need a dedicated strategy to deal with.)

crates/ty_ide/src/all_symbols.rs
BurntSushi: Seems sensible. Done!

PR #20208, opened 2025-09-02
Using Ruff as the default linter in Orionis Framework

rmunate:
Hello team! 👋
I’m truly delighted with everything Ruff brings to the table. Our open-source project, *Orionis Framework*, is evolving with many features inspired by Laravel’s philosophy but built in Python. Every new feature we develop is reviewed with Ruff, and in fact, one of the framework’s native commands will integrate Ruff as its default linter — explicitly pointing to this official repository. 🚀
Thanks for creating such an amazing tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,11 @@ of [FastAPI](https://github.com/tiangolo/fastapi):
 > Ruff is so fast that sometimes I add an intentional bug in the code just to confirm it's actually
 > running and checking the code.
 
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thank you for the kind words! We tend to limit the number of testimonials that we include here to keep the README short and impactful, so I'm going to omit this for now, but I do appreciate the PR and the encouragement.

Issue #20214, opened 2025-09-03
`os-path-expanduser` (PTH111) should be marked unsafe


dscorbett:
### Summary

The fix for [`os-path-expanduser` (PTH111)](https://docs.astral.sh/ruff/rules/os-path-expanduser/) should be marked unsafe because the replacement is not always equivalent. If `os.path.expanduser` can’t determine a home directory, it returns the input unchanged, whereas `pathlib.Path.expanduser` raises an error. One reason it might not be able to determine a home directory is if the indicated user does not exist, but that is not the only possible reason. [Example](https://play.ruff.rs/7b404540-09b0-48be-a3a0-3fb6ac57cd5d):
```console
$ cat >pth111.py <<'# EOF'
import os
try: print(os.path.expanduser("~pth111_nonexistent_user"))
except RuntimeError as e: print(e)
# EOF

$ python pth111.py
~pth111_nonexistent_user

$ ruff --isolated check pth111.py --select PTH111 --preview --fix
Found 1 error (1 fixed, 0 remaining).

$ cat pth111.py
import os
import pathlib
try: print(pathlib.Path("~pth111_nonexistent_user").expanduser())
except RuntimeError as e: print(e)

$ python pth111.py
Could not determine home directory.
```

### Version

ruff 0.12.11 (c2bc15bc1 2025-08-28)

PR #20215, opened 2025-09-03
[`flake8-use-pathlib`] Make `PTH111` fix unsafe because it can change behavior

chirizxc:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/20214

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
@@ -1,11 +1,13 @@
-use ruff_diagnostics::Applicability;
+use ruff_diagnostics::{Edit, Fix};
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::ExprCall;
+use ruff_text_size::Ranged;
 
 use crate::checkers::ast::Checker;
+use crate::importer::ImportRequest;
 use crate::preview::is_fix_os_path_abspath_enabled;
 use crate::rules::flake8_use_pathlib::helpers::{
-    check_os_pathlib_single_arg_calls, has_unknown_keywords_or_starred_expr,
+    has_unknown_keywords_or_starred_expr, is_pathlib_path_call,
 };
 use crate::{FixAvailability, Violation};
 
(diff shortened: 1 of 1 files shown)


dscorbett:
~[Apparently](https://github.com/search?q=language%3APython+%2Fexpanduser%5C%28%5Bfr%5D*%5B%27%22%5D%5B%5E%7E%7B%5D%2F&type=code), it’s not uncommon for `os.path.expanduser` to be called on a string literal that doesn’t start with `"~"`, in which case the PTH111 fix changes a (probably unintended) no-op into a guaranteed error. Should the fix be suppressed or changed in that case?~ This is not true.

chirizxc:
> [Apparently](https://github.com/search?q=language%3APython+%2Fexpanduser%5C%28%5Bfr%5D*%5B%27%22%5D%5B%5E%7E%7B%5D%2F&type=code), it’s not uncommon for `os.path.expanduser` to be called on a string literal that doesn’t start with `"~"`, in which case the PTH111 fix changes a (probably unintended) no-op into a guaranteed error. Should the fix be suppressed or changed in that case?

I think in that case it's better to offer only diagnostic

eli-schwartz:
> [Apparently](https://github.com/search?q=language%3APython+%2Fexpanduser%5C%28%5Bfr%5D*%5B%27%22%5D%5B%5E%7E%7B%5D%2F&type=code), it’s not uncommon for `os.path.expanduser` to be called on a string literal that doesn’t start with `"~"`, in which case the PTH111 fix changes a (probably unintended) no-op into a guaranteed error. Should the fix be suppressed or changed in that case?

It which?

```python
>>> pathlib.Path('.').expanduser()
PosixPath('.')
```

There's nothing to catch unless a genuine ~ exists.

ntBre:
@chirizxc is this ready for review? It sounds like the `~` question is resolved, so I think it's still fine just to mark the fix as unsafe.

chirizxc:
> @chirizxc is this ready for review? It sounds like the `~` question is resolved, so I think it's still fine just to mark the fix as unsafe.

yes

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
```diff
@@ -75,43 +73,17 @@ pub(crate) fn os_path_abspath(checker: &Checker, call: &ExprCall, segments: &[&s
...
-        } else {
-            format!("{binding}({arg_code}).resolve()")
-        };
-
-        Ok(Fix::unsafe_edits(
-            Edit::range_replacement(replacement, range),
-            [import_edit],
-        ))
-    });
+    check_os_pathlib_single_arg_calls(
```
ntBre: This is a different rule, not PTH111 right? I'd prefer moving this change to a separate PR to keep this one focused on the fix safety issue.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_expanduser.rs
```diff
@@ -35,7 +36,10 @@ use crate::{FixAvailability, Violation};
 /// especially on older versions of Python.
 ///
 /// ## Fix Safety
-/// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
+/// This rule's fix is always marked as unsafe because `os.path.expanduser` and `Path.expanduser` differ:
+/// if the home directory can't be resolved, `os.path.expanduser` returns the
+/// input unchanged, while `Path.expanduser` raises `RuntimeError`. This can
+/// change behavior in edge cases.
```
ntBre:
```suggestion
/// This rule's fix is always marked as unsafe because the behaviors of 
/// `os.path.expanduser` and `Path.expanduser` differ when a user's home 
/// directory can't be resolved: `os.path.expanduser` returns the
/// input unchanged, while `Path.expanduser` raises `RuntimeError`.
```

I think something like that sounds a little nicer, but feel free to tweak it further if you prefer.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
chirizxc: I've sent this change as a separate PR

ntBre (approved):

PR #20217, opened 2025-09-03
[ty] Add functions for revealing assignability/subtyping constraints

dcreager:


crates/ty_test/src/matcher.rs
```diff
@@ -297,21 +306,53 @@ impl Matcher {
                 }
             }
             ParsedAssertion::Revealed(expected_type) => {
-                #[cfg(not(debug_assertions))]
-                let expected_type = discard_todo_metadata(&expected_type);
+                let expected_type = discard_todo_metadata(expected_type);
```
dcreager, line 309: This makes sure that we have a consistent type in both debug and release modes, so that we can compare with `expected_type` down below

crates/ty_vendored/ty_extensions/ty_extensions.pyi
```diff
@@ -51,6 +51,12 @@ def is_disjoint_from(type_a: Any, type_b: Any) -> bool: ...
 def is_singleton(type: Any) -> bool: ...
 def is_single_valued(type: Any) -> bool: ...
 
+# These are the same as above, but instead of returning _whether_ the property
+# holds, it shows a diagnostic that describes under what constraints the
+# property holds.
+def reveal_when_assignable_to(type_target: Any, type_source: Any) -> None: ...
+def reveal_when_subtype_of(type_target: Any, type_source: Any) -> None: ...
```
sharkdp: Source and target seem to be swapped here? (and also above, in the existing function)

It's `is_assignable_to(source, target)` or `is_assignable_to(from, to)` or `is_assignable_to(rhs_of_assignment, lhs_of_assignment)`. https://play.ty.dev/47621237-8fec-475e-b668-56905b91252b

sharkdp (approved):
Cool, thank you!

crates/ty_vendored/ty_extensions/ty_extensions.pyi
dcreager: Yes you're right!  One of the other functions used the less opinionated `type_a` and `type_b`, so I standardized all of the signatures on that. (And cleaned up a couple that were using `type` as a parameter name, which even if it's not a syntax error is confusing as it conflicts with the new type alias keyword)

PR #20219, opened 2025-09-03
[ty] Add support for generic PEP695 type aliases

ibraheemdev:
## Summary

Adds support for generic PEP695 type aliases, e.g.,
```python
type A[T] = T
reveal_type(A[int]) # A[int]
```

I'm not at all confident in any of the new behavior defined on `Type::GenericAlias(GenericAlias::TypeAlias(_))`, so I would appreciate a thorough review here. In the ideal world it would act identically to a `Type::GenericAlias(GenericAlias::ClassLiteral(_)))`, but I believe it is possible for a generic type alias to refer to a non-generic class (e.g. `type A[T] = int`), so the behavior might differ.

Resolves https://github.com/astral-sh/ty/issues/677.


crates/ty_python_semantic/src/types.rs
```diff
@@ -1248,7 +1258,12 @@ impl<'db> Type<'db> {
                 Some(ClassType::NonGeneric(class_literal).into_callable(db))
             }
 
-            Type::GenericAlias(alias) => Some(ClassType::Generic(alias).into_callable(db)),
+            Type::GenericAlias(GenericAlias::ClassLiteral(alias)) => {
+                Some(ClassType::Generic(alias).into_callable(db))
+            }
+            Type::GenericAlias(GenericAlias::TypeAlias(alias)) => {
+                alias.value_type(db).into_callable(db)
```
ibraheemdev: For example, I'm not sure if delegating to `value_type` here is correct, because `value_type` returns a `Type::NominalInstance` for the RHS of the alias. I'm not very familiar with the typing hierarchy, but that doesn't seem correct to me.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -364,6 +464,24 @@ impl<'db> VarianceInferable<'db> for GenericAlias<'db> {
     }
 }
 
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
+pub struct GenericTypeAliasAlias<'db> {
```
ibraheemdev: Quite an unfortunate name here.

sharkdp:
Not a full review, but the new `too-many-positional-arguments` diagnostics in the ecosystem diff indicate that there is some problem with `ParamSpec`s. For example:
```py
from typing import Callable

type ReturnsInt[**P] = Callable[P, int]

def call[**P](callable: ReturnsInt[P]) -> None:  # ty: too-many-positional-arguments
    pass
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
sharkdp, line 1: We also have `resources/mdtest/pep695_type_aliases.md`. Should these two files be merged?

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -243,11 +243,22 @@ impl CodeGeneratorKind {
 /// # Ordering
 /// Ordering is based on the generic aliases's salsa-assigned id and not on its values.
 /// The id may change between runs, or when the alias was garbage collected and recreated.
-#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
-#[derive(PartialOrd, Ord)]
-pub struct GenericAlias<'db> {
```
dcreager: Did you consider keeping this a struct, and making just the `origin` field be an enum? Even if we need to add additional variants in the future, I can't imagine a generic alias not having a specialization, and it looks like a lot of the copied logic below is actually shared.

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -79,7 +79,14 @@ impl<'db> ClassBase<'db> {
...
             Type::Dynamic(dynamic) => Some(Self::Dynamic(dynamic)),
             Type::ClassLiteral(literal) => Some(Self::Class(literal.default_specialization(db))),
-            Type::GenericAlias(generic) => Some(Self::Class(ClassType::Generic(generic))),
+
+            Type::GenericAlias(generic) => match generic {
+                GenericAlias::ClassLiteral(generic) => {
+                    Some(Self::Class(ClassType::Generic(generic)))
+                }
+                GenericAlias::TypeAlias(_) => None,
+            },
```
dcreager: You can fold these match statements together like you do below in `display.rs`:

```suggestion
            Type::GenericAlias(GenericAlias::ClassLiteral(generic)) => {
                Some(Self::Class(ClassType::Generic(generic)))
            }
            Type::GenericAlias(GenericAlias::TypeAlias(_)) => None,
```

(Though if you revert `GenericAlias` back to being a struct per above, then you will need to keep this as two separate matches)

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -1734,7 +1736,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
     }
 
     fn infer_region_definition(&mut self, definition: Definition<'db>) {
-        match definition.kind(self.db()) {
+        match &definition.kind(self.db()) {
```
dcreager: Is this spurious? I don't see any other changes to this method or the type being matched on that would require this

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -8841,6 +8845,19 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         if let Type::SpecialForm(SpecialFormType::Tuple) = value_ty {
             return tuple_generic_alias(self.db(), self.infer_tuple_type_expression(slice));
         }
+        if let Type::KnownInstance(KnownInstanceType::TypeAliasType(TypeAliasType::PEP695(
+            type_alias,
+        ))) = value_ty
```
dcreager: I'd suggest adding a `TODO` comment that the limitation to PEP 695 type aliases is temporary, and only because that's the only kind of type alias that we've attached a generic context to yet.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -8853,6 +8870,50 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            specialize,
+        )
+    }
+
+    fn infer_explicit_callable_specialization(
+        &mut self,
+        subscript: &ast::ExprSubscript,
+        value_ty: Type<'db>,
+        generic_context: GenericContext<'db>,
+        specialize: impl FnOnce(&[Option<Type<'db>>]) -> Type<'db>,
```
dcreager, line 8889: I think the "keep `GenericAlias` as a struct" suggestion would simplify this a bit, too. You'd end up with a `GenericAliasOrigin` enum, which wraps either a class literal or a type alias, and you could update `infer_explicit_class_specialization` to take in that instead of a `ClassLiteral`. That would eliminate the need for this callback (and the two new methods here), since the two callbacks would be combined into a new `GenericAliasOrigin::apply_specialization` method.

ibraheemdev:
@sharkdp I believe that's https://github.com/astral-sh/ty/issues/157. All the generics infrastructure in this PR is reused from generic classes, which emit a similar error:

```py
class ReturnsInt[**P]:
    x: Callable[P, int]

def call[**P](callable: ReturnsInt[P]) -> None:  # ty: too-many-positional-arguments
    pass
````

**Review by dcreager — COMMENTED** [review]
Looking at how pyright [handles this](https://pyright-play.net/?enableExperimentalFeatures=true&code=MYGwhgzhAEDCDaAVAugLmgOiwKGwEwFMAzaAfSWQAph0EUBKVbaF6AJwIDcCwRSAXAJ4AHAtXq4O3XgJFjYEqTz5DRlBAEsAdv2QTcqgtAAaFaAF44FbIegBNC9Aj82uQiXIpKAD3SmGTKzsXMqyat767mSUguh2jMysSjKGMfrJKnKUxoohKVmm2rq50plq8UA), there might a simpler way to do this that punts on several of the questions we've had.

```py
class C[T]: ...

def _[T](c: C[T]):
    reveal_type(c)  # revealed: C[T@_]

reveal_type(C)  # revealed: type[C[Unknown]]
reveal_type(C[int])  # revealed: type[C[int]]


type X[T] = C[T]
type Y = str

def _[T](x: X[T]):
    reveal_type(x)  # revealed: C[T@_]

def _(y: Y):
    reveal_type(y)

reveal_type(X)  # revealed: TypeAliasType
reveal_type(X[int])  # revealed: TypeAliasType
reveal_type(Y)  # revealed: TypeAliasType
```

Note that in the last two function bodies, the revealed type does not mention the type alias at all. So in ty terms, `NominalInstance` does not need to wrap the (specialized) type alias.

Similarly, the revealed type of `X[int]` is `TypeAliasType`, and so `GenericAlias` doesn't need to wrap the type alias either.

This suggests that we can roll back all of the changes to `GenericAlias` and `NominalInstance`, and just make sure that `in_type_expression` propagates the generic context through when it resolves a type alias into its value type.

---

[I made several comments below about the current implementation, which might become moot given the above, but I'm keeping them here in case my analysis is wrong and we do need to keep those changes]

dcreager:
> and so `GenericAlias` doesn't need to wrap the type alias either

And in particular, this reminds me of an earlier draft of this that you had shown me, where `PEP695TypeAlias` had an `Option<Specialization>` field, since we do need _some_ way to distinguish `X` from `X[int]` in my example. And then `in_type_expression` would check if the type alias is generic, and if it is, fall back on the default specialization if that field is `None`.

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
carljm: Possibly, although this file is specific to generic PEP 695 aliases.

crates/ty_ide/src/hover.rs
```diff
@@ -1591,10 +1591,10 @@ def ab(a: int, *, c: int):
 
         // TODO: This should render T@Alias once we create GenericContexts for type alias scopes.
```
carljm: This comment should be removed.

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
```diff
@@ -0,0 +1,161 @@
...
+```py
+# error: [invalid-syntax] "duplicate type parameter"
+type RepeatedTypevar[T, T] = ...
+```
+
+You can only use typevars (TODO: or param specs or typevar tuples) in the alias's generic context.
+
+```py
+# TODO: error
+type GenericOfType[int] = ...
```
carljm: I don't think this example should be an error. In this case the name `int`, within the scope of this alias, is bound to a typevar, shadowing the global name `int`.

The names in brackets here are not references, they are targets to which a (typevar) value will be bound in this scope.

This shadowing is unnecessarily confusing, but it's not an error.

Here's this example [working in pyright](https://pyright-play.net/?code=C4TwDgpgBA4g2gSwHbALpQLxQDYIM7CIqoCwAUOQCYQBmUA%2BgBQDmAXLHAQE6oCUr5KEKhcIANwgBDbPVCQWvIA).

I think this test can just be removed? Any identifier is valid here, and anything invalid will be a syntax error.

crates/ty_python_semantic/src/types.rs
```diff
@@ -3589,7 +3602,7 @@ impl<'db> Type<'db> {
     /// is truthy or falsy in a context where Python doesn't make an implicit `bool` call.
     /// Use [`try_bool`](Self::try_bool) for type checking or implicit `bool` calls.
     pub(crate) fn bool(&self, db: &'db dyn Db) -> Truthiness {
-        self.try_bool_impl(db, true)
+        self.try_bool_impl(db, true, &TryBoolVisitor::new(Ok(Truthiness::Ambiguous)))
```
carljm, line 3605: We could implement `Default` for `TryBoolVisitor` to avoid repeating this a few places.

crates/ty_python_semantic/src/types.rs
```diff
@@ -5572,6 +5591,12 @@ impl<'db> Type<'db> {
             }),
 
             Type::KnownInstance(known_instance) => match known_instance {
+                KnownInstanceType::TypeAliasType(TypeAliasType::PEP695(alias))
+                    if alias.is_specialized(db) =>
+                {
+                    // Note that the RHS of a type alias statement is required to be in type-form.
+                    Ok(alias.value_type(db))
```
carljm: Is there a reason we need to eagerly unpack specialized aliases when they appear in type expressions, where we don't for non-generic aliases (see below, we wrap them in a lazy `Type::TypeAlias` instead).

I think this eager unpacking will result in Salsa query cycles for an example like `type Rec[T] = T | list[Rec[T]]`, because we will eagerly unpack the alias, recursively, and type inference of the RHS will diverge.

It looks to me like we can just delete this new arm entirely and just let the below arm do its thing?

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Specializing type aliases explicitly
```
carljm, line 43: What about when we use a generic type alias in a type expression without an explicit specialization? This should result in implicitly specializing to `Unknown`:

```py
type G[T] = list[T]

def _(g: G):
    reveal_type(g)  # should reveal `list[Unknown]`
```

But I don't see any tests for it, and I don't think the PR currently handles it correctly; instead we just fail to specialize at all, and allow a typevar to leak out, revealing `list[T@G]` above.

We should also have a test like the above but with a typevar default (`type G[T = int] = list[T]`) -- in this case we should specialize to `list[int]` when no explicit specialization is given.

(We already have code for creating the `default_specialization` for a generic context, so we shouldn't need to write new code to handle typevar defaults vs Unknown -- but we should have tests for both cases.)

crates/ty_python_semantic/src/types.rs
carljm: We might, though, need some additional handling here in case of a type alias with a generic context but without a specialization. Such an alias, if used in a type expression, should be implicitly "default specialized" (that is, specialized to the default for each typevar, or `Unknown` for typevars with no default.)

crates/ty_python_semantic/src/types.rs
```diff
@@ -6874,6 +6902,16 @@ impl<'db> KnownInstanceType<'db> {
                         f.write_str("typing.Generic")?;
                         generic_context.display(self.db).fmt(f)
                     }
+                    KnownInstanceType::TypeAliasType(TypeAliasType::PEP695(alias)) => {
+                        if let Some(specialization) = alias.specialization(self.db) {
```
carljm, line 6900: This is fine for now, but ultimately we will need to support all kinds of generic type aliases, so it might be a bit more forward-thinking to provide generalized API on `TypeAliasType` that matches variants internally, and avoid matching specifically on `TypeAliasType::PEP695` externally? But it's not a big deal, can always be changed later when we add support for other kinds of generic type aliases.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -8818,6 +8819,20 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         if let Type::SpecialForm(SpecialFormType::Tuple) = value_ty {
             return tuple_generic_alias(self.db(), self.infer_tuple_type_expression(slice));
         }
+        // TODO: Add support for non-PEP695 generic type aliases.
+        if let Type::KnownInstance(KnownInstanceType::TypeAliasType(TypeAliasType::PEP695(
```
carljm: Similar to my comment above, weak preference here to keep this TODO internal, not match here on specific kinds of aliases, and expose general API on `TypeAliasType` itself that we can use here without worrying about the details of which kinds of generic aliases are actually supported.

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
ibraheemdev: I took this from [the test in `generics/classes.md`](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/generics/pep695/classes.md?plain=1#L43). Is this different, or should that be removed as well?

crates/ty_python_semantic/src/types.rs
ibraheemdev: I wanted to avoid implementing `Default` for `Truthiness` because that felt like a potential footgun (and `TryBoolVisitor` is just a type alias so can't implement `Default` directly).

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
sharkdp: Yes. And I don't care too much about the structure/layout of the tests. Just wanted do let you know that there are some generics-related tests in that other file as well.

ibraheemdev:
The typing conformance suite is now panicking with `value_type(_): too many cycle iterations` on `aliases_type_statement.md` because of:
```py
type RecursiveTypeAlias1[T] = T | list[RecursiveTypeAlias1[T]]
r1_1: RecursiveTypeAlias1[int] = 1
```

This looks like https://github.com/astral-sh/ty/issues/256, which I believe is unrelated to this PR. The ecosystem failures also all seem to be cases where recognizing generic type aliases leads to more detailed diagnostics for underlying semantics that we do not support.

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
carljm: No, that test is similarly bogus and should be removed.

crates/ty_python_semantic/src/types.rs
carljm: I don't think you need to implement `Default` for `Truthiness` in order to implement it for `CycleDetector<TryBool, Type<'db>, Result<Truthiness, BoolError<'db>>>` (and you can `impl Default for TryBoolVisitor` as a shortcut for the latter). I could be missing something but I'm pretty sure I've done this before, see e.g. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8123cdc184c2e6a38c393f541dfbd151 (`V` doesn't implement `Default`)

carljm:
> The typing conformance suite is now panicking with `value_type(_): too many cycle iterations` on `aliases_type_statement.md`

Let me look at this. We avoid this already for PEP 695 aliases in general by creating a type alias type for the assignment without resolving the RHS, and deferring resolution of the RHS as part of `value_type` on the alias. It seems like something in this PR must be breaking this (i.e., resolving the RHS too eagerly) for generic aliases. If necessary we can fix it in a separate follow-up, but it is something that I would expect to work at this point, it's not an expected failure.

crates/ty_python_semantic/src/types.rs
ibraheemdev: You run into the orphan rule because `CycleDetector<_, _, R: Default>` implements `Default`, and `R` here is an external type (`Result<_, _>`). I don't think it's a huge deal but I'm happy to implement `Default` for `Truthiness` if you think that's better.

carljm: Ah makes sense! No I think it's fine as-is, just wanted to understand :)

carljm:
Hmm I see the issue -- it's not that we resolve the RHS too eagerly in general, it's that when subscripting we resolve the value-type of the alias at that point to apply the specialization. We may need to allow `Type::TypeAlias` to carry a lazy specialization with it in order to support this. In any case, having looked at it I agree this isn't an issue we should try to solve in this PR.

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
carljm: Good point. I think we can look at this as a follow-up.

PR #20221, opened 2025-09-03
[ty] Attribute access on top/bottom materializations

JelleZijlstra:
## Summary

Part of astral-sh/ty#994. The goal of this PR was to add correct behavior for attribute access on the top and bottom materializations. This is necessary for the end goal of using the top materialization for narrowing generics (`isinstance(x, list)`): we want methods like `x.append` to work correctly in that case.

It turned out to be convenient to represent materialization as a TypeMapping, so it can be stashed in the `type_mappings` list of a function object. This also allowed me to remove most concrete `materialize` methods, since they usually just delegate to the subparts of the type, the same as other type mappings. That is why the net effect of this PR is to remove a few hundred lines.

## Test Plan

I added a few more tests. Much of this PR is refactoring and covered by existing tests.

## Followups

Assigning to attributes of top materializations is not yet covered. This seems less important so I'd like to defer it.

I noticed that the `materialize` implementation of `Parameters` was wrong; it did the same for the top and bottom materializations. This PR makes the bottom materialization slightly more reasonable, but implementing this correctly will require extending the struct.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -820,7 +820,11 @@ impl<'db> Type<'db> {
     /// - In covariant position, it's replaced with `object`
     /// - In contravariant position, it's replaced with `Never`
     /// - In invariant position, it's replaced with an unresolved type variable
-    fn materialize(&self, db: &'db dyn Db, materialization_kind: MaterializationKind) -> Type<'db> {
+    pub(crate) fn materialize(
+        &self,
+        db: &'db dyn Db,
+        materialization_kind: MaterializationKind,
+    ) -> Type<'db> {
         match self {
             Type::Dynamic(_) => match materialization_kind {
                 MaterializationKind::Top => Type::object(db),
@@ -850,10 +854,14 @@ impl<'db> Type<'db> {
                 Type::PropertyInstance(property_instance.materialize(db, materialization_kind))
             }
 
-            Type::FunctionLiteral(_) | Type::BoundMethod(_) => {
+            Type::FunctionLiteral(_) => {
                 // TODO: Subtyping between function / methods with a callable accounts for the
                 // signature (parameters and return type), so we might need to do something here
                 *self
+            }
+
+            Type::BoundMethod(bound_method) => {
+                Type::BoundMethod(bound_method.materialize(db, materialization_kind))
             }
 
             Type::NominalInstance(instance) => instance.materialize(db, materialization_kind),
@@ -2792,7 +2800,14 @@ impl<'db> Type<'db> {
             }
 
             Type::GenericAlias(alias) => {
-                Some(ClassType::from(*alias).class_member(db, name, policy))
+                let attr = Some(ClassType::from(*alias).class_member(db, name, policy));
+                println!("Get name {}", name);
+                match alias.specialization(db).materialization_kind(db) {
+                    None => attr,
+                    Some(materialization_kind) => {
+                        attr.map(|attr| attr.materialize(db, materialization_kind))
+                    }
+                }
             }
 
             Type::SubclassOf(subclass_of_ty) => {
@@ -8915,6 +8930,18 @@ impl<'db> BoundMethodType<'db> {
                     .self_instance(db)
                     .is_equivalent_to_impl(db, self.self_instance(db), visitor)
             })
+    }
+
+    pub(crate) fn materialize(
+        self,
+        db: &'db dyn Db,
+        materialization_kind: MaterializationKind,
+    ) -> Self {
+        Self::new(
+            db,
+            self.function(db).materialize(db, materialization_kind),
+            self.self_instance(db).materialize(db, materialization_kind),
+        )
     }
 }
 
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -768,14 +768,22 @@ impl<'db> Type<'db> {
     /// most general form of the type that is fully static.
     #[must_use]
     pub(crate) fn top_materialization(&self, db: &'db dyn Db) -> Type<'db> {
-        self.materialize(db, MaterializationKind::Top)
+        self.materialize(
+            db,
+            MaterializationKind::Top,
+            &ApplyTypeMappingVisitor::default(),
+        )
     }
 
     /// Returns the bottom materialization (or lower bound materialization) of this type, which is
     /// the most specific form of the type that is fully static.
     #[must_use]
     pub(crate) fn bottom_materialization(&self, db: &'db dyn Db) -> Type<'db> {
-        self.materialize(db, MaterializationKind::Bottom)
+        self.materialize(
+            db,
+            MaterializationKind::Bottom,
+            &ApplyTypeMappingVisitor::default(),
+        )
     }
 
     /// If this type is an instance type where the class has a tuple spec, returns the tuple spec.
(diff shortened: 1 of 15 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -820,100 +810,12 @@ impl<'db> Type<'db> {
...
-                        .materialize(db, materialization_kind),
-                )
-            }
-            Type::TypedDict(_) => {
-                // TODO: Materialization of gradual TypedDicts
-                *self
-            }
-            Type::TypeAlias(alias) => alias.value_type(db).materialize(db, materialization_kind),
-        }
+    pub(crate) fn materialize(
```
carljm, line 830: I see some `materialize` methods in this PR, and some `materialize_impl`, with the same signature. What determines whether you chose to rename a given method to `materialize_impl`?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -608,6 +610,9 @@ impl<'db> Specialization<'db> {
         type_mapping: &TypeMapping<'a, 'db>,
         visitor: &ApplyTypeMappingVisitor<'db>,
     ) -> Self {
+        if let TypeMapping::Materialize(materialization_kind) = type_mapping {
+            return self.materialize_impl(db, *materialization_kind);
```
carljm: So I am currently working on recursive type alias support, and as part of that I have a local branch where I'd already added a cycle-detection visitor to all materialize methods. I think this PR is an improvement, but it will probably mean I need all `materialize_impl` methods (at least any that can be called from an `apply_type_mapping_impl` method, such as in this spot) to accept and pass on an `ApplyTypeMappingVisitor`.

I think that should be fine, and I can handle it as a follow-up, just mentioning it in case you see any potential issues.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -723,7 +728,11 @@ impl<'db> Specialization<'db> {
             .collect();
         let tuple_inner = self.tuple_inner(db).and_then(|tuple| {
             // Tuples are immutable, so tuple element types are always in covariant position.
-            tuple.materialize(db, materialization_kind)
+            tuple.apply_type_mapping_impl(
+                db,
+                &TypeMapping::Materialize(materialization_kind),
+                &ApplyTypeMappingVisitor::default(),
```
carljm: Here I think I will need this `materialize_impl` to already accept the `ApplyTypeMappingVisitor`, and I'll need to pass it along here, and in all the variance arms above.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -469,13 +442,19 @@ impl<'db> Signature<'db> {
         type_mapping: &TypeMapping<'a, 'db>,
         visitor: &ApplyTypeMappingVisitor<'db>,
     ) -> Self {
+        let flipped_mapping = match type_mapping {
+            TypeMapping::Materialize(materialization_kind) => {
+                &TypeMapping::Materialize(materialization_kind.flip())
+            }
+            _ => type_mapping,
+        };
```
carljm, line 450: I've seen this a couple places in this PR -- should it be a method on `TypeMapping`?

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -1254,13 +1222,24 @@ impl<'db> Parameters<'db> {
...
+        match type_mapping {
+            TypeMapping::Materialize(MaterializationKind::Top) if self.is_gradual => {
+                Parameters::object(db)
+            }
+            // TODO: This is wrong, the empty Parameters is not a subtype of all materializations.
+            // The bottom materialization is not currently representable and implementing it
+            // properly requires extending the Parameters struct.
+            TypeMapping::Materialize(MaterializationKind::Bottom) if self.is_gradual => {
+                Parameters::empty()
+            }
```
carljm, line 1237: Aren't these reversed? I think the bottom and top materializations are counter-intuitive here due to contravariance of parameters. Isn't `Parameters::object(db)` the subtype of all materializations of the gradual parameter-signature (it is a valid substitute for any parameter-signature, because it can accept any call arguments)? And the top materialization (the super-type of all materializations, for which any parameter-signature is a valid substitute) is more difficult to represent. Or is it actually just `*args: Never, **kwargs: Never`, which accepts no calls, and which any other parameter signature is a subtype of?

carljm:
Oh, I just noticed the stack overflow in mypy-primer. So it may be that you actually do need to add some of the visitor-threading I mentioned above in this PR, to avoid regressing. (Though I'd recommend reproducing the stack overflow locally to verify the exact cyclic stack in lldb before trying to fix.)

crates/ty_python_semantic/src/types/signatures.rs
JelleZijlstra: I think that may not be right because we flip the variance earlier (in whatever struct contains Parameters).

> Or is it actually just `*args: Never, **kwargs: Never`, which accepts no calls, and which any other parameter signature is a subtype of?

That signature would accept a call with no arguments (`f()`), so a signature that has required arguments wouldn't be a subtype of it.

JelleZijlstra:
Yes, I'll need to look into the crash. Intuitively it's not clear why materialization would diverge, maybe it's for recursive aliases? If so would it make sense to wait for your more general fix?

carljm:
I think it most likely is recursive aliases (though it could also be recursive protocols). The fix isn't really a "general" one; part of it is ensuring that we propagate cycle detection visitors through all recursive type visits (a stack overflow is usually caused by failing to do this thoroughly). It seems like this PR is regressing our existing cycle detection coverage in some way; ideally that would be fixed in this PR.

crates/ty_python_semantic/src/types/signatures.rs
carljm: > we flip the variance earlier (in whatever struct contains Parameters)

Ah, that makes sense -- might be worth a comment. Or it might be worth changing this so we handle the variance flip in `Parameters`? To the extent we end up defining the "top" and "bottom" materializations of a `Parameters`, I think I would find it more intuitive for those to correspond with the actual top and bottom signature, not be reversed. (I know that in my previous comment I called it counter-intuitive 🤷 ). Not an important point.

> That signature would accept a call with no arguments (`f()`), so a signature that has required arguments wouldn't be a subtype of it.

Oh, yeah, of course. So we will need (not in this PR) a dedicated representation for "parameters which accept no call".

crates/ty_python_semantic/src/types.rs
JelleZijlstra: I think a good rule is going to be that materialize calls apply_type_mapping(_impl), and apply_type_mapping calls materialize_impl. I believe I do that mostly consistently already but will double check.

crates/ty_python_semantic/src/types/signatures.rs
JelleZijlstra: Yeah maybe, I was struggling a bit with getting things to compile. I can give it another try.

JelleZijlstra: I couldn't get this to work, maybe I need to learn more Rust or something. The pattern only appears twice, so doesn't seem too bad to leave it duplicated for now.

crates/ty_python_semantic/src/types/signatures.rs
JelleZijlstra: I'll add a comment.

crates/ty_python_semantic/src/types/generics.rs
JelleZijlstra: Done

crates/ty_python_semantic/src/types/generics.rs
JelleZijlstra: I made them all accept an ApplyTypeMappingVisitor, that fixes the hang seen in mypy-primer though I won't pretend to understand how it actually helps.

PR #20223, opened 2025-09-04
[ty] Skip overload filtering if there are no participating parameters

dhruvmanila:


dhruvmanila:
~I plan on looking at the performance regression tomorrow morning but otherwise I think this is ready to get some initial feedback.~

It seems like it was just noise.

dcreager (approved):
Looks good!

Issue #15584, opened 2025-01-19
Document fix safety


dylwil3:
The rules found at the following locations need to have a `## Fix safety` section added to their doc comment. If it makes sense to lump together a few in a single PR, feel free to do so, but in general we should probably keep them separate to speed up review and implementation.

I have omitted rules that have unsafe edits _only_ when they intersect comments, since that is a general policy.

Please let me know if a file should be added or removed from the list - I did a regex and manual pruning, but I may have missed something or included something I shouldn't have.

- [ ] crates/ruff_linter/src/rules/ruff/rules/quadratic_list_summation.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/invalid_formatter_suppression_comment.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/collection_literal_concatenation.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/implicit_optional.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/unnecessary_round.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/missing_fstring_syntax.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/zip_instead_of_pairwise.rs
- [ ] crates/ruff_linter/src/rules/ruff/rules/post_init_default.rs
- [ ] crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/useless_import_alias.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/modified_iterating_set.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/unnecessary_dunder_call.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/sys_exit_alias.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/nested_min_max.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/unspecified_encoding.rs
- [ ] crates/ruff_linter/src/rules/pylint/rules/repeated_equality_comparison.rs
- [ ] crates/ruff_linter/src/rules/pycodestyle/rules/lambda_assignment.rs
- [ ] crates/ruff_linter/src/rules/pycodestyle/rules/trailing_whitespace.rs
- [ ] crates/ruff_linter/src/rules/flake8_pytest_style/rules/assertion.rs
[...]

PR #20227, opened 2025-09-04
[`refurb`] Fix safety section (`FURB132`)

VascoSch92:
The PR add the fix safety section for rule `FURB132` (#15584 )

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/check_and_remove_from_set.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/check_and_remove_from_set.rs
@@ -37,6 +37,10 @@ use crate::fix::snippet::SourceCodeSnippet;
 /// nums.discard(123)
 /// ```
 ///
+/// ## Fix safety
+/// The fix is always marked as unsafe because it can change the error behavior of a program, as
+/// `set.remove()` raises `KeyError` when the item doesn't exist, while `set.discard()` silently does nothing.
+///
 /// ## References
 /// - [Python documentation: `set.discard()`](https://docs.python.org/3/library/stdtypes.html?highlight=list#frozenset.discard)
 #[derive(ViolationMetadata)]


crates/ruff_linter/src/rules/refurb/rules/check_and_remove_from_set.rs
```diff
@@ -37,6 +37,10 @@ use crate::fix::snippet::SourceCodeSnippet;
 /// nums.discard(123)
 /// ```
 ///
+/// ## Fix safety
+/// The fix is always marked as unsafe because it can change the error behavior of a program, as
+/// `set.remove()` raises `KeyError` when the item doesn't exist, while `set.discard()` silently does nothing.
+///
```
dylwil3, line 43: Hmm... but won't the `if` check make that impossible?

VascoSch92: Actually, you are right. I checked and I could not find a situation where we have a change of behaviour in the program. I think the only problem is that it can delete comments

dylwil3:
If you're up for it I think it would make sense to open a different PR where (gated under preview) you make the fix safe if the edit range does not intersect comments, and unsafe otherwise.

What happened here and elsewhere is that long long ago the fix was implemented as "suggested". Later all "suggested" fixes were switched to "sometimes applies", and later still all "sometimes applies" fixes were switched to "unsafe". There's quite a big semantic gap between "suggested" and "unsafe", and I think that's why a lot of the rules labeled as unsafe have left us scratching our heads.

VascoSch92:
> If you're up for it I think it would make sense to open a different PR where (gated under preview) you make the fix safe if the edit range does not intersect comments, and unsafe otherwise.
> 
> What happened here and elsewhere is that long long ago the fix was implemented as "suggested". Later all "suggested" fixes were switched to "sometimes applies", and later still all "sometimes applies" fixes were switched to "unsafe". There's quite a big semantic gap between "suggested" and "unsafe", and I think that's why a lot of the rules labeled as unsafe have left us scratching our heads.

make sense... I will give a look :-)

PR #20231, opened 2025-09-04
Stabilize `in-empty-collection` (`RUF060`)

ntBre:
Tests look good, just two very small comma changes to the docs


ntBre:
https://github.com/astral-sh/ruff/issues/20238

dylwil3:
I'd be okay with stabilizing once #20249 merges

ntBre:
I usually consider any "recent" bug report and fix to be a blocker, but this one does seem fairly minor :shrug: I'm happy either way but probably lean toward leaving it in preview, it's only 120 days old anyway.

PR #20232, opened 2025-09-04
Stabilize `invalid-rule-code` (`RUF102`)

ntBre:
Docs and tests look good


ntBre:
Hmm, this might be a problem. The ecosystem checks are flagging two TCH001 noqas. It seems like the rule should respect our own redirects. This works fine, for example, so I don't think these are actually invalid:

```shell
> cat try.py

import pandas as pd  # noqa: TCH002


def func(df: pd.DataFrame) -> int:
    return len(df)


> ruff check try.py --select TC002
All checks passed!
```

PR #20233, opened 2025-09-04
Stabilize `unused-unpacked-variable` (`RUF059`)

ntBre:
The tests looked good. For the docs, I added a `## See also` section pointing to
the closely-related F841 (unused-variable) and the corresponding section to F841
pointing back to RUF059. It seems like you'd probably want both of these active
or at least to know about the other when reading the docs.


ntBre:
I went through the ecosystem check subset linked in the PR, and these all look correct to me. There are a lot of hits, but the rule also has an autofix to add a `_` to the variable name, which seems very helpful.

PR #20234, opened 2025-09-04
[ty] Consider a TypedDict type to be a subtype of itself

sharkdp:
## Summary

I only wanted to add some `TypedDict` instances to the property tests for later when we implement proper structural assignability and subtyping, but this immediately uncovered a bug where a `TypedDict` type was not considered to be a subtype of itself. Unions and intersections were also not handled correctly, because the `has_relation_to` branch was too high up.


sharkdp:
Closing this, but we should probably add some `TypedDict`s (and `dict[…]` variants) to property tests? Have you potentially looked into that @oconnor663?

oconnor663:
No not yet.

PR #20236, opened 2025-09-04
Stabilize `redundant-none-literal` (`PYI061`)

ntBre:
Tests and docs look good


ntBre:
I can kind of see why people would prefer `Literal[1, 2, 3, None]` over `Literal[1, 2, 3] | None` or `Optional[Literal[1, 2, 3]]`, but these all look like true positives to me. If you prefer the other style, I guess you just wouldn't activate the rule. `Literal[None]` -> `None` seems less contentious, but it does appear in the ecosystem check, mostly for `@overload`s.

ntBre:
Just to record the decision here, we're aware of https://github.com/astral-sh/ruff/issues/20265, but I agree with Alex that it doesn't need to block stabilization.

user27182:
> I can kind of see why people would prefer `Literal[1, 2, 3, None]` over `Literal[1, 2, 3] | None` or `Optional[Literal[1, 2, 3]]`, but these all look like true positives to me.

This is more than a style preference, as it modifies the output of `typing.get_args`, see #20729. This auto-fix caused test failures in the `pyvista` repo when trying to update the `ruff` pre-commit hook:
https://github.com/pyvista/pyvista/pull/7992/commits/d908bd738be48e4e514a07d87237b038a4bdbc26

ntBre:
Ah that's good to know. Thank you for the report!

PR #20241, opened 2025-09-04
[ty] TypedDict: Add support for `typing.ReadOnly`

sharkdp:
## Summary

Add support for `typing.ReadOnly` as a type qualifier to mark `TypedDict` fields as being read-only. If you try to mutate them, you get a new diagnostic:

image

## Test Plan

* New Markdown tests
* The typing conformance changes are all correct. There are some false negatives, but those are related to the missing support for the functional form of `TypedDict`, or to overriding of fields via inheritance. Both of these topics are tracked in https://github.com/astral-sh/ty/issues/154

### DIFF
--- a/crates/ty_python_semantic/src/types/typed_dict.rs
+++ b/crates/ty_python_semantic/src/types/typed_dict.rs
@@ -121,6 +121,10 @@ impl TypedDictAssignmentKind {
             Self::Constructor => &INVALID_ARGUMENT_TYPE,
         }
     }
+
+    const fn is_subscript(self) -> bool {
+        matches!(self, Self::Subscript)
+    }
 }
 
 /// Validates assignment of a value to a specific key on a `TypedDict`.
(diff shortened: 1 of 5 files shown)


carljm (approved):
Awesome!

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -152,6 +156,29 @@ pub(super) fn validate_typed_dict_key_assignment<'db, 'ast>(
...
 
+    if assignment_kind.is_subscript() && item.is_read_only() {
+        if let Some(builder) =
+            context.report_lint(assignment_kind.diagnostic_type(), key_node.into())
+        {
+            let typed_dict_ty = Type::TypedDict(typed_dict);
+            let typed_dict_d = typed_dict_ty.display(db);
+
+            let mut diagnostic = builder.into_diagnostic(format_args!(
+                "Can not assign to key \"{key}\" on TypedDict `{typed_dict_d}`",
```
AlexWaygood, line 167: nit: this should be

```suggestion
                "Cannot assign to key \"{key}\" on TypedDict `{typed_dict_d}`",
```

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -152,6 +156,29 @@ pub(super) fn validate_typed_dict_key_assignment<'db, 'ast>(
...
+            ));
+
+            diagnostic.set_primary_message(format_args!("key is marked read-only"));
+
+            diagnostic.annotate(
+                context
+                    .secondary(typed_dict_node.into())
+                    .message(format_args!("TypedDict `{typed_dict_d}`")),
+            );
```
AlexWaygood, line 176: could possibly add a subdiagnostic pointing back to the original `ReadOnly` declaration on the `TypedDict` class? But not essential

sharkdp: https://github.com/astral-sh/ruff/pull/20262

PR #20243, opened 2025-09-04
[syntax-errors]: multiple-starred-expressions (F622)

11happy:
## Summary

This PR implements https://docs.astral.sh/ruff/rules/multiple-starred-expressions/ as a semantic syntax error

## Test Plan

 I have added inline tests as directed in #17412


11happy:
Hii @ntBre , would love your feedback on this when you get a chance. 
Thank you : )

ntBre:
Thanks for the pings. Your PRs are in my inbox, but we've been focused on the minor release. I'll get to them soon.

11happy:
Sure, No worries at all, I completely understand.  Thank you for your time : )

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1362,6 +1404,7 @@ pub enum SemanticSyntaxErrorKind {
 
     /// Represents the use of `yield from` inside an asynchronous function.
     YieldFromInAsyncFunction,
+    MultipleStarredExpressions,
```
ntBre, line 1458: Could we add some documentation for this? I think this one could justify a larger writeup (like in the F622 docs) than some of the simpler errors like `YieldFromInAsyncFunction` directly above.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -104,6 +104,22 @@ impl SemanticSyntaxChecker {
                         *range,
                     );
                 }
+                for target in targets {
```
ntBre: I don't think this check is quite right. There's a new ecosystem hit for a case like this:

```py
(*_, normed), *_ = [(1,), 2]
```

which should be allowed. You may be able to take inspiration from the actual F622 implementation (which we'll also need to delete before adding this second implementation, or we'll get duplicate diagnostics. That's the reason for the change in the F622.py snapshot above).

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1230,7 +1266,10 @@ pub enum SemanticSyntaxErrorKind {
     /// to be very rare and not worth the additional complexity to detect.
     ///
     /// [#111123]: https://github.com/python/cpython/issues/111123
-    LoadBeforeGlobalDeclaration { name: String, start: TextSize },
+    LoadBeforeGlobalDeclaration {
+        name: String,
+        start: TextSize,
+    },
```
ntBre: I'd prefer not to reformat any unrelated code, but I guess if the formatting CI job doesn't complain it's okay.

crates/ruff_python_parser/src/semantic_errors.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs

crates/ruff_python_parser/src/semantic_errors.rs
11happy: Sure

crates/ruff_linter/src/checkers/ast/analyze/expression.rs
```diff
@@ -209,13 +209,11 @@ pub(crate) fn expression(expr: &Expr, checker: &Checker) {
...
                 let check_too_many_expressions =
                     checker.is_rule_enabled(Rule::ExpressionsInStarAssignment);
-                let check_two_starred_expressions =
-                    checker.is_rule_enabled(Rule::MultipleStarredExpressions);
                 pyflakes::rules::starred_expressions(
                     checker,
                     elts,
                     check_too_many_expressions,
-                    check_two_starred_expressions,
+                    false,
```
ntBre: Is this argument still used for anything, or could we delete the argument and the code that checks it too?

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -358,6 +361,38 @@ impl SemanticSyntaxChecker {
...
+    fn multiple_star_expression<Ctx: SemanticSyntaxContext>(expr: &Expr, ctx: &Ctx) {
+        match expr {
+            Expr::Tuple(ast::ExprTuple { elts, range, .. })
+            | Expr::List(ast::ExprList { elts, range, .. }) => {
+                let mut count = 0;
+                for elt in elts {
+                    if matches!(elt, Expr::Starred(_)) {
+                        count += 1;
+                    }
+                    Self::multiple_star_expression(elt, ctx);
```
ntBre: I think instead of recursing here, we could perform this `multiple_star_expression` check in `check_expr` instead of `check_stmt`. The parent visitor should already recurse automatically into the child expressions. That seems closer to how F622 was implemented originally. Concretely, I'd suggest something like this, which also uses the structure of the original F622 implementation:

<details><summary>Patch</summary>

```diff
diff --git a/crates/ruff_python_parser/src/semantic_errors.rs b/crates/ruff_python_parser/src/semantic_errors.rs
index 1a8fb3ad6f..b4c0ffb947 100644
--- a/crates/ruff_python_parser/src/semantic_errors.rs
+++ b/crates/ruff_python_parser/src/semantic_errors.rs
@@ -104,9 +104,6 @@ impl SemanticSyntaxChecker {
                         *range,
                     );
                 }
-                for target in targets {
-                    Self::multiple_star_expression(target, ctx);
-                }
 
                 // test_ok assign_stmt_starred_expr_value
                 // _ = 4
@@ -361,38 +358,6 @@ impl SemanticSyntaxChecker {
             );
         }
     }
-    fn multiple_star_expression<Ctx: SemanticSyntaxContext>(expr: &Expr, ctx: &Ctx) {
-        match expr {
-            Expr::Tuple(ast::ExprTuple { elts, range, .. })
-            | Expr::List(ast::ExprList { elts, range, .. }) => {
-                let mut count = 0;
-                for elt in elts {
-                    if matches!(elt, Expr::Starred(_)) {
-                        count += 1;
-                    }
-                    Self::multiple_star_expression(elt, ctx);
-                }
-                if count > 1 {
-                    // test_err multiple_starred_assignment_target
-                    // (*a, *b) = (1, 2)
-                    // [*a, *b] = (1, 2)
-                    // (*a, *b, c) = (1, 2, 3)
-                    // [*a, *b, c] = (1, 2, 3)
-
-                    // test_ok multiple_starred_assignment_target
-                    // (*a, b) = (1, 2)
-                    // (*_, normed), *_ = [(1,), 2]
-
-                    Self::add_error(
-                        ctx,
-                        SemanticSyntaxErrorKind::MultipleStarredExpressions,
-                        *range,
-                    );
-                }
-            }
-            _ => {}
-        }
-    }
 
     /// Check for [`SemanticSyntaxErrorKind::WriteToDebug`] in `stmt`.
     fn debug_shadowing<Ctx: SemanticSyntaxContext>(stmt: &ast::Stmt, ctx: &Ctx) {
@@ -765,6 +730,45 @@ impl SemanticSyntaxChecker {
             }) => {
                 Self::duplicate_parameter_name(parameters, ctx);
             }
+            Expr::Tuple(ast::ExprTuple {
+                elts,
+                ctx: expr_ctx,
+                range,
+                node_index: _,
+                parenthesized: _,
+            })
+            | Expr::List(ast::ExprList {
+                elts,
+                ctx: expr_ctx,
+                range,
+                node_index: _,
+            }) => {
+                if expr_ctx.is_store() {
+                    let mut has_starred: bool = false;
+                    for elt in elts {
+                        if elt.is_starred_expr() {
+                            if has_starred {
+                                // test_err multiple_starred_assignment_target
+                                // (*a, *b) = (1, 2)
+                                // [*a, *b] = (1, 2)
+                                // (*a, *b, c) = (1, 2, 3)
+                                // [*a, *b, c] = (1, 2, 3)
+
+                                // test_ok multiple_starred_assignment_target
+                                // (*a, b) = (1, 2)
+                                // (*_, normed), *_ = [(1,), 2]
+                                Self::add_error(
+                                    ctx,
+                                    SemanticSyntaxErrorKind::MultipleStarredExpressions,
+                                    *range,
+                                );
+                                return;
+                            }
+                            has_starred = true;
+                        }
+                    }
+                }
+            }
             _ => {}
         }
     }
```

</details>

That's pretty deeply indented, so it might still be a good idea to factor out a function.

Along these lines, another useful test case might be something like this:

```py
(*a, *b, (*c, *d)) = (1, 2)
```

We currently emit 2 F622 diagnostics for this, including with your changes. That would help make sure we get the recursion in somewhere. At first I thought we could just delete the recursive `multiple_star_expression` call.

ntBre (commented):
Thanks! I think the tests look good now, just a couple of simplification suggestions.

crates/ruff_linter/src/checkers/ast/analyze/expression.rs
11happy: done, yes I checked it can be removed.

crates/ruff_python_parser/src/semantic_errors.rs
11happy: Thank you , I have updated it accordingly : )

11happy:
The `cargo test (wasm)` failure seems unrelated to this PR changes.

Thank you

11happy:
Please let me know if there's any more changes required for this PR.

Thank you : )

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -388,6 +388,39 @@ impl SemanticSyntaxChecker {
             );
         }
     }
+    fn multiple_star_expression<Ctx: SemanticSyntaxContext>(
```
ntBre, line 392: tiny nit:


```suggestion

    fn multiple_star_expression<Ctx: SemanticSyntaxContext>(
```

ntBre (approved):
Thank you! I just had one tiny spacing nit that I'll apply and merge.

PR #20244, opened 2025-09-04
Stabilize `long-sleep-not-forever` (`ASYNC116`)

ntBre:
Tests and docs look good


ntBre:
The ecosystem changes are previously-unused noqa codes for ASYNC116 that are now used on stable. I took a look at these to see where a real project needed to use a noqa for this rule, but it was just trio intentionally testing long sleeps, with comments to support that:

```py
    # ignore ASYNC116, not sleep_forever, trying to test a large but finite sleep
    await sleep(100000)  # noqa: ASYNC116
```

PR #20245, opened 2025-09-04
[`RUF102`] Respect rule redirects in invalid rule code detection

### LINKED ISSUE #20235: `invalid-rule-code` (`RUF102`) doesn't respect rule redirects
### Summary

This came up in the [ecosystem check](https://github.com/astral-sh/ruff/pull/20232#issuecomment-3253818000) when trying to stabilize the rule for the 0.13 release. Using the `TC002` example with an added `TCH002` noqa (`TCH` has been remapped to `TC`):

```py

import pandas as pd  # noqa: TCH002


def func(df: pd.DataFrame) -> int:
    return len(df)
```

Selecting both `TC002` and `RUF102` shows that `TC002` is suppressed (the noqa code works and is not invalid), but `RUF102` still fires:

```shell
> ruff check try.py --select TC002 --select RUF102 --preview
RUF102 [*] Invalid rule code in `# noqa`: TCH002
 --> try.py:3:22
  |
1 | from __future__ import annotations
2 |
3 | import pandas as pd  # noqa: TCH002
  |                      ^^^^^^^^^^^^^^
  |
help: Remove the rule code

Found 1 error.
```

TaKO8Ki:
## Summary

Fixes #20235

• Fix `RUF102` to properly handle rule redirects when validating noqa codes
• Update `code_is_valid` to check redirect targets before determining validity
• Add test case for rule redirects (TCH002 in this case)

## Test Plan

I have added a test case for rule redirects to `crates/ruff_linter/resources/test/fixtures/ruff/RUF102.py`.

### DIFF
--- a/crates/ruff_linter/src/rules/ruff/rules/invalid_rule_code.rs
+++ b/crates/ruff_linter/src/rules/ruff/rules/invalid_rule_code.rs
@@ -6,6 +6,7 @@ use crate::checkers::ast::LintContext;
 use crate::noqa::{Code, Directive};
 use crate::noqa::{Codes, NoqaDirectives};
 use crate::registry::Rule;
+use crate::rule_redirects::get_redirect_target;
 use crate::{AlwaysFixableViolation, Edit, Fix};
 
 /// ## What it does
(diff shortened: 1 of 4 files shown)


crates/ruff_linter/src/rules/ruff/rules/invalid_rule_code.rs
```diff
@@ -81,7 +82,8 @@ pub(crate) fn invalid_noqa_code(
 
 fn code_is_valid(code: &Code, external: &[String]) -> bool {
     let code_str = code.as_str();
-    Rule::from_code(code_str).is_ok() || external.iter().any(|ext| code_str.starts_with(ext))
+    Rule::from_code(get_redirect_target(code.as_str()).unwrap_or(code.as_str())).is_ok()
+        || external.iter().any(|ext| code_str.starts_with(ext))
```
amyreese: Should both `code.as_str()` usages be `code_str` instead?

TaKO8Ki: Thank you. You're right. I have fixed it.

amyreese:
Thank you!

PR #18611, opened 2025-06-10
[`flake8-pyi`] Avoid syntax error in the case of starred and keyword arguments (`PYI059`)

ntBre:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/18602 by:
1. Avoiding a fix when `*args` are present
2. Inserting the `Generic` base class right before the first keyword argument, if one is present

In an intermediate commit, I also had special handling to avoid a fix in the `**kwargs` case, but this is treated (roughly) as a normal keyword, and I believe handling it properly falls out of the other keyword fix.

## Test Plan

Existing PYI059 cases, plus new tests based on the issue

PR #18601, opened 2025-06-09
Stabilize `generic-not-last-base-class` (`PYI059`)

ntBre:
Summary
--

The tests were already in the right place, I just updated the documentation slightly to reflect the discussion [here](https://github.com/astral-sh/ruff/pull/18519#issuecomment-2956344254) and on Discord.

Documentation: https://docs.astral.sh/ruff/rules/generic-not-last-base-class

Tests: https://github.com/astral-sh/ruff/blob/brent/release-0.12.0/crates/ruff_linter/src/rules/flake8_pyi/mod.rs#L51-L52

I considered mentioning the MRO explicitly and linking to [this](https://docs.python.org/3/howto/mro.html) but decided that might be a bit too technical for the rule docs.

Now I see that ty links to the [glossary](https://docs.python.org/3/glossary.html#term-method-resolution-order), which would be another option.

Test Plan
--

Existing tests

PR #20246, opened 2025-09-04
Stabilize `generic-not-last-base-class` (`PYI059`)

ntBre:
Tests and docs look good

We nearly stabilized this last time (https://github.com/astral-sh/ruff/pull/18601), but it needed one more bug fix and a documentation improvement (https://github.com/astral-sh/ruff/pull/18611)


ntBre:
This ecosystem hit from typeshed seems interesting, but maybe okay? Everything else looks as expected to me.

https://github.com/python/typeshed/blob/07557a4316d246b4315f600fd4c9734297d6bc92/stdlib/asyncio/queues.pyi#L27-L29

AlexWaygood (approved):

PR #20247, opened 2025-09-04
Stabilize `f-string-number-format` (`FURB116`)

ntBre:
Tests and docs look good


amyreese (approved):

ntBre:
Ah really?? I guess neither of them is really that readable, but I thought the f-string was slightly better :laughing: We can hold off if you feel strongly, I'm happy either way.

AlexWaygood:
I don't have a strong opinion. I don't think either `bin()` or `oct()` is _commonly_ used, so consistency is probably the most important thing 😄

PR #20248, opened 2025-09-04
[ty] Implement the legacy PEP-484 convention for indicating positional-only parameters

AlexWaygood:
## Test Plan

Mdtests added and updated

---

Co-authored-by: Carl Meyer <carl@astral.sh>

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/definition.rs
+++ b/crates/ty_python_semantic/src/semantic_index/definition.rs
@@ -1134,6 +1134,12 @@ impl From<&ast::StmtClassDef> for DefinitionNodeKey {
     }
 }
 
+impl From<&AstNodeRef<ast::StmtClassDef>> for DefinitionNodeKey {
+    fn from(node_ref: &AstNodeRef<ast::StmtClassDef>) -> Self {
+        Self(NodeKey::from_node_ref(node_ref))
+    }
+}
+
 impl From<&ast::StmtTypeAlias> for DefinitionNodeKey {
     fn from(node: &ast::StmtTypeAlias) -> Self {
         Self(NodeKey::from_node(node))
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -236,6 +236,22 @@ impl<'db> OverloadLiteral<'db> {
         self.has_known_decorator(db, FunctionDecorators::OVERLOAD)
     }
 
+    /// Returns true if this overload is decorated with `@staticmethod`, or if it is implicitly a
+    /// staticmethod.
+    pub(crate) fn is_staticmethod(self, db: &dyn Db) -> bool {
+        self.has_known_decorator(db, FunctionDecorators::STATICMETHOD) || self.name(db) == "__new__"
+    }
+
+    /// Returns true if this overload is decorated with `@classmethod`, or if it is implicitly a
+    /// classmethod.
+    pub(crate) fn is_classmethod(self, db: &dyn Db) -> bool {
+        self.has_known_decorator(db, FunctionDecorators::CLASSMETHOD)
+            || matches!(
+                self.name(db).as_str(),
+                "__init_subclass__" | "__class_getitem__"
+            )
+    }
+
     fn node<'ast>(
         self,
         db: &dyn Db,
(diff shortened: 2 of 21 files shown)


AlexWaygood:
The conformance suite diff is all positive: we're emitting new diagnostics on lines where we're meant to emit diagnostics.

The [new hits on `hydra-zen`](https://github.com/mit-ll-responsible-ai/hydra-zen/blob/4a9997c6a9907d0c85a39a422d8f5760d7358b68/tests/annotations/declarations.py#L858-L866) are also expected -- they're actually trying to _assert_ on these lines that certain calls cause the type checker to emit diagnostics.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -1181,23 +1192,61 @@ impl<'db> Parameters<'db> {
...
+                //
+                // TODO: should it also always be implicitly positional-only if the method definition
+                // is in a `Protocol` class? --Alex
+                positional_only.extend(pos_or_keyword_iter.next().map(pos_only_param));
+                positional_only.extend(
+                    pos_or_keyword_iter
+                        .peeking_take_while(|param| uses_pep_484_positional_only_convention(param))
+                        .map(pos_only_param),
+                );
+            }
```
AlexWaygood: I initially tried to see what would happen if we made `self` or `cls` _always_ implicitly positional-only, even if there were no other parameters in the function that used the PEP-484 convention. But there were a surprising number of ecosystem hits where people really are passing `self` or `cls` using keyword arguments.

The reason why this might be "interesting" for protocols is that [`Iterator.__iter__`](https://github.com/python/typeshed/blob/07557a4316d246b4315f600fd4c9734297d6bc92/stdlib/typing.pyi#L528) (for example) is not annotated as having a positional-only `self` parameter in typeshed. A naive implementation of protocol assignability/subtyping for method members might therefore lead us to conclude that this class isn't a subtype of `Iterable` (the protocol declares that it must be possible to pass `self` by keyword argument, and you can't do that with `Foo.__iter__`!), which is almost certainly going to be a counter-intuitive result for the user and go against their intent:

```py
from typing import Iterator

class Foo:
    def __iter__(self, /) -> Iterator[int]: ...
```

I suppose one alternative here might be to treat `self` as always implicitly positional-only for dunder methods? That doesn't feel like it would lead to particularly consistent or principled behaviour though.

I also think it's not essential to find a solution to this now, though. It's sort of a rare edge case, and anyway I'm pretty sure I can work around it fairly easily in our `Protocol` implementation

crates/ty_python_semantic/src/semantic_index/definition.rs
```diff
@@ -1134,6 +1134,12 @@ impl From<&ast::StmtClassDef> for DefinitionNodeKey {
     }
 }
 
+impl From<&AstNodeRef<ast::StmtClassDef>> for DefinitionNodeKey {
```
carljm: We could fill this out and add similar `From` implementations for the `AstNodeRef` wrapper of every AST node kind that we currently implement `From` for a direct reference to. I intended in our pairing session today that we'd come back and do this; it might save someone some confusion in future if they try to use a different kind of `AstNodeRef` as an `Into<DefinitionNodeKey>`. Seems sort of odd to arbitrarily support it only for class definitions, even if that's the only place we currently use it.

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -713,17 +732,15 @@ impl<'db> FunctionType<'db> {
     /// Returns true if this method is decorated with `@classmethod`, or if it is implicitly a
     /// classmethod.
     pub(crate) fn is_classmethod(self, db: &'db dyn Db) -> bool {
-        self.has_known_decorator(db, FunctionDecorators::CLASSMETHOD)
-            || matches!(
-                self.name(db).as_str(),
-                "__init_subclass__" | "__class_getitem__"
-            )
+        self.iter_overloads_and_implementation(db)
```
carljm, line 796: Nice! Glad to see we don't have to duplicate the source of truth for special-cased method names.

crates/ty_python_semantic/src/types/signatures.rs
carljm: It makes sense to me to imply this only for `Protocol` methods (as suggested in your `TODO` comment above.)

crates/ty_python_semantic/src/semantic_index/definition.rs
AlexWaygood: I found a way to generalise this impl (🥳) using higher-ranked trait bounds (😱) in https://github.com/astral-sh/ruff/pull/20248/commits/487334cb1498aa5a48065d00543a16b8b6d9d10c

crates/ty_python_semantic/src/types/signatures.rs
AlexWaygood: done in https://github.com/astral-sh/ruff/pull/20248/commits/3e1153a1a174fd4fdd9b79f865153be71394b994

AlexWaygood:
Requesting a re-review since this has changed a fair bit since you last took a look!

PR #20250, opened 2025-09-04
Stabilize the remaining Airflow rules

### LINKED ISSUE #17749: Stabilize Airflow 3.0 migration rules

ntBre:
- **Stabilize `airflow3-suggested-update` (`AIR311`)**
- **Stabilize `airflow3-suggested-to-move-to-provider` (`AIR312`)**
- **Stabilize `airflow3-removal` (`AIR301`)**
- **Stabilize `airflow3-moved-to-provider` (`AIR302`)**
- **Stabilize `airflow-dag-no-schedule-argument` (`AIR002`)**

I put this all in one PR to make it easier to double check with @Lee-W before we merge this. I also made a few minor documentation changes and updated one error message that I want to make sure are okay. But for the most part this just moves the rules from `RuleGroup::Preview` to `RuleGroup::Stable`!

Fixes #17749

### DIFF
--- a/crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
@@ -17,9 +17,9 @@ use ruff_text_size::TextRange;
 /// ## Why is this bad?
 /// Airflow 3.0 removed various deprecated functions, members, and other
 /// values. Some have more modern replacements. Others are considered too niche
-/// and not worth to be maintained in Airflow.
+/// and not worth continued maintenance in Airflow.
 /// Even though these symbols still work fine on Airflow 3.0, they are expected to be removed in a future version.
-/// The user is suggested to replace the original usage with the new ones.
+/// Where available, users should replace the removed functionality with the new alternatives.
 ///
 /// ## Example
 /// ```python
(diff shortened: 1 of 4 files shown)


crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
```diff
@@ -58,7 +58,7 @@ impl Violation for Airflow3SuggestedUpdate {
             | Replacement::AutoImport { module: _, name: _ }
             | Replacement::SourceModuleMoved { module: _, name: _ } => {
                 format!(
-                    "`{deprecated}` is removed in Airflow 3.0; \
+                    "`{deprecated}` is deprecated in Airflow 3.0; \
```
ntBre: This is the largest change I made. The second sentence "It still works..." made it sound like _removed_ wasn't exactly the right word. I'm happy to revert this if preferred, though.

Lee-W: It's technically been removed. 👀 We added an alias to the new place instead.

something like https://github.com/apache/airflow/blob/b9937fbbb3243b5ea25b7091bb0394cb8c238761/airflow-core/src/airflow/datasets/__init__.py#L32-L61

I like to `remove` a bit more. But I'm good with `deprecated` if it still makes more sense to you. 🙂

Lee-W:
also kinda wonder if the description here will also be updated. Thanks!

image

crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
ntBre: Ahh okay, I misunderstood. I'll put this back to `removed`. Thank you!

ntBre:
The documentation on the website is all generated from the docstrings and `message` implementations, so that should update to reflect whatever we change here after the release. Hmm, but it doesn't look like the AIR002 entry in that table updated after the patch release today. I'll look into that too.

Lee-W:
I think AIR302, AIR312 should probably be updated as well 🤔 let me know if there's anything I can help

ntBre:
The docs website should now be up to date! There was just a temporary issue in the release deployment yesterday.

PR #20251, opened 2025-09-04
Add support for sorting diagnostics without a range

amyreese:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

----

- Update ruff ordering compare to work with optional ranges (treating
  them as starting from zero)
- Add a simple from impl to convert std::Path to Span with empty range

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -1,7 +1,7 @@
 use std::{fmt::Formatter, path::Path, sync::Arc};
 
 use ruff_diagnostics::{Applicability, Fix};
-use ruff_source_file::{LineColumn, SourceCode, SourceFile};
+use ruff_source_file::{LineColumn, SourceCode, SourceFile, SourceFileBuilder};
 
 use ruff_annotate_snippets::Level as AnnotateLevel;
 use ruff_text_size::{Ranged, TextRange, TextSize};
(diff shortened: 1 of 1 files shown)


crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1211,6 +1214,14 @@ impl From<SourceFile> for Span {
...
 }
 
+impl From<&Path> for Span {
+    fn from(path: &Path) -> Self {
+        Span::from(
+            SourceFileBuilder::new(path.to_str().unwrap_or_default(), "".to_owned()).finish(),
+        )
+    }
+}
+
```
ntBre: Is this used in this PR, and I'm overlooking it? If not, I'd kind of rather group this with the code change that uses it since we're going to land this commit separately.

I'm also not sure this really belongs in `ruff_db`, which is generally agnostic to the `UnifiedFile` variant, while this is specific to a Ruff `SourceFile`. We can probably just inline this code, like we do for IO errors:

https://github.com/astral-sh/ruff/blob/df142def7a2796c9b48e1fa2314d27c4f3a94076/crates/ruff/src/diagnostics.rs#L67-L70

I think we can reuse the `Path::to_string_lossy` helper and avoid the `to_owned` call too.

But this looks like the right approach, we don't have access to any other `SourceFile` here 👍

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -504,9 +504,12 @@ impl Diagnostic {
     /// Panics if either diagnostic has no primary span, if the span has no range, or if its file is
     /// not a `SourceFile`.
     pub fn ruff_start_ordering(&self, other: &Self) -> std::cmp::Ordering {
-        (self.expect_ruff_source_file(), self.expect_range().start()).cmp(&(
+        fn start_or_zero(range: Option<TextRange>) -> TextSize {
+            range.map(|r| r.start()).unwrap_or_default()
+        }
```
ntBre: I think I'd probably just inline `self.range().unwrap_or_default().start()` (and `other...`) and not factor out an internal helper function, but either way is fine.

We should also update the docstring now that this doesn't panic if the span has no range.

More broadly, I'm wondering how helpful this change will be on its own. I checked the references for `Diagnostic::expect_range`, and there are quite a few more than I expected (sorry!). Do you think we should try to remove all of those at once? Otherwise it seems like we could potentially run into problems  beyond this sort call.

Most of the uses are in tests, so it's safe to slap an `unwrap()` on those, but a few of them will need more consideration.

PR #20252, opened 2025-09-04
Better handling for panics during linting

amyreese:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #20252
* #20251

----

- Convert panics to diagnostics with id `Panic`, severity `Fatal`, and
  the error as the diagnostic message, annotated with a `Span` with
  empty code block and no range.
- Updates the post-linting message diagnostic handling to track the
  maximum severity seen, and then prints the "report a bug in ruff"
  message only if the max severity was `Fatal`

This depends on the sorting changes since it creates diagnostics with no
range specified.

### DIFF
--- a/crates/ruff/src/commands/check.rs
+++ b/crates/ruff/src/commands/check.rs
@@ -11,7 +11,7 @@ use log::{debug, warn};
 use rayon::prelude::*;
 use rustc_hash::FxHashMap;
 
-use ruff_db::diagnostic::{Annotation, Diagnostic, DiagnosticId};
+use ruff_db::diagnostic::{Annotation, Diagnostic, DiagnosticId, Span};
 use ruff_db::panic::catch_unwind;
 use ruff_linter::package::PackageRoot;
 use ruff_linter::registry::Rule;
@@ -204,7 +204,8 @@ fn lint_path(
                 ruff_db::diagnostic::Severity::Fatal,
                 message,
             );
-            diagnostic.annotate(Annotation::primary(path.into()));
+            let span = Span::from(SourceFileBuilder::new(path.to_string_lossy(), "").finish());
+            diagnostic.annotate(Annotation::primary(span));
             Ok(Diagnostics::new(vec![diagnostic], FxHashMap::default()))
         }
     }
--- a/crates/ruff/src/lib.rs
+++ b/crates/ruff/src/lib.rs
@@ -13,7 +13,7 @@ use log::{error, warn};
 use notify::{RecursiveMode, Watcher, recommended_watcher};
 
 use args::{GlobalConfigArgs, ServerCommand};
-use ruff_db::diagnostic::Severity;
+use ruff_db::diagnostic::{Diagnostic, Severity};
 use ruff_linter::logging::{LogLevel, set_up_logging};
 use ruff_linter::settings::flags::FixMode;
 use ruff_linter::settings::types::OutputFormat;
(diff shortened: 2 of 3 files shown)


crates/ruff/src/lib.rs
```diff
@@ -444,6 +445,19 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
         }
 
         if !cli.exit_zero {
+            let mut max_severity = Severity::Info;
+            for diagnostic in &diagnostics.inner {
+                max_severity = max_severity.max(diagnostic.severity());
+            }
```
ntBre: I think we could use something like this from ty here:

https://github.com/astral-sh/ruff/blob/df142def7a2796c9b48e1fa2314d27c4f3a94076/crates/ty/src/lib.rs#L340-L344

I guess it's actually longer than the loop, but iterators are fun 😆

crates/ruff/src/lib.rs
```diff
@@ -444,6 +445,19 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
...
+            if max_severity == Severity::Fatal {
```
ntBre: Another idea from ty:

https://github.com/astral-sh/ruff/blob/df142def7a2796c9b48e1fa2314d27c4f3a94076/crates/ty/src/lib.rs#L365

crates/ruff/src/lib.rs
```diff
@@ -444,6 +445,19 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
...
+                let message = "Panic during linting indicates a bug in Ruff. If you could open an issue at:
+
+https://github.com/astral-sh/ruff/issues/new?title=%5BLinter%20panic%5D
+
+...with the relevant file contents, the `pyproject.toml` settings, and the stack trace above, we'd be very appreciative!
+";
+                error!("{message}");
+            }
```
ntBre, line 463: It doesn't look like this guarantees that we exit non-zero. For example, if `cli.diff` or `fix_only`, we only inspect `diagnostics.fixed.is_empty()` rather than `diagnostics.inner.is_empty()`. I think a panic should still cause `--diff` or `--fix-only` to exit non-zero. Maybe I'm missing something here, though.

crates/ruff/src/commands/check.rs
```diff
@@ -193,21 +193,19 @@ fn lint_path(
...
                 ":".bold()
             );
-
-            Ok(Diagnostics::default())
+            let mut diagnostic = Diagnostic::new(
+                DiagnosticId::Panic,
+                ruff_db::diagnostic::Severity::Fatal,
+                message,
+            );
+            diagnostic.annotate(Annotation::primary(path.into()));
```
ntBre: I wonder if we should follow ty again here, or maybe even factor out a common panic `Diagnostic` constructor:

https://github.com/astral-sh/ruff/blob/df142def7a2796c9b48e1fa2314d27c4f3a94076/crates/ty_project/src/lib.rs#L687-L697

Some of the code for constructing the main `message` might even be helpful for us too.

amyreese:
> Thanks! I had a few suggestions, mostly based on the related code in ty. I'd also be interested in adding a panicking test rule so that we can see this in action. 

I've got a commit in progress to do that, but I need to figure out how to get that panic to only happen on the right inputs or whatever, because otherwise it breaks/ruins all the existing text rule fixtures/snapshots. But also the current test rule doesn't go through the same code path, so it ends up creating a `ruff crashed` panic instead.

crates/ruff/src/commands/check.rs
amyreese: I think the benefit of putting this outside the diagnostic is that we get the clear error message at the bottom after the diagnostic messages, cluing the user into the existence of a panic.

crates/ruff/src/lib.rs
amyreese: Apparently in my various reworks/rebases, I forgot to include the exit code change. Oops.

crates/ruff/src/commands/check.rs
ntBre: Oh right, I was still thinking that we'd sort by ascending severity, like in `RenderingSortKey`, so that the panics would be at the end, but that makes sense if we don't want to sort.

ntBre: Oh that's not the main comparison in `RenderingSortKey` anyway, hmm. It looks like ty has a relatively short logging/tracing message but still most of this in the diagnostic:

https://github.com/astral-sh/ruff/blob/df142def7a2796c9b48e1fa2314d27c4f3a94076/crates/ty/src/lib.rs#L365-L369

Issue #16690, opened 2025-03-12
Consider adding a fix for `pytest-raises-ambiguous-pattern` (`RUF043`)


ntBre:
It's certainly a more noisy rule and it's unfortunate we can't provide an auto fix (because Ruff can't know what the desired behavior is). But I think it's fine to stabilize

Unless we want to add a heuristic for the auto fix. E.g.:

* A string ending with `.` is probably a sentence -> escape it
* A string with a sequence of metacharacters (e.g. `.*` or `.+`) is probably a regex -> make it raw. We can limit it to only support some specific patterns.

The first fix would have to be unsafe because it does change the program's semantics. The second one could be safe but I think it's still best to make user explicitly opt-in. We could document the heuristics in the rule and having a fix could reduce the churn for users. 

I'd postpone the promotion to stable if the suggested fix does sound reasonable (CC: @AlexWaygood)

_Originally posted by @MichaReiser in https://github.com/astral-sh/ruff/pull/16657#pullrequestreview-2677569437_

AlexWaygood:
We discussed when implementing the rule originally whether it should have an autofix or not and concluded "no" because adding backslash escapes may often be unwelcome (users will often prefer to make it a raw string), but it would probably be hard for Ruff to determine whether a raw string or an escape would be preferred. https://github.com/astral-sh/ruff/pull/14966#issuecomment-2543194280

I also don't _love_ the idea of adding so many heuristics to an autofix; I think it'll be harder for users to understand without reading the docs (and I doubt people go and read the docs until a rule goes and does something that they don't expect).

So I'd be inclined to leave the rule as it is and stabilise it. That said, I don't have a strong opinion; happy to defer to you and @MichaReiser if you think users would find an unsafe fix helpful here!

DimitriPapadopoulos:
This rule mixes the concepts of [regex metacharacters](https://docs.python.org/3/howto/regex.html#matching-characters) (`.` `^` `$` `*` `+` `?` `{` `}` `[` `]` `\` `|` `(` `)`) and [escape sequences](https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences) (`\`). I find this confusing.

In the documentation, example `match="A full sentence."` should be changed to `("A full sentence\\.")` in most cases, shouldn't it? I don't see how switching to a raw string helps in any way.

```
>>> "A full sentence."
'A full sentence.'
>>> 
>>> r"A full sentence."
'A full sentence.'
>>> 
[...]

PR #16657, opened 2025-03-12
[`ruff`] Stabilize `pytest-raises-ambiguous-pattern` (`RUF043`)

ntBre:
Summary
--

Stabilizes RUF043 and moves its test to the right place. The docs look good.

Test Plan
--

0 closed or open issues. 1 follow-up PR to fix typos in the test documentation and 1 bug fix to treat `)` as a regex meta-character on 2025-01-07.

This can be a pretty noisy rule (+503 violations in the initial PR), so we'll see how the ecosystem check looks now. I'm happy to hold off stabilizing if we think it's too noisy, just wanted to open the PR just in case.

PR #20253, opened 2025-09-04
Stabilize `pytest-raises-ambiguous-pattern` (`RUF043`)

ntBre:
This one has been a bit contentious in the past. It usually uncovers ~700 ecosystem hits. See:

- https://github.com/astral-sh/ruff/pull/16657
- https://github.com/astral-sh/ruff/issues/16690

But I think there's consensus that it's okay to merge as-is. We'd love an
autofix since it's so common, but we can't reliably tell what a user meant. The
pattern is ambiguous after all :laughing:

This is the first rule that actually needed its test case relocated, but the
docs looked good.


amyreese (approved):

ntBre:
After going through the ecosystem checks, I realized we should probably expand the docs slightly to explain why each `Use instead` example suppresses the lint. This was also discussed a bit on #16690.

All of the ecosystem hits look like true positives, though.

PR #20254, opened 2025-09-04
Stabilize `single-item-membership-test` (`FURB171`)

ntBre:
This rule has been a stabilization candidate at least as far back as 0.6, but it
had a few issues that blocked stabilization each time. The last issue was before
the 0.12 release, so I think it's worth trying this one again.

The docs and tests looked good.


ntBre:
#20255

PR #20256, opened 2025-09-04
[ty] Narrow specialized generics using isinstance()

JelleZijlstra:
Closes astral-sh/ty#456. Part of astral-sh/ty#994.

After all the foundational work, this is only a small change, but let's see if it exposes any unresolved issues.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/exhaustiveness_checking.md
+++ b/crates/ty_python_semantic/resources/mdtest/exhaustiveness_checking.md
@@ -281,12 +281,10 @@ def if_else_exhaustive(x: A[D] | B[E] | C[F]):
     elif isinstance(x, C):
         pass
     else:
-        # TODO: both of these are false positives (https://github.com/astral-sh/ty/issues/456)
-        no_diagnostic_here  # error: [unresolved-reference]
-        assert_never(x)  # error: [type-assertion-failure]
+        no_diagnostic_here
+        assert_never(x)
 
-# TODO: false-positive diagnostic (https://github.com/astral-sh/ty/issues/456)
-def if_else_exhaustive_no_assertion(x: A[D] | B[E] | C[F]) -> int:  # error: [invalid-return-type]
+def if_else_exhaustive_no_assertion(x: A[D] | B[E] | C[F]) -> int:
     if isinstance(x, A):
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
mypy-primer analysis:

* aio-redis ([link](https://github.com/aio-libs-abandoned/aioredis-py/blob/19be499015a8cf32580e937cbfd711fd48489eca/aioredis/client.py#L3429)): new errors are technically correct. It's in part because ty accounts for the possibility that something could be a Sequence and a Mapping at the same time, which is a bit pedantic but technically possible.
* anyio ([link](https://github.com/agronholm/anyio/blob/master/src/anyio/_core/_sockets.py)): new errors are due to ty preserving precise tuple types better. The errors should go away once ty supports narrowing by `len()` calls.
* nionutils ([link](https://github.com/nion-software/nionutils/blob/762ef25d29759c47610d0aa8e70600cc5fc5d785/nion/utils/Stream.py#L335)): similar to aioredis, technically correct new errors that depends on ty accounting for two classes that may not be disjoint.
* Expression ([link](https://github.com/dbrattli/Expression/blob/3981fd3a3ad60a13d00a4a3c170bdac76c1943fe/tests/test_option.py#L28)): didn't fully track down the types involved, but looks like ty is solving some generics nonoptimally.
* alerta ([link](https://github.com/alerta/alerta/blob/b7111d23ae9aefb5064ef107976f62e3836217b6/alerta/utils/api.py#L122)): lots of alerts going away due to better exhaustiveness checking. New errors would go away if we had `len()` narrowing on tuples.
* artigraph: minor change in error message
* starlette ([link](https://github.com/Kludex/starlette/blob/1ee8f857048368b3603f32f91a1214390e79d614/starlette/staticfiles.py#L76)): false positive goes away because `isinstance(x, tuple)` now works better
* koda_validate ([link](https://github.com/keithasaurus/koda-validate/blob/9e6c8985b59dceda54b18de7ae79c1059be4d6fc/koda_validate/_internal.py#L212)): ty seems technically correct, maybe something should remain Any though?

That's all for now.

carljm (approved):
Looks great! I'm satisfied with the ecosystem analysis you already did (thank you!).

PR #20257, opened 2025-09-05
Add support for sorting diagnostics without a range

amyreese:
- Update ruff ordering compare to work with optional ranges (treating
  them as starting from zero)

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -501,13 +501,18 @@ impl Diagnostic {
 
     /// Returns the ordering of diagnostics based on the start of their ranges, if they have any.
     ///
-    /// Panics if either diagnostic has no primary span, if the span has no range, or if its file is
-    /// not a `SourceFile`.
+    /// Panics if either diagnostic has no primary span, or if its file is not a `SourceFile`.
     pub fn ruff_start_ordering(&self, other: &Self) -> std::cmp::Ordering {
-        (self.expect_ruff_source_file(), self.expect_range().start()).cmp(&(
+        let a = (
+            self.expect_ruff_source_file(),
+            self.range().map(|r| r.start()),
+        );
+        let b = (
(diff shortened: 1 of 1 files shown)


amyreese:
Updated docstring. @ntBre am I correct that there aren't any unit tests covering the ordering of diagnostics where I could add test cases? Is that worth taking the time to add?

ntBre:
It looks like `test_contents` calls `ruff_start_ordering`, so I'm pretty sure all of our snapshot tests actually exercise this behavior, at least the range part since they're mostly single-file tests:

https://github.com/astral-sh/ruff/blob/16d674e63264acdf640349c81ec6263e813a3234/crates/ruff_linter/src/test.rs#L405

I'm guessing some of our CLI tests probably have multiple files and thus use the file part too.

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -501,13 +501,19 @@ impl Diagnostic {
...
     ///
-    /// Panics if either diagnostic has no primary span, if the span has no range, or if its file is
-    /// not a `SourceFile`.
+    /// Diagnostics with range of None will be treated as starting at zero.
+    /// Panics if either diagnostic has no primary span, or if its file is not a `SourceFile`.
     pub fn ruff_start_ordering(&self, other: &Self) -> std::cmp::Ordering {
-        (self.expect_ruff_source_file(), self.expect_range().start()).cmp(&(
+        let a = (
+            self.expect_ruff_source_file(),
+            self.range().map(|r| r.start()).unwrap_or_default(),
```
ntBre: I think we could avoid the `map` calls if we just `unwrap` the range first:


```rust
            self.range().unwrap_or_default().start(),
```

Alternatively, we can probably just drop the `unwrap_or_default`. I think `None` will sort below `Some(...)`, just like unwrapping to a `start` of zero.

```rust
            self.range().map(|range| range.start()),
```

PR #20259, opened 2025-09-05
[ty] propagate visitors and constraints through has_relation_in_invariant_position

carljm:
## Summary

The sub-checks for assignability and subtyping of materializations performed in `has_relation_in_invariant_position` and `is_subtype_in_invariant_position` need to propagate the `HasRelationToVisitor`, or we can stack overflow.

A side effect of this change is that we also propagate the `ConstraintSet` through, rather than using `C::from_bool`, which I think may also become important for correctness in cases involving type variables (though it isn't testable yet, since we aren't yet actually creating constraints other than always-true and always-false.)

## Test Plan

Added mdtest (derived from code found in pydantic) which stack-overflowed before this PR.

With this change incorporated, pydantic now checks successfully on my draft PR for PEP 613 TypeAlias support.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -287,18 +287,18 @@ def _(x: C):
     reveal_type(x)  # revealed: () -> C | None
 ```
 
-### Recursive relation in invariant position
+### Subtyping of materializations of cyclic aliases
 
 ```py
-from typing import Any, Callable
+from ty_extensions import static_assert, is_subtype_of, Bottom, Top
 
 type JsonValue = None | JsonDict
 type JsonDict = dict[str, JsonValue]
-type JsonSchemaExtraCallable = Callable[[JsonDict], None] | Callable[[JsonDict, type[Any]], None]
 
-def _update_class_schema(json_schema: dict[str, Any], json_schema_extra: JsonSchemaExtraCallable) -> None:
-    if isinstance(json_schema_extra, dict):
-        json_schema.update(json_schema_extra)
+static_assert(is_subtype_of(Top[JsonDict], Top[JsonDict]))
+static_assert(is_subtype_of(Top[JsonDict], Bottom[JsonDict]))
+static_assert(is_subtype_of(Bottom[JsonDict], Bottom[JsonDict]))
+static_assert(is_subtype_of(Bottom[JsonDict], Top[JsonDict]))
 ```
 
 ### Union inside generic
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -287,6 +287,20 @@ def _(x: C):
     reveal_type(x)  # revealed: () -> C | None
 ```
 
+### Recursive relation in invariant position
+
+```py
+from typing import Any, Callable
+
+type JsonValue = None | JsonDict
+type JsonDict = dict[str, JsonValue]
+type JsonSchemaExtraCallable = Callable[[JsonDict], None] | Callable[[JsonDict, type[Any]], None]
+
+def _update_class_schema(json_schema: dict[str, Any], json_schema_extra: JsonSchemaExtraCallable) -> None:
+    if isinstance(json_schema_extra, dict):
+        json_schema.update(json_schema_extra)
+```
+
 ### Union inside generic
 
 #### With old-style union
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -451,6 +451,7 @@ fn is_subtype_in_invariant_position<'db, C: Constraints<'db>>(
     derived_materialization: MaterializationKind,
     base_type: &Type<'db>,
     base_materialization: MaterializationKind,
+    visitor: &HasRelationToVisitor<'db, C>,
 ) -> C {
     let derived_top = derived_type.top_materialization(db);
     let derived_bottom = derived_type.bottom_materialization(db);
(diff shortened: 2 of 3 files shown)


carljm:
@JelleZijlstra I can't request your review here, as a non-member of the GH org, but I wouldn't mind your eyes on this if you have a chance, since this is updating code you wrote.

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -287,6 +287,20 @@ def _(x: C):
...
+
+```py
+from typing import Any, Callable
+
+type JsonValue = None | JsonDict
+type JsonDict = dict[str, JsonValue]
+type JsonSchemaExtraCallable = Callable[[JsonDict], None] | Callable[[JsonDict, type[Any]], None]
+
+def _update_class_schema(json_schema: dict[str, Any], json_schema_extra: JsonSchemaExtraCallable) -> None:
+    if isinstance(json_schema_extra, dict):
```
JelleZijlstra: This test case is a bit weird, because the `JsonSchemaExtraCallable` alias only contains callables. If this test case fails without your fix, it's fine, but wonder if something important got lost.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -459,37 +460,52 @@ fn is_subtype_in_invariant_position<'db, C: Constraints<'db>>(
...
-                    || (base_top.is_subtype_of(db, derived_top)
-                        && derived_bottom.is_subtype_of(db, base_top))),
-        ),
+        (MaterializationKind::Bottom, MaterializationKind::Top) => (base_bottom
+            .has_relation_to_impl(db, derived_bottom, TypeRelation::Subtyping, visitor)
+            .and(db, || {
+                derived_bottom.has_relation_to_impl(db, base_top, TypeRelation::Subtyping, visitor)
+            }))
+        .or(db, || {
+            base_bottom
```
JelleZijlstra: This becomes harder and harder to read, which is unfortunate as the logic is a bit tricky. But oh well.

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
carljm: Yes -- this doesn't simplify because `dict` and `Callable` are not disjoint, so it remains an intersection of `Top[dict[Unknown, Unknown]] & JsonSchemaExtraCallable` (which unpacks into a union of two intersections, since `JsonSchemaExtraCallable` is a union of two callable types -- the issue doesn't repro if I simplify it to a single callable type). Calling the overloaded `dict.update` (calling overloads can trigger materialization) with that union-of-intersections type is necessary to trigger the issue.

Clearly a lot of the original relevant code "got lost" in minimization. What's left is both sufficient and seemingly necessary to trigger the issue, but doesn't make a lot of sense as a standalone snippet.

I will spend some more time tomorrow understanding the precise mechanism more fully so I can write some more thorough and low-level tests.

crates/ty_python_semantic/src/types/generics.rs
carljm: I just pushed an update that I think helps a lot by defining an `is_subtype_of` closure helper, to remove all the repeated `TypeRelation::Subtyping, visitor` noise.

carljm:
@dcreager One thing to note here is that with invariance we sometimes end up checking type equivalence or subtyping as a sub-part of checking outer assignability, and with this PR we are taking constraints generated in an inner subtype or equivalence check and propagating them up to an outer assignability check. I think this is OK because the constraints we collect always use top and bottom materializations and fully static types, so the constraints should always be valid/interchangeable across assignability and subtyping? And I assume for equivalence we will create constraints with the same upper and lower bound, so those constraints are also valid across type relations. But let me know if there are issues here I'm not seeing.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -451,45 +451,50 @@ fn is_subtype_in_invariant_position<'db, C: Constraints<'db>>(
...
+            (is_subtype_of(base_bottom, derived_bottom)
+                .and(db, || is_subtype_of(derived_bottom, base_top)))
+            .or(db, || {
+                is_subtype_of(base_bottom, derived_top)
+                    .and(db, || is_subtype_of(derived_top, base_top))
+                    .or(db, || {
+                        is_subtype_of(base_top, derived_top)
+                            .and(db, || is_subtype_of(derived_bottom, base_top))
+                    })
+            })
```
AlexWaygood: could this also be written as

```suggestion
            .or(db, || {
                is_subtype_of(base_bottom, derived_top)
                    .and(db, || is_subtype_of(derived_top, base_top))
            })
            .or(db, || {
                is_subtype_of(base_top, derived_top)
                    .and(db, || is_subtype_of(derived_bottom, base_top))
            })
```

?

dcreager:
> One thing to note here is that with invariance we sometimes end up checking type equivalence or subtyping as a sub-part of checking outer assignability, and with this PR we are taking constraints generated in an inner subtype or equivalence check and propagating them up to an outer assignability check. I think this is OK because the constraints we collect always use top and bottom materializations and fully static types, so the constraints should always be valid/interchangeable across assignability and subtyping? And I assume for equivalence we will create constraints with the same upper and lower bound, so those constraints are also valid across type relations. But let me know if there are issues here I'm not seeing.

I think that should be okay, because the conditions under which that inner equivalence holds _should_ also be part of the conditions under which the outer assignability holds. (And you're right that we shouldn't end up with incompatible constraints or execution cycles since all of the checks that happen down inside the constraint set machinery work on fully static / materialized types.)

If we find that :point_up: isn't correct, then you can call `C::from_bool(is_equivalent_to)` instead of `when_equivalent_to`. (Though at that point we might consider adding `ensure_{always,sometimes}_satisfied` helpers on the `Constraints` trait.)

crates/ty_python_semantic/src/types/generics.rs
dcreager: > This becomes harder and harder to read

I had tried to use operator overloading when I first wrote these, but unfortunately you can't overload the short-circuiting `&&` and `||` operators. These closures are the only way I know of to get the rhs into a thunk so that it's lazily evaluated.

(Hmm...maybe a macro that would rewrite the boolean syntax into the right `and` and `or` calls? An interesting idea but not something to focus on atm)

carljm: Short of a macro or operator overloading, I did consider whether in nested cases like this a prefix-form helper would improve readability. So it would change `a.or(db, || b)` to `or(db, a, || b)`. 

I do think the prefix form is somewhat better visually for nested conditions. It doesn't allow us to get rid of the `||` since we still want short circuiting. I decided that limitation made it not worth it.

dcreager: Oh that's a good idea!  I think that will work without changing the trait or adding extra methods if you spell it as `C::or(a, db, || b)`.  (The weird parameter order is admittedly not great)

JelleZijlstra: Thanks, this already looks much better with the private helper.

carljm: Yeah I think what I've got is good enough here for now, I probably won't pursue `C::or` as part of this PR; the fact that we still can't get rid of the `db` or the `||` makes it feel less appealing.

PR #20261, opened 2025-09-05
[ty] Cover full range of annotated assignments

sharkdp:
## Summary

An annotated assignment `name: annotation` without a right-hand side was previously not covered by the range returned from `DefinitionKind::full_range`, because we did expand the range to include the right-hand side (if there was one), but failed to include the annotation.

## Test Plan

Updated snapshot tests

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/definition.rs
+++ b/crates/ty_python_semantic/src/semantic_index/definition.rs
@@ -769,13 +769,14 @@ impl DefinitionKind<'_> {
                 target_range.cover(value_range)
             }
             DefinitionKind::AnnotatedAssignment(assign) => {
-                let target_range = assign.target.node(module).range();
+                let mut full_range = assign.target.node(module).range();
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/semantic_index/definition.rs
```diff
@@ -769,13 +769,14 @@ impl DefinitionKind<'_> {
                 target_range.cover(value_range)
             }
             DefinitionKind::AnnotatedAssignment(assign) => {
-                let target_range = assign.target.node(module).range();
+                let mut full_range = assign.target.node(module).range();
+                full_range = full_range.cover(assign.annotation.node(module).range());
```
sharkdp, line 773: Due to how the syntax is structured, we could avoid this step if there is a right hand side (because covering the target and the right-hand side will necessarily also include the annotation). But this method is not performance critical, and I find it easier to reason about in this form: just always include the annotation.

PR #20262, opened 2025-09-05
[ty] Add backreferences to TypedDict items in diagnostics

sharkdp:
## Summary

Add backreferences to the original item declaration in TypedDict diagnostics:

image

image

Thanks to @AlexWaygood for the suggestion.

## Test Plan

Updated snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -9,12 +9,10 @@ use super::{
 use crate::FxOrderMap;
 use crate::module_resolver::KnownModule;
 use crate::semantic_index::definition::{Definition, DefinitionState};
-use crate::semantic_index::place::ScopedPlaceId;
 use crate::semantic_index::scope::NodeWithScopeKind;
 use crate::semantic_index::symbol::Symbol;
 use crate::semantic_index::{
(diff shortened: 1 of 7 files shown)


crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1283,6 +1283,9 @@ pub(crate) struct Field<'db> {
     pub(crate) declared_ty: Type<'db>,
     /// Kind-specific metadata for this field
     pub(crate) kind: FieldKind<'db>,
+    /// The original declaration of this field, if there is exactly one.
+    /// This field is used for backreferences in diagnostics.
+    pub(crate) single_declaration: Option<Definition<'db>>,
```
AlexWaygood, line 1286: Oh, nice. It might be possible to simplify this `NamedTuple` subdiagnostic using this new field:

https://github.com/astral-sh/ruff/blob/9e45bfa9fd730dc6a95bd94ced51147cc507cfe6/crates/ty_python_semantic/src/types/diagnostic.rs#L2914-L2938

sharkdp: Indeed! Would you mind taking a short look?

PR #20263, opened 2025-09-05
[ty] Include `python` folder in `environment.root` if it exists

sharkdp:
## Summary

I felt it was safer to add the `python` folder *in addition* to a possibly-existing `src` folder, even though the `src` folder only contains Rust code for `maturin`-based projects. There might be non-maturin projects where a `python` folder exists for other reasons, next to a normal `src` layout.

closes https://github.com/astral-sh/ty/issues/1120

## Test Plan

Tested locally on the egglog-python project.

### DIFF
--- a/crates/ty_project/src/metadata/options.rs
+++ b/crates/ty_project/src/metadata/options.rs
@@ -259,6 +259,19 @@ impl Options {
                 vec![project_root.to_path_buf()]
             };
 
+            let python = project_root.join("python");
+            if system.is_directory(&python) {
+                // If a `./python` directory exists, include it as a source root. This is the recommended layout
+                // maturin-based rust/python projects [1].
+                //
+                // https://github.com/PyO3/maturin/blob/979fe1db42bb9e58bc150fa6fc45360b377288bf/README.md?plain=1#L88-L99
+                tracing::debug!(
+                    "Including `./python` in `environment.root` because a `./python` directory exists"
+                );
+
+                roots.push(python);
+            }
+
             // Considering pytest test discovery conventions,
             // we also include the `tests` directory if it exists and is not a package.
             let tests_dir = project_root.join("tests");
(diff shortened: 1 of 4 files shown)


crates/ty_project/src/metadata/options.rs
```diff
@@ -259,6 +259,19 @@ impl Options {
                 vec![project_root.to_path_buf()]
             };
 
+            let python = project_root.join("python");
+            if system.is_directory(&python) {
```
AlexWaygood: I think we should do a similar thing to what we do with the `tests/` directory immediately below:
- Only add it to `roots` if it wasn't already explicitly specified by the user (otherwise we might end up with duplicate search paths, I guess?)
- Only add it to `roots` if `./python/__init__.py` does _not_ exist (if it _does_ exist, it's fairly clear that it's meant to be a top-level package rather than a first-party search path)

The existing `tests/` logic should probably also check for `./tests/__init__.pyi` as well as `./tests/__init__.py`. You could possibly also fix that while we're here, and also check for ./python/__init__.pyi` as well as `./python/__init__.py`.

crates/ty_project/src/metadata/options.rs
```diff
@@ -259,6 +259,19 @@ impl Options {
...
+                // If a `./python` directory exists, include it as a source root. This is the recommended layout
+                // maturin-based rust/python projects [1].
```
AlexWaygood:
```suggestion
                // for maturin-based rust/python projects [1].
```

AlexWaygood:
After landing this, we should also update the docs at https://github.com/astral-sh/ty/blob/main/docs/modules.md#first-party-modules. The docs at https://github.com/astral-sh/ty/blob/main/docs/reference/configuration.md#root don't need to be manually updated; they're generated from the doc-comment on the `Options` struct.

AlexWaygood (approved):
Thanks!

PR #20264, opened 2025-09-05
[flake8-async] Implement `blocking-path-method` (ASYNC240).

PieterCK:
## Summary
Adds a new rule to find and report use of `os.path` or `pathlib.Path` in
async functions.

Issue: #8451

## Test Plan

Using `cargo insta test`


ntBre:
Thanks! We're pretty focused on the minor release this week, so I may not get a chance to review until next week. But this looks really good from a quick skim :)

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,136 @@
...
+        "Async functions should not use pathlib.Path or os.path methods, use trio.Path or anyio.path".to_string()
+    }
+}
+
+const OS_PATH: [&str; 2] = ["os", "path"];
+const PATHLIB_PATH: [&str; 2] = ["pathlib", "Path"];
+
+fn is_blocking_path_method(segments: &[&str]) -> bool {
+    segments.starts_with(&OS_PATH) || segments.starts_with(&PATHLIB_PATH)
+}
```
amyreese: I'm not sure what the upstream rule does, but IMO this should also have either an allow list or an exclusion list, because there are a number of methods on Path objects that aren't blocking — at the very least, we should allow use [PurePath methods](https://docs.python.org/3/library/pathlib.html#methods-and-properties) or functions like `os.path.join` that don't care about the actual filesystem.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,136 @@
...
+        semantic: &ruff_python_semantic::SemanticModel,
+    ) -> bool {
+        let Expr::Call(ExprCall { func, .. }) = initializer else {
+            return false;
+        };
+
+        semantic
+            .resolve_qualified_name(func)
+            .is_some_and(|qualified_name| is_blocking_path_method(qualified_name.segments()))
+    }
```
amyreese, line 188: IIRC, the creation of a Path object itself isn't blocking, so this entire function can probably just be `false`;

amyreese: Although on second thought that might not catch calls like `os.path.exists(thing)`, so maybe instead make sure that eg, `pathlib.Path(...)` is allowed while `pathlib.Path(...).exists()` isn't.

ntBre: I think this is only used within the `check_type` infrastructure to check that the type of a variable (`p` in this case) is a `pathlib.Path` in a method call like:

```py
p = Path(".")
p.exists()
```

There's no diagnostic on line 49 of the test file, for example. So I think this is implemented correctly.

ntBre: Oh, I see what you mean. Yeah, I don't think we need to call `is_blocking_path_method` inside this method. We just need to check that it's a `Path`, which I think we might already have a `TypeChecker` for:

https://github.com/astral-sh/ruff/blob/16d674e63264acdf640349c81ec6263e813a3234/crates/ruff_python_semantic/src/analyze/typing.rs#L914-L917

This checks some sub/supeclasses too, we just need to double-check if we want that.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: The rule only mentions `os.path`, but yeah this makes sense. (I haven't looked into how they actually implemented it, though)

PieterCK: I've added an allow list for this (see `maybe_calling_io_operation`). It's probably not complete, but safer than an incomplete exclusion list of I/O methods.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: Ok! Switched to using `PathlibPathChecker`:
```diff
@@ -130,7 +168,14 @@ pub(crate) fn blocking_os_path(checker: &Checker, call: &ExprCall) {
         return;
     };
 
-    if check_type::<PathMethodInAsyncChecker>(binding, semantic) {
-        checker.report_diagnostic(BlockingPathMethodInAsyncFunction, call.func.range());
+    if check_type::<PathlibPathChecker>(binding, semantic) {
+        if maybe_calling_io_operation(attr.id.as_str()) {
+            checker.report_diagnostic(
+                BlockingPathMethodInAsyncFunction {
+                    path_library: "pathlib.Path".to_string(),
+                },
+                call.func.range(),
+            );
+        }
     }
```

Its `match_annotation` is missing a logic that checks if `Path` is in union and optional. Added that logic in this commit: **`Update PathlibPathChecker to check nested annotation.`**

PieterCK: > This checks some sub/supeclasses too, we just need to double-check if we want that.

There are some classes in that check that aren't relevant for checking `ASYNC240`; `PurePath`, `PurePosixPath` and `PureWindowsPath`. They all don't have I/O related methods.

Not excluding them from that check shouldn't be a problem though, expressions calling those objects methods will still be filtered out by `maybe_calling_io_operation` since they have exactly the same method as `PurePath`.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,181 @@
...
+            // Non I/O pathlib.Path or os.path methods:
+            | "join"
+            | "dirname"
+            | "basename"
+            | "splitroot"
+            | "splitdrive"
+            | "splitext"
+            | "split"
+            | "isabs"
+            | "normcase"
```
PieterCK: I've manually inspected these methods. They only do basic computation and don’t use any functions that `join` doesn’t use, so they’re probably safe.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,181 @@
...
+            | "parts"
+            | "relative_to"
+            | "root"
+            | "stem"
+            | "suffix"
+            | "suffixes"
+            | "with_name"
+            | "with_segments"
+            | "with_stem"
+            | "with_suffix"
```
PieterCK: This list is generated using a simple python (3.10.12) script:
```python
for attr in dir(pathlib.PurePath):
    if not attr.startswith("_"):
        print(f'"{attr}",')
```

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: > Although on second thought that might not catch calls like `os.path.exists(thing)`, so maybe instead make sure that eg, `pathlib.Path(...)` is allowed while `pathlib.Path(...).exists()` isn't.

The linter should be able to handle these cases now.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,181 @@
...
+    // calls an I/O method.
+    let Some(ast::ExprAttribute { value, attr, .. }) = call.func.as_attribute_expr() else {
+        return;
+    };
+
+    if let Some(ExprCall { func, .. }) = value.as_call_expr() {
+        if let Some(name) = func.as_name_expr() {
+            if semantic
+                .resolve_qualified_import_name("pathlib", name.id.as_str())
+                .is_none()
```
PieterCK, line 145: `Path(...).open()` parses as two separate expressions.

<details><summary>First expression</summary>

```
[crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs:130:9] call = ExprCall {
    node_index: NodeIndex(None),
    range: 1922..1945,
    func: Name(
        ExprName {
            node_index: NodeIndex(None),
            range: 1922..1926,
            id: Name("Path"),
            ctx: Load,
        },
    ),
    arguments: Arguments {
        range: 1926..1945,
        node_index: NodeIndex(None),
        args: [
            StringLiteral(
                ExprStringLiteral {
                    node_index: NodeIndex(None),
                    range: 1927..1944,
                    value: StringLiteralValue {
                        inner: Single(
                            StringLiteral {
                                range: 1927..1944,
                                node_index: NodeIndex(None),
                                value: "src/my_text.txt",
                                flags: StringLiteralFlags {
                                    quote_style: Double,
                                    prefix: Empty,
                                    triple_quoted: false,
                                },
                            },
                        ),
                    },
                },
            ),
        ],
        keywords: [],
    },
}
```
</details> 

<details><summary>Second expression</summary>

```
call = ExprCall {
    node_index: NodeIndex(None),
    range: 1922..1952,
    func: Attribute(
        ExprAttribute {
            node_index: NodeIndex(None),
            range: 1922..1950,
            value: Call(
                ExprCall {
                    node_index: NodeIndex(None),
                    range: 1922..1945,
                    func: Name(
                        ExprName {
                            node_index: NodeIndex(None),
                            range: 1922..1926,
                            id: Name("Path"),
                            ctx: Load,
                        },
                    ),
                    arguments: Arguments {
                        range: 1926..1945,
                        node_index: NodeIndex(None),
                        args: [
                            StringLiteral(
                                ExprStringLiteral {
                                    node_index: NodeIndex(None),
                                    range: 1927..1944,
                                    value: StringLiteralValue {
                                        inner: Single(
                                            StringLiteral {
                                                range: 1927..1944,
                                                node_index: NodeIndex(None),
                                                value: "src/my_text.txt",
                                                flags: StringLiteralFlags {
                                                    quote_style: Double,
                                                    prefix: Empty,
                                                    triple_quoted: false,
                                                },
                                            },
                                        ),
                                    },
                                },
                            ),
                        ],
                        keywords: [],
                    },
                },
            ),
            attr: Identifier {
                id: Name("open"),
                range: 1946..1950,
                node_index: NodeIndex(None),
            },
            ctx: Load,
        },
    ),
    arguments: Arguments {
        range: 1950..1952,
        node_index: NodeIndex(None),
        args: [],
        keywords: [],
    },
}
```
</details>

Since we can't do `resolve_qualified_name` on the second expression to see if it's calling a `pathlib.Path` object, I had to use `resolve_qualified_import_name` to do that check.

PieterCK:
Thanks for reviewing this! Marking this as ready for review again. What's the best way to show new PR changes here? I've updated the existing commit this time.

The main changes are around how the linter determines whether an expression is an `os.path` or `pathlib.Path` method call that performs I/O. Previously, even constructing a `Path` object in async function triggers violation. Comment thread: https://github.com/astral-sh/ruff/pull/20264#discussion_r2325468204


<details>

<summary>

git diff of the `blocking_path_methods.rs` file:</summary>

```diff
git diff be73d44f8b10 c2125edbbb8d -- crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
diff --git a/crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs b/crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
index c67ce67a8..50e7059bc 100644
--- a/crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
+++ b/crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
@@ -1,8 +1,8 @@
 use crate::Violation;
 use crate::checkers::ast::Checker;
 use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::{self as ast, Expr, ExprCall};
-use ruff_python_semantic::analyze::typing::{TypeChecker, check_type, traverse_union_and_optional};
+use ruff_python_ast::{self as ast, ExprCall};
+use ruff_python_semantic::analyze::typing::{PathlibPathChecker, check_type};
 use ruff_text_size::Ranged;
 
 /// ## What it does
@@ -35,68 +35,70 @@ use ruff_text_size::Ranged;
 ///     file_exists = await path.exists()
 /// ```
 #[derive(ViolationMetadata)]
-pub(crate) struct BlockingPathMethodInAsyncFunction;
+pub(crate) struct BlockingPathMethodInAsyncFunction {
+    path_library: String,
+}
 
 impl Violation for BlockingPathMethodInAsyncFunction {
     #[derive_message_formats]
     fn message(&self) -> String {
-        "Async functions should not use pathlib.Path or os.path methods, use trio.Path or anyio.path".to_string()
+        format!(
+            "Async functions should not use {path_library} methods, use trio.Path or anyio.path",
+            path_library = self.path_library
+        )
     }
 }
 
-const OS_PATH: [&str; 2] = ["os", "path"];
-const PATHLIB_PATH: [&str; 2] = ["pathlib", "Path"];
-
-fn is_blocking_path_method(segments: &[&str]) -> bool {
-    segments.starts_with(&OS_PATH) || segments.starts_with(&PATHLIB_PATH)
-}
-
-struct PathMethodInAsyncChecker;
-
-impl TypeChecker for PathMethodInAsyncChecker {
-    fn match_annotation(
-        annotation: &ruff_python_ast::Expr,
-        semantic: &ruff_python_semantic::SemanticModel,
-    ) -> bool {
-        // match base annotation directly
-        if semantic
-            .resolve_qualified_name(annotation)
-            .is_some_and(|qualified_name| is_blocking_path_method(qualified_name.segments()))
-        {
-            return true;
-        }
-
-        // otherwise traverse any union or optional annotation
-        let mut found = false;
-        traverse_union_and_optional(
-            &mut |inner_expr, _| {
-                if semantic
-                    .resolve_qualified_name(inner_expr)
-                    .is_some_and(|qualified_name| {
-                        is_blocking_path_method(qualified_name.segments())
-                    })
-                {
-                    found = true;
-                }
-            },
-            semantic,
-            annotation,
-        );
-        found
+fn is_calling_os_path_method(segments: &[&str]) -> bool {
+    if segments.len() != 3 {
+        return false;
     }
+    let Some(symbol_name) = segments.get(..2) else {
+        return false;
+    };
+    matches!(symbol_name, ["os", "path"])
+}
 
-    fn match_initializer(
-        initializer: &ruff_python_ast::Expr,
-        semantic: &ruff_python_semantic::SemanticModel,
-    ) -> bool {
-        let Expr::Call(ExprCall { func, .. }) = initializer else {
-            return false;
-        };
-
-        semantic
-            .resolve_qualified_name(func)
-            .is_some_and(|qualified_name| is_blocking_path_method(qualified_name.segments()))
-    }
+fn maybe_calling_io_operation(attr: &str) -> bool {
+    !matches!(
+        attr,
+        // Pure path objects provide path-handling operations which don’t actually
+        // access a filesystem.
+        // https://docs.python.org/3/library/pathlib.html#pure-paths
+        // pathlib.PurePath methods and properties:
+        "anchor"
+            | "as_posix"
+            | "as_uri"
+            | "drive"
+            | "is_absolute"
+            | "is_relative_to"
+            | "is_reserved"
+            | "joinpath"
+            | "match"
+            | "name"
+            | "parent"
+            | "parents"
+            | "parts"
+            | "relative_to"
+            | "root"
+            | "stem"
+            | "suffix"
+            | "suffixes"
+            | "with_name"
+            | "with_segments"
+            | "with_stem"
+            | "with_suffix"
+            // Non I/O pathlib.Path or os.path methods:
+            | "join"
+            | "dirname"
+            | "basename"
+            | "splitroot"
+            | "splitdrive"
+            | "splitext"
+            | "split"
+            | "isabs"
+            | "normcase"
+    )
 }
 
 /// ASYNC240
@@ -106,22 +108,53 @@ pub(crate) fn blocking_os_path(checker: &Checker, call: &ExprCall) {
         return;
     }
 
-    // Check if an expression is directly calling one of the blocking `path`
-    // objects.
+    // Check if an expression is calling I/O related os.path method.
+    // Just intializing pathlib.Path object is OK, we can return
+    // early in that scenario.
     if let Some(qualified_name) = semantic.resolve_qualified_name(call.func.as_ref()) {
         let segments = qualified_name.segments();
-        if is_blocking_path_method(segments) {
-            checker.report_diagnostic(BlockingPathMethodInAsyncFunction, call.func.range());
+        if !is_calling_os_path_method(segments) {
+            return;
+        }
+
+        let Some(os_path_method) = segments.last() else {
+            return;
+        };
+
+        if maybe_calling_io_operation(os_path_method) {
+            checker.report_diagnostic(
+                BlockingPathMethodInAsyncFunction {
+                    path_library: "os.path".to_string(),
+                },
+                call.func.range(),
+            );
         }
         return;
     }
 
-    // Past this, we're checking if a variable contains one of the blocking
-    // `path` objects.
-    let Some(ast::ExprAttribute { value, .. }) = call.func.as_attribute_expr() else {
+    // Check if an expression is a pathlib.Path constructor that directly
+    // calls an I/O method.
+    let Some(ast::ExprAttribute { value, attr, .. }) = call.func.as_attribute_expr() else {
         return;
     };
 
+    if let Some(ExprCall { func, .. }) = value.as_call_expr() {
+        if !PathlibPathChecker::is_pathlib_path_constructor(semantic, func) {
+            return;
+        }
+        if maybe_calling_io_operation(attr.id.as_str()) {
+            checker.report_diagnostic(
+                BlockingPathMethodInAsyncFunction {
+                    path_library: "pathlib.Path".to_string(),
+                },
+                call.func.range(),
+            );
+        }
+        return;
+    }
+
+    // Lastly, check if a variable is a pathlib.Path instance and it's
+    // calling an I/O method.
     let Some(name) = value.as_name_expr() else {
         return;
     };
@@ -130,7 +163,14 @@ pub(crate) fn blocking_os_path(checker: &Checker, call: &ExprCall) {
         return;
     };
 
-    if check_type::<PathMethodInAsyncChecker>(binding, semantic) {
-        checker.report_diagnostic(BlockingPathMethodInAsyncFunction, call.func.range());
+    if check_type::<PathlibPathChecker>(binding, semantic) {
+        if maybe_calling_io_operation(attr.id.as_str()) {
+            checker.report_diagnostic(
+                BlockingPathMethodInAsyncFunction {
+                    path_library: "pathlib.Path".to_string(),
+                },
+                call.func.range(),
+            );
+        }
     }
 }

```
</details>

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
jakkdl: The upstream implementation is not very fancy: https://github.com/python-trio/flake8-async/blob/d4c412ecc2ec2d4bdcd3908e0769adf60c1054ed/flake8_async/visitors/visitor2xx.py#L338

PieterCK:
Bumping this PR for review again.

ntBre:
> What's the best way to show new PR changes here?

I usually just push additional commits. That works better with GitHub's "Changes since your last review" button than comparing rebased commits. We always squash and merge PRs anyway. But it's not too big of a deal either way :) 

I'll give this another look now!

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+fn is_calling_os_path_method(segments: &[&str]) -> bool {
+    if segments.len() != 3 {
+        return false;
+    }
+    let Some(symbol_name) = segments.get(..2) else {
+        return false;
+    };
+    matches!(symbol_name, ["os", "path"])
```
ntBre: I think we can combine these three checks into one pattern match:


```suggestion
    matches!(segments, ["os", "path", _])
```

If we wanted to accept more than 3 components we could just use `..` instead of `_`. I don't think that helps here, just another neat trick.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
+use crate::Violation;
+use crate::checkers::ast::Checker;
+use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::{self as ast, ExprCall};
+use ruff_python_semantic::analyze::typing::{PathlibPathChecker, check_type};
+use ruff_text_size::Ranged;
+
+/// ## What it does
+/// Checks that async functions do not use `os.path` or `pathlib.Path`.
```
ntBre: I think we should clarify here and in the `Violation::message` implementation that only blocking functions are disallowed now that the modules as a whole aren't flagged.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
...
+            | "splitdrive"
+            | "splitext"
+            | "split"
+            | "isabs"
+            | "normcase"
+    )
+}
+
+/// ASYNC240
+pub(crate) fn blocking_os_path(checker: &Checker, call: &ExprCall) {
```
ntBre, line 65: nit: We usually put the main rule implementation function just below the `Violation` stuff and then helper functions after it.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
...
+        return false;
+    }
+    let Some(symbol_name) = segments.get(..2) else {
+        return false;
+    };
+    matches!(symbol_name, ["os", "path"])
+}
+
+fn maybe_calling_io_operation(attr: &str) -> bool {
+    !matches!(
```
ntBre: Are all of these shared between both `os.path` and `pathlib.Path`? I know sometimes those two modules have different names for similar functionality, but I didn't check all of these myself so it could be correct.

I'm also wondering if a deny list might be shorter than this allow list. If they're of comparable length, this is fine, just an idea.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
...
+    // Check if an expression is a pathlib.Path constructor that directly
+    // calls an I/O method.
+    let Some(ast::ExprAttribute { value, attr, .. }) = call.func.as_attribute_expr() else {
+        return;
+    };
+
+    if let Some(ExprCall { func, .. }) = value.as_call_expr() {
+        if !PathlibPathChecker::is_pathlib_path_constructor(semantic, func) {
+            return;
+        }
```
ntBre: It looks like if you use `PathlibPathChecker::match_initializer` instead of `is_pathlib_path_constructor`, it will handle the `ExprCall` check for you too:

https://github.com/astral-sh/ruff/blob/cb3c3ba94d329083abf430072d740acf2aa476c7/crates/ruff_python_semantic/src/analyze/typing.rs#L919-L925

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
```diff
@@ -0,0 +1,176 @@
...
+    let Some(name) = value.as_name_expr() else {
+        return;
+    };
+
+    let Some(binding) = semantic.only_binding(name).map(|id| semantic.binding(id)) else {
+        return;
+    };
+
+    if check_type::<PathlibPathChecker>(binding, semantic) {
+        if maybe_calling_io_operation(attr.id.as_str()) {
```
ntBre: This is the same `attr` as above right? I think we could do this check once before both the constructor and `check_type` checks, returning early if it's not one of the known attributes.

crates/ruff_linter/resources/test/fixtures/flake8_async/ASYNC240.py
```diff
@@ -0,0 +1,106 @@
...
+    with path.open() as f: # ASYNC240
+        ...
+
+async def pathlib_path_in_foo():
+    import pathlib
+
+    path = pathlib.Path("src/my_text.txt")
+    path.exists() # ASYNC240
+    with path.open() as f: # ASYNC240
+        ...
```
ntBre: I think we need to exclude `open` because it's covered by [blocking-open-call-in-async-function (ASYNC230)](https://docs.astral.sh/ruff/rules/blocking-open-call-in-async-function/#blocking-open-call-in-async-function-async230).

crates/ruff_python_semantic/src/analyze/typing.rs
```diff
@@ -913,7 +913,21 @@ impl PathlibPathChecker {
...
+        let mut found = false;
+        traverse_union_and_optional(
+            &mut |inner_expr, _| {
+                if Self::is_pathlib_path_constructor(semantic, inner_expr) {
+                    found = true;
+                }
+            },
+            semantic,
+            annotation,
+        );
```
ntBre: I do like having this check like in #20091, but I'm a little bit hesitant to add it here since it will affect other rules too. I would probably either revert this and accept that we miss option/union annotations for now, until we possibly update the `TypeChecker` infrastructure more generally in the future, or move this check into the rule itself, as @amyreese did for `ASYNC212`.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: The deny list version looks like this (excluding `open`):
```    
rglob
abspath
isjunction
isdir
stat
lstat
relpath
normpath
is_fifo
mkdir
iterdir
getmtime
islink
hardlink_to
samefile
absolute
walk
is_junction
realpath
is_symlink
getsize
is_socket
getctime
read_bytes
is_file
owner
is_dir
symlink_to
rmdir
sameopenfile
_path_normpath
isfile
chmod
resolve
is_block_device
is_mount
unlink
ismount
rename
glob
write_text
group
lexists
expanduser
readlink
cwd
replace
home
exists
is_char_device
lchmod
open
read_text
touch
getatime
write_bytes
_joinrealpath
```
Length = 57 methods

This is a list of combined async versions of `pathlib.Path` IO methods listed in trio and anyio documentation  + the [deny list from upstream](https://github.com/python-trio/flake8-async/blob/d4c412ecc2ec2d4bdcd3908e0769adf60c1054ed/flake8_async/visitors/visitor2xx.py#L338). I ran this through a Python script to make sure `pathlib.Path` or `os.path` have methods with the same names.

Reversing that and iterating over `pathlib.Path` + `os.path` methods to find non I/O methods (methods that aren't in the deny list and aren't private class methods)  yields about 50 methods, so a "complete" allow list would probably about the same length as the deny list.

PieterCK: How I got the list:
- anyio neatly listed all I/O methods that they support on their website: https://anyio.readthedocs.io/en/latest/api.html#anyio.Path

- for trio, I had to grep it from their website and clean up abit. `grep -oP '(?<=Like ).*(?=, but async)'`. website: https://trio.readthedocs.io/en/stable/reference-io.html

And the python script I used:

<details>

```python
from pathlib import Path

methods_to_check = {
    "absolute",
    "chmod",
    "cwd",
    "exists",
    "expanduser",
    "glob",
    "group",
    "hardlink_to",
    "home",
    "is_block_device",
    "is_char_device",
    "is_dir",
    "is_fifo",
    "is_file",
    "is_junction",
    "is_mount",
    "is_socket",
    "is_symlink",
    "iterdir",
    "lchmod",
    "link_to",
    "lstat",
    "mkdir",
    "open",
    "owner",
    "read_bytes",
    "read_text",
    "readlink",
    "rename",
    "replace",
    "resolve",
    "rglob",
    "rmdir",
    "samefile",
    "stat",
    "symlink_to",
    "touch",
    "unlink",
    "walk",
    "write_bytes",
    "write_text",
}

os_path_methods = {
    "_path_normpath",
    "normpath",
    "_joinrealpath",
    "islink",
    "lexists",
    "ismount",
    "realpath",
    "exists",
    "isdir",
    "isfile",
    "getatime",
    "getctime",
    "getmtime",
    "getsize",
    "samefile",
    "sameopenfile",
    "relpath",
}
import os

path_methods = dir(Path)
deny_list = []
for method in methods_to_check | os_path_methods:
    if method in path_methods or hasattr(os.path, method):
        deny_list.append(method)

print(len(deny_list))
non_blocking_list = []
os_methods = dir(os.path)
for method in set(path_methods) | set(os_methods):
    if method not in deny_list and not method.startswith("_"):
        non_blocking_list.append(method)

print(len(non_blocking_list))
```
</details>

PieterCK: Poking around `os.path` methods that doesn't match the deny list, I found a couple more suspects: `isjunction` and `abspath`.

Both methods look like they perform I/O and probably are `os.path` counterpart for `is_junction` and `absolute` (which are `pathlib` I/O methods). Might miss a few though since I only check suspicious looking ones, so here's the list:

```
['ALLOW_MISSING', 'altsep', 'basename', 'commonpath', 'commonprefix', 'curdir', 'defpath', 'devnull', 'dirname', 'expandvars', 'extsep', 'genericpath', 'isabs', 'join', 'normcase', 'os', 'pardir', 'pathsep', 'samestat', 'sep', 'split', 'splitdrive', 'splitext', 'splitroot', 'supports_unicode_filenames', 'sys']
```

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: Neat! Implemented this version

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: Okay, I've added another example using non-blocking methods in async function too.

crates/ruff_linter/resources/test/fixtures/flake8_async/ASYNC240.py
PieterCK: Ok! Now `open` passes `maybe_calling_io_operation`. There is a new test for this too

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: Great! Updated to use `match_initializer` for this check.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
PieterCK: Aha, good catch. I moved this a few checks above to just before the constructor check.

crates/ruff_python_semantic/src/analyze/typing.rs
PieterCK: Makes sense, for now I added a copy of `PathlibPathChecker` that has the traversing logic to this rule.

PieterCK:
Thank you for doing another pass on this! Will post another comment once this is ready for review again.

crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs
ntBre: Thanks for checking this so thoroughly! I'll defer to you on the allow vs deny list, whichever you think is easier.

amyreese: IMO I think an allowlist is "better" because any new methods added in the future couldn't silently slip by unnoticed by the linter.

PieterCK: Ok! Updated it to an allow list

PieterCK:
@ntBre -- Marking this as ready for you to take a look again!

crates/ruff_python_semantic/src/analyze/typing.rs
```diff
@@ -912,6 +912,11 @@ impl PathlibPathChecker {
 }
 
 impl TypeChecker for PathlibPathChecker {
+    // match_annotation doesn't check `Path` in union
+    // or `Optional`.
+    // TODO: Once we have updated it to handle those
+    // cases, update `blocking_path_method.rs`(ASYNC240)
+    // to use this version of `PathlibPathChecker`.
```
ntBre: I think it's good to record this TODO somewhere, but I'm kind of leaning toward opening a separate issue and deleting this comment. I feel like we should probably either always traverse unions or never traverse unions and apply that decision to all of our `TypeChecker`s rather than keep a mix of them.

PieterCK: Okay. This commit removes the comment, `Remove comment on PathlibPathChecker.`

PieterCK: And opened https://github.com/astral-sh/ruff/issues/20529 to keep track of this.

PieterCK:
Awesome! Thank you

PR #18799, opened 2025-06-19
[ty] eliminate is_fully_static

carljm:
## Summary

Having a recursive type method to check whether a type is fully static is inefficient, unnecessary, and makes us overly strict about subtyping relations.

It's inefficient because we end up re-walking the same types many times to check for fully-static-ness.

It's unnecessary because we can check relations involving the dynamic type appropriately, depending whether the relation is subtyping or assignability.

We use the subtyping relation to simplify unions and intersections. We can usefully consider that `S <: T` for gradual types also, as long as it remains true that `S | T` is equivalent to `T` and `S & T` is equivalent to `S`.

One conservative definition (implemented here) that satisfies this requirement is that we consider `S <: T` if, for every possible pair of materializations `S'` and `T'`, `S' <: T'`. Or put differently the top materialization of `S` (`S+` -- the union of all possible materializations of `S`) is a subtype of the bottom materialization of `T` (`T-` -- the intersection of all possible materializations of `T`). In the most basic cases we can usefully say that `Any <: object` and that `Never <: Any`, and we can handle more complex cases inductively from there.

This definition of subtyping for gradual subtypes is not reflexive (`Any` is not a subtype of `Any`).

As a corollary, we also remove `is_gradual_equivalent_to` -- `is_equivalent_to` now has the meaning that `is_gradual_equivalent_to` used to have. If necessary, we could restore an `is_fully_static_equivalent_to` or similar (which would not do an `is_fully_static` pre-check of the types, but would instead pass a relation-kind enum down through a recursive equivalence check, similar to `has_relation_to`), but so far this doesn't appear to be necessary.

Credit to @JelleZijlstra for the observation that `is_fully_static` is unnecessary and overly restrictive on subtyping.

[...]

PR #20267, opened 2025-09-05
[ty] Reimplement equivalence via mutual subtyping

AlexWaygood:


AlexWaygood:
### `beartype`

The [beartype change](https://github.com/beartype/beartype/blob/2c625cac1be01e0caa03bbba4bc2c9bb86a89136/beartype/_util/func/utilfunctest.py#L829-L908) seems to be desirable -- we now consider `((...) -> Unknown) & ~((...) -> Unknown)` to simplify to `Never`, which means things like this now have more intuitive behaviour:

```py
def f(x: object):
    if callable(x):
        reveal_type(x)
        if not callable(x):
            reveal_type(x)  # now `Never`, was previously `((...) -> Unknown) & ~((...) -> Unknown)`
```

### `cwltool`

I'm not totally sure why on `main` we infer the type of of `processobj` as `object` [here](https://github.com/common-workflow-language/cwltool/blob/256306a5da1eb0a8391d5f6734e7baae96922079/cwltool/load_tool.py#L249-L259) on `main`, but it seems like an improvement that we no longer do on this branch.

carljm:
We should narrow `if callable(x)` using the top materialization of `(...) -> Unknown` instead of using a gradual type (just like we recently started doing for invariant generics); that would give us the beartype improvement too.

carljm:
How do the property tests feel about this change?

carljm:
See https://github.com/astral-sh/ruff/pull/18799 for where I previously experimented with this; the PR description there explains why this breaks transitivity of subtyping.

JelleZijlstra:
> we now consider `((...) -> Unknown) & ~((...) -> Unknown)` to simplify to `Never`, which means things like this now have more intuitive behaviour:

That seems incorrect in principle. That type is equivalent to `((...) -> Unknown)`, for the same reason that `~Any` is the same as `Any`.

`callable()` is currently defined as returning `TypeIs[Callable[..., object]]`. It should really be `TypeIs[Top[Callable[..., object]]]`, though of course we can't do that in typeshed. Might be worth special-casing in ty though.

AlexWaygood:
I think you're right that making `Any` a subtype of `Any` probably isn't theoretically sound. The motivation here was to see if we could implement equivalence via mutual subtyping rather than having an entirely different logical approach to equivalence. That would mean that we could delete a lot of code and probably be more internally consistent about things (see the updated PR diff!). It would also make it easier to explore doing less eager union simplification.

Anyway, it's not possible to implement `a.is_equivalent_to(b)` as `a.is_subtype_of(b) && b.is_subtype_of(a)` unless `Any.is_subtype_of(Any)` holds, since our current logic (correctly, I think) views `Any` as equivalent to itself. @JelleZijlstra's idea in https://github.com/astral-sh/ty/issues/994 of implementing equivalence using `Top[]` and `Bottom[]` types might work better here. One more option is that we might consider introducing a third type relation -- sitting between `TypeRelation::Assignability` and `TypeRelation::Subtyping` -- `TypeRelation::UnionSimplification`. This could be used to reflect the fact that `Any` is not a subtype of `Any`, but `Any` is equivalent to `Any` and the union `Any | Any` can be simplified to `Any`.

AlexWaygood:
Anyway, I don't plan to spend much more time on this right now; it was just an experiment prompted by https://github.com/astral-sh/ty/issues/1132 (for which one solution would be to do less eager union simplification -- but doing less eager union simplification breaks our current implementation of equivalence).

> `callable()` is currently defined as returning `TypeIs[Callable[..., object]]`. It should really be `TypeIs[Top[Callable[..., object]]]`, though of course we can't do that in typeshed. Might be worth special-casing in ty though.

the actual code in beartype used a custom function with a `TypeIs[Callable]` return type FWIW, which would be harder to special-case; I just used `callable()` in my comment above for a smaller repro.

PR #20268, opened 2025-09-05
Stabilize new strategy for classifying imports as first party

dylwil3:
This stabilizes the behavior introduced in #16565 which (roughly) tries to match an import like `import a.b.c` to an actual directory path `a/b/c` in order to label it as first-party, rather than simply looking for a directory `a`.

Mainly this affects the sorting of imports in the presence of namespace packages, but a few other rules are affected as well.


crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -91,11 +88,6 @@ use crate::{Applicability, Fix, FixAvailability, Violation};
 ///     print("numpy is not installed")
 /// ```
 ///
-/// ## Preview
-/// When [preview](https://docs.astral.sh/ruff/preview/) is enabled,
-/// the criterion for determining whether an import is first-party
-/// is stricter, which could affect the suggested fix. See [this FAQ section](https://docs.astral.sh/ruff/faq/#how-does-ruff-determine-which-of-my-imports-are-first-party-third-party-etc) for more details.
```
ntBre, line 97: Would it still be helpful to link to the FAQ in some of these rules, just not in a `## Preview` section?

docs/faq.md
```diff
@@ -310,8 +310,7 @@ my_project
 When Ruff sees an import like `import foo`, it will then iterate over the `src` directories,
 looking for a corresponding Python module (in reality, a directory named `foo` or a file named
 `foo.py`). For module paths with multiple components like `import foo.bar`,
-the default behavior is to search only for a directory named `foo` or a file
-named `foo.py`. However, if `preview` is enabled, Ruff will require that the full relative path `foo/bar` exists as a directory, or that `foo/bar.py` or `foo/bar.pyi` exist as files. Finally, imports of the form `from foo import bar`, Ruff will only use `foo` when determining whether a module is first-party or third-party. 
+Ruff will require that the full relative path `foo/bar` exists as a directory, or that `foo/bar.py` or `foo/bar.pyi` exist as files. Finally, imports of the form `from foo import bar`, Ruff will only use `foo` when determining whether a module is first-party or third-party.
```
ntBre:
```suggestion
Ruff will require that the full relative path `foo/bar` exists as a directory, or that `foo/bar.py` or `foo/bar.pyi` exist as files. Finally, for imports of the form `from foo import bar`, Ruff will only use `foo` when determining whether a module is first-party or third-party.
```

ntBre (approved):
Thanks! I was slightly surprised to see a net increase in diagnostics, but I totally defer to your experience here. It looks like the same set from the original PR anyway.

dylwil3:
> Thanks! I was slightly surprised to see a net increase in diagnostics, but I totally defer to your experience here. It looks like the same set from the original PR anyway.

Yeah these are expected, mostly because folks presumably sorted using the "old" method already.

Issue #6474, opened 2023-08-10
INP001 false positive in tests folder


flying-sheep:
The recommended Python + Pytest layout is

- `src/`
  - `mypkg/`
    - `__init__.py`
    - `subpkg/`
      - `__init__.py`
- `tests/`
  - `test_thing.py`
  - `test_more/`
    - `test_other_thing.py`

Notably, `tests` and all files under it *aren’t packages*. The [recommended import mode](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code) enforces that you can’t import from them.

Therefore `INP001` shouldn’t trigger when `tests` is on top level of the project.

In fact, it would be great to have both a rule that enforces the opposite and to ensure that `tests/**/test_*.py` files don’t import things from each other.

tjkuson:
I think this could be related to #6114.

flying-sheep:
Pytest collects and imports individual files. That’s independent from Python’s namespace packages.

It’s possible that that other issue incidentally fixes this, but this is an independent issue.

charliermarsh:
Interesting. To what degree is this specific to pytest? On the surface it feels like per-file-ignores is the right solution here…

flying-sheep:
It’s very specific to Pytest. I’m coming from the perspective that when people have e.g.

```toml
[tool.ruff]
select = ['INP', 'PT']
```

```
project_dir
+- src/...
+- tests/
   +- test_foo.py
   +- test_bar/
      +- __init__.py
      +- test_bar_basic.py
```

then they should optimally see something like this:

```
tests/test_bar/__init__.py:1:1: PT030 File `tests/test_bar/__init__.py` found. Pytest test directories shouldn’t contain packages.
```

Currently they will see an instruction to do the opposite of the recommended route:

```
tests/test_foo.py:1:1: INP001 File `tests/test_foo.py` is part of an implicit namespace package. Add an `__init__.py`.
```

sanmai-NL:
See also #8690

Artalus:
For what it's worth, it seems that even _with_ the `__init__.py` file present, `ruff` still complains about the `INP001` (unlike the original flake linter):
```
[artalus@idpad foo]$ tree --gitignore
.
├── pyproject.toml
├── README.md
├── src
│   └── foo
│       └── __init__.py
├── tests
│   └── test_foo.py
└── uv.lock

[artalus@idpad foo]$ uv pip install flake8-no-pep420 ruff
[...]

PR #20269, opened 2025-09-05
[`flake8-no-pep420`] Stabilize allowing nested roots in `implicit-namespace-package` (`INP001`)

dylwil3:
Stabilizes the behavior from #14236

This was never documented, so we did not update documentation for the rule.


dylwil3:
Hmm. Looking at the ecosystem report, I guess this exacerbates the problem from https://github.com/astral-sh/ruff/issues/6474 so maybe better to resolve that first.

PR #20278, opened 2025-09-06
[ty] initial support for `slots=True` in dataclasses

thejchap:
## Summary

https://github.com/astral-sh/ty/issues/111
https://docs.python.org/3/library/dataclasses.html

this pr adds initial support for `slots=True` in the `dataclass` decorator. synthesizes `__slots__` as returning a fixed-length tuple of `str`s (field names)

## Test Plan

- new mdtests
- typing conformance diffs look to be true positives - https://diffswarm.dev/d-01k4eb7wx0wqecmxpvxxnch9ag

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
@@ -546,7 +546,29 @@ class A:
 
 ### `slots`
 
-To do
+If a dataclass is defined with `slots=True`, the `__slots__` attribute is generated as a tuple. It
+is not present otherwise.
+
+```py
+from dataclasses import dataclass
+from typing import Tuple
+
+@dataclass
+class A:
+    x: int
+    y: int
+
+# revealed: Unknown
+# error: [unresolved-attribute]
+reveal_type(A.__slots__)
+
+@dataclass(slots=True)
+class B:
+    x: int
+    y: int
+
+reveal_type(B.__slots__)  # revealed: tuple[Literal["x"], Literal["y"]]
+```
 
 ### `weakref_slot`
 
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -2282,6 +2282,13 @@ impl<'db> ClassLiteral<'db> {
                 }
                 None
             }
+            (CodeGeneratorKind::DataclassLike, "__slots__") => {
+                has_dataclass_param(DataclassParams::SLOTS).then(|| {
+                    let fields = self.fields(db, specialization, field_policy);
+                    let slots = fields.keys().map(|name| Type::string_literal(db, name));
+                    Type::heterogeneous_tuple(db, slots)
+                })
+            }
             (CodeGeneratorKind::TypedDict, "__setitem__") => {
                 let fields = self.fields(db, specialization, field_policy);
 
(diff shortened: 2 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -546,7 +546,29 @@ class A:
 
 ### `slots`
 
-To do
+If a dataclass is defined with `slots=True`, the `__slots__` attribute is generated as a `Tuple`. It
```
AlexWaygood:
```suggestion
If a dataclass is defined with `slots=True`, the `__slots__` attribute is generated as a tuple. It
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2282,6 +2282,15 @@ impl<'db> ClassLiteral<'db> {
...
                 None
             }
+            (CodeGeneratorKind::DataclassLike, "__slots__") => {
+                if has_dataclass_param(DataclassParams::SLOTS) {
+                    let field_count = self.fields(db, specialization, field_policy).len();
+                    let tuple_elements = vec![KnownClass::Str.to_instance(db); field_count];
+                    let tuple_type = Type::heterogeneous_tuple(db, tuple_elements);
+                    return Some(tuple_type);
+                }
+                None
```
AlexWaygood: At some point we will need to know the precise string values of `__slots__`, because at some point we will want to emit an diagnostic on code like this:

```py
@dataclass(slots=True)
class F:
    x: int

    def method(self):
        self.y = 42  # will raise an exception because `y` is not in `__slots__`
```

We don't need to add that diagnostic in this PR (and we'll want to emit the diagnostic on non-dataclasses as well as dataclasses). But ISTM that we might as well infer this now as a tuple of precise string-literal types, rather than as a tuple of `str`s? It will help support that future feature.

```suggestion
                has_dataclass_param(DataclassParams::SLOTS).then(|| {
                    let fields = self.fields(db, specialization, field_policy);
                    let slots = fields.keys().map(|name| Type::string_literal(db, name));
                    Type::heterogeneous_tuple(db, slots)
                })
```

AlexWaygood (commented):
Thanks, this is great! A couple of small points below.

Could you also add something like this test case to `https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/instance_layout_conflict.md` (it already passes on your branch!):

```py

@dataclass(slots=True)
class F: ...

@dataclass(slots=True)
class G: ...

class H(F, G): ...  # fine because both classes have empty `__slots__`

@dataclass(slots=True)
class I:
    x: int

@dataclass(slots=True)
class J:
    y: int

class K(I, J): ...  # error: [instance-layout-conflict]
```

And something like this test case to `https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/type_properties/is_disjoint_from.md` (it also already passes on your branch!):

```py
from ty_extensions import is_disjoint_from, static_assert

@dataclass(slots=True)
class F: ...

@dataclass(slots=True)
class G: ...

@dataclass(slots=True)
class I:
    x: int

@dataclass(slots=True)
class J:
    y: int

# A dataclass with empty `__slots__` is not disjoint from another dataclass with `__slots__`
static_assert(not is_disjoint_from(F, G))
static_assert(not is_disjoint_from(F, I))
static_assert(not is_disjoint_from(G, I))
static_assert(not is_disjoint_from(F, J))
static_assert(not is_disjoint_from(G, J))

# But two dataclasses with non-empty `__slots__` are disjoint
static_assert(is_disjoint_from(I, J))
```

crates/ty_python_semantic/src/types/class.rs
thejchap: ah! great - makes sense. thanks

AlexWaygood (approved):
Thanks!

PR #20279, opened 2025-09-06
[`refurb`] Mark `single-item-membership-test` fix as always unsafe (`FURB171`)

### LINKED ISSUE #20255: `single-item-membership-test` (FURB171) should be marked unsafe
### Summary

[`single-item-membership-test` (FURB171)](https://docs.astral.sh/ruff/rules/single-item-membership-test/) can change behavior in two ways. First, `in` tries both `is` and `==`, whereas the fix is just `==`, which changes behavior for values like NaN. There is no way to know in general which values it is safe for, although it would be possible for this rule to safely special-case certain expressions like `str` and number literals. Second, `in` returns a `bool` whereas `==` can return anything. The current fix is only safe in boolean contexts like a `bool` argument or an `if` condition. Otherwise, the fix should be marked unsafe or the fix should wrap the expression in a `bool` call (or use the `not` operator, in the negated case). [Example](https://play.ruff.rs/764834c7-4ee1-4126-b270-9ce50dc5f014):
```console
$ cat >furb171.py <<'# EOF'
import numpy as np
print(np.nan in [np.nan])
print(np.zeros(1) in [np.zeros(1)])
# EOF

$ python furb171.py
True
True

$ ruff --isolated check furb171.py --select FURB171 --preview --fix

$ cat furb171.py
import numpy as np
print(np.nan == np.nan)
print(np.zeros(1) == np.zeros(1))

$ python furb171.py
False
[ True]
```
When the container is a string literal and the fix is currently marked safe, that’s fine. This is only an issue for `list`, `tuple`, `set`, and `frozenset`.

### Version

TaKO8Ki:
## Summary

Fixes #20255

Mark single-item-membership-test fixes as always unsafe

- Always set `Applicability::Unsafe` for FURB171 fixes
- Update “Fix safety” docs to reflect always-unsafe behavior
- Expand tests (not in, nested set/frozenset, commented args)

## Test Plan

I have added new test cases to `crates/ruff_linter/resources/test/fixtures/refurb/FURB171_0.py` and `crates/ruff_linter/resources/test/fixtures/refurb/FURB171_1.py`.

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
@@ -6,7 +6,7 @@ use ruff_text_size::Ranged;
 
 use crate::checkers::ast::Checker;
 use crate::fix::edits::pad;
-use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
+use crate::{Edit, Fix, FixAvailability, Violation};
 
 /// ## What it does
 /// Checks for membership tests against single-item containers.
@@ -28,6 +28,9 @@ use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
 ///
 /// ## Fix safety
 /// The fix is always marked as unsafe.
+///
+/// When the right-hand side is a string, this fix can change the behavior of your program.
+/// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string,
 ///
 /// Converting `in`/`not in` against a single-item container to `==`/`!=` can
 /// change runtime behavior: `in` may consider identity (e.g., `NaN`), it always
@@ -99,9 +102,7 @@ pub(crate) fn single_item_membership_test(
     );
 
     // All supported cases can change runtime behavior; mark as unsafe.
-    let applicability = Applicability::Unsafe;
-
-    let fix = Fix::applicable_edit(edit, applicability);
+    let fix = Fix::unsafe_edit(edit);
 
     checker
         .report_diagnostic(SingleItemMembershipTest { membership_test }, expr.range())
--- a/crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
@@ -27,13 +27,11 @@ use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
 /// ```
 ///
 /// ## Fix safety
+/// The fix is always marked as unsafe.
 ///
-/// When the right-hand side is a string, the fix is marked as unsafe.
-/// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string,
-/// so the fix can change the behavior of your program in these cases.
-///
-/// Additionally, if there are comments within the fix's range,
-/// it will also be marked as unsafe.
+/// Converting `in`/`not in` against a single-item container to `==`/`!=` can
+/// change runtime behavior: `in` may consider identity (e.g., `NaN`), it always
+/// yields a `bool`, and comments within the replacement range may be removed.
 ///
 /// ## References
 /// - [Python documentation: Comparisons](https://docs.python.org/3/reference/expressions.html#comparisons)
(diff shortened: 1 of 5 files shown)


crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
```diff
@@ -100,12 +98,8 @@ pub(crate) fn single_item_membership_test(
...
     );
 
-    let applicability =
-        if right.is_string_literal_expr() || checker.comment_ranges().intersects(expr.range()) {
-            Applicability::Unsafe
-        } else {
-            Applicability::Safe
-        };
+    // All supported cases can change runtime behavior; mark as unsafe.
+    let applicability = Applicability::Unsafe;
```
TaKO8Ki: Previously, fixes were marked unsafe only for string literals and when comments intersected the edit range, but single‑item list/tuple/set/frozenset cases can also
  change behavior (e.g., NaN identity differences and non‑boolean eq), so we now mark all FURB171 fixes as unsafe.

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
```diff
@@ -100,12 +98,8 @@ pub(crate) fn single_item_membership_test(
...
 
     let fix = Fix::applicable_edit(edit, applicability);
```
ntBre: We could just use `Fix::unsafe_edit` now that it's always unsafe instead of a separate `applicability`.

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
```diff
@@ -27,13 +27,11 @@ use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
...
 ///
-/// When the right-hand side is a string, the fix is marked as unsafe.
-/// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string,
-/// so the fix can change the behavior of your program in these cases.
-///
-/// Additionally, if there are comments within the fix's range,
-/// it will also be marked as unsafe.
+/// Converting `in`/`not in` against a single-item container to `==`/`!=` can
+/// change runtime behavior: `in` may consider identity (e.g., `NaN`), it always
+/// yields a `bool`, and comments within the replacement range may be removed.
```
ntBre: I think we should probably preserve the details about strings from the original docs. The `in` vs `==` thing seems pretty niche, while the string case seems more likely, so it feels worth mentioning both.

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
TaKO8Ki: I've replaced it with `unsafe_edit`.

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
TaKO8Ki: I have revived a part of original docs.

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
```diff
@@ -27,13 +27,14 @@ use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
...
-/// When the right-hand side is a string, the fix is marked as unsafe.
+/// When the right-hand side is a string, this fix can change the behavior of your program.
 /// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string,
-/// so the fix can change the behavior of your program in these cases.
 ///
-/// Additionally, if there are comments within the fix's range,
-/// it will also be marked as unsafe.
+/// Converting `in`/`not in` against a single-item container to `==`/`!=` can
+/// change runtime behavior: `in` may consider identity (e.g., `NaN`), it always
+/// yields a `bool`, and comments within the replacement range may be removed.
```
ntBre:
```suggestion
/// Additionally, converting `in`/`not in` against a single-item container to `==`/`!=` can
/// change runtime behavior: `in` may consider identity (e.g., `NaN`) and always
/// yields a `bool`. 
///
/// Comments within the replacement range will also be removed.
```

crates/ruff_linter/src/rules/refurb/rules/single_item_membership_test.rs
```diff
@@ -27,13 +27,14 @@ use crate::{Applicability, Edit, Fix, FixAvailability, Violation};
 /// ```
 ///
 /// ## Fix safety
+/// The fix is always marked as unsafe.
 ///
-/// When the right-hand side is a string, the fix is marked as unsafe.
+/// When the right-hand side is a string, this fix can change the behavior of your program.
 /// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string,
```
ntBre:
```suggestion
/// This is because `c in "a"` is true both when `c` is `"a"` and when `c` is the empty string.
```

ntBre (approved):
Thank you! Just a couple of small docs nits that I'll apply and then merge this :)

PR #20283, opened 2025-09-07
[ty][WIP] make narrowing for enums

Renkai:
Finished todos left in https://github.com/astral-sh/ruff/pull/20164


Renkai:
though it works, it makes too many differences than I expected, can't vibe out a simpler one yet, will archive it first.

PR #20284, opened 2025-09-07
[ty] Treat `Hashable`, and similar protocols, equivalently to `object` for subtyping/assignability

AlexWaygood:

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1349,6 +1349,9 @@ impl<'db> Type<'db> {
             (_, Type::NominalInstance(instance)) if instance.is_object(db) => {
                 C::always_satisfiable(db)
             }
+            (_, Type::ProtocolInstance(_)) if target.normalized(db).is_object(db) => {
+                C::always_satisfiable(db)
+            }
 
             // `Never` is the bottom type, the empty set.
             // It is a subtype of all other types.
--- a/crates/ty_python_semantic/resources/mdtest/mro.md
+++ b/crates/ty_python_semantic/resources/mdtest/mro.md
@@ -165,7 +165,9 @@ from does_not_exist import DoesNotExist  # error: [unresolved-import]
 reveal_type(DoesNotExist)  # revealed: Unknown
 
 if hasattr(DoesNotExist, "__mro__"):
-    reveal_type(DoesNotExist)  # revealed: Unknown & <Protocol with members '__mro__'>
+    # TODO: this should be `Unknown & <Protocol with members '__mro__'>` or similar
+    # (The second part of the intersection is incorrectly simplified to `object` due to https://github.com/astral-sh/ty/issues/986)
+    reveal_type(DoesNotExist)  # revealed: Unknown
 
     class Foo(DoesNotExist): ...  # no error!
     reveal_type(Foo.__mro__)  # revealed: tuple[<class 'Foo'>, Unknown, <class 'object'>]
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1024,6 +1024,38 @@ static_assert(not is_assignable_to(HasX, Foo))
 static_assert(not is_subtype_of(HasX, Foo))
 ```
 
+Since `object` defines a `__hash__` method, this means that the standard-library `Hashable` protocol
+is currently understood by ty as being equivalent to `object`, much like `SupportsStr` and
+`UniversalSet` above:
+
+```py
+from typing import Hashable
+
+static_assert(is_equivalent_to(object, Hashable))
+static_assert(is_assignable_to(object, Hashable))
+static_assert(is_subtype_of(object, Hashable))
+```
+
+This means that any type considered assignable to `object` (which is all types) is considered by ty
+to be assignable to `Hashable`. This avoids false positives on code like this:
+
+```py
+def takes_hashable_or_sequence(x: Hashable | list[Hashable]): ...
+
+takes_hashable_or_sequence(["foo"])  # fine
+takes_hashable_or_sequence(None)  # fine
+```
+
+but means that ty currently does not detect errors on code like this, which is flagged by other type
+checkers:
+
+```py
+def needs_something_hashable(x: Hashable):
+    hash(x)
+
+needs_something_hashable([])
+```
+
 ## Diagnostics for protocols with invalid attribute members
 
 This is a short appendix to the previous section with the `snapshot-diagnostics` directive enabled


AlexWaygood:
Primer runs are currently failing due to GitLab being down for some planned maintenance, meaning some primer projects cannot be cloned... https://status.gitlab.com/

crates/ty_python_semantic/src/types.rs
```diff
@@ -1349,6 +1349,13 @@ impl<'db> Type<'db> {
             (_, Type::NominalInstance(instance)) if instance.is_object(db) => {
                 C::always_satisfiable(db)
             }
+            (_, Type::ProtocolInstance(target))
```
```suggestion

            // Any structural types that are "supertypes of `object`"
            // are in fact equivalent to `object`, and should therefore
            // be treated equivalently to `object` when it comes to assignability/subtyping
            (_, Type::ProtocolInstance(target))
```

AlexWaygood:
> ```diff
> trio (https://github.com/python-trio/trio)
> + src/trio/_deprecate.py:49:19: error[unresolved-attribute] Type `<Protocol with members '__qualname__'>` has no attribute `__module__`
> ```

This seems to be a pre-existing bug where we don't recognise attributes available on `object` as being available on synthesized protocols... I'm attempting to fix it in https://github.com/astral-sh/ruff/pull/20286, but not totally sure what the correct fix is.

AlexWaygood:
Applying this diff locally buys back basically all the lost performance on the incremental benchmark:

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 46aab242b2..fdaebb0bf9 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1350,21 +1350,19 @@ impl<'db> Type<'db> {
                 C::always_satisfiable(db)
             }
 
-            // Any structural types that are "supertypes of `object`"
-            // are in fact equivalent to `object`, and should therefore
-            // be treated equivalently to `object` when it comes to assignability/subtyping
-            (_, Type::ProtocolInstance(target))
-                if Type::object(db)
-                    .satisfies_protocol(db, target, relation, visitor)
-                    .is_always_satisfied(db) =>
-            {
-                C::always_satisfiable(db)
-            }
-
             // `Never` is the bottom type, the empty set.
             // It is a subtype of all other types.
             (Type::Never, _) => C::always_satisfiable(db),
 
+            // `Any` is only a subtype of `object` -- but if a protocol's interface can be satisfied by `object`,
+            // then `object` is equivalent to the protocol, meaning that `Any` can also be a subtype of that
+            // equivalent-to-object protocol.
+            (Type::Dynamic(_), Type::ProtocolInstance(proto)) => {
+                C::from_bool(db, relation.is_assignability()).or(db, || {
+                    Type::object(db).satisfies_protocol(db, proto, relation, visitor)
+                })
+            }
+
             // Dynamic is only a subtype of `object` and only a supertype of `Never`; both were
             // handled above. It's always assignable, though.
             (Type::Dynamic(_), _) | (_, Type::Dynamic(_)) => {
diff --git a/crates/ty_python_semantic/src/types/instance.rs b/crates/ty_python_semantic/src/types/instance.rs
index 8714c824aa..c278b181d9 100644
--- a/crates/ty_python_semantic/src/types/instance.rs
+++ b/crates/ty_python_semantic/src/types/instance.rs
@@ -114,6 +114,13 @@ impl<'db> Type<'db> {
             .when_all(db, |member| {
                 member.is_satisfied_by(db, self, relation, visitor)
             })
+            .or(db, || {
+                let object = Type::object(db);
+                if self == object {
+                    return C::unsatisfiable(db);
+                }
+                object.satisfies_protocol(db, protocol, relation, visitor)
+            })
     }
 }
 
@@ -524,13 +531,16 @@ impl<'db> ProtocolInstanceType<'db> {
         self,
         db: &'db dyn Db,
         other: Self,
-        _relation: TypeRelation,
+        relation: TypeRelation,
         visitor: &HasRelationToVisitor<'db, C>,
     ) -> C {
         other
             .inner
             .interface(db)
             .is_sub_interface_of(db, self.inner.interface(db), visitor)
+            .or(db, || {
+                Type::object(db).satisfies_protocol(db, other, relation, visitor)
+            })
     }
```

But I'm really not a fan of how repetitive that makes the logic -- it means encoding in three separate places that protocols which are supertypes of `object` should be treated equivalently to `object` :-(

carljm:
Another option that might buy back the performance also is to eagerly simplify when constructing the protocol type (as we elsewhere prefer to do with equivalent types when we can), so that an annotation of `Hashable` is eagerly translated to `object`, and we don't have to keep checking that `object` satisfies it on any subtype check. That has the downside that if you have `x: Hashable` and then `reveal_type(x)`, it'll say `object` -- but this is just accurately reflecting our understanding of the type.

There's probably yet another option where we keep "equivalent to object" as a special flag on the protocol struct, and just check it once when the protocol is constructed rather than on every subtype check.

Maybe those won't actually be net positive for performance, if there are lots of protocol types constructed but then never actually checked for assignability or subtyping?

I think if those aren't good options, it's probably worth doing the repetitive version here (with comments cross-referencing the location). The regression is only >5% on the incremental benchmark, but it shows up as 1-2% across the board. Having to check a protocol against object literally every time it is used in a subtype or assignability check is pretty significant.

carljm:
Actually there might be another option here that gets the best of both worlds performance-wise, plus the DRYness -- if we make `is_equivalent_to_object` a cached Salsa query on a `ProtocolInstance`?

AlexWaygood:
> Another option that might buy back the performance also is to eagerly simplify when constructing the protocol type (as we elsewhere prefer to do with equivalent types when we can), so that an annotation of `Hashable` is eagerly translated to `object`, and we don't have to keep checking that `object` satisfies it on any subtype check. That has the downside that if you have `x: Hashable` and then `reveal_type(x)`, it'll say `object` -- but this is just accurately reflecting our understanding of the type.

Yeah, I considered that, but I think it'll just be a really confusing user experience if we produce error messages like this:

```py
from typing import Hashable

def f(x: Hashable):
    x.foo  # error: object of type `object` has no attribute `foo`
```

From a user's perspective: why on earth are we talking about `object` there??

AlexWaygood:
> Actually there might be another option here that gets the best of both worlds performance-wise, plus the DRYness -- if we make `is_equivalent_to_object` a cached Salsa query on a `ProtocolInstance`?

Nice, that seems to have done the trick! We're now seeing speedups on some of the benchmarks 🥳

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -482,6 +482,43 @@ impl<'db> ProtocolInstanceType<'db> {
...
+            _result: &bool,
+            _count: u32,
+            _value: ProtocolInstanceType<'db>,
+            _: (),
+        ) -> salsa::CycleRecoveryAction<bool> {
+            salsa::CycleRecoveryAction::Iterate
+        }
+
+        fn initial<'db>(_db: &'db dyn Db, _value: ProtocolInstanceType<'db>, _: ()) -> bool {
+            false
```
carljm: Probably doesn't matter in practice, but I suspect this should be `true`, or we will fail to recognize some recursive protocols as equivalent to `object` when in fact they should be. The co-inductive assumption here should be `true` -- if there were any part of the protocol that `object` does not satisfy, we would still correctly end up with an overall `false`.

AlexWaygood: Hmm I'm not sure there *are* any possible recursive protocols that would be equivalent to `object`? But I haven't tried very hard to think of any 😄

carljm: If you take any protocol that is equivalent to object, and replace one of the types of its members with itself, I think that would do it?

AlexWaygood: Makes sense!

AlexWaygood: I added a regression test for recursive protocols that are supertypes of `object`. It didn't fail with the code as I had it before, but that may be just because we still don't look at the types of methods when doing assignability/subtyping checks for protocols (and this PR blocks doing that!). And the regression tests seem like a useful edge case to test explicitly, anyway.

I switched this to `true`, as per your suggestion!

PR #20285, opened 2025-09-07
[ty] equality narrowing on enums that don't override `__eq__` or `__ne__`

Renkai:
It's a TODO clean of PR https://github.com/astral-sh/ruff/pull/20164 that solved issue https://github.com/astral-sh/ty/issues/939


sharkdp:
Thank you for your contribution. It looks like a better format here would be a GitHub issue on the https://github.com/astral-sh/ty repo? Or is there a particular reason that you are opening a PR at this stage?

Renkai:
Hi, @sharkdp I want to clean some todos added in this PR:https://github.com/astral-sh/ruff/pull/20164
It's not finished yet so I mark this as `draft` in github. do you think we shall create an issue first? I'm not quite familiar with the best practice yet.

sharkdp:
If you plan to do more changes in this draft PR, that's totally fine. I just thought you were looking for feedback on that document you added here. In that case, I'll assume that we wait for this to move to in-review.

AlexWaygood:
Thanks! Could you possibly update your PR description to describe this change a little better?

crates/ty_python_semantic/src/types.rs
```diff
@@ -740,6 +740,43 @@ impl<'db> Type<'db> {
...
+                        | MemberLookupPolicy::MRO_NO_INT_OR_STR_LOOKUP,
+                );
+                let call_result = call_result.as_ref();
+                call_result.is_ok_and(|bindings| {
+                    bindings.return_type(db) == Type::BooleanLiteral(allowed_return_value)
+                }) || call_result
+                    .is_err_and(|err| matches!(err, CallDunderError::MethodNotAvailable))
+            };
+
+            check_dunder("__eq__", true) && check_dunder("__ne__", false)
```
carljm: We shouldn't duplicate all this code from `is_single_valued`; we should extract it as a separate method, e.g. `Type::overrides_equality`. (It doesn't even have to be specific to enums, even if we only call it for enums currently.)

crates/ty_python_semantic/src/types.rs
```diff
@@ -740,6 +740,43 @@ impl<'db> Type<'db> {
             .is_some_and(|instance| instance.class(db).is_known(db, KnownClass::Bool))
     }
 
+    /// Return true if this is an enum literal or enum class instance that doesn't override __eq__ or __ne__
+    fn is_simple_enum(&self, db: &'db dyn Db) -> bool {
+        let is_enum = match self {
+            Type::EnumLiteral(_) => true,
```
carljm: No mdtests fail if we remove this arm. I think that's because we only use this method as an additional check beyond `is_single_valued`, and enum literals are already always single-valued.

I think we can probably remove this `is_simple_enum` method altogether in favor of an `is_enum` method, and use `ty.is_enum() && !ty.overrides_equality()` everywhere the PR currently uses `ty.is_simple_enum()`. This isn't quite as concise but I think it's clearer, as its meaning is more self-evident, where the meaning of "simple enum" is not.

Renkai: Thanks for the review! I fixed it followed by your instruction, but cargo clippy forced me to use `!element.is_enum(self.db) || element.overrides_equality(self.db)` instead, hope it's still good to you.

PR #20289, opened 2025-09-07
[pycodestyle] Allow gi.require_version[s]() interspersed with imports, for PyGObject (E402) (fixes #20288)

### LINKED ISSUE #20288: E402: Exempt `gi.require_version[s]()` preceding imports
### Summary

Code written using [PyGObject](https://pygobject.gnome.org/) that imports versioned GObject-Introspection (`gi`) typelibs requires version-selection calls to precede the actual module imports, e.g.
    
```python
import gi

# Either    
gi.require_version('Gtk', '4.0')
gi.require_versions({'GObject': '2.0', 'Gio': '2.0'})

from gi.repository import GObject, Gio, Gtk
```
    
Because these calls are necessary and unavoidable with PyGObject, a narrow exception to rule E402 would be valuable.

**Note:** I already have a pull request ready to go that addresses this, but I wanted to formalize the need/request with an issue first.

ferdnyc:
## Summary

Exempt `gi.require_version()` and `gi.require_versions()` calls from triggering E402 if they appear interspersed with imports, because they are necessary version-selection calls required before importing from `gi.repository`, when working with PyGObject / GIRepository code.

This places `gi.require_version[s]()` calls in the same category as:

* `sys.path` modifications
* `os.environ` modifications
* `site.addsitedir()` calls

And, most similar in both nature and niche-ness:

* `pytest.importorskip()` calls
* `matplotlib.use()` calls
 
All of which are exempted from E402 in Ruff.

Fixes #20288 

## Test Plan

The PR contains a new test fixture `crates/ruff_linter/resources/test/fixtures/pycodestyle/E402_6.py`.

The sample code contained within triggers E402 without the changes in this PR.

With the PR changes in place, the corresponding test case (added to `crates/ruff_linter/src/rules/pycodestyle/mod.rs`) and `.snap` file result in a passing test.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pycodestyle/E402_6.py
+++ b/crates/ruff_linter/resources/test/fixtures/pycodestyle/E402_6.py
@@ -1,0 +1,7 @@
+# Issue: https://github.com/astral-sh/ruff/issues/20288
+import gi
+
+gi.require_version('Gtk', '4.0')
+gi.require_versions({'GObject': '2.0', 'Gio': '2.0'})
+
+from gi.repository import GObject, Gio, Gtk
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -842,7 +842,8 @@ impl<'a> Visitor<'a> for Checker<'a> {
                     || imports::is_sys_path_modification(stmt, self.semantic())
                     || imports::is_os_environ_modification(stmt, self.semantic())
                     || imports::is_pytest_importorskip(stmt, self.semantic())
-                    || imports::is_site_sys_path_modification(stmt, self.semantic()))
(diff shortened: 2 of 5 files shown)


ferdnyc:
### Notes

1. This PR doesn't currently limit the change to the `preview` feature, although I notice many PRs do. There wasn't any guidance in the contributing guide about that, so I decided I'd wait for guidance.
2. The PR also doesn't add any mention of the new exemption to the documentation. I noticed that the docs mention the "major" exemptions like `sys.path` and `os.environ`, but don't go into detail on the more focused ones like `pytest.importorskip()` and `matplotlib.use()`, so I followed suit.

MichaReiser:
Thank you for your contribution.

As discussed on the issue. I don't think we want to keep adding more library-specific exceptions. Mainly because it is very slippery terrain, where each exception justifies adding the next exception. 

Let's first decide on a path forward on the issue.

Issue #19962, opened 2025-08-18
`--output-format sarif` does not provide information on autofixes


alexanderlinne:
### Summary

When using `uv run ruff check --output-format json .`, the output for an issue contains a `fix` field if an autofix is available. When using `uv run ruff check --output-format sarif .`, the `results` however do not contain the autofix as specified in [the SARIF spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541319).

### Version

ruff 0.12.9

MichaReiser:
This should be relatively straightforward to add. I'll be happy to accept a contribution

nitram509:
Since this was fixed with #20300
the issue can be closed, yes?

ntBre:
Looks like it, thank you!

PR #20300, opened 2025-09-08
Extended `SarifEmitter` to include `fixes` in the SARIF ouput

njhearp:
## Summary

Addresses #19962

This PR extends `SarifEmitter` to include `fixes` in the SARIF ouput, if an auto fix is available. Before SARIF output did not include the `fixes` field. This makes Ruff's SARIF output more compliant with the spec.

Key changes:
- `SarifFix`, `SarifArtifactChange`, and related structs to represent fixes in SARIF.
- `fixes` is outputted only if an auto fix is available.

## Test Plan

Check for SARIF spec compliance and regressions.

### DIFF
--- a/crates/ruff_linter/src/message/sarif.rs
+++ b/crates/ruff_linter/src/message/sarif.rs
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
     start_column: OneIndexed,
     end_line: OneIndexed,
     end_column: OneIndexed,
+    fixes: Vec<SarifFix>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifFix {
+    description: RuleDescription,
+    #[serde(rename = "artifactChanges")]
+    artifact_changes: Vec<SarifArtifactChange>,
+}
+
+#[derive(Debug, Serialize)]
+
+struct RuleDescription {
+    text: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifArtifactChange {
+    #[serde(rename = "artifactLocation")]
+    uri: SarifArtifactLocation,
+    replacements: Vec<SarifReplacement>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifArtifactLocation {
+    uri: String,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifReplacement {
+    #[serde(rename = "deletedRegion")]
+    deleted_region: SarifRegion,
+    #[serde(rename = "insertedContent")]
+    inserted_content: InsertedContent,
+}
+
+#[derive(Debug, Serialize)]
+struct InsertedContent {
+    #[serde(rename = "text")]
+    inserted_content: String,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifRegion {
+    #[serde(rename = "startLine")]
+    start_line: OneIndexed,
+    #[serde(rename = "startColumn")]
+    start_column: OneIndexed,
+    #[serde(rename = "endLine")]
+    end_line: OneIndexed,
+    #[serde(rename = "endColumn")]
+    end_column: OneIndexed,
 }
 
 impl<'a> SarifResult<'a> {
@@ -161,17 +213,60 @@ impl<'a> SarifResult<'a> {
         let start_location = message.expect_ruff_start_location();
         let end_location = message.expect_ruff_end_location();
         let path = normalize_path(&*message.expect_ruff_filename());
+        let uri = url::Url::from_file_path(&path)
+            .map_err(|()| anyhow::anyhow!("Failed to convert path to URL: {}", path.display()))?
+            .to_string();
+        let fixes = message
+            .fix()
+            .map(|fix| {
+                let fix_description = message
+                    .first_help_text()
+                    .map(std::string::ToString::to_string);
+
+                let replacements: Vec<SarifReplacement> = fix
+                    .edits()
+                    .iter()
+                    .map(|edit| SarifReplacement {
+                        deleted_region: SarifRegion {
+                            start_line: start_location.line,
+                            start_column: start_location.column,
+                            end_line: end_location.line,
+                            end_column: end_location.column,
+                        },
+                        inserted_content: InsertedContent {
+                            inserted_content: edit.content().unwrap_or("").to_string(),
+                        },
+                    })
+                    .collect();
+
+                let mut artifact_changes = Vec::new();
+                if !replacements.is_empty() {
+                    artifact_changes.push(SarifArtifactChange {
+                        uri: SarifArtifactLocation { uri: uri.clone() },
+                        replacements,
+                    });
+                }
+
+                SarifFix {
+                    description: RuleDescription {
+                        text: fix_description,
+                    },
+                    artifact_changes,
+                }
+            })
+            .into_iter()
+            .collect();
+
         Ok(Self {
             code: RuleCode::from(message),
             level: "error".to_string(),
             message: message.body().to_string(),
-            uri: url::Url::from_file_path(&path)
-                .map_err(|()| anyhow::anyhow!("Failed to convert path to URL: {}", path.display()))?
-                .to_string(),
+            uri,
             start_line: start_location.line,
             start_column: start_location.column,
             end_line: end_location.line,
             end_column: end_location.column,
+            fixes,
         })
     }
 
@@ -199,7 +294,7 @@ impl Serialize for SarifResult<'_> {
     where
         S: Serializer,
     {
-        json!({
+        let mut value = json!({
             "level": self.level,
             "message": {
                 "text": self.message,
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+    uri: SarifArtifactLocation,
+    replacements: Vec<SarifReplacement>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifArtifactLocation {
+    uri: String,
+}
+
+#[derive(Debug, Serialize)]
```
MichaReiser, line 222: You can use `serde(rename_all="camelCase")` to avoid having to rename every field. See https://serde.rs/container-attrs.html#rename_all

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+
+struct RuleDescription {
+    text: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifArtifactChange {
+    #[serde(rename = "artifactLocation")]
+    uri: SarifArtifactLocation,
```
MichaReiser: I'd prefer if the fields use the same names as in the Sarif specification. It makes it easier to match the fields to their output.

```suggestion
    #[serde(rename = "artifactLocation")]
    artifactLocation: SarifArtifactLocation,
```

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
     start_column: OneIndexed,
     end_line: OneIndexed,
     end_column: OneIndexed,
+    fixes: Vec<SarifFix>,
+}
+
+#[derive(Debug, Serialize)]
```
MichaReiser, line 177: Unrelated to your PR: It would be great to include the link to `https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541319` somewhere in the documentation (on `SarifResult` or on `SarifEmitter`)

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+struct SarifFix {
+    description: RuleDescription,
+    #[serde(rename = "artifactChanges")]
+    artifact_changes: Vec<SarifArtifactChange>,
+}
+
+#[derive(Debug, Serialize)]
+
```
MichaReiser:
```suggestion
```

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+    #[serde(rename = "deletedRegion")]
+    deleted_region: SarifRegion,
+    #[serde(rename = "insertedContent")]
+    inserted_content: InsertedContent,
+}
+
+#[derive(Debug, Serialize)]
+struct InsertedContent {
+    #[serde(rename = "text")]
+    inserted_content: String,
```
MichaReiser:
```suggestion
    text: String,
```

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+struct SarifArtifactLocation {
+    uri: String,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifReplacement {
+    #[serde(rename = "deletedRegion")]
+    deleted_region: SarifRegion,
+    #[serde(rename = "insertedContent")]
+    inserted_content: InsertedContent,
```
MichaReiser: This should be an `Option` and be omitted when it is `None` (e.g. for a removal)

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -153,6 +153,58 @@ struct SarifResult<'a> {
...
+#[derive(Debug, Serialize)]
+struct SarifArtifactChange {
+    #[serde(rename = "artifactLocation")]
+    uri: SarifArtifactLocation,
+    replacements: Vec<SarifReplacement>,
+}
+
+#[derive(Debug, Serialize)]
+struct SarifArtifactLocation {
+    uri: String,
```
MichaReiser, line 219: Should this be an `Url`?

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -161,17 +213,60 @@ impl<'a> SarifResult<'a> {
...
+                    })
+                    .collect();
+
+                let mut artifact_changes = Vec::new();
+                if !replacements.is_empty() {
+                    artifact_changes.push(SarifArtifactChange {
+                        uri: SarifArtifactLocation { uri: uri.clone() },
+                        replacements,
+                    });
+                }
```
MichaReiser: A fix is guaranteed (statically) to always have at least one edit
```suggestion
                
                artifact_changes.push(SarifArtifactChange {
                    uri: SarifArtifactLocation { uri: uri.clone() },
                    replacements,
                });
                
```

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -181,15 +276,58 @@ impl<'a> SarifResult<'a> {
...
+
+                SarifFix {
+                    description: RuleDescription {
+                        text: fix_description,
+                    },
+                    artifact_changes,
+                }
+            })
+            .into_iter()
+            .collect();
```
MichaReiser: Let's extract this into a function to avoid repeating the same code twice

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -218,8 +356,11 @@ impl Serialize for SarifResult<'_> {
...
             }],
             "ruleId": self.code.as_str(),
-        })
-        .serialize(serializer)
+        });
+        if !self.fixes.is_empty() {
+            value["fixes"] = serde_json::to_value(&self.fixes).unwrap();
+        }
+        value.serialize(serializer)
     }
```
MichaReiser: Hmm, interesting. Can't we derive `Serialize` on `SarifResult` instead and skip serializing `fixes` if the container is empty (https://serde.rs/field-attrs.html#skip_serializing_if)?

crates/ruff_linter/src/message/sarif.rs
njhearp: I think the spec says `uri`. https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540865

crates/ruff_linter/src/message/sarif.rs
njhearp: Makes sense. I've changed it.

crates/ruff_linter/src/message/sarif.rs
njhearp: That's a great suggestion to make it more concise. I've changed this.

crates/ruff_linter/src/message/sarif.rs
njhearp: Thanks. I've updated it.

crates/ruff_linter/src/message/sarif.rs
njhearp: I had to do some additional changes and I've encountered a merge conflict from main. I'll work on fixing the merge conflict.

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -143,83 +157,163 @@ impl<'a> From<&'a Diagnostic> for RuleCode<'a> {
...
+                    .map(std::string::ToString::to_string);
+
+                let replacements: Vec<SarifReplacement> = fix
+                    .edits()
+                    .iter()
+                    .map(|edit| SarifReplacement {
+                        deleted_region,
+                        inserted_content: Some(InsertedContent {
+                            text: edit.content().unwrap_or("").to_string(),
+                        }),
```
MichaReiser: I think there are two issues with how we map edits:

1. `inserted_content` should be `None` if `edit.content()` is `None`. 
2. We need to use the range from the edit for the deleted range and not the diagnostic range. You can get the range by calling `edit.range()`. The range is guaranteed to be empty if it is an insertion. It's non empty for replacements or deletions

njhearp:
Sorry for the noisy update, I accidentally pushed extra commits when merging. I'm going to fix this to remove the noise.

crates/ruff_linter/src/message/sarif.rs
```diff
@@ -143,83 +158,189 @@ impl<'a> From<&'a Diagnostic> for RuleCode<'a> {
...
+            let start_location = source_code.line_column(range.start());
+            let end_location = source_code.line_column(range.end());
+
+            SarifRegion {
+                start_line: start_location.line,
+                start_column: start_location.column,
+                end_line: end_location.line,
+                end_column: end_location.column,
+            }
+        } else {
```
njhearp: I thought I would add error handling if there is a problem with the source file.

MichaReiser:
> Sorry for the noisy update, I accidentally pushed extra commits when merging. I'm going to fix this to remove the noise.

Don't worry. It happened to all of us :)

crates/ruff_linter/src/message/sarif.rs
MichaReiser: That makes sense. I changed the error handling to omit the fix when the source file is missing. Emitting a fix with made-up locations seems problematic, as it most likely will result in invalid code when applied.

MichaReiser (approved):

PR #20301, opened 2025-09-08
[`pep8-naming`] Fix formatting of `__all__` (`N816`)

onerandomusername:
## Summary

Noticed this was not escaped when writing a project that parses the result of `ruff rule --outputformat json`. This is visible here: <https://docs.astral.sh/ruff/rules/mixed-case-variable-in-global-scope/#why-is-this-bad>

## Test Plan

documentation only


crates/ruff_linter/src/rules/pep8_naming/rules/mixed_case_variable_in_global_scope.rs
```diff
@@ -19,7 +19,7 @@ use crate::rules::pep8_naming::helpers;
 /// > only.) The conventions are about the same as those for functions.
 /// >
 /// > Modules that are designed for use via from M import * should use the
-/// > __all__ mechanism to prevent exporting globals, or use the older
+/// > `__all__` mechanism to prevent exporting globals, or use the older
```
ntBre, line 22: Should we add code formatting to `from M import *` in the line above while we're here?

PR #18007, opened 2025-05-10
[ty] Assume type of self is `typing.Self` in method calls

Glyphack:
## Summary

Part of https://github.com/astral-sh/ty/issues/159

This PR only adjusts the signature of a method so if it has a `self` argument then that argument will have type of `Typing.Self` even if it's not specified. If user provides an explicit annotation then Ty will not override that annotation.

## Follow Ups

- https://github.com/astral-sh/ty/issues/1131
- https://github.com/astral-sh/ty/issues/1157
- https://github.com/astral-sh/ty/issues/1156
- https://github.com/astral-sh/ty/issues/1173
- https://github.com/astral-sh/ruff/pull/20328
- https://github.com/astral-sh/ty/issues/1163
- https://github.com/astral-sh/ty/issues/1196

## Test Plan

Added mdtests.
Also some tests need https://github.com/astral-sh/ruff/pull/18473 to work completely. So I added a todo for those new cases that I added.

PR #20302, opened 2025-09-08
[ty] Fix signature of `NamedTupleLike._make`

sharkdp:


AlexWaygood:
oops, yes, it was meant to be `cls: type[Self]`. Another data point for the argument that we should try to do as little special-casing as possible, and rely on typeshed as much as we can...

crates/ty_vendored/ty_extensions/ty_extensions.pyi
```diff
@@ -93,7 +93,7 @@ class NamedTupleLike(Protocol):
     _field_defaults: ClassVar[dict[str, Any]]
     _fields: ClassVar[tuple[str, ...]]
     @classmethod
-    def _make(self: Self, iterable: Iterable[Any]) -> Self: ...
+    def _make(cls, iterable: Iterable[Any]) -> Self: ...
```
AlexWaygood: this also would make sense for now, I think? But I'm also fine with what you have if #18007 is _just_ about to be merged ;)

```suggestion
    def _make(cls: type[Self], iterable: Iterable[Any]) -> Self: ...
```

AlexWaygood: I guess it's maybe a bit of a moot point while https://github.com/astral-sh/ty/issues/501 is still unimplemented, though

sharkdp: Changed to an explicit `type[Self]` for now.

PR #20304, opened 2025-09-08
[ty] Support "legacy" `typing.Self` in combination with PEP 695 generic contexts

sharkdp:
## Summary

Support cases like the following, where we need the generic context to include both `Self` and `T` (not just `T`):

```py
from typing import Self

class C:
    def method[T](self: Self, arg: T): ...

C().method(1)
```

closes https://github.com/astral-sh/ty/issues/1131

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/src/types/signatures.rs
+++ b/crates/ty_python_semantic/src/types/signatures.rs
@@ -376,12 +376,25 @@ impl<'db> Signature<'db> {
         let legacy_generic_context =
             GenericContext::from_function_params(db, definition, &parameters, return_ty);
 
-        if generic_context.is_some() && legacy_generic_context.is_some() {
-            // TODO: Raise a diagnostic!
-        }
+        let full_generic_context = match (legacy_generic_context, generic_context) {
+            (Some(legacy_ctx), Some(ctx)) => {
+                if legacy_ctx
+                    .variables(db)
+                    .iter()
+                    .exactly_one()
+                    .is_ok_and(|bound_typevar| bound_typevar.typevar(db).is_self(db))
+                {
+                    Some(legacy_ctx.merge(db, ctx))
+                } else {
+                    // TODO: Raise a diagnostic — mixing PEP 695 and legacy typevars is not allowed
+                    Some(ctx)
+                }
+            }
+            (left, right) => left.or(right),
+        };
 
         Self {
-            generic_context: generic_context.or(legacy_generic_context),
+            generic_context: full_generic_context,
             inherited_generic_context,
             definition: Some(definition),
             parameters,
--- a/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
@@ -494,3 +494,25 @@ age, name = team.employees[0]
 reveal_type(age)  # revealed: Age
 reveal_type(name)  # revealed: Name
 ```
+
+## `self` in PEP 695 generic methods
+
+When a generic method uses a PEP 695 generic context, an implict or explicit annotation of
+`self: Self` is still part of the full generic context:
+
+```py
+from typing import Self
+
+class C:
+    def explicit_self[T](self: Self, x: T) -> tuple[Self, T]:
+        return self, x
+
+    def implicit_self[T](self, x: T) -> tuple[Self, T]:
+        return self, x
+
+def _(x: int):
+    reveal_type(C().explicit_self(x))  # revealed: tuple[C, int]
+
+    # TODO: this should be `tuple[C, int]` as well, once we support implicit `self`
+    reveal_type(C().implicit_self(x))  # revealed: tuple[Unknown, int]
+```
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -134,6 +134,18 @@ impl<'db> GenericContext<'db> {
         Self::new(db, type_params.into_iter().collect::<FxOrderSet<_>>())
     }
 
+    /// Merge this generic context with another, returning a new generic context that
+    /// contains type variables from both contexts.
+    pub(crate) fn merge(self, db: &'db dyn Db, other: Self) -> Self {
+        Self::from_typevar_instances(
+            db,
+            self.variables(db)
+                .iter()
+                .chain(other.variables(db).iter())
+                .copied(),
+        )
+    }
+
     fn variable_from_type_param(
         db: &'db dyn Db,
         index: &'db SemanticIndex<'db>,


crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -376,12 +376,13 @@ impl<'db> Signature<'db> {
         let legacy_generic_context =
             GenericContext::from_function_params(db, definition, &parameters, return_ty);
 
-        if generic_context.is_some() && legacy_generic_context.is_some() {
-            // TODO: Raise a diagnostic!
```
sharkdp, line 380: The previous comment here indicates that this is probably not the right fix, but I don't see an immediate problem with merging legacy typevars and PEP 695 typevars in a single method, and it solves the problem with `self: Self`. So I'm opening this for discussion.

AlexWaygood: yes... I think it's invalid to combine legacy and PEP-695 type variables in general... but I think the sole exception to this rule is probably `Self`, which I _guess_ counts as a "legacy type variable", but is obviously different in several ways to most legacy type variables

sharkdp: Yes, https://peps.python.org/pep-0695/#compatibility-with-traditional-typevars states that this is not legal in all situations. Maybe this indicates that implicit `self` should work differently for functions with an explicit `[S, T, U, …]` PEP 695 generic context. Maybe those functions shouldn't use `typing.Self`, but rather explicitly add a new synthetic type variable to this context: `[_Self: C, S, T, U, …]`.

dcreager: The [section about `typing.Self`](https://typing.python.org/en/latest/spec/generics.html#self) doesn't seem to explicitly say one way or the other whether it's allowed in a PEP 695 generic function. I think it's worth supporting, and I think a slight modification to this approach is the right way to handle it.

Instead of merging the _entirety_ of the legacy generic context, I would suggest looking to see if it contains (only) `typing.Self`. If so, add it to the PEP 695 generic context. If not, continue to "TODO: raise a diagnostic".

(Note that we do have a distinct `TypeVarKind::TypingSelf` for `typing.Self`, so it should be easy to distinguish from "real" legacy typevars.)

AlexWaygood: > Maybe those functions shouldn't use `typing.Self`, but rather explicitly add a new synthetic type variable to this context: `[_Self: C, S, T, U, …]`.

Hmm, I don't think the intention was to deprecate `typing.Self` as part of PEP 695. I think the PEP would have stated this explicitly if that was the intention.

Adding `Self` to the type system didn't make it possible to express anything in type annotations that couldn't be expressed before (except _maybe_ in attribute annotations): it just made a common case much less verbose and much more ergonomic. Before you would have done this:

```py
from typing import TypeVar

_SelfT = TypeVar("_SelfT", bound="MyClass")

class MyClass:
    def method(self: _SelfT) -> SelfT:
        return self
```

and after the introduction of `Self`, you would do this:

```py
from typing import Self

class MyClass:
    def method(self) -> Self:
        return self
```

And the two mean exactly the same thing. The ergonomics argument for `Self` maybe isn't quite as strong for methods that use PEP-695 parameters, since there's no need to declare the type variable outside the body of the method anymore -- but I still think I'd find it a pain to have to explicitly include it in the type parameter list every time.

sharkdp: Yes, I didn't mean to imply that PEP 695 made `typing.Self` unnecessary. I was trying to illustrate the inner workings of a type checker that encounters `def pep695_method[T](self, x: T)`. And it makes sense to me that this would be syntactic sugar for `def pep695_method[_Self: MyClass, T](self: _Self, x: T)` instead of `def pep605_method[T](self: typing.Self, x: T)`, because the former wouldn't have to mix legacy and PEP 695 type variables.

dcreager: > Before you would have done this:
> 
> ```python
> from typing import TypeVar
> 
> _SelfT = TypeVar("_SelfT", bound="MyClass")
> 
> class MyClass:
>     def method(self: _SelfT) -> SelfT:
>         return self
> ```

I don't think they're equivalent, since `Self` does something more sophisticated with the bound:

```python
from typing import Self

class Base:
    def method(self: Self, other: Self) -> None: ...

class Sub(Base): ...

Sub().method(Base())  # error: [invalid-argument-type]
```

i.e. the upper bound of `Self` is that class that you access the method from (which can be different for different member accesses!), not the class where it's defined.

dcreager: (Though my point about them not being equivalent is a _stronger_ argument in favor of what you're suggesting, @AlexWaygood, which is that we should support `Self` in PEP 695 methods!)

AlexWaygood: > it makes sense to me that this would be syntactic sugar for `def pep695_method[_Self: MyClass, T](self: _Self, x: T)` instead of `def pep605_method[T](self: typing.Self, x: T)`, because the former wouldn't have to mix legacy and PEP 695 type variables.

Yeah, that makes sense to me too

AlexWaygood: > I don't think they're equivalent, since `Self` does something more sophisticated with the bound:

(We discussed this example on Discord -- other type checkers have the same behaviour on this snippet if you use a legacy TypeVar with an upper bound, and we should probably aim to have that behaviour too in the long run!)

sharkdp:
This change removes ~800 false positives across the ecosystem when based on https://github.com/astral-sh/ruff/pull/20303

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -376,12 +376,25 @@ impl<'db> Signature<'db> {
...
-        }
+        let full_generic_context = match (legacy_generic_context, generic_context) {
+            (Some(legacy_ctx), Some(ctx)) => {
+                if legacy_ctx
+                    .variables(db)
+                    .iter()
+                    .exactly_one()
```
dcreager, line 384: :sob: TIL `exactly_one`, I've been looking for an adapter like that forever!

crates/ty_python_semantic/src/types/signatures.rs
carljm: > we should probably aim to have that behaviour too in the long run

I don't think that behavior is right, either for `typing.Self` or for an explicit typevar -- discussed more on Discord.

> it makes sense to me that this would be syntactic sugar for `def pep695_method[_Self: MyClass, T](self: _Self, x: T)` instead of `def pep605_method[T](self: typing.Self, x: T)`, because the former wouldn't have to mix legacy and PEP 695 type variables.

Is this a distinction that matters in any practical way? Ultimately legacy and PEP 695 typevars are the same thing, the only difference is that they are declared differently -- and in the case of un-annotated `self`, there is no explicit declaration, so that sole difference disappears, too. And if explicit `typing.Self` annotations are allowed in a PEP 695 generic method signature, then the need to mix `typing.Self` with PEP 695 generic contexts exists regardless.

sharkdp: > Is this a distinction that matters in any practical way?

Not if `typing.Self`, `TypeVar("_SelfT", bound="MyClass")`, and `[_Self: MyClass]` all behave the same in all situations. From what I understand now, this is the case (which is a relief). I think I was maybe confused by the fact that `TypeVarKind::TypingSelf` exists, implying that it needs special treatment somehow (and @dcreager's example seemed to indicate that, but apparently that is just wrong behavior of other type checkers).

PR #20306, opened 2025-09-08
[ty] Use partial-order-friendly representation of typevar constraints

dcreager:
The constraint representation that we added in #19997 was subtly wrong, in that it didn't correctly model that type assignability is a _partial_ order — it's possible for two types to be incomparable, with neither a subtype of the other. That means the negation of a constraint like `T ≤ t` (typevar `T` must be a subtype of `t`) is **_not_** `t < T`, but rather `t < T ∨ T ≁ t` (using ≁ to mean "not comparable to").

That means we need to update our constraint representation to be an enum, so that we can track both _range_ constraints (upper/lower bound on the typevar), and these new _incomparable_ constraints.

Since we need an enum now, that also lets us simplify how we were modeling range constraints. Before, we let the lower/upper bounds be either open (<) or closed (≤). Now, range constraints are always closed, and we add a third kind of constraint for _not equivalent_ (≠). We can translate an open upper bound `T < t` into `T ≤ t ∧ T ≠ t`.

We already had the logic for doing adding _clauses_ to a _set_ by doing a pairwise simplification. We copy that over to where we add _constraints_ to a _clause_. To calculate the intersection or union of two constraints, the new enum representation makes it easy to break down all of the possibilities into a small number of cases: intersect range with range, intersect range with not-equivalent, etc. I've done the math [here](https://dcreager.net/theory/constraints/) to show that the simplifications for each of these cases is correct.


crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -48,29 +62,32 @@
...
+//! The first constraint in the union indicates that `V` can specialize to any type that is a
+//! subtype of `Never`, but not to `Never` itself.
+//!
+//! The second constraint in the union indicates that `V` can specialize to any type that is a
+//! supertype of `B` — but not to `B` itself. (For instance, it _can_ specialize to `A`.)
+//!
+//! There aren't any types that satisfy the first constraint in the union (the type would have to
+//! somehow contain a negative number of values). You can think of a constraint that cannot be
+//! satisfied as an empty set (of types), which means we can simplify it out of the union. That
+//! gives us a final constraint of `B ≤ V ∩ V ≠ B` for the negation of `T`'s constraint.
```
carljm: Doesn't this exposition still ignore the possibility that `V` is not comparable to `B` (or to `Never`, though in the particular case of `Never` that's impossible, all types are comparable to the bottom type by definition.)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -429,74 +439,116 @@ impl<'db> ConstraintClause<'db> {
...
 
+    fn is_satisfiable(&self) -> Satisfiable<()> {
+        if self.is_always() {
+            Satisfiable::Always
+        } else {
+            Satisfiable::Constrained(())
+        }
+    }
+
     /// Updates this clause to be the intersection of itself and an atomic constraint. Returns a
```
carljm: There are still many references to "atomic constraint" in doc comments, here and throughout (including on methods of `ConstrainedTypeVar` itself.) Is this language you want to keep using? I don't have a better idea, but maybe at some point it should be clarified that by "atomic constraint" you mean "ConstrainedTypeVar"?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -753,369 +805,589 @@ impl<'db> ConstraintClause<'db> {
...
+    ) -> Simplifiable<Constraint<'db>> {
+        if self.lower.is_equivalent_to(db, other.ty) || self.upper.is_equivalent_to(db, other.ty) {
+            return Simplifiable::NeverSatisfiable;
+        }
+
+        // A range constraint and a not-comparable constraint cannot be simplified.
+        Simplifiable::NotSimplified(
+            Constraint::Range(self.clone()),
+            Constraint::NotComparable(other.clone()),
+        )
```
carljm: I think more can be checked here? If we have the range constraint `L <: T <: U` and the not-comparable constraint `T nc X` (is there a standard symbol for not-comparable? doesn't seem like it), then if `U <: X` or `X <: L`, by transitivity of subtyping `T <: X` or `X <: T`, meaning the intersection is never satisfiable.

IOW I think the `is_equivalent_to` checks here should be `is_subtype_of` checks instead.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -753,369 +805,589 @@ impl<'db> ConstraintClause<'db> {
...
+    }
+
+    fn intersect_not_equivalent(
+        &self,
+        db: &'db dyn Db,
+        other: &NotEquivalentConstraint<'db>,
+    ) -> Simplifiable<Constraint<'db>> {
+        // If the range constraint says that the typevar must be equivalent to some type, and the
+        // not-equivalent type says that it must not, we have a contradiction.
+        if self.lower.is_equivalent_to(db, self.upper) && self.lower.is_equivalent_to(db, other.ty)
```
carljm, line 1034: What about the case where `other.ty` is neither a supertype of `self.lower` nor a subtype of `self.upper` -- therefore it's outside the range, and the intersection can simplify to just the range?

crates/ty_python_semantic/src/types/constraints.rs
dcreager: > is there a standard symbol for not-comparable? doesn't seem like it

I've been using ≁ (`\u2241`). I have a pretty [heavily customized](https://github.com/dcreager/dotfiles-public/blob/d98071250a4328504c6b6ed3275c157608b5e535/.XCompose.symlink#L25) Compose key on my linux boxes; not sure how to type that ergonomically on macos

dcreager: Yes, good catch! This was bad copy/pasta from the method above, I had subtype checks as you suggest when I did this in the math! [[lower bound](https://dcreager.net/theory/constraints/intersect-lower-ncmp/), [upper bound](https://dcreager.net/theory/constraints/intersect-upper-ncmp/)]

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Nice! It [actually works](https://dcreager.net/theory/constraints/intersect-range-neq/#Case-2) if _either_ `L ⊈ X` or `X ⊈ U`. Updated

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Reworded to use "individual constraint", which I was using in a couple of places already

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Yes, good catch! I had updated this to remove the open/closed distinction.  Added text about incomparable

dcreager:
> I'm sure if I tell you that all this math seems well-suited to a nice unit test suite, it won't be the first you've thought of it ;)

I'm tackling this in a separate PR, https://github.com/astral-sh/ruff/pull/20319

PR #20309, opened 2025-09-08
Refactor diagnostic start|end location helpers

amyreese:
- Renames functions to drop `expect_` from names.
- Make functions return `Option<LineColumn>` to appropriately signal
  when range is not available.
- Update existing consumers to use `unwrap_or_default()`. Uncertain if
  there are better fallback behaviors for individual consumers.

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -454,24 +454,26 @@ impl Diagnostic {
 
     /// Computes the start source location for the message.
     ///
-    /// Panics if the diagnostic has no primary span, if its file is not a `SourceFile`, or if the
-    /// span has no range.
-    pub fn expect_ruff_start_location(&self) -> LineColumn {
-        self.expect_primary_span()
-            .expect_ruff_file()
-            .to_source_code()
-            .line_column(self.expect_range().start())
+    /// Returns None if the diagnostic has no primary span, if its file is not a `SourceFile`,
+    /// or if the span has no range.
+    pub fn ruff_start_location(&self) -> Option<LineColumn> {
+        Some(
+            self.ruff_source_file()?
+                .to_source_code()
+                .line_column(self.range()?.start()),
+        )
     }
 
     /// Computes the end source location for the message.
     ///
-    /// Panics if the diagnostic has no primary span, if its file is not a `SourceFile`, or if the
-    /// span has no range.
-    pub fn expect_ruff_end_location(&self) -> LineColumn {
-        self.expect_primary_span()
-            .expect_ruff_file()
-            .to_source_code()
-            .line_column(self.expect_range().end())
+    /// Returns None if the diagnostic has no primary span, if its file is not a `SourceFile`,
+    /// or if the span has no range.
+    pub fn ruff_end_location(&self) -> Option<LineColumn> {
+        Some(
+            self.ruff_source_file()?
+                .to_source_code()
+                .line_column(self.range()?.end()),
+        )
     }
 
     /// Returns the [`SourceFile`] which the message belongs to.
--- a/crates/ruff_linter/src/message/github.rs
+++ b/crates/ruff_linter/src/message/github.rs
@@ -19,7 +19,7 @@ impl Emitter for GithubEmitter {
         context: &EmitterContext,
     ) -> anyhow::Result<()> {
         for diagnostic in diagnostics {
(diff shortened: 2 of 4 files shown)


crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -454,24 +454,26 @@ impl Diagnostic {
...
-            .expect_ruff_file()
-            .to_source_code()
-            .line_column(self.expect_range().end())
+    /// Panics if the diagnostic has no primary span, or if its file is not a `SourceFile`.
+    pub fn ruff_end_location(&self) -> Option<LineColumn> {
+        self.range().map(|range| {
+            self.expect_primary_span()
+                .expect_ruff_file()
+                .to_source_code()
+                .line_column(range.end())
```
ntBre: If we're already returning an `Option`, could we get rid of some of these other `expect` calls too? You might have to move them out of the closure, or use `and_then` instead of `map`, but ideally we'll actually get to remove _all_ the `expect` methods eventually and that seems like another step in that direction.

With a little more work, we could get a `SourceCode` from a ty file too. Something like this:

https://github.com/astral-sh/ruff/blob/d55edb3d74b8220dad93eb6f187489c246698089/crates/ruff_db/src/diagnostic/render/gitlab.rs#L73-L77

Ruff's `EmitterContext` implements the resolver trait needed by `UnifiedFile::diagnostic_source`. 

Then we could potentially use this helper in the output formats that have already been ported over to `ruff_db` too. I'd be okay deferring this part, though.

amyreese: > If we're already returning an `Option`, could we get rid of some of these other `expect` calls too? You might have to move them out of the closure, or use `and_then` instead of `map`, but ideally we'll actually get to remove _all_ the `expect` methods eventually and that seems like another step in that direction.

I was initially trying to do that with a couple of chained `.map()` calls, but ran into some issues with the borrow checker that I don't understand, and resorted to this for now.

ntBre: Something like this should work:

```rust
    pub fn ruff_start_location(&self) -> Option<LineColumn> {
        let span = self.primary_span()?;
        let source_file = span.as_ruff_file()?;
        self.range()
            .map(|range| source_file.to_source_code().line_column(range.start()))
    }
```

or 

```rust
    pub fn ruff_start_location(&self) -> Option<LineColumn> {
        self.range().and_then(|range| {
            Some(
                self.primary_span()?
                    .as_ruff_file()?
                    .to_source_code()
                    .line_column(range.start()),
            )
        })
    }
```

or I guess using question marks everywhere might be better at that point:

```rust
    pub fn ruff_start_location(&self) -> Option<LineColumn> {
        Some(
            self.primary_span()?
                .as_ruff_file()?
                .to_source_code()
                .line_column(self.range()?.start()),
        )
    }
```

but yeah, it's easy to run into lifetime issues in this area of the code base, especially with `FileResolver` and `DiagnosticSource`.

amyreese: Turns out that leaning on `ruff_source_file()` was the correct path as that already uses `primary_span_ref()?.as_ruff_file()` to bypass the lifetime issue. Someday I will understand how Rust works... 😅

amyreese:
Not sure why the clippy and wasm jobs are failing — do they run from different base revisions or code checkouts?

ntBre:
Ah, there are two very-nearly identical chunks of code in the SARIF format. You've modified the non-WASM code:

https://github.com/astral-sh/ruff/blob/c3f1f0b9f1ed8e522a6d22c7d1670997c19183de/crates/ruff_linter/src/message/sarif.rs#L159-L162

but not the WASM code below it:

https://github.com/astral-sh/ruff/blob/c3f1f0b9f1ed8e522a6d22c7d1670997c19183de/crates/ruff_linter/src/message/sarif.rs#L178-L182

The `#[cfg(...)]` stuff is for conditional compilation, so unless you explicitly pass a WASM flag locally that code won't be compiled. That's probably why they pass locally (assuming they do).

amyreese:
> Ah, there are two very-nearly identical chunks of code in the SARIF format. You've modified the non-WASM code:

Weird, I would have expected the <F2> rename refactoring to update those as well, or rust/clippy to complain that the code referenced removed functions, even if the lines weren't enabled in my current build. 🫢

ntBre (approved):
Thanks! And nice catch on the docs and using `ruff_source_file`

PR #19481, opened 2025-07-22
[ty] Detect enums if metaclass is a subtype of EnumType/EnumMeta

sharkdp:

PR #20311, opened 2025-09-09
[ty] infer `name` and `value` for enum members

thejchap:
## summary
- this pr implements the following attributes for `Enum` members:
  - `name`
  - `_name_`
  - `value`
  - `_value_`
- adds a TODO test for `my_enum_class_instance.name`
- only implements if the instance is a subclass of `Enum` re: this [comment](https://github.com/astral-sh/ruff/pull/19481#issuecomment-3103460307) and existing [test](https://github.com/thejchap/ruff/blob/c34449ed7ca4b043d514240e212335a2554d0fb1/crates/ty_python_semantic/resources/mdtest/enums.md?plain=1#L625)

### pointers
- https://github.com/astral-sh/ty/issues/876
- https://typing.python.org/en/latest/spec/enums.html#enum-definition
- https://github.com/astral-sh/ruff/pull/19481#issuecomment-3103460307

## test plan
- mdtests
- triaged conformance diffs here: https://diffswarm.dev/d-01k531ag4nee3xmdeq4f3j66pb
- triaged mypy primer diffs here for django-stubs: https://diffswarm.dev/d-01k5331n13k9yx8tvnxnkeawp3
  - added a TODO test for overriding `.value`
- discord diff seems reasonable

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/enums.md
+++ b/crates/ty_python_semantic/resources/mdtest/enums.md
@@ -51,40 +51,6 @@ class ColorStr(Enum):
 reveal_type(enum_members(ColorStr))
 ```
 
-### Generated `_name_` and `_value_` attributes
-
-```py
-from enum import Enum
-from typing import Literal
-
-class Color(Enum):
-    RED = 1
-    GREEN = 2
-    BLUE = 3
-
-reveal_type(Color.RED._name_)  # revealed: Literal["RED"]
-reveal_type(Color.RED._value_)  # revealed: Literal[1]
-```
-
-### `name` attribute Literal unions
-
-```py
-from enum import Enum
-from typing import Literal
-
-class Color(Enum):
-    RED = 1
-    GREEN = 2
-    BLUE = 3
-
-def func1(red_or_blue: Literal[Color.RED, Color.BLUE]):
-    reveal_type(red_or_blue.name)  # revealed: Literal["RED", "BLUE"]
-
-def func2(any_color: Color):
-    # TODO: Literal["RED", "GREEN", "BLUE"]
-    reveal_type(any_color.name)  # revealed: Any
-```
-
 ### When deriving from `IntEnum`
 
 ```py
@@ -531,6 +497,62 @@ callable: Callable[[str], None] = Printer.STDOUT
 callable("Hello again!")
 callable = Printer.STDERR
 callable("Another error!")
+```
+
+## Special attributes on enum members
+
+### `name` and `_name_`
+
+```py
+from enum import Enum
+from typing import Literal
+
+class Color(Enum):
+    RED = 1
+    GREEN = 2
+    BLUE = 3
+
+reveal_type(Color.RED._name_)  # revealed: Literal["RED"]
+
+def _(red_or_blue: Literal[Color.RED, Color.BLUE]):
+    reveal_type(red_or_blue.name)  # revealed: Literal["RED", "BLUE"]
+
+def _(any_color: Color):
+    # TODO: Literal["RED", "GREEN", "BLUE"]
+    reveal_type(any_color.name)  # revealed: Any
+```
+
+### `value` and `_value_`
+
+```toml
+[environment]
+python-version = "3.11"
+```
+
+```py
+from enum import Enum, StrEnum
+from typing import Literal
+
+class Color(Enum):
+    RED = 1
+    GREEN = 2
+    BLUE = 3
+
+reveal_type(Color.RED.value)  # revealed: Literal[1]
+reveal_type(Color.RED._value_)  # revealed: Literal[1]
+
+reveal_type(Color.GREEN.value)  # revealed: Literal[2]
+reveal_type(Color.GREEN._value_)  # revealed: Literal[2]
+
+class Answer(StrEnum):
+    YES = "yes"
+    NO = "no"
+
+reveal_type(Answer.YES.value)  # revealed: Literal["yes"]
+reveal_type(Answer.YES._value_)  # revealed: Literal["yes"]
+
+reveal_type(Answer.NO.value)  # revealed: Literal["no"]
+reveal_type(Answer.NO._value_)  # revealed: Literal["no"]
 ```
 
 ## Properties of enum types
--- a/crates/ty_python_semantic/resources/mdtest/enums.md
+++ b/crates/ty_python_semantic/resources/mdtest/enums.md
@@ -4,6 +4,7 @@
 
 ```py
 from enum import Enum
+from typing import Literal
 
 class Color(Enum):
     RED = 1
@@ -11,8 +12,8 @@ class Color(Enum):
     BLUE = 3
 
 reveal_type(Color.RED)  # revealed: Literal[Color.RED]
-# TODO: This could be `Literal[1]`
-reveal_type(Color.RED.value)  # revealed: Any
+reveal_type(Color.RED.name)  # revealed: Literal["RED"]
+reveal_type(Color.RED.value)  # revealed: Literal[1]
 
 # TODO: Should be `Color` or `Literal[Color.RED]`
 reveal_type(Color["RED"])  # revealed: Unknown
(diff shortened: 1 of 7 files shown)


thejchap:
@sharkdp could use some feedback on approach for `_value_`/`value` before i get further in

i am thinking to add a `value_ty` field to [EnumLiteralType](https://github.com/thejchap/ruff/blob/64c8a66390b14c6b3778a94dafa1a52eed82afc9/crates/ty_python_semantic/src/types.rs#L10209) that stores [this value](https://github.com/thejchap/ruff/blob/64c8a66390b14c6b3778a94dafa1a52eed82afc9/crates/ty_python_semantic/src/types/enums.rs#L123), which i think would require giving `EnumMetadata` the `'db` lifetime

anything im missing there or other recommendations?

sharkdp:
> i am thinking to add a `value_ty` field to [EnumLiteralType](https://github.com/thejchap/ruff/blob/64c8a66390b14c6b3778a94dafa1a52eed82afc9/crates/ty_python_semantic/src/types.rs#L10209) that stores [this value](https://github.com/thejchap/ruff/blob/64c8a66390b14c6b3778a94dafa1a52eed82afc9/crates/ty_python_semantic/src/types/enums.rs#L123)

I would rather expect that we add the value `Type` for enum members to `members` on `EnumMetadata`? Maybe by turning it to a `Name => Type` `HashMap`? Is that what you meant?

> which i think would require giving `EnumMetadata` the `'db` lifetime

Yes, it looks like that would require adding a lifetime parameter to `EnumMetadata`. 

> anything im missing there

I don't think so

crates/ty_python_semantic/src/types/enums.rs
```diff
@@ -11,44 +13,52 @@ use crate::{
...
-#[derive(Debug, PartialEq, Eq)]
-pub(crate) struct EnumMetadata {
-    pub(crate) members: FxOrderSet<Name>,
-    pub(crate) aliases: FxHashMap<Name, Name>,
+#[salsa::interned(debug, heap_size=EnumMetadata::heap_size)]
+pub(crate) struct EnumMetadata<'db> {
+    #[returns(ref)]
+    pub(crate) members: FxOrderMap<Name, Type<'db>>,
+    #[returns(ref)]
+    pub(crate) aliases: BTreeMap<Name, Name>,
```
thejchap: changed from `FxHashMap` to `BTreeMap` because `FxHashMap` doesn't implement `HashEqLike`, which it looks like is required for salsa to intern the struct

admittedly i am not 100% sure that interning the struct is necessary - i went down this road due to this error after adding the `'db` parameter to `EnumMetadata`, and it was the pattern i saw in other areas where structs with fields tied to the `'db` lifetime were getting returned from a tracked function (for example [GenericContext](https://github.com/thejchap/ruff/blob/1c247f9130837bf52fdc873479222be92523d6c3/crates/ty_python_semantic/src/types/generics.rs#L103))

```
62 | pub(crate) fn enum_metadata<'db>(
   |                             --- lifetime `'db` defined here
...
65 | ) -> Option<EnumMetadata<'db>> {
   |      ^^^^^^ requires that `'db` must outlive `'static`
```

for my own learning i'd appreciate a nudge in the right direction if this is the wrong approach, or a quick explanation as to why it was the right approach :)

crates/ty_python_semantic/resources/mdtest/enums.md
```diff
@@ -167,6 +182,16 @@ class Answer(Enum):
...
 reveal_type(enum_members(Answer))
+
+class Choices(Enum):
+    A = 1
+    B = 2
+
+    @enum_property
+    def value(self) -> Any: ...
+
+# TODO: This should be `Any` - overridden by `@enum_property`
```
thejchap, line 186: https://diffswarm.dev/d-01k5331n13k9yx8tvnxnkeawp3?search=c-01k533qx21fstdzy6ys15rnjza

sharkdp: Thank you for adding this test. Can we separate it from the one above (let it have it's own fenced code block) and add a small prose description above?

crates/ty_python_semantic/resources/mdtest/enums.md
```diff
@@ -21,6 +28,13 @@ reveal_type(Color["RED"])  # revealed: Unknown
 reveal_type(Color(1))  # revealed: Color
 
 reveal_type(Color.RED in Color)  # revealed: bool
+
+def func1(red_or_blue: Literal[Color.RED, Color.BLUE]):
+    reveal_type(red_or_blue.name)  # revealed: Literal["RED", "BLUE"]
+
+def func2(any_color: Color):
+    # TODO: Literal["RED", "GREEN", "BLUE"]
+    reveal_type(any_color.name)  # revealed: Any
```
sharkdp: Can we move these tests out of the "Basic" test, into a separate new section somewhere down below with a new section header?

crates/ty_python_semantic/resources/mdtest/enums.md
```diff
@@ -4,15 +4,22 @@
...
-# TODO: This could be `Literal[1]`
-reveal_type(Color.RED.value)  # revealed: Any
+
+reveal_type(Color.RED.value)  # revealed: Literal[1]
+
+reveal_type(Color.RED._value_)  # revealed: Literal[1]
+
+reveal_type(Color.RED.name)  # revealed: Literal["RED"]
+
+reveal_type(Color.RED._name_)  # revealed: Literal["RED"]
```
sharkdp: Similar to the comment below, let's maybe only keep `.value` and `.name` here and move the `_value_`/`_name_` tests to a dedicated test section below that deals with these attributes on enum members?

```suggestion
reveal_type(Color.RED.value)  # revealed: Literal[1]
reveal_type(Color.RED.name)  # revealed: Literal["RED"]
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -3528,6 +3528,29 @@ impl<'db> Type<'db> {
                 .value_type(db)
                 .member_lookup_with_policy(db, name, policy),
 
+            Type::EnumLiteral(enum_literal)
+                if matches!(name_str, "name" | "_name_")
+                    && Type::ClassLiteral(enum_literal.enum_class(db))
+                        .is_subtype_of(db, KnownClass::Enum.to_subclass_of(db)) =>
```
sharkdp, line 3534: Is this necessary? (similar below)

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -427,7 +427,10 @@ impl<'db> UnionBuilder<'db> {
                     .collect::<FxOrderSet<_>>();
 
                 let all_members_are_in_union = metadata
-                    .members
+                    .members(self.db)
+                    .keys()
+                    .cloned()
+                    .collect::<FxOrderSet<_>>()
```
sharkdp: This might be the cause for the performance regression? Is there any way we can prevent cloning and `collect`ing here?

crates/ty_python_semantic/src/types/enums.rs
sharkdp: The alternative would be to derive `salsa::Update` for `EnumMetadata`. I think that should also get rid of this (very confusing and unhelpful) error message. That might be worth exploring to keep the diff in this PR smaller?

crates/ty_python_semantic/src/types.rs
thejchap: without this, [these tests](https://github.com/astral-sh/ruff/pull/20311/files#diff-4ec0135eb3494415ddc8cee7f713314fcfcbc2b9d2722fa2292d75348791bd81R661) fail - i added this after reading through this [thread](https://github.com/astral-sh/ruff/pull/19481#issuecomment-3103460307)

crates/ty_python_semantic/src/types/enums.rs
thejchap: required switching `FxOrderMap` to `FxIndexMap` because salsa [does not provide](https://github.com/salsa-rs/salsa/blob/master/src/update.rs#L322) an `Update` impl for `FxOrderMap`

sharkdp (approved):
Thank you!

Glad we could get rid of the performance regression.

PR #17371, opened 2025-04-13
[ty] infer function's return type

mtshiba:
## Summary

This PR closes astral-sh/ty#128.

`FunctionType::infer_return_type` is added to infer the return type of a function when its return type is not specified.

## TODOs

- [x] infer simple function's return type
~~- [ ] infer generator's return type~~
~~- [ ] infer coroutine's return type~~
~~- [ ] infer lambda function's return type~~

## Test Plan

New test cases are added to `mdtest/function/return_type.md`.

PR #20312, opened 2025-09-09
[ty] introduce `DivergentType`

mtshiba:
## Summary

From #17371

In #17371, `DivergentType` was introduced to prevent type inference for recursive functions from diverging and causing panics. This turned out to be useful for other divergent type inferences (https://github.com/astral-sh/ruff/pull/17371#discussion_r2329337965), so I extracted the introduction part of `DivergentType` into this PR so that we can use it without waiting for the merge of #17371.

Note that this PR only introduces `DivergentType` and does not actually address divergent type inference yet. Please refer to https://github.com/astral-sh/ruff/pull/17371/files#diff-20b910c6e20faa962bb1642e111db1cbad8e66ace089bdd966ac9d7f9fa99ff2R542-R622 etc. when implementing handling of divergence suppression using this type.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -7546,6 +7546,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             // the result would then become Any or Unknown, respectively).
             (div @ Type::Dynamic(DynamicType::Divergent(_)), _, _)
             | (_, div @ Type::Dynamic(DynamicType::Divergent(_)), _) => Some(div),
+
             (any @ Type::Dynamic(DynamicType::Any), _, _)
             | (_, any @ Type::Dynamic(DynamicType::Any), _) => Some(any),
 


crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -7544,6 +7544,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
             // Non-todo Anys take precedence over Todos (as if we fix this `Todo` in the future,
             // the result would then become Any or Unknown, respectively).
+            (div @ Type::Dynamic(DynamicType::Divergent(_)), _, _)
+            | (_, div @ Type::Dynamic(DynamicType::Divergent(_)), _) => Some(div),
```
carljm, line 7548:
```suggestion
            | (_, div @ Type::Dynamic(DynamicType::Divergent(_)), _) => Some(div),

```

carljm (approved):
Looks good, thank you!  I put together a [proof-of-concept](https://github.com/astral-sh/ruff/pull/20333) to show how we can use this type to address an infinite-tuple case in implicit instance attribute (given a [small change in Salsa](https://github.com/salsa-rs/salsa/pull/991)) -- that's enough to convince me that this will be generally useful.

PR #20314, opened 2025-09-09
[ty] Allow protocols to participate in nominal subtyping as well as structural subtyping

AlexWaygood:
## Summary

(Stacked on top of https://github.com/astral-sh/ruff/pull/20284; review that PR first.)

Currently our implementation of subtyping and assignability checks against protocols only checks whether a given type is _structurally_ a subtype of/assignable to a given protocol. Other type checkers, however, allow protocols to participate in nominal subtyping _as well as_ structural subtyping. The most common example where this comes up is that all other type checkers consider `str` to be a subtype of `Container[str]` even though `str.__contains__` is less permissive in the types it accepts than `Container.__contains__`; the reason they consider it a subtype nonetheless is that `str` has `Container[str]` in its MRO. We currently also consider `str` to be a subtype of `Container[str]`, but this is because we still do not look at the types of method members when doing subtyping/assignability checks for protocols; an early version of https://github.com/astral-sh/ruff/pull/20165 revealed many ecosystem false positives due to our failure to understand `str` as being a subtype of `Container[str]`.

This PR therefore updates our subtyping/assignability logic for protocols so that, for a given class-based protocol, we also check whether a given other type could be considered a nominal subtype of that protocol before concluding that there exists no subtype relation between the other type and the protocol.

The implication of this is that one of our property tests must be removed: `all_type_assignable_to_iterable_are_iterable`. Just because a type can be assignable to `Iterable` does not mean it is necessarily iterable -- it could be a Liskov-uncompliant subclass of `Iterable`! I don't think we have any Liskov-uncompliant classes in our property-test generation right now, but this still seems like a potential cause for confusion if we added one (deliberately or not) in the future. As such, I've removed the property test as part of this PR.

## Test Plan

Mdtests added.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -1058,6 +1058,12 @@ impl<'db> Bindings<'db> {
                             // `tuple(range(42))` => `tuple[int, ...]`
                             // BUT `tuple((1, 2))` => `tuple[Literal[1], Literal[2]]` rather than `tuple[Literal[1, 2], ...]`
                             if let [Some(argument)] = overload.parameter_types() {
+                                // We deliberately use `.iterate()` here (falling back to `Unknown` if it isn't iterable)
+                                // rather than `.try_iterate().expect()`. Even though we know at this point that the input
+                                // type is assignable to `Iterable`, that doesn't mean that the input type is *actually*
+                                // iterable (it could be a Liskov-uncompliant subtype of the `Iterable` class that sets
+                                // `__iter__ = None`, for example). That would be badly written Python code, but we still
+                                // need to be able to handle it without crashing.
                                 overload.set_return_type(Type::tuple(TupleType::new(
                                     db,
                                     &argument.iterate(db),
--- a/crates/ty_python_semantic/src/types/property_tests.rs
+++ b/crates/ty_python_semantic/src/types/property_tests.rs
@@ -320,6 +320,15 @@ mod flaky {
     // iteration protocol as well as the new-style iteration protocol: not all objects that
     // we consider iterable are assignable to `Iterable[object]`.
     //
+    // Note also that (like other property tests in this module),
+    // this invariant will only hold true for Liskov-compliant types assignable to `Iterable`.
+    // Since protocols can participate in nominal assignability/subtyping as well as
+    // structural assignability/subtyping, it is possible to construct types that a type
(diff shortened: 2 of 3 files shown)


crates/ty_python_semantic/src/types/property_tests.rs
```diff
@@ -313,16 +311,4 @@ mod flaky {
...
-    //
-    // Note that the inverse is not true, due to the fact that we recognize the old-style
-    // iteration protocol as well as the new-style iteration protocol: not all objects that
-    // we consider iterable are assignable to `Iterable[object]`.
-    //
-    // Currently flaky due to <https://github.com/astral-sh/ty/issues/889>
-    type_property_test!(
-        all_type_assignable_to_iterable_are_iterable, db,
-        forall types t. t.is_assignable_to(db, KnownClass::Iterable.to_specialized_instance(db, [Type::object(db)])) => t.try_iterate(db).is_ok()
-    );
```
carljm: I'm not convinced by the rationale given in the PR description for removing this property test. I think there are many type properties that we want to verify for valid types, that can be invalidated by an invalid type (e.g. a type that violates Liskov), and we probably should never introduce such types into our type generation for the property tests. Or if we do, we should still keep some property tests with a pre-condition that they only apply to valid types.

Especially if there's no _current_ problem, I would not remove this property test. If anything, I'd just add a note to its docstring that it is not true for types which explicitly inherit `Iterable` but violate LSP on its interface.

AlexWaygood: hmm, I suppose we also have property tests that guarantee that subtyping is intransitive, and they would also be broken by Liskov-uncompliant classes interacting with protocols.

I still worry a bit that this could cause confusion for us at some point... there are a lot of Liskov-uncompliant classes in the standard library! Let's hope we're vigilant about keeping them out of our property tests...

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1058,17 +1058,9 @@ impl<'db> Bindings<'db> {
...
-                                        "try_iterate() should not fail on a type \
-                                            assignable to `Iterable`",
-                                    );
-                                    continue;
-                                };
                                 overload.set_return_type(Type::tuple(TupleType::new(
                                     db,
-                                    tuple_spec.as_ref(),
+                                    &argument.iterate(db),
                                 )));
```
AlexWaygood: I should add an mdtest here for calling `tuple()` on a Liskov-uncompliant subclass of `Iterable`

PR #20316, opened 2025-09-09
[ty] Add tests for protocols with generic method members

AlexWaygood:
## Summary

Protocols can have method members, and those method members can have generic contexts. The generic context of a protocol's method members can be scoped to the class or the method itself.

We currently have some "accidental integration tests" for generic method members in our named-tuple test suite, but no explicit unit tests in `protocols.md`. This PR adds them.

## Test Plan

`cargo test -p ty_python_semantic --test=mdtest`

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1801,6 +1801,98 @@ static_assert(is_assignable_to(str, SupportsLessThan))
 static_assert(is_assignable_to(int, Invertable))
 ```
 
+## Subtyping of protocols with generic method members
+
+Protocol method members can be generic. They can have generic contexts scoped to the class:
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+from typing_extensions import TypeVar, Self, Protocol
+from ty_extensions import is_equivalent_to, static_assert, is_assignable_to, is_subtype_of
+
+class NewStyle[T](Protocol):
+    def method(self, input: T) -> None: ...
+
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -1858,6 +1858,98 @@ static_assert(is_assignable_to(str, SupportsLessThan))
...
+    return x
+
+class Nominal:
+    def method(self, input: int) -> None: ...
+
+static_assert(is_assignable_to(Nominal, NewStyle))
+static_assert(is_assignable_to(Nominal, OldStyle))
+
+# `NewStyle` is implicitly `NewStyle[Unknown]`, which is not a fully static type,
+# hence there is no subtyping relation:
```
sharkdp: "non-fully-static => no subtyping relation" is not a correct conclusion anymore since https://github.com/astral-sh/ruff/pull/18799 has been merged, I think?

The test assertions are still fine, I think, but we could consider changing the comment?

AlexWaygood: uff, yes, thanks. I can never remember what our new rule is for when something can be assignable but not a subtype 🫣 let me read up on that again...

AlexWaygood:
```suggestion
# `NewStyle` is implicitly `NewStyle[Unknown]`,
# and there exist fully static materializations of `NewStyle[Unknown]`
# where `Nominal` would not be a subtype of the given materialization,
# hence there is no subtyping relation:
```

PR #20318, opened 2025-09-09
[`flake8-bugbear`] `unreliable-callable-check` always marked as unsafe (B004)

### LINKED ISSUE #20282: `unreliable-callable-check` (B004) should be marked unsafe
### Summary

The fix for [`unreliable-callable-check` (B004)](https://docs.astral.sh/ruff/rules/unreliable-callable-check/) should be marked unsafe because the replacement is not exactly equivalent to the original expression. (If they were equivalent, `callable` wouldn’t have been reintroduced in Python 3.2.) [Example](https://play.ruff.rs/5ff82084-76d5-45ac-a269-dd0ed920e885):
```console
$ cat >b004.py <<'# EOF'
import operator
print(hasattr(operator, "__call__"))

class C:
    def __init__(self): self.__call__ = None
print(hasattr(C(), "__call__"))
# EOF

$ python b004.py
True
True

$ ruff --isolated check b004.py --select B004 --fix

$ python b004.py
False
False
```

### Version

IDrokin117:
## Summary
Resolves #20282

Makes the rule fix always unsafe, because the replacement may not be semantically equivalent to the original expression, potentially changing the behavior of the code.

Updated docstring with examples.

## Test Plan
- Added two tests from issue and regenerated the snapshot


crates/ruff_linter/src/rules/flake8_bugbear/rules/unreliable_callable_check.rs
```diff
@@ -28,10 +28,29 @@ use crate::{Edit, Fix, FixAvailability, Violation};
...
+/// class A:
+///     def __init__(self):
+///         self.__call__ = None
+///
+///
+/// assert hasattr(A(), "__call__")
+/// assert callable(A()) is False
+/// ```
+///
+/// If there are comments in the `hasattr` call expression, then comments may be removed:
```
ntBre:
```suggestion
/// Additionally, if there are comments in the `hasattr` call expression, they may be removed:
```

PR #20319, opened 2025-09-09
[ty] Add mdtests that exercise constraint sets

dcreager:
This PR adds a new `ty_extensions.ConstraintSet` class, which is used to expose constraint sets to our mdtest framework. This lets us write a large collection of unit tests that exercise the invariants and rewrite rules of our constraint set implementation.

As part of this, `is_assignable_to` and friends are updated to return a `ConstraintSet` instead of a `bool`, and we implement `ConstraintSet.__bool__` to return when a constraint set is always satisfied. That lets us still use `static_assert(is_assignable_to(...))`, since the assertion will coerce the constraint set to a bool, and also lets us `reveal_type(is_assignable_to(...))` to see more detail about whether/when the two types are assignable. That lets us get rid of `reveal_when_assignable_to` and friends, since they are now redundant with the expanded capabilities of `is_assignable_to`.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -506,14 +506,28 @@ Otherwise, the union cannot be simplified.
 
 ```py
 def _[T]() -> None:
-    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Base) ∨ (Sub ≤ T@_ ≤ Super)]
-    reveal_type(range_constraint(SubSub, T, Base) | range_constraint(Sub, T, Super))
     # revealed: ty_extensions.ConstraintSet[(Sub ≤ T@_ ≤ Base) ∨ (Base ≤ T@_ ≤ Super)]
     reveal_type(range_constraint(Sub, T, Base) | range_constraint(Base, T, Super))
     # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Sub) ∨ (Base ≤ T@_ ≤ Super)]
     reveal_type(range_constraint(SubSub, T, Sub) | range_constraint(Base, T, Super))
     # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Sub) ∨ (Unrelated ≤ T@_)]
     reveal_type(range_constraint(SubSub, T, Sub) | range_constraint(Unrelated, T, object))
+```
+
+In particular, the following does not simplify, even though it seems like it could simplify to
+`SubSub ≤ T@_ ≤ Super`. The issue is that there are types that are within the bounds of
+`SubSub ≤ T@_ ≤ Super`, but which are not comparable to `Base` or `Sub`, and which therefore should
+not be included in the union. An example would be the type that contains all instances of `Super`,
+`Base`, and `SubSub` (but _not_ including instances of `Sub`). (We don't have a way to spell that
+type at the moment, but it is a valid type.) That type is not in `SubSub ≤ T ≤ Base`, since it
+includes `Super`, which is outside the range. It's also not in `Sub ≤ T ≤ Super`, because it does
+not include `Sub`. That means it should not be in the union. Since that type _is_ in
+`SubSub ≤ T ≤ Super`, it is not correct to simplify the union in this way.
+
+```py
+def _[T]() -> None:
+    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Base) ∨ (Sub ≤ T@_ ≤ Super)]
+    reveal_type(range_constraint(SubSub, T, Base) | range_constraint(Sub, T, Super))
 ```
 
 ### Union of range and not-equivalent


crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md
```diff
@@ -129,31 +129,31 @@ specialization. Thus, the typevar is a subtype of itself and of `object`, but no
...
-
-    reveal_when_subtype_of(T, T)  # revealed: always
-    reveal_when_subtype_of(T, object)  # revealed: always
-    reveal_when_subtype_of(T, Super)  # revealed: never
-    reveal_when_subtype_of(U, U)  # revealed: always
-    reveal_when_subtype_of(U, object)  # revealed: always
-    reveal_when_subtype_of(U, Super)  # revealed: never
-    reveal_when_subtype_of(T, U)  # revealed: never
-    reveal_when_subtype_of(U, T)  # revealed: never
+    reveal_type(is_assignable_to(T, T))  # revealed: ty_extensions.ConstraintSet[always]
```
carljm, line 140: These get quite a bit more verbose. Almost tempted to say we should keep `reveal_when_assignable_to` as a wrapper since it's so much more concise? But I also like reducing the API surface area of `ty_extensions`.

Since all of these tests reveal `always` or `never`, could they just use `static_assert(is_assignable_to(...))` and `static_assert(not is_assignable_to(...))` instead? Or is the idea that you want to verify we are returning `always` or `never`, rather than some set of conditions which we would convert to `false` along with `never`?

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -0,0 +1,749 @@
...
+def _[T]() -> None:
+    # revealed: ty_extensions.ConstraintSet[(T@_ ≠ Never)]
+    reveal_type(not_equivalent_constraint(T, Never))
+
+    # revealed: ty_extensions.ConstraintSet[(T@_ ≠ object)]
+    reveal_type(not_equivalent_constraint(T, object))
+```
+
+Constraints can only refer to fully static types, so the hole is transformed into its top
+materialization.
```
carljm: This seems arbitrary (the choice of top materialization), which I think is OK because I think it only ever matters when using `ty_extensions.not_equivalent_constraint`? Because otherwise a not-equivalent constraint would only ever be generated via negation of a range constraint, right? And the range constraint will already have fully-static upper and lower bounds, resulting in a fully-static hole from the start.

If this is all correct, would it make sense to clarify some of this more here?

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -0,0 +1,749 @@
...
+def _[T]() -> None:
+    # revealed: ty_extensions.ConstraintSet[never]
+    reveal_type(incomparable_constraint(T, Never))
+
+    # revealed: ty_extensions.ConstraintSet[never]
+    reveal_type(incomparable_constraint(T, object))
+```
+
+Constraints can only refer to fully static types, so the pivot is transformed into its top
+materialization.
```
carljm: Same comment as above.

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -0,0 +1,749 @@
...
+    reveal_type(range_constraint(SubSub, T, Super) | range_constraint(Sub, T, Base))
+    # revealed: ty_extensions.ConstraintSet[(Sub ≤ T@_ ≤ Super)]
+    reveal_type(range_constraint(Sub, T, Super) | range_constraint(Sub, T, Super))
+```
+
+Otherwise, the union cannot be simplified.
+
+```py
+def _[T]() -> None:
+    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Base) ∨ (Sub ≤ T@_ ≤ Super)]
```
carljm: Why doesn't this union simplify to `SubSub <= T <= Super`?

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -0,0 +1,749 @@
...
+    reveal_type(range_constraint(SubSub, T, Base) | range_constraint(Sub, T, Super))
+    # revealed: ty_extensions.ConstraintSet[(Sub ≤ T@_ ≤ Base) ∨ (Base ≤ T@_ ≤ Super)]
+    reveal_type(range_constraint(Sub, T, Base) | range_constraint(Base, T, Super))
```
carljm, line 510: Similarly it seems to me this should simplify to `Sub <= T <= Super`

carljm (approved):
Nice!

crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md
dcreager: I originally had left out the `ty_extensions.ConstraintSet` part, which would have left the reveal renderings exactly as they were before. But it seemed like we are trying to have the `reveal_type` output include the type name.

> Or is the idea that you want to verify we are returning `always` or `never`, rather than some set of conditions which we would convert to `false` along with `never`?

This, and more importantly, in #20093 some of these start returning constraint sets that are only sometimes satisfiable.

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
dcreager: There can be types that satisfy `SubSub ≤ T ≤ Super` but which aren't comparable to `Sub` or `Base`, and which therefore shouldn't be included in the union.

So with just set theory, if the possible values are 0, 1, and 2, you get a type lattice of

```
   012
01  02  12
 0   1   2
     ∅
```

Then we might have

```
SubSub = ∅
Sub    = 1
Base   = 12
Super  = 012

SubSub ≤ T ≤ Base  = {∅, 1, 2, 12}
Sub ≤ T ≤ Super    = {1, 01, 12, 012}
```

The union should be

```
(SubSub ≤ T ≤ Base) ∪ (Sub ≤ T ≤ Super) = {∅, 1, 2, 01, 12, 012}
```

but the simplification you suggest would be

```
SubSub ≤ T ≤ Super = {∅, 0, 1, 2, 01, 02, 12, 012}
```

`0` and `02` are included in the simplification but not in the actual union.

Translating that into Python, an example problem type would be `Super \ Sub*`, where `Sub*` indicates instances of just `Sub`, but not any of its subclasses. (So in English, the problem type is "instances of `Super`, or any subclass of `Super` other than `Sub`". That type is not in `SubSub ≤ T ≤ Base`, since it includes `Super`, which is outside the range. It's also not in `Sub ≤ T ≤ Super`, because it does not include `Sub`. But it is in `SubSub ≤ T ≤ Super`.

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
dcreager: > This seems arbitrary (the choice of top materialization),

Yeah, I wasn't sure which materialization to use here, or if the correct thing to do was to make a range from the bottom materialization to the top, and then negate that. (Even if that's more correct, I'm not at all sure what the analogue would be for incomparable below.)

But as you point out, we never create these directly, so it should be moot. I'll update the `ty_extensions` constructor to _require_ fully static inputs, and add commentary here explaining why.

crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
carljm: Makes sense, thank you! Might be worth recording a short version of this as a comment?

dcreager: Done

PR #20320, opened 2025-09-09
Move GitHub rendering to `ruff_db`

ntBre:
## Summary

This is the GitHub analog to #20117. This PR prepares to add a GitHub output format to ty by moving the implementation from `ruff_linter` to `ruff_db`. Hopefully this one is a bit easier to review commit-by-commit. Almost all of the refactoring this time is in the first commit, then the second commit adds the new `OutputFormat` variant and moves the file into `ruff_db`. The third commit is just a small touch up to use a private method that accommodates ty files so that we can run the tests and update/move the snapshots.

I had to push a fourth commit to fix and test diagnostics without a span/file.

## Test Plan

Existing tests

### DIFF
--- a/crates/ruff_db/src/diagnostic/render/github.rs
+++ b/crates/ruff_db/src/diagnostic/render/github.rs
@@ -53,17 +53,19 @@ impl<'a> GithubRenderer<'a> {
 
                 write!(
                     f,
-                    "{path}:{row}:{column}:",
+                    "{path}:{row}:{column}: ",
                     path = file.relative_path(self.resolver).display(),
                     row = start_location.line,
                     column = start_location.column,
                 )?;
+            } else {
+                write!(f, "::")?;
             }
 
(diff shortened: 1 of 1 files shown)


crates/ruff_db/src/diagnostic/render/github.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+                .unwrap_or_default();
+
+                write!(
+                    f,
+                    ",line={row},col={column},endLine={end_row},endColumn={end_column}::",
+                    row = start_location.line,
+                    column = start_location.column,
+                    end_row = end_location.line,
+                    end_column = end_location.column,
```
ntBre, line 51: I think we could omit these entirely if there's no range, based on the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message), but I stuck with `unwrap_or_default` for now since that's what we were already doing for notebooks.

PR #20322, opened 2025-09-09
Remove Diagnostic::expect_range and all consumers

amyreese:
Replace usage with `range().unwrap_or_default()` or more appropriate
alternatives based on context.


crates/ruff_linter/src/rules/flake8_commas/rules/trailing_commas.rs
```diff
@@ -364,9 +364,10 @@ fn check_token(
         if let Some(mut diagnostic) =
             lint_context.report_diagnostic_if_enabled(ProhibitedTrailingComma, prev.range())
         {
-            let range = diagnostic.expect_range();
-            diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
-            return;
+            if let Some(range) = diagnostic.range() {
+                diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
+                return;
+            }
```
ntBre: I think in these rule cases we should try to use the range passed to `report_diagnostic` or `report_diagnostic_if_enabled`. We can just use `prev.range()` again or factor out a shared `range` variable in this instance.

It works either way since the range will always be `Some`, but it seems more clear to me.

crates/ruff_linter/src/rules/isort/mod.rs
```diff
@@ -657,7 +657,7 @@ mod tests {
                 ..LinterSettings::for_rule(Rule::UnsortedImports)
             },
         )?;
-        diagnostics.sort_by_key(|diagnostic| diagnostic.expect_range().start());
+        diagnostics.sort_by_key(|diagnostic| diagnostic.range().unwrap_or_default().start());
```
ntBre: Just for fun, would you try deleting one of these sort calls entirely? I'm pretty sure the diagnostics from `test_path` should already be sorted, so we may be able to drop all of these `sort_by_key` and `unwrap_or_default` calls.

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -172,8 +172,9 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
                             WhitespaceAfterOpenBracket { symbol },
                             TextRange::at(token.end(), trailing_len),
                         ) {
-                            let range = diagnostic.expect_range();
-                            diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
+                            if let Some(range) = diagnostic.range() {
```
ntBre: Same as above, we can just use the `range` passed to `report_diagnostic_if_enabled`.

crates/ruff_linter/src/rules/pyflakes/mod.rs
```diff
@@ -788,7 +788,7 @@ mod tests {
             &parsed,
             target_version,
         );
-        messages.sort_by_key(|diagnostic| diagnostic.expect_range().start());
+        messages.sort_by_key(|diagnostic| diagnostic.range().unwrap_or_default().start());
```
ntBre: Unlike isort, I think we actually do need this one since pyflakes uses a slightly different test runner. It might make sense to use `Diagnostic::ruff_sort_ordering` again if it doesn't change any snapshots, though. Hopefully we'll unify these with the other tests soon anyway (https://github.com/astral-sh/ruff/pull/19977).

crates/ruff_linter/src/linter.rs
```diff
@@ -513,10 +513,12 @@ fn diagnostics_to_messages(
...
-            let noqa_offset = directives
-                .noqa_line_for
-                .resolve(diagnostic.expect_range().start());
-            diagnostic.set_noqa_offset(noqa_offset);
+            if let Some(noqa_offset) = diagnostic
+                .range()
+                .map(|range| directives.noqa_line_for.resolve(range.start()))
+            {
+                diagnostic.set_noqa_offset(noqa_offset);
+            }
```
ntBre: I think I'd probably immediately if-let on the range itself, but what you have is fine too:


```suggestion
            if let Some(range) = diagnostic.range() {
                let noqa_offset = directives.noqa_line_for.resolve(range.start());
                diagnostic.set_noqa_offset(noqa_offset);
            }
```

crates/ruff_linter/src/linter.rs
```diff
@@ -983,7 +985,7 @@ mod tests {
             &parsed,
             target_version,
         );
-        diagnostics.sort_by_key(|diagnostic| diagnostic.expect_range().start());
+        diagnostics.sort_by_key(|diagnostic| diagnostic.range().unwrap_or_default().start());
```
ntBre: Maybe another good use for `Diagnostic::ruff_sort_ordering`

crates/ruff_linter/src/noqa.rs
```diff
@@ -886,7 +886,10 @@ fn find_noqa_comments<'a>(
             }
         }
 
-        let noqa_offset = noqa_line_for.resolve(message.expect_range().start());
+        let noqa_offset = message
+            .range()
+            .map(|range| noqa_line_for.resolve(range.start()))
+            .unwrap_or_default();
```
ntBre, line 892: I'm actually not sure, but I think we don't want to unwrap this one. That would put a noqa offset at the start of a file for any diagnostic without a range, if I understand correctly. I think we should just put the references to `noqa_offset` in an if-let here.

crates/ruff_server/src/lint.rs
```diff
@@ -234,7 +234,7 @@ fn to_lsp_diagnostic(
     index: &LineIndex,
     encoding: PositionEncoding,
 ) -> (usize, lsp_types::Diagnostic) {
-    let diagnostic_range = diagnostic.expect_range();
+    let diagnostic_range = diagnostic.range().unwrap_or_default();
```
ntBre, line 237: I double checked what ty does for this, and it looks like ranges are required for `lsp_types::Diagnostic`. We fall back to the default range in ty too:

https://github.com/astral-sh/ruff/blob/7a75702237e6f0bc605d3d8e9470c4ca9f37c69e/crates/ty_server/src/server/api/diagnostics.rs#L300-L302

so this looks good!

crates/ruff_wasm/src/lib.rs
```diff
@@ -231,8 +231,12 @@ impl Workspace {
...
                 code: msg.secondary_code_or_id().to_string(),
                 message: msg.body().to_string(),
-                start_location: source_code.line_column(msg.expect_range().start()).into(),
-                end_location: source_code.line_column(msg.expect_range().end()).into(),
+                start_location: source_code
+                    .line_column(msg.range().unwrap_or_default().start())
+                    .into(),
+                end_location: source_code
+                    .line_column(msg.range().unwrap_or_default().end())
+                    .into(),
```
ntBre, line 239: This also seems right to me. In ty, it looks like we preserve the optionality on the Rust side but unwrap to a default value in TypeScript:

https://github.com/astral-sh/ruff/blob/7a75702237e6f0bc605d3d8e9470c4ca9f37c69e/playground/ty/src/Editor/Diagnostics.tsx#L74-L75

crates/ruff_linter/src/rules/isort/mod.rs
amyreese: Yep, removing these did not break tests.

crates/ruff_linter/src/noqa.rs
amyreese: This one was tricky to try and unravel what's going on, and I *think* this is actually correct as-is:

It's using the range to look for an associated noqa directive (`find_lines_with_directive` takes the range and does a binary search against noqa directives that it knows about). With a default range, IIUC it should never find a directive, and therefore fall through to the following code at line 915, which reports a `None` directive using the line from the default range.

With an if-let defining `noqa_offset` guarding the use of `find_line_with_directive`, then `noqa_offset` is no longer defined and usable by the block at line 915.

ntBre: Hmm, maybe you're right. I'm honestly not that familiar with this part of the code. I was just afraid that you'd be able to put a noqa comment at the start of a file for a diagnostic without a range. But maybe that's even correct?

Anyway, this is actually unreachable until we add any diagnostics without ranges, so this seems fine. Even the panics shouldn't hit this path because they don't have rule/noqa codes anyway.

Thanks for looking into it!

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
ntBre: Sorry I stopped marking these, but I was hoping we could do a fix similar to `crates/ruff_linter/src/rules/flake8_commas/rules/trailing_commas.rs` for each of the `diagnostic.range()` calls immediately after constructing a diagnostic. In this case it would look something like:


```suggestion
                        let range = TextRange::at(token.end(), trailing_len);
                        if let Some(mut diagnostic) = ... {
                            diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
                        }
```

(with some of the context from above added to the suggestion since github won't let me select it all)

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -188,8 +189,9 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
                                 WhitespaceBeforeCloseBracket { symbol },
                                 TextRange::at(token.start() - offset, offset),
                             ) {
-                                let range = diagnostic.expect_range();
-                                diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
+                                if let Some(range) = diagnostic.range() {
+                                    diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
+                                }
```
ntBre: Same as above:

```rust
let range = TextRange::at(token.start() - offset, offset);
...
diagnostic.set_fix(Fix::safe_edit(Edit::range_deletion(range)));
```

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -216,9 +218,11 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
                                             TextRange::at(token.start() - offset, offset),
                                         )
                                     {
-                                        let range = diagnostic.expect_range();
-                                        diagnostic
-                                            .set_fix(Fix::safe_edit(Edit::range_deletion(range)));
+                                        if let Some(range) = diagnostic.range() {
```
ntBre: Same as above

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -233,10 +237,11 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
                                                 TextRange::at(token.start() - offset, offset),
                                             )
                                         {
-                                            let range = diagnostic.expect_range();
-                                            diagnostic.set_fix(Fix::safe_edit(
-                                                Edit::range_deletion(range),
-                                            ));
+                                            if let Some(range) = diagnostic.range() {
```
ntBre: Same as above

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -261,14 +266,15 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
...
                                         {
-                                            let range = diagnostic.expect_range();
-                                            diagnostic.set_fix(Fix::safe_edits(
-                                                Edit::range_deletion(range),
-                                                [Edit::insertion(
-                                                    " ".into(),
-                                                    token.start() - offset,
-                                                )],
-                                            ));
+                                            if let Some(range) = diagnostic.range() {
```
ntBre: Same as above

crates/ruff_linter/src/rules/pycodestyle/rules/logical_lines/extraneous_whitespace.rs
```diff
@@ -284,10 +290,11 @@ pub(crate) fn extraneous_whitespace(line: &LogicalLine, context: &LintContext) {
                                                 TextRange::at(token.start() - offset, offset),
                                             )
                                         {
-                                            let range = diagnostic.expect_range();
-                                            diagnostic.set_fix(Fix::safe_edit(
-                                                Edit::range_deletion(range),
-                                            ));
+                                            if let Some(range) = diagnostic.range() {
```
ntBre: Same as above

crates/ruff_linter/src/test.rs
```diff
@@ -383,7 +383,7 @@ Either ensure you always emit a fix or change `Violation::FIX_AVAILABILITY` to e
 
             // Not strictly necessary but adds some coverage for this code path by overriding the
             // noqa offset and the source file
-            let range = diagnostic.expect_range();
+            let range = diagnostic.range().unwrap();
```
ntBre: This one could probably be an if-let too. I _think_ we could hit this one with the panics, depending on how we end up testing them (or if a rule panics in a future test, but you'd probably only hit that while trying to fix a bug).

amyreese: I also added if lets guarding the other use of `range.unwrap()` here, since those are also likely to cause issues in the same way.

ntBre: Good call!

ntBre (approved):

PR #20324, opened 2025-09-10
[ty] Simplify unions of enum literals and subtypes thereof

sharkdp:
## Summary

When adding an enum literal `E = Literal[Color.RED]` to a union which already contained a subtype of that enum literal(!), we were previously not simplifying the union correctly. My assumption is that our property tests didn't catch that earlier, because the only possible non-trivial subytpe of an enum literal that I can think of is `Any & E`. And in order for that to be detected by the property tests, it would have to randomly generate `Any & E | E` and then also compare that with `E` on the other side (in an equivalence test, or the subtyping-antisymmetry test).

closes https://github.com/astral-sh/ty/issues/1155

## Test Plan

* Added a regression test.
* I also ran the property tests for a while, but probably not for two months worth of daily CI runs.

### DIFF
--- a/crates/ty_python_semantic/src/types/builder.rs
+++ b/crates/ty_python_semantic/src/types/builder.rs
@@ -444,8 +444,7 @@ impl<'db> UnionBuilder<'db> {
                     .filter_map(UnionElement::to_type_element)
                     .any(|ty| Type::EnumLiteral(enum_member_to_add).is_subtype_of(self.db, ty))
                 {
-                    self.elements
-                        .push(UnionElement::Type(Type::EnumLiteral(enum_member_to_add)));
+                    self.push_type(Type::EnumLiteral(enum_member_to_add), seen_aliases);
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -444,93 +444,96 @@ impl<'db> UnionBuilder<'db> {
                     .filter_map(UnionElement::to_type_element)
                     .any(|ty| Type::EnumLiteral(enum_member_to_add).is_subtype_of(self.db, ty))
                 {
-                    self.elements
-                        .push(UnionElement::Type(Type::EnumLiteral(enum_member_to_add)));
```
sharkdp, line 448: This was the problem. Instead of just using `self.elements.push`, we should go through the whole simplification logic in the `_ => {…}` branch, which I have now moved to a separate function. Every other change in this file is just the "moving out to a separate function" part.

PR #20325, opened 2025-09-10
[ty] Use 'unknown' specialization for upper bound on Self

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1156

## Test Plan

Added a regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/self.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/self.md
@@ -2,7 +2,7 @@
 
 ```toml
 [environment]
-python-version = "3.11"
+python-version = "3.13"
 ```
 
 `Self` is treated as if it were a `TypeVar` bound to the class it's being used on.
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -5708,9 +5708,7 @@ impl<'db> Type<'db> {
                             ],
                         });
                     };
-                    let instance = Type::ClassLiteral(class).to_instance(db).expect(
-                        "nearest_enclosing_class must return type that can be instantiated",
-                    );
```
sharkdp, line 5713: This would previously use the "default" specialization, and so we would end up with an upper bound of `Container[bytes]` for the example in the mdtest. With the change here, we instead use `Container[Unknown]` as the upper bound.

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -2,7 +2,7 @@
 
 ```toml
 [environment]
-python-version = "3.11"
+python-version = "3.13"
```
sharkdp, line 5: Default values for type parameters have only been introduced recently (https://peps.python.org/pep-0696/)

crates/ty_python_semantic/src/types.rs
```diff
@@ -5708,9 +5708,7 @@ impl<'db> Type<'db> {
...
+                    let instance = Type::instance(db, class.unknown_specialization(db));
```
AlexWaygood, line 5711: I wondered if `class.top_materialization()` would work here, but it looks like under the hood that uses `default_specialization()`, so it suffers from the same issue as our current logic on `main`. (It probably _shouldn't_ use `default_specialization()`?)

dcreager: > It probably _shouldn't_ use `default_specialization()`

I think you're right — it should translate typevars into the bound/constraints in covariant position, and to `Never` (the implicit lower bound of every typevar) in contravariant position. (With the caveat that we can't really express the constraints of a constrained typevar as a single type — we'd need a "one-of" connective instead of union, and an "instance of this type but not any subtypes".)

sharkdp: I opened https://github.com/astral-sh/ty/issues/1164

PR #20328, opened 2025-09-10
[ty] Patch (implicit) `Self` on fallback types

sharkdp:


crates/ty_python_semantic/src/types.rs
```diff
@@ -10817,6 +10856,29 @@ impl<'db> TypeIsType<'db> {
...
+        if is_known_base(base) {
+            break;
+        }
+
+        if let Some(superclass) = base.into_class() {
+            upper_bound = superclass;
+        }
+    }
+
+    Type::instance(db, upper_bound)
```
dcreager: I think you can use `take_while` to simplify this:

```suggestion
    let upper_bound = class_literal
        .iter_mro(db, specialization)
        .take_while(|base| !is_known_base(base))
        .filter_map(ClassBase::into_class)
        .last()
        .unwrap_or_else(|| class_literal.unknown_specialization(db));
    Type::instance(db, upper_bound)
```

crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -270,18 +270,36 @@ reveal_type(Person._field_defaults)  # revealed: dict[str, Any]
 reveal_type(Person._fields)  # revealed: tuple[str, ...]
 reveal_type(Person._make)  # revealed: bound method <class 'Person'>._make(iterable: Iterable[Any]) -> Person
 reveal_type(Person._asdict)  # revealed: def _asdict(self) -> dict[str, Any]
-reveal_type(Person._replace)  # revealed: def _replace(self, **kwargs: Any) -> Self@_replace
+reveal_type(Person._replace)  # revealed: def _replace(self, **kwargs: Any) -> Self
```
dcreager: This loses the `@_replace` suffix because you give the newly synthesized typevar a binding context of `BindingContext::Synthesized`. I think it would be better to copy the binding context over from the `NamedTupleFallback` method. (The binding context is primarily used to differentiate between different uses of the same legacy typevar, so I don't think anything will _break_ without this change, but I think it's (a) more correct and (b) not substantially more difficult to implement, so I think it's worth doing.)

To do this, you'd need to update your new `TypeMapping::ReplaceSelf` to take in the new upper bound, rather than a fully constructed `BoundTypeVarInstance`, so that you could copy over all of the other fields from the typevar being replaced.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -466,7 +466,22 @@ impl<'db> Signature<'db> {
...
+                                *new_self
+                            } else {
+                                *var
+                            }
+                        }),
+                    )
+                })
+            } else {
+                self.generic_context
+            },
```
dcreager: :+1: You might consider a new `TypeMapping::updates_generic_context` method or similar (which would currently only return true for `ReplaceSelf`), and use that in the `if` check. That would hopefully make it more obvious how to add any future type mappings that also need to engage this new logic.

sharkdp: Yes, good suggestion. There might be a second type mapping that also needs to modify the generic context (as you mentioned in one of our chats, `BindSelf` should remove `Self` from the generic context, see https://github.com/astral-sh/ruff/pull/20364).

sharkdp:
I will merge this without approval, because I want to have an easier way of rebasing https://github.com/astral-sh/ruff/pull/18007 on this. Happy to address any post-merge comments.

dcreager:
> I will merge this without approval

:+1: Looks good!

PR #20333, opened 2025-09-10
[ty] use Type::Divergent to avoid panic in infinitely-nested-tuple implicit attribute

carljm:
## Summary

Use `Type::Divergent` to avoid "too many iterations" panic on an infinitely-nested tuple in an implicit instance attribute.

The regression here is from checking all tuple elements to see if they contain a Divergent type. It's 5% on one project, 1% on another, and zero on the rest. I spent some time looking into eliminating this regression by tracking a flag on inference results to note if they could possibly contain any Divergent type, but this doesn't really work -- there are too many different ways a type containing a Divergent type could enter an inference result. Still thinking about whether there are other ways to reduce this. One option is if we see certain kinds of non-atomic types that are commonly expensive to check for Divergent, we could make `has_divergent_type` a Salsa query on those types.

## Test Plan

Added mdtest.

Co-authored-by: Alex Waygood <Alex.Waygood@Gmail.com>

### DIFF
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -387,19 +387,20 @@ impl<'db> InferenceRegion<'db> {
     }
 }
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq, get_size2::GetSize)]
-enum CycleRecovery {
+#[derive(Debug, Clone, Copy, Eq, PartialEq, get_size2::GetSize, salsa::Update)]
+enum CycleRecovery<'db> {
     /// An initial-value for fixpoint iteration; all types are `Type::Never`.
     Initial,
     /// A divergence-fallback value for fixpoint iteration; all types are `Divergent`.
-    Divergent,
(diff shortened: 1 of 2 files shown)


carljm:
> All tests seem to pass if I apply this refactor to your branch... is there anything conceptually wrong with this?

There sort of is; I'd actually started the same refactor and then stopped when I reached the "conceptually wrong" thing, which is the implementation of `CycleRecovery::merge`. When we merge two `CycleRecovery::Divergent` together, with potentially different scopes, that should never change the scope of whichever one we're mutating. It makes `CycleRecovery::merge` not a symmetric operation.

That said, I _think_ your implementation actually achieves the necessary invariant in practice, but it's a bit subtle, and due solely to the way the clauses of the match pattern union are ordered, and the fact that `TypeInferenceBuilder::extend_cycle_recovery` always calls `CycleRecovery::merge` on `self.cycle_recovery`, not `other.cycle_recovery`. So it may still make sense to apply this change, just with a bit more explicit commentary about that being an important invariant.

PR #20337, opened 2025-09-10
[ty] Integrate type context for bidirectional inference

ibraheemdev:
## Summary

This may need some performance work.


crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -311,34 +327,103 @@ pub(super) fn infer_same_file_expression_type<'db>(
 ///
 /// Use [`infer_same_file_expression_type`] if it is guaranteed that  `expression` is in the same
 /// to avoid unnecessary salsa ingredients. This is normally the case inside the `TypeInferenceBuilder`.
-#[salsa::tracked(cycle_fn=single_expression_cycle_recover, cycle_initial=single_expression_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
 pub(crate) fn infer_expression_type<'db>(
     db: &'db dyn Db,
     expression: Expression<'db>,
+    tcx: TypeContext<'db>,
```
ibraheemdev, line 260: We never actually call `infer_expression_type` with a non-empty `TypeContext` currently, I'm not sure if that's something that will change in the future?

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -2084,8 +2177,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 },
             ) = node
             {
-                let value_ty = self.infer_expression(value);
-                let slice_ty = self.infer_expression(slice);
+                let value_ty = self.infer_expression(value, TypeContext::default());
+                let slice_ty = self.infer_expression(slice, TypeContext::default());
```
ibraheemdev: The functions (and salsa queries) that now take a `TypeContext` are most often called with an empty one, but I'm not sure if we want to have separate functions or make the lack of a `TypeContext` explicit on every call.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -6263,12 +6415,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
         let mut call_arguments =
             CallArguments::from_arguments(self.db(), arguments, |argument, splatted_value| {
-                let ty = self.infer_expression(splatted_value);
+                let ty = self.infer_expression(splatted_value, TypeContext::default());
                 self.store_expression_type(argument, ty);
                 ty
             });
 
-        let callable_type = self.infer_maybe_standalone_expression(func);
+        // TODO: Use the type context for more precise inference.
```
ibraheemdev: I left this comment in a few notable places where we now have a `TypeContext` that we could use for more precise inference.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -312,34 +328,108 @@ pub(super) fn infer_same_file_expression_type<'db>(
...
     db: &'db dyn Db,
     expression: Expression<'db>,
+    tcx: TypeContext<'db>,
 ) -> Type<'db> {
-    let file = expression.file(db);
+    infer_expression_type_impl(db, InferExpression::new(db, expression, tcx))
+}
+
+#[salsa::tracked(cycle_fn=single_expression_cycle_recover, cycle_initial=single_expression_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
+fn infer_expression_type_impl<'db>(db: &'db dyn Db, input: InferExpression<'db>) -> Type<'db> {
```
ibraheemdev, line 266: I moved the queries into private functions because the `InferExpression` supertype is mostly an internal implementation detail.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -312,34 +328,108 @@ pub(super) fn infer_same_file_expression_type<'db>(
...
 ) -> Type<'db> {
     Type::Never
 }
 
+/// An `Expression` with an optional `TypeContext`.
+///
+/// This is a Salsa supertype used as the input to `infer_expression_types` to avoid
+/// interning an `ExpressionWithContext` unnecessarily when no type context is provided.
+#[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, salsa::Supertype, salsa::Update)]
+enum InferExpression<'db> {
```
ibraheemdev, line 296: We also discussed splitting the queries into variants with and without a `TypeContext` as a potential workaround, but this supertype is effectively a more convenient way of doing that (internally, Salsa essentially creates a query for each variant of the input enum). The performance impact of this PR seems to be very minor with this optimization (and without it, there was a noticeable 2-3% regression).

crates/ty_python_semantic/src/types/infer.rs
carljm: Hmm. This PR is great, and it's awesome that it has negligible performance impact. And now this question is making me wonder if we don't even need this and I mis-guided you last week? I was thinking the RHS of assignment statements was a standalone expression, but it only is for un-annotated assignment statements, not for annotated assignments. And it's only annotated assignments and call arguments where we currently intend to apply a type context, but neither of those are standalone expressions. So it seems like at least for the initial scope here, we only need to pass type contexts internally in a `TypeInferenceBuilder`?

Unless I'm missing something here (@dcreager?), maybe the thing to take from this PR is that if we ever need additional context on `infer_type_expression`, now we know it can be done without a major performance regression -- but for now we should put it on ice and focus on just the internal type-contexts we need to actually implement bidirectional inference within an annotated assignment and for call arguments. (This will still involve the need to accommodate potentially multiple inferred types for a single expression, in the case of call arguments to unions of callables or overloaded callables.)

crates/ty_python_semantic/src/types/infer.rs
dcreager: > this supertype is effectively a more convenient way of doing that (internally, Salsa essentially creates a query for each variant of the input enum).

That's a nice trick, I didn't realize that about how salsa handles the enum input

crates/ty_python_semantic/src/types/infer.rs
dcreager: > Unless I'm missing something here

I don't see anything you're missing

sharkdp: > I was thinking the RHS of assignment statements was a standalone expression, but it only is for un-annotated assignment statements, not for annotated assignments.

Isn't the right hand side of something like `self.t: MyTypedDict = {…}` a standalone expression (see [here](https://github.com/astral-sh/ruff/blob/c6b92b918e4d350e4f8aed95a41985f80ca7dfc9/crates/ty_python_semantic/src/semantic_index/builder.rs#L1647) and [here](https://github.com/astral-sh/ruff/blob/8a0edf0da8342ae3b9fcf8d183fc289b52d64f86/crates/ty_python_semantic/src/types/infer.rs#L4629) and [here](https://github.com/astral-sh/ruff/blob/8a0edf0da8342ae3b9fcf8d183fc289b52d64f86/crates/ty_python_semantic/src/types/infer.rs#L4721))?

Maybe that's only needed for something like `self.MY_CONSTANT: Final = 17` where the annotation contains no type (only a qualifier) though? Which could mean that it's not important for purposes of bidirectional inference.

carljm: Aha, I _was_ missing something, should have looked more carefully last night, thanks @sharkdp. I found the `Final` case yesterday but we indeed don't need to apply any type context in that scenario (at least I don't think we currently have plans for a bare `Final` or `ClassVar` annotation to affect inference of the RHS). But that doesn't matter -- we now make annotated assignment RHS a standalone expression whenever it's inside a method, so as you point out, even `self.x: MyTypedDict = {...}` inside a method (or for that matter, `x: MyTypedDict = {...}` inside a method) will have a standalone expression on the RHS. So I think we do need this PR. And we should (at some point) pass a non-empty `TypeContext` through in those places in `infer_annotated_assignment_statement` and `infer_annotated_assignment_definition` that @sharkdp linked.

crates/ty_python_semantic/src/types/infer.rs
carljm: I don't have strong feelings here -- there's maybe some advantage to making it explicit, and avoiding another axis on which to proliferate separate inference functions?

crates/ty_python_semantic/src/types/infer.rs
carljm: It looks like already in this PR, we are (correctly) passing a non-empty `TypeContext` into `infer_maybe_standalone_expression` from `infer_annotated_assignment_statement` and `infer_annotated_assignment_definition`, which passes it through to `infer_standalone_expression_impl`, which passes it to `infer_expression_types`.

I think I mis-interpreted this question from @ibraheemdev entirely last night -- it's actually a narrow question specifically about this `infer_expression_type` wrapper function, not about the `infer_expression_types` Salsa query generally. I think the answer to that question is that I don't immediately see where we'll use `infer_expression_type` with a non-empty `TypeContext`, but it still seems reasonable that it should accept one for consistency.

Sorry for mis-reading things here and not reviewing more carefully here last night; shouldn't do reviews when short on time. But I think we are back to concluding that we should go ahead with this PR as-is, and hopefully some of the discussion above was useful context anyway 😆

crates/ty_python_semantic/src/types/infer.rs
MichaReiser: This is a nice solution!

PR #20338, opened 2025-09-10
[ty] Fix `CallableTypeOf[…]` for bound methods

sharkdp:
## Summary

Very confusing if the tools you are using to debug something are themselves buggy :smile:. `CallableTypeOf[bound_method]` would previously bind `self` to the bound method type itself, instead of binding it to the instance type stored inside the bound method type.

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/ty_extensions.md
+++ b/crates/ty_python_semantic/resources/mdtest/ty_extensions.md
@@ -469,6 +469,8 @@ c4: CallableTypeOf[()]
 Using it in annotation to reveal the signature of the callable object:
 
 ```py
+from typing_extensions import Self
+
 class Foo:
     def __init__(self, x: int) -> None:
         pass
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/ty_extensions.md
```diff
@@ -491,4 +497,6 @@ def _(
     reveal_type(c4)  # revealed: (...) -> Foo
 
     reveal_type(c5)  #  revealed: (x: int) -> str
+
+    reveal_type(c6)  # revealed: (x: int) -> Foo
```
sharkdp, line 501: This would previously reveal `(x: int) -> bound method Foo.returns_self(x: int) -> Foo` instead of `(x: int) -> Foo` (https://play.ty.dev/f4b4b1a6-ae47-4338-9be4-07134afc237a)

AlexWaygood (approved):

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -10989,8 +10989,10 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     .expect("`Bindings` should have at least one `CallableBinding`");
 
                 let mut signature_iter = callable_binding.into_iter().map(|binding| {
-                    if argument_type.is_bound_method() {
-                        binding.signature.bind_self(self.db(), Some(argument_type))
+                    if let Some(bound_method) = argument_type.into_bound_method() {
+                        binding
+                            .signature
+                            .bind_self(self.db(), Some(bound_method.self_instance(db)))
```
sharkdp, line 10995: I just noticed that I might want to call `bound_method.typing_self_type(db)` instead to handle classmethods correctly. Will take a look.

PR #20340, opened 2025-09-10
[ty] Add dedicated variant for `NominalInstance(object)`

dcreager:
Previously, `Type::object` would find the definition of the `object` class in typeshed, load that in (to produce a `ClassLiteral` and `ClassType`), and then create a `NominalInstance` of that class.

It's possible that we are using a typeshed that doesn't define `object`. We will not be able to do much useful work with that kind of typeshed, but it's still a possibility that we have to support at least without panicking. Previously, we would handle this situation by falling back on `Unknown`.

In most cases, that's a perfectly fine fallback! But `object` is also our top type — the type of all values. `Unknown` is _not_ an acceptable stand-in for the top type.

This PR adds a new `NominalInstance` variant for "instances of `object`". Unlike other nominal instances, we do not need to load in `object`'s `ClassType` to instantiate this variant. We will use this new variant even when the current typeshed does not define an `object` class, ensuring that we have a fully static representation of our top type at all times.

There are several operations that need access to a nominal instance's class, and for this new `object` variant we load it lazily only when it's needed. That means this operation is now fallible, since this is where the "typeshed doesn't define `object`" failure shows up.

This new approach also has the benefit of avoiding some salsa cycles that were cropping up while I was debugging #20093, since the new constraint set representation was trying to instantiate `Type::object` while in the middle of processing its definition in typeshed. Cycle handling was kicking in correctly and returning the `Unknown` fallback mentioned above. But the constraint set implementation depends on `Type::object` being a distinct and fully static type, highlighting that this is a correctness fix, not just an optimization fix.

### DIFF
--- a/crates/ty_python_semantic/src/types/instance.rs
+++ b/crates/ty_python_semantic/src/types/instance.rs
@@ -382,6 +382,9 @@ impl<'db> NominalInstanceType<'db> {
         other: Self,
         visitor: &IsDisjointVisitor<'db, C>,
     ) -> C {
+        if self.is_object() || other.is_object() {
+            return C::unsatisfiable(db);
+        }
         let mut result = C::unsatisfiable(db);
         if let Some(self_spec) = self.tuple_spec(db) {
             if let Some(other_spec) = other.tuple_spec(db) {


crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -153,15 +181,39 @@ pub(super) fn walk_nominal_instance_type<'db, V: super::visitor::TypeVisitor<'db
...
             NominalInstanceInner::NonTuple(class) => class,
-        }
+            NominalInstanceInner::Object => return KnownClass::Object.try_to_class_literal(db),
+        };
+        let (class_literal, _) = class.class_literal(db);
+        Some(class_literal)
+    }
+
+    pub(super) fn known(&self, db: &'db dyn Db) -> Option<KnownClass> {
+        self.class(db).and_then(|class| class.known(db))
```
carljm: Technically we wouldn't need to load the class from typeshed here, for `Object`? We can just directly return `KnownClass::Object`. Unless you're concerned about inconsistency if `object` is missing from typeshed? (Frankly I'd be fine if we panicked in that scenario anyway.)

It just seems there could be some perf hiding here still, since checking whether a type is a certain known-class is a somewhat common operation.

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -153,15 +181,39 @@ pub(super) fn walk_nominal_instance_type<'db, V: super::visitor::TypeVisitor<'db
...
-    visitor.visit_type(db, nominal.class(db).into());
+    let Some(class) = nominal.class(db) else {
+        return;
+    };
+    visitor.visit_type(db, class.into());
 }
 
 impl<'db> NominalInstanceType<'db> {
-    pub(super) fn class(&self, db: &'db dyn Db) -> ClassType<'db> {
+    pub(super) fn class(&self, db: &'db dyn Db) -> Option<ClassType<'db>> {
```
carljm: Personally I think it would be fine if we panicked if there is no `object` at all in typeshed (or alternatively, fell back to some empty `object` type defined in `ty_extensions`?) It would certainly be more convenient if this didn't have to become fallible.

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -12,32 +12,51 @@ use crate::types::enums::is_single_member_enum;
...
 use crate::{Db, FxOrderSet};
 
 pub(super) use synthesized_protocol::SynthesizedProtocolType;
 
 impl<'db> Type<'db> {
+    pub(crate) fn object() -> Self {
+        Type::NominalInstance(NominalInstanceType(NominalInstanceInner::Object))
+    }
+
+    pub(crate) fn is_object(&self) -> bool {
```
AlexWaygood: nit

```suggestion
    pub(crate) const fn object() -> Self {
        Type::NominalInstance(NominalInstanceType(NominalInstanceInner::Object))
    }

    pub(crate) const fn is_object(&self) -> bool {
```

crates/ty_python_semantic/src/types/instance.rs
AlexWaygood: I'll co-sign this, and go further: I think I'd actually strongly prefer it if we panicked here when there was no `object` at all in typeshed. If it's a custom typeshed without `object`, I strongly suspect we'll panic in some way with a confusing error message -- and even if we don't, we won't have intuitive behaviour at all when checking any user code. Much better to panic early here with an intelligible error message than attempting to go further. Other type checkers such as mypy also crash quite early on if you omit certain fundamental types from typeshed, so we'd be far from alone in doing this. There's also an internal precedent for this in `TupleType::to_class_type()`: https://github.com/astral-sh/ruff/blob/59c8fda3f8f3bf2cc5c1ae34e7ca9dbea4d0278f/crates/ty_python_semantic/src/types/tuple.rs#L203-L207

As well as that, I just don't think the idea of a "nominal instance type with no class" makes much conceptual sense. The thing that makes an instance type nominal is the fact that the presence of a subtyping relation between two such instances is determined based on whether there is a subclass relations between their underlying classes. I think it's much easier to understand the concept of a nominal instance type -- and therefore much less error-prone for us to work with that concept -- if we maintain the invariant that a nominal instance type must always have a class backing it (even if that class is sometimes only lazily materialized).

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -153,15 +181,39 @@ pub(super) fn walk_nominal_instance_type<'db, V: super::visitor::TypeVisitor<'db
...
+        };
+        let (class_literal, _) = class.class_literal(db);
+        Some(class_literal)
+    }
+
+    pub(super) fn known(&self, db: &'db dyn Db) -> Option<KnownClass> {
+        self.class(db).and_then(|class| class.known(db))
+    }
+
+    pub(super) fn is_known(&self, db: &'db dyn Db, known_class: KnownClass) -> bool {
```
AlexWaygood: Could we rename this to:

```suggestion
    pub(super) fn has_known_class(&self, db: &'db dyn Db, known_class: KnownClass) -> bool {
```

?

An instance type might _have_ a known class, but that's a different question to asking whether it _is_ a known class

crates/ty_python_semantic/src/types/instance.rs
AlexWaygood: We could probably return `KnownClass::Tuple` here if the inner variant is `NominalInstanceInner::Tuple` too, without actually materializing the `ClassType`.

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -153,15 +181,39 @@ pub(super) fn walk_nominal_instance_type<'db, V: super::visitor::TypeVisitor<'db
...
             NominalInstanceInner::ExactTuple(tuple) => tuple.to_class_type(db),
             NominalInstanceInner::NonTuple(class) => class,
-        }
+            NominalInstanceInner::Object => return KnownClass::Object.try_to_class_literal(db),
+        };
+        let (class_literal, _) = class.class_literal(db);
+        Some(class_literal)
+    }
+
+    pub(super) fn known(&self, db: &'db dyn Db) -> Option<KnownClass> {
```
AlexWaygood: Similar to my comment regarding `is_known`, I'd prefer it if this could be named

```suggestion
    pub(super) fn known_class(&self, db: &'db dyn Db) -> Option<KnownClass> {
```

so that it's clear that it's the underlying class that is "known" rather than the _instance type_. (There could be multiple non-equivalent instance types with the same known class backing them, if that class is generic!)

crates/ty_python_semantic/src/types/instance.rs
dcreager: Good ideas, done

crates/ty_python_semantic/src/types/instance.rs
dcreager: Works for me! Especially given the precedent with `tuple`

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -161,9 +182,39 @@ impl<'db> NominalInstanceType<'db> {
...
+        match self.0 {
+            NominalInstanceInner::ExactTuple(_) => Some(KnownClass::Tuple),
+            NominalInstanceInner::NonTuple(class) => class.known(db),
+            NominalInstanceInner::Object => Some(KnownClass::Object),
+        }
+    }
+
+    pub(super) fn has_known_class(&self, db: &'db dyn Db, known_class: KnownClass) -> bool {
+        self.known_class(db) == Some(known_class)
+    }
```
AlexWaygood: nit: Add some short doc-comments for these APIs?

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -203,15 +254,13 @@ impl<'db> NominalInstanceType<'db> {
...
         }
     }
 
     /// Return `true` if this type represents instances of the class `builtins.object`.
-    pub(super) fn is_object(self, db: &'db dyn Db) -> bool {
-        match self.0 {
-            NominalInstanceInner::ExactTuple(_) => false,
-            NominalInstanceInner::NonTuple(class) => class.is_object(db),
-        }
+    pub(super) fn is_object(self) -> bool {
```
AlexWaygood:
```suggestion
    pub(super) const fn is_object(self) -> bool {
```

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -312,9 +362,10 @@ impl<'db> NominalInstanceType<'db> {
                 NominalInstanceInner::ExactTuple(tuple1),
                 NominalInstanceInner::ExactTuple(tuple2),
             ) => tuple1.is_equivalent_to_impl(db, tuple2, visitor),
-            (NominalInstanceInner::NonTuple(class1), NominalInstanceInner::NonTuple(class2)) => {
-                class1.is_equivalent_to_impl(db, class2, visitor)
-            }
+            (
+                NominalInstanceInner::Object | NominalInstanceInner::NonTuple(_),
+                NominalInstanceInner::Object | NominalInstanceInner::NonTuple(_),
+            ) => (self.class(db)).is_equivalent_to_impl(db, other.class(db), visitor),
```
AlexWaygood: we know that `object` is not generic; we can do this more cheaply:

```suggestion
            (NominalInstanceInner::Object, NominalInstanceInner::Object) => C::always_satisfiable(db),
            (NominalInstanceInner::NonTuple(class1), NominalInstanceInner::NonTuple(class2)) => {
                class1.is_equivalent_to_impl(db, class2, visitor)
            }
```

AlexWaygood:
> Memory usage changes were detected when running on open source projects
> 
> ```diff
> trio (https://github.com/python-trio/trio)
> -     memo metadata = ~23MB
> +     memo metadata = ~22MB
> 
> sphinx (https://github.com/sphinx-doc/sphinx)
> -     memo metadata = ~40MB
> +     memo metadata = ~38MB
> ```

This is pretty cool!

crates/ty_python_semantic/src/types/instance.rs
dcreager: I added a similar case up in `has_relation_to_impl` for `(_, Object)`, since we know that is always true without having to walk the MROs.

crates/ty_python_semantic/src/types/instance.rs
dcreager: Done

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -291,6 +344,7 @@ impl<'db> NominalInstanceType<'db> {
         visitor: &HasRelationToVisitor<'db, C>,
     ) -> C {
         match (self.0, other.0) {
+            (_, NominalInstanceInner::Object) => C::always_satisfiable(db),
```
AlexWaygood, line 347: you can add a similar short-circuit branch for `is_disjoint_from_impl`: if either `self` or `other` is `object` in `NominalInstanceType::is_disjoint_from_impl`, you can immediately return `C::unsatisfiable(db)`

dcreager: Done

PR #20343, opened 2025-09-11
[`flake8-simplify`] Detect unnecessary `None` default for additional key expression types (`SIM910`)

### LINKED ISSUE #20341: SIM910 doesn't recognize `get` access on dictionary when the key is a property/method call
### Summary

For this case:

```python
ages = {"Tom": 23, "Maria": 23, "Dog": 11}
key_source = {"Thomas": "Tom"}
age = ages.get(key_source.get("Thomas", "Tom"), None)
```

I'd expect SIM910 to detect that the `None` is unnecessary

### Version

ruff 0.13.0

danparizher:
## Summary

Fixes #20341


crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM910_preview.py
```diff
@@ -0,0 +1,32 @@
...
+        self.name = "Tom"
+
+data = Data()
+ages = {"Tom": 23, "Maria": 23, "Dog": 11}
+age = ages.get(data.name, None)
+
+# Complex expression as key
+ages = {"Tom": 23, "Maria": 23, "Dog": 11}
+key = "Tom" if True else "Maria"
+age = ages.get(key, None)
```
ntBre: This is actually just another `Expr::Name` as key since it's assigned to a variable. Even the if expression itself is only a single `Expr::If`, but this is fine either way!

ntBre: (I inlined this too while I was at it)

PR #20344, opened 2025-09-11
[ty] Rename "possibly-unbound" diagnostics to "possibly-undeclared"

Renkai:
this fixes https://github.com/astral-sh/ty/issues/636


Renkai:
will use other names

PR #20347, opened 2025-09-11
[ty] Minor fixes to `Protocol` tests

AlexWaygood:


crates/ty_python_semantic/resources/mdtest/expression/len.md
```diff
@@ -260,7 +260,7 @@ class SecondRequiredArgument:
     def __len__(self, v: int) -> Literal[1]:
         return 1
 
-# TODO: Emit a diagnostic
+# this is fine: the call succeeds at runtime since the second argument is optional
 reveal_type(len(SecondOptionalArgument()))  # revealed: Literal[0]
```
AlexWaygood, line 264: we could consider emitting a diagnostic on the definition of `__len__` itself here (but that should be a disabled-by-default lint rule, IMO!), but I see no reason why we should emit a diagnostic at the `len()` call itself here. `SecondOptionalArgument` is a subtype of `Sized`.

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -77,7 +77,7 @@ from typing import Protocol, TypeVar
 T = TypeVar("T")
 
 class CanIndex(Protocol[T]):
-    def __getitem__(self, index: int) -> T: ...
+    def __getitem__(self, index: int, /) -> T: ...
```
AlexWaygood, line 80: The intent of this test is that `list[str]` should be a subtype of `CanIndex`; below, an instance of `list[str]` is passed into `takes_in_protocol`. But as written, `list[str]` is not a subtype of `CanIndex`, since all parameters of `list.__getitem__` are positional-only

crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -1813,9 +1813,28 @@ class Foo:
...
+class Meta(type):
+    def __iter__(self) -> Iterator[int]:
+        yield from range(42)
+
+class IterableClass(metaclass=Meta):
+    def __iter__(self) -> Iterator[str]:
+        yield from "abc"
+
+static_assert(is_subtype_of(TypeOf[IterableClass], Iterable[int]))
+```
```
AlexWaygood: A missing test that demonstrates another reason why it's important that dunder methods especially must be looked up on the meta-type when determining protocol subtyping/assignability

crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -1813,9 +1813,28 @@ class Foo:
 static_assert(not is_assignable_to(Foo, Iterable[Any]))
 ```
 
-Because method members must always be available on the class, it is safe to access a method on
-`type[P]`, where `P` is a protocol class, just like it is generally safe to access a method on
-`type[C]` where `C` is a nominal class:
+Because method members are always looked up on the meta-type of an object when testing assignability
+and subtyping, we understand that `IterableClass` here is a subtype of `Iterable` even though
+`Foo.__iter__` resolves to a type with the wrong signature:
```
carljm:
```suggestion
and subtyping, we understand that `IterableClass` here is a subtype of `Iterable[int]` even though
`IterableClass.__iter__` has the wrong signature:
```

PR #20349, opened 2025-09-11
[ty] Inlay hint call argument location

MatthewMckee4:
## Summary

Adds location to the inlay hint label part for call arguments. This allows "goto" on the inlay hint.

[inlay-hint-call-argument-location.webm](https://github.com/user-attachments/assets/9458c4ce-8cd1-4edb-bac6-4379ae913075)

This has not been added to the playground as Monaco doesn't support it.

## Test Plan

updated a test in `ty_server`. But other than that im not sure the best way to approach tests for this.

### DIFF
--- a/crates/ty_ide/src/signature_help.rs
+++ b/crates/ty_ide/src/signature_help.rs
@@ -382,7 +382,7 @@ mod tests {
             f = func_a
         else:
             f = func_b
-
+        
         f(<CURSOR>
         "#,
         );
@@ -427,10 +427,10 @@ mod tests {
 
         @overload
         def process(value: int) -> str: ...
-
+        
         @overload
         def process(value: str) -> int: ...
-
+        
         def process(value):
             if isinstance(value, int):
                 return str(value)
@@ -827,10 +827,10 @@ def ab(a: int, *, c: int):
             r#"
         class Point:
             """A simple point class representing a 2D coordinate."""
-
+            
             def __init__(self, x: int, y: int):
                 """Initialize a point with x and y coordinates.
-
+                
                 Args:
                     x: The x-coordinate
                     y: The y-coordinate
@@ -962,12 +962,12 @@ def ab(a: int, *, c: int):
             r#"
         from typing import overload
 
-        @overload
+        @overload  
         def process(value: int) -> str: ...
-
+        
         @overload
         def process(value: str, flag: bool) -> int: ...
-
+        
         def process(value, flag=None):
             if isinstance(value, int):
                 return str(value)


MatthewMckee4:
Getting this error in playground. It seems like i have broken the goto as well as the inlay hint location still not working. will look into this more

```bash
Uncaught Error: can't access property "monaco", this is undefined

mapNavigationTarget@http://localhost:5173/src/Editor/Editor.tsx:464:13
mapNavigationTargets@http://localhost:5173/src/Editor/Editor.tsx:495:22
provideDefinition@http://localhost:5173/src/Editor/Editor.tsx:391:21
```

dhruvmanila:
> Getting this error in playground. It seems like i have broken the goto as well as the inlay hint location still not working. will look into this more

I'd be happy to keep the playground changes in a separate PR to get the LSP changes in.

MatthewMckee4:
I fixed this error, was a small issue, I fixed in the "Fix Playground" commit

MichaReiser:
> I fixed this error, was a small issue, I fixed in the "Fix Playground" commit

Nice. Does that mean this PR is now ready for review or is there more to do?

MatthewMckee4:
I am just looking over it for the final time now, will make it ready for review today

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -893,15 +922,22 @@ pub fn find_active_signature_from_details(
...
-    pub argument_names: HashMap<usize, String>,
+pub struct InlayHintCallArgumentDetails {
+    /// The file containing the signature derived from the [`ast::ExprCall`].
+    ///
+    /// This is [`None`] if the signature does not have a definition.
+    /// This should only happen if [`definition_parameter_offsets`] is incomplete.
+    pub target_signature_file: Option<File>,
+
+    /// The position of the arguments mapped to their name and the range of the argument definition in the signature.
+    pub argument_names: HashMap<usize, (String, Option<TextRange>)>,
```
MatthewMckee4: I could clean this up more by using `Option<FileRange>`, but this requires a lot more cloning

MatthewMckee4: Yeah I'm not too sure about this. I'm not sure if the cloning of a `File` is "too" expensive. This would simplify the code a bit but not too much really

MichaReiser:
@Gankra is this a PR you could take a look at? If not, that's fine. It might just take me a few days before I get to it because I'm still catching up with notifications after my PTO (and have some other time sensitive things that I need to do first)

crates/ty_python_semantic/src/types/ide_support.rs
Gankra: Cloning this kind of File is essentially free, aiui.

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -804,6 +807,23 @@ pub struct CallSignatureDetails<'db> {
...
+) -> Option<HashMap<String, TextRange>> {
+    match definition_kind {
+        DefinitionKind::Function(node) => Some(
+            node.node(module_ref)
+                .parameters
+                .iter()
+                .map(|param| (param.name().to_string(), param.name().range()))
+                .collect(),
+        ),
+        // Todo: lambda functions
```
Gankra:
```suggestion
        // TODO: lambda functions
```

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -791,6 +791,9 @@ pub struct CallSignatureDetails<'db> {
     /// within the full signature string.
     pub parameter_label_offsets: Vec<TextRange>,
 
+    /// Offsets for each parameter in the signature definition.
+    pub definition_parameter_offsets: Option<HashMap<String, TextRange>>,
```
Gankra: Hm is anything added by this being an Option?

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -935,16 +976,30 @@ pub fn inlay_hint_function_argument_details<'db>(
             continue;
         };
 
+        let parameter_label_offset =
+            definition_parameter_offsets
+                .as_ref()
+                .and_then(|offsets| match param.name() {
+                    Some(name) => offsets.get(&name.to_string()),
+                    None => None,
+                });
```
Gankra: I think the body of the and_then can be `offsets.get(&param.name()?.to_string())`

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: No not really, removed the option now

MatthewMckee4:
hmm there seems to be something quite wrong about the location for the parameters (in playground), and i think there's an issue with my video above

MatthewMckee4:
working fine in vscode, just need to look into the playground not working

MatthewMckee4:
really have no clue why it isnt working

crates/ty_python_semantic/src/types/ide_support.rs
MichaReiser: `FileRange` is even marked as `Copy` (it's as expensive as copying a u128)

crates/ty_wasm/src/lib.rs
```diff
@@ -918,6 +923,28 @@ pub struct LocationLink {
...
+    target: &NavigationTarget,
+    db: &dyn Db,
+    position_encoding: PositionEncoding,
+    source_range: Option<Range>,
+) -> LocationLink {
+    LocationLink {
+        path: target.file().path(db).to_string(),
+        full_range: Range::from_file_range(
+            db,
+            FileRange::new(target.file(), target.full_range()),
```
MichaReiser: Nit: We could make this a method on `target` (`target.full_file_range`)

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: Okay I'll update this code then

playground/ty/src/Editor/Editor.tsx
```diff
@@ -433,7 +434,13 @@ class PlaygroundServer
     return {
       dispose: () => {},
       hints: inlayHints.map((hint) => ({
-        label: hint.markdown,
+        label: hint.label.map((part) => ({
+          label: part.label,
+          location:
```
MichaReiser: Given that this doesn't work in Monaco (it seems) I'm inclined to not set location and instead at a comment saying that. As of today (date), location isn't supported by Monaco which is why the field is ommited in the playground.

MatthewMckee4: yeah that makes sense, will update. Thanks

MatthewMckee4: I think it would be good to keep all of the updates to `ty_wasm`, just incase in the future monaco supports this, what do you think?

MichaReiser: Yeah, I think that's fine. I also don't expect the ty_wasm code to add too much overhead. It only maps a few locations

MichaReiser:
@Gankra would you mind taking another look at this PR (The playground changes look good to me)

MatthewMckee4:
I understand this isn't important for the beta but just making sure this doesn't get lost.

MichaReiser:
Thanks. I think it got lost 😓 I'll take another look tomorrow and another ping to @Gankra

MatthewMckee4:
Thank you

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -468,6 +553,18 @@ mod tests {
...
+        info[inlay-hint-location]: Inlay Hint Target
+         --> main.py:3:24
+          |
+        2 | class A:
+        3 |     def __init__(self, y):
+          |                        ^
+        4 |         self.x = 1
+        5 |         self.y = y
+          |
+        info: For inlay hint label 'y' at 112..113
```
MichaReiser: I find the byte offsets here hard to understand. This seems to be a perfect use case for a diagnostic with two spans:

* The first span points to the source (the `y` in the call)
* The second span points to the target (the `y` in the function declaration)

I think that will make it much easier to read those snapshots

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -869,6 +869,9 @@ pub struct CallSignatureDetails<'db> {
     /// within the full signature string.
     pub parameter_label_offsets: Vec<TextRange>,
 
+    /// Offsets for each parameter in the signature definition.
+    pub definition_parameter_offsets: HashMap<String, FileRange>,
```
MichaReiser: Do we need to use a `FileRange` here? How does `parameter_label_offset` get away with using `TextRange`? Does it use the `Definition.file`?

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -913,12 +933,29 @@ pub fn call_signature_details<'db>(
                 let display_details = signature.display(db).to_string_parts();
                 let parameter_label_offsets = display_details.parameter_ranges;
                 let parameter_names = display_details.parameter_names;
+                let definition_parameter_offsets = signature
+                    .definition()
+                    .and_then(|definition| {
```
MichaReiser: Given that `CallSignatureDetails` stores the function's definition: Can't we compute the definition to parameter offset mapping on demand rather than storing it on `CallSignatureDetails`? What I mean by that: We could have a method `get_parameter_range(self, db, name: &str) -> Option<FileRange>` that calls `parsed_module` ad-hoc, searches the parameter, and returns it. 

Or we can build this map closer to where we use it rather than building it for every `CallSignatureDetails` even when it's never used.

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -1099,16 +1140,21 @@ pub fn inlay_hint_function_argument_details<'db>(
             continue;
         };
 
+        let parameter_label_offset = definition_parameter_offsets.get(&param.name()?.to_string());
```
MichaReiser: Given that we already map the parameters. Could we look up the parameter range by param_index instead of by name?

MatthewMckee4: The issue i was having was with bound methods with `self`. If i had a list then the index would be one off.

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: Yes we need the `File` and `TextRange` for `NavigationTarget`

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: Yes, good point

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: Yeah, fair point

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: Ah the one hard part about this was that when i annotate with a span
```rs
source.annotate(Annotation::primary(
                Span::from(self.source.file()).with_range(self.source.range()),
            ));
```

I dont get the actual inlay hints in it, so we just point to a "random" part in the file

crates/ty_python_semantic/src/types/ide_support.rs
MichaReiser: Fair, but can't we use `definition.file` instead of repeating the `File` for each parameter (which I think is always the same?)

crates/ty_ide/src/inlay_hints.rs
MichaReiser: > I dont get the actual inlay hints in it, so we just point to a "random" part in the file

Can you elaborate on this? Why would we get random parts in the file?

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: Yeah, i had that before. Changed in https://github.com/astral-sh/ruff/pull/20349/commits/20364f173946173210255be731b938b0bf71bdea. The code is a lot cleaner like this, since definition is an `Option` in `CallSignatureDetails`.

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: Yes sorry.Our extra inlay hint labels that we add in the output `[a=]`. We want to be able to add a `source` section in the diagnostics section of the snapshots. But when we try to use the original file to show the source of the inlay hint goto `a` is not in the file. So we cant really show a source, unless there is another way to do this for an arbitrary `String`.

MatthewMckee4: Sorry if this is hard to read, this is the new snapshots from when i try to include the source. We see thee `^` points to random places where the inlay hint label "would" be.
```
   18    25 │   |                                      ^
   19    26 │ 3 | foo(1, 'pos', 3.14, False, e=42)
   20    27 │ 4 | foo(1, 'pos', 3.14, e=42, f='custom')
   21    28 │   |
   22       │-info: For inlay hint label 'd' at 115..116
         29 │+info: Source
         30 │+ --> main.py:3:25
         31 │+  |
         32 │+2 | def foo(a: int, b: str, /, c: float, d: bool = True, *, e: int, f: str = 'default'): pass
         33 │+3 | foo(1, 'pos', 3.14, False, e=42)
         34 │+  |                         ^
         35 │+4 | foo(1, 'pos', 3.14, e=42, f='custom')
         36 │+  |
   23    37 │
   24    38 │ info[inlay-hint-location]: Inlay Hint Target
   25    39 │  --> main.py:2:28
   26    40 │   |
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
   28    42 │   |                            ^
   29    43 │ 3 | foo(1, 'pos', 3.14, False, e=42)
   30    44 │ 4 | foo(1, 'pos', 3.14, e=42, f='custom')
   31    45 │   |
   32       │-info: For inlay hint label 'c' at 146..147
         46 │+info: Source
         47 │+ --> main.py:4:23
         48 │+  |
         49 │+2 | def foo(a: int, b: str, /, c: float, d: bool = True, *, e: int, f: str = 'default'): pass
         50 │+3 | foo(1, 'pos', 3.14, False, e=42)
         51 │+4 | foo(1, 'pos', 3.14, e=42, f='custom')
         52 │+  |                       ^
         53 │+  |
```

MichaReiser: What about something like this

```patch
Index: crates/ty_ide/src/inlay_hints.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/crates/ty_ide/src/inlay_hints.rs b/crates/ty_ide/src/inlay_hints.rs
--- a/crates/ty_ide/src/inlay_hints.rs	(revision 1522cbd742f0e725a1d4725ecfcd465a4b50f7b1)
+++ b/crates/ty_ide/src/inlay_hints.rs	(date 1761135752041)
@@ -435,17 +435,10 @@
                 for part in hint.label.parts() {
                     hint_str.push_str(part.text());
 
-                    let label_length = part.text().len();
-
                     if let Some(target) = part.target() {
-                        let label_range = TextRange::new(
-                            TextSize::try_from(end_position).unwrap(),
-                            TextSize::try_from(end_position + label_length).unwrap(),
-                        );
-
                         diagnostics.push(InlayHintLocationDiagnostic::new(
                             part.text().to_string(),
-                            label_range,
+                            FileRange::new(self.file, TextRange::at(hint.position, TextSize::ZERO)),
                             target,
                         ));
                     }
@@ -555,16 +548,19 @@
         a.y[: Literal[3]] = 3
 
         ---------------------------------------------
-        info[inlay-hint-location]: Inlay Hint Target
+        info[inlay-hint-location]: Inlay Hint part y
          --> main.py:3:24
           |
         2 | class A:
         3 |     def __init__(self, y):
-          |                        ^
+          |                        - points here
         4 |         self.x = 1
         5 |         self.y = y
+        6 |
+        7 | a = A(2)
+          |       ^
+        8 | a.y = 3
           |
-        info: For inlay hint label 'y' at 112..113
         ");
     }
 
@@ -1132,7 +1128,7 @@
          6 |     return y
          7 |
          8 | def baz(a: int, b: str, c: bool): pass
-           |         ^
+           |     ^
          9 |
         10 | baz(foo(5), bar(bar('test')), True)
            |
@@ -1140,11 +1136,11 @@
 
         info[inlay-hint-location]: Inlay Hint Target
          --> main.py:2:9
-          |
-        2 | def foo(x: int) -> int:
+           |
+         2 | def foo(x: int) -> int:
           |         ^
-        3 |     return x * 2
-          |
+         3 |     return x * 2
+           |
         info: For inlay hint label 'x' at 133..134
 
         info[inlay-hint-location]: Inlay Hint Target
@@ -1153,7 +1149,7 @@
          6 |     return y
          7 |
          8 | def baz(a: int, b: str, c: bool): pass
-           |                 ^
+           |             ^
          9 |
         10 | baz(foo(5), bar(bar('test')), True)
            |
@@ -1161,24 +1157,24 @@
 
         info[inlay-hint-location]: Inlay Hint Target
          --> main.py:5:9
-          |
-        3 |     return x * 2
-        4 |
-        5 | def bar(y: str) -> str:
+           |
+         3 |     return x * 2
+         4 |
+         5 | def bar(y: str) -> str:
           |         ^
-        6 |     return y
-          |
+         6 |     return y
+           |
         info: For inlay hint label 'y' at 149..150
 
         info[inlay-hint-location]: Inlay Hint Target
          --> main.py:5:9
-          |
-        3 |     return x * 2
-        4 |
-        5 | def bar(y: str) -> str:
+           |
+         3 |     return x * 2
+         4 |
+         5 | def bar(y: str) -> str:
           |         ^
-        6 |     return y
-          |
+         6 |     return y
+           |
         info: For inlay hint label 'y' at 157..158
 
         info[inlay-hint-location]: Inlay Hint Target
@@ -1400,24 +1396,24 @@
         ---------------------------------------------
         info[inlay-hint-location]: Inlay Hint Target
          --> main.py:5:9
-          |
-        4 | @overload
-        5 | def foo(x: int) -> str: ...
+           |
+         4 | @overload
+         5 | def foo(x: int) -> str: ...
           |         ^
-        6 | @overload
-        7 | def foo(x: str) -> int: ...
-          |
+         6 | @overload
+         7 | def foo(x: str) -> int: ...
+           |
         info: For inlay hint label 'x' at 136..137
 
         info[inlay-hint-location]: Inlay Hint Target
          --> main.py:5:9
-          |
-        4 | @overload
-        5 | def foo(x: int) -> str: ...
+           |
+         4 | @overload
+         5 | def foo(x: int) -> str: ...
           |         ^
-        6 | @overload
-        7 | def foo(x: str) -> int: ...
-          |
+         6 | @overload
+         7 | def foo(x: str) -> int: ...
+           |
         info: For inlay hint label 'x' at 148..149
         ");
     }
@@ -1558,7 +1554,7 @@
                 "Inlay Hint Target".to_string(),
             );
             main.annotate(Annotation::primary(
-                Span::from(self.target.file()).with_range(self.target.range()),
+                    Span::from(self.target.file()).with_range(self.target.range()),
             ));
             main.info(format!(
                 "For inlay hint label '{}' at {:?}",
```

The alternative is that you:
* iterate over all hints and collect all parts
* Once you have the final buffer, write a new file to the in-memory file system
* Now create the diagnostics for all parts

MatthewMckee4: That looks good, updated now

MichaReiser:
Let me know when this is ready for re-review (you can request a review)

crates/ty_python_semantic/src/types/ide_support.rs
MatthewMckee4: added this now

PR #20351, opened 2025-09-11
[`flake8-simplify`] Mark guaranteed-bool fixes as safe `SIM103`

### LINKED ISSUE #20287: `needless-bool` (SIM103) could be safe for expressions guaranteed to return `bool`s
### Summary

The fix for [`needless-bool` (SIM103)](https://docs.astral.sh/ruff/rules/needless-bool/) is marked unsafe, but it could be marked safe when the replacement is a `bool` call, `not` expression, `in` expression, `not in` expression, `is` expression, or `is not` expression, all of which are guaranteed to return a `bool`. [Example](https://play.ruff.rs/c9353445-2d3a-4578-b1c4-00619a2bbf1f):
```console
$ cat >sim103.py <<'# EOF'
def f(x):
    if x:
        return True
    return False
print(f(0), f(1))

def f(x):
    if x:
        return False
    return True
print(f(0), f(1))

def f(x):
    if x in [0]:
        return True
    return False
print(f(0), f(1))

def f(x):
    if x in [0]:
        return False
    return True
print(f(0), f(1))

def f(x):
    if x not in [0]:
        return True
    return False
print(f(0), f(1))

def f(x):
    if x not in [0]:
        return False
    return True
print(f(0), f(1))

def f(x):
    if x is x:
        return True
    return False
print(f(0))

def f(x):
    if x is x:
        return False
    return True
print(f(0))

def f(x):
    if x is not x:
        return True
    return False
print(f(0))

def f(x):
    if x is not x:
        return False
    return True
print(f(0))
# EOF

$ python sim103.py
False True
True False
True False
False True
False True
True False
True
False
False
True

$ ruff --isolated check sim103.py --select SIM103 --unsafe-fixes --fix
Found 10 errors (10 fixed, 0 remaining).

$ python sim103.py
False True
True False
True False
False True
False True
True False
True
False
False
True
```

### Version

TaKO8Ki:
## Summary

Fixes #20287

Mark guaranteed-bool fixes as safe.

- Mark fixes as `Applicability::Safe` when the replacement is guaranteed boolean
    - not <expr>, identity/membership compares (is/is not/in/not in), builtin bool(...)
- Keep `Applicability::Unsafe` for equality/inequality (==/!=) and other non‑guaranteed cases
- Avoid offering a fix when bool is shadowed and expression isn’t guaranteed boolean

## Test Plan

Update the snapshot of `SIM103.py`

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_simplify/rules/needless_bool.rs
+++ b/crates/ruff_linter/src/rules/flake8_simplify/rules/needless_bool.rs
@@ -1,3 +1,4 @@
+use ruff_diagnostics::Applicability;
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::name::Name;
 use ruff_python_ast::traversal;
@@ -48,10 +49,15 @@ use crate::{Edit, Fix, FixAvailability, Violation};
 ///
 /// ## Fix safety
 ///
-/// This fix is marked as unsafe because it may change the program’s behavior if the condition does not
-/// return a proper Boolean. While the fix will try to wrap non-boolean values in a call to bool,
-/// custom implementations of comparison functions like `__eq__` can avoid the bool call and still
-/// lead to altered behavior.
+/// This rule provides safe fixes when the replacement expression is guaranteed to evaluate
+/// to a real boolean value – for example, a logical negation (`not`), an identity or
+/// membership comparison (`is`, `is not`, `in`, `not in`), or a call to the builtin
+/// `bool(...)`.
+///
+/// In other cases, the fix is marked as unsafe because it can change runtime behavior. In
+/// particular, equality and inequality comparisons (`==`, `!=`) may be overloaded to return
+/// non-boolean values. When `bool` is shadowed and the expression is not guaranteed to be
+/// boolean, no fix is offered.
 ///
 /// ## References
 /// - [Python documentation: Truth Value Testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/flake8_simplify/rules/needless_bool.rs
```diff
@@ -334,6 +274,155 @@ enum Bool {
...
+///
+/// Safety rules:
+/// - Safe if the replacement is guaranteed boolean:
+///   - a negation (`not <expr>`),
+///   - an identity/membership comparison (`is`, `is not`, `in`, `not in`), or
+///   - a call to the builtin `bool(<expr>)` (only when `bool` is not shadowed).
+/// - Unsafe for equality comparisons (`==`, `!=`), because user types may overload them to return
+///   non-boolean values.
+/// - When `bool` is shadowed and the expression is not guaranteed boolean, no fix is provided.
+fn build_replacement_and_safety(
```
TaKO8Ki, line 287: While some `if/else` chains were refactored into match expressions, the core logic for constructing the condition expression remains unchanged. This change only annotates applicability.

crates/ruff_linter/src/rules/flake8_simplify/rules/needless_bool.rs
```diff
@@ -334,6 +274,155 @@ enum Bool {
...
+        // `==`/`!=` which can be overloaded to return non-bool.
+        (
+            true,
+            Expr::Compare(ast::ExprCompare {
+                ops,
+                left,
+                comparators,
+                ..
+            }),
+        ) => match ops.as_ref() {
```
ntBre, line 342: Should we make this a guard like in the original code? That should allow us to combine the `_ => ` cases on lines 370 and 382 again.

crates/ruff_linter/src/rules/flake8_simplify/rules/needless_bool.rs
```diff
@@ -334,6 +274,155 @@ enum Bool {
...
+                op: ast::UnaryOp::Not,
+                ..
+            }) => (Some((**operand).clone()), Applicability::Safe),
+            Expr::Compare(ast::ExprCompare { ops, .. }) => {
+                let app = if is_identity_or_membership_ops(ops.as_ref()) {
+                    Applicability::Safe
+                } else {
+                    Applicability::Unsafe
+                };
+                (Some((**operand).clone()), app)
```
ntBre, line 326: I think we could make this a guard too:


```suggestion
            Expr::Compare(ast::ExprCompare { ops, .. }) if is_identity_or_membership_ops(ops.as_ref()) => {
                (Some((**operand).clone()), Applicability::Safe)
```

and fall through to the unsafe case below if the condition is false.

ntBre (commented):
Thank you! This looks good to me with only a couple of small nits on the code.

However, I think we need to make this a preview change because SIM103 is a stable rule. We can only demote fix safety in a patch release, not upgrade unsafe to safe. You can see [`preview.rs`](https://github.com/astral-sh/ruff/blob/c0fb235a703684bca1cc1767015bb7e19ab1c22f/crates/ruff_linter/src/preview.rs#L16-L19) for some examples of preview functions.


We'll need to update both the implementation and the docs to reflect the preview behavior. I like your new docs section, so I would probably recommend leaving the old one and just adding your new paragraphs, starting with `In preview, ...`.

dscorbett:
I now think that the fix should stay unsafe when it inserts a `bool` call. Although it preserves the original behavior, it is often unnecessary and is probably not wanted. Keeping it unsafe encourages manual review. For example, this:
```python
def is_odd_or_zero(x):
    if x % 2 != 0 or x == 0:
        return True
    return False
```
is fixed to this:
```python
def is_odd_or_zero(x):
    return bool(x % 2 != 0 or x == 0)
```
but what was almost certainly intended was this:
```python
def is_odd_or_zero(x):
    return x % 2 != 0 or x == 0
```

MichaReiser:
@TaKO8Ki sorry for another ping. Do you think you'll have time in the near future to look into the PR feedback? If not, that's completely fine, I'm just trying to understand the status of this PR

PR #20355, opened 2025-09-11
[ty] Remove the `Constraints` trait

dcreager:
This PR removes the `Constraints` trait. We removed the `bool` implementation several weeks back, and are using `ConstraintSet` everywhere. There have been discussions about trying to include the reason for an assignability failure as part of the result, but that there are no concrete plans to do so soon, and it's not clear that we'll need the `Constraints` trait to do that. (We can ideally just update the `ConstraintSet` type directly.)

In the meantime, this just complicates the code for no good reason.

This PR is a pure refactoring, and contains no behavioral changes.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -43,7 +43,7 @@
 //! — must be fully static. We take the bottom and top materializations of the types to remove any
 //! gradual forms if needed.
 //!
-//! NOTE: This module is currently in a transitional state: added the DNF [`ConstraintSet`]
+//! NOTE: This module is currently in a transitional state. We've added the DNF [`ConstraintSet`]
 //! representation, and updated all of our property checks to build up a constraint set and then
 //! check whether it is ever or always satisfiable, as appropriate. We are not yet inferring
 //! specializations from those constraints.


crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -110,86 +108,29 @@ fn incomparable<'db>(db: &'db dyn Db, left: Type<'db>, right: Type<'db>) -> bool
     !comparable(db, left, right)
 }
 
-/// Encodes the constraints under which a type property (e.g. assignability) holds.
-pub(crate) trait Constraints<'db>: Clone + Sized + std::fmt::Debug {
-    /// Returns a constraint set that never holds
-    fn unsatisfiable(db: &'db dyn Db) -> Self;
-
-    /// Returns a constraint set that always holds
-    fn always_satisfiable(db: &'db dyn Db) -> Self;
```
dcreager, line 119: These methods (and by extension, `from_bool`) didn't actually use their `db` parameter, so I've simplified the API a bit.  `from_bool` is now replaced with a `From<bool>` impl, and these two constructors are replaced with `ConstraintSet::from(true)` and `from(false)`.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -259,13 +216,67 @@ pub struct ConstraintSet<'db> {
...
+    pub(crate) fn union(&mut self, db: &'db dyn Db, other: Self) -> &Self {
+        self.union_set(db, other);
+        self
+    }
+
+    /// Updates this constraint set to hold the intersection of itself and another constraint set.
+    pub(crate) fn intersect(&mut self, db: &'db dyn Db, other: &Self) -> &Self {
+        self.intersect_set(db, other);
+        self
+    }
```
dcreager, line 249: There's now an annoying difference here, where `union` takes in an owned `other`, while `intersect` takes in a reference. This is purely due to internal details about when we can/cannot reuse the existing storage. This didn't exist before, because clippy couldn't see through the trait to trip the "you can pass this by reference" lint. I decided to go ahead and accept clippy's recommendation. If folks prefer, I would be equally happy putting an `#[expect]` here so that both methods take in an owned `other`.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -43,12 +43,10 @@
 //! — must be fully static. We take the bottom and top materializations of the types to remove any
 //! gradual forms if needed.
 //!
-//! NOTE: This module is currently in a transitional state: we've added a [`Constraints`] trait,
-//! and updated all of our type property implementations to work on any impl of that trait. We have
-//! added the DNF [`ConstraintSet`] representation, and updated all of our property checks to build
-//! up a constraint set and then check whether it is ever or always satisfiable, as appropriate. We
-//! are not yet inferring specializations from those constraints, and we will likely remove the
-//! [`Constraints`] trait once everything has stabilized.
+//! NOTE: This module is currently in a transitional state: added the DNF [`ConstraintSet`]
```
```suggestion
//! NOTE: This module is currently in a transitional state. We've added the DNF [`ConstraintSet`]
```

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: In terms of readability, I think I weakly prefer the old spelling, actually! `ConstraintSet::unsatisfiable()` seems clearer to me about what it means than `ConstraintSet::from(false)`, even if the constructor is strictly-speaking redundant with the new `From<bool>` implementation. Same for `ConstraintSet::always_satisfiable`.

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: I'm happy to genuflect to Clippy here!

crates/ty_python_semantic/src/types/constraints.rs
carljm: If it's a weak preference, I'll put in a vote in favor of the new spelling (or at least, in favor of not taking time to reintroduce the old methods and updating the PR to use them everywhere.)

dcreager: > or at least, in favor of not taking time to reintroduce the old methods and updating the PR to use them everywhere

I did that change in a separate commit, so the cost of reverting it would be ~zero if that sways your vote

carljm: I don't feel strongly either, but I think the new spelling is pretty clear, and I like that it generalizes well to cases where we already have a boolean value, not literal `true` or `false` (e.g. `ConstraintSet::from_bool(relation.is_assignability)`). So it seems like we will need to have (and be able to read and understand the concept of) `ConstraintSet::from_bool` anyway, which makes me dis-inclined to pad the API with special-case methods for literal `true` and `false`.

PR #20356, opened 2025-09-11
[`ty`] Include `NamedTupleFallback` members in `NamedTuple` instance completions

TaKO8Ki:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1161

Include `NamedTupleFallback` members in `NamedTuple` instance completions.

- Augment instance attribute completions when completing on NamedTuple instances by merging members from `_typeshed._type_checker_internals.NamedTupleFallback`

## Test Plan

Adds a minimal completion test `namedtuple_fallback_instance_methods`

### DIFF
--- a/crates/ty_python_semantic/src/types/ide_support.rs
+++ b/crates/ty_python_semantic/src/types/ide_support.rs
@@ -120,11 +120,11 @@ impl<'db> AllMembers<'db> {
             }
 
             Type::ClassLiteral(class_literal) => {
+                self.extend_with_class_members(db, ty, class_literal);
+
                 if CodeGeneratorKind::NamedTuple.matches(db, class_literal) {
                     self.extend_with_type(db, KnownClass::NamedTupleFallback.to_class_literal(db));
                 }
-
-                self.extend_with_class_members(db, ty, class_literal);
 
                 if let Type::ClassLiteral(meta_class_literal) = ty.to_meta_type(db) {
                     self.extend_with_class_members(db, ty, meta_class_literal);
@@ -140,8 +140,16 @@ impl<'db> AllMembers<'db> {
             }
 
             Type::SubclassOf(subclass_of_type) => {
-                if let Some(class_literal) = subclass_of_type.subclass_of().into_class() {
-                    self.extend_with_class_members(db, ty, class_literal.class_literal(db).0);
+                if let Some(class_type) = subclass_of_type.subclass_of().into_class() {
+                    let class_literal = class_type.class_literal(db).0;
+                    self.extend_with_class_members(db, ty, class_literal);
+
+                    if CodeGeneratorKind::NamedTuple.matches(db, class_literal) {
+                        self.extend_with_type(
+                            db,
+                            KnownClass::NamedTupleFallback.to_class_literal(db),
+                        );
+                    }
                 }
             }
 
--- a/crates/ty_python_semantic/resources/mdtest/ide_support/all_members.md
+++ b/crates/ty_python_semantic/resources/mdtest/ide_support/all_members.md
@@ -250,15 +250,26 @@ class Person(NamedTuple):
 
 static_assert(has_member(Person, "id"))
 static_assert(has_member(Person, "name"))
-static_assert(has_member(Person(1, "Name"), "id"))
-static_assert(has_member(Person(1, "Name"), "name"))
 
 static_assert(has_member(Person, "_make"))
 static_assert(has_member(Person, "_asdict"))
 static_assert(has_member(Person, "_replace"))
-static_assert(has_member(Person(1, "Name"), "_make"))
-static_assert(has_member(Person(1, "Name"), "_asdict"))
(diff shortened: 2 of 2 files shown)


crates/ty_ide/src/completion.rs
```diff
@@ -1753,6 +1753,29 @@ Quux.<CURSOR>
         ");
     }
 
+    #[test]
+    fn namedtuple_fallback_instance_methods() {
```
sharkdp: Can we please add tests for accessing methods (and classmethods) on `P` directly as well? And also add an example with a generic `NamedTuple`? It looks to me like those cases are not yet handled (see what we do for `TypedDict`).

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -95,7 +98,17 @@ impl<'db> AllMembers<'db> {
...
-                self.extend_with_instance_members(db, ty, instance.class_literal(db));
+                let class_lit = instance.class_literal(db);
+                self.extend_with_instance_members(db, ty, class_lit);
+
+                // If this is a NamedTuple instance, include members from NamedTupleFallback
+                if CodeGeneratorKind::NamedTuple.matches(db, class_lit) {
+                    if let Type::ClassLiteral(fallback) =
+                        KnownClass::NamedTupleFallback.to_class_literal(db)
+                    {
+                        self.extend_with_instance_members(db, ty, fallback);
```
sharkdp: Is `extend_with_instance_members` correct here, or should it be `self.extend_with_type(db, KnownClass::NamedTupleFallback.to_instance(db));`? For example, is the classmethod `_make` available on instances of `NamedTuple`s?

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -95,7 +98,17 @@ impl<'db> AllMembers<'db> {
             ),
 
             Type::NominalInstance(instance) => {
-                self.extend_with_instance_members(db, ty, instance.class_literal(db));
+                let class_lit = instance.class_literal(db);
```
sharkdp: Minor: maybe use `class_literal` instead of this abbreviation, like elsewhere in this method?

crates/ty_python_semantic/src/types/ide_support.rs
AlexWaygood: > is the classmethod `_make` available on instances of `NamedTuple`s?

Yes, it is. `NamedTuple` is pretty different from `TypedDict` in this regard — it's conceptually more similar to a dataclass. Inheriting from NamedTuple creates a tuple subclass that has a number of convenient methods monkey-patched onto it by the runtime, including `_make` etc. Instantiating that namedtuple class creates an instance of that namedtuple class just like instantiating any normal Python class would, so classmethods like `_make` are available on instances of namedtuple classes in the same way that classmethods are generally available on instances

sharkdp: > > is the classmethod `_make` available on instances of `NamedTuple`s?
> 
> Yes, it is.

Yeah, sorry. I meant to say: is `_make` available *as a completion* on instances of `NamedTuple`s (with the current implementation here; as it should be).

AlexWaygood: Ohhh sorry! I totally misinterpreted that as a question about what happens at runtime; my apologies 😄

crates/ty_ide/src/completion.rs
TaKO8Ki: Ok. Is it better to move these tests into the mdtest like `TypedDict` instead?

https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/ide_support/all_members.md

sharkdp: That might be a bit easier, and more consistent with what we already have, yes.

crates/ty_python_semantic/src/types/ide_support.rs
TaKO8Ki: I've changed it to class_literal.

crates/ty_ide/src/completion.rs
TaKO8Ki: I've added a mdtest for NamedTuple including generic NamedTuple example.

TaKO8Ki:
Is the CI error related to the changes? It seems to happen in a different branch. https://github.com/astral-sh/ruff/actions/runs/17681728186

AlexWaygood:
> Is the CI error related to the changes? It seems to happen in a different branch. [astral-sh/ruff/actions/runs/17681728186](https://github.com/astral-sh/ruff/actions/runs/17681728186)

no, it's not related to your change. We're not yet sure why that's started happening on all PRs, but you don't need to worry about it.

carljm:
The CI error is now fixed in main, so a rebase and force-push (or merge from main) will fix it on this PR.

PR #20357, opened 2025-09-11
Track t-strings and f-strings for token-based rules and suppression comments

### LINKED ISSUE #20310: ISC rules do not support t-strings
### Summary

[The ISC rules](https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc) don’t support t-strings. [Example](https://play.ruff.rs/ec0ae1e6-5942-4f87-9a1f-20a16f734fd4):
```console
$ cat >isc.py <<'# EOF'
# ISC001
t"The quick " t"brown fox."

# ISC002
t"The quick brown fox jumps over the lazy "\
t"dog."

# ISC003
(
    t"The quick brown fox jumps over the lazy "
    + t"dog"
)
# EOF

$ ruff --isolated check isc.py --select ISC --preview --target-version py314
All checks passed!

$ sed 's/t"/"/g' isc.py | ruff --isolated check - --select ISC --output-format concise -q
-:2:1: ISC001 [*] Implicitly concatenated string literals on one line
-:5:1: ISC002 Implicitly concatenated string literals over multiple lines
-:10:5: ISC003 [*] Explicitly concatenated string should be implicitly concatenated
```

### Version

dylwil3:
Our token-based rules and `noqa` extraction used an `Indexer` that kept track of f-string ranges but not t-strings. We've updated the `Indexer` and downstream uses thereof to handle both f-strings and t-strings.

Most of the diff is renaming and adding tests.

Note that much of the "new" logic gets to be naive because the lexer has already ensured that f and t-string "starts" are paired with their respective "ends", even amidst nesting and so on.

Finally: one could imagine wanting to know if a given interpolated string range corresponds to an f-string or a t-string, but I didn't find a place where we actually needed this.

Closes #20310

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_implicit_str_concat/snapshots/ruff_linter__rules__flake8_implicit_str_concat__tests__multiline_ISC001_ISC.py.snap
+++ b/crates/ruff_linter/src/rules/flake8_implicit_str_concat/snapshots/ruff_linter__rules__flake8_implicit_str_concat__tests__multiline_ISC001_ISC.py.snap
@@ -485,7 +485,7 @@ help: Combine string literals
 95 | 
 96 | # Mixed literal + non-literal scenarios
 
-ISC001 Implicitly concatenated string literals on one line
+ISC001 [*] Implicitly concatenated string literals on one line
    --> ISC.py:193:1
     |
 191 | # https://github.com/astral-sh/ruff/issues/20310
@@ -496,6 +496,14 @@ ISC001 Implicitly concatenated string literals on one line
 195 | # ISC002
     |
 help: Combine string literals
+190 | 
+191 | # https://github.com/astral-sh/ruff/issues/20310
+192 | # ISC001
+    - t"The quick " t"brown fox."
+193 + t"The quick brown fox."
+194 | 
+195 | # ISC002
+196 | t"The quick brown fox jumps over the lazy "\
 
 ISC001 Implicitly concatenated string literals on one line
    --> ISC.py:206:5
@@ -519,7 +527,7 @@ ISC001 Implicitly concatenated string literals on one line
     |
 help: Combine string literals
 
-ISC001 Implicitly concatenated string literals on one line
+ISC001 [*] Implicitly concatenated string literals on one line
    --> ISC.py:206:14
     |
 205 | # nested examples with both t and f-strings
@@ -529,6 +537,14 @@ ISC001 Implicitly concatenated string literals on one line
 208 | _ = f"b {t"abc" \
     |
 help: Combine string literals
+203 | )
+204 | 
+205 | # nested examples with both t and f-strings
+    - _ = "a" f"b {t"c" t"d"} e" "f"
+206 + _ = "a" f"b {t"cd"} e" "f"
+207 | _ = t"b {f"c" f"d {t"e" t"f"} g"} h"
+208 | _ = f"b {t"abc" \
+209 |     t"def"} g"
 
 ISC001 [*] Implicitly concatenated string literals on one line
    --> ISC.py:207:10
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs
```diff
@@ -123,21 +123,34 @@ pub(crate) fn implicit(
...
+                {
                     Some(a_range) => (a_range, b_token.range()),
                     None => continue,
                 }
             }
-            (TokenKind::FStringEnd, TokenKind::FStringStart) => {
+            (
+                TokenKind::FStringEnd | TokenKind::TStringEnd,
+                TokenKind::FStringStart | TokenKind::TStringStart,
+            ) => {
```
ntBre: I think you addressed this in the summary, and it sounds like it should be okay, but it would feel slightly more natural to me to write it as:


```suggestion
            (TokenKind::FStringEnd, TokenKind::FStringStart)
             | (TokenKind::TStringEnd, TokenKind::TStringStart) => {
```

or can they actually be mixed?

crates/ruff_linter/src/directives.rs
```diff
@@ -497,12 +498,29 @@ end'''
...
+            NoqaMapping::from_iter([TextRange::new(TextSize::from(6), TextSize::from(82))])
+        );
+
         let contents = "x = 1
 y = f'normal'
 z = f'another but {f'nested but {f'still single line'} nested'}'
 ";
         assert_eq!(noqa_mappings(contents), NoqaMapping::default());
 
+        let contents = "x = 1
```
ntBre, line 518: This looks like a test for an f-string in a t-string then a t-string in an f-string. Do we need one for a t-string in a t-string? These are probably the trickier cases, just an idea

crates/ruff_python_index/src/indexer.rs
```diff
@@ -9,15 +9,17 @@ use ruff_python_trivia::{
...
+};
 use crate::multiline_ranges::{MultilineRanges, MultilineRangesBuilder};
 
 pub struct Indexer {
     /// Stores the start offset of continuation lines.
     continuation_lines: Vec<TextSize>,
 
-    /// The range of all f-string in the source document.
-    fstring_ranges: FStringRanges,
+    /// The range of all interpolated string in the source document.
```
ntBre:
```suggestion
    /// The range of all interpolated strings in the source document.
```

I guess we might as well fix an existing typo while we're here

crates/ruff_linter/src/rules/flake8_implicit_str_concat/snapshots/ruff_linter__rules__flake8_implicit_str_concat__tests__multiline_ISC001_ISC.py.snap
```diff
@@ -484,3 +484,80 @@ help: Combine string literals
...
+    |          ^^^^^^^^^^^^^^^^^^^^^^^
+208 | _ = f"b {t"abc" \
+209 |     t"def"} g"
+    |
+help: Combine string literals
+204 | 
+205 | # nested examples with both t and f-strings
+206 | _ = "a" f"b {t"c" t"d"} e" "f"
+    - _ = t"b {f"c" f"d {t"e" t"f"} g"} h"
+207 + _ = t"b {f"cd {t"e" t"f"} g"} h"
```
ntBre, line 564: Just curious, why is this the only case with a fix? I'm assuming that's an existing feature of the rule, but I hadn't noticed it before.

dylwil3: Oh good catch! Fixed

crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs
dylwil3: Your way is better (but they are equivalent on valid code because of the explanation in the PR summary)

PR #20358, opened 2025-09-11
[ty] Add GitHub output format

ntBre:
## Summary

This PR wires up the GitHub output format moved to `ruff_db` in #20320 to the ty CLI.

It's a bit smaller than the GitLab version (#20155) because some of the helpers were already in place, but I did factor out a few `DisplayDiagnosticConfig` constructor calls in Ruff. I also exposed the `GithubRenderer` and a wrapper `DisplayGithubDiagnostics` type because we needed a way to configure the program name displayed in the GitHub diagnostics. This was previously hard-coded to `Ruff`:

image

Another option would be to drop the program name in the output format, but I think it can be helpful in workflows with multiple programs emitting annotations (such as Ruff and ty!)

## Test Plan

New CLI test, and a manual test with `--config 'terminal.output-format = "github"'`

### DIFF
--- a/crates/ruff_db/src/diagnostic/mod.rs
+++ b/crates/ruff_db/src/diagnostic/mod.rs
@@ -8,6 +8,7 @@ use ruff_text_size::{Ranged, TextRange, TextSize};
 
 pub use self::render::{
     DisplayDiagnostic, DisplayDiagnostics, FileResolver, Input, ceil_char_boundary,
+    github::{DisplayGithubDiagnostics, GithubRenderer},
 };
 use crate::{Db, files::File};
 
@@ -1312,10 +1313,6 @@ pub struct DisplayDiagnosticConfig {
     show_fix_diff: bool,
     /// The lowest applicability that should be shown when reporting diagnostics.
     fix_applicability: Applicability,
-    /// The program for which diagnostics are being rendered (Ruff or ty).
-    ///
-    /// This is currently only used by the GitHub output format and defaults to `Program::Ty`.
-    program: Program,
 }
 
 impl DisplayDiagnosticConfig {
@@ -1381,13 +1378,6 @@ impl DisplayDiagnosticConfig {
             ..self
         }
     }
-
-    /// Set the [`Program`] for which diagnostics are being displayed.
-    ///
-    /// This is currently only used by [`DiagnosticFormat::Github`].
-    pub fn program(self, program: Program) -> DisplayDiagnosticConfig {
-        DisplayDiagnosticConfig { program, ..self }
-    }
 }
 
 impl Default for DisplayDiagnosticConfig {
@@ -1401,7 +1391,6 @@ impl Default for DisplayDiagnosticConfig {
             show_fix_status: false,
             show_fix_diff: false,
             fix_applicability: Applicability::Safe,
-            program: Program::Ty,
         }
     }
 }
@@ -1464,21 +1453,6 @@ pub enum DiagnosticFormat {
     ///
     /// [GitHub Actions]: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
     Github,
-}
-
-#[derive(Clone, Debug)]
-pub enum Program {
-    Ruff,
-    Ty,
-}
-
-impl std::fmt::Display for Program {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Program::Ruff => f.write_str("Ruff"),
-            Program::Ty => f.write_str("ty"),
-        }
-    }
 }
 
 /// A representation of the kinds of messages inside a diagnostic.
--- a/crates/ruff_db/src/diagnostic/render/github.rs
+++ b/crates/ruff_db/src/diagnostic/render/github.rs
@@ -1,13 +1,13 @@
-use crate::diagnostic::{Diagnostic, DisplayDiagnosticConfig, FileResolver, Severity};
+use crate::diagnostic::{Diagnostic, FileResolver, Severity};
 
-pub(super) struct GithubRenderer<'a> {
+pub struct GithubRenderer<'a> {
     resolver: &'a dyn FileResolver,
-    config: &'a DisplayDiagnosticConfig,
+    program: &'a str,
 }
 
 impl<'a> GithubRenderer<'a> {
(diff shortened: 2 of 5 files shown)


crates/ruff_db/src/diagnostic/render/snapshots/ruff_db__diagnostic__render__github__tests__output.snap
```diff
@@ -2,6 +2,6 @@
 source: crates/ruff_db/src/diagnostic/render/github.rs
 expression: env.render_diagnostics(&diagnostics)
 ---
-::error title=Ruff (F401),file=fib.py,line=1,col=8,endLine=1,endColumn=10::fib.py:1:8: F401 `os` imported but unused
-::error title=Ruff (F841),file=fib.py,line=6,col=5,endLine=6,endColumn=6::fib.py:6:5: F841 Local variable `x` is assigned to but never used
-::error title=Ruff (F821),file=undef.py,line=1,col=4,endLine=1,endColumn=5::undef.py:1:4: F821 Undefined name `a`
+::error title=ty (F401),file=fib.py,line=1,col=8,endLine=1,endColumn=10::fib.py:1:8: F401 `os` imported but unused
+::error title=ty (F841),file=fib.py,line=6,col=5,endLine=6,endColumn=6::fib.py:6:5: F841 Local variable `x` is assigned to but never used
+::error title=ty (F821),file=undef.py,line=1,col=4,endLine=1,endColumn=5::undef.py:1:4: F821 Undefined name `a`
```
dhruvmanila, line 7: Just confirming that this change is basically just a side effect of having ty as the default for `Program` and is not because there's someplace where the program is not being set correctly?

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1390,6 +1401,7 @@ impl Default for DisplayDiagnosticConfig {
             show_fix_status: false,
             show_fix_diff: false,
             fix_applicability: Applicability::Safe,
+            program: Program::Ty,
```
dhruvmanila: I think at this point, it might be useful to refactor this into a `new` method that takes in a `program: Program` as the sole parameter because there's no valid default value for this field and the caller should always provide the program for which this rendering is to be done for. What do you think?

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1312,6 +1312,10 @@ pub struct DisplayDiagnosticConfig {
     show_fix_diff: bool,
     /// The lowest applicability that should be shown when reporting diagnostics.
     fix_applicability: Applicability,
+    /// The program for which diagnostics are being rendered (Ruff or ty).
+    ///
+    /// This is currently only used by the GitHub output format and defaults to `Program::Ty`.
```
dhruvmanila: Refer to my other comment, I'd probably remove the default value for this field.

dhruvmanila (approved):
This looks great!

I just have one minor concern, refer to inline comment.

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1454,6 +1466,21 @@ pub enum DiagnosticFormat {
     Github,
 }
 
+#[derive(Clone, Debug)]
+pub enum Program {
+    Ruff,
+    Ty,
+}
```
MichaReiser: Nit: It would probably be fine to just use a `String` here

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1312,6 +1312,10 @@ pub struct DisplayDiagnosticConfig {
     show_fix_diff: bool,
     /// The lowest applicability that should be shown when reporting diagnostics.
     fix_applicability: Applicability,
+    /// The program for which diagnostics are being rendered (Ruff or ty).
+    ///
+    /// This is currently only used by the GitHub output format and defaults to `Program::Ty`.
+    program: Program,
```
MichaReiser: I think I brought this up before and it isn't something that we need to change in this PR but I personally find it a bit difficult to reason about which settings are used for which output format (most settings are only used by one or very few formats). 

The main reason I see today why we need a single `Config` struct is because `DisplayDiagnostics` matches on the output format, which I'm not sure we should keep because we also match over the output format in ruff's rendering code (meaning, we first match to create the generic configuration only to then match on the format again to call into specific renderers). To me, it would feel simpler to just have a single match in ruff's printer that calls the corresponding renderer directly

crates/ruff_db/src/diagnostic/render/github.rs
```diff
@@ -17,7 +18,8 @@ impl<'a> GithubRenderer<'a> {
         for diagnostic in diagnostics {
             write!(
                 f,
-                "::error title=Ruff ({code})",
+                "::error title={program} ({code})",
```
MichaReiser: I think we want to use the correct severity for ty diagnostics and not always default to error

crates/ruff_db/src/diagnostic/mod.rs
ntBre: I thought eventually we would be able to hoist the `Config` out of the match in Ruff and have something like `DisplayDiagnosticsConfig::from(ruff_output_format)` and a single `DisplayDiagnostics` call, but I guess you're right. We may not want to port over all of the output formats, and some other logic is mixed in on the Ruff side, such as the `SHOW_FIX_SUMMARY` flag and writing of summary text. I can follow up on this.

crates/ruff_db/src/diagnostic/render/snapshots/ruff_db__diagnostic__render__github__tests__output.snap
ntBre: Yes, that's right!

ntBre:
Thanks for the reviews! I pushed one commit to use the real severity. I think Dhruv's suggestions about not having a default `Program` make a lot of sense, but it looks like the `Program` config field can go away entirely with Micha's restructuring suggestion (we can just pass a `&str` to `GithubRenderer::new` instead of a config). 

I briefly started that refactor on this branch, but I think it would help to finally make the `DisplayDiagnostics` rendering take a `std::io::Write` instead of a `std::fmt::Write`, so it seems best to address these all at once in an immediate follow-up PR.

PR #20359, opened 2025-09-12
[ty] no more diverging query cycles in type expressions

carljm:
## Summary

Use `Type::Divergent` to short-circuit diverging types in type expressions. This avoids panicking in a wide variety of cases of recursive type expressions.

Avoids many panics (but not yet all -- I'll be tracking down the rest) from https://github.com/astral-sh/ty/issues/256 by falling back to Divergent. For many of these recursive type aliases, we'd like to support them properly (i.e. really understand the recursive nature of the type, not just fall back to Divergent) but that will be future work.

This switches `Type::has_divergent_type` from using `any_over_type` to a custom set of visit methods, because `any_over_type` visits more than we need to visit, and exercises some lazy attributes of type, causing significantly more work. This change means this diff doesn't regress perf; it even reclaims some of the perf regression from https://github.com/astral-sh/ruff/pull/20333.

## Test Plan

Added mdtest for recursive type alias that panics on main.

Verified that we can now type-check `packaging` (and projects depending on it) without panic; this will allow moving a number of mypy-primer projects from `bad.txt` to `good.txt` in a subsequent PR.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -24,7 +24,10 @@ f(1)
 ```py
 from typing import Union
 
-T = list[Union["T", None]]
+Recursive = list[Union["Recursive", None]]
+
+def _(r: Recursive):
+    reveal_type(r)  # revealed: list[Divergent]
 ```
 
 ### New union syntax
@@ -35,5 +38,8 @@ python-version = "3.12"
 ```
 
 ```py
-T = list["T" | None]
+Recursive = list["Recursive" | None]
+
+def _(r: Recursive):
+    reveal_type(r)  # revealed: list[Divergent]
 ```
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -1,0 +1,39 @@
+# Implicit type aliases
+
+Implicit type aliases are the earliest form of type alias, introduced in PEP 484. They have no
+special marker, just an ordinary assignment statement.
+
+## Basic
+
+We support simple type aliases with no extra effort, when the "value type" of the RHS is still a
+valid type for use in a type expression:
+
+```py
+MyInt = int
+
+def f(x: MyInt):
+    reveal_type(x)  # revealed: int
+
+f(1)
+```
+
+## Recursive
+
+### Old union syntax
+
+```py
+from typing import Union
+
+T = list[Union["T", None]]
+```
+
+### New union syntax
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+T = list["T" | None]
+```
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6521,8 +6521,11 @@ impl<'db> Type<'db> {
         div: Type<'db>,
         visitor: &HasDivergentTypeVisitor<'db>,
     ) -> bool {
-        // We don't use `any_over_type` here because we don't need/want to descend into lazy parts
-        // of types (typevar bounds/constraints, type alias values, etc) here.
+        // We don't use `any_over_type` here because we don't need/want to descend into lazy
+        // parts of types (typevar bounds/constraints, type alias values, etc) here. If it's a lazy
+        // aspect of the type, it won't cause a query cycle in type inference, and we shouldn't
+        // trigger unnecessary inference of it. (TODO: bounds/constraints of legacy typevars aren't
+        // yet lazy, but they should be.)
         match self {
             Type::Dynamic(DynamicType::Divergent(_)) => self == div,
             Type::Union(union) => union.has_divergent_type_impl(db, div, visitor),
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -65,7 +65,6 @@ use crate::types::signatures::{Parameter, ParameterForm, Parameters, walk_signat
 use crate::types::tuple::TupleSpec;
 pub(crate) use crate::types::typed_dict::{TypedDictParams, TypedDictType, walk_typed_dict_type};
 use crate::types::variance::{TypeVarVariance, VarianceInferable};
-use crate::types::visitor::any_over_type;
 use crate::unpack::EvaluationMode;
 pub use crate::util::diagnostics::add_inferred_python_version_hint_to_diagnostic;
 use crate::{Db, FxOrderSet, Module, Program};
@@ -221,6 +220,10 @@ pub(crate) struct TryBool;
 /// A [`TypeTransformer`] that is used in `normalized` methods.
 pub(crate) type NormalizedVisitor<'db> = TypeTransformer<'db, Normalized>;
 pub(crate) struct Normalized;
+
+/// A [`CycleDetector`] that is used in `has_divergent_type` methods.
+pub(crate) type HasDivergentTypeVisitor<'db> = CycleDetector<HasDivergentType, Type<'db>, bool>;
+pub(crate) struct HasDivergentType;
 
 /// How a generic type has been specialized.
 ///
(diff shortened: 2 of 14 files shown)


carljm:
@mtshiba I can't request you as reviewer, but would be happy for you to take a look at this PR.

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -111,12 +111,18 @@ pub(crate) fn infer_definition_types<'db>(
...
-    _db: &'db dyn Db,
+    db: &'db dyn Db,
     _value: &DefinitionInference<'db>,
-    _count: u32,
-    _definition: Definition<'db>,
+    count: u32,
+    definition: Definition<'db>,
 ) -> salsa::CycleRecoveryAction<DefinitionInference<'db>> {
-    salsa::CycleRecoveryAction::Iterate
+    if count == ITERATIONS_BEFORE_FALLBACK {
```
mtshiba, line 122: This approach (sets an upper limit on cycles and falls back to inference to set the cycle initial value to `Divergent` instead of `Never` when it is detected that the cycle is "not likely to converge") seems to have already been adopted in #20333, but I think it isn't the best.
For example, suppose a divergent function has `n+m` branches for its return values, of which `m(>1)` are divergent. After the 10th cycle, the size of the return type will be `n(m^10 - 1)/(m-1)` (10th value of `size(n) = m size(n-1) + n; size(1) = n`).
Although cases where `m` is large are likely to be rare, if they do occur, it will become a bottleneck.

```
1st: T1 | T2 | ... | Tn (| list[Never] | tuple[Never])
2nd: T1 | T2 | ... | | Tn | list[T1 | T2 | ... | Tn] | tuple[T1 | T2 | ... | Tn]
3rd: T1 | T2 | ... | Tn | list[T1 | T2 | ... | Tn | list[T1 | T2 | ... | Tn] | tuple[T1 | T2 | ... | Tn]] | tuple[T1 | T2 | ... | Tn | list[T1 | T2 | ... | Tn] | tuple[T1 | T2 | ... | Tn]]
...
```

In fact, while working on #17371, I also felt it would be nice for salsa to do multi-cycle convergence checks even in fallbacks, but finally decided it wasn't necessary in this case.

I think that in this kind of inference, the initial cycle value should be set to the `Divergent` type from the beginning.
`Divergent` has the identity property with respect to union, like `Never`. So `Divergent` will not appear in the result type if the recursive inference is not actually divergent.

```
int | Divergent = int | (int | (int | ...)) = int
```

This means that we should add special rules for `Divergent` to the union type reduction rules. They have already been discussed [here](https://github.com/astral-sh/ruff/pull/17371#issuecomment-3279706332).
I plan to split #17371 again so we can use the rules here.

AlexWaygood:
> For many of these recursive type aliases, we'd like to support them properly (i.e. really understand the recursive nature of the type, not just fall back to Divergent) but that will be future work.

Will it still be easy to identify type aliases in user code that we don't properly support, if they no longer cause us to panic? The panics are obviously really bad UX, but the nice thing about them is it makes it very easy for us to spot where we have bugs

carljm:
> > For many of these recursive type aliases, we'd like to support them properly (i.e. really understand the recursive nature of the type, not just fall back to Divergent) but that will be future work.
> 
> Will it still be easy to identify type aliases in user code that we don't properly support, if they no longer cause us to panic? The panics are obviously really bad UX, but the nice thing about them is it makes it very easy for us to spot where we have bugs

I agree this is a potential downside, but a) I think it's outweighed by the value of eliminating the cycles ASAP, and b) I'm not that worried about it. We already know the core recursive-alias cases that we need to support, and if there are a few edge-case bugs hiding, if they matter to anyone we'll discover them.

It's also quite easy to temporarily remove the `Type::Divergent` fallback, if we want to do that and run it on ecosystem projects as a bug-finding tool.

AlexWaygood:
Sounds good!

crates/ty_python_semantic/src/types/infer.rs
carljm: I agree that the arbitrary choice of 10 cycles before falling back is not ideal. I think in practice it will be unlikely to be a problem, though.

I think the idea that we could make `Divergent` behave sufficiently similarly to `Never` that we could use it as the initial iteration type is interesting, but I'm not confident it will be able to handle all cases correctly.

For the moment I plan to continue pushing forward on addressing existing panics with this approach, but I'll be very interested to take a look if you propose a PR that can improve this.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Recursive
+
+### Old union syntax
+
+```py
+from typing import Union
+
+T = list[Union["T", None]]
```
sharkdp: Should we maybe add a reveal-type assertion here to "get notified" when this changes?
```suggestion
Recursive = list[Union["Recursive", None]]

def _(r: Recursive):
    reveal_type(r)  # revealed: list[Divergent]
```

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -0,0 +1,39 @@
...
+
+### New union syntax
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+T = list["T" | None]
```
sharkdp: Similar here?

```suggestion
Recursive = list["Recursive" | None]

def f(r: Recursive):
    reveal_type(r)  # revealed: list[Divergent]
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -6516,10 +6532,74 @@ impl<'db> Type<'db> {
...
+    }
+
+    pub(super) fn has_divergent_type_impl(
+        self,
+        db: &'db dyn Db,
+        div: Type<'db>,
+        visitor: &HasDivergentTypeVisitor<'db>,
+    ) -> bool {
+        // We don't use `any_over_type` here because we don't need/want to descend into lazy parts
+        // of types (typevar bounds/constraints, type alias values, etc) here.
```
sharkdp: It's not immediately clear to me why descending into e.g. typevar bounds/constraints is not necessary. Could the reason be explained in this comment?

crates/ty_python_semantic/src/types.rs
```diff
@@ -6516,10 +6532,74 @@ impl<'db> Type<'db> {
     }
 
     pub(super) fn has_divergent_type(self, db: &'db dyn Db, div: Type<'db>) -> bool {
-        any_over_type(db, self, &|ty| match ty {
-            Type::Dynamic(DynamicType::Divergent(_)) => ty == div,
-            _ => false,
-        })
+        self.has_divergent_type_impl(db, div, &HasDivergentTypeVisitor::new(false))
```
sharkdp: Have you thought about implementing `has_divergent_type` using a custom `TypeVisitor` (similar to what `any_over_type` does), that would overwrite selected methods like `visit_bound_type_var_type` in order to skip visiting the parts that this needs/wants to skip?

marcelroed:
Here's an unhandled cycle using the latest commit:

```python
from typing import Generic, TypeVar

AT = TypeVar("AT", bound="A")
BT = TypeVar("BT", bound="B")

class A(Generic[BT]):
    ...

class B(Generic[AT]):
    ...
```

This is from an ML modeling library where type A (model configs) need to be able to construct type B (model instances) and type B (model instances) need to be able to return type A (model configs). Feel free to delete/hide this comment if it should be a part of a separate issue instead.

carljm:
@marcelroed Yes, thanks -- I'm aware this PR doesn't fix all the cases in https://github.com/astral-sh/ty/issues/256. Your case is already mentioned in https://github.com/astral-sh/ty/issues/256#issuecomment-3140776680, so it will be addressed before we close that issue.

crates/ty_python_semantic/src/types.rs
carljm: It's because they are lazy/deferred (not part of the core identity of the type), which means they won't cause a query cycle in type inference, and we don't want to trigger their inference unnecessarily.

Currently this is actually only true for PEP 695 type vars, but it should become true for all typevars; that's on my roadmap.

Added a comment about this.

crates/ty_python_semantic/src/types.rs
carljm: Thanks for the prompt. I tried this but was slightly unhappy with how many types the visitor needed to acquire detailed knowledge of the internals of in order to correctly implement visiting types without visiting lazily-inferred type attributes. So I ended up creating a general visitor flag for this, so that each type's `walk_` function can implement this distinction internally.

PR #20360, opened 2025-09-12
[ty] Infer more precise types for collection literals 

ibraheemdev:
## Summary

Part of https://github.com/astral-sh/ty/issues/168. Infer more precise types for collection literals (currently, only `list` and `set`). For example,

```py
x = [1, 2, 3] # revealed: list[Unknown | int]
y: list[int] = [1, 2, 3] # revealed: list[int]
```

This could easily be extended to `dict` literals, but I am intentionally limiting scope for now.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -79,6 +79,49 @@ b: tuple[int] = ("foo",)
 c: tuple[str | int, str] = ([], "foo")
 ```
 
+## Collection literal annotations are understood
+
+`module.py`:
+
+```py
+import typing
+
+a: list[int] = [1, 2, 3]
+b: list[int | str] = [1, 2, 3]
+c: typing.List[int] = [1, 2, 3]
+d: list[typing.Any] = []
+
+e: set[int] = {1, 2, 3}
+f: set[int | str] = {1, 2, 3}
+g: typing.Set[int] = {1, 2, 3}
+```
+
+`script.py`:
+
+```py
+import typing
+from module import a, b, c, d, e, f, g
+
+reveal_type(a)  # revealed: list[int]
+reveal_type(b)  # revealed: list[int | str]
+reveal_type(c)  # revealed: list[int]
+reveal_type(d)  # revealed: list[Any]
+
+reveal_type(e)  # revealed: set[int]
+reveal_type(f)  # revealed: set[int | str]
+reveal_type(g)  # revealed: set[int]
+```
+
+## Incorrect collection literal assignments are complained aobut
+
+```py
+# error: [invalid-assignment] "Object of type `list[str | Literal[1, 2, 3]]` is not assignable to `list[str]`"
+a: list[str] = [1, 2, 3]
+
+# error: [invalid-assignment] "Object of type `set[int | Literal["3"]]` is not assignable to `set[int]`"
+b: set[int] = {1, 2, "3"}
+```
+
 ## PEP-604 annotations are supported
 
 ```py
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -129,6 +129,16 @@ type IntList = list[int]
 
 m: IntList = [1, 2, 3]
 reveal_type(m)  # revealed: list[int]
+
+# TODO: this should type-check and avoid literal promotion
+# error: [invalid-assignment] "Object of type `list[int]` is not assignable to `list[Literal[1, 2, 3]]`"
+n: list[typing.Literal[1, 2, 3]] = [1, 2, 3]
+reveal_type(n)  # revealed: list[Literal[1, 2, 3]]
+
+# TODO: this should type-check and avoid literal promotion
+# error: [invalid-assignment] "Object of type `list[str]` is not assignable to `list[LiteralString]`"
+o: list[typing.LiteralString] = ["a", "b", "c"]
+reveal_type(o)  # revealed: list[LiteralString]
 ```
 
 ## Incorrect collection literal assignments are complained aobut
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -81,35 +81,41 @@ c: tuple[str | int, str] = ([], "foo")
 
 ## Collection literal annotations are understood
 
-`module.py`:
-
 ```py
 import typing
 
 a: list[int] = [1, 2, 3]
-b: list[int | str] = [1, 2, 3]
-c: typing.List[int] = [1, 2, 3]
-d: list[typing.Any] = []
-
-e: set[int] = {1, 2, 3}
-f: set[int | str] = {1, 2, 3}
-g: typing.Set[int] = {1, 2, 3}
-```
-
-`script.py`:
-
-```py
-import typing
-from module import a, b, c, d, e, f, g
-
 reveal_type(a)  # revealed: list[int]
+
+b: list[int | str] = [1, 2, 3]
 reveal_type(b)  # revealed: list[int | str]
+
+c: typing.List[int] = [1, 2, 3]
 reveal_type(c)  # revealed: list[int]
+
+d: list[typing.Any] = []
 reveal_type(d)  # revealed: list[Any]
 
+e: set[int] = {1, 2, 3}
 reveal_type(e)  # revealed: set[int]
+
+f: set[int | str] = {1, 2, 3}
 reveal_type(f)  # revealed: set[int | str]
+
+g: typing.Set[int] = {1, 2, 3}
 reveal_type(g)  # revealed: set[int]
+
+h: list[list[int]] = [[], [42]]
+# TODO: revealed: list[list[int]]
+reveal_type(h)  # revealed: list[list[int] | list[@Todo(list literal element type)]]
+
+i: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
+reveal_type(i)  # revealed: list[tuple[str | int, ...]]
+
+j: list[tuple[list[typing.Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["bar"])]
+# TODO: revealed: list[tuple[list[typing.Any], ...]]
+# revealed: list[tuple[list[Any], ...] | tuple[list[@Todo(list literal element type)]] | tuple[list[@Todo(list literal element type)], list[@Todo(list literal element type)]]]
+reveal_type(j)
 ```
 
 ## Incorrect collection literal assignments are complained aobut
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -111,21 +111,24 @@ g: typing.Set[int] = {1, 2, 3}
 reveal_type(g)  # revealed: set[int]
 
 h: list[list[int]] = [[], [42]]
-# TODO: revealed: list[list[int]]
-reveal_type(h)  # revealed: list[list[int] | list[Unknown] | list[Unknown | int]]
+reveal_type(h)  # revealed: list[list[int]]
 
-i: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
-reveal_type(i)  # revealed: list[tuple[str | int, ...]]
+i: list[typing.Any] = [1, 2, "3", ([4],)]
+reveal_type(i)  # revealed: list[Any | int | str | tuple[list[Unknown | int]]]
 
-j: list[tuple[list[typing.Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["bar"])]
-# TODO: revealed: list[tuple[list[typing.Any], ...]]
-# revealed: list[tuple[list[Any], ...] | tuple[list[Unknown]] | tuple[list[Unknown | int], list[Unknown | int]] | tuple[list[Unknown | str], list[Unknown | str]]]
-reveal_type(j)
+j: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
+reveal_type(j)  # revealed: list[tuple[str | int, ...]]
+
+k: list[tuple[list[int], ...]] = [([],), ([1, 2], [3, 4]), ([5], [6], [7])]
+reveal_type(k)  # revealed: list[tuple[list[int], ...]]
+
+l: tuple[list[int], *tuple[list[typing.Any], ...], list[str]] = ([1, 2, 3], [4, 5, 6], [7, 8, 9], ["10", "11", "12"])
+reveal_type(l)  # revealed: tuple[list[int], list[Any | int], list[Any | int], list[str]]
 
 type IntList = list[int]
 
-k: IntList = [1, 2, 3]
-reveal_type(a)  # revealed: list[int]
+m: IntList = [1, 2, 3]
+reveal_type(m)  # revealed: list[int]
 ```
 
 ## Incorrect collection literal assignments are complained aobut
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -81,6 +81,11 @@ c: tuple[str | int, str] = ([], "foo")
 
 ## Collection literal annotations are understood
 
+```toml
+[environment]
+python-version = "3.12"
+```
+
 ```py
 import typing
 
@@ -116,6 +121,11 @@ j: list[tuple[list[typing.Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["ba
 # TODO: revealed: list[tuple[list[typing.Any], ...]]
 # revealed: list[tuple[list[Any], ...] | tuple[list[Unknown]] | tuple[list[Unknown | Literal[1, 2]], list[Unknown | Literal[3, 4]]] | tuple[list[Unknown | Literal["foo"]], list[Unknown | Literal["bar"]]]]
 reveal_type(j)
+
+type IntList = list[int]
+
+k: IntList = [1, 2, 3]
+reveal_type(a)  # revealed: list[int]
 ```
 
 ## Incorrect collection literal assignments are complained aobut
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -112,14 +112,14 @@ reveal_type(g)  # revealed: set[int]
 
 h: list[list[int]] = [[], [42]]
 # TODO: revealed: list[list[int]]
-reveal_type(h)  # revealed: list[list[int] | list[Unknown] | list[Unknown | Literal[42]]]
+reveal_type(h)  # revealed: list[list[int] | list[Unknown] | list[Unknown | int]]
 
 i: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
 reveal_type(i)  # revealed: list[tuple[str | int, ...]]
 
 j: list[tuple[list[typing.Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["bar"])]
 # TODO: revealed: list[tuple[list[typing.Any], ...]]
-# revealed: list[tuple[list[Any], ...] | tuple[list[Unknown]] | tuple[list[Unknown | Literal[1, 2]], list[Unknown | Literal[3, 4]]] | tuple[list[Unknown | Literal["foo"]], list[Unknown | Literal["bar"]]]]
+# revealed: list[tuple[list[Any], ...] | tuple[list[Unknown]] | tuple[list[Unknown | int], list[Unknown | int]] | tuple[list[Unknown | str], list[Unknown | str]]]
 reveal_type(j)
 
 type IntList = list[int]
@@ -131,10 +131,10 @@ reveal_type(a)  # revealed: list[int]
 ## Incorrect collection literal assignments are complained aobut
 
 ```py
-# error: [invalid-assignment] "Object of type `list[Literal[1, 2, 3]]` is not assignable to `list[str]`"
+# error: [invalid-assignment] "Object of type `list[int]` is not assignable to `list[str]`"
 a: list[str] = [1, 2, 3]
 
-# error: [invalid-assignment] "Object of type `set[Literal[1, 2, "3"]]` is not assignable to `set[int]`"
+# error: [invalid-assignment] "Object of type `set[int | str]` is not assignable to `set[int]`"
 b: set[int] = {1, 2, "3"}
 ```
 
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -107,24 +107,24 @@ reveal_type(g)  # revealed: set[int]
 
 h: list[list[int]] = [[], [42]]
 # TODO: revealed: list[list[int]]
-reveal_type(h)  # revealed: list[list[int] | list[@Todo(list literal element type)]]
+reveal_type(h)  # revealed: list[list[int] | list[Unknown] | list[Unknown | Literal[42]]]
 
 i: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
 reveal_type(i)  # revealed: list[tuple[str | int, ...]]
 
 j: list[tuple[list[typing.Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["bar"])]
 # TODO: revealed: list[tuple[list[typing.Any], ...]]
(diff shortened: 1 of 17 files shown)


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
AlexWaygood, line 1: I'd love to see some tests for nested collections here -- e.g.

```py
x: list[list[int]] = [[], [42]]
y: list[tuple[str | int, ...]] = [(1, 2), ("foo", "bar"), ()]
z: list[tuple[list[Any], ...]] = [([],), ([1, 2], [3, 4]), (["foo"], ["bar"])]
```

It's obviously okay if they don't all pass right now -- we can add TODOs next to them

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -79,6 +79,49 @@ b: tuple[int] = ("foo",)
...
+from module import a, b, c, d, e, f, g
+
+reveal_type(a)  # revealed: list[int]
+reveal_type(b)  # revealed: list[int | str]
+reveal_type(c)  # revealed: list[int]
+reveal_type(d)  # revealed: list[Any]
+
+reveal_type(e)  # revealed: set[int]
+reveal_type(f)  # revealed: set[int | str]
+reveal_type(g)  # revealed: set[int]
```
AlexWaygood: these all pass on `main` (https://play.ty.dev/9844d26d-f2b1-45ca-a6d9-4bd317376ffd) because we will always prefer the declared type over the inferred type for a symbol's "public type" as observed by an external scope. What doesn't pass on `main` is an mdtest like this, where all the `reveal_type` calls are in the same scope that the collection literals are created in:

```py
import typing

a: list[int] = [1, 2, 3]
reveal_type(a)  # revealed: list[int]

b: list[int | str] = [1, 2, 3]
reveal_type(b)  # revealed: list[int | str]

c: typing.List[int] = [1, 2, 3]
reveal_type(c)  # revealed: list[int]

d: list[typing.Any] = []
reveal_type(d)  # revealed: list[Any]

e: set[int] = {1, 2, 3}
reveal_type(e)  # revealed: set[int]

f: set[int | str] = {1, 2, 3}
reveal_type(f)  # revealed: set[int | str]

g: typing.Set[int] = {1, 2, 3}
reveal_type(g)  # revealed: set[int]
```

https://play.ty.dev/eadf4952-5909-41ee-b836-09f1bef7f6ac

That's because the thing that this PR is changing is that we now use the declared type as context to inform the inferred type that we produce for local-scope accesses of a variable

AlexWaygood:
Would you be able to look through the mypy_primer diff and check that it's mostly as expected? If you use the ecosystem analyzer rich diff [here](https://ibraheem-bidirectional-infer.ecosystem-663.pages.dev/diff), it links directly to the line each new diagnostic is being emitted on

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -79,6 +79,49 @@ b: tuple[int] = ("foo",)
...
+```
+
+## Incorrect collection literal assignments are complained aobut
+
+```py
+# error: [invalid-assignment] "Object of type `list[str | Literal[1, 2, 3]]` is not assignable to `list[str]`"
+a: list[str] = [1, 2, 3]
```
ibraheemdev, line 148: The error message here is slightly odd, I would expect it to be "Object of type `list[Literal[1, 2, 3]]` is not...". Should we infer the RHS without a type context if it's not assignable to the declared type?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5270,21 +5274,66 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             .to_specialized_instance(self.db(), [todo_type!("list literal element type")])
```
carljm: Even in the short term (e.g. for the beta) we'll want to improve this to `list[Unknown | Literal[1, 2, 3]]` instead of the todo type. That obviously doesn't necessarily have to be done in the same PR, I only mention it because it suggests that we might want to organize the code in such a way that "inference with a context" and "inference without a context" are less separated into different paths, because we'll need the inferred types of the elements either way. (But this can obviously also be changed in a future PR.)

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
carljm: Yes, probably if any of the inferred literal types are not subtypes of the context element type, we should infer without regard for the context, in order to have a more intuitive error message.

carljm:
Oh, I do think some analysis of the ecosystem report here would be good, if there are any common causes of false positives we should prioritize addressing next. I did check the conformance suite report -- the changes are both good.

ibraheemdev:
I ended up rewriting most of the code to support better inference for unannotated collection literals as well. We now infer the type of a collection literal based on the inferred type of each element, as well as the type annotation if provided, or `Unknown` if not. For example,

```py
x = [1, 2, 3] # revealed: list[Unknown | Literal[1, 2, 3]]
x: list[int] = [1, 2, 3] # revealed: list[int]
```

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -79,6 +79,55 @@ b: tuple[int] = ("foo",)
...
+
+f: set[int | str] = {1, 2, 3}
+reveal_type(f)  # revealed: set[int | str]
+
+g: typing.Set[int] = {1, 2, 3}
+reveal_type(g)  # revealed: set[int]
+
+h: list[list[int]] = [[], [42]]
+# TODO: revealed: list[list[int]]
+reveal_type(h)  # revealed: list[list[int] | list[Unknown] | list[Unknown | Literal[42]]]
```
ibraheemdev: @dcreager is this something that you expect the new constraint solver to be able to simplify?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5253,36 +5253,132 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
         }
 
-        KnownClass::Set.to_specialized_instance(self.db(), [todo_type!("set literal element type")])
+        // The inferred type of each element acts as an additional constraint on `T`.
+        for inferred_elt_ty in inferred_elt_tys {
+            // Use the fallback instance type for element literals to avoid excessively large unions
+            // for large nested list literals, which the constraint solver struggles with.
+            let inferred_elt_ty = inferred_elt_ty
+                .literal_fallback_instance(self.db())
+                .unwrap_or(inferred_elt_ty);
```
ibraheemdev: Without this, ty (seemingly) hangs when checking [files that contain large 2D list literals](https://github.com/colour-science/colour/blob/develop/colour/models/datasets/pointer_gamut.py#L52), even in release mode. This means `[1, 2, 3]` is inferred as `list[Unknown | int]` instead of `list[Unknown | Literal[1, 2, 3]]`.

We could potentially use a heuristic before falling back?

ibraheemdev: This does mean we now fail to typecheck:
```py
x: list[LiteralString] = ["a", "b"]
```
Though we may be able to special case this.

ibraheemdev: Hmm, this seems to still be problematic for large lists of tuples. The runtime of `ty check` on `rich` has increased from 0.6s to 4s, largely due to [this file](https://github.com/Textualize/rich/blob/master/rich/_cell_widths.py). Checking that file takes >3s in release mode, and >80s in debug mode, which is why CI is timing out.

Note that the initial commit, which was restricted to annotated collection literals, does not seem to introduce any regressions (as annotated collection literals are typically quite simple).

carljm: > This does mean we now fail to typecheck:
> 
> ```python
> x: list[LiteralString] = ["a", "b"]
> ```

Seems like we should be able to use `LiteralString` instead of `str` as the fallback type for string literals?

> this seems to still be problematic for large lists of tuples

We may need to disallow inferring precise literal types for tuple members when nested? It seems pyright does something like this: `(0, 0, 0)` is inferred as `tuple[Literal[0], Literal[0], Literal[0]]`, but as soon as it is even `[(0, 0, 0)]` it becomes `list[tuple[int, int, int]]`.

carljm: I could also see potentially doing it at the union level, so that we simplify two tuples of literals to a single tuple of the common fallback type at each element position? This is related to https://github.com/astral-sh/ty/issues/493, but also seems like it might get complex more quickly.

It would also be OK to pare this PR back down to just the annotated case, if you want to get that landed before looking more deeply into this performance issue.

ibraheemdev: The tuple literal promotion seems to fix most of the hangs, but scipy seems to still be hanging in mypy-primer and the ecosystem-analyzer. Weirdly, mypy-primer passes for me locally, without a large regression compared to main, if I run the exact same command as CI, so I'm not exactly sure why it's hanging.

ibraheemdev: I'm able to reproduce it now, I suspect that the new types introduced by this PR are exposing a combinatorial explosion bug somewhere else when checking scipy. From the backtrace it seems related to `Type::normalized_impl`.

carljm: It's a surprise to get infinite hang from `Type::normalized_impl` without getting a stack overflow. But it may be possible. Or it may just be taking a very long time to normalize some deeply-nested type, but not actually be in an infinite loop?

Typically infinite loops in `Type::normalized_impl` should be caught by the `NormalizedVisitor`.

carljm: Let me know (and post the repro instructions) if you'd like a second pair of eyes on this.

ibraheemdev: It looks like the culprit is [this call to `enumerate`](https://github.com/scipy/scipy/blob/c5c862df0fbe70f1c0dbeb07c85c5d98ec348e96/scipy/optimize/_tstutils.py#L968). The inferred type of the argument after this PR is:
```py
list[
  Unknown
    | list[Unknown | (def fun1(x) -> Unknown) | list[Unknown | int] | float | int]
    | list[Unknown | (def fun1(x) -> Unknown) | list[Unknown | int | float] | float | int]
    | list[Unknown | (def fun2(x) -> Unknown) | list[Unknown | float] | int]
    | list[Unknown | (def fun3(x) -> Unknown) | list[Unknown | int] | int]
    | list[Unknown | (def fun3(x) -> Unknown) | list[Unknown | int | float] | int]
    | list[Unknown | (def fun4(x) -> Unknown) | list[Unknown | int] | int]
    | list[Unknown | (def fun4(x) -> Unknown) | list[Unknown | int | float] | int]
    | list[Unknown | (def fun5(x) -> Unknown) | list[Unknown | int] | int]
    | list[Unknown | (def fun6(x) -> Unknown) | list[Unknown | int | float] | int]
    | list[Unknown | (def fun7(x) -> Unknown) | list[Unknown | int] | int]
    | list[Unknown | (def fun8(x) -> Unknown) | list[Unknown | float | int] | float | int]
    | list[Unknown | (def fun9(x) -> Unknown) | list[Unknown | float | int] | float | int]
]
```

And the backtrace shows deeply nested calls to `Type::normalized_impl` inside `infer_call_expression`. Running `ty check scipy/optimize/_tstutils.py` from this PR takes 40s in release mode, and much longer in debug mode.

ibraheemdev: I suppose promoting the function types to their `Callable` signature would also fix this?

carljm: Not sure; it depends on where the runtime explosion in `normalized` is coming from. That type doesn't look unreasonably large. And abstracting the function types would shrink it some, but wouldn't reduce its nesting level at all.

ibraheemdev: That seems to have fixed it.

crates/ty_python_semantic/resources/mdtest/narrow/conditionals/nested.md
```diff
@@ -311,15 +311,15 @@ def f(l: list[str | None]):
     if l[0] is not None:
         def _():
             # TODO: should be `str | None`
-            reveal_type(l[0])  # revealed: str | None | @Todo(list literal element type)
+            reveal_type(l[0])  # revealed: str | None | Unknown
         # TODO: should be of type `list[None]`
```
carljm: I would remove these TODOs; they now refer just to removing the `Unknown |` from un-annotated generic collection literals, and there are a lot more tests than just these that will change if/when we do that.

crates/ty_python_semantic/resources/mdtest/narrow/conditionals/nested.md
```diff
@@ -311,15 +311,15 @@ def f(l: list[str | None]):
...
         l = [None]
 
 def f(l: list[str | None]):
     l[0] = "a"
     def _():
         # TODO: should be `str | None`
-        reveal_type(l[0])  # revealed: str | None | @Todo(list literal element type)
+        reveal_type(l[0])  # revealed: str | None | Unknown
     # TODO: should be of type `list[None]`
```
carljm: Also these TODOs

crates/ty_python_semantic/src/types.rs
```diff
@@ -1133,6 +1133,24 @@ impl<'db> Type<'db> {
...
     }
 
+    /// If this type is a literal, returns a type that would be appropriate as a type annotation for
+    /// an instance of the literal.
+    ///
+    /// Notably, this converts `def _() -> int` to `Callable[[], int]`, which `literal_fallback_instance`
+    /// does not.
+    pub(crate) fn literal_annotation_type(self, db: &'db dyn Db) -> Option<Type<'db>> {
+        match self {
+            Type::StringLiteral(_) | Type::LiteralString => Some(KnownClass::Str.to_instance(db)),
```
carljm, line 1144: Does this mean we still error on `x: list[LiteralString] = ["foo"]`? Can we add a test for that (with TODO if we still error on it)?

crates/ty_python_semantic/src/types.rs
```diff
@@ -1133,6 +1133,24 @@ impl<'db> Type<'db> {
         }
     }
 
+    /// If this type is a literal, returns a type that would be appropriate as a type annotation for
+    /// an instance of the literal.
+    ///
+    /// Notably, this converts `def _() -> int` to `Callable[[], int]`, which `literal_fallback_instance`
```
carljm: Given the comment on `literal_fallback_instance` above, I think we could probably just add the `FunctionLiteral` fallback there and keep it to just the one method? Or did you try that and observe undesirable effects?

AFAIK `literal_fallback_instance` is just used for inference of generic classes from constructor calls, which is closely related to what we use `literal_annotation_type` for. So I don't see much semantic justification for separating them.

crates/ty_python_semantic/src/types.rs
```diff
@@ -5984,8 +6002,10 @@ impl<'db> Type<'db> {
                         self
                     }
                 }
-                TypeMapping::PromoteLiterals | TypeMapping::BindLegacyTypevars(_) |
-                TypeMapping::MarkTypeVarsInferable(_) => self,
+                TypeMapping::LiteralToInstance
+                    | TypeMapping::LiteralToAnnotation
```
carljm: Similar to above, maybe just keep it to one `PromoteLiterals` type mapping, unless there are strong reasons we need both?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5253,36 +5254,130 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                Type::TypeAlias(alias) => alias.value_type(db).into_nominal_instance()?,
+                _ => return None,
+            }
+            .class(db);
+
+            if !class_type.is_known(db, collection_class) {
+                return None;
+            }
+            let specialization = class_type.into_generic_alias()?.specialization(db);
+            let [annotated_elts_ty] = specialization.types(db) else {
```
carljm: This will need some generalization (in a subsequent PR) to handle dict literals (with two type parameters).

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5253,36 +5254,130 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            panic!("Typeshed should always have a `{name}` class in `builtins.pyi` with a single type variable")
+        });
+
+        let mut elements_are_assignable = true;
+        let mut inferred_elt_tys = Vec::with_capacity(elts.len());
+
+        // Infer the type of each element in the collection literal.
         for elt in elts {
-            self.infer_expression(elt, TypeContext::default());
+            let inferred_elt_ty = self.infer_expression(elt, TypeContext::default());
```
carljm: Should we have a TODO here for pushing down the nested context? I saw some related TODOs in tests.

carljm (approved):
Looks great, thank you! Just a few nits inline, but I think it's ready to land.

crates/ty_python_semantic/src/types.rs
ibraheemdev: The reason I separated them is that the expected `literal_fallback_instance` for a `FunctionLiteral` would be `FunctionType`. Returning `Callable` is slightly different, and I thought that might be confusing, e.g. [see this branch](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types.rs#L1713) that I would expect to be able to group with the `literal_fallback_instance` branch above it, if `literal_fallback_instance` had an implementation for `Type::FunctionLiteral`.

I also wasn't sure if we want to change the semantics of generic class constructors, which is the one place we currently use `TypeMapping::PromoteLiteral`. If you think it's still fine to change the semantics of `TypeMapping::PromoteLiteral` instead of add a separate operation, I'm happy to do that.

carljm: I see, makes sense.

I do think we only need one `PromoteLiterals` mapping. The way that we promote literals should be consistent between generic class instantiations and generic collection inference (the latter is really just kind of a special case of the former), and I think, for that case, promoting functions to callable types is better than promoting them to `FunctionType`. (There could be an argument for "callable-type & types.FunctionType" as the best promotion and fallback type for functions, but that would require that we special-case around the overly-forgiving definition of `types.FunctionType.__call__` in typeshed; definitely something for another time.)

All that said, I see your point that `FunctionType` would also be a reasonable "instance fallback", and that we have at least one case where the fallback to `FunctionType` is what we'd want for functions. So I could see keeping two methods here, in which case a) I'd name them `literal_fallback_instance` and `literal_promotion_type` (I think the latter more clearly describes how it is used than "literal_annotation_type"), and b) I'd probably go ahead and add `FunctionType` fallback to the former as well, and clean up the spot you linked (though that's maybe scope creep for this PR and could be done separately.)

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: I hadn't considered that supporting nested context was as simple as just passing the inner type down here! I implemented that and removed the TODOs.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5238,11 +5241,39 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            // Extend variadic tuple annotations to match the length of the RHS.
+                            let variable_elements = elts
+                                .len()
+                                .saturating_sub(tuple.prefix.len())
+                                .saturating_sub(tuple.suffix.len());
+
+                            let elements = tuple
+                                .prefix_elements()
+                                .chain(std::iter::repeat_n(&tuple.variable, variable_elements))
+                                .chain(tuple.suffix_elements());
```
ibraheemdev: I'm a little surprised we don't follow this pattern anywhere else for type-checking a tuple, is there a better way of doing this?

crates/ty_python_semantic/src/types.rs
ibraheemdev: Yeah, I'm a little hesitant about `LiteralString` being the default promotion type, because I think this is actually a bigger issue with unconditionally promoting the types of the RHS. For example, `x: list[Literal["a"]] = ["a"]` should also type-check (and mypy and pyright both handle this correctly). I think to make this work properly we would need to thread the type-context down to the literal promotions visitor, which seems non-trivial. I left a test with a TODO for now.

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: I don't think we currently have any cases where we need to check a tuple spec against a known fixed set of elements; we are generally comparing a full tuple spec to another full tuple spec (e.g. because the RHS is another tuple type). So in a sense this scenario is kind of the `Tuple::Fixed` branch of `VariableLengthTuple::has_relation_to_impl` (or maybe it's the `Tuple::Variable` branch of `FixedLengthTuple::has_relation_to_impl`?) -- but even in those cases we handle it a bit differently, we don't ever create a single iterator where we repeat variadic elements to stretch to a certain length.

I think the way you are doing this here makes sense. I do think there'd be an argument for encapsulating it as a new method `Tuple::elements_as_length` (maybe there's a better name?), which would make this method shorter and maybe clearer. But we could also wait to do that until/unless we have a second need for it.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -79,6 +79,78 @@ b: tuple[int] = ("foo",)
...
+# TODO: this should type-check and avoid literal promotion
+# error: [invalid-assignment] "Object of type `list[int]` is not assignable to `list[Literal[1, 2, 3]]`"
+n: list[typing.Literal[1, 2, 3]] = [1, 2, 3]
+reveal_type(n)  # revealed: list[Literal[1, 2, 3]]
+
+# TODO: this should type-check and avoid literal promotion
+# error: [invalid-assignment] "Object of type `list[str]` is not assignable to `list[LiteralString]`"
+o: list[typing.LiteralString] = ["a", "b", "c"]
+reveal_type(o)  # revealed: list[LiteralString]
+```
```
carljm, line 142: Have you done a review of the ecosystem analyzer results? It's good to do in general just to catch bugs, but specifically it would be good to know if this is actually coming up in the ecosystem, to get a sense of how much we should prioritize fixing it.

crates/ty_python_semantic/src/types.rs
carljm: Makes sense.

> I think to make this work properly we would need to thread the type-context down to the literal promotions visitor

I guess this is because we might want to promote some literals but not others, e.g. if we have `x: list[int | Literal["a"]] = ["a", <massive list of integer literals>]`?

It seems like the simpler cases could be handled via a heuristic of "if annotated element type is a literal, don't promote literals on the RHS", but that wouldn't handle the mixed union case well.

carljm (approved):

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: This looks to me very much like the code we have in `VariableLengthTuple::resize()`. You might be able to avoid `match`ing on the inner tuple altogether by just using `Tuple::resize()`.

crates/ty_python_semantic/src/types.rs
ibraheemdev: > I guess this is because we might want to promote some literals but not others, e.g. if we have `x: list[int | Literal["a"]] = ["a", <massive list of integer literals>]`?

I think there are more straightforwards examples than the union case, e.g. `x: list[tuple[Literal[1]]] = [(1,)]`.

One of the issues is that we promote the literals of the RHS _recursively_ when inferring the elements of a list/set, because we currently infer `(1,)` to be `tuple[Literal[1]]`, but `[(1,)]` to be `list[tuple[int]]`. That means the logic for any heuristic would need to go in the type-mapping visitor, which would then need access to the type-context.

This is actually already an issue in the other place we use `PromoteLiterals`:
```py
class X[T]:
    def __init__(self, x: T):
        self.x = x

# ty: Object of type `tuple[X[int]]` is not assignable to `tuple[X[Literal[1]]]`
x: tuple[X[Literal[1]]] = (X(1), )
```

So i think the bug is somewhat unrelated to the changes here.

ibraheemdev:
I went through the ecosystem report, and most of the changes seem to be cases where we are able to correctly catch more type errors.

There were one or two cases where the literal promotion issue came up, but as discussed I think that can be addressed in a separate PR.

There were also a couple instances where lists were being used as heterogenous types, e.g.,
```py
def y(a: int, b: int, c: str): ...
y(*[1, 2, "3"])
``` 

pyright actually doesn't error on this, because it infers `list[Unknown]`, but we infer `list[Unknown | int | str]`, and so error.

carljm:
> pyright actually doesn't error on this, because it infers `list[Unknown]`

Interesting that pyright gives up in that situation. Mypy infers `list[object]` and also errors.

crates/ty_python_semantic/src/types.rs
carljm: https://github.com/astral-sh/ty/issues/1198

PR #20363, opened 2025-09-12
[ty] Fix subtyping/assignability of function- and class-literal types to callback protocols

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/377.

We were treating any function as being assignable to any callback protocol, because we were trying to figure out a type's `Callable` supertype by looking up the `__call__` attribute on the type's meta-type. But a function-literal's meta-type is `types.FunctionType`, and `types.FunctionType.__call__` is `(...) -> Any`, which is not very helpful!

While working on this PR, I also realised that assignability between class-literals and callback protocols was somewhat broken too, so I fixed that at the same time.

## Test Plan

Added mdtests

### DIFF
--- a/crates/ty_python_semantic/src/types/protocol_class.rs
+++ b/crates/ty_python_semantic/src/types/protocol_class.rs
@@ -541,6 +541,15 @@ impl<'a, 'db> ProtocolMember<'a, 'db> {
     ) -> ConstraintSet<'db> {
         match &self.kind {
             ProtocolMemberKind::Method(method) => {
+                // `__call__` members must be special cased for several reasons:
+                //
+                // 1. Looking up `__call__` on the meta-type of a `Callable` type returns `Place::Unbound` currently
+                // 2. Looking up `__call__` on the meta-type of a function-literal type currently returns a type that
+                //    has an extremely vague signature (`(*args, **kwargs) -> Any`), which is not useful for protocol
+                //    checking.
+                // 3. Looking up `__call__` on the meta-type of a class-literal, generic-alias or subclass-of type is
+                //    unfortunately not sufficient to obtain the `Callable` supertypes of these types, due to the
+                //    complex interaction between `__new__`, `__init__` and metaclass `__call__`.
                 let attribute_type = if self.name == "__call__" {
                     let Some(attribute_type) = other.into_callable(db) else {
                         return ConstraintSet::from(false);


AlexWaygood:
The conformance suite diff looks excellent -- the new diagnostics are all on lines that have `# E` next to them!

AlexWaygood:
# Mypy_primer analysis

## psycopg

```diff
psycopg (https://github.com/psycopg/psycopg)
+ tests/typing_example.py:100:21: error[invalid-argument-type] Argument to bound method `connect` is incorrect: Expected `RowFactory[tuple[Any, ...]] | None`, found `def int_row_factory(cursor: Cursor[Any] | AsyncCursor[Any]) -> (Sequence[int], /) -> int`
+ tests/typing_example.py:110:21: error[invalid-argument-type] Argument to bound method `connect` is incorrect: Expected `RowFactory[tuple[Any, ...]] | None`, found `bound method <class 'Person'>.row_factory(cursor: Cursor[Any] | AsyncCursor[Any]) -> (Sequence[str], /) -> Person`
+ tests/typing_example.py:134:43: error[invalid-argument-type] Argument to bound method `connect` is incorrect: Expected `AsyncRowFactory[tuple[Any, ...]] | None`, found `def int_row_factory(cursor: Cursor[Any] | AsyncCursor[Any]) -> (Sequence[int], /) -> int`
+ tests/typing_example.py:144:43: error[invalid-argument-type] Argument to bound method `connect` is incorrect: Expected `AsyncRowFactory[tuple[Any, ...]] | None`, found `bound method <class 'Person'>.row_factory(cursor: Cursor[Any] | AsyncCursor[Any]) -> (Sequence[str], /) -> Person`
```

This is pretty interesting. Here's a minimal repro of what's going on here:

```py
from typing import Protocol, Any

def int_maker() -> int:
    raise NotImplementedError

class RowMaker[Row = tuple[Any, ...]](Protocol):
    def __call__(self) -> Row: ...

class Connection[Row = tuple[Any, ...]]:
    @classmethod
    def connect(cls, factory: RowMaker[Row]) -> Row:
        raise NotImplementedError

connect = Connection.connect
reveal_type(connect(int_maker))
```

Mypy issues no complaints on this code and reveals `int` on the last line. Pyright, however, says:

```
Argument of type "() -> int" cannot be assigned to parameter "factory" of type "RowMaker[tuple[Any, ...]]" in function "connect"
  Type "() -> int" is not assignable to type "() -> tuple[Any, ...]"
    Function return type "int" is incompatible with type "tuple[Any, ...]"
      "int" is not assignable to "tuple[Any, ...]"  (reportArgumentType)
Type of "connect(int_maker)" is "tuple[Any, ...]"
```

And with this PR, ty would match pyright's behaviour. The difference comes down to what the inferred type of `connect` is after the `connect = Connection.connect` assignment. Both ty and pyright specialize the method as a result of the binding, and they use the TypeVar default to specialize the method since no explicit specialization was provided for the `Connect` class. This then causes the `connect(int_maker)` call to fail, since `int` is not assignable to `tuple[Any, ...]`, the default specialization of `Connection`. But mypy appears to defer solving the TypeVar somehow, which means that it is able to solve the type variable to `int`.

Anyway, I think the fact that we are emulating pyright's behaviour here means that this is pretty defensible behaviour from us!

## werkzeug

```diff
werkzeug (https://github.com/pallets/werkzeug)
+ tests/test_formparser.py:437:45: error[invalid-argument-type] Argument to bound method `__init__` is incorrect: Expected `TStreamFactory | None`, found `def stream_factory() -> Unknown`
```

This seems like a clear true positive. [`TStreamFactory`](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/formparser.py#L40-L47) has a `__call__` method with 3 required parameters, and [`stream_factory`](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/tests/test_formparser.py#L434-L435) will fail at runtime if you provide any parameters to it.

---

Not sure I have the energy on a Friday evening to trawl through the pandas codebase figuring out what the new diagnostics there are about...

PR #20364, opened 2025-09-12
[ty] Remove `Self` from generic context when binding `Self`

sharkdp:
## Summary

This mainly removes an internal inconsistency, where we didn't remove the `Self` type variable when eagerly binding `Self` to an instance type. It has no observable effect, apparently.

builds on top of https://github.com/astral-sh/ruff/pull/20328

## Test Plan

None


sharkdp:
> Oof, good catch!

You told me that this is might be a problem :-)

dcreager:
> You told me that this is might be a problem

And then promptly forgot to look for it in #20328 :sweat_smile:

PR #20367, opened 2025-09-12
[ty] Temporary hack to reduce false positives around `builtins.open()`

AlexWaygood:
## Summary

https://github.com/astral-sh/ruff/pull/20165 added a lot of false positives around calls to `builtins.open()`, because our missing support for PEP-613 type aliases means that we don't understand typeshed's overloads for `builtins.open()` at all yet, and therefore always select the first overload. This didn't use to matter very much, but now that we have a much stricter implementation of protocol assignability/subtyping it matters a lot, because most of the stdlib functions dealing with I/O (`pickle`, `marshal`, `io`, `json`, etc.) are annotated in typeshed as taking in protocols of some kind.

In lieu of full PEP-613 support, which is blocked on various things and might not land in time for our next alpha release, this PR adds some temporary special-casing for `builtins.open()` to avoid the false positives. We just infer `Todo` for anything that isn't meant to match typeshed's first `open()` overload. This should be easy to rip out again once we have proper support for PEP-613 type aliases, which hopefully should be pretty soon!

## Test Plan

Added an mdtest

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/builtins.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/builtins.md
@@ -162,3 +162,22 @@ def _(x: A | B, y: list[int]):
         reveal_type(x)  # revealed: B & ~A
         reveal_type(isinstance(x, B))  # revealed: Literal[True]
 ```
+
+## Calls to `open()`
+
+We do not fully understand typeshed's overloads for `open()` yet, due to missing support for PEP-613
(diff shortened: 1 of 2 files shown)


AlexWaygood:
That looks like the kind of ecosystem report I expected!

AlexWaygood:
(The mypy_primer panic on sympy is happening on multiple PRs right now so is almost certainly unrelated)

AlexWaygood:
This just reduces the fallout of a change that's unreleased, so I'll add the `internal` label to keep it out of the changelog

PR #20368, opened 2025-09-12
[ty] Improve assignability/subtyping between two protocol types

AlexWaygood:
## Summary

We currently have reasonably good assignability and subtyping checks between nominal types and protocol types if the protocol has a method member or an attribute member. For example, these assertions all pass:

```py
from ty_extensions import static_assert, is_subtype_of
from typing import Iterable, Iterator, Callable

class IsIterable:
    def __iter__(self) -> Iterator[str]: ...

class NotIterable:
    __iter__ = None

class AlsoNotIterable:
    def __init__(self):
        self.__init__: Callable[[], Iterator[int]] = lambda: iter(range(42))

static_assert(is_subtype_of(IsIterable, Iterable[str]))
static_assert(not is_subtype_of(NotIterable, Iterable[str]))
static_assert(not is_subtype_of(AlsoNotIterable, Iterator[int]))
```

However, we do not currently have a good implementation of subtyping and assignability _between_ protocol types: we currently think that `Iterator[str]` is a subtype of `Iterator[int]`, and vice versa, for example, because both protocols have `__iter__` and `__next__` method members, and we do not currently compare the types of these members when doing subtyping/assignability checks _between_ protocols (we only check for the presence of those members).

This PR fixes that.

Fixes https://github.com/astral-sh/ty/issues/306
Fixes https://github.com/astral-sh/ty/issues/733
Fixes https://github.com/astral-sh/ty/issues/1089

## Implementation details

The change that I _wanted_ to make in this PR is the change to `protocol_class.rs`. Unfortunately, I quickly ran into stack overflows after making that change, however, due to large cycles that were spread across `is_subtype_of` and `is_disjoint_from`. Looking at backtraces, the root of the problem was clear: `has_relation_to` for intersections calls `is_disjoint_from`:

https://github.com/astral-sh/ruff/blob/3771f1567c778776777679674b8cffdc90aa51aa/crates/ty_python_semantic/src/types.rs#L1653-L1665

...and `is_disjoint_from` for intersections calls `has_relation_to`...

https://github.com/astral-sh/ruff/blob/3771f1567c778776777679674b8cffdc90aa51aa/crates/ty_python_semantic/src/types.rs#L2196-L2209

...meaning that existing mdtests ended up running into an infinite loop when checking whether intersections of protocols were subtypes of other types. This infinite loop could not be caught by our existing cycle-detector infrastructure, because a fresh cycle detector would be created in the call to `has_relation_to` call from `is_disjoint_from_impl`, and this cycle detector would not be passed into the `is_disjoint_from` call inside `has_relation_to`.

I therefore had to make a much more invasive change here than I wanted to, in order to fix these stack overflows. Rather than accepting only one cycle detector as an input parameter, `has_relation_to_impl` must now accept two cycle detectors: one `HasRelationToVisitor` and one `IsDisjointFromVisitor`. This allows `has_relation_to` to call `is_disjoint_from_impl` rather than `is_disjoint_from`, meaning that the same cycle detector can be passed between `has_relation_to` calls and `is_disjoint_from` calls. The cycle can therefore be detected and stopped in its tracks before a stack overflow occurs.

Similar changes also needed to be made to `is_disjoint_from_impl`: it now accepts a `HasRelationToVisitor` as well as an `IsDisjointFromVisitor`.

## Performance

There are some regressions of up to 5% or so on some benchmarks. I think this is just because we're doing more work now, but I'm happy to hear suggestions if anybody has any ideas for how to optimise this.

## Conformances suite impact

See https://github.com/astral-sh/ruff/pull/20368#issuecomment-3286597297

## Ecosystem impact

See https://github.com/astral-sh/ruff/pull/20368#issuecomment-3368395654

## Test Plan

Mdtests updated

### DIFF
--- a/crates/ty_python_semantic/src/types/instance.rs
+++ b/crates/ty_python_semantic/src/types/instance.rs
@@ -157,6 +157,10 @@ impl<'db> Type<'db> {
         // recognise `str` as a subtype of `Container[str]`.
         structurally_satisfied.or(db, || {
             if let Protocol::FromClass(class) = protocol.inner {
+                // if `self` and `other` are *both* protocols, we also need to treat `self` as if it
+                // were a nominal type, or we won't consider a protocol `P` that explicitly inherits
+                // from a protocol `Q` to be a subtype of `Q` to be a subtype of `Q` if it overrides
+                // `Q`'s members in a Liskov-incompatible way.
                 let type_to_test = if let Type::ProtocolInstance(ProtocolInstanceType {
                     inner: Protocol::FromClass(class),
                     ..


AlexWaygood:
> ```diff
> +callables_annotation.py:157:5: error[invalid-assignment] Object of type `Proto7` is not assignable to `Proto6`
> ```

According to the conformance test, we're not meant to emit an error on this one ([here](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/callables_annotation.py#L157)). All other new diagnostics on the conformance suite look correct.

Edit: this is a pre-existing bug in `Callable` assignability. I opened a separate issue for this (https://github.com/astral-sh/ty/issues/1257)

AlexWaygood:
# Ecosystem analysis

I was expecting this PR to reduce diagnostics across the ecosystem overall, but it seems like on net it leads to more diagnostics, which is unfortunate. I _have_ spent a while now trying to reduce the number of underlying issues here, and the diagnostic diff _is_ much better than it was when I first opened the PR (we've gone from +297,-128 a few weeks ago, to +280,-157 now). PRs such as https://github.com/astral-sh/ruff/pull/20602, https://github.com/astral-sh/ruff/pull/20629, https://github.com/astral-sh/ruff/pull/20603, https://github.com/astral-sh/ruff/pull/20591 and github.com/astral-sh/ruff/pull/20593 have all helped reduce the impact of this PR by fixing underlying issues. It's possible I could carry on with this a little bit, but I think the remaining diagnostics are mostly acceptable and/or hard to fix.

Many new diagnostics on this PR are due to narrowing behaviour that is technically correct, but which I worry may appear quite pedantic to users. For example, an adapted repro from `aiohttp` is:

```py
from typing import Iterable
import socket

def f(x: Iterable[socket.socket] | socket.socket):
    if isinstance(x, Iterable):
        reveal_type(x)  # revealed: Iterable[socket] | (socket & Iterable[object])
        for s in x:
            reveal_type(s)  # revealed: object
            needs_a_socket(s)  # error: "Expected `socket`, found `object`"
    else:
        needs_a_socket(x)

def needs_a_socket(s: socket.socket): ...
```

This is, of course, strictly correct: typeshed (correctly) does not mark `socket.socket` as `@final`, since it can be subclassed at runtime. It's entirely possible that a user _could_ create a subclass of `socket.socket` that was iterable, so inferring `object` for `s` here is entirely correct. Unfortunately it also differs from the behaviour of all currently existing type checkers, and I'm not totally sure how we can tailor our diagnostics so that it's clear to users why we're unexpectedly inferring `object` in many places. Still, this is consistent behaviour with how we do narrowing elsewhere, so it's not really a _bug_ in this PR.

There are lots of diagnostics like this going away, due to https://github.com/astral-sh/ty/issues/733 being fixed by this PR:

```
[error] invalid-return-type - Return type does not match returned value: expected `str`, found `Literal[b""]`
```

And there's also lots of churning regarding diagnostics for `str.join`, due to https://github.com/astral-sh/ty/issues/306 being fixed by this PR.

Not an exhaustive analysis, but here's a spot check of some of the other diagnostics in the ecosystem that stood out to me while I skimmed through:

## bokeh

One of the bokeh hits is:

```
[error] invalid-return-type - [:510:16](https://github.com/bokeh/bokeh/blob/17c46de8f0df63051a08e1c756587702e249ea7f/src/bokeh/model/model.py#L510) - Return type does not match returned value: expected `Iterable[bokeh.model.model.Model]`, found `Iterable[bokeh.model.model.Model]`
```

Debug-printing the types gives:

<details>

```
error[invalid-return-type]: Return type does not match returned value
   --> src/bokeh/model/model.py:510:16
    |
508 |         '''
509 |         from ..core.query import find
510 |         return find(self.references(), selector)
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `ProtocolInstance(
    ProtocolInstanceType {
        inner: FromClass(
            Generic(
                GenericAlias {
                    origin: ClassLiteral {
                        name: Name("Iterable"),
                        body_scope: ScopeId {
                            [salsa id]: Id(534a),
                            file: File {
                                path: Vendored(
                                    VendoredPathBuf(
                                        "stdlib/typing.pyi",
                                    ),
                                ),
                                status: Exists,
                                permissions: Some(
                                    292,
                                ),
                                revision: FileRevision(
                                    1689436316,
                                ),
                            },
                            file_scope_id: FileScopeId(
                                106,
                            ),
                        },
                        known: Some(
                            Iterable,
                        ),
                        deprecated: None,
                        dataclass_params: None,
                        dataclass_transformer_params: None,
                    },
                    specialization: Specialization {
                        generic_context: GenericContext {
                            variables_inner: {
                                BoundTypeVarInstance {
                                    typevar: TypeVarInstance {
                                        name: Name("_T_co"),
                                        definition: Some(
                                            Definition {
                                                [salsa id]: Id(6c3d),
                                                file: File {
                                                    path: Vendored(
                                                        VendoredPathBuf(
                                                            "stdlib/typing.pyi",
                                                        ),
                                                    ),
                                                    status: Exists,
                                                    permissions: Some(
                                                        292,
                                                    ),
                                                    revision: FileRevision(
                                                        1689436316,
                                                    ),
                                                },
                                                file_scope: FileScopeId(
                                                    0,
                                                ),
                                                place: Symbol(
                                                    ScopedSymbolId(
                                                        74,
                                                    ),
                                                ),
                                                kind: Assignment(
                                                    AssignmentDefinitionKind {
                                                        target_kind: Single,
                                                        value: AstNodeRef {
                                                            kind: ExprCall,
                                                            range: 21869..21901,
                                                        },
                                                        target: AstNodeRef {
                                                            kind: ExprName,
                                                            range: 21861..21866,
                                                        },
                                                    },
                                                ),
                                                is_reexported: true,
                                            },
                                        ),
                                        _bound_or_constraints: None,
                                        explicit_variance: Some(
                                            Covariant,
                                        ),
                                        _default: None,
                                        kind: Legacy,
                                    },
                                    binding_context: Definition(
                                        Definition {
                                            [salsa id]: Id(6c8f),
                                            file: File {
                                                path: Vendored(
                                                    VendoredPathBuf(
                                                        "stdlib/typing.pyi",
                                                    ),
                                                ),
                                                status: Exists,
                                                permissions: Some(
                                                    292,
                                                ),
                                                revision: FileRevision(
                                                    1689436316,
                                                ),
                                            },
                                            file_scope: FileScopeId(
                                                0,
                                            ),
                                            place: Symbol(
                                                ScopedSymbolId(
                                                    108,
                                                ),
                                            ),
                                            kind: Class(
                                                AstNodeRef {
                                                    kind: StmtClassDef,
                                                    range: 27489..27607,
                                                },
                                            ),
                                            is_reexported: true,
                                        },
                                    ),
                                }: GenericContextTypeVarOptions {
                                    should_promote_literals: false,
                                },
                            },
                        },
                        types: [
                            NominalInstance(
                                NominalInstanceType(
                                    NonTuple(
                                        NonGeneric(
                                            ClassLiteral {
                                                name: Name("Model"),
                                                body_scope: ScopeId {
                                                    [salsa id]: Id(1001),
                                                    file: File {
                                                        path: System(
                                                            "/Users/alexw/dev/bokeh/src/bokeh/model/model.py",
                                                        ),
                                                        status: Exists,
                                                        permissions: Some(
                                                            33188,
                                                        ),
                                                        revision: FileRevision(
                                                            32450546247609911550292892181,
                                                        ),
                                                    },
                                                    file_scope_id: FileScopeId(
                                                        1,
                                                    ),
                                                },
                                                known: None,
                                                deprecated: None,
                                                dataclass_params: None,
                                                dataclass_transformer_params: None,
                                            },
                                        ),
                                    ),
                                ),
                            ),
                        ],
                        materialization_kind: None,
                        tuple_inner: None,
                    },
                },
            ),
        ),
        _phantom: PhantomData<()>,
    },
)`, found `ProtocolInstance(
    ProtocolInstanceType {
        inner: FromClass(
            Generic(
                GenericAlias {
                    origin: ClassLiteral {
                        name: Name("Iterable"),
                        body_scope: ScopeId {
                            [salsa id]: Id(534a),
                            file: File {
                                path: Vendored(
                                    VendoredPathBuf(
                                        "stdlib/typing.pyi",
                                    ),
                                ),
                                status: Exists,
                                permissions: Some(
                                    292,
                                ),
                                revision: FileRevision(
                                    1689436316,
                                ),
                            },
                            file_scope_id: FileScopeId(
                                106,
                            ),
                        },
                        known: Some(
                            Iterable,
                        ),
                        deprecated: None,
                        dataclass_params: None,
                        dataclass_transformer_params: None,
                    },
                    specialization: Specialization {
                        generic_context: GenericContext {
                            variables_inner: {
                                BoundTypeVarInstance {
                                    typevar: TypeVarInstance {
                                        name: Name("_T_co"),
                                        definition: Some(
                                            Definition {
                                                [salsa id]: Id(6c3d),
                                                file: File {
                                                    path: Vendored(
                                                        VendoredPathBuf(
                                                            "stdlib/typing.pyi",
                                                        ),
                                                    ),
                                                    status: Exists,
                                                    permissions: Some(
                                                        292,
                                                    ),
                                                    revision: FileRevision(
                                                        1689436316,
                                                    ),
                                                },
                                                file_scope: FileScopeId(
                                                    0,
                                                ),
                                                place: Symbol(
                                                    ScopedSymbolId(
                                                        74,
                                                    ),
                                                ),
                                                kind: Assignment(
                                                    AssignmentDefinitionKind {
                                                        target_kind: Single,
                                                        value: AstNodeRef {
                                                            kind: ExprCall,
                                                            range: 21869..21901,
                                                        },
                                                        target: AstNodeRef {
                                                            kind: ExprName,
                                                            range: 21861..21866,
                                                        },
                                                    },
                                                ),
                                                is_reexported: true,
                                            },
                                        ),
                                        _bound_or_constraints: None,
                                        explicit_variance: Some(
                                            Covariant,
                                        ),
                                        _default: None,
                                        kind: Legacy,
                                    },
                                    binding_context: Definition(
                                        Definition {
                                            [salsa id]: Id(6c8f),
                                            file: File {
                                                path: Vendored(
                                                    VendoredPathBuf(
                                                        "stdlib/typing.pyi",
                                                    ),
                                                ),
                                                status: Exists,
                                                permissions: Some(
                                                    292,
                                                ),
                                                revision: FileRevision(
                                                    1689436316,
                                                ),
                                            },
                                            file_scope: FileScopeId(
                                                0,
                                            ),
                                            place: Symbol(
                                                ScopedSymbolId(
                                                    108,
                                                ),
                                            ),
                                            kind: Class(
                                                AstNodeRef {
                                                    kind: StmtClassDef,
                                                    range: 27489..27607,
                                                },
                                            ),
                                            is_reexported: true,
                                        },
                                    ),
                                }: GenericContextTypeVarOptions {
                                    should_promote_literals: false,
                                },
                            },
                        },
                        types: [
                            NominalInstance(
                                NominalInstanceType(
                                    NonTuple(
                                        NonGeneric(
                                            ClassLiteral {
                                                name: Name("Model"),
                                                body_scope: ScopeId {
                                                    [salsa id]: Id(c983),
                                                    file: File {
                                                        path: System(
                                                            "/Users/alexw/dev/bokeh/src/bokeh/model/model.pyi",
                                                        ),
                                                        status: Exists,
                                                        permissions: Some(
                                                            33188,
                                                        ),
                                                        revision: FileRevision(
                                                            32450546247609911550292943140,
                                                        ),
                                                    },
                                                    file_scope_id: FileScopeId(
                                                        1,
                                                    ),
                                                },
                                                known: None,
                                                deprecated: None,
                                                dataclass_params: Some(
                                                    DataclassParams(
                                                        REPR | EQ | MATCH_ARGS,
                                                    ),
                                                ),
                                                dataclass_transformer_params: None,
                                            },
                                        ),
                                    ),
                                ),
                            ),
                        ],
                        materialization_kind: None,
                        tuple_inner: None,
                    },
                },
            ),
        ),
        _phantom: PhantomData<()>,
    },
)`
```

</details>

Here's the diff between the two types: https://www.diffchecker.com/ZxRy9rMI/. It appears that one of the `Model` types comes from `bokeh/src/bokeh/model/model.pyi`, and the other `Model` type comes from `bokeh/src/bokeh/model/model.py` (notice the difference in file extension).

I suppose that this is only an issue because the diagnostic is being emitted while checking `model/model.py` itself.

This is quite strange but seems unrelated to this PR, and it was a pre-existing problem. I've opened https://github.com/astral-sh/ty/issues/1306.

## homeassistant

There's lots of new diagnostics like this:

```
[error] invalid-argument-type - :58:9 - Argument to bound method `__init__` is incorrect: Expected `(() -> Awaitable[dict[str, Any]]) | None`, found `def async_update() -> CoroutineType[Any, Any, IssData]`
```

This relates to https://github.com/astral-sh/ty/issues/500 (so I suppose it will be solved by https://github.com/astral-sh/ty/issues/623). A [repro of this false positive](https://play.ty.dev/9b4dee70-b60e-4fc0-8eb4-bd3439de9120) that occurs on `main` and doesn't involve protocols is:

```py
from typing import Any, Callable

class Foo[T = dict[str, Any]]:
    def __init__(self, data: Callable[..., T]): ...

def returns_bool() -> bool:
    return True

# error: [invalid-argument-type] "Argument to bound method `__init__` is incorrect: Expected `(...) -> dict[str, Any]`, found `def returns_bool() -> bool`"
# revealed: `Foo[dict[str, Any]]`
reveal_type(Foo(returns_bool))
```

## ibis

```
ibis/common/egraph.py
[error] unresolved-attribute - :732:44 - Type `list[Rewrite]` has no attribute `matcher`
[error] unresolved-attribute - :733:25 - Type `list[Rewrite]` has no attribute `applier`
```

These are because on `main` we infer this:

```py
from typing import Iterable

def f[T](x: T | Iterable[T]) -> T:
    raise NotImplementedError

reveal_type(f(["foo"]))  # revealed: Unknown
```

whereas with this PR, we infer this:

```py
from typing import Iterable

def f[T](x: T | Iterable[T]) -> T:
    raise NotImplementedError

reveal_type(f(["foo"]))  # revealed: list[Unknown | str]
```

when what is desired is of course:

```py
from typing import Iterable

def f[T](x: T | Iterable[T]) -> T:
    raise NotImplementedError

reveal_type(f(["foo"]))  # revealed: `str`
```

So this should also hopefully be solved by https://github.com/astral-sh/ty/issues/623.

## isort

Here's a minimal repro of what's going on here:

```py
from typing import Iterable, Self

class chain[T]:
    def __new__(cls, *iterables: Iterable[T]) -> Self:
        raise NotImplementedError

def yield_ints() -> Iterable[int]:
    yield 42

# revealed: chain[None]
reveal_type(chain((None,), (None,)))

# revealed: chain[Unknown]
reveal_type(chain(yield_ints(), yield_ints()))

# revealed: chain[Unknown]
# error: [invalid-argument-type] "Argument to function `__new__` is incorrect: Expected `Iterable[None]`, found `Iterable[int]`"
reveal_type(chain(yield_ints(), (None,)))
```

On `main`, the final line reveals `chain[None]` and the error on the final line is not emitted. It's possible that this might be related to https://github.com/astral-sh/ty/issues/1308 ? Not sure.

### meson

```
unittests/allplatformstests.py
[error] index-out-of-bounds - :3861:26 - Index 2 is out of bounds for tuple `tuple[str, (json_node) -> Unknown]` with length 2
[error] index-out-of-bounds - :3861:26 - Index 2 is out of bounds for tuple `tuple[str, (kwargs) -> Unknown]` with length 2
[error] index-out-of-bounds - :3861:26 - Index 2 is out of bounds for tuple `tuple[str, (node_list) -> Unknown]` with length 2
```

These are all https://github.com/astral-sh/ty/issues/1248: it's an unannotated heterogeneous dictionary.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -545,7 +546,8 @@ impl<'db> ClassType<'db> {
         db: &'db dyn Db,
         other: Self,
         relation: TypeRelation,
-        visitor: &HasRelationToVisitor<'db>,
+        relation_visitor: &HasRelationToVisitor<'db>,
+        disjointness_visitor: &IsDisjointVisitor<'db>,
```
carljm, line 550: Disjointness is also a relation. If we added TypeRelation::Disjoint and used `has_relation_to_impl` for implementing `is_disjoint_from`, we could reduce this back to a single `HasRelationToVisitor` and eliminate `IsDisjointVisitor` entirely.

In an ideal world I might say we should prefer to do that refactor as a prior PR, and then make this a much smaller PR on top of it. But I won't argue if you prefer to land this PR as-is and consider that as a possible follow-up. Maybe it's not even worth it; there's just a _big_ increase in LoC in this PR from all the `has_relation_to_impl` calls that now have to explode onto seven lines.

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -149,11 +157,22 @@ impl<'db> Type<'db> {
...
-                self.has_relation_to_impl(
+                let type_to_test = if let Type::ProtocolInstance(ProtocolInstanceType {
+                    inner: Protocol::FromClass(class),
+                    ..
+                }) = self
+                {
+                    Type::non_tuple_instance(db, class)
+                } else {
+                    self
+                };
```
carljm, line 172: This seems notable -- can we add a comment explaining the rationale for this?

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: > Disjointness is also a relation.

That's true, but in terms of the implementation it's very different from the others (at least currently). I think what you'd be looking at would be greatly expanding the size of the `Type::has_relation_to_impl` function, which is already very complicated, hard to refactor, and hard to understand.

> there's just a _big_ increase in LoC in this PR from all the `has_relation_to_impl` calls that now have to explode onto seven lines.

That's true, but I feel less bad about it given what Doug has in store for us in https://github.com/astral-sh/ruff/pull/20677 ;)

PR #20369, opened 2025-09-12
[ty] add cycle handling to BoundMethodType::into_callable_type()

carljm:
## Summary

This looks like it should fix the errors that we've been seeing in sympy in recent mypy-primer runs.

## Test Plan

I wasn't able to reproduce the sympy failures locally; it looks like there is probably a dependency on the order in which files are checked. So I don't have a minimal reproducible example, and wasn't able to add a test :/ Obviously I would be happier if we could commit a regression test here, but since the change is straightforward and clearly desirable, I'm not sure how many hours it's worth trying to track it down.

Mypy-primer is still failing in CI on this PR, because it fails on the "old" ty commit already (i.e. on main). But it passes [on a no-op PR stacked on top of this](https://github.com/astral-sh/ruff/pull/20370), which strongly suggests this PR fixes the problem.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8938,6 +8938,7 @@ fn walk_bound_method_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
     visitor.visit_type(db, method.self_instance(db));
 }
 
+#[allow(clippy::trivially_copy_pass_by_ref)]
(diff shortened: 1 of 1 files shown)


carljm:
I'm going to go ahead and merge this to clear up CI for other PRs. Post-land review is welcome, and if anyone wants to try to track down a minimized repro so we can add a regression test, I certainly won't complain!

PR #20375, opened 2025-09-13
[`ruff`] Add `analyze.string-imports-min-dots` to settings

### LINKED ISSUE #20110: Allow configuring `--min-dots` using configuration file
`ruff analyze graph` has both `--detect-string-imports` and `--min-dots`, but [`tool.ruff.analyze`/`analyze`](https://docs.astral.sh/ruff/settings/#analyze) only has [`detect-string-imports`](https://docs.astral.sh/ruff/settings/#analyze_detect-string-imports).

`analyze.min-dots` should be a thing, if only for consistency.

TaKO8Ki:
## Summary

Fixes #20110

`string_imports_min_dots` was introduced in #19538, so the option has been available since then; this PR adds documentation and a config test.

## Test Plan

Add integration test `string_detection_from_config` to verify the config setting works.

### DIFF
--- a/crates/ruff/tests/analyze_graph.rs
+++ b/crates/ruff/tests/analyze_graph.rs
@@ -247,6 +247,59 @@ fn string_detection() -> Result<()> {
 }
 
 #[test]
+fn string_detection_from_config() -> Result<()> {
+    let tempdir = TempDir::new()?;
+
+    let root = ChildPath::new(tempdir.path());
+
+    // Configure string import detection with a lower min-dots via ruff.toml
+    root.child("ruff.toml").write_str(indoc::indoc! {r#"
+        [analyze]
+        detect-string-imports = true
+        string-imports-min-dots = 1
+    "#})?;
+
+    root.child("ruff").child("__init__.py").write_str("")?;
+    root.child("ruff")
+        .child("a.py")
+        .write_str(indoc::indoc! {r#"
+        import ruff.b
+    "#})?;
+    root.child("ruff")
+        .child("b.py")
+        .write_str(indoc::indoc! {r#"
+        import importlib
+
+        importlib.import_module("ruff.c")
+    "#})?;
+    root.child("ruff").child("c.py").write_str("")?;
+
+    insta::with_settings!({
+        filters => INSTA_FILTERS.to_vec(),
+    }, {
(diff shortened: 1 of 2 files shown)


crates/ruff/tests/analyze_graph.rs
```diff
@@ -246,6 +246,59 @@ fn string_detection() -> Result<()> {
...
+fn string_detection_from_config() -> Result<()> {
+    let tempdir = TempDir::new()?;
+
+    let root = ChildPath::new(tempdir.path());
+
+    // Configure string import detection with a lower min-dots via ruff.toml
+    root.child("ruff.toml").write_str(indoc::indoc! {r#"
+        [analyze]
+        detect-string-imports = true
+        string-imports-min-dots = 1
```
MichaReiser, line 259: I think this should be `min-dots` for consistency with the CLI

```
Options:
      --direction <DIRECTION>            The direction of the import map. By default, generates a dependency map, i.e., a map from file to files that it depends on. Use `--direction
                                         dependents` to generate a map from file to files that depend on it [default: dependencies] [possible values: dependencies, dependents]
      --detect-string-imports            Attempt to detect imports from string literals
      --min-dots <MIN_DOTS>              The minimum number of dots in a string import to consider it a valid import
      --preview                          Enable preview mode. Use `--no-preview` to disable
      --target-version <TARGET_VERSION>  The minimum Python version that should be supported [possible values: py37, py38, py39, py310, py311, py312, py313, py314]
      --python <PYTHON>                  Path to a virtual environment to use for resolving additional dependencies
  -h, --help
```

MichaReiser (approved):
This looks good but I think we should rename the option to match the CLI

crates/ruff/tests/analyze_graph.rs
TaKO8Ki: Ok. I have renamed it to `min-dots`!

MichaReiser:
Oh wait, I made a mistake. I assumed you added the new option but that isn't the case. This is a pre-existing option (if I understand your summary correctly) and this PR only adds documentation. 

If that's the case, then we shouldn't rename the option because that would break existing users

TaKO8Ki:
@MichaReiser Thanks for the clarification. I had the same concern, so I didn’t rename the option; I’ll keep the existing name as is.

PR #13280, opened 2024-09-07
[`pydoclint`] Add `docstring-missing-parameter` and `docstring-extraneous-parameter` (`DOC101`, `DOC102`)

augustelalande:
## Summary

Add `docstring-missing-parameter` and `docstring-extraneous-parameter` (`DOC101`, `DOC102`). These rules check that the parameters defined in a functions signature match thos defined in the docstring.

Part of #12434.

## Test Plan

Test cases added.

Issue #12434, opened 2024-07-21
Implement `pydoclint`


augustelalande:
This issue tracks the implementation of [pydoclint](https://jsh9.github.io/pydoclint/violation_codes.html). Following discussion in #11471 it was decided to prefer `pydoclint` over `darglint` for implementation is ruff since they cover mostly the same ruleset, and `pydoclint` is still actively developed (unlike `darglint`). This issue should then supersede #458.

## Rules
* [ ] `DOC101` Docstring contains fewer arguments than in function signature
* [ ] `DOC102` Docstring contains more arguments than in function signature
* [ ] `DOC103` Docstring arguments are different from function arguments
* [ ] `DOC104` Arguments are the same in the docstring and the function signature, but are in a different order
* [ ] `DOC105` Argument names match, but type hints do not match
* [ ] `DOC106` The option `--arg-type-hints-in-signature` is `True` but there are no argument type hints in the signature
* [ ] `DOC107` The option `--arg-type-hints-in-signature` is `True` but not all args in the signature have type hints
* [ ] `DOC108` The option `--arg-type-hints-in-signature` is `False` but there are argument type hints in the signature
* [ ] `DOC109` The option `--arg-type-hints-in-docstring` is `True` but there are no type hints in the docstring arg list
* [ ] `DOC110` The option `--arg-type-hints-in-docstring` is `True` but not all args in the docstring arg list have type hints
* [ ] `DOC111` The option `--arg-type-hints-in-docstring` is `False` but there are type hints in the docstring arg list
* [ ] `DOC201` Function/method does not have a return section in docstring
* [ ] `DOC202` Function/method has a return section in docstring, but there are no return statements or annotations
* [ ] `DOC203` Return type(s) in the docstring not consistent with the return annotation
* [ ] `DOC301` `__init__()` should not have a docstring; please combine it with the docstring of the class
* [ ] `DOC302` The class docstring does not need a "Returns" section, because `__init__()` cannot return anything
* [ ] `DOC303` The `__init__()` docstring does not need a "Returns" section, because it cannot return anything
* [ ] `DOC304` Class docstring has an argument/parameter section; please put it in the `__init__()` docstring
* [ ] `DOC305` Class docstring has a "Raises" section; please put it in the `__init__()` docstring
* [ ] `DOC306` The class docstring does not need a "Yields" section, because `__init__()` cannot yield anything
[...]

PR #20376, opened 2025-09-13
[`pydoclint`] Implement `docstring-extraneous-parameter` (`DOC102`)

augustelalande:
## Summary

Implement `docstring-extraneous-parameter` (`DOC102`). This rule checks that all parameters present in a functions docstring are also present in its signature.

Split from #13280, per this [comment](https://github.com/astral-sh/ruff/pull/13280#issuecomment-3280575506).

Part of #12434.

## Test Plan

Test cases added.

### DIFF
--- a/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
+++ b/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
@@ -7,8 +7,8 @@ use ruff_python_ast::{self as ast, Expr, Stmt, visitor};
 use ruff_python_semantic::analyze::{function_type, visibility};
 use ruff_python_semantic::{Definition, SemanticModel};
 use ruff_python_stdlib::identifiers::is_identifier;
-use ruff_source_file::NewlineWithTrailingNewline;
-use ruff_text_size::{Ranged, TextRange, TextSize};
+use ruff_source_file::{LineRanges, NewlineWithTrailingNewline};
+use ruff_text_size::{Ranged, TextLen, TextRange, TextSize};
 
 use crate::Violation;
 use crate::checkers::ast::Checker;
@@ -588,10 +588,10 @@ fn parse_parameters_google(content: &str, content_start: TextSize) -> Vec<Parame
     };
     let indentation = &first_arg[..first_arg.len() - first_arg.trim_start().len()];
 
-    let mut current_pos = 0;
+    let mut current_pos = TextSize::ZERO;
     for line in content.lines() {
         let line_start = current_pos;
-        let line_end = current_pos + line.len();
+        current_pos = content.full_line_end(line_start);
 
         if let Some(entry) = line.strip_prefix(indentation) {
             if entry
@@ -600,27 +600,25 @@ fn parse_parameters_google(content: &str, content_start: TextSize) -> Vec<Parame
                 .is_some_and(|first_char| !first_char.is_whitespace())
             {
                 let Some((before_colon, _)) = entry.split_once(':') else {
-                    current_pos = line_end + 1;
                     continue;
                 };
                 if let Some(param) = before_colon.split_whitespace().next() {
                     let param_name = param.trim_start_matches('*');
                     if is_identifier(param_name) {
-                        let param_start = line_start + indentation.len();
-                        let param_end = param_start + param.len();
+                        let param_start = line_start + indentation.text_len();
+                        let param_end = param_start + param.text_len();
 
                         entries.push(ParameterEntry {
                             name: param_name,
                             range: TextRange::new(
-                                content_start + TextSize::try_from(param_start).unwrap(),
-                                content_start + TextSize::try_from(param_end).unwrap(),
+                                content_start + param_start,
+                                content_start + param_end,
                             ),
                         });
                     }
                 }
             }
         }
-        current_pos = line_end + 1;
     }
     entries
 }
@@ -643,10 +641,10 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
     };
     let indentation = &dashes[..dashes.len() - dashes.trim_start().len()];
 
-    let mut current_pos = dashes.len() + 1;
+    let mut current_pos = content.full_line_end(dashes.text_len());
     for potential in lines {
         let line_start = current_pos;
-        let line_end = current_pos + potential.len();
+        current_pos = content.full_line_end(line_start);
 
         if let Some(entry) = potential.strip_prefix(indentation) {
             if entry
@@ -658,21 +656,20 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
                     let param = before_colon.trim_end();
                     let param_name = param.trim_start_matches('*');
                     if is_identifier(param_name) {
-                        let param_start = line_start + indentation.len();
-                        let param_end = param_start + param.len();
+                        let param_start = line_start + indentation.text_len();
+                        let param_end = param_start + param.text_len();
 
                         entries.push(ParameterEntry {
                             name: param_name,
                             range: TextRange::new(
-                                content_start + TextSize::try_from(param_start).unwrap(),
-                                content_start + TextSize::try_from(param_end).unwrap(),
+                                content_start + param_start,
+                                content_start + param_end,
                             ),
                         });
                     }
                 }
             }
         }
-        current_pos = line_end + 1;
     }
     entries
 }


crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -448,33 +545,118 @@ impl<'a> DocstringSections<'a> {
...
+///     a (int): The first number to add.
+///     b (int): The second number to add.
+/// ```
+fn parse_parameters_google(content: &str) -> Vec<&str> {
+    let mut entries: Vec<&str> = Vec::new();
+    // Find first entry to determine indentation
+    let Some(first_arg) = content.lines().next() else {
+        return entries;
+    };
+    let indentation = &first_arg[..first_arg.len() - first_arg.trim_start().len()];
```
ntBre, line 589: This doesn't seem like a big deal either way, but I noticed that `parse_raises_google` doesn't trim indentation first, unlike `parse_raises_numpy`. Could we get away with just an immediate `split_once` call followed by a `trim()` call to strip the whitespace? Again, it seems fine either way, just an idea to more closely mirror the `raises` version.

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-extraneous-parameter_DOC102_google.py.snap
```diff
@@ -0,0 +1,142 @@
...
+160 |             number (int or float): The number to add to the current value.
+    |
+help: Remove the extraneous parameter from the docstring
+
+DOC102 Documented parameter `value_str` is not in the function's signature
+   --> DOC102_google.py:174:9
+    |
+172 |         Creates a Calculator instance from a string representation of a number.
+173 |
+174 |         Args:
```
ntBre, line 153: Do you think it would be helpful to show a sub-diagnostic or secondary annotation highlighting the parameter list itself? Obviously they're nearby in the source code, but it might be nice to see them both in the diagnostic. Just an idea since we have support for multiple spans now.

crates/ruff_linter/resources/test/fixtures/pydoclint/DOC102_numpy.py
```diff
@@ -0,0 +1,274 @@
...
+def add_numbers(b):
+    """
+    Adds two numbers and returns the result.
+
+    Parameters
+    ----------
+    a : int
+        The first number to add.
+    b : int
+        The second number to add.
```
ntBre, line 11: Could we add a test case without a type? It looked like the code would handle it correctly, but I didn't see a test for it.

ntBre (commented):
Thanks for your work here and for splitting off the PR! This looks great to me overall, just a few minor ideas/suggestions. Did you look into reusing any code from `D417`? That seemed like the only [comment](https://github.com/astral-sh/ruff/pull/13280#issuecomment-2362997178) from the other PR that wasn't fully resolved.

I'd also like @MichaReiser to have a chance to look since he reviewed #13280 originally.

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-extraneous-parameter_DOC102_google.py.snap
augustelalande: lol the original PR predates multiple spans. Can you suggest an example I can reference?

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
augustelalande: Depends if we care about enforcing indentation. You are right that I'm being inconsistent between raises and parameters. Not sure which is the better approach.

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-extraneous-parameter_DOC102_google.py.snap
ntBre: Sure, I copied this from another comment I wrote so hopefully the  links aren't too outdated:

The easiest way is  with `secondary_annotation` on the guard returned by `report_diagnostic`:

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
augustelalande: I think I prefer enforcing consistent indentation. I would rather update parse_raises in a separate PR

ntBre: That sounds good to me

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-extraneous-parameter_DOC102_google.py.snap
```diff
@@ -33,6 +37,9 @@ DOC102 Documented parameter `numbers` is not in the function's signature
 36 |     Args:
    |     ^^^^
 37 |         numbers (list of int): A list of integers to search through.
+   |         ------- `numbers` is not in the signature
```
ntBre: Oh wow, I was actually suggesting to show the parameters in the function definition, which I thought would be a bit easier to implement. Something like:

```
DOC102 Documented parameter `multiplier` is not in the function's signature
  --> DOC102_google.py:21:5
   |
19 |     Multiplies each element in a list by a given multiplier.
20 |
21 |     Args:
   |     ^^^^
22 |         lst (list of int): A list of integers.
23 |         multiplier (int): The multiplier for each element in the list.
24 |
25 |     Returns:
   |
   | ------------------------------------------------------------
   |
16 | # DOC102
17 | def multiply_list_elements(lst):
   |                           ----- parameters defined here
18 | """
   |
help: Remove the extraneous parameter from the docstring
```

Now that you've implemented this, though, should we make it the main diagnostic range instead of the section heading? I saw some previous discussion about this [here](https://github.com/astral-sh/ruff/pull/13280#discussion_r1752784612).

After mocking up my original/intended suggestion above, I think it might have been a bad idea, but using the ranges you're computing now as the main diagnostic ranges would be great. What do you think? At that point we could also emit one diagnostic per parameter as Micha suggested in the other PR.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -607,25 +647,41 @@ fn parse_parameters_google(content: &str) -> Vec<&str> {
...
-                    entries.push(param.trim_end().trim_start_matches('*'));
+                    let param_name = param.trim_end().trim_start_matches('*');
+                    let param_start = line_start + indentation.len();
+                    let param_end = param_start + param.len();
+
+                    entries.push(ParameterEntry {
+                        name: param_name,
+                        range: TextRange::new(
+                            content_start + TextSize::new(param_start as u32),
+                            content_start + TextSize::new(param_end as u32),
```
ntBre: I think we try to avoid `as` casts since they can silently misbehave. This should be safe since the ranges are from within the document, and the document has to be less than `u32::MAX` in length, but let's unwrap in case that assumption ever breaks down:

```suggestion
                            content_start + TextSize::try_from(param_start as u32).unwrap(),
                            content_start + TextSize::try_from(param_end).unwrap(),
```

Ah this is what clippy is complaining about too.

ntBre (commented):
Thank you! You went above and beyond my diagnostic suggestion. I think we could use that for the main diagnostic now. The code looks good to me otherwise.

However, I took a quick peek at the ecosystem results, and this one in particular looks pretty suspicious:

[disnake/client.py:805:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L805) DOC102 These documented parameters are not in the function's signature: `Example`, `--------`, `.. code-block`

I know we have some open issues around sphinx docstrings, but ideally we'd eliminate these false positives at least.

This one also looks interesting, I wonder if we should consider skipping `*args` as well as `**kwargs`:

[disnake/ext/commands/context.py:310:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/ext/commands/context.py#L310) DOC102 Documented parameter `entity` is not in the function's signature

This one looks like it should have been straightforward, but we have a false positive on `cls`:

[disnake/ext/commands/flag_converter.py:556:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/ext/commands/flag_converter.py#L556) DOC102 Documented parameter `cls` is not in the function's signature

We'll need to go through all of the ecosystem results, but those are a few I pulled out from the `disnake` run.

augustelalande:
I fixed:

[disnake/ext/commands/context.py:310:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/ext/commands/context.py#L310) DOC102 Documented parameter entity is not in the function's signature

and

[disnake/ext/commands/flag_converter.py:556:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/ext/commands/flag_converter.py#L556) DOC102 Documented parameter cls is not in the function's signature

however the problem with

[disnake/client.py:805:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L805) DOC102 These documented parameters are not in the function's signature: Example, --------, .. code-block

is that `Example` is not an official numpy docstring section, where `Examples` is prefered. As defined [here](https://github.com/astral-sh/ruff/blob/main/crates/ruff_linter/src/docstrings/numpy.rs) and [here](https://numpydoc.readthedocs.io/en/latest/format.html#examples)

ntBre:
Thanks! Have you looked through the other ecosystem results? It looks like airflow has some YAML blocks in docstrings that we're trying to parse as parameter names (lots of `-` diagnostics), and there are a couple of splitting issues in bokeh like this:

> [src/bokeh/core/has_props.py:435:13:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/src/bokeh/core/has_props.py#L435) DOC102 Documented parameter `setter(ClientSession` is not in the function's signature

I wonder if we should be a little more permissive with the numpy example(s) section. I don't really like the look of emitting diagnostics like this:

+ [disnake/client.py:812:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L812) DOC102 Documented parameter `Example` is not in the function's signature
+ [disnake/client.py:813:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L813) DOC102 Documented parameter `--------` is not in the function's signature
+ [disnake/client.py:815:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L815) DOC102 Documented parameter `.. code-block` is not in the function's signature

It may help in several of these cases to filter out candidates that aren't valid Python identifiers, if we can't avoid checking them in the first place.

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-extraneous-parameter_DOC102_numpy.py.snap
```diff
@@ -2,214 +2,187 @@
...
+  --> DOC102_numpy.py:8:5
    |
- 4 |     Adds two numbers and returns the result.
- 5 |
  6 |     Parameters
-   |     ^^^^^^^^^^
  7 |     ----------
  8 |     a : int
-   |     -- `a` is not in the signature
+   |     ^^
```
ntBre: These new diagnostics look really good! It looks like we're underlining one additional character in some cases, though. The `a` example above in the Google file looks good, but this one has an extra space.

augustelalande:
> It may help in several of these cases to filter out candidates that aren't valid Python identifiers, if we can't avoid checking them in the first place.

I have limited the detection of parameters to those beginning with valid characters.

>I wonder if we should be a little more permissive with the numpy example(s) section. I don't really like the look of emitting diagnostics like this:
    [disnake/client.py:812:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L812) DOC102 Documented parameter Example is not in the function's signature
    [disnake/client.py:813:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L813) DOC102 Documented parameter -------- is not in the function's signature
    [disnake/client.py:815:9:](https://github.com/DisnakeDev/disnake/blob/7fa65774938b88565c236474973d1ce942081e9e/disnake/client.py#L815) DOC102 Documented parameter .. code-block is not in the function's signature

We can, and it is easy to change, but it seems out of scope for this PR

>[src/bokeh/core/has_props.py:435:13:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/src/bokeh/core/has_props.py#L435) DOC102 Documented parameter setter(ClientSession is not in the function's signature

As far as I'm concerned this one is a problem with their formatting, they should have a space between the parameter and the type.

>lots of - diagnostics

This is also a problem with their docstrings, they aren't using google or numpy style, so we can't parse them properly.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -448,33 +549,150 @@ impl<'a> DocstringSections<'a> {
...
+    let indentation = &first_arg[..first_arg.len() - first_arg.trim_start().len()];
+
+    let mut current_pos = 0;
+    for line in content.lines() {
+        let line_start = current_pos;
+        let line_end = current_pos + line.len();
+
+        if let Some(entry) = line.strip_prefix(indentation) {
+            if entry.chars().next().is_some_and(|first_char| {
+                first_char.is_ascii_alphabetic() || first_char == '_' || first_char == '*'
```
ntBre: I imagine that this would cover the vast majority of cases, but it's not strictly correct. We should try to use something like [`is_identifier`](https://github.com/astral-sh/ruff/blob/caf48f4bfc54bb587342bf794a4f5167d855f529/crates/ruff_python_stdlib/src/identifiers.rs#L7) or the helper functions it uses to cover non-ascii cases and any other nuances.

I'm kind of wondering if a regex might be helpful here. Something like `\*{0,2}([\w_]+) *:` seems like it would work? Then we could apply `is_identifier` to the capture group and reject any non-identifiers. Then we can be a bit more flexible with missing spaces and such.

augustelalande:
> Thanks! I've been thinking about how to handle the false positives. I think I agree with you that these are technically issues in the ecosystem docs, but they'll still look like false positives in Ruff, so I'd like to err on the side of filtering them out.
> 
> I had a couple of ideas about identifier handling in an inline comment that I think could help with some of them.
> 
> I'm somewhat less concerned about the `Example` case and the inline YAML docstrings, but one idea that came to mind is if we can bail out if we're not in a known `Args` or `Parameters` section? That seems like it would filter out at least the YAML cases I looked at, which don't fall under any known section heading as far as I can tell.

If by YAML docstring you are referring to something like this https://github.com/apache/superset/blob/4b71adaa9c89925676e7fed6af0c56c5b0ece863/superset/dashboards/api.py#L1711, the issue is that one of the keys is [`parameters`](https://github.com/apache/superset/blob/4b71adaa9c89925676e7fed6af0c56c5b0ece863/superset/dashboards/api.py#L1705C11-L1705C22) which causes our parser to think it is in a parameters section. We already split the docstring into sections, so we are not processing anything outside of identified sections.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
augustelalande: Ok I should now only be considering valid identifiers. That should get rid of more false positives.

augustelalande:
@ntBre Any further comments?

ntBre:
No, I think you've convinced me. I might do one more pass over the code, but I think we should land this.

There are still a few false positives for unusual docstring formats, but the ecosystem check is also showing quite a few true positives. I think it's a helpful rule.

augustelalande:
@ntBre Sorry to be pushy, but it's been a year since I made the initial pull request, it's been reviewed 3 or 4 times, and every time the conversation just goes silent with no more comments and no merge. I'm fine with making more changes, and fine with it not being merged, but I would really appreciate some transparency about what is holding things up.

Best,

Auguste

ntBre:
Sorry, just a bit behind on my notifications. This is still in my inbox to do another once-over on the code, but I didn't quite get to it today. I'll make this my first review tomorrow.

augustelalande:
@ntBre Thanks for the quick reply, and appreciate everything you guys do

PR #20377, opened 2025-09-14
[ty] Heterogeneous unpacking support for unions

AlexWaygood:

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -62,7 +62,7 @@ use crate::types::infer::infer_unpack_types;
 use crate::types::mro::{Mro, MroError, MroIterator};
 pub(crate) use crate::types::narrow::infer_narrowing_constraint;
 use crate::types::signatures::{Parameter, ParameterForm, Parameters, walk_signature};
-use crate::types::tuple::TupleSpec;
+use crate::types::tuple::{TupleSpec, TupleSpecBuilder};
 pub(crate) use crate::types::typed_dict::{TypedDictParams, TypedDictType, walk_typed_dict_type};
 use crate::types::variance::{TypeVarVariance, VarianceInferable};
 use crate::types::visitor::any_over_type;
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -5277,11 +5277,117 @@ impl<'db> Type<'db> {
...
     ) -> Result<Cow<'db, TupleSpec<'db>>, IterationError<'db>> {
-        // We will not infer precise heterogeneous tuple specs for literals with lengths above this threshold.
-        // The threshold here is somewhat arbitrary and conservative; it could be increased if needed.
-        // However, it's probably very rare to need heterogeneous unpacking inference for long string literals
-        // or bytes literals, and creating long heterogeneous tuple specs has a performance cost.
-        const MAX_TUPLE_LENGTH: usize = 128;
+        fn non_async_special_case<'db>(
+            db: &'db dyn Db,
+            ty: Type<'db>,
+        ) -> Option<Cow<'db, TupleSpec<'db>>> {
```
AlexWaygood, line 5534: other than the `Type::Union` branch, this is exactly the same code that used to exist in the `match` statement slightly lower down. It's just been extracted into a standalone function (so that it can be called recursively), and moved higher up to satisfy Clippy.

dcreager:
We've supported unpacking of unions like this [in assignments](https://github.com/astral-sh/ruff/blob/5e08e5451df62398caf16034ae50621e46688641/crates/ty_python_semantic/resources/mdtest/unpacking.md#L762), but it looks like this implementation might be more general. Can we replace the older union-related code in `types/unpacker.rs` with calls to this new logic?

AlexWaygood:
> We've supported unpacking of unions like this [in assignments](https://github.com/astral-sh/ruff/blob/5e08e5451df62398caf16034ae50621e46688641/crates/ty_python_semantic/resources/mdtest/unpacking.md#L762), but it looks like this implementation might be more general. Can we replace the older union-related code in `types/unpacker.rs` with calls to this new logic?

That's a great question. I've just spent a while looking at this (probably longer than I should have!), and my answer is... I don't think so, unfortunately. Obtaining a different tuple spec for each union member, like `unpacker.rs` is doing, is still preferable to calling `try_iterate()` directly on a union type -- it's still less lossy, yields better fallback results in the case of errors, and gives better error messages.

I experimented with having `Type::try_iterate()` return `Result<IterationOutcome<'db>, IterationError<'db>>`, where `IterationOutcome` is an abstraction that looks like this:

```rs
struct IterationOutcome<'db>(smallvec::SmallVec<[Cow<'db, TupleSpec<'db>>; 1]>);
```

but even that is too lossy a representation for `unpacker.rs`. `unpacker.rs` really needs to know whether each individual union element is iterable, and what kind of diagnostic it emits if it's not. So for `unpacker.rs` to be able to use `try_iterate` without manually mapping over the union elements, `try_iterate` would need to return `IterationOutcome<'db>`, where `IterationOutcome` looks like

```rs
struct IterationOutcome<'db>(
    smallvec::SmallVec<[Result<Cow<'db, TupleSpec<'db>>, IterationError<'db>>; 1]>
);
```

At which point the abstraction becomes so complicated, that I think it's probably not worth it anymore.

So the question is: what does this PR actually get us?

Firstly it gets us much better call-binding for calls with variadic parameters: unlike in `unpacker.rs`, we do not map over the union elements manually in `call/bind.rs`. Maybe we _should_ map over the union elements there, like in `unpacker.rs`, but it looks awkward to pull off: https://github.com/astral-sh/ruff/blob/ac2c5303775bcf52d40eea406939a6961774e6af/crates/ty_python_semantic/src/types/call/bind.rs#L2283-L2291

Secondly, it gets us much more precise inference even if all you want is the _homogeneous_ element type. For example:

```py
from typing import Literal

def f(x: Literal["abc", "def"]):
    for item in x:
        # main: LiteralString
        # This PR: Literal["a", "b", "c", "d", "e", "f"]
        reveal_type(item)
```

crates/ty_python_semantic/resources/mdtest/call/function.md
```diff
@@ -694,6 +694,51 @@ def _(
...
+
+my_args = (("a", "b"), ("c", "d"), ("e", "f"))
+
+for tup in my_args:
+    f4(*tup, e=None)  # fine
+
+my_other_args = (
+    ("a", "b", "c", "d", "e"),
+    ("f", "g", "h", "i", "k"),
+)
```
dcreager: bit of a nit: Using `"a"`, etc as the values in `my_args` and `my_other_args` made me assume at first that you would be exercising `**` unpacking, where it's important that some of the values line up with parameter names. Maybe use integer literals instead of string literals to eliminate that confusion?

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1100,10 +1100,14 @@ impl<'db> Bindings<'db> {
...
-                                    &argument.iterate(db),
-                                )));
+                                let return_type = if let Type::Union(union) = argument {
+                                    union.map(db, |element| {
+                                        Type::tuple(TupleType::new(db, &element.iterate(db)))
+                                    })
+                                } else {
+                                    Type::tuple(TupleType::new(db, &argument.iterate(db)))
+                                };
+                                overload.set_return_type(return_type);
```
dcreager, line 1116: Not blocking for this PR, but this pattern seems to come up enough to deserve a helper method on `Type` — a `map_over_union` that applies the function to each union element if it's a union, or to the individual type if not.

crates/ty_python_semantic/src/types/tuple.rs
```diff
@@ -1552,6 +1552,59 @@ impl<'db> TupleSpecBuilder<'db> {
         }
     }
 
+    fn all_elements(&self) -> impl Iterator<Item = &Type<'db>> {
```
dcreager, line 1586: nit: Can you throw a `copied` onto each iterator so that this can return `Iterator<Item = Type>`? I find it good karma to not make the caller have to worry about that

crates/ty_python_semantic/src/types/tuple.rs
```diff
@@ -1552,6 +1552,59 @@ impl<'db> TupleSpecBuilder<'db> {
...
+                self
+            }
+
+            // We *could* have a branch here where both `self` and `other` are mixed tuples
+            // with same-length prefixes and same-length suffixes. We *could* zip the two
+            // `prefix` vecs together, unioning each pair of elements to create a new `prefix`
+            // vec, and do the same for the `suffix` vecs. This would preserve the tuple specs
+            // of the union elements more closely. But it's hard to think of a test where this
+            // would actually lead to more precise inference, so it's probably not worth the
+            // complexity.
```
dcreager, line 1626: To be fair, such a branch wouldn't be _that_ much longer than the explanation for why it's not present... 😄  But I don't disagree with your argument, and am fine with leaving it out.

AlexWaygood: Yeah, I had it originally, but then couldn't figure out a test for it that actually reflected something that might come up in real-world code, and the virtue of a comment is it doesn't need tests 😆 so I ripped it out

crates/ty_python_semantic/src/types/tuple.rs
AlexWaygood: Hmm, I feel like here it's more ergonomic to have it return `Iterator<Item = &Type<'db>>`, because then it matches the signature of `TupleSpec::all_elements` (and, indeed, the two other `all_elements()` methods in this module!):

https://github.com/astral-sh/ruff/blob/6f468ae8e33bf3ad1d342d29c049fb620c701ca0/crates/ty_python_semantic/src/types/tuple.rs#L343-L345

it seems like strictly more code for not much gain here, if I make it return `Iterator<Item = Type<'db>>` rather than `Iterator<Item = &Type<'db>>`?

```diff
diff --git a/crates/ty_python_semantic/src/types/tuple.rs b/crates/ty_python_semantic/src/types/tuple.rs
index dcb3df675d..cc39e696f8 100644
--- a/crates/ty_python_semantic/src/types/tuple.rs
+++ b/crates/ty_python_semantic/src/types/tuple.rs
@@ -1583,7 +1583,7 @@ impl<'db> TupleSpecBuilder<'db> {
         }
     }
 
-    fn all_elements(&self) -> impl Iterator<Item = &Type<'db>> {
+    fn all_elements(&self) -> impl Iterator<Item = Type<'db>> {
         match self {
             TupleSpecBuilder::Fixed(elements) => Either::Left(elements.iter()),
             TupleSpecBuilder::Variable {
@@ -1592,6 +1592,7 @@ impl<'db> TupleSpecBuilder<'db> {
                 suffix,
             } => Either::Right(prefix.iter().chain(std::iter::once(variable)).chain(suffix)),
         }
+        .copied()
     }
 
     /// Return a new tuple-spec builder that reflects the union of this tuple and another tuple.
@@ -1625,8 +1626,10 @@ impl<'db> TupleSpecBuilder<'db> {
             // would actually lead to more precise inference, so it's probably not worth the
             // complexity.
             _ => {
-                let unioned =
-                    UnionType::from_elements(db, self.all_elements().chain(other.all_elements()));
+                let unioned = UnionType::from_elements(
+                    db,
+                    self.all_elements().chain(other.all_elements().copied()),
+                );
                 TupleSpecBuilder::Variable {
                     prefix: vec![],
                     variable: unioned,
```

PR #20378, opened 2025-09-14
bump mypy_primer pin

AlexWaygood:
Pulls in https://github.com/hauntsaninja/mypy_primer/commit/06849fda409cbdc3233c8125a2cf9881bb9447a6


sharkdp:
We should also remove `materialize` from `good.txt`. mypy_primer doesn't care if there are projects in the filter that are not available, but ecosystem-analyzer complains. I'll fix that.

PR #20379, opened 2025-09-14
[ty] Add LSP debug information command

Guillaume-Fgt:
## Summary

I implemented a Debug command for the ty_server in response to issue https://github.com/astral-sh/ty/issues/974.

I took inspiration from the Ruff server mechanism. The command can be executed with  `ty.printDebugInformation`. It accepts three different arguments: "short", "full" and "mypy_primer".

## Test Plan
I wrote tests for the command with the three different possible arguments. I also used it with Nvim with this configuration in my lua:

```
--command to trigger ty debug command for information about the session.
vim.api.nvim_create_autocmd('FileType', {
  pattern = 'python',
  callback = function()
    vim.keymap.set('n', '<leader>td', function()
      local clients = vim.lsp.get_active_clients({ bufnr = 0 })
      for _, client in ipairs(clients) do
        if client.name == "ty" then
          client.request("workspace/executeCommand", {
            command = "ty.printDebugInformation",
            arguments = {"short"}
          }, function(err, result)
            if err then
              vim.notify("Error: " .. err.message, vim.log.levels.ERROR)
            else
              vim.notify("Debug info: " .. vim.inspect(result), vim.log.levels.INFO)
            end
          end)
          break
        end
      end
    end, { buffer = true, desc = "Print Ty Debug Information" })
  end,
})
```

This is the result in action:

https://github.com/user-attachments/assets/07d9a070-86c8-4723-84d8-f082d966c99a

### DIFF
--- a/crates/ty_server/src/server/api/requests/execute_command.rs
+++ b/crates/ty_server/src/server/api/requests/execute_command.rs
@@ -54,18 +54,18 @@ fn debug_information(session: &Session) -> crate::Result<String> {
         "Open text documents: {}",
         session.text_document_keys().count()
     )?;
-    writeln!(buffer, "")?;
+    writeln!(buffer)?;
 
     for (root, workspace) in session.workspaces() {
         writeln!(buffer, "Workspace {root} ({})", workspace.url())?;
         writeln!(buffer, "Settings: {:#?}", workspace.settings())?;
-        writeln!(buffer, "")?;
+        writeln!(buffer)?;
     }
 
     for db in session.project_dbs() {
         writeln!(buffer, "Project at {}", db.project().root(db))?;
         writeln!(buffer, "Settings: {:#?}", db.project().settings(db))?;
-        writeln!(buffer, "")?;
+        writeln!(buffer)?;
         writeln!(
             buffer,
             "Memory report:\n{}",


MichaReiser:
I plan to look into this but I'm not sure if I get to it before the end of the week (I'm still catching up after my PTO)

crates/ty_server/src/server/api/requests/execute_command.rs
```diff
@@ -0,0 +1,65 @@
...
+                writeln!(buffer, "{db_str}")?;
+            }
+            "mypy_primer" => {
+                let db_str = db.salsa_memory_dump().display_mypy_primer().to_string();
+                writeln!(buffer, "{db_str}")?;
+            }
+            "full" => {
+                let db_str = db.salsa_memory_dump().display_full().to_string();
+                writeln!(buffer, "{db_str}")?;
+            }
```
MichaReiser: I think it's fine if this command only supports the `full` output format. It's a debug command and too much output isn't really an issue.

crates/ty_server/src/server/api/requests/execute_command.rs
```diff
@@ -0,0 +1,65 @@
...
+        Ok(None)
+    }
+}
+
+/// Returns a string with detailed memory usage.
+fn debug_information(session: &Session, report_type: &str) -> crate::Result<String> {
+    let mut buffer = String::new();
+
+    writeln!(buffer, "report type: {report_type}")?;
+    let db = session.project_dbs().next();
```
MichaReiser: It would be great to show the memory usage for all databases rather than just the first. I think there's also some additional information that's useful to show, like the resolved settings. 

This is untested, but maybe something like this

```rust
    let mut buffer = String::new();

    writeln!(buffer, "Client capabilities: {:#?}", session.client_capabilities());
    writeln!(buffer, "Position encoding: {:#?}", session.position_encoding());
    writeln!(buffer, "Global settings: {:#?}", session.global_settings());
    writeln!(buffer, "Open text documents: {}", session.text_document_keys().count());

    for (root, workspace) in session.workspaces() {
        writeln!(buffer, "Workspace {root} ({})", workspace.url())?;
        writeln!(buffer, "  Settings:\n\n{:#?}", workspace.settings())?;
        writeln!(buffer, "\n")?;
    }

    let mut first_db = true;

    for db in session.project_dbs() {
        writeln!(buffer, "\n")?;

        first_db = false;

        writeln!(buffer, "Project at {}", db.project().root(db))?;
        writeln!(buffer, "  Settings:\n{:#?}", db.project().settings(db))?;
        writeln!(buffer, "  Memory report:\n{}", db.salsa_memory_dump().display_full())?;
    }
```

crates/ty_server/tests/e2e/commands.rs
```diff
@@ -0,0 +1,102 @@
...
+        .build()?
+        .wait_until_workspaces_are_initialized()?;
+
+    let response = execute_command(
+        &mut server,
+        "ty.printDebugInformation".to_string(),
+        vec![json!("full")],
+    )?;
+
+    insta::assert_debug_snapshot!(response);
```
MichaReiser: This is great that you added tests. 

I do think we'll need to redact the snapshot to prevent the test from becoming noisy. 

The easiest could be to simply omit the memory usage if `session.in_test` is true (you might need to expose the field with a `pub(crate)` method.

The alternative (and somewhat nicer) approach is to use insta's filter feature. We could add a filter that replaces all `[0-9]+.[0-9]+MB` with (`<X.XXMB>`)

MichaReiser:
This is great. Thank you!

Guillaume-Fgt:
Hi Micha, I was planning to make detailed explanations of my code modifications but my commits automatically triggered review apparently, I wasn't expected that. Thanks you for your comments on the code, it helped me a lot as I am still discovering the repo.

> Are you also interested in contributing the necessary VS code changes:
> 
>     1. Adding the command in the `package.json` ([source](https://github.com/astral-sh/ruff-vscode/blob/67f5c516ae2175a32a468aaaa05a614344230ad0/package.json#L447-L451))
> 
>     2. Add a handler that automatically opens a window in VS code ([source](https://github.com/astral-sh/ruff-vscode/blob/67f5c516ae2175a32a468aaaa05a614344230ad0/src/common/commands.ts#L1-L124))

I am less familiar with VS code but I am interested. I will work on it and open a PR or an issue if I need help 👍.

PR #20327, opened 2025-09-10
[`pyupgrade`] Prevent infinite loop with `I002` and `UP026`

TaKO8Ki:
## Summary

Fixes #19842

Prevent infinite loop with I002 and UP026

- Implement isort-aware handling for UP026 (deprecated mock import):
- Add CLI integration tests in crates/ruff/tests/lint.rs:

## Test Plan

I have added two integration tests `pyupgrade_up026_respects_isort_required_import_fix` and `pyupgrade_up026_respects_isort_required_import_from_fix` in `crates/ruff/tests/lint.rs`.

PR #19413, opened 2025-07-17
[`pyupgrade`] Prevent infinite loop with `I002` (`UP010`, `UP035`)

danparizher:
## Summary

Fixes #18729 and fixes #16802

## Test Plan

Manually verified via CLI that Ruff no longer enters an infinite loop by running:
```sh
echo 1 | ruff --isolated check - --select I002,UP010 --fix
```
with `required-imports = ["from __future__ import generator_stop"]` set in the config, confirming “All checks passed!” and no snapshots were generated.

PR #20381, opened 2025-09-15
[`isort`] Fix infinite loop when checking equivalent imports (`I002`, `PLR0402`)

### LINKED ISSUE #20380: [Infinite loop] I002 and PLR0402
### Summary

Ruff fails to converge when [`missing-required-import` (I002)](https://docs.astral.sh/ruff/rules/missing-required-import/) and [`manual-from-import` (PLR0402)](https://docs.astral.sh/ruff/rules/manual-from-import/) are selected and a redundant alias is required. [Example](https://play.ruff.rs/788c22f4-b157-4e9d-aef4-29528b22dd71):
```console
$ echo 1 | ruff --isolated check - --select I002,PLR0402 --config 'lint.isort.required-imports=["import concurrent.futures as futures"]' --diff

...quoting the contents of `-`, the rule codes I002, along with the `pyproject.toml` settings and executed command, we'd be very appreciative!

@@ -1 +1,51 @@
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
+from concurrent import futures
 1

Would fix 100 errors.
```

### Version

danparizher:
## Summary

Fixes #20380

The fix changes the `includes_import` function in `I002` to recognize equivalent imports based on bound name and qualified name rather than exact syntax matching.

This allows the function to correctly identify that `from concurrent import futures` satisfies the requirement for `import concurrent.futures as futures`.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/isort/required_imports/equivalent_imports.py
+++ b/crates/ruff_linter/resources/test/fixtures/isort/required_imports/equivalent_imports.py
@@ -1,0 +1,3 @@
+from concurrent import futures
+
+1
--- a/crates/ruff_linter/src/rules/isort/mod.rs
+++ b/crates/ruff_linter/src/rules/isort/mod.rs
@@ -991,6 +991,29 @@ mod tests {
         Ok(())
     }
 
+    #[test_case(Path::new("equivalent_imports.py"))]
+    fn required_import_equivalent_imports(path: &Path) -> Result<()> {
+        let snapshot = format!("required_import_equivalent_imports_{}", path.to_string_lossy());
+        let diagnostics = test_path(
+            Path::new("isort/required_imports").join(path).as_path(),
+            &LinterSettings {
+                src: vec![test_resource_path("fixtures/isort")],
+                isort: super::settings::Settings {
+                    required_imports: BTreeSet::from_iter([NameImport::Import(
+                        ModuleNameImport::alias(
+                            "concurrent.futures".to_string(),
+                            "futures".to_string(),
+                        ),
+                    )]),
+                    ..super::settings::Settings::default()
+                },
+                ..LinterSettings::for_rules([Rule::MissingRequiredImport, Rule::ManualFromImport])
+            },
+        )?;
+        assert_diagnostics!(snapshot, diagnostics);
+        Ok(())
+    }
+
     #[test_case(Path::new("from_first.py"))]
     fn from_first(path: &Path) -> Result<()> {
         let snapshot = format!("from_first_{}", path.to_string_lossy());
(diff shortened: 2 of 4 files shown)


dylwil3:
Hmmm... I'm a bit conflicted here about whether `I002` or `PLR0402` should take precedence to resolve the infinite loop. Generally it feels like the more "customized" or "manually entered" choice should take precedence. Since a user enters required imports exactly as they'd like them to appear, whereas `PLR0402` is a "general" rule, I'm tempted to say that it's `PLR0402` that should be modified here.

Would it be possible to have `PLR0402` skip if it's triggering on a required import instead? I would also be open to adding either a warning or a debug message explaining why the rule is being skipped in this case. (Probably a debug message since we wouldn't have a good way to override the warning).

What do you think?

ntBre:
I agree with Dylan, that seems closer to how we've handled the other `I002` conflicts recently (https://github.com/astral-sh/ruff/pull/20327, https://github.com/astral-sh/ruff/pull/19413). Possibly we can even reuse the helper method you added in #19413 :)

crates/ruff_linter/src/rules/pyupgrade/rules/unnecessary_future_import.rs
```diff
@@ -95,20 +95,32 @@ pub(crate) fn is_import_required_by_isort(
...
+            required_imports
+                .iter()
+                .any(|required_import| required_import.qualified_name().segments() == segments)
+        }
+        StmtRef::Import(ast::StmtImport { .. }) => {
+            // For import statements, we need to check if the full module path matches
+            // For example, "concurrent.futures" should match ["concurrent", "futures"]
+            let segments: Vec<&str> = alias.name.as_str().split('.').collect();
+            required_imports.iter().any(|required_import| {
+                required_import.qualified_name().segments() == segments.as_slice()
```
dylwil3: Could you compare qualified names directly and use this method? https://github.com/astral-sh/ruff/blob/0c7cfd2a8d60e9e09cfd75c9d521f71a8be0b592/crates/ruff_python_ast/src/name.rs#L247

I believe it's designed to not allocate on the heap when there's a small number of segments.

PR #20391, opened 2025-09-15
[`ruff`] `FURB164` Replace -nan with nan when using the value to construct `Decimal`

### LINKED ISSUE #19699: FURB164 fix changes behavior for negative NaN
### Summary

When the argument to `Decimal.from_float` is negative NaN, the fix for [`unnecessary-from-float` (FURB164)](https://docs.astral.sh/ruff/rules/unnecessary-from-float/) should replace it with positive NaN to preserve the program’s behavior. [Example](https://play.ruff.rs/9ac93a35-6395-4766-be06-26af9938879d):
```console
$ cat >furb164_2.py <<'# EOF'
from decimal import Decimal
print(Decimal.from_float(float(" -Nan\n")))
# EOF

$ python furb164_2.py
NaN

$ ruff --isolated check furb164_2.py --select FURB164 --preview --fix

$ cat furb164_2.py
from decimal import Decimal
print(Decimal("-nan"))
# EOF

$ python furb164_2.py
-NaN
```
This was originally reported in #19460 and marked as fixed by #19468, but that PR added normalization for non-finite float strings, which is unnecessary and doesn’t fix the problem. [`verbose-decimal-constructor` (FURB157)](https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/) handles this case correctly and with minimal normalization.

### Version

ruff 0.12.7 (c5ac99889 2025-07-29)

TaKO8Ki:
## Summary

Fixes #19699

Normalize `Decimal(float("-nan"))` to `Decimal("nan")`.

The same handling is implemented in https://github.com/TaKO8Ki/ruff/blob/c3e873dd827b9ec8dd49cff723089852ce2301e1/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs#L165.

## Test Plan

I've updated `ruff_linter__rules__refurb__tests__FURB164_FURB164.py.snap`.

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/unnecessary_from_float.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/unnecessary_from_float.rs
@@ -293,7 +293,19 @@ fn handle_non_finite_float_special_case(
         return None;
     };
     let normalized = as_non_finite_float_string_literal(float_arg)?;
-    let replacement_text = format!(r#"{constructor_name}("{normalized}")"#);
+    let replacement_text = format!(
+        r#"{constructor_name}("{}")"#,
+        // `Decimal.from_float(float(" -nan")) == Decimal("nan")`
+        if normalized == "-nan" {
+            // Here we do not attempt to remove just the '-' character.
+            // It may have been encoded (e.g. as '\N{hyphen-minus}')
+            // in the original source slice, and the added complexity
+            // does not make sense for this edge case.
+            "nan"
+        } else {
+            normalized
+        }
+    );
     Some(Edit::range_replacement(replacement_text, call.range()))
 }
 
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/refurb/rules/unnecessary_from_float.rs
```diff
@@ -293,7 +293,19 @@ fn handle_non_finite_float_special_case(
         return None;
     };
     let normalized = as_non_finite_float_string_literal(float_arg)?;
-    let replacement_text = format!(r#"{constructor_name}("{normalized}")"#);
+    let replacement_text = format!(
+        r#"{constructor_name}("{}")"#,
+        // `Decimal.from_float(float(" -nan")) == Decimal("nan")`
+        if normalized == "-nan" {
```
TaKO8Ki, line 299: The same handling is implemented in: https://github.com/astral-sh/ruff/blob/c3e873dd827b9ec8dd49cff723089852ce2301e1/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs#L165

Is it better to embed this handling into `as_non_finite_float_string_literal`? Since Apart from unnecessary_from_float, the function is used only in the following place, it does not affect other parts, but I feel it's too specific.
https://github.com/astral-sh/ruff/blob/c3e873dd827b9ec8dd49cff723089852ce2301e1/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs#L159

Issue #15555, opened 2025-01-17
Autofix for flake8-logging-format (G) rules


justinchuby:
Hi! I was wondering if there are any plans to create autofix for (G) rules? We have a lot of violations that prevents enabling the rule. Having autofix capabilities (even if unsafe) would greatly help with adoption of the (G) rules in our codebase.

MichaReiser:
Are you looking for a specific rule or is it really all rules? The format fixes tend to be rather difficult to implement because of all the edge cases and adding more fixes isn't a priority for us right now (not saying that it wouldn't be great to have them, it's just there are so many other improvements)

justinchuby:
The current rule we need the most in microsoft/onnxruntime is G004. Because everything else can be turned into an f-string first.

ktbarrett:
I will second really only needing an autofix for G004.

G001 looks trivial as both logging methods and `str.format` use `string.Formatter`.
G002 could involve transforming to `str.format` first, then trivially converting to the logging call.

bmos:
If we're just talking about G004, it seems like it should be pretty easy to copy the fix for EM102 with small changes. Maybe I'm wrong, but just a suggestion.

ntBre:
> If we're just talking about G004, it seems like it should be pretty easy to copy the fix for EM102 with small changes.

I thought the same thing when I looked into this a few weeks ago, but the point of G004 is not simply to move the f-string out of the `logging` call, it's to avoid formatting the string at all when the log level is set high enough. See this part of the rule docs in particular:

> Using f-strings to format a logging message requires that Python eagerly format the string, even if the logging statement is never executed (e.g., if the log level is above the level of the logging statement), whereas using the extra keyword argument defers formatting until required.

Moving the f-string out of the `logging` call like in EM102 just eagerly formats the string a little earlier.

Properly fixing G004 requires inspecting the contents of the f-string and converting them to the appropriate `logging` method call with printf-style arguments and placeholders, at least as far as I can tell.

Now, if you simply want to defeat the rule, moving the f-string out of the `logging` call *does* help because the rule doesn't do any type inference on its argument. It only checks for literal f-strings and other formatting calls. But at that point you might as well `noqa` or ignore the rule entirely.

ktbarrett:
So what I'm hearing is that this is a valid fix?

```
if self.log.isEnabledFor(logging.DEBUG):
    self.log.debug(f"{name=}")
```

/s

justinchuby:
[...]

PR #11019, opened 2024-04-19
[UP031] When encountering `"%s" % var` offer unsafe fix

plredmond:
Resolves #10187

<details>
<summary>Old PR description; accurate through commit e86dd7d; probably best to leave this fold closed</summary>

## Description of change

In the case of a printf-style format string with only one %-placeholder and a variable at right (e.g.  `"%s" % var`):

* The new behavior attempts to dereference the variable and then match on the bound expression to distinguish between a 1-tuple (fix), n-tuple (bug :bug:), or a non-tuple (fix). Dereferencing is via `analyze::typing::find_binding_value`.
* If the variable cannot be dereferenced, then the type-analysis routine is called to distinguish only tuple (no-fix) or non-tuple (fix). Type analysis is via `analyze::typing::is_tuple`.
* If any of the above fails, the rule still fires, but no fix is offered.

## Alternatives

* If the reviewers think that singling out the 1-tuple case is too complicated, I will remove that.
* The ecosystem results show that no new fixes are detected. So I could probably delete all the variable dereferencing code and code that tries to generate fixes, tbh.

## Changes to existing behavior

**All the previous rule-firings and fixes are unchanged except for** the "false negatives" in `crates/ruff_linter/resources/test/fixtures/pyupgrade/UP031_1.py`. Those previous "false negatives" are now true positives and so I moved them to `crates/ruff_linter/resources/test/fixtures/pyupgrade/UP031_0.py`.

<details>
<summary>Existing false negatives that are now true positives</summary>

```
crates/ruff_linter/resources/test/fixtures/pyupgrade/UP031_0.py:134:1: UP031 Use format specifiers instead of percent format
    |
133 | # UP031 (no longer false negatives)
134 | 'Hello %s' % bar
    | ^^^^^^^^^^^^^^^^ UP031
135 |
136 | 'Hello %s' % bar.baz
    |
    = help: Replace with format specifiers

crates/ruff_linter/resources/test/fixtures/pyupgrade/UP031_0.py:136:1: UP031 Use format specifiers instead of percent format
    |
134 | 'Hello %s' % bar
135 |
136 | 'Hello %s' % bar.baz
    | ^^^^^^^^^^^^^^^^^^^^ UP031
137 |
138 | 'Hello %s' % bar['bop']
    |
    = help: Replace with format specifiers

crates/ruff_linter/resources/test/fixtures/pyupgrade/UP031_0.py:138:1: UP031 Use format specifiers instead of percent format
    |
[...]

PR #20416, opened 2025-09-15
[`pyupgrade`] Mark `printf-string-formatting` fixes as safe (`UP031`)

auscompgeek:
## Summary

This marks the subset of fixes emitted by UP031, where the RHS operand is unambiguous (see https://docs.astral.sh/ruff/rules/printf-string-formatting/#fix-safety), as safe.

## Test Plan

`cargo test`, and inspect snapshot diff to see the new safe fixes

### DIFF
--- a/crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs
+++ b/crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::fmt::Write;
 use std::str::FromStr;
 
+use ruff_diagnostics::Applicability;
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::{self as ast, AnyStringFlags, Expr, StringFlags, whitespace::indentation};
 use ruff_python_codegen::Stylist;
@@ -416,22 +417,31 @@ pub(crate) fn printf_string_formatting(
     }
 
     // Parse the parameters.
-    let params_string = match right {
+    let (applicability, params_string) = match right {
         Expr::StringLiteral(_)
         | Expr::BytesLiteral(_)
         | Expr::NumberLiteral(_)
         | Expr::BooleanLiteral(_)
         | Expr::NoneLiteral(_)
         | Expr::EllipsisLiteral(_)
-        | Expr::FString(_) => Cow::Owned(format!("({})", checker.locator().slice(right))),
+        | Expr::FString(_) => (
+            Applicability::Safe,
+            Cow::Owned(format!("({})", checker.locator().slice(right))),
+        ),
         Expr::Name(_) | Expr::Attribute(_) | Expr::Subscript(_) | Expr::Call(_) => {
             if num_keyword_arguments > 0 {
                 // If we have _any_ named fields, assume the right-hand side is a mapping.
-                Cow::Owned(format!("(**{})", checker.locator().slice(right)))
+                (
+                    Applicability::Safe,
+                    Cow::Owned(format!("(**{})", checker.locator().slice(right))),
+                )
             } else if num_positional_arguments > 1 {
                 // If we have multiple fields, but no named fields, assume the right-hand side is a
                 // tuple.
-                Cow::Owned(format!("(*{})", checker.locator().slice(right)))
+                (
+                    Applicability::Safe,
+                    Cow::Owned(format!("(*{})", checker.locator().slice(right))),
+                )
             } else {
                 // Otherwise, if we have a single field, we can't make any assumptions about the
                 // right-hand side. It _could_ be a tuple, but it could also be a single value,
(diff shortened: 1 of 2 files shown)


dylwil3:
Thanks for working on this! Would you mind gating this change behind preview? You'll want to add a function to https://github.com/astral-sh/ruff/blob/e8b445012507762e794f6a5f6247856e093e53e9/crates/ruff_linter/src/preview.rs with a comment linking to this PR and then call that function in the module implementing the rule

ntBre:
Thanks for working on this, but I'm not sure we want to make this change. The fix has always been unsafe, even before we started handling variables (https://github.com/astral-sh/ruff/pull/11019), so I think there could be a good reason. For example, I think [this comment](https://github.com/astral-sh/ruff/issues/15555#issuecomment-3282365625) is relevant here too:

> `%s` is not equivalent to passing `None` to `__format__` except by convention.

As an example of this:

```pycon
>>> class C:
...     def __format__(*args, **kwargs):
...         print("called my format!")
...         return ""
...
>>> "{}".format(C())
called my format!
''
>>> "%s" % C()
'<__main__.C object at 0x7fd0f5757250>'
```

So I think this fix should remain unsafe, even if this is pretty reliable in practice.

auscompgeek:
> As an example of this:

Ah, that's a fun example. I think there shouldn't be any examples of this being unsafe for a literal RHS though, right?

ntBre:
Hmm, yeah it's probably safe for literals. But it also seems relatively uncommon to pass a literal to a formatting call instead of using a literal string. I'd probably just lean toward leaving the whole thing unsafe for the sake of simplicity, but I'm open to other thoughts. Making the fix safe for literals in preview, as Dylan said, sounds technically okay.

auscompgeek:
I agree that formatting only a single literal on the RHS would be quite uncommon. I can see it being useful with multiple values though.

I've also added a comment about `__format__` to the fix safety docs for the rule, in case that question comes up again :smile:

crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs
```diff
@@ -71,6 +75,19 @@ use crate::{Edit, Fix, FixAvailability, Violation};
 /// print("{}".format(val))  # "(1,)"
 /// ```
 ///
+/// It is also possible for the right-hand side to have values that implement
+/// `__format__` to not behave the same as `__str__` when `None` is passed:
```
ntBre, line 79: Not sure exactly the best phrasing here, maybe something like this?

```suggestion
/// Additionally, while `__format__` typically calls `__str__`, this is not always
/// the case. If `__format__` has been overridden, replacing `%s` with the
/// empty `format` placeholder will change the program's behavior:
```

I also think we should mention explicitly when the fix _is_ safe. To me that feels most natural right at the beginning of the section ("The fix is only marked as safe when the value being formatted is a literal"). We should also try to work `In preview, ...` in there somewhere.

crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs
```diff
@@ -444,17 +473,36 @@ pub(crate) fn printf_string_formatting(
...
                 Cow::Owned(format!("({})", checker.locator().slice(right)))
-            }
-        }
-        Expr::Tuple(_) => clean_params_tuple(right, checker.locator()),
-        Expr::Dict(_) => {
+            },
+        ),
+        Expr::Tuple(tuple) => (
+            if is_literal_safe_fix_printf_string_formatting_enabled(checker.settings())
+                && tuple.iter().all(is_constant)
```
ntBre, line 480: It looks like `is_constant` already works for tuples:

https://github.com/astral-sh/ruff/blob/d8b4b38d0bae57e72bf7c59eef68e40c5ffafa76/crates/ruff_python_ast/src/helpers.rs#L661-L663

crates/ruff_linter/src/rules/pyupgrade/rules/printf_string_formatting.rs
```diff
@@ -444,17 +473,36 @@ pub(crate) fn printf_string_formatting(
...
+                clean_params_dictionary(dict, checker.locator(), checker.stylist())
             else {
                 checker.report_diagnostic(PrintfStringFormatting, string_expr.range());
                 return;
             };
-            Cow::Owned(params_string)
+            let applicability =
+                if is_literal_safe_fix_printf_string_formatting_enabled(checker.settings())
+                    && dict.iter().all(|item| {
+                        item.key.as_ref().is_some_and(is_constant) && is_constant(&item.value)
```
ntBre, line 498: I think we've already checked the keys in `clean_params_dictionary`. We're transforming a dict literal into keyword arguments like `key=value`, and we bail out without a `params_string` if the key isn't a string literal. But we do need to check the values with `is_constant`; it doesn't handle `Expr::Dict`s automatically like tuples.

MichaReiser:
@auscompgeek do you think you'll have time in the near future to address the PR feedback. If not, that's fine. I think we could follow up with the feedback ourselves to get this landed.

PR #20418, opened 2025-09-15
Avoid syntax error when formatting attribute expressions with outer parentheses, parenthesized value, and trailing comment on value

### LINKED ISSUE #19350: Formatter: Parenthesized expression with comment on right hand side in assignment produces invalid syntax
> ```python
> variable = (
>     (something)  # a comment
>     .fist_method("some string")
> )
> ```
> gets reformatted to
> ```python
> variable = (something)  # a comment
> .fist_method("some string")
> ```
> which is a syntax error, so there's a bug in ruff here.
>  

 _Originally posted by @olofahlen in [#8598](https://github.com/astral-sh/ruff/issues/8598#issuecomment-3072991517)_

dylwil3:
Closes #19350 

I'm not sure if this is the expected/desired output. Without parentheses around the value, we have:

```python
# unformatted
variable = (
    something # a comment
    .first_method("some string")
)

# formatted
variable = something.first_method("some string")  # a comment
```

It's a little unclear to me... it almost feels like in _both_ cases (when `something` does or does not have parentheses) we should maintain the break, since that's what I would expect from a cursory reading of https://docs.astral.sh/ruff/formatter/black/#trailing-end-of-line-comments

### DIFF
--- a/crates/ruff_python_formatter/src/expression/expr_attribute.rs
+++ b/crates/ruff_python_formatter/src/expression/expr_attribute.rs
@@ -179,6 +179,12 @@ impl NeedsParentheses for ExprAttribute {
             context.comments().ranges(),
             context.source(),
         ) {
+            // We have to avoid creating syntax errors like
+            // ```python
+            // variable = (something) # trailing
+            // .my_attribute
+            // ```
+            // See https://github.com/astral-sh/ruff/issues/19350
             if context
                 .comments()
                 .trailing(self.value.as_ref())


MichaReiser:
Thank you. 

The comment placement is a bit weird if `something` isn't parenthesized, but the comment is very long:

```python
variable = (
    (something) # a commentdddddddddddddddddddddddddddddd
    .first_method("some string")
)


variable = (
    something # a commentdddddddddddddddddddddddddddddd
    .first_method("some string")
)
```

Formatted

```
variable = (
    (something)  # a commentdddddddddddddddddddddddddddddd
    .first_method("some string")
)


variable = something.first_method(  # a commentdddddddddddddddddddddddddddddd
    "some string"
)
```

We might want to spent a little time to see if we can preserve the comment placement closer to where it used to be (by e.g. inserting a hard line break after the attribute name or using a line suffix boundary). But this might be harder than it looks and is something we could fix separately.

The fix otherwise looks good, e.g. it also fixes this code that previously produced invalid syntax (may be worth adding as a test to showcase that the issue isn't specific to assignment statements):

```python
if (
    (something)  # a commentdddddddddddddddddddddddddddddd
    .first_method("some string")
): pass
```

crates/ruff_python_formatter/src/expression/expr_attribute.rs
```diff
@@ -179,7 +179,11 @@ impl NeedsParentheses for ExprAttribute {
             context.comments().ranges(),
             context.source(),
         ) {
-            OptionalParentheses::Never
+            if context.comments().has_trailing(self.value.as_ref()) {
```
MichaReiser: We'll now also use the new layout if `value` is parenthesized and has a trailing comment like this

```python
variable = (
    (something # a comment  
    ).fist_method("some string")
)
```

I think that's fine but it might be worth adding a few more tests.

For example, this snippet now gets formatted differently:


```python
if (
    (something # a comment  
    ).fist_method("some string") # second comment
): pass
```


Overall, maybe add a few more permutations with multiple comments (or at least play with it in the playground)

dylwil3:
I've added some more tests, and I'm personally okay with the edge cases that produce slightly awkward formatting (in the name of getting rid of this syntax error). But let me know if you want me to try to tweak anything as part of this PR.

crates/ruff_python_formatter/src/expression/expr_attribute.rs
MichaReiser: Do we need to limit this to end of line comments only? Can you add a test where a value has both (unless that's not possible)

MichaReiser (approved):

crates/ruff_python_formatter/src/expression/expr_attribute.rs
dylwil3: It's possible to have both (and we get a syntax error on main after formatting, but not in this PR). But you're right that we only need to use `Multiline` here if there's a trailing end of line comment, so I've narrowed the condition.

Added a test with both and a test just with trailing own-line comment (the latter formats with the same IR as on `main`, and without a syntax error, now that we narrowed the condition).

PR #20419, opened 2025-09-15
Revert "Update CodSpeedHQ/action action to v4"

AlexWaygood:
Reverts astral-sh/ruff#20409

Prior to https://github.com/astral-sh/ruff/commit/25c13ea91c7535f14bda2d23595fa3ba311752fe, the walltime benchmarks CI job took around 10-11 minutes. It now seems to be taking around 17-18 minutes to complete, which is a pretty major slowdown and somewhat annoying if you're waiting for benchmark results on a PR! See e.g. https://github.com/astral-sh/ruff/actions/runs/17738649543/job/50411336184

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -933,9 +933,8 @@ jobs:
         run: cargo codspeed build --features "codspeed,instrumented" --no-default-features -p ruff_benchmark
 
       - name: "Run benchmarks"
-        uses: CodSpeedHQ/action@653fdc30e6c40ffd9739e40c8a0576f4f4523ca1 # v4.0.1
+        uses: CodSpeedHQ/action@76578c2a7ddd928664caa737f0e962e3085d4e7c # v3.8.1
         with:
-          mode: instrumentation
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
 
@@ -967,8 +966,7 @@ jobs:
         run: cargo codspeed build --features "codspeed,walltime" --no-default-features -p ruff_benchmark
 
       - name: "Run benchmarks"
-        uses: CodSpeedHQ/action@653fdc30e6c40ffd9739e40c8a0576f4f4523ca1 # v4.0.1
+        uses: CodSpeedHQ/action@76578c2a7ddd928664caa737f0e962e3085d4e7c # v3.8.1
         with:
-          mode: walltime
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}


silamon:
Maybe try running with environment variable `CODSPEED_PERF_ENABLED=false`, since it has been enabled by default, while it was not before.

.github/workflows/ci.yaml
```diff
@@ -933,9 +933,8 @@ jobs:
         run: cargo codspeed build --features "codspeed,instrumented" --no-default-features -p ruff_benchmark
 
       - name: "Run benchmarks"
-        uses: CodSpeedHQ/action@653fdc30e6c40ffd9739e40c8a0576f4f4523ca1 # v4.0.1
+        uses: CodSpeedHQ/action@76578c2a7ddd928664caa737f0e962e3085d4e7c # v3.8.1
         with:
-          mode: instrumentation
```
MichaReiser: Can we keep the mode lines, it makes upgrading easier (a rennovate PR rather than having to do the upgrade manually)

MichaReiser:
I pinged the codspeed folks. I'd suggest to maybe give them a few hours to fix this on their side.

MichaReiser (approved):

MichaReiser:
I think @silamon is on to something. Codspeed now enables perf profiles for wallspeed benchmarks (you now see a flamegraph for walltime benchmarks too). I can't say just yet if they'll be useful or if our flamegraphs are too large but it's certainly nice to have more than just: It got slower or faster

AlexWaygood:
https://codspeed.io/astral-sh/ruff/branches/alex%2Fsubtyping-between-protocols?runnerMode=WallTime has some flamegraphs for https://github.com/astral-sh/ruff/pull/20368#issuecomment-3292968246 but I don't find them useful personally. The regression in CI time is very significant, however; I'm finding it very frustrating waiting for CI to finish on #20368!

> We should try `CODSPEED_PERF_ENABLED=false`

I've pushed a commit trying this out

AlexWaygood:
Nice, we're back to the walltime benchmarks completing in 10-11 minutes. Though there are now a lot of lines like this in [the logs](https://github.com/astral-sh/ruff/actions/runs/17759661197/job/50469322114?pr=20419) that I don't know the meaning of:

```
  Failed to send benchmark URI to runner: FIFO does not exist: /tmp/runner.ctl.fifo
```

MichaReiser (approved):
Thank you. I'm fine disabling this for now. It might be worth adding a comment explaining why we disable the perf uploads (because they would be nice)

.github/workflows/ci.yaml
```diff
@@ -968,6 +968,8 @@ jobs:
 
       - name: "Run benchmarks"
         uses: CodSpeedHQ/action@653fdc30e6c40ffd9739e40c8a0576f4f4523ca1 # v4.0.1
+        env:
+          CODSPEED_PERF_ENABLED: false
```
AlexWaygood, line 975:
```suggestion
          # enabling flamegraphs adds ~5 minutes to the CI time,
          # and they don't appear to provide much useful data for our benchmarks right now
          # (see https://github.com/astral-sh/ruff/pull/20419)
          CODSPEED_PERF_ENABLED: false
```

PR #20420, opened 2025-09-15
Simplify `Program` configuration for GitHub output

ntBre:
## Summary

This is a follow-up to https://github.com/astral-sh/ruff/pull/20358 to avoid adding a new `DisplayDiagnosticConfig` option (with a surprising default value) just to print the program name for GitHub diagnostics.

Initially I expected this to be a larger refactor including making the various `Renderer::render` methods take a `std::io::Write` instead of a `std::fmt::Formatter` to make them easier to call directly in Ruff. However, this was a bit more painful than expected because we use the `Display` implementation of `DisplayDiagnostics` for more than writing directly to stdout. Implementing `Display` on the individual renderers seemed like a smaller step in this direction.

At this point it might make sense just to merge this into https://github.com/astral-sh/ruff/pull/20358. Alternatively, I could apply the same transformation to the other affected renderers. For example, the JSON and JSON-lines formats only use their `config` fields for the `preview` setting. Or we could update all of the renderers and call them directly in Ruff instead of assembling a config at all. I just wanted to make sure this approach looked reasonable before applying it to everything.

## Test Plan

Existing tests

### DIFF
--- a/crates/ruff_db/src/diagnostic/render/github.rs
+++ b/crates/ruff_db/src/diagnostic/render/github.rs
@@ -3,19 +3,26 @@ use crate::diagnostic::{Diagnostic, DisplayDiagnosticConfig, FileResolver, Sever
 pub(super) struct GithubRenderer<'a> {
     resolver: &'a dyn FileResolver,
     config: &'a DisplayDiagnosticConfig,
+    diagnostics: &'a [Diagnostic],
 }
 
 impl<'a> GithubRenderer<'a> {
-    pub(super) fn new(resolver: &'a dyn FileResolver, config: &'a DisplayDiagnosticConfig) -> Self {
-        Self { resolver, config }
+    pub(super) fn new(
+        resolver: &'a dyn FileResolver,
+        config: &'a DisplayDiagnosticConfig,
+        diagnostics: &'a [Diagnostic],
+    ) -> Self {
+        Self {
+            resolver,
+            config,
+            diagnostics,
+        }
     }
+}
 
-    pub(super) fn render(
-        &self,
-        f: &mut std::fmt::Formatter,
-        diagnostics: &[Diagnostic],
-    ) -> std::fmt::Result {
-        for diagnostic in diagnostics {
+impl std::fmt::Display for GithubRenderer<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        for diagnostic in self.diagnostics {
             let severity = match diagnostic.severity() {
                 Severity::Info => "notice",
                 Severity::Warning => "warning",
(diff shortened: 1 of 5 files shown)


ntBre:
I'm pretty sure the ecosystem and benchmark results are just a quirk of the base branch, but I'll keep an eye on that for sure.

crates/ruff_db/src/diagnostic/render/github.rs
```diff
@@ -1,21 +1,28 @@
...
     }
+}
 
-    pub(super) fn render(
-        &self,
-        f: &mut std::fmt::Formatter,
-        diagnostics: &[Diagnostic],
-    ) -> std::fmt::Result {
-        for diagnostic in diagnostics {
+impl std::fmt::Display for GithubRenderer<'_> {
```
MichaReiser: It wasn't clear to me from the summary what the motivation was for implementing `Display` for `Renderer` but, after playing with the code, the issue seems to be that ty uses an `std::fmt::Write` whereas Ruff uses an `std::io::Write`. 

I do find it odd that a `Renderer` can be displayed, especially because its display implementation shows the diagnostic and not the renderer. 

1. Introduce a `Renderer` trait that has a single `render` method that takes a `std::fmt::Formatter` and a `&[Diagnostic]` and create a new `DisplayDiagnostics<R: Renderer>` that implements `Display` by forwarding to the wrapped renderer. The main challenge with this is that we already have a `DisplayDiagnostics` type in `ruff_db`. My opinion is still that we should delete the old `DisplayDiagnostics` type and move `DiagnosticFormat` out of `ruff_db`, as which formats are supported is a concern of ty/ruff. But this is a bit a larger refactor (but I think it would be worth it because it also helps clarify which program supports which `DiagnosticFormat`s). 
2. Rename `GithubRenderer` to `GithubDiagnostics`. That's obviously a much simpler change but it's a bit more awkward because it no longer fits into the `render` module.

For this PR, I think what I'd do is to do a mix of 1 and 2 by:

* Restore the `render` method on `GithubRenderer`
* Introduce a new `DisplayGithubDiagnostics` type which takes a renderer and diagnostics as input and implements `Display`. We can use this type in places where we need to write to a `std::io::Write`

ntBre: Ah yeah, sorry that wasn't clear. Yes, Ruff's `Emitter`s take a `&mut dyn std::io::Write`, but `DisplayDiagnostics` and the `Renderer`s in ty operate via `std::fmt::Display` and `std::fmt::Formatter`s. As I mentioned on https://github.com/astral-sh/ruff/pull/20358, I thought the best fix was moving `DisplayDiagnostics` and the renderers to use `std::io::Write` too, but there were a few places where this wasn't an easy switch. These are two of the less important cases since they're in tests:

https://github.com/astral-sh/ruff/blob/aa63c24b8f4a82f5f54dc90c4e6b5adadbf3e1b2/crates/ruff_db/src/diagnostic/render.rs#L2749-L2760

This seemed like the more painful case that made me rethink the approach and try implementing `Display` for each renderer:

https://github.com/astral-sh/ruff/blob/34ba738c19c557653567defc5cdaaa75445ad0d4/crates/ty_project/src/metadata/options.rs#L1465-L1472

I just tried fixing that one and more keep popping up, so I think your suggestions sound very helpful. Thank you!

PR #20426, opened 2025-09-15
[syntax-errors] Name is parameter and global

11happy:
## Summary

This PR implements a new semantic syntax error where name is parameter & global.

## Test Plan

I have written inline test as directed in #17412


crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -76,12 +76,35 @@ impl SemanticSyntaxChecker {
...
                 }
                 Self::duplicate_parameter_name(parameters, ctx);
+
+                let mut param_names = FxHashSet::default();
+                for param in parameters {
+                    let cur_name = param.name().as_str();
+                    param_names.insert(cur_name);
+                }
+
+                for stmt in body {
```
ntBre: This kind of traversal isn't fully correct. It will only catch top-level `global` statements like the test case you included, but there are other ways of nesting `global` statements within other statements that are still syntax errors. A couple of simple examples are `if` and `try` statements:

```pycon
>>> def f(a):
...     if True:
...         global a
...
  File "<python-input-3>", line 3
    global a
    ^^^^^^^^
SyntaxError: name 'a' is parameter and global
>>> def f(a):
...     try:
...         global a
...     except:
...         pass
...
  File "<python-input-8>", line 3
    global a
    ^^^^^^^^
SyntaxError: name 'a' is parameter and global
```

I added these examples on your branch locally, and they were not identified as errors.

The way we handle this for `LoadBeforeGlobalDeclaration` is by using the `SemanticSyntaxContext::global` method, which checks whether a given binding has been declared `global`:

https://github.com/astral-sh/ruff/blob/c3f2187fdae2f706488ec193db395800ec7a9b82/crates/ruff_python_parser/src/semantic_errors.rs#L717-L720

Unfortunately that doesn't seem to work here either because we're not yet inside the function scope when traversing the function definition, and the `global` method checks the current scope in Ruff's semantic model.

I think the approach I would try would be to perform this check when visiting a `Stmt::Global` instead of when visiting the `Stmt::FunctionDef` and check if any of the names in the `global` statement shadows a parameter. It looks like we track this kind of information in the [`BindingKind::Argument`](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_semantic/src/binding.rs#L532-L537) variant in Ruff and possibly [`Scope::shadowed_bindings`](https://github.com/astral-sh/ruff/blob/c3f2187fdae2f706488ec193db395800ec7a9b82/crates/ruff_python_semantic/src/scope.rs#L38). Note that the shadowed check may be necessary because cases like this still emit the `parameter and global` syntax error (this would be a good test case to have too):

```pycon
>>> def f(a):
...     a = 1
...     global a
...
  File "<python-input-9>", line 3
    global a
    ^^^^^^^^
SyntaxError: name 'a' is parameter and global
```

I think we can do something similar with ty's [`DefinitionKind::Parameter`](https://github.com/astral-sh/ruff/blob/c3f2187fdae2f706488ec193db395800ec7a9b82/crates/ty_python_semantic/src/semantic_index/definition.rs#L671), but I'm less sure about that.

If we can't do something like that, we'd need a `Visitor` to traverse the full function definition looking for `global` statements. However, that could be quite expensive, so the approach I outlined above would be preferable.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1230,7 +1256,10 @@ pub enum SemanticSyntaxErrorKind {
     /// to be very rare and not worth the additional complexity to detect.
     ///
     /// [#111123]: https://github.com/python/cpython/issues/111123
-    LoadBeforeGlobalDeclaration { name: String, start: TextSize },
+    LoadBeforeGlobalDeclaration {
```
ntBre: I'd prefer not to reformat unrelated sections of code, as I mentioned on your other PR

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1362,6 +1394,8 @@ pub enum SemanticSyntaxErrorKind {
 
     /// Represents the use of `yield from` inside an asynchronous function.
     YieldFromInAsyncFunction,
+
+    GlobalParameter(String),
```
ntBre, line 1542: We'll need some docs here too

crates/ruff_python_parser/src/semantic_errors.rs
11happy: sure , actually I am not changing these , these are automatically added by pre-commit jobs

crates/ruff_python_parser/src/semantic_errors.rs
11happy: sure will add

crates/ruff_python_parser/src/semantic_errors.rs
11happy: Hii @ntBre , I have tried implementation based on above pointers,  can you please take a look whenever you get a chance.

Thank you : )

crates/ruff_python_parser/src/semantic_errors.rs
11happy: done, I have added docs

crates/ruff_python_parser/src/semantic_errors.rs
11happy: Have fixed the CI errors. 
: )

crates/ruff_linter/resources/test/fixtures/syntax_errors/global_parameter.py
```diff
@@ -0,0 +1,16 @@
+def f(a):
+    global a # error
+
+def g(a):
+    if True:
+        global a # error
+
+def h(a):
+    def inner():
+        global a # error
```
ntBre: This one is actually not an error in the REPL:

```pycon
>>> def h(a):
...     def inner():
...         global a
...
>>> h
<function h at 0x7f684173fd80>
```

crates/ruff_linter/src/rules/ruff/rules/global_parameter.rs
ntBre, line 1: We don't need to define a new Ruff _rule_ in this case, we can just emit this as a syntax error. The only reason we've had rules/`Violation`s in your previous PRs is that those syntax errors were previously implemented as lint rules. New syntax errors that don't correspond to existing lint rules can just be standalone syntax errors like the ones here:

https://github.com/astral-sh/ruff/blob/cacfb09a59eef3d7af8a6067d981c8e85ae5a104/crates/ruff_linter/src/checkers/ast/mod.rs#L713-L730

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -811,6 +817,25 @@ impl SemanticSyntaxContext for Checker<'_> {
             }
         )
     }
+    fn is_binded_parameter(&self, name: &str) -> bool {
```
ntBre: nit: I think a name like `is_parameter` or `is_bound_parameter` would be a bit better.

More importantly, and related to my comment on the test case above, I don't think this is quite right. We need to limit our parameter check to the nearest-enclosing function scope. In other words, we need to search through `self.semantic.current_scopes` more like we do in some of the other context methods:

https://github.com/astral-sh/ruff/blob/bedfc6fd063e08f4d1e76ae8a780162404079298/crates/ruff_linter/src/checkers/ast/mod.rs#L737-L749

until we find the first function scope, and then only check for shadowed bindings within that scope. We can also return early if we find a class scope, for example, because this is also valid, like the nested function case I showed above:

```pycon
>>> def h(a):
...     class C:
...         global a
...
>>>
```

I think the checks on the binding below look roughly correct, but we need to be more careful about identifying the scope holding the relevant binding.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -2777,6 +2777,9 @@ impl SemanticSyntaxContext for SemanticIndexBuilder<'_, '_> {
             self.semantic_syntax_errors.borrow_mut().push(error);
         }
     }
+    fn is_binded_parameter(&self, _name: &str) -> bool {
+        false
```
ntBre, line 2793: Ideally we would implement this for ty too, but we can leave it for a follow-up if it looks complicated or if you aren't interested. That's totally fine :)

ntBre (commented):
Thank you, this is looking good. We just need to be a bit more careful with our scope handling and revert the unnecessary `Violation` changes.

crates/ruff_linter/resources/test/fixtures/syntax_errors/global_parameter.py
11happy: yeah, now its working fine.

crates/ruff_linter/src/rules/ruff/rules/global_parameter.rs
11happy: done

crates/ruff_linter/src/checkers/ast/mod.rs

crates/ty_python_semantic/src/semantic_index/builder.rs
11happy: Sure, I am interested to implement this for ty , I would prefer to keep it as a follow-up if that sounds fine ?.

Thank you : )

crates/ruff_python_parser/tests/fixtures.rs
```diff
@@ -571,6 +571,9 @@ impl SemanticSyntaxContext for SemanticSyntaxCheckerVisitor<'_> {
     fn in_generator_scope(&self) -> bool {
         true
     }
+    fn is_bound_parameter(&self, _name: &str) -> bool {
```
ntBre, line 579: nit:


```suggestion

    fn is_bound_parameter(&self, _name: &str) -> bool {
```

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -2777,6 +2777,9 @@ impl SemanticSyntaxContext for SemanticIndexBuilder<'_, '_> {
             self.semantic_syntax_errors.borrow_mut().push(error);
         }
     }
+    fn is_bound_parameter(&self, _name: &str) -> bool {
```
ntBre, line 2792:
```suggestion

    fn is_bound_parameter(&self, _name: &str) -> bool {
```

crates/ty_python_semantic/src/semantic_index/builder.rs
ntBre: That sounds fine!

11happy:
So the shadowed_bindings in `SemanticModel` and `Scope` are different, respectively. As per the example given in the docstring of `Scope`
```
/// A map from binding ID to binding ID that it shadows.
///
/// For example:
/// ```python
/// def f():
///     x = 1
///     x = 2
/// ```
///
/// In this case, the binding created by `x = 2` shadows the binding created by `x = 1`.
shadowed_bindings: FxHashMap<BindingId, BindingId>,
```
We can utilize this to trace back whether a global declaration is originally an argument or not — and I’ve done that accordingly. I also think we don’t really need to interact with rebindings_scope here. I manually checked with an example, and it was coming out to be empty in the following case:
```
def f(a):
    a = 1
    a = 2
    global a
```
Also, while trying to solve this issue, I read a lot about bindings, shadowed bindings, etc., and I’m genuinely thankful for the opportunity to work on this. The tests are working as expected now

Thank you : )

crates/ruff_python_parser/tests/fixtures.rs
11happy: done

crates/ty_python_semantic/src/semantic_index/builder.rs
11happy: done

11happy:
is the `CI/fuzz parser` error related to this PR changes ?

crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__semantic_syntax_error_GlobalParameter_global_parameter_3.10.snap
```diff
@@ -32,3 +32,34 @@ invalid-syntax: name `a` is parameter and global
...
+28 | def f(a):
+   |
+
+invalid-syntax: name `a` is parameter and global
+  --> <filename>:30:16
+   |
+28 | def f(a):
+29 |     class Inner:
+30 |         global a
+   |                ^
```
ntBre: This is what the fuzzer seeds that are failing look like. This is not supposed to be an error:

```pycon
>>> def f(a):
...     class Inner:
...         global a
...
>>> f
<function f at 0x7fbec085be20>
```

11happy: done , have fixed it & fixed the merge conflict as well (rebased)
Thank you : )

11happy: Currently we are doing an early return for class scope , please let me know if there are any more edge cases you would like me to test , would be happy to add them.

Thank you

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -839,6 +840,38 @@ impl SemanticSyntaxContext for Checker<'_> {
...
         false
     }
+
+    fn is_bound_parameter(&self, name: &str) -> bool {
+        for scope in self.semantic.current_scopes() {
+            match scope.kind {
+                ScopeKind::Class(_) => {
+                    return false;
+                }
+                ScopeKind::Function(_) | ScopeKind::Lambda(_) => {
```
ntBre: Ah, I overlooked this initially, but the `ScopeKind`s here actually contain these nodes' definitions. We should be able to extract the parameters from that and avoid this binding stuff entirely. I always get a little nervous when I see a `loop`, so I looked a bit harder at this today.

```rust
    fn is_bound_parameter(&self, name: &str) -> bool {
        for scope in self.semantic.current_scopes() {
            match scope.kind {
                ScopeKind::Class(_) => return false,
                ScopeKind::Function(ast::StmtFunctionDef { parameters, .. })
                | ScopeKind::Lambda(ast::ExprLambda {
                    parameters: Some(parameters),
                    ..
                }) => return parameters.includes(name),
                ScopeKind::Lambda(_)
                | ScopeKind::Generator { .. }
                | ScopeKind::Module
                | ScopeKind::Type
                | ScopeKind::DunderClassCell => {}
            }
        }

        false
    }
```

I believe it's actually impossible for a lambda to contain a `global` since it can only contain a single expression and not a statement, so we could probably eliminate the lambda branch here too.

crates/ruff_linter/src/linter.rs
```diff
@@ -1041,6 +1041,42 @@ mod tests {
         PythonVersion::PY310,
         "DuplicateMatchKey"
     )]
+    #[test_case(
```
ntBre, line 1043: We shouldn't do this in this PR, but I think it would be nice to move these test cases out to separate test files. It's getting a bit unwieldy. Feel free to follow up on that if you're interested :)

crates/ruff_linter/src/checkers/ast/mod.rs

crates/ruff_linter/src/linter.rs
11happy: Sure, would open a follow up PR.

Thank you

crates/ruff_linter/src/checkers/ast/mod.rs
11happy: Now that I’m looking at this and the previous iterations, I feel like if we’re starting to make the code more complex, something’s probably off with the approach. From now on, I’ll be more careful and focus on keeping things simpler.

Thank you : )

crates/ruff_linter/src/linter.rs
11happy: I have created a PR here https://github.com/astral-sh/ruff/pull/20926

PR #20427, opened 2025-09-15
[`pyupgrade`] Fix false positive when class name is shadowed by local variable (`UP008`)

### LINKED ISSUE #20422: UP008 should not apply when the class variable is shadowed
### Summary

When the first argument to `super` in a method is a local variable, [`super-call-with-parameters` (UP008)](https://docs.astral.sh/ruff/rules/super-call-with-parameters/) should not apply, even if the local variable has the same name as the class. [Example](https://play.ruff.rs/02304bc1-1e7f-4fa4-893b-07fc4903489a):
```console
$ cat >up008.py <<'# EOF'
class A:
    def f(self):
        return 1
class B(A):
    def f(self):
        return 2
class C(B):
    def f(self):
        C = B
        return super(C, self).f()
print(C().f())
# EOF

$ python up008.py
1

$ ruff --isolated check up008.py --select UP008 --unsafe-fixes --fix

$ python up008.py
2
```

### Version

danparizher:
## Summary

Fixes #20422

### DIFF
--- a/crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
+++ b/crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
@@ -139,19 +139,12 @@ pub(crate) fn super_call_with_parameters(checker: &Checker, call: &ast::ExprCall
         return;
     };
 
-    if !((first_arg_id == "__class__" || first_arg_id == parent_name.as_str())
+    if !((first_arg_id == "__class__"
+        || (first_arg_id == parent_name.as_str()
+            && !checker.semantic().current_scope().has(first_arg_id)))
         && second_arg_id == parent_arg.name().as_str())
     {
         return;
-    }
-
-    // If the first argument matches the class name, check if it's a local variable
-    // that shadows the class name. If so, don't apply UP008.
-    if first_arg_id == parent_name.as_str() {
-        let scope = checker.semantic().current_scope();
-        if scope.has(first_arg_id) {
-            return;
-        }
     }
 
     drop(parents);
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
```diff
@@ -145,6 +145,15 @@ pub(crate) fn super_call_with_parameters(checker: &Checker, call: &ast::ExprCall
...
     }
 
+    // If the first argument matches the class name, check if it's a local variable
+    // that shadows the class name. If so, don't apply UP008.
+    if first_arg_id == parent_name.as_str() {
+        let scope = checker.semantic().current_scope();
+        if scope.has(first_arg_id) {
+            return;
+        }
+    }
```
ntBre: Can we combine this with the other `first_arg_id == parent_name.as_str()` check to avoid having to check that twice?

```rust
    if !((first_arg_id == "__class__"
        || (first_arg_id == parent_name.as_str()
            && !checker.semantic().current_scope().has(first_arg_id)))
        && second_arg_id == parent_arg.name().as_str())
    {
        return;
    }
```

This condition is definitely getting a bit unwieldy, but I don't think it's too much worse.

An alternative approach here would be to resolve the binding of the first argument name and compare that to the binding of the parent class, but I think this check is okay because any intervening scope should also disable the rule. The current approach is a bit simpler.

crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
```diff
@@ -139,21 +139,14 @@ pub(crate) fn super_call_with_parameters(checker: &Checker, call: &ast::ExprCall
         return;
     };
 
-    if !((first_arg_id == "__class__" || first_arg_id == parent_name.as_str())
+    if !((first_arg_id == "__class__"
+        || (first_arg_id == parent_name.as_str()
+            && !checker.semantic().current_scope().has(first_arg_id)))
```
ntBre, line 146:
```suggestion
            // If the first argument matches the class name, check if it's a local variable
            // that shadows the class name. If so, don't apply UP008.
            && !checker.semantic().current_scope().has(first_arg_id)))
```

Just adding back your comment, we'll see if I got the formatting right in the `suggestion` box!

PR #20428, opened 2025-09-16
[ty] Make TypeIs invariant in its type argument

ericmarkmartin:
## Summary

What it says on the tin. See the [typing spec](https://docs.python.org/3/library/typing.html#typing.TypeIs) for justification.

## Test Plan

Add more tests to PEP 695 `variance.md` suite.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/generics/pep695/variance.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/pep695/variance.md
@@ -752,6 +752,36 @@ b_container = ClassContainer[B](B)
 a_instance: A = use_a_class_container(b_container)  # This should work
 ```
 
+## TypeIs
+
+```toml
+[environment]
+python-version = "3.13"
+```
+
+`TypeIs[T]` is invariant in `T`. See the [typing spec][typeis-spec] for a justification.
+
+```py
+from typing import TypeIs
+from ty_extensions import is_assignable_to, is_subtype_of, static_assert
+
+class A:
+    pass
+
+class B(A):
+    pass
+
+class C[T]:
+    def check(x: object) -> TypeIs[T]:
+        # this is a bad check, but we only care about it type-checking
+        return False
+
+static_assert(not is_subtype_of(C[B], C[A]))
+static_assert(not is_subtype_of(C[A], C[B]))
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/generics/pep695/variance.md
```diff
@@ -752,6 +752,31 @@ b_container = ClassContainer[B](B)
...
+    pass
+
+class B(A):
+    pass
+
+assert not is_subtype_of(TypeIs[B], TypeIs[A])
+assert not is_subtype_of(TypeIs[A], TypeIs[B])
+assert not is_assignable_to(TypeIs[B], TypeIs[A])
+assert not is_assignable_to(TypeIs[A], TypeIs[B])
+```
```
InSyncWithFoo: Isn't `TypeIs`'s variance already tested [in `is_subtype_of.md`](https://github.com/astral-sh/ruff/blob/2a6dde4acb6cb6a8ef4426ca3d792bdb1c477f21/crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of.md#typeguard-and-typeis)?

ericmarkmartin: Oh this is interesting. The pre-existence of these tests implies the invariance was already being enforced? I'll have to look further

carljm: Do the tests you added here fail before the code change in this PR?

It seems to me that what you are testing for here is assignability/subtyping of `TypeIs` types themselves (which I think was already correctly invariant), but the code change you made would not impact that, it would instead affect variance inference when you have e.g. `TypeIs[T]` as part of a method signature or as an attribute on a PEP695 generic class with typevar `T`. Which seems like a rare/unusual case, but still worth getting right.

ericmarkmartin: Yeah, I missed that there was existing logic for the in variance of `TypeIs`. Changed the tests, and also fixed `assert` -> `static_assert` 🙈

carljm (approved):
Thank you!

PR #20429, opened 2025-09-16
[`flake8-bugbear`] Add `B912`: `map()` without an explicit `strict=` parameter

### LINKED ISSUE #20057: New rule: `B912` require `strict=` argument for `map(...)` with >=2 iterables on Python 3.14+
### Summary

cf https://github.com/PyCQA/flake8-bugbear/pull/517; related to https://github.com/astral-sh/ruff/pull/19333

danparizher:
## Summary

Implements new rule `B912` that requires the `strict=` argument for `map(...)` calls with two or more iterables on Python 3.14+, following the same pattern as `B905` for `zip()`.

Closes #20057


crates/ruff_linter/src/rules/flake8_bugbear/rules/map_without_explicit_strict.rs
```diff
@@ -0,0 +1,134 @@
...
+use ruff_python_semantic::SemanticModel;
+use ruff_text_size::Ranged;
+
+use crate::checkers::ast::Checker;
+use crate::fix::edits::add_argument;
+use crate::{AlwaysFixableViolation, Applicability, Fix};
+
+/// ## What it does
+/// Checks for `map` calls without an explicit `strict` parameter when called with two or more iterables.
+///
```
dylwil3, line 13: Can you add somewhere in the documentation that this only applies to versions Python 3.14 and later? And maybe add a link to the section of "What's new" that has this change?

crates/ruff_linter/src/rules/flake8_bugbear/rules/map_without_explicit_strict.rs
dylwil3, line 1: This looks very close to `zip_without_explicit_strict.rs`. Could we extract the common logic into a helper function somehow, or are they subtly different?

crates/ruff_linter/src/rules/flake8_bugbear/rules/map_without_explicit_strict.rs
```diff
@@ -0,0 +1,134 @@
...
+                } else {
+                    Applicability::Safe
+                },
+            ));
+    }
+}
+
+/// Return `true` if the [`Expr`] appears to be an infinite iterator (e.g., a call to
+/// `itertools.cycle` or similar).
+pub(crate) fn is_infinite_iterable_for_map(arg: &Expr, semantic: &SemanticModel) -> bool {
```
dylwil3: Can we not re-use (or alter and then re-use) the crate-visible function `is_infinite_iterable` here?

danparizher:
I extracted common helpers into `strict_utils.rs`:
  - `is_infinite_iterable(...)`
  - `any_infinite_iterables(...)`

Applied across `B905`, `B911`, and `B912` to remove duplicated logic

Let me know if that makes the most sense, happy to adjust naming/placement of the helpers if you prefer a different structure

PR #20430, opened 2025-09-16
[ty] Add support for `**kwargs`

dhruvmanila:


dhruvmanila:
## Ecosystem analysis

In `boostedblob` and `mkosi`, all of the new diagnostics are due to the fact that the assignment has been annotated as `dict[str, Any]` but the inferred type is `dict[str, <class 'int'> | str | Unknown | int]`, so ty will try to check assignment between `<class 'int'> | str | Unknown | int` and the matching parameter type.

In `colour`, all of the new diagnostics are due to:
```python
def optional[T](value: T | None, default: T) -> T: ...

def _(value: dict | None = None):
    # Here, `None` shouldn't be present
    # revealed: dict[Unknown, Unknown] | None 
    reveal_type(optional(value), {})
```

In `home-assistant/core`, there's a case which requires support for narrowing on `TypedDict` (?). The scenario is that the type of a variable is a union of two `TypedDict` and the `if` condition is checking whether a key exists in the union type and narrows based on that. For example:
```python
from typing import TypedDict


class Foo1(TypedDict):
    a: int


class Foo2(TypedDict):
    b: str


def _(f: Foo1 | Foo2):
    if "a" in f:
        # This should reveal `Foo1`
        reveal_type(f)
    else:
        # This should reveal `Foo2`
        reveal_type(f)
```

In `hydra-zen`, all of the new diagnostics are because ty doesn't support `TypeGuard` narrowing.

In `pandas`, the diagnostics are valid, Pyright raises them as well but the specific rule has been disabled in the project config for Pyright.

~_I'm still looking at a few more but this is ready for review._~

dhruvmanila:
The benchmark is failing because ty is now raising more number of diagnostics than the current limit for `static-frame`

https://github.com/astral-sh/ruff/blob/fed9d6d02cc97c078bee3992820034f16c3e6f49/crates/ruff_benchmark/benches/ty_walltime.rs#L221-L237

I'll increase the limit to 600.

The new diagnostics seems valid and Pyright raises them too. They're due to the following pattern where the `data` is being splatted to the parameters but the value type that ty infers is going to be a union of all the value types present in the `dict` creation:
```py
data = dict(...)

func(**data)
```

Here's an example in the `static-frame` repository: https://github.com/static-frame/static-frame/blob/9021541074a1b9b3f51fbb4cab545ad2ddbc7d0b/static_frame/core/interface.py#L1525-L1537

Pyright raises these errors as well but it's being silenced using `# pyright: ignore`.

sharkdp:
> there's a case which requires support for narrowing on `TypedDict` (?). The scenario is that the type of a variable is a union of two `TypedDict` and the `if` condition is checking whether a key exists in the union type and narrows based on that.

This sounds *somewhat* similar to https://github.com/astral-sh/ty/issues/643. Maybe we could add a comment there that this is another use-case?

carljm:
Looking at the conformance suite, it seems the issues there are related to still lacking `Unpack` and `ParamSpec` support?

carljm:
> In `boostedblob` and `mkosi`, all of the new diagnostics are due to the fact that the assignment has been annotated as `dict[str, Any]` but the inferred type is `dict[str, <class 'int'> | str | Unknown | int]`

So this would be another case to consider in https://github.com/astral-sh/ty/issues/136

> In `colour`, all of the new diagnostics are due to:
> 
> ```python
> def optional[T](value: T | None, default: T) -> T: ...
> 
> def _(value: dict | None = None):
>     # Here, `None` shouldn't be present
>     # revealed: dict[Unknown, Unknown] | None 
>     reveal_type(optional(value), {})
> ```

And this is https://github.com/astral-sh/ty/issues/623

crates/ty_python_semantic/resources/mdtest/call/function.md
```diff
@@ -871,3 +871,285 @@ is_subtype_of(int, int, int)
...
+
+# error: [invalid-argument-type] "Argument expression after ** must be a mapping with `str` key type: Found `int`"
+f(**DictSubclass())
+# error: [invalid-argument-type] "Argument expression after ** must be a mapping with `str` key type: Found `int`"
+f(**MappingSubclass())
+# error: [invalid-argument-type] "Argument expression after ** must be a mapping with `str` key type: Found `int`"
+f(**MappingProtocol())
+```
+
+The key can also be a custom type that inherits from `str`.
```
carljm, line 1051: What about if the key type is `Any` or `Unknown`? (That is, something assignable to `str` but not a subtype of `str`.) It looks like this is handled, but not tested.

crates/ty_python_semantic/resources/mdtest/call/function.md
```diff
@@ -871,3 +871,285 @@ is_subtype_of(int, int, int)
...
+
+```py
+from ty_extensions import Unknown
+
+def f(**kwargs: int) -> None: ...
+def _(kwargs: Unknown):
+    f(**kwargs)
+```
+
+### Key is not a mapping
```
carljm: It's not the "key" that needs to be a mapping.

```suggestion
### Not a mapping
```

dhruvmanila:
> Looking at the conformance suite, it seems the issues there are related to still lacking `Unpack` and `ParamSpec` support?

Yeah, that's correct.

crates/ty_python_semantic/resources/mdtest/call/function.md
dhruvmanila: Added a test case for these cases, thanks!

crates/ty_python_semantic/resources/mdtest/call/function.md
dhruvmanila: Right! Thanks for catching it

dhruvmanila:
(Happy to address any other feedback on this PR in a follow-up :))

PR #20433, opened 2025-09-16
[ty] Update mypy_primer

sharkdp:
## Summary

Revert the materialize-changes, see https://github.com/hauntsaninja/mypy_primer/pull/208


AlexWaygood (approved):

sharkdp:
Something very strange is going on. CI fails with "pip install failed for **materialize**", but I can't see the actual error. If I update the mypy_primer commit to point to https://github.com/hauntsaninja/mypy_primer/pull/209, I see a proper traceback, but now I get "pip install failed for **cwltool**"! And I still can't see the stdout/stderr of `uv pip install`. I can reproduce neither of these problems locally.

sharkdp:
Ok, now suddenly everything works fine. Not sure what happened yesterday. Some kind of caching effect?

PR #20434, opened 2025-09-16
Generator preferred quote style

### LINKED ISSUE #20432: Can't configure quote style for `ruff_python_codegen::Generator`
There's support for creating a `Generator` from a `Stylist` instance:

https://github.com/astral-sh/ruff/blob/2a6dde4acb6cb6a8ef4426ca3d792bdb1c477f21/crates/ruff_python_codegen/src/stylist.rs#L11-L17

https://github.com/astral-sh/ruff/blob/2a6dde4acb6cb6a8ef4426ca3d792bdb1c477f21/crates/ruff_python_codegen/src/generator.rs#L76-L87

And while `indentation` and `line_ending` are being respected, `quote` is ignored:/

---

Motivation for issue: https://github.com/RustPython/RustPython/pull/6124#discussion_r2320979701

ShaharNaveh:
Closes #20432


crates/ruff_python_codegen/src/generator.rs
```diff
@@ -1585,10 +1603,14 @@ mod tests {
     fn round_trip_with(
         indentation: &Indentation,
         line_ending: LineEnding,
+        quote: Option<Quote>,
```
MichaReiser: The documentation now needs updating

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -2047,9 +2074,43 @@ if True:
...
                 LineEnding::Cr,
+                None,
                 "if True:\n    print(42)",
             ),
             "if True:\r    print(42)",
         );
     }
+
+    #[test]
+    fn set_quote() {
```
MichaReiser:
```suggestion
    fn preferred_quote() {
```

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -67,6 +68,8 @@ pub struct Generator<'a> {
     indent: &'a Indentation,
     /// The line ending to use.
     line_ending: LineEnding,
+    /// Forced quote style to use.
```
MichaReiser: The comment here is inaccurate. Setting `quote` doesn't force the generator to use the given quote everywhere. For example, setting `Quote::Single` for `f'test{f"nested"}'` doesn't change the quotes of the inner f-string. I also recommend adding a test showcasing this behavior

That's why I'd call this field `preferred_quote`. We should also extend the documentation to explain in more detail what it does (When `None`, the generator preserves the quote style when possible. When `Some`, the generator prefers the set quote style, disregarding the quote style used in the source)

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -100,6 +105,13 @@ impl<'a> Generator<'a> {
         }
     }
 
+    /// Force quote style for generated source code.
+    #[must_use]
+    pub fn with_quote(mut self, quote: Quote) -> Self {
```
MichaReiser: Same as on the field documentation
```suggestion
    pub fn with_preferred_quote(mut self, quote: Quote) -> Self {
```

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -1393,7 +1405,12 @@ impl<'a> Generator<'a> {
 
     fn unparse_string_literal(&mut self, string_literal: &ast::StringLiteral) {
         let ast::StringLiteral { value, flags, .. } = string_literal;
-        self.p_str_repr(value, *flags);
+        let flags = if let Some(quote) = self.quote {
+            flags.with_quote_style(quote)
+        } else {
+            *flags
+        };
+        self.p_str_repr(value, flags);
```
MichaReiser: I think it's clearer if we move the `self.quote` override to `self.p_str_repr` instead of "patching" the flags here.

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: that's a better name for sure

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: that's a better name for sure

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -169,13 +181,18 @@ impl<'a> Generator<'a> {
...
             write!(self.buffer, "{}", flags.display_contents(s))
                 .expect("Writing to a String buffer should never fail");
             return;
         }
         self.p(flags.prefix().as_str());
+
+        if let Some(preferred_quote) = self.preferred_quote {
+            flags = flags.with_quote_style(preferred_quote);
+        }
+
```
MichaReiser: Instead of patching the flags here, can we do something like this:

```rust
let quote_style = self.preferred_quote.unwrap_or(flags.quote_style());
let escape = UnicodeEscape::with_preferred_quote(s, quote_style);
```

Do we need to add the same handling to:

* `unparse_interpolated_string_literal_element`
* `p_bytes_repr`
* `unparse_interpolated_string`

ShaharNaveh: That's a nice cleanup. will look into the other methods

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -100,6 +105,17 @@ impl<'a> Generator<'a> {
         }
     }
 
+    /// Set a preferred quote style for generated source code.
+    ///
+    /// - If [`None`], the generator will attempt to preserve the existing quote style whenever possible.
+    /// - If [`Some`], the generator will always use the specified quote style,
+    /// ignoring the one found in the source.
+    #[must_use]
+    pub fn with_preferred_quote(mut self, quote: Option<Quote>) -> Self {
```
ShaharNaveh, line 113: @MichaReiser I've changed the signature to be able to take `Option<Quote>`, I think it makes more sense as it gives the ability to unset the preferred quote style & better aligns with the method's doc (IMO)

PR #20437, opened 2025-09-16
Shard instrumented benchmark

MichaReiser:
Shard the instrumented benchmarks and only run the ruff/ty benchmarks when there were corresponding code changes.

I tested that the corresponding benchmarks run when making changes to ruff or ty

* https://github.com/astral-sh/ruff/pull/20450
* https://github.com/astral-sh/ruff/pull/20451

This PR increases the overall CI time when making changes that apply to both ruff and ty, as it now requires building the shared crates twice. However, it should reduce CI time when only changing code for one project because the benchmark jobs now only build the relevant benchmarks (e.g. ty doesn't need to build `ruff_linter` and `ruff`).

Sharding the walltime benchmarks is a bit more involved because codspeed doesn't support filtering the benchmarks by name but this is something they plan to add soon.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -906,21 +906,14 @@ jobs:
         run: npm run fmt:check
         working-directory: playground
 
-  benchmarks-instrumented:
+  benchmarks-instrumented-ruff:
+    name: "benchmarks instrumented (ruff)"
     runs-on: ubuntu-24.04
     needs: determine_changes
     if: |
       github.ref == 'refs/heads/main' || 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -907,10 +906,13 @@ jobs:
         run: npm run fmt:check
         working-directory: playground
 
-  benchmarks-instrumented:
+  benchmarks-instrumented-ruff:
```
MichaReiser, line 909: I tried to use a matrix to reduce code duplication but matrix doesn't provide an easy way to conditionally skip an entire job.

MichaReiser:
Let's give this a try. It's easy enough to revert

PR #18455, opened 2025-06-04
[ty] ty_ide: Hotfix for `expression_scope_id` panics

sharkdp:
## Summary

Implement a hotfix for the playground/LSP crashes related to missing `expression_scope_id`s.

relates to: https://github.com/astral-sh/ty/issues/572

## Test Plan

* Regression tests from https://github.com/astral-sh/ruff/pull/18441
* Ran the playground locally to check if panics occur / completions still work.

PR #20439, opened 2025-09-16
[ty] Add imports when an unimported completion is selected

BurntSushi:
This PR does the work necessary to make the "import"
part of "auto-import" work. That is, when an unimported
completion is found, it is now returned with an LSP text
edit for inserting an import that should bring that symbol
into scope.

The diffstat here is quite large, but a bulk of that are
tests. And this PR should be reviewed commit-by-commit.
Many of the commits are somewhat smaller refactorings.
The last commit is where the new `Importer` abstraction is
introduced, and it's where (arguably) most of the weeds are.

For now, I've not concerned myself with performance in order
to avoid getting distracted by it. In particular, every
unimported completion gets an edit generated for it. This
strikes me as sub-optimal, and indeed, it looks like the
LSP protocol has support for running "commands" in response
to a completion being selected. Future work might involve
taking advantage of that, but it's not clear to me yet how
expensive (relatively speaking) generating these edits
actually is.

### DIFF
--- a/crates/ty_python_semantic/src/semantic_model.rs
+++ b/crates/ty_python_semantic/src/semantic_model.rs
@@ -267,10 +267,6 @@ impl<'db> SemanticModel<'db> {
     fn scope(&self, node: ast::AnyNodeRef<'_>) -> Option<FileScopeId> {
         let index = semantic_index(self.db, self.file);
 
-        // TODO: We currently use `try_expression_scope_id` here as a hotfix for [1].
-        // Revert this to use `expression_scope_id` once a proper fix is in place.
-        //
-        // [1] https://github.com/astral-sh/ty/issues/572
         match node {
             ast::AnyNodeRef::Identifier(identifier) => index.try_expression_scope_id(identifier),
             node => match node.as_expr_ref() {


BurntSushi:
Demo:

https://github.com/user-attachments/assets/e381082a-bc82-4f7e-b21b-e9f4a094a3cc

crates/ruff_python_importer/src/insertion.rs
```diff
@@ -135,6 +135,57 @@ impl<'a> Insertion<'a> {
...
+            tokens
+                .before(at)
+                .last()
+                .map(ruff_python_parser::Token::kind),
+            Some(TokenKind::Import)
+        ) {
+            return None;
+        }
+        Some(Insertion::inline(" ", at, ""))
+    }
```
BurntSushi, line 180: This is how I avoided using `libcst`. I believe this is the only use of libcst in ruff's existing importer abstraction: https://github.com/astral-sh/ruff/blob/aa63c24b8f4a82f5f54dc90c4e6b5adadbf3e1b2/crates/ruff_linter/src/importer/mod.rs#L470-L488

crates/ruff_diagnostics/src/edit.rs
```diff
@@ -62,6 +62,11 @@ impl Edit {
         self.content.as_deref()
     }
 
+    /// Returns the range of this edit.
+    pub fn range(&self) -> TextRange {
+        self.range
+    }
+
```
MichaReiser: Is this only so that you don't need to import `Ranged` to call `Range` because `Edit` implements `Ranged` (and, therefore, has a `edit.range` method)

crates/ruff_source_file/src/locator.rs
```diff
@@ -0,0 +1,225 @@
...
+use crate::{LineColumn, LineIndex, LineRanges, OneIndexed, SourceCode};
+
+#[derive(Debug)]
+pub struct Locator<'a> {
+    contents: &'a str,
+    index: OnceCell<LineIndex>,
+}
+
+impl<'a> Locator<'a> {
+    pub const fn new(contents: &'a str) -> Self {
```
MichaReiser: Hmm, I was hoping we could eventually remove `Locator`. What I dislike about the type is that its lazy computation of the index is a bit of a footgun (which is why all those methods are marked as deprecated). 

But `Locator` isn't much more than a `&str` if you remove all methods that don't use a `LineIndex` (`&str` also implements `LineRanges`). The only difference are the helper methods like `after`, `before` that work with `TextRanges`. 

I don't think it's worth spending too much time but is there maybe a way we could decouple `Importer` from `Locator` (e.g. by just using a `&str` and exposing some of those helper methods with a different trait on `&str` which I think could be useful in other places too)

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub use self::insertion::Insertion;
+
+mod insertion;
+
+/// A text edit to be applied to a source file.
+///
+/// Inserts, deletes, or replaces content at a given location.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Edit {
```
MichaReiser: The `convert_edit` calls confused me a fair bit because I didn't realize that there are now two `Edit` types. 

Given that `ruff_diagnostics` has very few dependencies, could we just reuse that type instead of duplicating its definition?

crates/ruff_python_codegen/src/stylist.rs
```diff
@@ -10,34 +11,43 @@ use ruff_text_size::Ranged;
...
     }
 
+    pub fn into_owned(self) -> Stylist<'static> {
+        Stylist {
+            source: Cow::Owned(self.source.into_owned()),
+            indentation: self.indentation,
+            quote: self.quote,
+            line_ending: self.line_ending,
+        }
+    }
```
MichaReiser, line 44: Neat! 

But I'm not sure if this is actually needed, given that we only need `into_owned` in tests and only in tests that call `import` (or `import_from`) of which, I think, most tests only contain exactly one call. 

I'm leaning towards creating the `Stylist` ad-hoc on demand. The only downside is that we compute the `line_ending` a few more times than necessary but that should be neglectable, given that most test files have very short lines (it only searches for the first line ending)

crates/ruff_python_importer/src/insertion.rs
MichaReiser: I'm only mid-through reviewing this PR and I'm going commit-by-commit as suggested so I don't know if you did this in a later commit.

I suggest adding extensive testing for this function, specifically in the presence of comments. 

```py
from collections import (
	Collector # comment
)

from collections import (
	Collector, # comment
)

from collections import (
	Collector # comment
    ,
)

from collections import (
	Collector
    # comment
    ,
)
```

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -30,35 +30,55 @@ use resolve_definition::{find_symbol_in_scope, resolve_definition};
...
                 })
         })
         .chain(use_def_map.all_end_of_scope_symbol_bindings().filter_map(
             move |(symbol_id, bindings)| {
+                // It's not clear to AG how to using a bindings
+                // iterator here to get the correct definition for
+                // this binding. Below, we look through all bindings
+                // with a definition and only take one if there is
+                // exactly one. I don't think this can be wrong, but
+                // it's probably omitting definitions in some cases.
```
MichaReiser, line 62: @sharkdp do you see any issues with this?

crates/ty_ide/src/all_symbols.rs
```diff
@@ -56,10 +57,15 @@ pub fn all_symbols(db: &dyn Db, query: &str) -> Vec<AllSymbolInfo> {
...
+pub struct AllSymbolInfo<'db> {
+    /// The symbol information.
     pub symbol: SymbolInfo<'static>,
-    /// The file containing the symbol
+    /// The module containing the symbol.
+    pub module: Module<'db>,
+    /// The file containing the symbol.
+    ///
+    /// This `File` is guaranteed to be the same
+    /// as the `File` underlying `module`.
```
MichaReiser, line 68: Given that this invariant exists, should we remove file and instead use `module.file(db)` to get the file or do we want to keep `file` because it is guaranteed that `Module` is never a `NamespacePackage`?

Another alternative would be to use `file_to_module(db, file)`, but I guess this has the same downside that it requires you to handle the `None` case

crates/ty_wasm/src/lib.rs
```diff
@@ -425,14 +425,11 @@ impl Workspace {
             .into_iter()
             .map(|completion| Completion {
                 kind: completion.kind(&self.db).map(CompletionKind::from),
-                name: completion.inner.name.into(),
+                name: completion.name.into(),
```
MichaReiser: Would you mind adding the new fields here too (and wire them up in the playground?)

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+This importer is based on a similar abstraction in `ruff_linter::importer`.
+Both of them use the lower level `ruff_python_importer::{Edit, Insertion}`
+primitives. The main differences here are:
+
+1. This work's with ty's semantic model instead of ruff's.
+2. This owns the task of visiting AST to extract imports. This
+   design was chosen because it's currently only used for inserting
+   imports for unimported completion suggestions. If it needs to be
+   used more broadly, it might make sense to roll construction of an
+   `Impoter` into ty's `SemanticIndex`.
```
MichaReiser: We should consider this very carefully, given that the `SemanticIndex` is constructed for every file but adding imports is only used very rarely. I think re-walking the tree here is perfectly fine (and should be very fast too).

An alternative is to introduce a `top_level_imports` query that caches the result but I don't think that's worth it because it needs to re-run on every AST change anyway

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// edit (if any) produced by this importer.
+    ///
+    /// The `file` given should correspond to the module that we want
+    /// to insert an import statement into.
+    ///
+    /// The `Locator` is used to get access to the original source
+    /// text for `file`, which is used to help produce code edits (if
+    /// any). It's also used to re-parse code using `libcst` in order
+    /// to produce code edits that retain "trivia" in the source code
+    /// (i.e., whitespace and comments).
```
MichaReiser: Didn't you abstract away the only libcst usage (which would be great, because libcst is pretty large and chunky to build)

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// providing an empty map is generally fine. But it does mean that
+    /// the resulting import may shadow (or be shadowed by) some other
+    /// symbol.
+    ///
+    /// The `available_at` offset is the byte offset into this
+    /// importer's module source where the symbol being imported should
+    /// be made available. This is necessary to ensure correctness.
+    /// For example, the importer reuses existing imports whenever
+    /// possible, but some imports may not be in scope in some parts of
+    /// the module.
```
MichaReiser: Should this documentation be moved to `members.at`?

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            // is possible the import statement *could* satisfy the
+            // request. For example, if `available_at` is inside a
+            // function defined before the import statement. But this
+            // only works if the function is known to be called *after*
+            // the import statement executes. So... it's complicated.
+            // In the worst case, we'll end up inserting a superfluous
+            // import statement at the top of the module.
+            //
+            // Also, we can stop here since our import statements are
+            // sorted by source sorted.
```
MichaReiser:
```suggestion
            // sorted by their start location in the source.
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            if import.stmt.start() >= available_at {
+                return choice;
+            }
+            if let Some(response) = import.satisfies(self.db, self.file, request) {
+                if choice
+                    .as_ref()
+                    .map_or(true, |c| !c.kind.is_prioritized_over(&response.kind))
```
MichaReiser:
```suggestion
                if choice
                    .is_none_or(|c| !c.kind.is_prioritized_over(&response.kind))
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+                // compared to the more refined scope tracking in the
+                // semantic index builder. However, I don't think
+                // we need anything more refined here. We only care
+                // about top-level imports. So as soon as we get into
+                // something nested, we can bail out.
+                //
+                // Although, this does mean, e.g.,
+                //
+                //     if predicate:
+                //         import whatever
```
MichaReiser, line 726: I think we'll want to support those for type-checking only imports. I think it's totally fine that this PR ignores those, but we should think about when we should add support for them. 

Unlike Rust, importing a module has a runtime cost. Because of that, Python supports type-checking only imports, to avoid importing modules that are only needed to type some API. 

The way type-checking only imports work in Python is that the imports are gated by an `if TYPE_CHECKING` guard (it's a well-known constant by type checkers or it can be imported from the typing module):

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import foo
```

Now, symbols imported in a `TYPE_CHECKING` block aren't available at runtime. That means, they should only be suggested when in a type position (and e.g. not in the condition of an `if` statement). 

It's also preferred to add the imports to an `if TYPE_CHECKING` block if the import is only used for type checking. 

I don't know how far existing LSPs go here but Ruff has a few rules to help with this: https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+        visitor.visit_body(parsed.suite());
+        // Sort so that when we search for imports,
+        // we can bail early based on their position
+        // in the source code.
+        visitor.imports.sort_by_key(|import| import.stmt.start());
+        visitor.imports
+    }
+}
+
+impl<'ast> StatementVisitor<'ast> for TopLevelImports<'ast> {
```
MichaReiser: You could consider implementing `SourceOrderVisitor` instead to remove the need to sort the import statements after visiting because the ordering is already guaranteed by the visitor. 

The only drawback is that there's no statements only equivalent of it but it should be relatively easy to skip unnecessary nodes by overriding `enter_node`

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// way that is correct.
+    ///
+    /// At time of writing (2025-09-15), we just insert a
+    /// qualified import anyway, even though this will result
+    /// in what is likely incorrect code. This seems better
+    /// than some alternatives:
+    ///
+    /// 1. Silently do nothing.
+    /// 2. Silently omit the symbol from completions.
+    ///
```
MichaReiser, line 1356: I think there's a third option to alias the import. That's probably what I would expect but I also think it's fine to leave this as is for now

crates/ty_python_semantic/src/semantic_model.rs
```diff
@@ -217,26 +248,16 @@ impl<'db> SemanticModel<'db> {
...
-            ast::AnyNodeRef::Identifier(identifier) => index.try_expression_scope_id(identifier),
-            node => match node.as_expr_ref() {
-                // If we couldn't identify a specific
-                // expression that we're in, then just
-                // fall back to the global scope.
-                None => Some(FileScopeId::global()),
-                Some(expr) => index.try_expression_scope_id(&expr),
-            },
-        }) else {
+        let Some(file_scope) = self.scope(node) else {
```
MichaReiser, line 247: Can we remove the TODO here, considering that it's repeated in `scope`?

MichaReiser (approved):
This is great! 

I think it would be nice to make use of the LSPs lazy resolution of completion metadata (e.g. lazily resolve the `Type` of a `Completion) but I agree with your scoping that we should leave this to a separate improvement. 

We probably also want some more advanced sorting of completion items to e.g. use the closest (and most high-level) re-export of a symbol over importing it from a very deeply nested submodule (or from a test!). But again, these are incremental improvements that we can make later.

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
+/*!
+Low level helpers for manipulating Python import statements.
```
AlexWaygood:
```suggestion
Low-level helpers for manipulating Python import statements.
```

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// When this edit is a deletion, then this returns `None`.
+    pub fn content(&self) -> Option<&str> {
+        self.content.as_deref()
+    }
+
+    /// Consumes this edit to give back an owned value of the new content for
+    /// an insertion or replacement.
+    ///
+    /// When this edit is a deletion, then this returns `None`.
```
AlexWaygood:
```suggestion
    /// Returns `None` if this edit is a deletion.
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
+#![allow(warnings)]
+
+/*!
+An abstraction for adding new imports to a single Python source file.
+
+This importer is based on a similar abstraction in `ruff_linter::importer`.
+Both of them use the lower level `ruff_python_importer::{Edit, Insertion}`
```
AlexWaygood:
```suggestion
Both of them use the lower-level `ruff_python_importer::{Edit, Insertion}`
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+primitives. The main differences here are:
+
+1. This work's with ty's semantic model instead of ruff's.
```
AlexWaygood:
```suggestion
1. This works with ty's semantic model instead of ruff's.
```

crates/ty_ide/src/importer.rs
AlexWaygood:
```suggestion
   `Importer` into ty's `SemanticIndex`.
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// possible, but some imports may not be in scope in some parts of
+    /// the module.
+    ///
+    /// The "import action" returned includes an edit for inserting
+    /// the actual import (if necessary) along with the symbol text
+    /// that should be used to refer to the imported symbol. While
+    /// the symbol text may be expected to just be equivalent to the
+    /// request's `member`, it can be different. For example, if there
+    /// is an alias or if the corresponding module is already imported
+    /// in a qualified way.
```
AlexWaygood:
```suggestion
    /// request's `member`, it can be different. For example, there
    /// might be an alias, or the corresponding module might already
    /// be imported in a qualified way.
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            }
+            return ImportAction {
+                import: Some(import),
+                symbol_text,
+            };
+        };
+        match response.kind {
+            ImportResponseKind::Unqualified { ast, alias } => {
+                let member = alias.asname.as_ref().unwrap_or(&alias.name).as_str();
+                // As long as it's not a glob, we use whatever name
```
AlexWaygood:
```suggestion
                // As long as it's not a wildcard import, we use whatever name
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            }
+        }
+        choice
+    }
+}
+
+/// A map of symbols in scope at a particular location in a module.
+///
+/// Users of an `Importer` must create this map via
+/// `Importer::members_in_scope_at` in order to use the `Importer::import` API.
```
AlexWaygood:
```suggestion
/// [`Importer::members_in_scope_at`] in order to use the [`Importer::import`] API.
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    fn avoid_conflicts(self, db: &dyn Db, importing_file: File, members: &MembersInScope) -> Self {
+        match (members.map.get(self.module), members.map.get(self.member)) {
+            // Neither symbol exists, so we can just proceed as
+            // normal.
+            (None, None) => self,
+            // The symbol we want to import already exists but
+            // the module symbol does not, so we can import the
+            // symbol in a qualified way safely.
+            (None, Some(member)) => {
+                // ... unless the symbol we want is just already
```
AlexWaygood:
```suggestion
                // ...unless the symbol we want is already
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+                    style: ImportStyle::Import,
+                    force_style: true,
+                    ..self
+                }
+            }
+            // The symbol we want to import doesn't exist but
+            // the module does. So we can import the symbol we
+            // want *unqualified* safely.
+            //
+            // ... unless the module symbol we found here is
```
AlexWaygood, line 530:
```suggestion
            // ...unless the module symbol we found here is
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            ) => self,
+            (Some(_), None) => Self {
+                style: ImportStyle::ImportFrom,
+                force_style: true,
+                ..self
+            },
+            // Both the module and the member symbols are in
+            // scope. We *assume* that the module symbol is in
+            // scope because it is imported. Since the member
+            // symbol is definitively in scope, we attempt a
```
AlexWaygood, line 547:
```suggestion
            // symbol is definitely in scope, we attempt a
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+#[derive(Debug)]
+struct ImportResponse<'importer, 'ast> {
+    import: &'importer AstImport<'ast>,
+    kind: ImportResponseKind<'ast>,
+}
+
+/// The kind of response to an import request.
+///
+/// This encodes the answer to the question: how does a given import
+/// statement satisfy an `ImportRequest`? This encodes the different
```
AlexWaygood:
```suggestion
/// statement satisfy an [`ImportRequest`]? This encodes the different
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+        alias: &'ast ast::Alias,
+    },
+    /// The necessary module is imported via `from module import ...`,
+    /// but the desired symbol is not listed in `...`.
+    ///
+    /// This always corresponds to a `from <...> import <...>`
+    /// statement.
+    ///
+    /// It is guaranteed that this never contains a glob import.
+    /// (Because otherwise this import wouldn't be partial.)
```
AlexWaygood:
```suggestion
    /// It is guaranteed that this never contains a wildcard import
    /// (otherwise, this import wouldn't be partial).
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+            "\
+from collections import Counter, \\
+  OrderedDict
+<CURSOR>
+        ",
+        );
+        // N.B. The snapshot here has a `/` instead of
+        // a `\`, but this is because of an Insta filter
+        // we have enabled. (I'm unsure of its original
+        // motivation.) ---AG
```
AlexWaygood: the motivation there is almost certainly so that a snapshot that includes paths in some ways will pass on both Unix and Windows systems. But it feels like we probably shouldn't have that filter enabled on this snapshot assertion?

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    #[test]
+    fn out_of_scope_ordering_top_level() {
+        let test = cursor_test(
+            "\
+<CURSOR>
+from collections import defaultdict
+        ",
+        );
+        // Since the import came after the cursor,
+        // we add another import at the top-level
```
AlexWaygood, line 1240:
```suggestion
        // we add another import at the top level
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    fn out_of_scope_ordering_within_function_add_import() {
+        let test = cursor_test(
+            "\
+def foo():
+    <CURSOR>
+from collections import defaultdict
+        ",
+        );
+        // Since the import came after the cursor,
+        // we add another import at the top-level
```
AlexWaygood, line 1266:
```suggestion
        // we add another import at the top level
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// At time of writing (2025-09-15), we just insert a
+    /// qualified import anyway, even though this will result
+    /// in what is likely incorrect code. This seems better
+    /// than some alternatives:
+    ///
+    /// 1. Silently do nothing.
+    /// 2. Silently omit the symbol from completions.
+    ///
+    /// I think it would perhaps be ideal if we could somehow
+    /// prompt the user for what they wany to do. But I think
```
AlexWaygood:
```suggestion
    /// prompt the user for what they want to do. But I think
```

crates/ty_ide/src/lib.rs
```diff
@@ -350,11 +355,17 @@ mod tests {
         }
     }
 
-    /// The file and offset into that file containing
-    /// a `<CURSOR>` marker.
+    /// The file, offset into that file containing a `<CURSOR>` marker.
```
AlexWaygood:
```suggestion
    /// The file, and the offset into that file where the `<CURSOR>` marker is located.
```

crates/ty_python_semantic/src/semantic_model.rs
```diff
@@ -246,6 +267,32 @@ impl<'db> SemanticModel<'db> {
...
         completions
     }
+
+    fn scope(&self, node: ast::AnyNodeRef<'_>) -> Option<FileScopeId> {
+        let index = semantic_index(self.db, self.file);
+
+        // TODO: We currently use `try_expression_scope_id` here as a hotfix for [1].
+        // Revert this to use `expression_scope_id` once a proper fix is in place.
+        //
+        // [1] https://github.com/astral-sh/ty/issues/572
```
AlexWaygood: As @carljm said in https://github.com/astral-sh/ruff/pull/18455#discussion_r2126833137, I think we should just remove this TODO. Using a non-panicking method here seems better in the long term; I don't think we should ever change this.

```suggestion
```

crates/ruff_python_importer/src/insertion.rs
```diff
@@ -135,6 +135,57 @@ impl<'a> Insertion<'a> {
...
+    ///
+    /// from collections import Counter
+    ///
+    ///
+    /// def foo():
+    ///     pass
+    /// ```
+    ///
+    /// The insertion returned will begin after `Counter` but before the
+    /// newline terminator. Callers can then call `Insertion::into_edit`
```
AlexWaygood:
```suggestion
    /// newline terminator. Callers can then call [`Insertion::into_edit`]
```

crates/ruff_python_importer/src/insertion.rs
```diff
@@ -135,6 +135,57 @@ impl<'a> Insertion<'a> {
...
+            return None;
+        };
+        if let Some(at) = import_from.names.last().map(Ranged::end) {
+            return Some(Insertion::inline(", ", at, ""));
+        }
+        // Our AST can deal with partial `from ... import`
+        // statements, so we might not have any members
+        // yet. In this case, we don't need the comma.
+        //
+        // ... however, unless we can be certain that
```
AlexWaygood, line 166:
```suggestion
        // ...however, unless we can be certain that
```

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self {
+            content: None,
+            range,
+        }
+    }
+
+    /// Creates an edit that replaces the content in `range` with `content`.
+    ///
+    /// When `content` is empty, this is equivalent to calling
+    /// `Edit::deletion` with the given range.
```
AlexWaygood:
```suggestion
    /// [`Edit::deletion`] with the given range.
```

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+                content: Some(content),
+                range,
+            }
+        }
+    }
+
+    /// Creates an edit that inserts `content` at the [`TextSize`] `at`.
+    ///
+    /// When `content` is empty, this is equivalent to calling
+    /// `Edit::deletion` with the given range.
```
AlexWaygood:
```suggestion
    /// [`Edit::deletion`] with the given range.
```

crates/ruff_python_importer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// When this edit is a deletion, then this returns `None`.
+    pub fn into_content(self) -> Option<Box<str>> {
+        self.content
+    }
+
+    /// Returns the range of this edit.
+    pub fn range(&self) -> TextRange {
+        self.range
+    }
```
AlexWaygood: Or you could implement `Ranged` for `Edit`? Then you'd also get `start()` and `end()` methods for free

crates/ty_ide/src/completion.rs
```diff
@@ -2,16 +2,199 @@ use std::cmp::Ordering;
...
+    /// type of all such symbols could be quite expensive.
+    pub ty: Option<Type<'db>>,
+    /// The "kind" of this completion.
+    ///
+    /// When this is set, it takes priority over any kind
+    /// inferred from `ty`.
+    ///
+    /// Usually this is set when `ty` is `None`, since it
+    /// may be cheaper to compute at scale. (e.g., For
+    /// unimported symbol completions.)
```
```suggestion
    /// may be cheaper to compute at scale (e.g., for
    /// unimported symbol completions).
```

crates/ty_ide/src/completion.rs
```diff
@@ -2,16 +2,199 @@ use std::cmp::Ordering;
...
+    ///
+    /// Callers should use `Completion::kind` to get the
```
```suggestion
    /// Callers should use [`Completion::kind`] to get the
```

crates/ty_ide/src/completion.rs
```diff
@@ -2,16 +2,199 @@ use std::cmp::Ordering;
...
+    /// An import statement to insert (or ensure is already
+    /// present) when this completion is selected.
+    pub import: Option<Edit>,
+    /// Whether this suggestion came from builtins or not.
+    ///
+    /// At time of writing (2025-06-26), this information
+    /// doesn't make it into the LSP response. Instead, we
+    /// use it mainly in tests so that we can write less
+    /// noisy tests.
+    pub builtin: bool,
```
AlexWaygood, line 66: Just to double-check: you're using "builtins" here to refer to the `builtins` module/namespace in the stdlib rather than "any stdlib module from our vendored typeshed", right?

(If so, you're using the term correctly, but I've seen it used incorrectly in some places in ty, so I'm just double-checking ;) )

crates/ty_ide/src/completion.rs
```diff
@@ -2,16 +2,199 @@ use std::cmp::Ordering;
...
+        })
+    }
+}
+
+/// The "kind" of a completion.
+///
+/// This is taken directly from the LSP completion specification:
+/// <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind>
+///
+/// The idea here is that `Completion::kind` defines the mapping to this from
```
```suggestion
/// The idea here is that [`Completion::kind`] defines the mapping to this from
```

crates/ty_ide/src/completion.rs
```diff
@@ -33,64 +216,80 @@ pub fn completion<'db>(
...
+    let source = source_text(db, file);
+    let line_index = LineIndex::from_source_text(source.as_str());
+    let stylist = Stylist::from_tokens(parsed.tokens(), source.as_str());
+    let importer = Importer::new(db, &stylist, file, source.as_str(), line_index, parsed);
+    let members = importer.members_in_scope_at(scoped.node, scoped.node.start());
+
+    for symbol in all_symbols(db, typed) {
+        let request =
+            ImportRequest::import_from(symbol.module.name(db).as_str(), &symbol.symbol.name);
+        // FIXME: `all_symbols` doesn't account for glob imports. Since
```
```suggestion
        // FIXME: `all_symbols` doesn't account for wildcard imports. Since
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// The [`Stylist`] for the Python AST.
+    stylist: &'a Stylist<'a>,
+    /// The list of visited, top-level runtime imports in the Python AST.
+    imports: Vec<AstImport<'a>>,
+}
+
+impl<'a> Importer<'a> {
+    /// Create a new importer.
+    ///
+    /// The `Stylist` dictates the code formatting options of any code
```
AlexWaygood:
```suggestion
    /// The [`Stylist`] dictates the code formatting options of any code
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    /// edit (if any) produced by this importer.
+    ///
+    /// The `file` given should correspond to the module that we want
+    /// to insert an import statement into.
+    ///
+    /// The `Locator` is used to get access to the original source
```
AlexWaygood:
```suggestion
    /// The [`Locator`] is used to get access to the original source
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+    }
+
+    /// Builds a set of members in scope at the given AST node and position.
+    ///
+    /// Callers should use this routine to build "in scope members" to be used
+    /// with repeated calls to `Importer::import`. This does some work up-front
+    /// to avoid doing it for every call to `Importer::import`.
+    ///
+    /// In general, `at` should be equivalent to `node.start()` (from the
+    /// `ruff_text_size::Ranged` trait). However, in some cases, identifying
```
AlexWaygood:
```suggestion
    /// [`ruff_text_size::Ranged`] trait). However, in some cases, identifying
```

crates/ty_ide/src/importer.rs
```diff
@@ -0,0 +1,1727 @@
...
+impl ImportAction {
+    /// Returns an edit to insert an import statement.
+    pub(crate) fn import(&self) -> Option<&Edit> {
+        self.import.as_ref()
+    }
+
+    /// Returns the symbol text that should be used.
+    ///
+    /// Usually this is identical to the symbol text given to the corresponding
+    /// `ImportRequest`, but this may sometimes be fully qualified based on
```
AlexWaygood:
```suggestion
    /// [`ImportRequest`], but this may sometimes be fully qualified based on
```

crates/ty_python_semantic/src/module_name.rs
```diff
@@ -273,7 +273,14 @@ impl ModuleName {
         std::iter::successors(Some(self.clone()), Self::parent)
     }
 
-    pub(crate) fn from_import_statement<'db>(
+    /// Extracts a module name from the AST of a `from <module> import ...`
+    /// statement.
+    ///
+    /// `importing_file` must be the `File` that contains the import
```
AlexWaygood:
```suggestion
    /// `importing_file` must be the [`File`] that contains the import
```

crates/ruff_diagnostics/src/edit.rs
BurntSushi: Nope, just completely missed the `Ranged` impl. :P

crates/ruff_source_file/src/locator.rs
BurntSushi: Yeah I noticed this. `Locator` struck me as odd for the reasons you cite. But I was trying to find a balance between "make progress" and "refactor." :P 

The only reason I'm using it is because `Insertion` uses it.

But I think you're right. It looks like `Insertion` isn't using any of the methods that use the `LineIndex`. If not, I can refactor that.

BurntSushi: Done! I also moved `Locator` back into `ruff_linter`. _And_ I removed `Locator` from `ruff_linter`'s `Importer`.

crates/ruff_python_importer/src/lib.rs
BurntSushi: I guess that's fine with me. I was trying to avoid bringing `ruff_diagnostics` into the ty world. And the type is very simple, so I don't think duplicating it is terrible. Moreover, I feel like `ruff_python_importer` should probably own the types that it returns, but that's a vague API design nit. But this is done now.

crates/ruff_python_codegen/src/stylist.rs
BurntSushi: It's a little more subtle than that. An `Importer` wants a `&Stylist`. So if we create a `Stylist` on-demand, it can't be tied to the stack frame that creates an `Importer` _and_ returns that importer.

There are a lot of different ways around this. A `Stylist` could always own its source for example. Or the tests could always be required to create an `Importer` in the same stack frame (or above) as where it's used.

But I found this to be simple and it follows a standard API pattern in Rust libraries for creating borrowed/owned versions of the same type. And I don't think the cost here is that big.

crates/ty_ide/src/all_symbols.rs
BurntSushi: Yeah I kept both to avoid the unwraps and because I perceived there to be little-to-no cost in doing so. And if and when we ever refactor this to use namespace packages, the types will help a bit more I think since it will force us to confront the fact that a `File` may be missing.

crates/ty_wasm/src/lib.rs
BurntSushi: Yes! Done.

crates/ty_ide/src/importer.rs
BurntSushi: Yeah that is actually the same instinct that led me to _not_ try to add it to the semantic index. Here, I was only talking about doing so _if_ for some reason this was needed more broadly. But yeah, I can't think of why we would need it more broadly.

crates/ty_ide/src/importer.rs
BurntSushi: Yeah I wrote this comment when I thought I wouldn't be able to do that. Fixed now. :-)

crates/ty_ide/src/importer.rs
BurntSushi: Yup, done. It was already moved (with some edits), but I forgot to delete it here.

crates/ty_ide/src/importer.rs
BurntSushi: Blech. Done.

crates/ty_ide/src/importer.rs
BurntSushi: Yeah good call out! I thought about this when working on the importer, but the "how do you know when it should go into a `TYPE_CHECKING` block" seemed like it was big enough to be worth punting on for now. I've added this to my notes for completions.

crates/ty_ide/src/importer.rs
BurntSushi: All righty. Seems sensible. Done.

crates/ty_ide/src/importer.rs
BurntSushi: Yeah I was talking about that with Zanie. It's definitely not clear to me that aliasing would be better... But we can wait for user feedback on this one. Either way, I think if we did want to go the alias route, it wouldn't be hard to bolt that on to what I have here.

crates/ty_python_semantic/src/semantic_model.rs
BurntSushi: Sure, done!

crates/ty_ide/src/importer.rs
BurntSushi: Hmmm I prefer the space after the `...` here. :-)

crates/ty_ide/src/importer.rs
BurntSushi: I meant "definitively" here.

crates/ty_ide/src/importer.rs
BurntSushi: Hmmm. The filter here is

```
insta_settings.add_filter(r#"\\(\w\w|\s|\.|")"#, "/$1");
```

That `\s` is what threw me off here, since I don't think that usually helps with file paths. I'd guess that was probably copied from elsewhere.

In any case, just removing the `\s` alternation fixed things and makes this snapshot read better.

crates/ty_ide/src/importer.rs
BurntSushi: I think if `low-level` is supposed to be hyphenated then `top-level` ought to be as well.

crates/ty_python_semantic/src/semantic_model.rs
BurntSushi: Fair. Done!

crates/ruff_python_importer/src/lib.rs
BurntSushi: Yeah I ended up just deleting this type.

crates/ty_ide/src/completion.rs
BurntSushi: I believe so, yes. Specifically, however ty uses "builtin." `builtin` here is `true` when `module.is_known(self.db, KnownModule::Builtins)` is `true`.

crates/ruff_python_importer/src/insertion.rs
BurntSushi: Done! I figured I was forgetting something obvious. Perhaps surprising, my implementation passes all the tests. (In the sense that it still generates valid Python source, although I wouldn't call the formatting ideal. Although perhaps some of your examples don't have ideal formatting in the first place.)

BurntSushi:
Going to bring this in, but I'm happy to address additional feedback in follow-ups. :-)

crates/ty_ide/src/importer.rs
AlexWaygood: Unfortunately whether either "top-level" or "low-level" should be hyphenated or not depends on the context of whether it's being used as a compound adjective (but this is definitely not worth worrying about :-)

crates/ty_python_semantic/src/types/ide_support.rs
sharkdp: I only added the `single_declaration` field in the return type of `place_from_declarations` recently. We could probably do something similar and add a `single_binding` or `single_definition` field in the return type of `place_from_bindings`, so as to avoid the additional traversal here?

A case where this (and the `single_declaration` / `single_binding` fields) would fail to return a single definition would be:
```py
if flag:
    x = 1
else:
    x = 2

# there are two live bindings of `x` at the end of the scope here, each with its own definition
```

I haven't thought through the implications here (I don't know what we use the definition field for, downstream), but it's maybe worth at least adding some tests for?

sharkdp: *Somewhat* related discussion that describes another shortcoming of our `place_from_{declarations,bindings}` API: https://github.com/astral-sh/ty/issues/1051

BurntSushi: Yeah https://github.com/astral-sh/ty/issues/1051 looks very wise here.

I explored this a bit by adding several test cases in this PR: https://github.com/astral-sh/ruff/pull/20467

rushter:
> Demo:
> 
>  even-slicker-import-demo.mp4

What lsp feature displays the import path (green text)?

In helix, I just see suggested names without imports. They are from different modules and there is no way to tell which is correct.

image

BurntSushi:
We put it in [`CompletionItemLabelDetails.detail`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails). It does require LSP 3.17 though.

rushter:
> We put it in [`CompletionItemLabelDetails.detail`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails). It does require LSP 3.17 though.

Weird, I think they do support it. I tested ty on `0.0.1-alpha.21`, not from GitHub branch.

https://github.com/helix-editor/helix/blob/0ae37dc52ba715100893c327414bcb1a1924a4c3/helix-lsp-types/src/completion.rs#L559


I can see them in other LSPs.

image

rushter:
I'm getting the same behavior in Zed as well. Will test it more tomorrow.

image

BurntSushi:
> Weird, I think they do support it. I tested ty on `0.0.1-alpha.21`, not from GitHub branch.

Hmmm, [this](https://github.com/astral-sh/ruff/pull/20466) might be related and I don't think made it into `0.0.1-alpha.21`? Although that shouldn't prevent it from showing up at all.

> Will test it more tomorrow.

Thank you!!! <3

> I can see them in other LSPs.

Hmmm that's not showing the import path though as far as I can see. It's just showing the kind. Although ty should be including the kind too, and that doesn't seem to be showing up for you either.

rushter:
> > Weird, I think they do support it. I tested ty on `0.0.1-alpha.21`, not from GitHub branch.
> 
> Hmmm, [this](https://github.com/astral-sh/ruff/pull/20466) might be related and I don't think made it into `0.0.1-alpha.21`? Although that shouldn't prevent it from showing up at all.
> 
> > Will test it more tomorrow.
> 
> Thank you!!! <3
> 
> > I can see them in other LSPs.
> 
> Hmmm that's not showing the import path though as far as I can see. It's just showing the kind. Although ty should be including the kind too, and that doesn't seem to be showing up for you either.

Installed from the main branch. Here is what I see in LSP logs of helix:

```
 {
      "additionalTextEdits": [
        {
          "newText": "from selectolax.parser import HTMLParser\n",
          "range": { "end": { "character": 0, "line": 7 }, "start": { "character": 0, "line": 7 } }
        }
      ],
      "insertText": "HTMLParser",
      "kind": 7,
      "label": "HTMLParser",
      "labelDetails": { "detail": " (import selectolax.parser)" },
      "sortText": " 56"
    },
    {
      "additionalTextEdits": [
        {
          "newText": "from html.parser import HTMLParser\n",
          "range": { "end": { "character": 0, "line": 7 }, "start": { "character": 0, "line": 7 } }
        }
      ],
      "insertText": "HTMLParser",
      "kind": 7,
      "label": "HTMLParser",
      "labelDetails": { "detail": " (import html.parser)" },
      "sortText": " 57"
    },
```
    
   So I guess label details are not fully supported. Not sure why kind does not work though...

rushter:
Checked how it's done in other LSP.s 
Rust analyzer just appends it to label:
```
        "label": "PatternMatchAs(use ruff_python_ast::PatternMatchAs)",
```
gopls puts it in the description (if it's a module import).:

```
"label": "unsafeheader",
"kind": 9,
 "detail": "\"internal/unsafeheader\"",
```


It looks like even Zed does not support labelDetails yet, so perhaps it's better to put it somewhere else for now.

BurntSushi:
Oh nice find! I missed that in r-a. I think you're right that stuffing it somewhere else if the LSP doesn't support `labelDetails` would be wise.

BurntSushi:
@rushter I didn't forget about your bug! It should be fixed in https://github.com/astral-sh/ruff/pull/21109 :-)

rushter:
> @rushter I didn't forget about your bug! It should be fixed in #21109 :-)

Thanks! I was checking the recent ty changes yesterday for autocomplete improvements :)

PR #20441, opened 2025-09-16
Teach TY to search PYTHONPATH to find modules

mmlb:
This brings ty in line with other type checkers and python interpreters, including on by default and no options to disable. Can be disabled by clearing the PYTHONPATH environment variable.

## Summary

Prepend PYTHONPATH to extra_paths so ty behaves similar to other type checkers and the interpreters. Fixes https://github.com/astral-sh/ty/issues/547

## Test Plan

I built ty locally and ran it in nix-shell environment that sets PYTHONPATH, without this PR ty errors with `unresolved-attribute`.


mmlb:
This is modification of #20016, I dropped the namespace package logic and opt-in as requested. Instead of opt-out via flag I went with opt-out by clearing PYTHONPATH as I think that makes sense but would be happy to add an explicit cli arg. Thanks to @natelust for the initial code/PR.

natelust:
Thanks for picking this up, I just have been underwater dealing with things lately, I really appreciate it.

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +278,36 @@ impl Options {
...
 
-        let settings = SearchPathSettings {
-            extra_paths: environment
+        // collect the existing site packages
+        let mut extra_paths: Vec<SystemPathBuf> = Vec::new();
+
+        // read all the paths off the PYTHONPATH environment variable, check
+        // they exists, and add them to the vec of extra_paths as they should
+        // be checked before site-packages just like python interpreter does
+        if let Ok(python_path) = env::var("PYTHONPATH") {
```
MichaReiser: As reported by clippy, we should use `system::env_var` here instead. This ensures proper test isolation.

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +278,36 @@ impl Options {
...
+            let splits: Vec<SystemPathBuf> = python_path
+                .split(':')
+                .filter_map(|str_path| {
+                    let possible_path = std::path::PathBuf::from(str_path);
+                    match possible_path.try_exists() {
+                        Ok(true) => SystemPathBuf::from_path_buf(possible_path).ok(),
+                        _ => None,
+                    }
```
MichaReiser:
```suggestion
                .filter_map(|str_path| {
                    let possible_path = SystemPathBuf::from(str_path);
                    system.path_exists(&possible_path).then_some(possible_path)
                })
```

MichaReiser: Same here, we should use `system.path_exists` over `path.exists` to properly support in-memory file systems and WASM.  Or even better, use `is_directory` because adding a file would later fail

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +278,36 @@ impl Options {
...
+                })
+                .collect();
+            extra_paths.extend(splits);
```
MichaReiser: It's not really necessary to collect before calling `extend`. You can pass the iterator directly. However, as mentioned by @carljm on the issue, we should inform users that ty's adding paths from `PYTHONPATH`.

That's why I would rewrite this as:

```rust
for path in python_path.split(':') {
	let possible_path = SystemPathBuf::from(str_path);

	if system.path_exists(&possible_path) {
		tracing::debug!("Adding `{possible_path}` from the `PYTHONPATH` environment variable to `extra_paths`);
		extra_paths.push(possible_path);
	} else {
		tracing::debug!("Skipping `{possible_path}` listed in `PYTHONPATH` because the path doesn't exist)"
	}
}
```

We may even want

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +278,36 @@ impl Options {
             roots
         };
 
-        let settings = SearchPathSettings {
-            extra_paths: environment
+        // collect the existing site packages
+        let mut extra_paths: Vec<SystemPathBuf> = Vec::new();
```
MichaReiser, line 299: I think we should only add `PYTHONPATH` if the user didn't explicitly set `extra_paths` (when `extra_paths` is `None`).

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +278,36 @@ impl Options {
...
                 .extra_paths
                 .as_deref()
                 .unwrap_or_default()
                 .iter()
-                .map(|path| path.absolute(project_root, system))
-                .collect(),
+                .map(|path| path.absolute(project_root, system)),
+        );
+
+        let settings = SearchPathSettings {
```
MichaReiser, line 331: @AlexWaygood any thoughts on whether we should move this logic into `SearchPaths::from_settings`?

MichaReiser (changes requested):
Thank you. I left a few comments. We should also add a CLI test for this in https://github.com/astral-sh/ruff/blob/1d2128f918a2315a5fc81042b5417f9d8cf34282/crates/ty/tests/cli/python_environment.rs#L9

crates/ty_project/src/metadata/options.rs
carljm: I'm not sure about this. I think I would find it surprising for extra paths option on CLI to override `PYTHONPATH` entirely, since they aren't named the same. I think if we want a way to disable respecting `PYTHONPATH` it should be its own separate option.  There may be use cases for respecting `PYTHONPATH` and setting unrelated additional extra paths in config or CLI.

crates/ty_project/src/metadata/options.rs
mmlb: fixed

crates/ty_project/src/metadata/options.rs
mmlb: fixed, also converted to `is_directory`

crates/ty_project/src/metadata/options.rs
mmlb: you cut off there at the end, but I did grab from your re-write.

crates/ty_project/src/metadata/options.rs
mmlb: fixed, I also added a log message for when both --extra-paths and PYTHONPATH are set

mmlb:
> Thank you. I left a few comments. We should also add a CLI test for this in
> 
> https://github.com/astral-sh/ruff/blob/1d2128f918a2315a5fc81042b5417f9d8cf34282/crates/ty/tests/cli/python_environment.rs#L9

I'll take a look at this and give it a go.

crates/ty_project/src/metadata/options.rs
AlexWaygood: Hmm, I agree with @carljm though -- I'm not sure I understand why we should ignore `PYTHONPATH` just because `extra-_paths` has been set

crates/ty_project/src/metadata/options.rs
AlexWaygood: No strong opinion really!

crates/ty_project/src/metadata/options.rs
mmlb: Hmm yeah I can see that. I was going off of cli-args > env vars but it can be confusing if the names don't match. I think the intention was that PYTHONPATH and --extra-paths both set extra_paths. I can see it either way, though I think I do prefer avoiding the implentation leakage and instead just allow both PYTHONPATH and --extra-paths, the `extra` part is pretty telling.

carljm: Yes -- for me, using `extra-paths` internally to implement respecting `PYTHONPATH` is an implementation detail, just because the behavior of the paths happens to be the same. It doesn't mean that `PYTHONPATH` is simply an alias of the extra-paths configuration; they are separate features.

mmlb: I think I'll go back to respecting both PYTHONPATH and --extra-paths, PYTHONPATH first --extra-paths second. I want to continue avoiding adding an argument to ignore PYTHONPATH, I think it'll be confusing and you can wait until someone shows a need for it.

mmlb: Back to taking both PYTHONPATH and --extra-paths.

crates/ty_project/src/metadata/options.rs
```diff
@@ -277,14 +277,41 @@ impl Options {
...
+        // interpreter does
+        if let Ok(python_path) = system.env_var("PYTHONPATH") {
+            for path in python_path.split(':') {
+                let possible_path = SystemPathBuf::from(path);
+
+                if system.is_directory(&possible_path) {
+                    tracing::debug!(
+                        "Adding `{possible_path}` from the `PYTHONPATH` environment variable to `extra_paths`"
+                    );
+                    extra_paths.push(possible_path);
```
MichaReiser, line 313: Let's make `possible_path` absolute here (or on line 289), similar to what we do on line 310.

mmlb: done

MichaReiser:
Thank you. This looks good now and we can merge it once there's a CLI test.

mmlb:
> Thank you. This looks good now and we can merge it once there's a CLI test.

Done!

MichaReiser (approved):
Perfect, thank you

Issue #20411, opened 2025-09-15
Unfork annotate-snippets


MichaReiser:
The initial reason for forking annotate-snippets was that it didn't support [omitting a header](https://github.com/rust-lang/annotate-snippets-rs/issues/167). This seems to be supported with 0.12 or newer. 

However, we've since made new changes. We should explore if we can migrate back to the upstream version.

pajod:
Using 0.12.0 or above might also fix `␀` rendering in console output.. at least [changelog](https://github.com/rust-lang/annotate-snippets-rs/blob/master/CHANGELOG.md#0120---2025-08-28) promises "Various rendering fixes".

Issue #14452, opened 2024-11-19
setting output format hides output in ruff check


kaddkaka:
When running a `ruff check --fix-only` there is no output about what errors were fixed. This surprised me, I thought it would still print the fixed error since the [documentation about `--fix-only`](https://docs.astral.sh/ruff/settings/#fix-only) says "Like [fix](https://docs.astral.sh/ruff/settings/#fix), but disables reporting on leftover violation. Implies [fix](https://docs.astral.sh/ruff/settings/#fix)."

Maybe supplying `--show-fixes` will fix it? Yes,

`ruff check --fix-only --show-fixes` shows the fixed errors, great!

Can we also get this info in json output format? No,

`ruff check --fix-only --show-fixes --output-format json`  print nothing! `--output-format junit` has the same issue.

Ruff version: 0.7.0

zanieb:
```
❯ ruff check example.py --select I
example.py:1:1: I001 [*] Import block is un-sorted or un-formatted
  |
1 | / import b, a
  |
  = help: Organize imports

Found 1 error.
[*] 1 fixable with the `--fix` option.

❯ ruff check example.py --select I --fix-only
Fixed 1 error.

❯ echo "import b, a" > example.py
❯ ruff check example.py --select I --fix
Found 1 error (1 fixed, 0 remaining).

❯ ruff check example.py --select I --fix-only --output-format json

❯ echo "import b, a" > example.py
❯ ruff check example.py --select I --fix --output-format json
[]%                        

❯ echo "import b, a" > example.py
❯ ruff check example.py --select I --fix --output-format json --show-fixes
[]%                          

❯ echo "import b, a" > example.py
❯ ruff check example.py --select I --fix-only --show-fixes

Fixed 1 error:
- example.py:
    1 × I001 (unsorted-imports)

Fixed 1 error.                                                                                                                                  
```

I don't think this has to do with `--fix-only`, i.e., it applies equally to `--fix`. I think we should improve the JSON output here, though I don't think it should rely on the `--show-fixes` flag.

MichaReiser:
I suspect that Ruff throws away the diagnostics after each fix cycle. We would need to preserve them and pass them through to the json emitter. This likely requires some special casing in the CLI because we don't want to emit the diagnostics for all output formats (which makes the experience somewhat inconsistent)

kaddkaka:
[...]

PR #20443, opened 2025-09-16
[WIP] Use `Diagnostic`s for rendering formatting results

ntBre:
## Summary

This is a rough draft/prototype of using the new `Diagnostic` type for rendering formatter diagnostics. This allows the formatter to inherit all of the output formats already implemented in the linter and ty. For example, here's the new `full` output format, with the formatting diff displayed using the same infrastructure as the linter:

image

There are several limitiations/todos here still, especially around the `OutputFormat` type:
- A few literal `todo!`s for the remaining `OutputFormat`s without matching `DiagnosticFormat`s
- The default output format is `full` instead of something more concise like the current output
- Some of the output formats (namely JSON) have information that doesn't make much sense for these diagnostics

A solution to all of these could just be using a separate `OutputFormat` type from the one in the linter. I reused that for now just for convenience.

More broadly, it seems pretty expensive to add the entire file's contents as an `Edit`. I'm guessing that might show up in the benchmarks on this PR. On a related note, this is a pretty shallow conversion, only constructing the `Diagnostic`s right before rendering. There might be a better way to use the new infrastructure more. We could also use them for rendering `FormatCommandError`s, not just `FormatPathResult`s.

## Test Plan

Some new CLI tests and manual testing. Another TODO is adding something like the `output_format` test for the linter with snapshots for all of the supported output formats.


MichaReiser:
> More broadly, it seems pretty expensive to add the entire file's contents as an Edit. I'm guessing that might show up in the benchmarks on this PR. On a related note, this is a pretty shallow conversion, only constructing the Diagnostics right before rendering. There might be a better way to use the new infrastructure more. We could also use them for rendering FormatCommandErrors, not just FormatPathResults.

A solution here could be to add a `diff` field to `Diagnostic` so that the "edit" is computed lazily or the diff is rendered directly instead of using the edit rendering.


> The default output format is full instead of something more concise like the current output

Hmm, that's an interesting find. This needs some design work.

ntBre:
I think this is mostly ready for review now. I did a lot of squashing today to make it easier to review commit-by-commit. 

The first 7 commits are all very small, standalone refactors used in later steps. The 8th commit is the biggest, converting `FormatPathResult`s into diagnostics. The 9th commit is an incremental improvement on that to split notebook `Edit`s by cell so that they can actually be rendered in the `full` output format (otherwise they never satisfy [this check](https://github.com/astral-sh/ruff/blob/main/crates/ruff_db/src/diagnostic/render/full.rs#L141)). Finally, the last commit emits errors as diagnostics too. 

I still have a bunch of TODO comments on the errors. I did my best to match up the error variants with `DiagnosticId`s, including adding two new `DiagnosticId`s, but I think there's still room for improvement. I should probably also add a test for some or all of these.

I haven't tried to truncate the lines in the diff yet either. I know that came up in the design discussion, so I'm happy to tackle it now or leave it for a follow up if desired.

Hopefully it's not too glaring, but in the screenshot in the summary you can see that the `|` for line numbers doesn't line up with the middle of the filename `-->` like it does for lint diagnostics. I don't think there's really a good way around this since the arrow alignment comes from `annotate-snippets` and only gets indented if `annotate-snippets` renders line numbers too. I'm sure we could hack another option into `annotate-snippets`, but it would likely make https://github.com/astral-sh/ruff/issues/20411 harder still.

ntBre:
I'm marking this ready for review. I pushed a few more commits resolving the main issues I noted above:
- I added a `lineno_offset`/`header_offset` field for improving the `-->` alignment:
  
  image

  It can still go wrong and makes it harder to un-fork `annotate-snippets`, so I'm happy to revert this, but I think it's a visual improvement in general.
- I added tests for all of the `FormatCommandError` variants and fixed a couple of oversights in their formatting, including factoring out and using some of ty's panic rendering. There are still TODOs on a few of these, but I think I could use some input on the best `DiagnosticId`s to use. The tests at least show what they look like to help us iterate on them.

I think the performance hit is acceptable. It's only ~10 ms (~5%) for the `concise` output on a large project, and the larger discrepancy between `full` output and `--diff` seems justified given the additional work of computing more granular diffs. I am seeing slightly worse performance today compared to the last time I ran the benchmarks, so it may be worth a bit more profiling, but I think the point still stands.

<details><summary>New benchmark table</summary>
<p>

| Command           |     Mean [ms] | Min [ms] | Max [ms] |     Relative |
| :---------------- | ------------: | -------: | -------: | -----------: |
| `stable`          |   200.0 ± 2.3 |    196.6 |    203.5 |         1.00 |
| `preview-full`    | 9134.8 ± 32.1 |   9108.0 |   9197.5 | 45.67 ± 0.55 |
| `preview-concise` |   232.6 ± 4.2 |    227.0 |    240.7 |  1.16 ± 0.02 |
| `stable-diff`     | 3354.8 ± 15.9 |   3333.7 |   3383.6 | 16.77 ± 0.21 |

`stable` is basically identical to last time, but both `preview` versions are ~20 ms slower. `--diff` is almost 50 ms slower, which really doesn't make sense.

</p>
</details> 

Similarly, I think we may want to truncate large diffs at some point (Zanie even mentioned possibly making the limit configurable), but I think we can hold off on that for now.

Oh, one other TODO is that it would be nice to emit a warning if the `output-format` is set without `preview`, but I don't think that's so easy to do since `output-format` is a global option and will have already been `unwrap_or_default`ed by the time we see it.

crates/ruff/src/args.rs
```diff
@@ -537,6 +537,11 @@ pub struct FormatCommand {
     /// Exit with a non-zero status code if any files were modified via format, even if all files were formatted successfully.
     #[arg(long, help_heading = "Miscellaneous", alias = "exit-non-zero-on-fix")]
     pub exit_non_zero_on_format: bool,
+
+    /// Output serialization format for violations.
+    /// The default serialization format is "full".
```
MichaReiser: We should document that this option is only respected in preview mode. Ideally, we'd emit a warning message if it is set when not using preview

crates/ruff/src/commands/format.rs
```diff
@@ -321,15 +350,24 @@ pub(crate) fn format_path(
 #[derive(Debug)]
 pub(crate) enum FormattedSource {
     /// The source was formatted, and the [`SourceKind`] contains the transformed source code.
-    Formatted(SourceKind),
+    Formatted {
+        formatted: SourceKind,
+        unformatted: SourceKind,
```
MichaReiser: It seems the only reason that we need to add `unformatted` here is for the `From<FormattedSource> for FormatResult` implementation. This seems unfortunate because we do have an owned `unformatted` in all code paths where we convert from `FormattedSource` to `FormatResult`. 

I suggest removing `unformatted` here and making `From<FormattedSource>` a method on `FormattedSource` instead:

```rust
impl FormattedSource {
	fn into_format_result(unformatted: SourceKind) -> FormatResult { ... }
```

crates/ruff/src/commands/format.rs
```diff
@@ -478,7 +520,10 @@ pub(crate) fn format_source(
 #[derive(Debug, Clone, is_macro::Is)]
 pub(crate) enum FormatResult {
     /// The file was formatted.
-    Formatted,
+    Formatted {
+        unformatted: SourceKind,
+        formatted: SourceKind,
+    },
```
MichaReiser: Do we still need both `Formatted` and `Diff`, given that they now represent the same?

This makes me wonder: Do your changes regress the performance of `format` (without `--check` or `--diff`). I suspect it could because we're now always capturing `formatted` and `unformatted` even if the changes were written right to disk. 

I think we should keep both those variants but use `Diff` when using `--check` but keep using `Formatted` when using a bare `format` command.

crates/ruff/src/commands/format.rs
```diff
@@ -566,14 +611,106 @@ impl<'a> FormatResults<'a> {
         Ok(())
     }
 
+    /// Write a list of the files that would be changed and any errors to the given writer.
+    ///
+    /// Errors are reported first in the order they are provided, followed by the remaining
+    /// diagnostics sorted by file name.
```
MichaReiser: This seems a bit surprising to me. I saw that `self. diagnostics` does sort the diagnostics. What's the reason that we don't sort the entire `diagnostics` instead?

crates/ruff/src/commands/format.rs
```diff
@@ -566,14 +611,106 @@ impl<'a> FormatResults<'a> {
...
+            }
+            OutputFormat::Full => {
+                let config = config
+                    .format(DiagnosticFormat::Full)
+                    .hide_severity(true)
+                    .show_fix_diff(true)
+                    .color(!cfg!(test) && colored::control::SHOULD_COLORIZE.should_colorize());
+                let value = DisplayDiagnostics::new(&context, &config, &diagnostics);
+                write!(f, "{value}")?;
+            }
```
MichaReiser: Nit: It seems unfortunate that we now have the same match statement in the linter and formatter. What makes it challenging to share this match statement?

crates/ruff/src/commands/format.rs
```diff
@@ -664,6 +889,91 @@ impl FormatCommandError {
...
+            FormatCommandError::Parse(display_parse_error) => Diagnostic::new(
+                DiagnosticId::InvalidSyntax,
+                Severity::Error,
+                display_parse_error,
+            ),
+            FormatCommandError::Panic(path, panic_error) => {
+                let mut diagnostic = Diagnostic::new(
+                    DiagnosticId::Panic,
+                    Severity::Fatal,
+                    panic_error.to_diagnostic_message(path.as_ref().map(|path| path.display())),
```
MichaReiser: Nit: Could we unify some of the code here with what we have in the linter?

crates/ruff/src/commands/format.rs
```diff
@@ -664,6 +889,91 @@ impl FormatCommandError {
...
+            annotation
+        });
+
+        let mut diagnostic = match error {
+            // TODO(brent) also not sure about this one. Most of the ignore::Error variants
+            // do fit pretty well under Io errors, but we could match and be even more
+            // specific if we wanted. We already have a `DiagnosticId::InvalidGlob`, which
+            // could map to Error::Glob, for example.
+            FormatCommandError::Ignore(error) => {
+                Diagnostic::new(DiagnosticId::Io, Severity::Error, error)
```
MichaReiser, line 868: Using IO here is probably fine and roughly matches what we do in the lint command. 

We match on all `ignore::Error` in ty and:

* Log a warning for any gitignore related errors but keep going (unlike Ruff)
* Create IO errors for all other errors

https://github.com/astral-sh/ruff/blob/959f55ae057843f8b3fe59a23577f5ae3948a903/crates/ruff_db/src/system/os.rs#L527-L533

What's special about the ignore errors is that the path isn't guaranteed to be the current file path. E.g. the error could be related to the gitignore file in the same or even a parent directory.

crates/ruff/src/commands/format.rs
```diff
@@ -664,6 +889,91 @@ impl FormatCommandError {
...
+                    &parse_error.error,
+                ),
+                FormatModuleError::FormatError(format_error) => {
+                    Diagnostic::new(DiagnosticId::FormatError, Severity::Error, format_error)
+                }
+                // TODO(brent) this might need a new DiagnosticId, or we could reuse
+                // `FormatError`, which also has an InvalidDocument variant, like PrintError
+                FormatModuleError::PrintError(print_error) => {
+                    Diagnostic::new(DiagnosticId::InvalidSyntax, Severity::Error, print_error)
+                }
```
MichaReiser, line 893: `FormatError` would be more correct as I consider this a formatter bug (The constructed IR is invalid).

crates/ruff/src/commands/format.rs
```diff
@@ -664,6 +889,91 @@ impl FormatCommandError {
...
+            // TODO(brent) also not sure about this one. Most of the ignore::Error variants
+            // do fit pretty well under Io errors, but we could match and be even more
+            // specific if we wanted. We already have a `DiagnosticId::InvalidGlob`, which
+            // could map to Error::Glob, for example.
+            FormatCommandError::Ignore(error) => {
+                Diagnostic::new(DiagnosticId::Io, Severity::Error, error)
+            }
+            // TODO(brent) not sure if this is correct, DisplayParseError includes some
+            // color and other formatting. I think we'd rather have access to the `path` and
+            // underlying ParseError directly, like in other variants here.
```
MichaReiser: I agree that using a `ParseError` here directly would be better. How hard is it to change `FormatCommandError::Parse` to store a `ParseError` instead ?

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +765,94 @@ impl<'a> FormatResults<'a> {
...
+                let mut diagnostic = Diagnostic::new(
+                    DiagnosticId::Unformatted,
+                    Severity::Error,
+                    "File would be reformatted",
+                );
+
+                let path = result.path.to_string_lossy();
+                // For now, report the edit as a replacement of the whole file's contents. For
+                // scripts, this is a single `Edit`, but notebook edits must be split by cell in
+                // order to render them as diffs.
```
MichaReiser: I think it's worth trying to do a little better here and trim the edit to the range that is different. This can be done in `O(n)` by searching the first (or last) character that's different (I'm not sure if we should then move forward/backward to find the start/end of the line or if that doesn't matter). I think that would also remove the need to use `set_file_level(true)` and for `set_header_offset`?

While not as good as separate Edits, it should already narrow the `Edits significantly in practice.

crates/ruff/src/commands/format.rs
```diff
@@ -664,6 +889,91 @@ impl FormatCommandError {
...
+            }
+
+            // I/O errors
+            FormatCommandError::Read(_, source_error)
+            | FormatCommandError::Write(_, source_error) => {
+                Diagnostic::new(DiagnosticId::Io, Severity::Error, source_error)
+            }
+            FormatCommandError::Diff(_, error) => {
+                Diagnostic::new(DiagnosticId::Io, Severity::Error, error)
+            }
```
MichaReiser: Can we remove this error? It seems that it is only created if we fail to write the diff to stdout... We've already lost if that happens.

I think we can simply ignore that error or log a warning with tracing. It might also be worth checking what ruff check does.

crates/ruff_db/src/panic.rs
```diff
@@ -31,6 +31,29 @@ impl Payload {
     }
 }
 
+impl PanicError {
+    pub fn to_diagnostic_message(&self, path: Option<impl std::fmt::Display>) -> String {
```
MichaReiser, line 35: Should we use this in ruff linter too?

crates/ruff_formatter/src/diagnostics.rs
```diff
@@ -38,12 +38,12 @@ impl std::fmt::Display for FormatError {
             ),
             FormatError::InvalidDocument(error) => std::write!(
                 fmt,
-                "Invalid document: {error}\n\n This is an internal Rome error. Please report if necessary."
```
MichaReiser, line 41: lol

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +765,94 @@ impl<'a> FormatResults<'a> {
...
+                    let line_count = formatted
+                        .source_code()
+                        .count_lines(TextRange::up_to(formatted.source_code().text_len()));
+                    (fix, line_count)
+                };
+
+                let source_file = SourceFileBuilder::new(path, unformatted.source_code()).finish();
+                let span = Span::from(source_file);
+                let mut annotation = Annotation::primary(span);
+                annotation.set_file_level(true);
```
MichaReiser: Using `set_file_level` here seems odd to me. I understand that the motivation for `set_file_level` was to suppress code snippets for diagnostics starting at 1:1 (or don't have a range).

But we do want to show the code snippet for formatter diagnostics and I believe it's the reason why the `-->` arrow gets misplaced. 

Why do we need `set_file_level` here, could we avoid using it as it seems a misuse?

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1016,6 +1032,17 @@ pub enum DiagnosticId {
 
     /// Use of a deprecated setting.
     DeprecatedSetting,
+
+    /// The code needs to be formatted.
+    Unformatted,
+
+    /// Range formatting isn't supported for notebooks.
+    RangeFormatNotebook,
```
MichaReiser: I don't think it's worth having a diagnostic ID for every CLI option that can possibly be invalid. Maybe `invalid-cli-option`?

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -1016,6 +1032,17 @@ pub enum DiagnosticId {
...
+
+    /// Something went wrong while formatting.
+    ///
+    /// This often indicates a bug in the formatter.
+    FormatError,
```
MichaReiser: Similar to `RangeFormatNotebook`. I wonder if we should just have one `internal-error` code. The main advantage I see is that it makes it very clear that this is a bug in Ruff vs. some user error (we should probalby add a note saying that too)

MichaReiser (commented):
Nice: I've a few suggestions. The part I feel most unsure about is if using `set_file_level` is correct for format diagnostics.

crates/ruff/src/args.rs
ntBre: Yeah, I thought about the warning, but I wasn't sure how easy it would be to add since the output format is a global setting. I can go look again too, but is there a place where we know we're in the `format` sub-command without preview and before unwrapping to a default output format?

It seems a bit easier if we want to check for non-default output formats, at least.

crates/ruff/src/commands/format.rs
ntBre: I was just trying to preserve the stable behavior of emitting errors first, but it makes sense to me to sort them all together here.

crates/ruff/src/args.rs
MichaReiser: I think this is a bit different because this option is specific to the format command and, if set, it should always take precedence over any other configuration option. 

We should be able to check this directly in the format command

crates/ruff/src/commands/format.rs
ntBre: The snippet in your comment is the main difference. In the linter we still use the `TextEmitter` for these output formats and mix in a few other checks in the `match`:

https://github.com/astral-sh/ruff/blob/2af8c53110227598cb4758a777667dbe18bc7f15/crates/ruff/src/printer.rs#L251-L268

We also use at least one different configuration setting in the formatter to always suppress the fix availability icon.

crates/ruff/src/args.rs
ntBre: Ohhh okay, yes a warning for the new CLI option does make sense. I was over-complicating it by trying to check for the configuration option too.

crates/ruff/src/commands/format.rs
ntBre: I wasn't sure that we do want to show the code snippet for these diagnostics as it also requires us to underline/annotate a range in the file. I didn't want to add a multi-line annotation possibly spanning the entire file, but maybe we can underline the first changed line? That could work well with your other suggestion about narrowing the edit range. It does seem slightly redundant to render any code if we're immediately going to render it again in the diff, but I can give it a try.

But yes, you're right. If we set a real range and thus render a snippet we can avoid making these file-level diagnostics, and I can also revert my arrow changes.

crates/ruff/src/commands/format.rs
ntBre: Yes, I think so. I liked how ty rendered panics, so I pulled from there instead, but I can either switch to how the linter constructs panics or port this to the linter too.

crates/ruff/src/commands/format.rs
MichaReiser: Hmm, why do we still need `TextEmitter`? I worry that it will be difficult to understand why some of those differences exist in the future.

crates/ruff/src/commands/format.rs
MichaReiser: Oh, I see. The difference is that we don't actually render a code snippet, we only render a fix (which is something different in annotated snippet?)

And the issue is that the rendering of `-->` isn't aligned with the rendering of the fix.

crates/ruff/src/args.rs
MichaReiser: Yeah, warning on the option seems difficult.

crates/ruff/src/commands/format.rs
ntBre: Yep, the fixes/diffs aren't actually rendered through `annotate-snippets`, they're my best attempt to emulate `annotate-snippets` after we render the rest of the diagnostic. That's why the arrow header can't align itself, the two steps are separate.

crates/ruff/src/commands/format.rs
ntBre: That's a good point, I can look into refactoring the linter calls and then reusing something here.

crates/ruff/src/commands/format.rs
MichaReiser: I see. I think we should rename `set_file_level` to `hide_snippet` or similar to make the distinction clear now that it's used for more than just hiding snippets for file level diagnostics. 

For the alignment. We can probably get it "right" if we compute the line number where the first change is. 

An alternative would be to remove the need for the alignment altogether. Unfortunately, this might be another breaking change except if all we do is change the number of whitespace before after the arrow

ntBre: > I see. I think we should rename `set_file_level` to `hide_snippet` or similar to make the distinction clear now that it's used for more than just hiding snippets for file level diagnostics.

I guess I was still thinking of these conceptually as file-level diagnostics. That's what they are on stable anyway, "File would be reformatted." But I'm happy to rename the setting, `hide_snippet` makes the actual effect more clear for sure.

> For the alignment. We can probably get it "right" if we compute the line number where the first change is.

That's roughly what I'm doing now, except it's a bit better to compute the line number of the _last_ change (currently the end of the file) since that's likely to need a wider alignment than the start. Notebooks are also more complicated since the width is for the per-cell line number.

> An alternative would be to remove the need for the alignment altogether. Unfortunately, this might be another breaking change except if all we do is change the number of whitespace before after the arrow

That's interesting. I do think it looks nice aligning the middle of the arrow with the line number separator, but maybe it's not worth the trouble.

crates/ruff/src/commands/format.rs
ntBre: Hmm that's a good point. They do look mostly the same comparing results from my branch release build and my system ruff 0.13.0 binary, but the branch version is a little slower:

| Command  |   Mean [ms] | Min [ms] | Max [ms] |    Relative |
| :------- | ----------: | -------: | -------: | ----------: |
| `branch` | 213.1 ± 2.5 |    209.7 |    217.4 | 1.00 ± 0.02 |
| `stable` | 212.4 ± 3.5 |    206.4 |    217.6 |        1.00 |

It makes sense to me to reuse `Diff` anyway, though.

<details><summary>hyperfine</summary>
<p>

```shell
hyperfine --ignore-failure --warmup 10 --export-markdown /tmp/table.md \
  --prepare "git restore ."  \
  -n branch -n stable \
  "~/astral/ruff/target/release/ruff format . --no-cache" \
  "ruff format . --no-cache"
```

</p>
</details>

crates/ruff/src/commands/format.rs
ntBre: It looks like reusing the `Diff` variant below also let me revert this!

crates/ruff/src/commands/format.rs
ntBre: Ah that makes sense, thank you! I was wondering if the path might be a bit redundant in both the header and in the diagnostic message, but I didn't realize the two paths could be different. I'll just delete this todo for now then, although it makes sense to me to warn and keep going like in ty at some point.

crates/ruff/src/commands/format.rs
ntBre: `DisplayParseError` does a fair amount of work to compute its `ErrorLocation`, and we still use the `Display` impl in a couple of places. For now I just added an `error` method to retrieve the underlying `ParseError`.

crates/ruff/src/commands/format.rs
ntBre: Sounds good, I had an earlier version that did this before I rebased over it. I'll add it back.

crates/ruff/src/commands/format.rs
ntBre: Oh I think you're right. Can we just reuse the `Write` variant? That's what we do for the other failed stdout writes in the only place this is constructed:

https://github.com/astral-sh/ruff/blob/6b3c493cff01ad1ab51102048d0d4f489b1e1aa4/crates/ruff/src/commands/format_stdin.rs#L111-L116

In the linter I think we just bubble up an exit code that ends up reported here:

https://github.com/astral-sh/ruff/blob/6b3c493cff01ad1ab51102048d0d4f489b1e1aa4/crates/ty/src/main.rs#L28

```shell
$ ruff check try.py > /dev/full
ruff failed
  Cause: No space left on device (os error 28)
$ ruff format --check try.py > /dev/full
ruff failed
  Cause: No space left on device (os error 28)
```

I guess that's effectively the same in the formatter since any of our attempts to write diagnostics will fail.

crates/ruff/src/commands/format.rs
ntBre: I renamed the file-level references to `hide_snippet` and updated my todo about removing it. I also added an expanded new todo about the offset hack for now.

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +739,127 @@ impl<'a> FormatResults<'a> {
...
+    ) -> impl Iterator<Item = Diagnostic> {
+        self.results.iter().filter_map(|result| {
+            let FormatResult::Diff {
+                unformatted,
+                formatted,
+            } = &result.result
+            else {
+                return None;
+            };
+
```
MichaReiser: Should we use a match here that explicitly skips over `Unchanged` and `Skipped`, panics for `Formatted` (at least in debug builds), and returns `unformatted` and `formatted` for diff?

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +739,127 @@ impl<'a> FormatResults<'a> {
...
+                    .zip(formatted.source_code().chars())
+                {
+                    if old != new {
+                        if start.is_none() {
+                            start = Some(offset);
+                        } else {
+                            end = Some(offset);
+                        }
+                    }
+                }
```
MichaReiser: I think we should use this range for the Edit as well if we already computed it anyway. It can help to reduce the size of the edit in common cases where only a few files are incorrectly formatted

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +739,127 @@ impl<'a> FormatResults<'a> {
...
+                ));
+                let line_count = formatted
+                    .source_code()
+                    .count_lines(TextRange::up_to(formatted.source_code().text_len()));
+                (fix, line_count)
+            };
+
+            // Locate the first and last characters that differ to use as the diagnostic
+            // range.
+            let range = {
```
MichaReiser: I think we can do better than traversing the entire file by:

1. Search from the beginning and find the first changed character
2. Slice the remaining document to `&source[start..]`
3. Search from the end to find the last changed character


This is still `O(n)` in the worst case, but should do much better if a large portion of the file changed. It might be worth extracting this logic into a helper function to not distract the reading flow

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +739,127 @@ impl<'a> FormatResults<'a> {
...
+            // is rendered by our fork of `annotate-snippets`, to align with our manually-rendered
+            // diff. `annotate-snippets` computes the alignment of the arrow in the header based on
+            // the maximum line number width in its rendered snippet. However, we don't have a
+            // reasonable range to underline in an annotation, so we don't send `annotate-snippets`
+            // a snippet to measure. If we commit to staying on our fork, a more robust way of
+            // handling this would be to move the diff rendering in
+            // `ruff_db::diagnostic::render::full` into `annotate-snippets`, likely as another
+            // `DisplayLine` variant and update the `lineno_width` calculation in
+            // `DisplayList::fmt`. That would handle this offset "automatically."
+            let lines = OneIndexed::new(line_count as usize).unwrap_or_default();
```
MichaReiser, line 816: We should use the line number of the `Edit`'s end + rendered context window size. The arrow is otherwise over intended if only the first line changed in a 1000 lines file (which I think looks worse than always aligning with two spaces).

crates/ruff/src/commands/format.rs
```diff
@@ -658,12 +889,65 @@ impl FormatCommandError {
...
+            | FormatCommandError::Write(_, source_error) => {
+                Diagnostic::new(DiagnosticId::Io, Severity::Error, source_error)
+            }
+            FormatCommandError::Format(_, format_module_error) => match format_module_error {
+                FormatModuleError::ParseError(parse_error) => Diagnostic::new(
+                    DiagnosticId::InvalidSyntax,
+                    Severity::Error,
+                    &parse_error.error,
+                ),
+                FormatModuleError::FormatError(format_error) => {
```
MichaReiser, line 888: Nit: I'd also classify this as an internal error. At least, how we're using it in the formatter. 

The way we use this error variant is to avoid panicking if we didn't see a specific syntax element when doing some ad-hoc lexing. This means, we incorrectly assumed some invariant that isn't true because the formatter is never called on code with syntax errors. This is always an implementation error -> `InternalError`)

The original intended use case for this error variant comes from biome where the formatter supports formatting code with syntax error. The way it works is that it formats all code without syntax errors. For sections that contain syntax errors, it falls back to preserving the formatting from the source by catching `FormatError::SyntaxError` at the statement level. But even for Biome, a `SyntaxError` should never be propagated passt the statement formatting. If it does, than that's consider an internal error too.

MichaReiser (approved):
I've a few more suggestions about the computed diagnostic and `Edit` range and on how to compute the offset for the arrow.

I still very much dislike the offset hack but I think we can get it "correct" enough outside annotated snippet. However, it might be worth spending like a day to see if we can move the diff rendering to annotated snippets, removign the need for the offset workaround.

crates/ruff/src/commands/format.rs
ntBre: Sure, I was just following the logic in `write_changed`, but a debug assertion makes sense to me.

https://github.com/astral-sh/ruff/blob/00c8851ef8d643434340577f1b1ee65b5e728ac1/crates/ruff/src/commands/format.rs#L555-L559

crates/ruff/src/commands/format.rs
ntBre: Oops, of course, thank you! I found my earlier commit in the reflog but obviously didn't finish integrating it with the other new code.

crates/ruff/src/commands/format.rs
ntBre: Ah good call. This was a bit surprising to me, so I just wanted to note it here: I think the context window size is up to 6 lines. We pass `3` to similar's `grouped_ops` function here:

https://github.com/astral-sh/ruff/blob/1d3e4a91534448121c6db99c2baf83c1ccd2d909/crates/ruff_db/src/diagnostic/render/full.rs#L157-L159

which is used in similar [here](https://github.com/mitsuhiko/similar/blob/a169e29954502d67e26c9e766f2f719e99a52559/src/common.rs#L134-L136):

```rust
            // End the current group and start a new one whenever
            // there is a large range with no changes.
            if len > n * 2 {
```

I guess in a strict sense we should also `min` this with the actual number of lines in the file, so we don't end up with 105 in a 99 line file, for example.

ntBre:
Thank you for the reviews! 

I think this could use one more look. I was over-complicating the range calculations for a while today, but they felt kind of tricky, at least when I was trying to use `zip` like the `start` calculation[^1]. The new `ModifiedRange` type seems to be working well now, though, unless I missed any edge cases.

The other changes seemed relatively straightforward, and I also merged the changes from #20595.

I'll timebox trying to move our diff rendering to annotate-snippets to one day later this week.

[^1]: The tricky part was that zipping and `find`ing the first different character could easily fail if one of the snippets was shorter than the other; at that point it's not clear which one was shorter and caused the failure. We also want an exclusive range, so we've gone one character too far by finding the first different character rather than the last common character. Now I just loop from the end and track the length of the common suffix, which we can subtract from both `text_len`s despite the actual offsets likely being different.

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +670,155 @@ impl<'a> FormatResults<'a> {
...
+                        } else {
+                            None
+                        }
+                    })
+                    .max()
+                    .unwrap_or_default();
+                (fix, line_count)
+            } else {
+                let formatted_code = &formatted.source_code()[modified_range.formatted];
+                let edit = if formatted_code.is_empty() {
```
MichaReiser, line 782: It shouldn't matter so I think it's fine leaving as is but there's the third case where `modified_range.unformatted` is empty (e.g. when adding blank lines between two classes), in which case its an insertion. We could add a `Edit::from_text_and_range(new_text, range)` (with a better name) that does this dance.

crates/ruff/src/commands/format.rs
```diff
@@ -628,6 +670,155 @@ impl<'a> FormatResults<'a> {
...
+            // reasonable range to underline in an annotation, so we don't send `annotate-snippets`
+            // a snippet to measure. If we commit to staying on our fork, a more robust way of
+            // handling this would be to move the diff rendering in
+            // `ruff_db::diagnostic::render::full` into `annotate-snippets`, likely as another
+            // `DisplayLine` variant and update the `lineno_width` calculation in
+            // `DisplayList::fmt`. That would handle this offset "automatically."
+            let line_count = (line_count + CONTEXT_LINES).min(
+                formatted
+                    .source_code()
+                    .count_lines(TextRange::up_to(formatted.source_code().text_len())),
```
MichaReiser, line 814: Is using `formatted` here correct? What if unformatted is shorter?

crates/ruff/src/commands/format.rs
```diff
@@ -792,6 +1018,56 @@ impl Display for FormatCommandError {
...
+
+        let start = unformatted
+            .char_indices()
+            .zip(formatted.chars())
+            .find_map(|((offset, old), new)| {
+                (old != new).then_some(TextSize::try_from(offset).unwrap())
+            })
+            // Fall back on the shorter text length if one of the strings is a strict prefix of the
+            // other (i.e. the zip iterator ended before finding a difference).
+            .unwrap_or_else(|| unformatted.text_len().min(formatted.text_len()));
```
MichaReiser: I wonder if a regular loop would have been easier here (similar to what you have below):

```rust
let mut prefix_length = TextSize::ZERO;

for (unformatted, formatted) in unformatted.chars().zip(formatted.chars()) {
    if unformatted != formatted {
        break;
    }

    prefix_length += unformatted.text_len();
}
```

MichaReiser (approved):

crates/ruff/src/commands/format.rs
ntBre: We discussed this in our 1:1, but we only display line numbers for the new code, so I think this is correct. Somewhat random lint rule example:

```
help: Replace for loop with list comprehension
1 | original = list(range(10000))
  - filtered = []
  - for i in original:
  -     if i % 2:
  -         filtered.append(i)
2 + filtered = [i for i in original if i % 2]
```

This is a deletion but still shows that we only render  the new line numbers.

crates/ruff/src/commands/format.rs
ntBre: I think I'll leave this for now. I only split out deletions because I hit the `debug_assert!` that `content` is not empty in `Edit::range_replacement`. Insertions seem okay to group with full replacements since `content` is `Some` in both cases, and we already have a `TextRange`, which `Edit::insertion` would otherwise construct.

crates/ruff/src/commands/format.rs
ntBre: That is nicer, thanks. I used the guarded break for the suffix too instead of the if-else.

kaddkaka:
Does this PR take any steps closer to solving https://github.com/astral-sh/ruff/issues/14452 ?

ntBre:
No, I don't think so. This PR was for the `format` subcommand, so it shouldn't affect `ruff check` (the lint subcommand) at all. Micha's comment about Ruff only reporting leftover diagnostics after fixes have been applied is still accurate, as far as I know.

pygarap:
@ntBre 

Now, `ruff format --check`, become like `ruff format --diff`? But with more features.

So, `ruff format --diff` is deprecated for preview mode now?

And why the docs don't mention it?

```
      --check
          Avoid writing any formatted files back; instead, exit with a non-zero
          status code if any files would have been modified, and zero otherwise
      --diff
          Avoid writing any formatted files back; instead, exit with a non-zero
          status code and the difference between the current file and how the
          formatted file would look like
```

Looks like the `--check` CLI flag didn't change in the docs, But it did.

ntBre:
`--diff` produces a standalone diff that can still be useful for applying as a patch, for example, so it's not deprecated. The "diff" shown by the default `format --check` output uses the same format that's in preview for lint rules, which is a bit different.

I also think the `--check` help message is still accurate. The new information is in the new `--output-format` entry in the CLI help.

joukewitteveen:
This would be even more powerful when `ruff format --check` (and `ty check`) would also support `--output-file`. Currently, only `ruff check` supports that. Redirection (`ruff format --check [...] > DIR/OUTFILE`) doesn't fully solve the issue, since the directory containing the output file is required to exist.

kaddkaka:
> No, I don't think so. This PR was for the `format` subcommand, so it shouldn't affect `ruff check` (the lint subcommand) at all. Micha's comment about Ruff only reporting leftover diagnostics after fixes have been applied is still accurate, as far as I know.

Which comment are you referring to?

ntBre:
> Which comment are you referring to?

I think I was referring to this comment: https://github.com/astral-sh/ruff/issues/14452#issuecomment-2487825730

PR #20444, opened 2025-09-17
[ty] move primer projects to good.txt

carljm:
## Summary

After https://github.com/astral-sh/ruff/pull/20359 we can move all but three remaining projects over to `good.txt`.

## Test Plan

CI


AlexWaygood (approved):
Fantastic work!

PR #20447, opened 2025-09-17
[ty] move graphql-core to good.txt

carljm:
## Summary

With https://github.com/astral-sh/ruff/pull/20446, graphql-core now checks without error; we can move it to `good.txt`.

## Test Plan

CI


sharkdp:
Fantastic!

Just FYI: It's hopefully not that relevant anymore in the future, but you can include `bad.txt`/`good.txt` changes in your actual PR and use a ecosystem-analyzer CI run to verify the changes instead of mypy_primer. It uses the old `good.txt` for the old ty version, and the new `good.txt` for the new ty version.

Issue #13691, opened 2024-10-09
Audit how we determine whether a file is a "Python source file"


AlexWaygood:
I noticed in #13682 that there's some inconsistency regarding how we determine whether a file is a "Python source file" currently. In the code for `ruff server` (and the red-knot port of the server), we take care to do case-insensitive matching when figuring out whether something is a notebook file or not:

https://github.com/astral-sh/ruff/blob/5b4afd30caebd6e2c5c27bbf5debc1663cbb28a2/crates/ruff_server/src/session/index.rs#L124-L126

https://github.com/astral-sh/ruff/blob/5b4afd30caebd6e2c5c27bbf5debc1663cbb28a2/crates/red_knot_server/src/session/index.rs#L75-L79

Elsewhere, however, we mostly use case-sensitive matching:

https://github.com/astral-sh/ruff/blob/5b4afd30caebd6e2c5c27bbf5debc1663cbb28a2/crates/ruff_python_ast/src/lib.rs#L91-L101

https://github.com/astral-sh/ruff/blob/5b4afd30caebd6e2c5c27bbf5debc1663cbb28a2/crates/red_knot_python_semantic/src/module_resolver/path.rs#L41-L60

For places like the red-knot module resolver, it's likely correct to do case-sensitive matching (Python recognises `foo.py` as an importable module, but not `foo.PY`), but in other places it may not be. We should audit the code to make sure we're using case-sensitive matching and case-insensitive matching for file extensions in the correct places. We should also add comments to the places where there might be a subtle reason why case-(in)sensitive matching is required, rather than vice versa.

MichaReiser:
> For places like the red-knot module resolver, it's likely correct to do case-sensitive matching (Python recognises foo.py as an importable module, but not foo.PY

Note: I think that depends on the file system's case sensitivity.

Other than that. I think we should ignore casing when matching files because that's what most desktop environments to. They use the same application to open a `test.jpg` or `test.JPG`.

AlexWaygood:
Here, for example, we might want to do case-sensitive matching, since the rule only applies to the names of Python modules that are intended to be importable by other Python modules: https://github.com/astral-sh/ruff/blob/5b4afd30caebd6e2c5c27bbf5debc1663cbb28a2/crates/ruff_linter/src/rules/pep8_naming/rules/invalid_module_name.rs#L51-L59

[...]

PR #20458, opened 2025-09-18
[ruff] include `.pyw` files by default when linting and formatting

amyreese:
- **[ruff] Add tests to validate if ruff discovers .pyw files**
- **[ruff] include `.pyw` files by default when linting and formatting**

### DIFF
--- a/crates/ruff/tests/lint.rs
+++ b/crates/ruff/tests/lint.rs
@@ -2438,6 +2438,318 @@ requires-python = ">= 3.11"
 /// └── test.py
 /// ```
 #[test]
+fn requires_python_no_tool_preview_enabled() -> Result<()> {
+    let tempdir = TempDir::new()?;
+    let project_dir = dunce::canonicalize(tempdir.path())?;
+    let ruff_toml = tempdir.path().join("pyproject.toml");
+    fs::write(
+        &ruff_toml,
+        r#"[project]
+requires-python = ">= 3.11"
+"#,
+    )?;
+
+    let testpy = tempdir.path().join("test.py");
+    fs::write(
+        &testpy,
+        r#"from typing import Union;foo: Union[int, str] = 1"#,
+    )?;
+    insta::with_settings!({
+        filters => vec![(tempdir_filter(&project_dir).as_str(), "[TMP]/")]
+    }, {
+        assert_cmd_snapshot!(Command::new(get_cargo_bin(BIN_NAME))
+            .args(STDIN_BASE_OPTIONS)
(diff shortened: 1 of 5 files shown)


MichaReiser:
The basics here look good, but there are a few more places we have to account for `pyw` files.

One such place is `PySourceType`. It's heavily used internally to query whether a file is a Python file, a stub file, or a notebook. We need to decide if we need to add a new variant or if it's okay to add `pyw` to `py` (might not be okay if `PySourceType` is used in any module resolution logic as `pyw` files aren't importable as far as I know). This will require reviewing where `PySourceType` is used (and how).

It might also be good to do a quick search for ".py" (or "pyi" because there are fewer instances) to see if there are other places where we check for the extension that need updating.

This is related to https://github.com/astral-sh/ruff/issues/13691

amyreese:
> One such place is `PySourceType`. It's heavily used internally to query whether a file is a Python file, a stub file, or a notebook. We need to decide if we need to add a new variant or if it's okay to add `pyw` to `py` (might not be okay if `PySourceType` is used in any module resolution logic as `pyw` files aren't importable as far as I know). This will require reviewing where `PySourceType` is used (and how).

I tried looking through various places where this is used, but there really is an overwhelming number of places where it's used that I don't have context on, but what I've looked at so far usually seems reasonable. 

I went ahead and added the mapping to the existing `Python` enum, and tests are passing, but I don't have a good idea what sort of tests I could add to help give signal on whether that is sufficient, or how to compare that with the alternative of adding a new enum type.

That said, I'm also not sure what you mean by "module resolution logic", so if you can suggest specific crates/files to look at more closely, I'd be happy to dig further.

> It might also be good to do a quick search for ".py" (or "pyi" because there are fewer instances) to see if there are other places where we check for the extension that need updating.

I did do that originally, though I searched for `ipynb` instead, but apparently I missed the exact one place that `PySourceType` showed up in the gigantic list of results. 😅🫠

MichaReiser:
> I tried looking through various places where this is used, but there really is an overwhelming number of places where it's used that I don't have context on,

Yeah, same here. We don't need to study each line for a very long time. It's okay if we miss something. 

I only did a quick search. Most usages indeed do look good. The one I think we need to take a closer look is 

https://github.com/astral-sh/ruff/blob/b0279265e82903b57bb32f8bf04a2962740c5a8f/crates/ruff_linter/src/rules/flake8_no_pep420/rules/implicit_namespace_package.rs#L72-L73

because it tries to resemble python's module resolution at runtime.

PR #20459, opened 2025-09-18
[`flake8-simplify`] Fix diagnostic to show correct method name for rsplit calls (`SIM905`)

### LINKED ISSUE #20456: SIM905 diagnostic mentions `str.split` instead of `str.rsplit`
### Summary

The diagnostic for [`split-static-string` (SIM905)](https://docs.astral.sh/ruff/rules/split-static-string/) mentions `str.split` even when the relevant method is `str.rsplit`. [Example](https://play.ruff.rs/730e5360-1356-485f-848e-c8ebda0b9ebc):
```console
$ cat >sim905.py <<'# EOF'
"a,b,c,d".rsplit(",")
# EOF

$ ruff --isolated check sim905.py --select SIM905 --output-format concise -q
sim905.py:1:1: SIM905 [*] Consider using a list literal instead of `str.split`
```

### Version

danparizher:
## Summary

Fixes #20456


MichaReiser (approved):
Thank you

PR #20461, opened 2025-09-18
[ty] when analyzing a .py file, do not error if there's also a .pyi for that module

Gankra:
This is a naive(?) implementation of the approach @MichaReiser originally suggested to me in https://github.com/astral-sh/ty/issues/869

Fixes https://github.com/astral-sh/ty/issues/869
Fixes https://github.com/astral-sh/ty/issues/1195

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/resolver.rs
+++ b/crates/ty_python_semantic/src/module_resolver/resolver.rs
@@ -19,7 +19,7 @@ use rustc_hash::{FxBuildHasher, FxHashSet};
 use ruff_db::files::{File, FilePath, FileRootKind};
 use ruff_db::system::{DirectoryEntry, System, SystemPath, SystemPathBuf};
 use ruff_db::vendored::VendoredFileSystem;
-use ruff_python_ast::PythonVersion;
+use ruff_python_ast::{PySourceType, PythonVersion};
 
 use crate::db::Db;
 use crate::module_name::ModuleName;
@@ -155,26 +155,27 @@ pub(crate) fn file_to_module(db: &dyn Db, file: File) -> Option<Module<'_>> {
     let module_file = module.file(db)?;
 
     if file.path(db) == module_file.path(db) {
-        Some(module)
-    } else {
+        return Some(module);
+    } else if file.source_type(db) == PySourceType::Python
+        && module_file.source_type(db) == PySourceType::Stub
+    {
         // If a .py and .pyi are both defined, the .pyi will be the one returned by `resolve_module().file`,
         // which would make us erroneously believe the `.py` is *not* also this module (breaking things
         // like relative imports). So here we try `resolve_real_module().file` to cover both cases.
         let module = resolve_real_module(db, &module_name)?;
         let module_file = module.file(db)?;
         if file.path(db) == module_file.path(db) {
-            Some(module)
-        } else {
-            // This path is for a module with the same name but with a different precedence. For example:
-            // ```
-            // src/foo.py
-            // src/foo/__init__.py
-            // ```
-            // The module name of `src/foo.py` is `foo`, but the module loaded by Python is `src/foo/__init__.py`.
-            // That means we need to ignore `src/foo.py` even though it resolves to the same module name.
-            None
+            return Some(module);
         }
     }
+    // This path is for a module with the same name but with a different precedence. For example:
+    // ```
+    // src/foo.py
+    // src/foo/__init__.py
+    // ```
+    // The module name of `src/foo.py` is `foo`, but the module loaded by Python is `src/foo/__init__.py`.
+    // That means we need to ignore `src/foo.py` even though it resolves to the same module name.
+    None
 }
 
 pub(crate) fn search_paths(db: &dyn Db, resolve_mode: ModuleResolveMode) -> SearchPathIterator<'_> {


crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1576,16 +1585,18 @@ mod tests {
...
         assert_eq!(&src, foo.search_path(&db).unwrap());
         assert_eq!(&foo_stub, foo.file(&db).unwrap().path(&db));
 
         assert_eq!(Some(foo), path_to_module(&db, &FilePath::System(foo_stub)));
         assert_eq!(
-            None,
+            Some(foo_real),
             path_to_module(&db, &FilePath::System(src.join("foo.py")))
         );
+        assert!(foo_real != foo);
```
Gankra, line 1600: I'm hoping the fact that file_to_module returns two *different* modules that happen to have the same name Contains The Chaos.

Gankra:
Wow that's a lot of fixed imports in mypy_primer

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -157,14 +157,23 @@ pub(crate) fn file_to_module(db: &dyn Db, file: File) -> Option<Module<'_>> {
...
-        // src/foo.py
-        // src/foo/__init__.py
-        // ```
-        // The module name of `src/foo.py` is `foo`, but the module loaded by Python is `src/foo/__init__.py`.
-        // That means we need to ignore `src/foo.py` even though it resolves to the same module name.
-        None
+        // If a .py and .pyi are both defined, the .pyi will be the one returned by `resolve_module().file`,
+        // which would make us erroneously believe the `.py` is *not* also this module (breaking things
+        // like relative imports). So here we try `resolve_real_module().file` to cover both cases.
+        let module = resolve_real_module(db, &module_name)?;
```
MichaReiser, line 165: Should we limit this to only run if `module_file.path` has a `pyi` extension (and `file` has an extension other than `pyi`)?

Gankra: Semantically that won't have any effect, but I could add it if you're concerned about perf?

MichaReiser: I wouldn't say I'm concerned about performance given that unresolved imports should be rare. But it seems trivial enough and it would save a fair amount of IO operations in cases where both files are py files. That's why I think a simple check that catches most cases where we can skip would be nice to add but it's fine if it doesn't bail early in all cases where it could

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -157,14 +157,23 @@ pub(crate) fn file_to_module(db: &dyn Db, file: File) -> Option<Module<'_>> {
...
+        } else {
+            // This path is for a module with the same name but with a different precedence. For example:
+            // ```
+            // src/foo.py
+            // src/foo/__init__.py
+            // ```
+            // The module name of `src/foo.py` is `foo`, but the module loaded by Python is `src/foo/__init__.py`.
+            // That means we need to ignore `src/foo.py` even though it resolves to the same module name.
+            None
+        }
```
carljm, line 169: I wonder if there's really solid justification for doing this only in the .pyi vs .py case, which requires this additional somewhat expensive "re-import without stubs" check. The alternative would be to remove the existing `resolve_module` check (so making this method generally simpler and more efficient) and just accept that this will always return the module name that a file _would_ have, were it not shadowed by some other file with the same module name.

It seems like even for the `src/foo.py` vs `src/foo/__init__.py` case (or the two different root search paths case), this is arguably still an improvement -- if you open up `src/foo.py` in your editor, wouldn't we still want to be able to resolve relative imports in it?

But I haven't had my head deeply in this code for a long time, so maybe there are other negative side effects this would have? It at least seems worth a try, given it would simplify this method rather than add to it, and give more consistent behavior across all cases of "file is import-shadowed by another file."

MichaReiser: The `src/foo.py` case is interesting. But I do think it might result in incorrect results for relative imports if the module is from a different search root (because we then incorrectly resolve the import relative to the module rather than the file's folder). This is probably not horrible in practice and fine in the editor, but probably not what we want in type inference?

carljm: It's hard to say what is the "right" behavior here for type inference, since the runtime behavior is that an "import-shadowed" module will never be imported in the first place, so it's sort of "unreachable". Given that, it's not clear to me that there is a wrong answer for how we resolve relative imports in that module, and either answer is probably better than failing to resolve them at all? Edit: it does kind of make intuitive sense in the separate-search-paths case that maybe you'd want to resolve "as if the other search path weren't there" -- I'm sure we could implement that, if it shows up in practice as something that trips people up.

Gankra:
> Not for this PR but did you look into some of the new ecosystem diagnostics if they look correct or if they show a similar issue (e.g that star imports don't work or I don't know what).

From a random skimming of the new issues, they seem to consist entirely of "foo.py is using private internals of bar.py that bar.pyi does not declare and therefore we do not believe exist"

PR #20466, opened 2025-09-18
[ty] Swap `detail` and `description` fields for `CompletionItemLabelDetails`

### LINKED ISSUE #1200: I001: Support for magic trailing commas
Black auto-explodes imports, collections, function calls, etc. when they contain a trailing comma:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

This means that even though this would fit into one line:
```py
from seven_dwwarfs import (
    Grumpy,
    Happy,
    Sleepy,
    Bashful,
    Sneezy,
    Dopey,
    Doc,
)
```
Black will NOT convert it into this:
```py
from seven_dwwarfs import Grumpy, Happy, Sleepy, Bashful, Sneezy, Dopey, Doc
```
As long as the import list contains a trailing comma.

Ruff currently does not have the magic trailing comma handling as far as I can tell and will collapse the import list if it can fit within the line length limit.

See relevant fixed (but not released) issue in isort: https://github.com/PyCQA/isort/issues/1683

BurntSushi:
This seems to be more consistent with how other LSPs work (like
`rust-analyzer`), and also I think is more consistent with how
`CompletionItem.detail` is itself rendered. Namely, in VS Code, it
is right-aligned. And it's also where we put the type signature.
But `CompletionItemLabelDetails.detail` is left-aligned where as
`CompletionItemLabelDetails.description` is right-aligned. So let's
swap them such that type signatures go in the latter and not the
former.

Fixes #1200


BurntSushi:
Demo:

https://github.com/user-attachments/assets/f0f1a5bd-332f-4009-ae0b-da98e683331d

MichaReiser:
Thanks for looking into this. 

It now seems unfortunate that the module name comes directly (without any space) after the completion text which seems expected acording to the LSP specification

```ts
/**
 * Additional details for a completion item label.
 *
 * @since 3.17.0
 */
export interface [CompletionItemLabelDetails](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails) {

	/**
	 * An optional string which is rendered less prominently directly after
	 * {@link CompletionItem.label label}, without any spacing. Should be
	 * used for function signatures or type annotations.
	 */
	detail?: string;

	/**
	 * An optional string which is rendered less prominently after
	 * {@link CompletionItemLabelDetails.detail}. Should be used for fully qualified
	 * names or file path.
	 */
	description?: string;
}
```

I think we should at least insert a space at the front of `detail` if we want to use that field.

Or we keep using `detail` for the type but change the syntax to `: {type}`, similar to what we show in inlay hints. Wdyt? 

Edit: I only just now saw your detailed write-up on the issue. I think what you did makes sense. I'm only wondering if we should introduce a space to make this look less squeezed together

Edit2: It seems that r-a solves this visual issue by using `(use <module name>)` for the left aligned content which helps a bit.

BurntSushi:
Yeah and r-a inserts a leading space too.

I added the leading space and added `(import <name>)`, which I think also helps (similar r-a's `(use <name>)`).

Demo:


https://github.com/user-attachments/assets/6e6a4019-6552-4b7a-b787-416695b2692f

MichaReiser (approved):

PR #20470, opened 2025-09-18
Bump MSRV to Rust 1.88

MichaReiser:
## Summary

Bumps the Rust MSRV to Rust 1.88 in accordance to our versioning policy. 

This needs https://github.com/conda-forge/ruff-feedstock/pull/296

### DIFF
--- a/crates/ruff_db/src/diagnostic/render/full.rs
+++ b/crates/ruff_db/src/diagnostic/render/full.rs
@@ -58,10 +58,11 @@ impl<'a> FullRenderer<'a> {
                 writeln!(f, "{}", renderer.render(diag.to_annotate()))?;
             }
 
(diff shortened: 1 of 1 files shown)


crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -58,10 +58,11 @@ impl<'a> FullRenderer<'a> {
                 writeln!(f, "{}", renderer.render(diag.to_annotate()))?;
             }
 
-            if self.config.show_fix_diff && diag.has_applicable_fix(self.config) {
-                if let Some(diff) = Diff::from_diagnostic(diag, &stylesheet, self.resolver) {
-                    write!(f, "{diff}")?;
-                }
+            if self.config.show_fix_diff
```
MichaReiser, line 61: Funny how this is the only place where clippy identified the possibility to use if-let chains

PR #20474, opened 2025-09-18
[ty] Implement additional implicit re-exports idiom for `__init__.pyi`

Gankra:
Relative imports like `from . import submodule` in `__init.pyi__` are pervasively used to define `mypackage.submodule` and should be regarded as intentional (if not intentional, you can use `from mypackage import submodule` to opt out).

A simple example of this in the wild that requires this is:

```
import msgspec
reveal_type(msgspec.json)
```

Because msgspec's `__init.pyi__` (and `__init.py__`) contain `from . import json`

The more subtle one is `from .submodule import whatever`. In this case pylance considers both `submodule` and `whatever` to be re-exported. Pylance [documents and implements this behaviour](https://microsoft.github.io/pyright/#/import-statements). However I don't (yet) have an in-the-wild example of this. As such I have not yet implemented it, pending more testing (really the worst part of this whole feature has just been making sure I'm actually testing the behaviour I mean to test because there's a dozen subtly different situations).

### DIFF
--- a/crates/ruff_db/src/files.rs
+++ b/crates/ruff_db/src/files.rs
@@ -461,25 +461,8 @@ impl File {
 
     /// Returns `true` if the file is an `__init__.py(i)`
     pub fn is_init(self, db: &dyn Db) -> bool {
-        let (name, extension) = match self.path(db) {
-            FilePath::System(path) => (
-                path.file_stem(),
-                path.extension().and_then(PySourceType::try_from_extension),
-            ),
-            FilePath::Vendored(path) => (
-                path.file_stem(),
-                path.extension().and_then(PySourceType::try_from_extension),
-            ),
-            FilePath::SystemVirtual(path) => (
-                path.file_stem(),
-                path.extension().and_then(PySourceType::try_from_extension),
-            ),
-        };
-        name == Some("__init__")
-            && matches!(
-                extension,
-                Some(PySourceType::Python) | Some(PySourceType::Stub)
-            )
+        let path = self.path(db).as_str();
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/place.rs
```diff
@@ -1320,21 +1326,26 @@ fn place_from_declarations_impl<'db>(
 // This will first check if the definition is using the "redundant alias" pattern like `import foo
 // as foo` or `from foo import bar as bar`. If it's not, it will check whether the symbol is being
 // exported via `__all__`.
-fn is_reexported(db: &dyn Db, definition: Definition<'_>) -> bool {
+fn is_reexported(db: &dyn Db, definition: Definition<'_>) -> ReExportKind {
```
Gankra, line 1329: I plan to rename `is_reexported` everywhere to `reexport` or something

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1562,6 +1569,15 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                         (&alias.name.id, false)
                     };
 
+                    let is_reexported = if is_reexported {
+                        ReExportKind::Yes
+                    } else if node.level == 1 {
+                        // `from . import a`
+                        ReExportKind::InitIdiom
+                    } else {
```
Gankra, line 1577: Possibly this should be `>= 1`? Need to test that situation more.

Gankra:
Obsolete

PR #20476, opened 2025-09-18
[ty] Use type context for inference of generic function calls

ibraheemdev:
## Summary

Part of https://github.com/astral-sh/ty/issues/168. This PR allows widening the type of a generic function call based on a type annotation, e.g.,
```py
def f[T](x: T) -> list[T]:
    return [x]

v: list[int] = f(True)
reveal_type(v)  # list[int]
```

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -131,12 +131,12 @@ m: IntList = [1, 2, 3]
 reveal_type(m)  # revealed: list[int]
 
 # TODO: this should type-check and avoid literal promotion
-# error: [invalid-assignment] "Object of type `list[int]` is not assignable to `list[Literal[1, 2, 3]]`"
+# error: [invalid-assignment] "Object of type `list[Unknown | int]` is not assignable to `list[Literal[1, 2, 3]]`"
 n: list[typing.Literal[1, 2, 3]] = [1, 2, 3]
 reveal_type(n)  # revealed: list[Literal[1, 2, 3]]
 
 # TODO: this should type-check and avoid literal promotion
-# error: [invalid-assignment] "Object of type `list[str]` is not assignable to `list[LiteralString]`"
+# error: [invalid-assignment] "Object of type `list[Unknown | str]` is not assignable to `list[LiteralString]`"
 o: list[typing.LiteralString] = ["a", "b", "c"]
 reveal_type(o)  # revealed: list[LiteralString]
 ```
@@ -144,10 +144,10 @@ reveal_type(o)  # revealed: list[LiteralString]
 ## Incorrect collection literal assignments are complained aobut
 
 ```py
-# error: [invalid-assignment] "Object of type `list[int]` is not assignable to `list[str]`"
+# error: [invalid-assignment] "Object of type `list[Unknown | int]` is not assignable to `list[str]`"
 a: list[str] = [1, 2, 3]
 
-# error: [invalid-assignment] "Object of type `set[int | str]` is not assignable to `set[int]`"
+# error: [invalid-assignment] "Object of type `set[Unknown | int | str]` is not assignable to `set[int]`"
 b: set[int] = {1, 2, "3"}
 ```
 
@@ -263,8 +263,7 @@ reveal_type(d)  # revealed: list[int | tuple[int, int]]
 e: list[int] = f(True)
 reveal_type(e)  # revealed: list[int]
 
-# TODO: the RHS should be inferred as `list[Literal["a"]]` here
-# error: [invalid-assignment] "Object of type `list[int | Literal["a"]]` is not assignable to `list[int]`"
+# error: [invalid-assignment] "Object of type `list[Literal["a"]]` is not assignable to `list[int]`"
 g: list[int] = f("a")
 
 # error: [invalid-assignment] "Object of type `list[Literal["a"]]` is not assignable to `tuple[int]`"
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -234,3 +234,48 @@ reveal_type(x)  # revealed: Foo
 x: int = 1
 reveal_type(x) # revealed: Literal[1]
 ```
+
+## Annotations influence generic call inference
+
+```toml
+[environment]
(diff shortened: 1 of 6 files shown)


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -234,3 +234,48 @@ reveal_type(x)  # revealed: Foo
...
+
+d: list[int | tuple[int, int]] = f((1, 2))
+reveal_type(d)  # revealed: list[int | tuple[int, int]]
+
+e: list[int] = f(True)
+reveal_type(e)  # revealed: list[int]
+
+# TODO: the RHS should be inferred as `list[Literal["a"]]` here
+# error: [invalid-assignment] "Object of type `list[int | Literal["a"]]` is not assignable to `list[int]`"
+g: list[int] = f("a")
```
ibraheemdev, line 267: Ideally we would infer the RHS without the type context if it is not assignable to the type annotation, like we do for collection literals. However, this is a little difficult as we also want the type annotation to be the first type that we use for inference, so that the inferred type more closely matches the type annotation.

I think the only way this can work is if we perform inference twice, first without the type context, and then again with it only if the unannotated inferred type is assignable to the type annotation. However, I'm not sure that's worth the extra work for this specific error message.

Maybe we could instead generalize this to performing type inference on the RHS directly when emitting the diagnostic?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -234,3 +234,48 @@ reveal_type(x)  # revealed: Foo
...
+h: tuple[int] = f("a")
+
+def f2[T: int](x: T) -> T:
+    return x
+
+i: int = f2(True)
+reveal_type(i)  # revealed: int
+
+j: int | str = f2(True)
+reveal_type(j)  # revealed: Literal[True]
```
ibraheemdev, line 279: This works because we ignore specialization errors on the return type. pyright also infers `bool` here, but mypy interestingly still infers `int | str`.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
carljm: Yeah, I think what would make sense here is to re-infer the RHS without context if we are emitting a diagnostic. Some extra cost is fine, if it's only paid when actually emitting a diagnostic.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -120,16 +121,21 @@ impl<'db> Bindings<'db> {
     /// You must provide an `argument_types` that was created from the same `arguments` that you
     /// provided to [`match_parameters`][Self::match_parameters].
     ///
+    /// The return type annotation is also used to infer the specialization of generic calls.
```
carljm: Not sure if I'm understanding this comment right, but the wording doesn't quite make sense to me. By "return type annotation" I would generally mean the `-> T` in `def f[T](x: T) -> T:`. But `return_tcx` here is not that, it's the annotated type context of the entire call expression, which would be used to help infer the type of `T`.

If I'm getting this right, I think I would name the argument `call_expression_tcx` not `return_tcx` (though I understand the rationale for the latter name, since it's the return type which will become the type of the entire call expression), and would reword this comment to something like
```suggestion
    /// The type context of the call expression is also used to infer the specialization of generic calls.
```

carljm (approved):
Nice! Have you done any analysis of the ecosystem results?

ibraheemdev:
There seems to be a common ecosystem error where we fail to type-check calls to `dataclasses.field`. This can be minimized to:
```py
class X: ...

def field[T](default_factory: Callable[[], T]) -> T:
    return default_factory()

# error[invalid-argument-type] Argument to function `field` is incorrect: Expected `() -> X`, found `<class 'X'>`
x: X = field(X)
```

However, without the type annotation, we currently infer `Unknown` for `x`, so I'm not sure if the issue is related to this PR?

carljm:
The fact that we can't infer this typevar is https://github.com/astral-sh/ty/issues/500, which we closed as duplicate of the new constraint solver.

It's a bit strange to me that we display the callable as `() -> X` and think the class `X` is not assignable to that; I'm not clear what's going on there. I do think it's probably not directly related to this PR, but clearly this PR is introducing the false positives in this case, and I think these are probably important false positives to understand and fix. Maybe the new constraint solver will just take care of this?

crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -217,6 +218,24 @@ fn infer_expression_types_impl<'db>(
     .finish_expression()
 }
 
+/// Infer the type of an non-standalone expression in isolation.
+///
+/// The type returned by this function may be different than the type of the expression
+/// if it was inferred within its region, as it does not account for surrounding type context.
+/// This can be useful to re-infer the type of an expression for diagnostics.
+pub(crate) fn infer_isolated_expression<'db>(
```
ibraheemdev, line 226: Don't love the name here.

ibraheemdev:
> It's a bit strange to me that we display the callable as `() -> X` and think the `class X` is not assignable to that; I'm not clear what's going on there. I do think it's probably not directly related to this PR, but clearly this PR is introducing the false positives in this case, and I think these are probably important false positives to understand and fix. Maybe the new constraint solver will just take care of this?

It looks like we correctly infer the type variable now due to the type annotation, but the issue remains without any generics, so I think this is unrelated:
```py
class X:
    ...

def f(callable: Callable[[], X]) -> X:
    return callable()

# error[invalid-argument-type]: Argument to function `f` is incorrect
x = f(X)
```
It seems to work if I add an explicit `__new__` or `__init__` method to `X`, I'm not sure if there's an open issue about this.

Edit: Opened https://github.com/astral-sh/ty/issues/1210.

ibraheemdev:
There's another issue here with:
```py
def f[T](callable: Callable[[], T]) -> T:
    return callable()

# Expected `() -> Mapping[str, str]`, found `<class 'dict'>`
x: Mapping[str, str] = f(dict)
```

After this PR, we infer `T` to be `Mapping[str, str]` based on the type annotation, and so expect a `Callable[[], Mapping[str, str]]`. This one should be fixed by https://github.com/astral-sh/ty/issues/500, because we should also account for the return type of the provided callable (i.e. `dict`).

Given that both of those issues are unrelated to the changes here, I think this PR should be good to merge.

PR #20479, opened 2025-09-19
[ty] Use `C[T]` instead of `C[Unknown]` for the upper bound of `Self`

sharkdp:
# do not review yet


sharkdp:
It's just an endless rabbithole. Now we pick the wrong overload of `datetime.__sub__` when called with a `timedelta` for some reason. Probably because of the `skip_assignability_check` hack — which allows `timedelta` to be incorrectly assigned to `Self@datetime`

crates/ty_python_semantic/resources/mdtest/overloads.md
```diff
@@ -126,6 +126,18 @@ foo2 = Foo2()
...
+    @overload
+    def takes_self_or_int(self: Self, x: Self) -> Self: ...
+    @overload
+    def takes_self_or_int(self: Self, x: int) -> int: ...
+    def takes_self_or_int(self: Self, x: Self | int) -> Self | int:
+        return x
+
+foo3 = Foo3()
+reveal_type(foo3.takes_self_or_int(foo3))  # revealed: Foo3
+reveal_type(foo3.takes_self_or_int(1))  # revealed: int
```
sharkdp, line 140: This is a new test case adapted from a `datetime`/`timedelta` example observed in the ecosystem.

The lower of these two assertions was failing with the induct-into-bounds-of-typevars approach, because we would pick the first overload and return `Foo3 | Literal[1]`, if assignability to the upper bound of `Self` was not properly checked. And if we properly check the assignability to the upper bound of `Self`, we're getting `false` from the missing `has_relation_to_impl` branch, and can't fix https://github.com/astral-sh/ty/issues/1196. So at that point, we need to bring back all of the subtyping/assignability adjustments. And then the induction into the bounds of `Self` doesn't have any observable effect anymore, so I reverted it.

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -366,6 +366,31 @@ reveal_type(f(g("a")))  # revealed: tuple[Literal["a"] | None, int]
 reveal_type(g(f("a")))  # revealed: tuple[Literal["a"], int] | None
 ```
 
+## Passing generic functions to generics functions
```
carljm:
```suggestion
## Passing generic functions to generic functions
```

crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
```diff
@@ -321,8 +321,10 @@ a covariant generic, this is equivalent to using the upper bound of the type par
 `object`):
 
 ```py
+from typing import Self
+
 class Covariant[T]:
-    def get(self) -> T:
+    def get(self: Self) -> T:
```
carljm, line 328: Why do we need the explicit `Self` annotations in this test?

Is this the right way to write the tests long-term (i.e. once we support implicit-type-of-self properly)? If not, should we have a TODO here to remove these annotations?

crates/ty_python_semantic/src/types.rs
```diff
@@ -6695,8 +6730,9 @@ pub enum TypeMapping<'a, 'db> {
     BindSelf(Type<'db>),
     /// Replaces occurrences of `typing.Self` with a new `Self` type variable with the given upper bound.
     ReplaceSelf { new_upper_bound: Type<'db> },
-    /// Marks the typevars that are bound by a generic class or function as inferable.
-    MarkTypeVarsInferable(BindingContext<'db>),
+    /// Marks the typevars that are bound by a generic class or function as inferable. If the parameter
+    /// is `None`, *all* typevars are marked as inferable.
```
dcreager: We discussed this synchronously, but can you add a bit more context here for when the `None` case will be invoked? (The `Some` case is because a generic function might reference other typevars from enclosing generic scopes, and we only want to make _that function's_ typevars as inferable. But once we have decided that a particular typevar should be marked inferable, then any typevars mentioned in its default values (and bound/constraints if that approach had turned out to be fruitful) should be unconditionally marked as inferable)

crates/ty_python_semantic/src/types.rs
```diff
@@ -7637,6 +7673,40 @@ impl<'db> TypeVarInstance<'db> {
         )
     }
 
+    fn mark_typevars_inferable(
+        self,
+        db: &'db dyn Db,
+        visitor: &ApplyTypeMappingVisitor<'db>,
+    ) -> Self {
+        let type_mapping = &TypeMapping::MarkTypeVarsInferable(None);
```
dcreager, line 7692: And then a bit of copy/paste from that commentary here where we switch from the `Some` case to the `None` case

crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
sharkdp: > Why do we need the explicit `Self` annotations in this test?

With these explicit `Self` annotations, these all become regression tests for https://github.com/astral-sh/ty/issues/1196. I could have added new tests, but would have had to duplicate the different variance examples from here. Hopefully `self` and `self: Self` will soon be exactly equivalent (?), so yes, we will be able to remove these soon. I added TODO comments.

PR #20481, opened 2025-09-19
[ty] Remove unnecessary `FileScopeId` to `ScopeId` conversion

MichaReiser:
I noticed this when debugging https://github.com/astral-sh/ty/issues/1111 because I saw many repeated `semantic_index` calls.

Going to `ScopeId` here is unnecessary (and only introduces more work). We can retrieve the `node` directly from `index.scope(id).node`

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/scope.rs
+++ b/crates/ty_python_semantic/src/semantic_index/scope.rs
@@ -26,11 +26,6 @@ pub struct ScopeId<'db> {
 impl get_size2::GetSize for ScopeId<'_> {}
 
 impl<'db> ScopeId<'db> {
-    #[expect(unused)]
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/semantic_index/scope.rs
```diff
@@ -26,6 +26,7 @@ pub struct ScopeId<'db> {
 impl get_size2::GetSize for ScopeId<'_> {}
 
 impl<'db> ScopeId<'db> {
+    #[expect(unused)]
     pub(crate) fn is_function_like(self, db: &'db dyn Db) -> bool {
         self.node(db).scope_kind().is_function_like()
     }
```
AlexWaygood: I'd just delete this if it's now unused -- it seems pretty trivial? But no strong opinion

PR #20486, opened 2025-09-19
[`ruff`] Fix `B004` to skip invalid `hasattr`/`getattr` calls

### LINKED ISSUE #20440: B004 false positives for invalid or obfuscated `hasattr` and `getattr` calls
### Summary

[`unreliable-callable-check` (B004)](https://docs.astral.sh/ruff/rules/unreliable-callable-check/) doesn’t validate the arguments of `hasattr` or `getattr`.

It should be suppressed when `hasattr` is called with anything another than two positional non-starred arguments, lest it suppress or introduce an exception. [Example](https://play.ruff.rs/f11372d2-11f5-4800-a5ab-5fb7f39f86cb):
```console
$ cat >b004.py <<'# EOF'
try: print(hasattr(0, "__call__", 0, x=0))
except TypeError as e: print(e)

try: print(hasattr(*(), "__call__", "split"))
except TypeError as e: print(e)
# EOF

$ python b004.py
hasattr() takes no keyword arguments
True

$ ruff --isolated check b004.py --select B004 --fix

$ cat b004.py
try: print(callable(0))
except TypeError as e: print(e)

try: print(callable(*()))
except TypeError as e: print(e)

$ python b004.py
False
callable() takes exactly one argument (0 given)
```

Similarly, B004 should only apply when `getattr` is called with two or three positional non-starred arguments.

### Version

TaKO8Ki:
## Summary

Fixes #20440

Fix B004 to skip invalid hasattr/getattr calls

- Add argument validation for `hasattr` and `getattr`
- Skip B004 rule when function calls have invalid argument patterns

## Test Plan

Add comprehensive test cases for invalid argument patterns including unpacking to `B004.py`

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_bugbear/rules/unreliable_callable_check.rs
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/rules/unreliable_callable_check.rs
@@ -92,6 +92,9 @@ pub(crate) fn unreliable_callable_check(
     args: &[Expr],
     keywords: &[ast::Keyword],
 ) {
+    if !keywords.is_empty() {
+        return;
+    }
     let [obj, attr, ..] = args else {
         return;
     };
@@ -109,11 +112,11 @@ pub(crate) fn unreliable_callable_check(
     let valid_args = match builtins_function {
         "hasattr" => {
             // hasattr should have exactly 2 positional arguments and no keywords
-            args.len() == 2 && keywords.is_empty()
+            args.len() == 2
         }
         "getattr" => {
             // getattr should have 2 or 3 positional arguments and no keywords
-            (args.len() == 2 || args.len() == 3) && keywords.is_empty()
+            args.len() == 2 || args.len() == 3
         }
         _ => return,
     };


crates/ruff_linter/src/rules/flake8_bugbear/rules/unreliable_callable_check.rs
```diff
@@ -103,7 +104,21 @@ pub(crate) fn unreliable_callable_check(
...
+        "hasattr" => {
+            // hasattr should have exactly 2 positional arguments and no keywords
+            args.len() == 2 && keywords.is_empty()
+        }
+        "getattr" => {
+            // getattr should have 2 or 3 positional arguments and no keywords
+            (args.len() == 2 || args.len() == 3) && keywords.is_empty()
+        }
+        _ => return,
+    };
```
dylwil3, line 122: We may as well move the `keywords` check all the way to the beginning of the function. (You could even leave the original function unmodified and do the `keywords` check at the call site - but it's a little more self-documenting to keep the logic inside the function, I guess).

TaKO8Ki: I've moved the keywords count check to the beginning of `unreliable_callable_check`.

PR #20489, opened 2025-09-20
Upgrade rustix to 1.1.1

WhyNotHugo:
Rustix fails to build on loongarch64 due to ambiguous glob imports for the same symbol. This was fixed in 1.1.1.

Upgrade pinned version to 1.1.1 to fix build failures in transitive dependency.

See: https://github.com/bytecodealliance/rustix/issues/1502
See: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/89436


MichaReiser:
Thank you. I went ahead and bumped all our transitive dependencies in https://github.com/astral-sh/ruff/pull/20513 This also includes the radix upgrade.

WhyNotHugo:
Thanks!

PR #20490, opened 2025-09-20
[ty]: Add PYTHONPATH to EnvVars and fix on Windows

mmlb:
## Summary

* Add `PYTHONPATH` to `ty_static::EnvVars` so it shows up in envvar docs.
* Remove custom env path splitting with hard coded separator and use stdlib instead, fixing windows. 

## Test Plan

No new tests added.


mmlb:
The EnvVar bit came out of https://github.com/astral-sh/ty/issues/1219 and coming up with a doc for that issue I checked CPython's own docs and noted the mention of os.pathsep and :man_facepalming: so fixed it here too. Not sure how to add a windows test and also unsure if the `panic!` is the right call in system.rs

crates/ruff_db/src/system.rs
```diff
@@ -191,6 +191,11 @@ pub trait System: Debug + Sync + Send {
         Err(std::env::VarError::NotPresent)
     }
 
+    fn split_paths<'a>(&self, name: &'a str) -> std::env::SplitPaths<'a> {
+        let _ = name;
+        panic!("split_paths is not implemented for this system")
```
mmlb: not sure if there's a better alternative

crates/ty_project/src/metadata/options.rs
```diff
@@ -302,9 +303,17 @@ impl Options {
         // they exist as a directory, and add them to the vec of extra_paths
         // as they should be checked before site-packages just like python
         // interpreter does
-        if let Ok(python_path) = system.env_var("PYTHONPATH") {
-            for path in python_path.split(':') {
-                let possible_path = SystemPath::absolute(path, system.current_directory());
+        if let Ok(python_path) = system.env_var(EnvVars::PYTHONPATH) {
+            for path in system.split_paths(python_path.as_str()) {
+                let possible_path = if let Some(val) = path.to_str() {
```
mmlb: should this be `to_string_lossy` instead?

crates/ruff_db/src/system.rs
MichaReiser: We shouldn't provide a default implementation that always panics. Instead, each `System` implementation should add its own implementation. 
* Implementing it for `TestSystem` is trivial because we can forward to the inner system (the same as all other methods do). 
* Implementing it for `LSPSystem` is trivial too, simply forward to the `native_system`
* Implementing it for `MemorySystem` is a bit trickier. I'd be inclined to simply split by `:` 

The method also can't return `std::env::SplitPaths` because it means that the only possible implementation is `std::env::split_paths`. The return type either has to be a `vec[SystemPath]` or `Box<dyn Iterator<Item=&'a SystemPath>>`


Having said all this. We've only added methods to `System` when we had a need to abstract over them. For example, we only added `env_var` because we ran into issues with test isolation. All file system operations exist because we need to "patch" some in the LSP case AND because we wanted to use a memory file system in tests. 

That's why I'd suggest removing the method for now. We can always add it to `System` when the need arises and it simplifies our lifes now :)

There are other platform-specific behaviors we're currently not normalizing by using `System`. For example, `SystemPath` uses the underlying `Path` type. That means Windows and Unix use different path representations. We could normalize all paths to Unix paths but we haven't seen a need for it just yet.

crates/ty_project/src/metadata/options.rs
```diff
@@ -302,9 +303,17 @@ impl Options {
         // they exist as a directory, and add them to the vec of extra_paths
         // as they should be checked before site-packages just like python
         // interpreter does
-        if let Ok(python_path) = system.env_var("PYTHONPATH") {
-            for path in python_path.split(':') {
-                let possible_path = SystemPath::absolute(path, system.current_directory());
+        if let Ok(python_path) = system.env_var(EnvVars::PYTHONPATH) {
+            for path in system.split_paths(python_path.as_str()) {
```
MichaReiser: Oh nice find with `split_paths`. I think we should update our CLI test to correctly set the `PYTHONPATH` variable depending on the platform (by using `std::env::join_paths`)

crates/ty_project/src/metadata/options.rs
MichaReiser: What you have here is fine but I suggest using `let Ok(path) = ... else {` and using `SystemPathBuf::from_std_path` instead (which avoids any allocations if the path is a valid UTF8 path)


```rust
                let Ok(path) = SystemPathBuf::from_path_buf(path) else {
                    tracing::debug!(
                        "Skipping `{path}` listed in `PYTHONPATH` because it is not a valid UTF-8 path."
                    );
                    continue;
                };

                let possible_path = SystemPath::absolute(path, system.current_directory());
```

crates/ty_project/src/metadata/options.rs
```diff
@@ -302,9 +303,17 @@ impl Options {
...
+                let possible_path = if let Some(val) = path.to_str() {
+                    SystemPath::absolute(val, system.current_directory())
+                } else {
+                    let lossy = path.to_string_lossy();
+                    tracing::debug!(
+                        "Skipping `{lossy}` listed in `PYTHONPATH` because it contains invalid utf-8 sequences"
+                    );
+                    continue;
+                };
```
MichaReiser, line 323: Reading this code again. I'm wondering what the precedence between `PYTHONPATH` and `extra_paths` should be. What's implemented now is that `PYTHONPATH` takes precedence over the user-provided `extra_paths`. This contradicts the extra paths documentation.

https://github.com/astral-sh/ruff/blob/2fec3f75ee42a149ccb39e498470f1300ec93466/crates/ty_project/src/metadata/options.rs#L546-L548

@AlexWaygood do you think it makes sense that `PYTHONPATH` takes precedence over `extra_paths`, or should it be the other way round?

crates/ty_static/src/env_vars.rs
```diff
@@ -42,6 +42,9 @@ impl EnvVars {
     /// Used to detect an activated virtual environment.
     pub const VIRTUAL_ENV: &'static str = "VIRTUAL_ENV";
 
+    /// Augment the default search path for module files.
```
MichaReiser: I don't think I'd know what this means as a ty user (what does augmented mean in this case?)

uv's documentation for `PYTHONPATH` is 

```
Adds directories to Python module search path (e.g., PYTHONPATH=/path/to/modules).
```

I suggest something similar

Adds additional directories to ty's search paths. The format is the same as the shell’s PATH: one or more directory pathnames separated by [os.pathsep](https://docs.python.org/3/library/os.html#os.pathsep) (e.g. colons on Unix or semicolons on Windows).

MichaReiser (commented):
Thank you. A few nits and suggestions around the documentation. I suggest removing `split_paths` from `System` for now unless there's a need today (e.g. the wasm tests are failing in CI)

crates/ty_project/src/metadata/options.rs
AlexWaygood: I agree with you — I think ty's explicit configuration (`extra_paths`) should take precedence over paths in `PYTHONPATH`, which is an external environment variable (used for configuring Python itself) that isn't ty-specific

crates/ruff_db/src/system.rs
mmlb: my initial thought was to just use os.pathsep equivalent, but its not exposed. I flirted with just defining our own and using with the split I had already but then switched over to the stdlib's function. I'll just go back to my first PR with platform specific separator, that seemed the cleanest way.

crates/ty_project/src/metadata/options.rs
mmlb: Going back to previous code that was simpler, so this is unnecessary.

mmlb:
I've updated the PR by using a platform specific separator instead of the hard coded `:` and added join_paths based test to ensure correctness. I've also switched the precedence between PYTHONPATH and extra-paths as requested. PTAL.

mmlb:
Looks like windows is failing to replace `C:\...` with `<temp_dir>`. Not sure if its because `insta::with_filters` needs `C:\` or if I did something weird.

crates/ty/tests/cli/python_environment.rs
```diff
@@ -1915,7 +1915,77 @@ fn pythonpath_is_respected() -> anyhow::Result<()> {
...
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    All checks passed!
+
+    ----- stderr -----
+    WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
+    "#);
+
+    let case = CliTest::with_files([
```
MichaReiser, line 1934: I'm not sure what's different with this test that the path filtering doesn't work. Can you try splitting this second case into its own test, just to ensure that two `CliTest` don't interfere with each other (in some unfortunate way?)

MichaReiser: Glad that it worked. I'm not sure why because `CliTest` does use `add_filter`

mmlb: Hmm looks like that worked for some reason...

crates/ty_project/src/metadata/options.rs
```diff
@@ -296,14 +297,22 @@ impl Options {
...
+            .collect();
 
         // read all the paths off the PYTHONPATH environment variable, check
         // they exist as a directory, and add them to the vec of extra_paths
         // as they should be checked before site-packages just like python
         // interpreter does
-        if let Ok(python_path) = system.env_var("PYTHONPATH") {
-            for path in python_path.split(':') {
+        if let Ok(python_path) = system.env_var(EnvVars::PYTHONPATH) {
+            const SEP: &str = if cfg!(windows) { ";" } else { ":" };
```
MichaReiser: I think we may still want to use `split_paths` here for:

> This also performs unquoting on Windows.

mmlb: Ah I didn't see that previously.

mmlb: So with that known, I'm thinking of going back to std::env::split_paths will that work for wasm (I think that's the only potential issue, LSP, Test, Memory Systems don't really influence here specifically)? I'll convert to SystemPath too.

mmlb:
Ok looks like CI is happy and I've gone back to std::env::split_paths with `SystemPath::from_std_path` to avoid allocating on valid utf8. If thats all good I'll squash the shouty commits into appropriate parent and re-push.

PR #20495, opened 2025-09-21
[ty] Add positional-only-parameter-as-kwarg error

fgiacome:
## Summary

Fixes astral-sh/ty#1129

It adds a new lint `positional-only-parameter-as-kwarg` which is shown in place of `missing-argument` and `unknown-parameter` when a positional-only parameter is supplied as a keyword argument.

## Test Plan

Test cases for this error existed, so the existing test cases were updated to reflect the new output.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -19,8 +19,8 @@ use crate::place::{Boundness, Place};
 use crate::types::call::arguments::{Expansion, is_expandable_type};
 use crate::types::diagnostic::{
     CALL_NON_CALLABLE, CONFLICTING_ARGUMENT_FORMS, INVALID_ARGUMENT_TYPE, MISSING_ARGUMENT,
-    NO_MATCHING_OVERLOAD, PARAMETER_ALREADY_ASSIGNED, TOO_MANY_POSITIONAL_ARGUMENTS,
-    UNKNOWN_ARGUMENT,
+    NO_MATCHING_OVERLOAD, PARAMETER_ALREADY_ASSIGNED, POSITIONAL_ONLY_PARAMETER_AS_KWARG,
+    TOO_MANY_POSITIONAL_ARGUMENTS, UNKNOWN_ARGUMENT,
 };
 use crate::types::enums::is_enum_class;
 use crate::types::function::{
@@ -1991,6 +1991,7 @@ struct ArgumentMatcher<'a, 'db> {
 
     argument_matches: Vec<MatchedArgument<'db>>,
     parameter_matched: Vec<bool>,
+    suppress_missing_error: Vec<bool>,
     next_positional: usize,
     first_excess_positional: Option<usize>,
     num_synthetic_args: usize,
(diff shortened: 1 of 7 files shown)


MichaReiser:
Can you tell me a bit more about the motivation for a new diagnostic ID?

I think I'd prefer to keep this an `unknown-argument` diagnostic but with a more specific message (the message your new diagnostic uses). 

I'm also curious to hear other opinions on if this should be a new diagnostic kind or not.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1991,6 +1991,7 @@ struct ArgumentMatcher<'a, 'db> {
 
     argument_matches: Vec<MatchedArgument<'db>>,
     parameter_matched: Vec<bool>,
+    suppress_missing_error: Vec<bool>,
```
AlexWaygood, line 1994: we might be able to do this with fewer allocations if we had a single field

```suggestion
    argument_info: Vec<ArgumentInfo<'db>>,
```

where `ArgumentInfo` is a new struct that looks like this:

```rs
struct ArgumentInfo<'db> {
    argument_matches: MatchedArgument<'db>,
    parameter_matched: bool,
    suppress_missing_error: bool,
}
```

this issue sort-of predates your PR, though, so it's fine to disregard this comment -- we can tackle it in a followup!

fgiacome:
Thank you for your feedback!

> Can you tell me a bit more about the motivation for a new diagnostic ID?
> 
> I think I'd prefer to keep this an `unknown-argument` diagnostic but with a more specific message (the message your new diagnostic uses).

My thinking was that `unknown-argument` does not suggest that we might be suppressing a `missing-argument` error for the sake of brevity, while the new diagnostic redirects the user to check the required positional parameters and not only the arguments. What I'm unsure of is the name, it feels inelegant but I couldn't come up with something better (I suppose that `positional-only-parameter-passed-as-keyword-argument` is too long).

crates/ty_python_semantic/src/types/call/bind.rs
fgiacome: Thank you! I will update the PR as soon as I find the time, but I might be busy for the next few days. I can also follow up in a new PR if everything else looks good.

carljm: This seems mostly independent of this PR, so I'd rather merge this now and consider that as a follow-up rather than hold this PR for that improvement.

MichaReiser:
@AlexWaygood's reasoning makes sense to me. I'd be open to an `invalid-argument` rule name if we have other use cases where an extra rule code seems a bit much. For example, what about the opposite where a keyword argument is passed for a positional-only parameter?

AlexWaygood:
> For example, what about the opposite where a keyword argument is passed for a positional-only parameter?

I think our behaviour on `main` is already reasonable for that case:

```py
def f(x, /, *, y): ...

# error[missing-argument]: No argument provided for required parameter `y` of function `f`
# error[too-many-positional-arguments] Too many positional arguments to function `f`: expected 1, got 2
f(42, 36)
```

It doesn't feel to me like we can be confident here in saying that the additional positional-only argument in this call was definitely meant to be for the `y` parameter. We can be certain of that when a positional-only argument is passed by keyword, because you have to explicitly name the parameter the argument is intended to satisfy when you're passing a parameter by keyword. But the same isn't true when passing an argument positionally.

MichaReiser:
Alright. I'm fine going ahead with this

PR #20497, opened 2025-09-21
[`pyupgrade`] Fix `UP008` to not apply when `__class__` is a local variable (`UP008`)

### LINKED ISSUE #20491: UP008 should not apply when `__class__` is a local variable
### Summary

When the first argument to `super` in a method is a local variable named `__class__`, [`super-call-with-parameters` (UP008)](https://docs.astral.sh/ruff/rules/super-call-with-parameters/) should not apply. [Example](https://play.ruff.rs/9f505e06-2c92-4c72-a50d-5f92b2931a69):
```console
$ cat >up008.py <<'# EOF'
class A:
    def f(self):
        return 1
class B(A):
    def f(self):
        return 2
class C(B):
    def f(self):
        __class__ = B
        return super(__class__, self).f()
print(C().f())
# EOF

$ python up008.py
1

$ ruff --isolated check up008.py --select UP008 --unsafe-fixes --fix

$ python up008.py 2>&1 | tail -n 1
RuntimeError: super(): __class__ cell not found
```

This isn’t a problem in the usual case where `__class__` is nonlocal, whether explicitly via `nonlocal __class__` or implicitly by not being assigned to.

### Version

danparizher:
## Summary

Fixes #20491

### DIFF
--- a/crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
+++ b/crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
@@ -140,6 +140,9 @@ pub(crate) fn super_call_with_parameters(checker: &Checker, call: &ast::ExprCall
     };
 
     if !((first_arg_id == "__class__"
+            // If the first argument is __class__, check if it's a local variable.
+            // If so, don't apply UP008.
+            && !checker.semantic().current_scope().has("__class__")
         || (first_arg_id == parent_name.as_str()
             // If the first argument matches the class name, check if it's a local variable
             // that shadows the class name. If so, don't apply UP008.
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/pyupgrade/rules/super_call_with_parameters.rs
```diff
@@ -140,6 +140,9 @@ pub(crate) fn super_call_with_parameters(checker: &Checker, call: &ast::ExprCall
     };
 
     if !((first_arg_id == "__class__"
+            // If the first argument is __class__, check if it's a local variable.
```
ntBre: Actually, I think we can now refactor this condition to

```rust
    if !((first_arg_id == "__class__" || first_arg_id == parent_name.as_str())
        && !checker.semantic().current_scope().has(first_arg_id)
        && second_arg_id == parent_arg.name().as_str())
```

Since `first_arg_id` will equal `__class__` in that case, we can just reuse the `has(first_arg_id)` check, unless I'm missing something.

I dropped the comments for convenience while editing this, but I think it's nice to have them here, possibly combining them if the check is combined too.

PR #20507, opened 2025-09-22
[`flake8-bugbear`] Fix false positive when `lambda` parameters have same name as loop variables (`B023`)

### LINKED ISSUE #15716: B023 False positive when there is a list comprehension with the same variable name in a loop.
### Description

__minimal code snippet:__
```py
for _ in range(3):
    [x for x in []]
    def func():
        lambda x: x
```
__command:__
```
ruff check --isolated --select B023 demo.py
```
__log:__
```
demo.py:5:19: B023 Function definition does not bind loop variable `x`
  |
4 |     def func():
5 |         lambda x: x
  |                   ^ B023
  |

Found 1 error.
```
__ruff version:__ 0.9.2

This is NOT a duplicate of https://github.com/astral-sh/ruff/issues/7847, because `x` is only used in a list comprehension in the same loop. Ruff should not raise a false positive in this situation.

danparizher:
## Summary

Fixes #15716

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
@@ -192,7 +192,7 @@ for _ in range(3):
         lambda x: x  # Should not trigger B023 - x is a lambda parameter
 
 
-# Test case from issue comment - pandas apply (should still trigger B023 for legitimate cases)
+# Test case from issue comment - pandas apply (should NOT trigger B023 - apply is safe like map)
 import pandas as pd
 
 data = pd.DataFrame()
@@ -205,16 +205,16 @@ def modifier(value):
 
 for _i in range(0, 4):
     data[f"v{_i}"] = data["hex"].apply(
-        lambda x: modifier(x[2 * _i : 2 * _i + 2]),  # Should trigger B023 - _i is not bound
+        lambda x: modifier(x[2 * _i : 2 * _i + 2]),  # Should NOT trigger B023 - apply is safe
     )
 
 
-# Test case from issue comment - nested function (should still trigger B023 for legitimate cases)
+# Test case from issue comment - nested function (should NOT trigger B023 - value is function parameter)
 for _ in range(2):
 
     def add_one():
         def _add_one_inner(value):
-            return value + 1  # Should trigger B023 - value is not bound to loop variable
+            return value + 1  # Should NOT trigger B023 - value is function parameter
 
         return _add_one_inner
 
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
@@ -183,3 +183,41 @@ for val in range(3):
 
 
     funcs.append(make_func())
+
+
+# Test cases for issue #15716 - false positives with lambda parameters
+for _ in range(3):
+    [x for x in []]
+    def func():
+        lambda x: x  # Should not trigger B023 - x is a lambda parameter
+
+
+# Test case from issue comment - pandas apply (should still trigger B023 for legitimate cases)
+import pandas as pd
+
+data = pd.DataFrame()
+data.loc[0, "hex"] = "72756666"
+data.loc[1, "hex"] = "75767576"
+
+def modifier(value):
+    return chr(int(value, 16))
+
+
+for _i in range(0, 4):
+    data[f"v{_i}"] = data["hex"].apply(
+        lambda x: modifier(x[2 * _i : 2 * _i + 2]),  # Should trigger B023 - _i is not bound
+    )
+
+
+# Test case from issue comment - nested function (should still trigger B023 for legitimate cases)
+for _ in range(2):
+
+    def add_one():
+        def _add_one_inner(value):
+            return value + 1  # Should trigger B023 - value is not bound to loop variable
+
+        return _add_one_inner
+
+    for value in range(5):
+        result = add_one()(value)
+        print(result)
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
@@ -221,3 +221,15 @@ for _ in range(2):
     for value in range(5):
         result = add_one()(value)
         print(result)
+
+
+# nested function that captures loop variable (SHOULD trigger B023)
+lst = []
+for value in range(2):
+    def add_one():
+        def _add_one_inner():
+            return value + 1  # Should trigger B023 - value is loop variable, not bound
+
+        return _add_one_inner
+
+    lst.append(add_one())
--- a/crates/ruff_linter/src/rules/flake8_bugbear/rules/function_uses_loop_variable.rs
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/rules/function_uses_loop_variable.rs
@@ -64,7 +64,7 @@ struct LoadedNamesVisitor<'a> {
 /// `Visitor` to collect all used identifiers in a statement.
 impl<'a> Visitor<'a> for LoadedNamesVisitor<'a> {
     fn visit_stmt(&mut self, stmt: &'a Stmt) {
-        // Don't visit nested function definitions
+        // Skip nested function definitions - they are handled separately by `SuspiciousVariablesVisitor`
         if stmt.is_function_def_stmt() {
             return;
         }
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
```diff
@@ -183,3 +183,41 @@ def tmp():
...
+data.loc[1, "hex"] = "75767576"
+
+def modifier(value):
+    return chr(int(value, 16))
+
+
+for _i in range(0, 4):
+    data[f"v{_i}"] = data["hex"].apply(
+        lambda x: modifier(x[2 * _i : 2 * _i + 2]),  # Should trigger B023 - _i is not bound
+    )
```
ntBre, line 209: I actually think this _is_ a false positive that should be treated like the `min`, `max`, and `sorted` examples starting on line 124 of this file. Based on the issue comment, this works correctly, I assume because `apply` uses the lambda immediately rather than storing a reference to it like the problematic cases.

crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B023.py
```diff
@@ -183,3 +183,41 @@ def tmp():
...
+
+    def add_one():
+        def _add_one_inner(value):
+            return value + 1  # Should trigger B023 - value is not bound to loop variable
+
+        return _add_one_inner
+
+    for value in range(5):
+        result = add_one()(value)
+        print(result)
```
ntBre: I also don't think this case should trigger the rule, as mentioned on the issue too.

ntBre (changes requested):
Thanks for tackling such a long-standing issue! However, I don't think this resolves the problems reported in #15716. Two of the three false positives are still present, and I don't think explicitly filtering out lambda parameters is really the right solution to that false positive either.

I consider the problem in the initial report to be the "leaking" of the comprehension scope. `x` is not the `for` loop variable, and the `x` in the comprehension isn't actually in scope here:

```py
for _ in range(3):
    [x for x in []]
    def func():
        lambda x: x
```

```pycon
>>> [x for x in []]
[]
>>> x
    x
```

I haven't really looked closely enough at the rule to have a good suggestion for an alternative, but the idea that comes to mind from reading the comments in `function_uses_rule_variable` is that we're missing a check that the variable is actually in scope. That may not require explicit scope tracking; maybe the `SuspiciousVariableVisitor` and/or `AssignedNamesVisitor` are too eager in some way. That's what I would look into instead of tracking a separate vec of lambda parameters. This also triggers B023 currently without being a lambda:

```py
for _ in range(3):
    [x for x in []]
    def func():
        def f(x): x
```

I didn't look as much into the other false positives, but we should try to see how we filter out the other cases (`min`, `max`, `sorted`, etc.) and see if we can apply that here too

crates/ruff_linter/src/rules/flake8_bugbear/rules/function_uses_loop_variable.rs
```diff
@@ -167,6 +176,13 @@ impl<'a> Visitor<'a> for SuspiciousVariablesVisitor<'a> {
                                     }
                                 }
                             }
+                        } else if attr == "apply" {
```
ntBre, line 180: Ah I didn't realize we had a list of allowed functions. I'm not totally sure we should add a special case for pandas, but we do have other pandas-specific rules so it's not without precedent.

If we do add this, we should at least check that pandas has been imported. I think that's what we do in some other cases where we inspect pandas attributes. I believe as written this would allow any type with an `apply` attribute, which is a bit more permissive than necessary.

Ideally we would resolve the type of the value and see that it's a pandas object with a known `apply` method.

crates/ruff_linter/src/rules/flake8_bugbear/snapshots/ruff_linter__rules__flake8_bugbear__tests__B023_B023.py.snap
```diff
@@ -244,3 +244,14 @@ B023 Function definition does not bind loop variable `i`
...
+B023 Function definition does not bind loop variable `val`
+   --> B023.py:180:26
+    |
+178 |     def make_func(val=val):
+179 |         def tmp():
+180 |             return print(val)
+    |                          ^^^
+181 |
+182 |         return tmp
+    |
```
ntBre: This is a new false positive compared to main on a pre-existing test.

ntBre (changes requested):
I'm still skeptical that filtering out a list of lambda parameters is the right approach, and there's a new issue caused by the last round of changes.

crates/ruff_linter/src/rules/flake8_bugbear/rules/function_uses_loop_variable.rs
```diff
@@ -319,8 +303,31 @@ impl<'a> Visitor<'a> for AssignedNamesVisitor<'a> {
...
+    }
+
+    // Collect suspicious variables
     let suspicious_variables = {
-        let mut visitor = SuspiciousVariablesVisitor::default();
+        let mut visitor = SuspiciousVariablesVisitor {
+            names: Vec::new(),
+            safe_functions: safe_functions.clone(),
+            apply_calls: Vec::new(),
+        };
```
ntBre: We definitely shouldn't run the visitor twice. The `seen_module` check is just a check of a bit flag, and the method could even be const from what I can tell (though it isn't marked as such currently), so it should be fine either to check it each time we hit `apply` or to save the result of calling it once. Either way should be much cheaper than two visits.

crates/ruff_linter/src/rules/flake8_bugbear/rules/function_uses_loop_variable.rs
```diff
@@ -62,22 +63,31 @@ struct LoadedNamesVisitor<'a> {
 
 /// `Visitor` to collect all used identifiers in a statement.
 impl<'a> Visitor<'a> for LoadedNamesVisitor<'a> {
+    fn visit_stmt(&mut self, stmt: &'a Stmt) {
+        // Don't visit nested function definitions
+        if stmt.is_function_def_stmt() {
+            return;
+        }
+        visitor::walk_stmt(self, stmt);
+    }
```
ntBre, line 72: I think this check is also incorrect. Using a slight variation on one of the examples from the issue:

```py
lst = []
for value in range(2):
    def add_one():
        def _add_one_inner():
            return value + 1

        return _add_one_inner

    lst.append(add_one())

for l in lst:
    print(l())
```

This _should_ trigger B023 because the inner function captures the loop variable. Running this code produces:

```pycon
2
2
```

but no diagnostic is emitted on this branch.

ntBre (changes requested):
I still don't think either of the function changes is the right fix. I still stand by my [first comment](https://github.com/astral-sh/ruff/pull/20507#pullrequestreview-3290071170) saying that the leaking scope is  the real problem. Have you looked into that at all instead of trying to filter out lambda parameters and avoiding visiting nested functions?

The  pandas issue seems a bit different and might be worth landing separately, at least once we avoid visiting twice as noted inline. But that will not close #15716, which is really about the scope problem above. I'm still not 100% sold on that either because the other functions we special-case here are from the standard library. It could be a slippery slope to start adding third-party functions.

ntBre:
@amyreese and I took another look at this today, and I think the approach here makes a bit more sense to me now. However, we were wondering if there might be another solution, potentially with a larger refactor of the rule itself. Going back to the main case from the issue:

```py
for _ in range(3):
    [x for x in []]
    def func():
        lambda x: x
```

the `x` in the lambda body is currently flagged as suspicious by the first visitor in the rule, the `SuspiciousVariablesVisitor`, because it's a load reference to a variable without a store reference inside the body of `func`. Then, the second visitor, `AssignedNamesVisitor` finds a _different_ `x` within the loop, the target of the list comprehension.

These two are only equal, and thus trigger the `reassigned_in_loop.contains` check, because the rule tracks all of the names as simple strings instead of preserving any scope or even range information.

I guess this is similar to my first suggestion, but to me, the better fix will be something like confirming that the unmatched load reference actually loads a variable defined in the loop and accessible in the same scope as the load reference. The `x` in the comprehension here is not accessible to the lambda, and the same is true in the loop case from https://github.com/astral-sh/ruff/issues/15716#issuecomment-3304010685.

I think the test cases here are good and helpful, but I would still prefer a different, more robust implementation that fixes the underlying issue. We could also consider spinning off the pandas fix, which is pretty separate from the main issue in #15716 to me, if we want to land that earlier.

ntBre:
I'll close this for now due to inactivity.

PR #20508, opened 2025-09-22
[`flake8-pyi`] Fix operator precedence by adding parentheses when needed (`PYI061`)

### LINKED ISSUE #20265: PYI061 fix can change operator precedence
### Summary

The fix for [`redundant-none-literal` (PYI061)](https://docs.astral.sh/ruff/rules/redundant-none-literal/) in Python 3.10 and later changes a subscription into a `|` expression, which has different precedence, which can change the program’s behavior outside of typing contexts. The fix should insert parentheses when necessary to maintain the correct precedence, i.e. `(Literal[1] | None).__dict__` in [this example](https://play.ruff.rs/4622390b-e4a0-44c0-bce3-e592c36c1f6d):
```console 
$ cat >pyi061.py <<'# EOF'
from typing import Literal
print(Literal[1, None].__dict__)
# EOF

$ python pyi061.py
{'_inst': True, '_name': None, '__origin__': typing.Literal, '__slots__': None, '__args__': (1, None), '__parameters__': (), '__module__': 'typing'}

$ ruff --isolated check pyi061.py --select PYI061 --preview --target-version py310 --fix

$ cat pyi061.py 
from typing import Literal
print(Literal[1] | None.__dict__)

$ python pyi061.py 2>&1 | tail -n 1
AttributeError: 'NoneType' object has no attribute '__dict__'. Did you mean: '__dir__'?
```

### Version

danparizher:
## Summary

Fixes #20265

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI061.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI061.py
@@ -78,3 +78,10 @@ b: None | Literal[None] | None
 c: (None | Literal[None]) | None
 d: None | (Literal[None] | None)
 e: None | ((None | Literal[None]) | None) | None
+
+# Test cases for operator precedence issue (https://github.com/astral-sh/ruff/issues/20265)
+print(Literal[1, None].__dict__)  # Should become (Literal[1] | None).__dict__
+print(Literal[1, None].method())  # Should become (Literal[1] | None).method()
+print(Literal[1, None][0])  # Should become (Literal[1] | None)[0]
+print(Literal[1, None] + 1)  # Should become (Literal[1] | None) + 1
+print(Literal[1, None] * 2)  # Should become (Literal[1] | None) * 2
--- a/crates/ruff_linter/src/rules/flake8_pyi/rules/redundant_none_literal.rs
+++ b/crates/ruff_linter/src/rules/flake8_pyi/rules/redundant_none_literal.rs
@@ -4,6 +4,7 @@ use ruff_python_ast::{
     self as ast, Expr, ExprBinOp, ExprContext, ExprNoneLiteral, Operator, PythonVersion,
     helpers::{pep_604_union, typing_optional},
     name::Name,
+    operator_precedence::OperatorPrecedence,
 };
 use ruff_python_semantic::analyze::typing::{traverse_literal, traverse_union};
 use ruff_text_size::{Ranged, TextRange};
@@ -238,7 +239,14 @@ fn create_fix(
                 node_index: ruff_python_ast::AtomicNodeIndex::NONE,
             });
             let union_expr = pep_604_union(&[new_literal_expr, none_expr]);
-            let content = checker.generator().expr(&union_expr);
+
+            // Check if we need parentheses to preserve operator precedence
+            let content = if needs_parentheses_for_precedence(semantic, &union_expr) {
+                format!("({})", checker.generator().expr(&union_expr))
+            } else {
+                checker.generator().expr(&union_expr)
+            };
+
             let union_edit = Edit::range_replacement(content, literal_expr.range());
             Fix::applicable_edit(union_edit, applicability)
         }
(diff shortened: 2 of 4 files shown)


crates/ruff_linter/src/rules/flake8_pyi/rules/redundant_none_literal.rs
```diff
@@ -256,3 +264,35 @@ enum UnionKind {
     TypingOptional,
     BitOr,
 }
+
+/// Check if the union expression needs parentheses to preserve operator precedence.
+/// This is needed when the union is part of a larger expression where the `|` operator
+/// has lower precedence than the surrounding operations (like attribute access).
+fn needs_parentheses_for_precedence(
+    semantic: &ruff_python_semantic::SemanticModel,
+    _union_expr: &Expr,
```
ntBre: If this variable isn't used we should remove it from the signature. Or should it have been used?

crates/ruff_linter/src/rules/flake8_pyi/rules/redundant_none_literal.rs
```diff
@@ -256,3 +264,35 @@ enum UnionKind {
...
+) -> bool {
+    // Get the parent expression to check if we're in a context that needs parentheses
+    let Some(parent_expr) = semantic.current_expression_parent() else {
+        return false;
+    };
+
+    // Check if the parent expression is an attribute access, call, or subscript
+    // These operations have higher precedence than the `|` operator
+    match parent_expr {
```
ntBre: Do we need to match here? It looks like `OperatorPrecedence::CallAttribute` would work fine if we just did `OperatorPrecedence::from(parent_expr)`, but I could be missing something.

crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI061.py
```diff
@@ -78,3 +78,10 @@ def good_func(arg1: Literal[int] | None):
 c: (None | Literal[None]) | None
 d: None | (Literal[None] | None)
 e: None | ((None | Literal[None]) | None) | None
+
+# Test cases for operator precedence issue (https://github.com/astral-sh/ruff/issues/20265)
+print(Literal[1, None].__dict__)  # Should become (Literal[1] | None).__dict__
+print(Literal[1, None].method())  # Should become (Literal[1] | None).method()
+print(Literal[1, None][0])  # Should become (Literal[1] | None)[0]
+print(Literal[1, None] + 1)  # Should become (Literal[1] | None) + 1
+print(Literal[1, None] * 2)  # Should become (Literal[1] | None) * 2
```
ntBre, line 87: One additional test case would be one where the `Literal` is already parenthesized, to make sure we don't add a duplicate set.

crates/ruff_linter/src/rules/flake8_pyi/snapshots/ruff_linter__rules__flake8_pyi__tests__PYI061_PYI061.py.snap
```diff
@@ -510,10 +513,27 @@ PYI061 [*] Use `Literal[...] | None` rather than `Literal[None, ...]`
...
+87 | print(Literal[1, None] * 2)  # Should become (Literal[1] | None) * 2
+88 | print((Literal[1, None]).__dict__)  # Should become ((Literal[1] | None)).__dict__
+   |                   ^^^^
+   |
+help: Replace with `Literal[...] | None`
+85 | print(Literal[1, None][0])  # Should become (Literal[1] | None)[0]
+86 | print(Literal[1, None] + 1)  # Should become (Literal[1] | None) + 1
+87 | print(Literal[1, None] * 2)  # Should become (Literal[1] | None) * 2
+   - print((Literal[1, None]).__dict__)  # Should become ((Literal[1] | None)).__dict__
+88 + print(((Literal[1] | None)).__dict__)  # Should become ((Literal[1] | None)).__dict__
```
ntBre: This added another layer of parentheses, which I was afraid of. If the expression is already parenthesized we should avoid adding another layer. I think you can use the `parenthesized_range` helper to check.

ntBre (approved):

PR #20511, opened 2025-09-22
[ty] Match variadic argument to variadic parameter

dhruvmanila:
## Summary

Closes: https://github.com/astral-sh/ty/issues/1236

This PR fixes a bug where the variadic argument wouldn't match against the variadic parameter in certain scenarios.

This was happening because I didn't realize that the `all_elements` iterator wouldn't keep on returning the variable element (which is correct, I just didn't realize it back then).

I don't think we can use the `resize` method here because we don't know how many parameters this variadic argument is matching against as this is where the actual parameter matching occurs.

## Test Plan

Expand test cases to consider a few more combinations of arguments and parameters which are variadic.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/function.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/function.md
@@ -694,11 +694,23 @@ def _(
 
 ### Mixed argument and parameter containing variadic
 
+```toml
+[environment]
+python-version = "3.11"
+```
+
 ```py
 def f(x: int, *args: str) -> int:
     return 1
 
-def _(args1: list[int], args2: tuple[int], args3: tuple[int, int], args4: tuple[int, ...]) -> None:
+def _(
+    args1: list[int],
+    args2: tuple[int],
+    args3: tuple[int, int],
+    args4: tuple[int, ...],
+    args5: tuple[int, *tuple[str, ...]],
+    args6: tuple[int, int, *tuple[str, ...]],
+) -> None:
     # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
     reveal_type(f(*args1))  # revealed: int
 
@@ -711,6 +723,13 @@ def _(args1: list[int], args2: tuple[int], args3: tuple[int, int], args4: tuple[
 
     # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
     reveal_type(f(*args4))  # revealed: int
+
+    # The first element of the tuple matches the required argument;
+    # all subsequent elements match the variadic argument
+    reveal_type(f(*args5))  # revealed: int
+
+    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
+    reveal_type(f(*args6))  # revealed: int
 ```
 
 ### Keyword argument, positional-or-keyword parameter


AlexWaygood (commented):
Nice! I think this approach still has some false negatives on "mixed tuples" with fixed-length suffixes. E.g. here's the current behaviour on this branch:

```py
def func(*args: str): ...

def _(
    a: tuple[str, ...],
    b: tuple[str, *tuple[str, ...]],
    c: tuple[str, *tuple[str, ...], str],
    d: tuple[int, *tuple[str, ...]],
    e: tuple[int, *tuple[str, ...], str],
    f: tuple[*tuple[str, ...], str],
    g: tuple[*tuple[str, ...], int],
    h: tuple[int, *tuple[str, ...], int],
):
    func(*a)  # no error (good)
    func(*b)  # no error (good)
    func(*c)  # no error (good)
    func(*d)  # error (good)
    func(*d)  # error (good)
    func(*e)  # error (good)
    func(*f)  # no error (good)
    func(*g)  # <-------------  no error (bad!)
    func(*h)  # error (good)
```

dhruvmanila:
I think the core issue here is that we've a way to structure a one-to-many mapping like `*args` argument to multiple parameters via the `MatchedArgument` but we don't have a way to structure a many-to-one mapping which will occur in this scenario where a variadic argument is being unpacked into a variadic parameter. I'm thinking if it's actually required or not, and if so how to structure it.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2465,37 +2493,15 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+            self.argument_matches[argument_index].iter()
         {
             self.check_argument_type(
                 adjusted_argument_index,
                 argument,
-                *argument_type,
-                *parameter_index,
+                variadic_argument_type.unwrap_or_else(Type::unknown),
+                parameter_index,
             );
```
dhruvmanila, line 2505: The reason this isn't required now is because the unpacking of a variadic argument is being done during parameter matching itself. This was added in https://github.com/astral-sh/ruff/pull/20130 to make sure that specialization builder has access to the correct type for each parameter.

The `MatchedArgument` already stores the unpacked types for a variadic argument so we can directly use that here instead.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2135,24 +2135,36 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
...
 
+    /// Match a variadic argument to the remaining positional, standard or variadic parameters.
     fn match_variadic(
         &mut self,
         db: &'db dyn Db,
         argument_index: usize,
         argument: Argument<'a>,
         argument_type: Option<Type<'db>>,
-        length: TupleLength,
     ) -> Result<(), ()> {
```
dhruvmanila, line 2145: The length isn't required here because we already have access to `Tuple<Type>` in this function so we can use that directly to get the `TupleLength`. This also simplifies the overload evaluation where we needed to change this length when argument type expansion lead to retrying of parameter matching.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2163,7 +2175,24 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
...
+        // because a variadic parameter is optional, so if this was done unconditionally, ty could
+        // raise a false positive as "too many arguments".
+        if self.parameters.variadic().is_some() {
+            if let Some(argument_type) = argument_types.next().or(variable_element) {
+                self.match_positional(argument_index, argument, Some(argument_type))?;
+                for argument_type in argument_types {
+                    self.match_positional(argument_index, argument, Some(argument_type))?;
+                }
             }
         }
```
dhruvmanila, line 2197: This is the core logic that would actually match the variadic argument with the variadic parameter in cases where it's missed today. Refer to mdtest examples.

dhruvmanila:
## Ecosystem analysis

All of the new diagnostics in `ignite` are, I think, correct. The annotation of the variadic parameter for `register_events` function is `*event_names: Union[List[str], List[EventEnum]]` which suggests that each element of an unpacking should be a list but the provided parameter is unpacking the enum class directly.

```py
from enum import Enum

class Foo(Enum):
    foo = "foo"
    bar = "bar"

def f(*args: list[Foo]): ...

# This should raise a diagnostic because it unpacks into `Foo` variants but the
# annotated parameter type expects each element to be a `list[Foo]`
f(*Foo)
```

The diagnostic in `vision` is correct as well:
```py
import itertools
import string


def _(*digits: str):
    if not digits:
        digits = string.ascii_lowercase
    else:
        # ty: Object of type `str` is not assignable to `tuple[str, ...]` [invalid-assignment]
        digits = "".join(itertools.chain(*digits))
```

The diagnostic in `bokeh` seems to be related to generics which, if I remember correctly, was brought up during an early discussion regarding how to consider the entire union to map the concrete type so that it doesn't map it to `T` and instead considers `T | list[T]` (the details are a bit hazy in my mind right now).

```py
def f[T](*args: T | list[T]) -> list[T]:
    return []


class Foo: ...


def _(*args: Foo | list[Foo]):
	# This PR reveals this to be `list[Foo | list[Foo]]` and in `bokeh` it then creates a
    # diagnostic when it's being used where `list[Foo]` is expected.
    reveal_type(f(*args))
```

dhruvmanila:
(This is something that I'd love to get @dcreager review on specifically as the original author for single-starred argument.)

crates/ty_python_semantic/resources/mdtest/call/function.md
```diff
@@ -642,6 +642,77 @@ def f(*args: int) -> int:
...
+
+    # This shouldn't raise an error because the unpacking doesn't match the variadic parameter.
+    reveal_type(f(*args2))  # revealed: int
+
+    # But, this should because the second tuple element is not assignable.
+    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
+    reveal_type(f(*args3))  # revealed: int
+
+    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
+    reveal_type(f(*args4))  # revealed: int
```
AlexWaygood: Could possibly consider adding some mixed-tuple cases here too (but not essential!)

```suggestion
def _(args1: list[int], args2: tuple[int], args3: tuple[int, int], args4: tuple[int, ...], args5: tuple[int, *tuple[str, ...]], args6: tuple[int, int, *tuple[str, ...]]) -> None:
    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
    reveal_type(f(*args1))  # revealed: int

    # This shouldn't raise an error because the unpacking doesn't match the variadic parameter.
    reveal_type(f(*args2))  # revealed: int

    # But, this should because the second tuple element is not assignable.
    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
    reveal_type(f(*args3))  # revealed: int

    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
    reveal_type(f(*args4))  # revealed: int
    
    # The first element of the tuple matches the required argument;
    # all subsequent elements match the variadic argument
    reveal_type(f(*args5))  # revealed: int
    
    # error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `str`, found `int`"
    reveal_type(f(*args6))
```

AlexWaygood (commented):
This looks reasonable to me, thanks! I agree it would be great to get @dcreager's eyes on it, though

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2135,24 +2135,36 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
...
     ) -> Result<(), ()> {
         let tuple = argument_type.map(|ty| ty.iterate(db));
-        let mut argument_types = match tuple.as_ref() {
-            Some(tuple) => Either::Left(tuple.all_elements().copied()),
-            None => Either::Right(std::iter::empty()),
+        let (mut argument_types, length, variable_element) = match tuple.as_ref() {
+            Some(tuple) => (
+                Either::Left(tuple.all_elements().copied()),
+                tuple.len(),
+                tuple.variable_element().copied(),
```
dcreager, line 2151: It looks like you only ever use the variable element to extend the `all_elements` iterator with a never-ending stream of `variable_element` items. You could do the same thing by chaining a `std::iter::repeat` to the end of the `all_elements` iterator. Then you wouldn't need the `.or(variable_element)` calls below. (You might also consider adding that as an additional method on `Tuple` or `TupleSpec`)

crates/ty_python_semantic/src/types/call/bind.rs
dcreager: That's a great catch!

dcreager (approved):
This approach looks good! We still need to know the type of the splatted argument before we can perform parameter matching; otherwise we don't know how many elements are in the argument and therefore how many parameters it will be matched to. I think that's unavoidable, though. (And @ibraheemdev, I think this the main thing that might wreak havoc on your work to use bidirectional type checking with function arguments.)

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: That's a good idea although there's one place where I don't use the `.or(...)` call at the end which just loops over the remaining argument types if any. This is usually the case for suffix elements.

dhruvmanila:
> This approach looks good! We still need to know the type of the splatted argument before we can perform parameter matching; otherwise we don't know how many elements are in the argument and therefore how many parameters it will be matched to. I think that's unavoidable, though.

Yeah, I agree that this is unavoidable.

crates/ty_python_semantic/resources/mdtest/call/function.md
dhruvmanila: Thanks, I've added them.

PR #20513, opened 2025-09-22
Update transitive dependencies

MichaReiser:


sharkdp:
What? Why is there a mypy_primer change?

MichaReiser:
Yeah, I scheduled another mypy primer run because I don't understand this as well. Maybe something something hash map and unstable iteration order?

MichaReiser:
The only two dependencies that stand out are Boxcar and Inventory, as they're used by Salsa.

MichaReiser:
Now it's all green...

PR #20516, opened 2025-09-22
[ty] Enum narrowing eq ne

Guillaume-Fgt:
## Summary

In response to https://github.com/astral-sh/ty/issues/1177. This is my attempt at trying to modify the narrow mechanism of enum.

I implemented it this way because I didn't want to alter `fn is_single_valued` signature as it is used in 103 different places. Instead, I created a `fn is_single_valued_enum` and an optional parameter to modify if necessary the dunder methods checked by `fn overrides_equality`.

## Test Plan
I added a test with the case exposed in the issue mentioned above.


Guillaume-Fgt:
Thank you for your review. I understand the reservations you made. I close the PR.

PR #20517, opened 2025-09-22
[ty] Use `typing.Self` for the first parameter of instance methods

sharkdp:
## Summary

Modify the (external) signature of instance methods such that the first parameter uses `Self` unless it is explicitly annotated.

part of https://github.com/astral-sh/ty/issues/159


crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -208,6 +301,53 @@ class MyMetaclass(type):
...
+`Self` treatment.
+
+```toml
+[environment]
+python-version = "3.11"
+```
+
+```py
+class Explicit:
+    # TODO: Should warn the user if self is overriden with a type that is not subtype of the class
```
Glyphack: This should be supertype instead of subtype?

This is what pyright says:
https://pyright-play.net/?pyrightVersion=1.1.405&code=MYGwhgzhAECCBcBYAUNN0AmBTAZtARmBgBQRYg7zQBEAwtQJTQC0AfNAHID2AdlkqnRCAdKJQpQkGLWKwGVUcKA

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -325,7 +325,7 @@ type A = list[Union["A", str]]
 def f(x: A):
     reveal_type(x)  # revealed: list[A | str]
     for item in x:
-        reveal_type(item)  # revealed: list[A | str] | str
+        reveal_type(item)  # revealed: list[Any | str] | str
```
sharkdp, line 328: This regression is tracked in https://github.com/astral-sh/ty/issues/1157

crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -277,9 +277,9 @@ reveal_type(Person._make(("Alice", 42)))  # revealed: Unknown
 
 person = Person("Alice", 42)
 
+# error: [invalid-argument-type] "Argument to bound method `_asdict` is incorrect: Expected `NamedTupleFallback`, found `Person`"
```
sharkdp: Hm. I looks like the https://github.com/astral-sh/ruff/pull/20328 fix doesn't work here?

crates/ty_python_semantic/resources/mdtest/annotations/self.md
sharkdp: Hmm, in a generic class `C[T]`, `self` may be also be annotated with e.g. `C[int]` (see last example in https://typing.python.org/en/latest/spec/generics.html#use-in-generic-classes). I'm not sure I understand pyright's error message here?

sharkdp: I talked with @AlexWaygood and it seems to us that we can hardly put any restrictions on the annotated type of `self`, so it may be fine to remove this TODO for now. Exceptions might be `self: B` annotation on a class `A`, where `B` is disjoint from `A`. So we could also update the TODO comment to mention this example.
```suggestion
    # TODO: Potentially emit a warning if the annotated type of `self` is disjoint from the method's class
```

crates/ty_python_semantic/resources/mdtest/named_tuple.md
sharkdp: This also seems to cause new diagnostics in the ecosystem, so this should be fixed.

sharkdp: Oh, we still use the `Unknown`-specialization in `determine_upper_bound`

sharkdp: No, the problem was that we eagerly "baked" `NamedTupleFallback` into the signature with the new "no generics here" trick. Fixed now.

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,144 @@ class Outer:
...
+    @property
+    def a_property(self) -> Self:
+        # TODO: Should reveal Self@a_property
+        reveal_type(self)  # revealed: Unknown
+        return self
+
+reveal_type(B().name_does_not_matter())  # revealed: B
+
+# TODO: this should be B
+reveal_type(B().positional_only(1))  # revealed: Unknown
```
sharkdp: This should be fixed. Implicit `self` only works for positional-or-keyword parameters, at the moment.

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,144 @@ class Outer:
...
+
+reveal_type(B().keyword_only(x=1))  # revealed: B
+
+# TODO: this should be B
+reveal_type(B().a_property)  # revealed: Unknown
```
sharkdp, line 156: This should also be fixed. I have not looked into it yet. Maybe it can be deferred to a follow-up.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -1219,17 +1259,84 @@ impl<'db> Parameters<'db> {
...
+                    && parameters.index(arg.name().id()) == Some(0)
+                {
+                    let method_has_self_in_generic_context =
+                        method.signature(db).overloads.iter().any(|s| {
+                            if let Some(context) = s.generic_context {
+                                context
+                                    .variables(db)
+                                    .iter()
+                                    .any(|v| v.typevar(db).kind(db) == TypeVarKind::TypingSelf);
+                                true
```
sharkdp: it looks like this was supposed to be
```suggestion
                                context
                                    .variables(db)
                                    .iter()
                                    .any(|v| v.typevar(db).kind(db) == TypeVarKind::TypingSelf)
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
sharkdp: Fixed.

crates/ty_python_semantic/src/types/signatures.rs
sharkdp: Fixed.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -1188,16 +1226,83 @@ impl<'db> Parameters<'db> {
...
+        let is_static_or_classmethod = method_info
+            .is_some_and(|f| f.method.is_staticmethod(db) || f.method.is_classmethod(db));
+
+        let inferred_annotation = |arg: &ParameterWithDefault| {
+            if let Some(MethodInformation { method, class }) = method_info
+                && !is_static_or_classmethod
+                && arg.parameter.annotation().is_none()
+                && parameters.index(arg.name().id()) == Some(0)
+                // TODO: find out why we break protocol type property tests when we include them here:
+                && !class.is_protocol(db)
```
sharkdp: Will talk to @AlexWaygood about this. I assume that we break assignability/subtyping for protocol since we create `Self` with a bound that refers to the protocol class itself (instead of doing something that is more appropriate for structural types?).

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -4174,6 +4193,91 @@ impl KnownClass {
...
+            | KnownClass::Deque
+            | KnownClass::OrderedDict
+            | KnownClass::VersionInfo
+            | KnownClass::Field
+            | KnownClass::KwOnly
+            | KnownClass::NamedTupleLike
+            | KnownClass::Template
+            | KnownClass::Path
+            | KnownClass::ConstraintSet
+            | KnownClass::InitVar => false,
```
sharkdp, line 4276: I am also fine with moving this down and making it a catch-all case (so we don't have to update it every time we add a `KnownClass`). Seems unlikely that new fallback classes are (ever) added. On the other hand, *if* we ever add another fallback class, it would be important to get this right. And not getting it right can lead to subtle and confusing errors.

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,143 @@ class Outer:
...
+    @staticmethod
+    def a_staticmethod(x: int): ...
+
+a = A()
+
+reveal_type(a.implicit_self())  # revealed: A
+reveal_type(a.implicit_self)  # revealed: bound method A.implicit_self() -> A
+```
+
+Calling an instance method explicitly verifies the first argument:
```
AlexWaygood, line 93: But we also verify the type of the first argument if `self` is passed implicitly -- for example (and I'm very happy to see that we emit a diagnostic here!):

```py
from typing import Never

class Foo:
    def f(self: Never) -> None: ...

# error[invalid-argument-type]: Argument to bound method `f` is incorrect: Expected `Never`, found `Foo`
Foo().f()
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,143 @@ class Outer:
...
+
+```py
+A.implicit_self(a)
+
+# error: [invalid-argument-type] "Argument to function `implicit_self` is incorrect: Argument type `Literal[1]` does not satisfy upper bound `A` of type variable `Self`"
+A.implicit_self(1)
+```
+
+If the method is a class or static method then first argument is not self:
```
AlexWaygood:
```suggestion
If the method is a class or static method then first argument is not inferred as `Self`:
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,143 @@ class Outer:
...
+
+```py
+A.a_classmethod()
+
+A.a_classmethod(a)  # error: [too-many-positional-arguments]
+
+A.a_staticmethod(1)
+a.a_staticmethod(1)
+
+A.a_staticmethod(a)  # error: [invalid-argument-type]
```
AlexWaygood: nit


```suggestion
A.a_classmethod()
A.a_classmethod(a)  # error: [too-many-positional-arguments]
A.a_staticmethod(1)
a.a_staticmethod(1)
A.a_staticmethod(a)  # error: [invalid-argument-type]
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -53,6 +48,143 @@ class Outer:
...
+    def a_property(self) -> Self:
+        # TODO: Should reveal Self@a_property
+        reveal_type(self)  # revealed: Unknown
+        return self
+
+reveal_type(B().name_does_not_matter())  # revealed: B
+
+reveal_type(B().positional_only(1))  # revealed: B
+
+reveal_type(B().keyword_only(x=1))  # revealed: B
```
AlexWaygood, line 153:
```suggestion
reveal_type(B().name_does_not_matter())  # revealed: B
reveal_type(B().positional_only(1))  # revealed: B
reveal_type(B().keyword_only(x=1))  # revealed: B
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -208,6 +340,42 @@ class MyMetaclass(type):
...
+[environment]
+python-version = "3.12"
+```
+
+```py
+from __future__ import annotations
+
+class Explicit:
+    # TODO: We could emit a warning if the annotated type of `self` is disjoint from `Explicit`
+    def bad(self: int) -> None:
```
AlexWaygood: `int` is not disjoint from `Explicit` -- you could maybe have another method with `self` annotated as `bool` (which _is_ disjoint), and put the TODO next to that method instead?

crates/ty_python_semantic/resources/mdtest/annotations/self.md
```diff
@@ -208,6 +340,42 @@ class MyMetaclass(type):
...
+        reveal_type(self)  # revealed: int
+
+    def forward(self: Explicit) -> None:
+        reveal_type(self)  # revealed: Explicit
+
+# error: [invalid-argument-type] "Argument to bound method `bad` is incorrect: Expected `int`, found `Explicit`"
```
AlexWaygood: I think it comes to the same thing from a soundness perspective, but a more user-friendly diagnostic might be to say

```suggestion
# error: [invalid-argument-type] "Argument to bound method `bad` is incorrect: Expected `int & Explicit`, found `Explicit`"
```

to make it clearer to the user that the only valid way to call the method from an instance would be on an instance some class that is _both_ a subclass of `Explicit` and a subclass of `int`

I don't think you need to change anything here now, though you could also add a TODO -- I'll leave it up to you :-)

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: I think this is fine; it's not too burdensome to have to update a few `match`es when we add `KnownClass`es (which anyway isn't very common). And we've had errors elsewhere in the past due to not having exhaustive matches, and then forgetting to update things when we add new variants

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1227,7 +1229,9 @@ impl Display for DisplayParameter<'_> {
         } else if let Some(ty) = self.param.annotated_type() {
             // This case is specifically for the `Callable` signature where name and default value
             // cannot be provided.
-            ty.display_with(self.db, self.settings.clone()).fmt(f)?;
+            if self.param.should_annotation_be_displayed() {
+                ty.display_with(self.db, self.settings.clone()).fmt(f)?;
+            }
```
AlexWaygood: hmm, what do we end up displaying if the parameter is unnamed and `self.param.should_annotation_be_displayed()` returns `false`? That seems like it could produce some interesting results 😄

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -101,6 +101,47 @@ pub(crate) fn bind_typevar<'db>(
...
 
+/// Create a `typing.Self` type variable for a given class.
+pub(crate) fn typing_self<'db>(
+    db: &'db dyn Db,
+    scope_id: ScopeId,
+    typevar_binding_context: Option<Definition<'db>>,
+    class: ClassLiteral<'db>,
+    typevar_to_type: &impl Fn(BoundTypeVarInstance<'db>) -> Type<'db>,
+) -> Option<Type<'db>> {
+    let module = parsed_module(db, scope_id.file(db)).load(db);
```
AlexWaygood: (if we merge https://github.com/astral-sh/ruff/pull/20630 then this call won't be necessary, since you're only passing it to `bind_typevar()` down below)

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -101,6 +101,47 @@ pub(crate) fn bind_typevar<'db>(
...
+    let index = semantic_index(db, scope_id.file(db));
+
+    let typevar = TypeVarInstance::new(
+        db,
+        ast::name::Name::new_static("Self"),
```
AlexWaygood, line 113: if we're looking for speedups: one micro-optimisation (that probably won't make a difference, but might be worth trying?) might be to initialise this as a `const` variable at compile time and then clone it here rather than calling `Name::new_static("Self")` every time

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -1185,16 +1223,72 @@ impl<'db> Parameters<'db> {
...
+                {
+                    let scope_id = definition.scope(db);
+                    let typevar_binding_context = Some(definition);
+                    let index = semantic_index(db, scope_id.file(db));
+                    let class = nearest_enclosing_class(db, index, scope_id).unwrap();
+
+                    Some(
+                        typing_self(db, scope_id, typevar_binding_context, class, &Type::TypeVar)
+                            .unwrap(),
+                    )
```
AlexWaygood, line 1258: nit: could we add some SAFETY comments here about why the `.unwrap()`s are okay? (Or switch to using `.expect()`? Obviously even better would be if we could avoid the `.unwrap()`s entirely ;) )

AlexWaygood:
And thank you so much @Glyphack as well!

crates/ty_python_semantic/resources/mdtest/annotations/self.md
sharkdp: Added this test:
```py
from typing import Never

class Strange:
    def can_not_be_called(self: Never) -> None: ...

# error: [invalid-argument-type] "Argument to bound method `can_not_be_called` is incorrect: Expected `Never`, found `Strange`"
Strange().can_not_be_called()
```

crates/ty_python_semantic/resources/mdtest/annotations/self.md
sharkdp: Decided to leave everything as is for now. Not sure if the intersection type in the error message really helps.

crates/ty_python_semantic/src/types/display.rs
sharkdp: None of our mdtests currently hit the (implicit) `else` branch here, at the moment. I removed the check for now. I thought about adding `debug_assert!(self.param.should_annotation_be_displayed())`, but it seems too annoying even for a debug build to crash because of that.

crates/ty_python_semantic/src/types/generics.rs
sharkdp: We use `Name::new_static` in lots of places. I somehow can't imagine that this will help with performance, but we can try that separately, I guess.

sharkdp:
The performance regressions look severe, but I am going to merge this. We can always work on performance later, and it's probably also important to keep in mind that across the *whole* ecosystem, this is more of a [3% slowdown](https://david-implicit-self-signatur.ecosystem-663.pages.dev/timing).

PR #20518, opened 2025-09-22
[`pylint`] Clarify fix safety to include left-hand hashability (`PLR6201`)

### LINKED ISSUE #20510: PLR6201 should only apply to hashable types
### Summary

# Ruff version

ruff 0.12.12

# Ruff settings

PLR6201 enabled

# Problem

Fix safety description of PLR6201 is misleading:

> [Fix safety](https://docs.astral.sh/ruff/rules/literal-membership/#fix-safety)
> This rule's fix is marked as unsafe, as the use of a set literal will error at runtime if the sequence contains unhashable elements (like lists or dictionaries). While Ruff will attempt to infer the hashability of the elements, it may not always be able to do so.

This reads as if **only** the hashability of the rhs-sequence was the problem. But in my opinion, the hashability of the lhs is more problematic in most use cases. F.i.

```
def foo(val: str | list[str] | dict[str, str]) -> bool:
    return val in {"a", "b", "c"}
```

### Version

ruff 0.12.12

danparizher:
## Summary

Fixes #20510

### DIFF
--- a/crates/ruff_linter/src/rules/pylint/rules/literal_membership.rs
+++ b/crates/ruff_linter/src/rules/pylint/rules/literal_membership.rs
@@ -25,9 +25,11 @@ use crate::{AlwaysFixableViolation, Edit, Fix};
 ///
 /// ## Fix safety
 /// This rule's fix is marked as unsafe, as the use of a `set` literal will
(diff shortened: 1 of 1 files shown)


ntBre (approved):
Perfect, thanks!

PR #19404, opened 2025-07-17
[`flake8-use-pathlib`] Add autofix for `PTH101`, `PTH104`, `PTH105`, `PTH121`

chirizxc:
## Summary

Part of https://github.com/astral-sh/ruff/issues/2331

## Test Plan

`cargo nextest run flake8_use_pathlib`

PR #20520, opened 2025-09-22
[`refurb`] Add fixes for `FURB101`, `FURB103`

chirizxc:
## Summary

Part of `PTH-*` fixes: https://github.com/astral-sh/ruff/pull/19404#issuecomment-3089639686

## Test Plan
`cargo nextest run furb`


chirizxc:
@ntBre I think this PR can include furb103 as well, in essence the code will be almost identical

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -42,11 +45,23 @@ pub(crate) struct ReadWholeFile {
...
     fn message(&self) -> String {
-        let filename = self.filename.truncated_display();
-        let suggestion = self.suggestion.truncated_display();
-        format!("`open` and `read` should be replaced by `Path({filename}).{suggestion}`")
+        format!(
+            "`open` and `read` should be replaced by `Path({}).{}`",
+            self.filename.truncated_display(),
+            self.suggestion.truncated_display(),
+        )
+    }
```
ntBre: I think we might want to preserve the old `message` with the `{filename}` and `{suggestion}` placeholders because they get rendered in the docs website:

Image

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -42,11 +45,23 @@ pub(crate) struct ReadWholeFile {
...
+
+    fn fix_title(&self) -> Option<String> {
```
ntBre, line 59: It's fine here, though. I don't think these are rendered anywhere.

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -126,14 +169,22 @@ fn match_read_call(expr: &Expr) -> Option<&Expr> {
...
-        func: Box::new(name.into()),
+        func: Box::new(
+            ast::ExprName {
+                id: name,
+                ctx: ast::ExprContext::Load,
+                range: TextRange::default(),
+                node_index: ruff_python_ast::AtomicNodeIndex::NONE,
+            }
+            .into(),
+        ),
```
ntBre: I don't really understand this change, it looks like it will result in the same return value. Is that true, or am I missing something?

crates/ruff_linter/src/rules/refurb/rules/write_whole_file.rs
```diff
@@ -43,11 +48,22 @@ pub(crate) struct WriteWholeFile {
...
 impl Violation for WriteWholeFile {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
-        let filename = self.filename.truncated_display();
-        let suggestion = self.suggestion.truncated_display();
-        format!("`open` and `write` should be replaced by `Path({filename}).{suggestion}`")
+        format!(
+            "`open` and `write` should be replaced by `Path({}).{}`",
```
ntBre: Same suggestion as above for the `message` placeholders.

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__preview_FURB103.py.snap
```diff
@@ -0,0 +1,260 @@
...
+37 |     # The body of `with` is non-trivial, but the recommendation holds.
+38 |     bar("pre")
+   |
+help: Replace with `Path(foo()).write_bytes(bar())`
+
+FURB103 `open` and `write` should be replaced by `Path("a.txt").write_text(x)`
+  --> FURB103.py:44:6
+   |
+43 | # FURB103
+44 | with open("a.txt", "w") as a, open("b.txt", "wb") as b:
```
ntBre, line 163: Nice! I was curious how we'd handle this case, but I think it makes sense only to fix the cases where we can remove the whole `with`. It might be worth adding a `## Fix availability` section pointing that out, though.

I think we probably do have all the information needed to remove parts of the `with`, but we can always expand the fix later.

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -1,14 +1,13 @@
-use std::borrow::Cow;
-
-use ruff_python_ast::name::Name;
-use ruff_python_ast::{self as ast, Expr, parenthesize::parenthesized_range};
+use ruff_python_ast::PythonVersion;
+use ruff_python_ast::{self as ast, Expr, Stmt, name::Name, parenthesize::parenthesized_range};
 use ruff_python_codegen::Generator;
 use ruff_python_semantic::{BindingId, ResolvedReference, SemanticModel};
 use ruff_text_size::{Ranged, TextRange};
+use std::borrow::Cow;
```
ntBre: nit: I would put this back at the top, regrouping std, ruff, and crate imports.

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -362,3 +357,78 @@ pub(super) enum IterLocation {
...
+    let replacement = match target {
+        Some(var) => {
+            format!(
+                "{} = {}({}).{}({})",
+                var,
+                binding,
+                filename_code,
+                open.mode.pathlib_method(),
+                args
+            )
```
ntBre: I might be missing something, but I thought we already constructed a replacement using the `Generator`. Can we reuse that snippet that we passed into `SourceCodeSnippet`?

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -362,3 +357,78 @@ pub(super) enum IterLocation {
     Call,
     Comprehension,
 }
+
+pub(super) fn generate_furb_101_103_fix(
```
ntBre: I'm not sure how helpful this helper function really is. There will be a little duplication, but I think the two `match`es on `target` justify keeping these code paths separate. And there will be a bit less duplication in any case if we can reuse the `SourceCodeSnippet` as I mention below.

crates/ruff_linter/src/rules/refurb/mod.rs
```diff
@@ -62,6 +63,24 @@ mod tests {
         Ok(())
     }
 
+    #[test_case(Path::new("FURB101.py"))]
+    #[test_case(Path::new("FURB103.py"))]
+    fn preview_rules(path: &Path) -> Result<()> {
```
ntBre: nit: we might want to make the signature match `rules` above. It's not really helpful right now, but that will make it easier to reuse the function for other rules in the future, if needed.

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
chirizxc: yes

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__preview_FURB103.py.snap
chirizxc: I don't think we should write about it, as it will break the code when trying to fix it

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -42,12 +47,22 @@ pub(crate) struct ReadWholeFile {
...
+
     #[derive_message_formats]
     fn message(&self) -> String {
         let filename = self.filename.truncated_display();
         let suggestion = self.suggestion.truncated_display();
         format!("`open` and `read` should be replaced by `Path({filename}).{suggestion}`")
     }
+
+    fn fix_title(&self) -> Option<String> {
+        Some(format!(
```
chirizxc, line 60: Should we follow the same style in these places?

ntBre: No I think these are okay because only the `message` is rendered in the docs table.

crates/ruff_linter/src/rules/refurb/rules/write_whole_file.rs
```diff
@@ -165,3 +198,46 @@ fn make_suggestion(open: &FileOpen<'_>, arg: &Expr, generator: Generator) -> Sou
...
+
+    if !is_valid_block {
+        return None;
+    }
+
+    let locator = checker.locator();
+    let filename_code = locator.slice(open.filename.range());
+    let call = suggestion
+        .full_display()
+        .unwrap_or(suggestion.truncated_display());
```
ntBre: I don't think we can use this here because it will truncate the `call` with `...` if it's too long. That's fine for the error message but not for the suggested fix, which always needs to be valid code.

I meant that we should pass the `&str` we pass to `SourceCodeSnippet::from_str` into this function as well, not reuse the `SourceCodeSnippet` itself. I think we can just return a `String` from `make_suggestion` to reuse here and only wrap it in a `SourceCodeSnippet` where that's needed.

ntBre (commented):
Thanks! I just had one clarification about earlier suggestion, but this looks good to me otherwise.

PR #20521, opened 2025-09-22
[ty] Fix class literal subtyping with object fallback

MatthewMckee4:
## Summary

@ibraheemdev notes this example failed

```py
from typing import Callable

class X:
    ...

def f(callable: Callable[[], X]) -> X:
    return callable()

x = f(X)
```

Resolves https://github.com/astral-sh/ty/issues/1210

The issue was that we set the `Self` to the class type instead of the instance type of the class.

## Test Plan

Fix tests in `is_subtype_of.md`


MatthewMckee4:
All of the 
```
error[no-matching-overload] No overload of function `field` matches arguments
```
 diagnostics look good, seems like they are mostly for something of the form:

```py
a: A = field(default_factory=A)
```

and the other diagnostics basically saying
```
error[invalid-argument-type] Argument to function ... is incorrect: Expected `() -> A`, found `<class 'A'>`
```

ibraheemdev (approved):

PR #20522, opened 2025-09-22
[ty] Warn incompatible type annotation for self

Glyphack:
## Summary

Part of: https://github.com/astral-sh/ty/issues/159

Report incompatible type annotations for `self` argument.
The type that is used should be either `typing.Self` or super type of the class.


Glyphack:
Closing because of: https://github.com/astral-sh/ruff/pull/20517#discussion_r2372560363
I feel this is not going to find a lot of diagnostics in codebases overall.

PR #20523, opened 2025-09-22
[ty] More precise type inference for dictionary literals

ibraheemdev:
## Summary

Extends https://github.com/astral-sh/ruff/pull/20360 to dictionary literals. This also improves our `TypeDict` support by passing through nested type context.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -4112,7 +4112,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 value,
                 TypeContext::new(Some(declared.inner_type())),
             );
-            let mut inferred_ty = if target
+            let inferred_ty = if target
                 .as_name_expr()
                 .is_some_and(|name| &name.id == "TYPE_CHECKING")
             {
@@ -4122,24 +4122,6 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             } else {
                 inferred_ty
             };
-
-            // Validate `TypedDict` dictionary literal assignments
-            if let Some(typed_dict) = declared.inner_type().into_typed_dict() {
-                if let Some(dict_expr) = value.as_dict_expr() {
-                    validate_typed_dict_dict_literal(
-                        &self.context,
-                        typed_dict,
-                        dict_expr,
-                        target.into(),
-                        |expr| self.expression_type(expr),
-                    );
-
-                    // Override the inferred type of the dict literal to be the `TypedDict` type
-                    // This ensures that the dict literal gets the correct type for key access
-                    let typed_dict_type = Type::TypedDict(typed_dict);
-                    inferred_ty = typed_dict_type;
-                }
-            }
 
             self.add_declaration_with_binding(
                 target.into(),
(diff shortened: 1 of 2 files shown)


ibraheemdev:
It looks like mypy-primer is timing out due to some more issue related to union normalization. I opened https://github.com/astral-sh/ty/issues/1237 to track this issue more generally.

carljm:
It looks like the conformance suite regressions are because we don't understand the function form for creating a `TypedDict` yet. It may be that in order to prevent adding similar false positives in user code, we either need to add basic support for understanding the function form of `TypedDict` (shouldn't be too hard, I don't think?) or at least recognize it and insert a `@Todo` type. (Will be interesting to see if this also shows up as primer false positives, once we get a primer run.)

carljm:
I think we should switch primer to run in release mode and see how much that helps here. IIRC running it in debug mode originally wasn't really due to needing debug info in the run, it was because at the time it was fast enough in debug mode that saving the build time made it faster overall. I doubt that's the case today. We have enough real perf issues to sort out, we don't need to be also blocked by issues that only show up in debug builds.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5303,95 +5306,131 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            items,
+        } = dict;
+
+        let items = items
+            .iter()
+            .map(|item| [item.key.as_ref(), Some(&item.value)]);
+
+        self.infer_collection_literal(items, tcx, KnownClass::Dict)
+            .unwrap_or_else(|| {
+                KnownClass::Set.to_specialized_instance(self.db(), [Type::unknown()])
```
carljm: This looks like a copy-paste error?

We should add a test that would catch this.

ibraheemdev: I don't actually know if this case is ever hit, because the constraint solver only fails if an inferred type fails to meet the upper bound of a type variable.

AlexWaygood:
you probably want to rebase on top of `main` now https://github.com/astral-sh/ruff/pull/20540 has landed

Edit: argh, and I see you filed https://github.com/astral-sh/ruff/pull/20537 before I made that PR, and I didn't see it 🤦

ibraheemdev:
From a quick skim, a lot of the errors seem to be of the form:
```py
def f(a: int, b: str): ...

x = { "a": 1, "b": "2" }

# Expected `int` found `Unknown | int | str`
f(x["a"], x["b"])
# or
f(**x)
```
There were similar errors in the list/set inference PR, but it seems like a more common pattern in the ecosystem for dictionaries. pyright again accepts this, as it doesn't even attempt to infer a union and falls back to `dict[str, Unknown]`, but mypy widens to `dict[str, object]`, and also errors.

carljm:
I can see the rationale for pyright's behavior. This is a case where we can emit a lot of false positives on perfectly working untyped code, just because we don't infer a sufficiently precise type for the dict literal (which is implicitly used as a heterogeneous `TypedDict`). Ideally maybe this would be solved by a more precise internal "implicit TypedDict" type, but that gets really complex to handle with subsequent mutations.

I think for now we should probably go ahead with this as-is, but in the future in "gradual guarantee" mode we may want to emulate pyright for container literals that "look heterogeneous".

carljm (approved):

carljm:
Filed https://github.com/astral-sh/ty/issues/1248 for follow-up later.

ibraheemdev:
> we either need to add basic support for understanding the function form of TypedDict (shouldn't be too hard, I don't think?) or at least recognize it and insert a `@Todo` type

I think this is actually tricky, because we don't currently have infrastructure for any other functional type forms (e.g. `NamedTuple`, or `NewType`), so we should probably wait till something like https://github.com/astral-sh/ruff/pull/20126 lands.

I added a fallback to `dict[Unknown, Unknown]` if the type annotation is a `@Todo` type for now.

carljm:
> I think this is actually tricky, because we don't currently have infrastructure for any other functional type forms

I think the workaround you put in place is fine for this PR.

I am curious what the trickiest part of functional `TypedDict` and `NamedTuple` would be. I think the hard parts of the `NewType` PR are in how to represent the "new-type" itself, which I don't think would be common infrastructure shared with `TypedDict` and `NamedTuple`? I think we'd need a fully-materialized (as opposed to "lazy based on a class definition") representation of those types; similar to `Protocol::Synthesized` vs `Protocol::FromClass`.

Nothing actionable for this PR, I'm just not convinced that landing the `NewType` PR is a pre-requisite to support for functional `TypedDict` or `NamedTuple`. (Which maybe doesn't matter either, since `NewType` may be just as high priority anyway.)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5303,95 +5288,167 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                } else {
+                    self.infer_expression(&item.value, TypeContext::default());
+                }
+            }
+
+            validate_typed_dict_dict_literal(
+                &self.context,
+                typed_dict,
+                dict,
+                dict.into(),
```
ibraheemdev, line 5326: We don't have access to the target binding anymore, so this changes diagnostics slightly. Before:
```
error[missing-typed-dict-key]: Missing required key 'name' in TypedDict `Movie` constructor
 --> x.py:8:5
  |
7 | def func1(variable_key: str):
8 |     movie: Movie = {variable_key: "", "year": 1900}
  |     ^^^^^
```
After:
```
error[missing-typed-dict-key]: Missing required key 'name' in TypedDict `Movie` constructor
 --> x.py:8:20
  |
7 | def func1(variable_key: str):
8 |     movie: Movie = {variable_key: "", "year": 1900}
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I think this makes sense, as it extends nicely to errors in nested `TypedDict` literals.

carljm: I think this is fine.

ibraheemdev:
The performance regression on `freqtrade` comes from inferring the type of [this massive dictionary literal](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/config_schema/config_schema.py#L31), which we were previously ignoring. I'm not sure there's much that can be done there.

PR #20525, opened 2025-09-22
[`ruff`] Suppress diagnostic for f-string interpolations with debug text (`RUF010`)

### LINKED ISSUE #20519: RUF010 diagnostic is unhelpful for an interpolation with debug text
### Summary

[`explicit-f-string-type-conversion` (RUF010)](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/) should be suppressed for interpolations that use `=` because there is no way to use a conversion flag to get the same output, which the user presumably intended, so the diagnostic is not helpful.
```console
$ cat >ruf010.py <<'# EOF'
print(f"{ascii(1)=}")
# EOF

$ python ruf010.py
ascii(1)='1'

$ ruff --isolated check ruf010.py --select RUF010 --output-format concise -q
ruf010.py:1:10: RUF010 Use explicit conversion flag
```
Alternatively, if the diagnostic _is_ deemed helpful, I suggest reverting that piece of #18690 and marking the fix unsafe in this case. Either works; it’s just this middle ground of hinting at a change but not showing it that seems unhelpful.

### Version

danparizher:
## Summary

Fixes #20519

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/ruff/RUF010.py
+++ b/crates/ruff_linter/resources/test/fixtures/ruff/RUF010.py
@@ -56,3 +56,11 @@ f"{str(object=3)}"
 f"{str(x for x in [])}"
 
 f"{str((x for x in []))}"
+
+# Debug text cases - should not trigger RUF010
+f"{str(1)=}"
+f"{ascii(1)=}"
+f"{repr(1)=}"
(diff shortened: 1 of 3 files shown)


ntBre (approved):
Thank you!

PR #20526, opened 2025-09-23
[`playground`]: Fix non‑BMP code point handling in quick‑fixes and markers

### LINKED ISSUE #16266: Playground counts each non-BMP code point as two code points
### Description

The playground counts each non-BMP code point as two code points when applying fixes, leaving characters there that should have been replaced as part of the fix. [Here is an example.](https://play.ruff.rs/8aef5a51-fa9e-4d87-9634-537659f3c65a) The input is:
```python
dict(𠤪=1)
```
Applying the quick fix for [`unnecessary-collection-call` (C408)](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/) incorrectly produces:
```python
{"𠤪": 1})
```
On the command line, though, Ruff 0.9.6 correctly fixes it to:
```python
{"𠤪": 1}
```

danparizher:
## Summary

Fixes #16266

Updated the playground to convert UTF‑32 (code point) columns from the WASM diagnostics/fixes into Monaco’s UTF‑16 columns when constructing both quick‑fix edit ranges and diagnostic markers in `playground/ruff/src/Editor/SourceEditor.tsx`.

With this change, applying the `C408` quick fix on `dict(𠤪=1)` now correctly yields `{"𠤪": 1}` without leaving a stray `)`.

### DIFF
--- a/crates/ruff_wasm/src/lib.rs
+++ b/crates/ruff_wasm/src/lib.rs
@@ -127,7 +127,7 @@ impl Workspace {
     }
 
     #[wasm_bindgen(constructor)]
-    pub fn new(position_encoding: PositionEncoding, options: JsValue) -> Result<Workspace, Error> {
+    pub fn new(options: JsValue, position_encoding: Option<PositionEncoding>) -> Result<Workspace, Error> {
         let options: Options = serde_wasm_bindgen::from_value(options).map_err(into_error)?;
         let configuration =
             Configuration::from_options(options, Some(Path::new(".")), Path::new("."))
@@ -138,7 +138,7 @@ impl Workspace {
 
         Ok(Workspace {
             settings,
-            position_encoding: position_encoding.into(),
+            position_encoding: position_encoding.unwrap_or_default().into(),
         })
     }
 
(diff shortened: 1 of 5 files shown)


crates/ruff_wasm/src/lib.rs
```diff
@@ -228,27 +228,47 @@ impl Workspace {
...
-                            location: source_code.line_column(edit.start()).into(),
-                            end_location: source_code.line_column(edit.end()).into(),
-                            content: edit.content().map(ToString::to_string),
-                        })
-                        .collect(),
-                }),
+            .map(|msg| {
+                let range = msg.range().unwrap_or_default();
+                let start_loc = source_code.source_location(range.start(), PositionEncoding::Utf16);
+                let end_loc = source_code.source_location(range.end(), PositionEncoding::Utf16);
```
MichaReiser: This works but it will break existing clients that assumed UTF8. 

I suggest adding `positionEncoding` a new argument to `new` (similar to ty) and storing it on the `Workspace` and then using it here.

This allows each user to use the encoding that best suits their use case.

PR #20527, opened 2025-09-23
[`playground`] Allow hover quick fixes to appear for overlapping diagnostics

### LINKED ISSUE #19655: ISC002 makes W605 unfixable on playground
### Summary

I'm not sure why this happens, but if both [invalid-escape-sequence (W605)](https://docs.astral.sh/ruff/rules/invalid-escape-sequence/#invalid-escape-sequence-w605) and [multi-line-implicit-string-concatenation (ISC002)](https://docs.astral.sh/ruff/rules/multi-line-implicit-string-concatenation/#multi-line-implicit-string-concatenation-isc002) trigger on the same string on the playground, `W605` will not be fixable using the On Hover context menu -> Quick Fix.

https://play.ruff.rs/83a23eba-a1c5-4394-aab9-dcd20e098a55

Image

This seems to just happen with the playground hover menu, the fix works fine locally in PowerShell with `--fix`, the "Quick Fix" shows up on hover locally in VSC with the VSC extension, and when using the `ctrl+.` shortcut with the cursor over the `W605` error to open the Quick Fix menu directly.

No clue if this is something that is fixable/is a problem on ruff's end, it could also be something in Monaco.

### Version

Playground

danparizher:
## Summary

Fixes #19655

- Switched quick-fix filtering from line-equality to proper range intersection to show all relevant fixes for the hovered range.
- Set `code` on markers with `diagnostic.code ?? undefined` to satisfy Monaco’s `IMarkerData` type and ensure the rule code shows up without type errors.

### DIFF
--- a/playground/ruff/src/Editor/SourceEditor.tsx
+++ b/playground/ruff/src/Editor/SourceEditor.tsx
@@ -125,49 +125,17 @@ class RuffCodeActionProvider implements CodeActionProvider {
   ): languages.ProviderResult<languages.CodeActionList> {
     const actions = this.diagnostics
       // Show fixes for any diagnostic whose range intersects the requested range
-      .filter((check) => {
-        const diagStartLine = check.start_location.row;
-        const diagStartCol = check.start_location.column;
-        const diagEndLine = check.end_location.row;
-        const diagEndCol = check.end_location.column;
-
-        const rangeStartLine = range.startLineNumber;
-        const rangeStartCol = range.startColumn;
-        const rangeEndLine = range.endLineNumber;
-        const rangeEndCol = range.endColumn;
-
-        // Helper to compare positions
-        const isBeforeOrEqual = (
-          lineA: number,
-          colA: number,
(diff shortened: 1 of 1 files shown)


playground/ruff/src/Editor/SourceEditor.tsx
```diff
@@ -124,7 +124,50 @@ class RuffCodeActionProvider implements CodeActionProvider {
...
+          diagStartCol,
+          rangeEndLine,
+          rangeEndCol,
+        );
+        const diagEndsAfterOrAtRangeStart = isAfterOrEqual(
+          diagEndLine,
+          diagEndCol,
+          rangeStartLine,
+          rangeStartCol,
+        );
```
MichaReiser: Can we use `range.containsRange(other_range)` here instead (or the other way round or any other range method)

danparizher:
I did some digging and found `Range.areIntersecting()` which is perfect for this case.

PR #20528, opened 2025-09-23
[ty] bidirectional type inference using function return type annotations

mtshiba:
## Summary

Implements bidirectional type inference using function return type annotations.

This PR was originally proposed to solve astral-sh/ty#1167, but this does not fully resolve it on its own.
Additionally, I believe we need to allow dataclasses to generate their own `__new__` methods, [use constructor return types ​​for inference](https://github.com/mtshiba/ruff/blob/5844c0103d30de4f8890414b38ce1a8d92d6f2d2/crates/ty_python_semantic/src/types.rs#L5326-L5328), and a mechanism to discard type narrowing like `& ~AlwaysFalsy` if necessary (at a more general level than this PR).

## Test Plan

`mdtest/bidirectional.md` is added.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/typed_dict.md
+++ b/crates/ty_python_semantic/resources/mdtest/typed_dict.md
@@ -125,9 +125,10 @@ def homogeneous_list[T](*args: T) -> list[T]:
 reveal_type(homogeneous_list(1, 2, 3))  # revealed: list[Literal[1, 2, 3]]
 plot2: Plot = {"y": homogeneous_list(1, 2, 3), "x": None}
 reveal_type(plot2["y"])  # revealed: list[int]
-# TODO: no error
-# error: [invalid-argument-type]
+
 plot3: Plot = {"y": homogeneous_list(1, 2, 3), "x": homogeneous_list(1, 2, 3)}
+reveal_type(plot3["y"])  # revealed: list[int]
+reveal_type(plot3["x"])  # revealed: list[int] | None
 
 Y = "y"
 X = "x"


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1079,7 +1079,20 @@ impl<'db> SpecializationBuilder<'db> {
...
+            actual = union.filter(self.db, |actual_elem| {
+                !actual_elem.is_disjoint_from(self.db, formal)
+            });
+        }
+
         match (formal, actual) {
+            (Type::Union(_), Type::Union(_)) => {
+                // TODO: We need to infer specializations appropriately.
+                // e.g.
+                // `formal: list[T] | T | U, actual: V | int | list[V]` => `T = V, U = int`
```
mtshiba: I decided that this case requires elaborate special handling, that falling back to the case below would lead to false positive errors, and that we should do nothing at this time.

In the case below, specialization is performed only when there is only one type variable in `formal`, but even so, the following code will not be handled correctly.


```python
def h[T](x: T, cond: bool) -> T | list[T]:
    # error: [invalid-return-type] "Return type does not match returned value: expected `T@h | list[T@h]`, found `T@h | list[T@h] | list[T@h | list[T@h]]`"
    return i(x, cond)

def i[T](x: T, cond: bool) -> T | list[T]:
    return x if cond else [x]
```

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,139 @@
...
+l4: list[int] | None = list1(1)
+reveal_type(l4)  # revealed: list[int]
+
+def _(l: list[int] | None = None):
+    l1 = l or list()
+    reveal_type(l1)  # revealed: (list[int] & ~AlwaysFalsy) | list[Unknown]
+
+    l2: list[int] = l or list()
+    # TODO: should be `list[int]`
+    reveal_type(l2)  # revealed: (list[int] & ~AlwaysFalsy) | list[Unknown]
```
AlexWaygood: I think that's maybe not an awful type to infer here, actually? `(list[int] & ~AlwaysFalsy) | list[Unknown]` is assignable to `list[int]` (and the upper bound of `list[int]` is the only constraint here), so it seems reasonable to infer the more precise type here

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,139 @@
...
+
+def f[T](x: T, cond: bool) -> T | list[T]:
+    return x if cond else [x]
+
+# TODO: no error
+# error: [invalid-assignment] "Object of type `Literal[1] | list[Literal[1]]` is not assignable to `int | list[int]`"
+l5: int | list[int] = f(1, True)
+```
+
+```py
```
AlexWaygood, line 60: I suggest giving this snippet an explicit filename, or it will be implicitly merged with the previous snippet by the mdtest framework, which might produce some surprising results:


````suggestion

`typed_dict.py`:

```py
````

File: crates/ty_python_semantic/resources/mdtest/bidirectional.md [hunk 4]
```diff
@@ -0,0 +1,139 @@
...
+    return {}
+
+def wrap_data() -> list[dict]:
+    if not (res := get_data()):
+        return list1({})
+    reveal_type(list1(res))  # revealed: list[dict[Unknown, Unknown] & ~AlwaysFalsy]
+    # `list[dict[Unknown, Unknown] & ~AlwaysFalsy]` and `list[dict[Unknown, Unknown]]` are incompatible,
+    # but the return type check passes here because the inferred return type is widened
+    # by bidirectional type inference.
+    return list1(res)
```
AlexWaygood: I don't think this quite accurately describes what's going on here. It's not that we infer one type, and then that type is later widened to a second type by bidirectional type inference. Instead, we use bidirectional type inference to ensure that we never infer the narrower type at all -- the outer context is used to ensure that we infer a broader type when solving the type variable in the second `list1()` call

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,139 @@
...
+        reveal_type(list1(res))  # revealed: list[dict[Unknown, Unknown] & ~AlwaysFalsy]
+        return list1(res)
+    return inner
+
+@overload
+def f(x: int) -> list[int]: ...
+@overload
+def f(x: str) -> list[str]: ...
+def f(x: int | str) -> list[int] | list[str]:
+    # `list[int] | list[str]` is a different type than `list[int | str]`.
```
AlexWaygood:
```suggestion
    # `list[int] | list[str]` is disjoint from `list[int | str]`.
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4770,9 +4771,20 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
 
     fn infer_return_statement(&mut self, ret: &ast::StmtReturn) {
-        if let Some(ty) =
-            self.infer_optional_expression(ret.value.as_deref(), TypeContext::default())
-        {
+        let annotated_return_type = |_| {
+            nearest_enclosing_function(self.db(), self.index, self.scope()).map(|func| {
+                // When inferring expressions within a function body,
+                // the expected type passed should be the "raw" type,
+                // i.e. type variables in the return type are non-inferable, and the return types of async functions are not wrapped in `CoroutineType[...]`.
```
```suggestion
                // i.e. type variables in the return type are non-inferable,
                // and the return types of async functions are not wrapped in `CoroutineType[...]`.
```

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -408,23 +435,31 @@ impl<'db> OverloadLiteral<'db> {
...
             index,
-        );
-
+        ) {
+            flags |= SignatureFlags::HAS_IMPLICITLY_POSITIONAL_FIRST_PARAMETER;
+        }
+        if !raw {
+            flags |= SignatureFlags::MARK_TYPEVARS_INFERABLE;
+            flags |= SignatureFlags::COROUTINE_RETURN_TYPE;
+        }
```
dcreager: These flags are becoming fairly complex, which I consider a code smell. Could they instead become methods on `Signature`, that the caller can decide to call if they need to? There would still be a new `raw_signature`, but that's what `Signature::from_function` would return, and then `OverloadLiteral::signature` would call new `mark_typevars_inferable` and `wrap_coroutine_return_type` methods on the result before returning.

crates/ty_python_semantic/resources/mdtest/bidirectional.md
mtshiba: Yes, this typing is acceptable in this case.
The reason I marked this as TODO is because I think bidirectional inference could be used for inferring boolean operations. Here the expected type of `l2` is `list[int]`, so the type of `list()` should be inferred to be `list[int]`.

AlexWaygood: But I think we'd only use bidirectional inference to ensure that the inferred type is _assignable_ to the declared type. Here the "natural" inferred type (`(list[int] & ~AlwaysFalsy) | list[Unknown]`) [is already assignable](https://play.ty.dev/71778f42-1772-4e62-b233-2d9e9737cdd7) to the declared type (`list[int]`), so I don't think there's any reason for bidirectional inference to alter our inferred type. It's the same principle by which we infer `Literal[3]` here, even though the declared type is `int` -- `Literal[3]` is more precise than `int`, and the type `Literal[3]` is assignable to `int`, so it's acceptable here for our inferred type to be narrower than the declared type:

```py
x: int = 3
reveal_type(x)  # revealed: Literal[3]
```

mtshiba:
~~local mdtest is now failing as shown below, and I don't know why (CI reports no issues).~~
~~I think it might be related to #20533, do you know anything about it ? @dcreager~~

```
running 1 test

constraints.md - Constraints - Intersection - Intersection of a ra… (27656aaba27b10e9)

  crates\ty_python_semantic\resources\mdtest\type_properties\constraints.md:309 unmatched assertion: revealed: ty_extensions.ConstraintSet[(¬(Sub ≤ T@_ ≤ Base) ∧ (SubSub ≤ T@_ ≤ Base))]
  crates\ty_python_semantic\resources\mdtest\type_properties\constraints.md:309 unexpected error: 17 [revealed-type] "Revealed type: `ty_extensions.ConstraintSet[((SubSub ≤ T@_ ≤ Base) ∧ ¬(Sub ≤ T@_ ≤ Base))]`"

To rerun this specific test, set the environment variable: MDTEST_TEST_FILTER='constraints.md - Constraints - Intersection - Intersection of a ra… (27656aaba27b10e9)'
MDTEST_TEST_FILTER='constraints.md - Constraints - Intersection - Intersection of a ra… (27656aaba27b10e9)' cargo test -p ty_python_semantic --test mdtest -- mdtest__type_properties_constraints

constraints.md - Constraints - Union - Union of two ranges (3b2c9218d1b1205)

  crates\ty_python_semantic\resources\mdtest\type_properties\constraints.md:421 unmatched assertion: revealed: ty_extensions.ConstraintSet[(Base ≤ T@_ ≤ Super) ∨ (Sub ≤ T@_ ≤ Base)]
  crates\ty_python_semantic\resources\mdtest\type_properties\constraints.md:421 unexpected error: 17 [revealed-type] "Revealed type: `ty_extensions.ConstraintSet[(Sub ≤ T@_ ≤ Base) ∨ (Base ≤ T@_ ≤ Super)]`"

To rerun this specific test, set the environment variable: MDTEST_TEST_FILTER='constraints.md - Constraints - Union - Union of two ranges (3b2c9218d1b1205)'
MDTEST_TEST_FILTER='constraints.md - Constraints - Union - Union of two ranges (3b2c9218d1b1205)' cargo test -p ty_python_semantic --test mdtest -- mdtest__type_properties_constraints

--------------------------------------------------

test mdtest__type_properties_constraints ... FAILED
```

Seems to have been fixed in #20653.

crates/ty_python_semantic/resources/mdtest/bidirectional.md
mtshiba: Hmm, you're right. This typing indeed seems consistent with the current behavior of ty.

However, it seems counterintuitive that even though we typed `l2` as the fully-static type `list[int]` here, a gradual type is mixed into the inferred type.

This is a related issue to astral-sh/ty#136. For now, I'll mark this comment as just a note, not a TODO.

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -342,6 +342,50 @@ impl<'db> OverloadLiteral<'db> {
...
+            GenericContext::from_type_params(db, index, self.definition(db), type_params)
+        });
+        let legacy_ctx = GenericContext::from_function_params(
+            db,
+            self.definition(db),
+            signature.parameters(),
+            signature.return_ty,
+        );
+        signature.generic_context =
+            GenericContext::merge_pep695_and_legacy(db, pep695_ctx, legacy_ctx);
```
dcreager, line 370: I _think_ this might be redundant with work that is happening in `raw_signature` and `Signature::from_function`. If you remove this snippet does everything still work? (Or is there something being modified in the generic context that I'm missing? If so, this deserves a comment describing that.)

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1026,7 +1049,20 @@ impl<'db> SpecializationBuilder<'db> {
             return Ok(());
         }
 
+        if let Type::Union(union) = actual {
+            // For example, if `formal` is `list[T]` and `actual` is `list[int] | None`, we want to specialize `T` to `int`.
+            // So, here we remove the union elements that are not related to `formal`.
+            actual = union.filter(self.db, |actual_elem| {
+                !actual_elem.is_disjoint_from(self.db, formal)
+            });
+        }
```
ibraheemdev: Can you add a test [to this snippet](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md#collection-literal-annotations-are-understood) that `l: list[int | str] | None = [1]` is inferred as `list[int | str]` with this change?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4750,9 +4751,21 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                // i.e. type variables in the return type are non-inferable,
+                // and the return types of async functions are not wrapped in `CoroutineType[...]`.
+                TypeContext::new(func.last_definition_raw_signature(self.db()).return_ty)
+            })
+        };
+        let tcx = ret
+            .value
+            .as_ref()
+            .and_then(annotated_return_type)
+            .unwrap_or_default();
```
ibraheemdev: I find the way this is written slightly confusing.
```suggestion
        let tcx = if ret.value.is_some() {
            nearest_enclosing_function(self.db(), self.index, self.scope()).map(|func| {
                // When inferring expressions within a function body,
                // the expected type passed should be the "raw" type,
                // i.e. type variables in the return type are non-inferable,
                // and the return types of async functions are not wrapped in `CoroutineType[...]`.
                TypeContext::new(func.last_definition_raw_signature(self.db()).return_ty)
            })
        } else {
            TypeContext::default()
        };
```

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,140 @@
...
+intermediate = list1(1)
+# TODO: the error will not occur if we can infer the type of `intermediate` to be `list[int]`
+# error: [invalid-assignment] "Object of type `list[Literal[1]]` is not assignable to `list[int]`"
+l3: list[int] = intermediate
+# TODO: it would be nice if this were `list[int]`
+reveal_type(intermediate)  # revealed: list[Literal[1]]
+reveal_type(l3)  # revealed: list[int]
+
+l4: list[int] | None = list1(1)
+reveal_type(l4)  # revealed: list[int]
```
ibraheemdev: I'd prefer to avoid the overlap with literal promotion here.

```suggestion
l4: list[int | str] | None = list1(1)
reveal_type(l4)  # revealed: list[int | str]
```

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,140 @@
...
+    if isinstance(x, int):
+        return list1(x)
+    else:
+        return list1(x)
+
+reveal_type(f(1))  # revealed: list[int]
+reveal_type(f("a"))  # revealed: list[str]
+
+async def g() -> list[int]:
+    return list1(1)
```
ibraheemdev: Similarly here.
```suggestion
async def g() -> list[int | str]:
    return list1(1)
```

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -0,0 +1,140 @@
...
+reveal_type(l2)  # revealed: list[int]
+
+# `list[Literal[1]]` and `list[int]` are incompatible, since `list[T]` is invariant in `T`.
+# error: [invalid-assignment] "Object of type `list[Literal[1]]` is not assignable to `list[int]`"
+l2 = l1
+
+intermediate = list1(1)
+# TODO: the error will not occur if we can infer the type of `intermediate` to be `list[int]`
+# error: [invalid-assignment] "Object of type `list[Literal[1]]` is not assignable to `list[int]`"
+l3: list[int] = intermediate
```
ibraheemdev, line 34: Hmm.. it seems we need some form of post-specialization literal promotion here.

ibraheemdev (approved):

crates/ty_python_semantic/src/types/function.rs
mtshiba: We need to update `signature.generic_context` here, because type variables in `GenericContext::variables` are still non-inferable.

~~But I realized that this handling can be omitted at the `Signature::from_function` stage, so I did it (9507d7274b82567cd000bbfaefb4412251165ffb).~~
Hmm, this change seems to contrary make performance worse.

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -323,6 +323,10 @@ def union_param(x: T | None) -> T:
 reveal_type(union_param("a"))  # revealed: Literal["a"]
 reveal_type(union_param(1))  # revealed: Literal[1]
 reveal_type(union_param(None))  # revealed: Unknown
+
+def _(x: int | None):
+    # TODO: should be `int`
+    reveal_type(union_param(x))  # revealed: Unknown
```
ibraheemdev: Is this a regression?

mtshiba: This is revealed as `int | None` in main. The change is due to the addition of a [branch](https://github.com/mtshiba/ruff/blob/8834fec17728a1531821309ed2fa89c311fce242/crates/ty_python_semantic/src/types/generics.rs#L1167-L1168) that does nothing when matching in `SpecializationBuilder::infer`.
As [already mentioned](https://github.com/astral-sh/ruff/pull/20528/files#r2374660055), specialization between unions requires special handling, and falling back to the  `(Type::Union(_), _) => ...` branch could result in incorrect specialization. The default mapping  `T = Unknown` is suboptimal but not incorrect, and the mapping `T = int | None` is also suboptimal, so I'm not sure whether this should be considered a regression.

Specialization between unions will be a future work.

dcreager: This is timely — I just added that same branch with an initial "better than nothing" heuristic in https://github.com/astral-sh/ruff/pull/20749.

dcreager: (Note that the heuristic in 20749 will not handle @mtshiba's example, since it only engages when `formal` has precisely one top-level typevar, and `actual` has none.)

ibraheemdev: I think that's fine, this should be eventually fixed by @dcreager's constraint solver work, so I wouldn't worry too much about it.

mtshiba: OK, I merged the latest changes and made a minor fix to prevent incorrect specialization.

https://github.com/mtshiba/ruff/blob/1b42e0576a4493b1c5576624e8764b2708de410c/crates/ty_python_semantic/src/types/generics.rs#L1191-L1201

mtshiba:
I think I addressed all review comments.

codspeed reports a slight performance regression, but for the same reason as https://github.com/astral-sh/ruff/pull/20635, I believe it is difficult to avoid this.

The ecosystem reports are generally good, some false positive errors are due to TODOs in `TypedDict` and the unification solver.

ibraheemdev (approved):

carljm:
I think this might need minor updating now that I merged #20806 (since it should fix a test TODO added in that PR)? @ibraheemdev feel free to merge this if it looks ready to you.

ibraheemdev:
This looks great, thanks.

PR #20533, opened 2025-09-23
[ty] Change to BDD representation for constraint sets

dcreager:
While working on #20093, I kept running into test failures due to constraint sets not simplifying as much as they could, and therefore not being easily testable against "always true" and "always false".

This PR updates our constraint set representation to use BDDs. Because BDDs are reduced and ordered, they are canonical — equivalent boolean formulas are represented by the same interned BDD node.

That said, there is a wrinkle, in that the "variables" that we use in these BDDs — the individual constraints like `Lower ≤ T ≤ Upper` are not always independent of each other.

As an example, given types `A ≤ B ≤ C ≤ D` and a typevar `T`, the constraints `A ≤ T ≤ C` and `B ≤ T ≤ D` "overlap" — their intersection is non-empty. So we should be able to simplify

```
(A ≤ T ≤ C) ∧ (B ≤ T ≤ D) == (B ≤ T ≤ C)
```

That's not a simplification that the BDD structure can perform itself, since those three constraints are modeled as separate BDD variables, and are therefore "opaque" to the BDD algorithms.

That means we need to perform this kind of simplification ourselves. We look at pairs of constraints that appear in a BDD and see if they can be simplified relative to each other, and if so, replace the pair with the simplification. A large part of the toil of getting this PR to work was identifying all of those patterns and getting that substitution logic correct.

With this new representation, all existing tests pass, as well as some new ones that represent test failures that were occuring on #20093.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -562,3 +562,29 @@ def _[T]() -> None:
     # revealed: ty_extensions.ConstraintSet[always]
     reveal_type(constraint | ~constraint)
 ```
+
+### Negation of constraints involving two variables
+
+```py
+from typing import Never
+from ty_extensions import range_constraint
+
+class Base: ...
+
+def _[T, U]() -> None:
+    # revealed: ty_extensions.ConstraintSet[¬(T@_ ≤ Base) ∨ ¬(U@_ ≤ Base)]
+    reveal_type(~(range_constraint(Never, T, Base) & range_constraint(Never, U, Base)))
+```
+
+The union of a range constraint and its negation should always be satisfiable.
+
+```py
+def _[T, U]() -> None:
+    constraint = range_constraint(Never, T, Base) & range_constraint(Never, U, Base)
+    # TODO: revealed: ty_extensions.ConstraintSet[always]
+    # revealed: ty_extensions.ConstraintSet[((T@_ ≤ Base) ∧ (U@_ ≤ Base)) ∨ ¬(T@_ ≤ Base) ∨ ¬(U@_ ≤ Base)]
+    reveal_type(constraint | ~constraint)
+    # TODO: revealed: ty_extensions.ConstraintSet[always]
+    # revealed: ty_extensions.ConstraintSet[¬(T@_ ≤ Base) ∨ ¬(U@_ ≤ Base) ∨ ((T@_ ≤ Base) ∧ (U@_ ≤ Base))]
+    reveal_type(~constraint | constraint)
+```
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -305,7 +305,7 @@ range.
 
 ```py
 def _[T]() -> None:
-    # revealed: ty_extensions.ConstraintSet[((SubSub ≤ T@_ ≤ Base) ∧ ¬(Sub ≤ T@_ ≤ Base))]
+    # revealed: ty_extensions.ConstraintSet[(¬(Sub ≤ T@_ ≤ Base) ∧ (SubSub ≤ T@_ ≤ Base))]
     reveal_type(range_constraint(SubSub, T, Base) & negated_range_constraint(Sub, T, Super))
     # revealed: ty_extensions.ConstraintSet[(¬(Sub ≤ T@_ ≤ Base) ∧ (SubSub ≤ T@_ ≤ Super))]
     reveal_type(range_constraint(SubSub, T, Super) & negated_range_constraint(Sub, T, Base))
@@ -339,7 +339,7 @@ Otherwise, the union cannot be simplified.
 
 ```py
 def _[T]() -> None:
-    # revealed: ty_extensions.ConstraintSet[(¬(Sub ≤ T@_ ≤ Base) ∧ ¬(Base ≤ T@_ ≤ Super))]
+    # revealed: ty_extensions.ConstraintSet[(¬(Base ≤ T@_ ≤ Super) ∧ ¬(Sub ≤ T@_ ≤ Base))]
     reveal_type(negated_range_constraint(Sub, T, Base) & negated_range_constraint(Base, T, Super))
     # revealed: ty_extensions.ConstraintSet[(¬(Base ≤ T@_ ≤ Super) ∧ ¬(SubSub ≤ T@_ ≤ Sub))]
     reveal_type(negated_range_constraint(SubSub, T, Sub) & negated_range_constraint(Base, T, Super))
@@ -438,9 +438,9 @@ Otherwise, the union cannot be simplified.
 def _[T]() -> None:
     # revealed: ty_extensions.ConstraintSet[(Base ≤ T@_ ≤ Super) ∨ (Sub ≤ T@_ ≤ Base)]
     reveal_type(range_constraint(Sub, T, Base) | range_constraint(Base, T, Super))
-    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Sub) ∨ (Base ≤ T@_ ≤ Super)]
+    # revealed: ty_extensions.ConstraintSet[(Base ≤ T@_ ≤ Super) ∨ (SubSub ≤ T@_ ≤ Sub)]
     reveal_type(range_constraint(SubSub, T, Sub) | range_constraint(Base, T, Super))
-    # revealed: ty_extensions.ConstraintSet[(Unrelated ≤ T@_) ∨ (SubSub ≤ T@_ ≤ Sub)]
+    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Sub) ∨ (Unrelated ≤ T@_)]
     reveal_type(range_constraint(SubSub, T, Sub) | range_constraint(Unrelated, T, object))
 ```
 
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -389,6 +389,25 @@ def _[T, U]() -> None:
     reveal_type(negated_range_constraint(Sub, T, Base) | negated_range_constraint(Sub, U, Base))
 ```
 
+### fwomp
+
+```py
+from typing import final
+from ty_extensions import range_constraint
+
+class Super: ...
+class Base(Super): ...
+class Sub(Base): ...
+class SubSub(Sub): ...
+
+@final
+class Unrelated: ...
+
+def _[T]() -> None:
+    # revealed: ty_extensions.ConstraintSet[(SubSub ≤ T@_ ≤ Super)]
+    reveal_type(range_constraint(SubSub, T, Super) | range_constraint(Sub, T, Base))
+```
+
 ### Union of two ranges
 
 When one of the bounds is entirely contained within the other, the union simplifies to the larger
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -307,7 +307,7 @@ range.
 def _[T]() -> None:
     # revealed: ty_extensions.ConstraintSet[((SubSub ≤ T@_ ≤ Base) ∧ ¬(Sub ≤ T@_ ≤ Base))]
     reveal_type(range_constraint(SubSub, T, Base) & negated_range_constraint(Sub, T, Super))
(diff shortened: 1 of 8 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -562,3 +562,42 @@ def _[T]() -> None:
...
+    # revealed: ty_extensions.ConstraintSet[always]
+    reveal_type(c2 | ~c2)
+    # revealed: ty_extensions.ConstraintSet[always]
+    reveal_type(~c2 | c2)
+
+    union = c1 | c2
+    # revealed: ty_extensions.ConstraintSet[always]
+    reveal_type(union | ~union)
+    # revealed: ty_extensions.ConstraintSet[always]
+    reveal_type(~union | union)
```
dcreager, line 602: These tests all failed to simplify with the old representation

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1142 @@ impl<'db> ConstraintSet<'db> {
...
-            Satisfiable::Constrained(clause) => clause,
-        };
+/// An individual constraint in a constraint set. This restricts a single typevar to be within a
+/// lower and upper bound.
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
+pub(crate) struct ConstrainedTypeVar<'db> {
+    typevar: BoundTypeVarInstance<'db>,
+    range: RangeConstraint<'db>,
+}
```
carljm, line 264: I'm not sure I understand why `RangeConstraint` and `ConstrainedTypeVar` should be separate structs, given that the only kind of constraint a `ConstrainedTypeVar` can contain is a `RangeConstraint`. It seems to result in a lot of awkward entanglement between the types:

* We have `RangeConstraint::new_node`, which returns a `Node` containing a `ConstrainedTypeVar` containing a `RangeConstraint` -- why is this method located on `RangeConstraint`?
*  `RangeConstraint::display` has to accept an external typevar.
* Various methods (e.g. `RangeConstraint::intersect`) are not sensible unless the constraints are applied to the same typevar, but `RangeConstraint` can't validate this itself.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1142 @@ impl<'db> ConstraintSet<'db> {
...
-        //
-        // TODO: Consider checking both directions in one pass, possibly via a tri-valued return
-        // value.
-        if self.subsumes_via_intersection(db, &other) {
-            return Simplifiable::Simplified(other);
-        }
-        if other.subsumes_via_intersection(db, &self) {
-            return Simplifiable::Simplified(self);
+    /// Returns the `if-then-else` of three BDDs: when `self` evaluates to `true`, it returns when
+    /// `then_node` evaluates to; otherwise it returns when `else_node` evaluates to.
```
```suggestion
    /// Returns the `if-then-else` of three BDDs: when `self` evaluates to `true`, it returns what
    /// `then_node` evaluates to; otherwise it returns what `else_node` evaluates to.
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
-        // See the notes in `simplify_clauses` for more details on subsumption, including Python
-        // examples that cause it to fire.
+    /// Returns a new BDD that returns the same results as `self`, but with some inputs fixed to
+    /// particular values. (Those variables will not be checked when evaluating the result, and
+    /// will not be present in the result.)
+    fn restrict(
+        self,
+        db: &'db dyn Db,
+        assignment: impl IntoIterator<Item = ConstraintAssignment<'db>>,
+    ) -> Node<'db> {
```
```suggestion
    ) -> Self {
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+    }
 
-        if self.constraints.len() != other.constraints.len() {
-            return false;
+    /// Returns a new BDD that returns the same results as `self`, but with one input fixed to a
+    /// particular value. (That variable will be not be checked when evaluating the result, and
+    /// will not be present in the result.)
+    ///
+    /// Also returns whether the restricted variable appeared in the BDD.
+    fn restrict_one(self, db: &'db dyn Db, assignment: ConstraintAssignment<'db>) -> Node<'db> {
```
```suggestion
    fn restrict_one(self, db: &'db dyn Db, assignment: ConstraintAssignment<'db>) -> Self {
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+        match self {
+            ConstraintAssignment::Positive(constraint) => constraint,
+            ConstraintAssignment::Negative(constraint) => constraint,
         }
-        for self_negative in &self.negative {
-            for other_negative in &other.negative {
-                results.add_negated_range(db, self_negative.union_negative(db, other_negative));
+    }
+
+    fn flipped(self) -> Self {
```
carljm: Why is this named `flipped()` instead of `negated()`?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+
+#[salsa::tracked]
+impl<'db> InteriorNode<'db> {
+    #[salsa::tracked(heap_size=ruff_memory_usage::heap_size)]
+    fn negate(self, db: &'db dyn Db) -> Node<'db> {
+        Node::new(
+            db,
+            self.constraint(db),
+            self.if_true(db).negate(db),
+            self.if_false(db).negate(db),
```
carljm, line 795: Couldn't this just swap `if_true` and `if_false` instead of recursively negating them? Or would that violate some invariant of the BDD?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+        // Lastly, verify that the result is consistent with the input. (It must produce the same
+        // results when `left ∧ right`.) If it doesn't, the substitution isn't valid, and we should
+        // return the original BDD unmodified.
+        let validity = replacement.iff(db, left_node.and(db, right_node));
+        let constrained_original = self.and(db, validity);
+        let constrained_replacement = result.and(db, validity);
+        if constrained_original == constrained_replacement {
+            result
+        } else {
+            self
```
carljm, line 631: When would we attempt a simplification that this validation would reject? Are some of the simplifications attempted in `InteriorNode::simplify` speculative, or possibly invalid? If so, maybe we should mark those more clearly than we do now?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+        let left_node = Node::new_satisfied_constraint(db, left);
+        let right_node = Node::new_satisfied_constraint(db, right);
+        let validity = replacement.iff(db, left_node.or(db, right_node));
+        let constrained_original = self.and(db, validity);
+        let constrained_replacement = result.and(db, validity);
+        if constrained_original == constrained_replacement {
+            result
+        } else {
+            self
         }
```
carljm, line 689: Same question as above.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
+                    Node::new_satisfied_constraint(db, larger_constraint.when_true()),
+                );
+
+                // larger ∧ smaller = smaller
+                simplified = simplified.substitute_intersection(
+                    db,
+                    larger_constraint.when_false(),
+                    smaller_constraint.when_false(),
+                    Node::new_satisfied_constraint(db, larger_constraint.when_false()),
+                );
```
carljm: I might be misunderstanding, but it seems like the attempted simplification here doesn't match the comment? The simplification I'd expect to match the comment would be this one:

```suggestion
                // larger ∧ smaller = smaller
                simplified = simplified.substitute_intersection(
                    db,
                    larger_constraint.when_true(),
                    smaller_constraint.when_true(),
                    Node::new_satisfied_constraint(db, smaller_constraint.when_true()),
                );
```

It seems like the currently implemented simplification is `¬smaller ∧ ¬larger = ¬larger`, which is also correct, but not the same thing.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -273,1231 +239,1147 @@ impl<'db> ConstraintSet<'db> {
...
-                        self.constraint.hole.lower.display(self.db)
-                    );
+    /// Simplifies the DNF representation, removing redundancies that do not change the underlying
+    /// function. (This is used when displaying a BDD, to make sure that the representation that we
+    /// show is as simple as possible while still producing the same results.)
+    fn simplify(&mut self) {
+        while self.simplify_one_round() {
+            // Keep going
+        }
+    }
```
carljm, line 1266: Is it not possible to sufficiently simplify the underlying BDD so we don't need additional simplification of the `SatisfiedConstraints`?

carljm (approved):
Nice! It looks like a few tests are failing with the latest commit.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: No, this is unrelated to the simplifications that we make to the BDD. Those are needed since the BDD variables are not independent of each other.

This issue is true of BDDs in general, even those with nicely independent variables. The problem is that the naive DNF of a BDD includes extra information, because it's just walking the BDD node graph. The simplest example is if you create the BDD for `x ∨ y`:

```
    x
  ₀/ \₁
  y   1
₀/ \₁
0   1
```

and then render that by turning it into DNF, you get `x ∨ (¬x ∧ y)`. The DNF formula is just a straight rendering of each path that leads to a 1 terminal. But when evaluating that formula, if `x=1`, we short circuit and don't need to evaluate the second clause. Since we can _assume_ that `x=0` if we get to the point where we need to look at the second clause, we can remove `¬x` from it. That's what this simplification is doing.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: In a way all of the simplifications are speculative — not because they sometimes don't hold, but because we don't know in advance whether a particular BDD contains the clause that we're trying to simplify. For instance, we might try to substitute `x || y` with `y` (because we determined that `y` fully contains `x`). If we do that to the function `x`, without this check, the result we get from the substitution is `y`. But the result should be `x`, since `x || y` doesn't appear in the function. In general there isn't a way to see if a BDD contains a clause like `x || y`, since those clauses aren't stored directly.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: No, if you have `a ∧ b` that would produce `a ∧ ¬b`:

```
a ∧ b =  a
       ₀/ \₁
       0   b
         ₀/ \₁
         0   1
```

Flipping the true/false edges of the root node gives:

```
         a
       ₀/ \₁
       b   0
     ₀/ \₁
     0   1
```

which is `¬a ∧ b`. Instead we want to walk the entire graph and change each terminal 0 or 1 to its opposite.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Oh yeah good catch! I'll rename to be consistent with the other types

crates/ty_python_semantic/src/types/constraints.rs
dcreager: This is a holdover from the previous representation, where there were different kinds of constraint that we built up into a `ConstrainedTypeVar`. I was mildly leaning towards keeping the diff smaller where possible, but I take your point. I'll combine them.

crates/ty_python_semantic/src/types/constraints.rs
carljm: I see, makes sense. It still somehow feels like we ought to be able to do this "on the fly" while building the DNF representation, by tracking the information we can assume and not adding clauses that can be assumed? But maybe not, or maybe doing it after-the-fact is just as good in practice.

dcreager: We might be able to, but this code path is only executed as part of displaying constraint sets, which only happens in our test cases, so I didn't think it was worth spending too much time optimizing.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: It's the comment that's wrong (which I just fixed). The `larger ∧ smaller = smaller` case holds for any two non-disjoint types, not just for strict containment, so it's handled in the stanza below.

PR #20534, opened 2025-09-23
[`flake8-bandit`] Clarify the supported hashing functions (`S324`)

### LINKED ISSUE #16572: S324 false negatives for synonyms of insecure algorithm names
### Summary

[hashlib-insecure-hash-function (S324)](https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/) misses some insecure algorithm names that are equivalent to names that it currently flags. `"ssl3-md5"` and `"1.2.840.113549.2.5"` are synonyms of `"md5"`. `"sha-1"`, `"ssl3-sha1"`, and `"1.3.14.3.2.26"` are synonyms of `"sha1"`. Trailing spaces are ignored after any name. The details depend on the version of OpenSSL that Python is using, but those are the false negatives I’ve found on my machine.

```python
import hashlib
print(hashlib.new("md5  ").hexdigest())
print(hashlib.new("sha-1").hexdigest())
print(hashlib.new("ssl3-md5").hexdigest())
print(hashlib.new("ssl3-sha1").hexdigest())
print(hashlib.new("1.3.14.3.2.26").hexdigest())
print(hashlib.new("1.2.840.113549.2.5").hexdigest())
```

### Version

ruff 0.9.10 (0dfa810e9 2025-03-07)

danparizher:
## Summary

Fixes #16572

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_bandit/rules/hashlib_insecure_hash_functions.rs
+++ b/crates/ruff_linter/src/rules/flake8_bandit/rules/hashlib_insecure_hash_functions.rs
@@ -69,6 +69,7 @@ use crate::rules::flake8_bandit::helpers::string_literal;
 /// ## References
 /// - [Python documentation: `hashlib` — Secure hashes and message digests](https://docs.python.org/3/library/hashlib.html)
 /// - [Python documentation: `crypt` — Function to check Unix passwords](https://docs.python.org/3/library/crypt.html)
+/// - [Python documentation: `FIPS` - FIPS compliant hashlib implementation](https://docs.python.org/3/library/hashlib.html#hashlib.algorithms_guaranteed)
 /// - [Common Weakness Enumeration: CWE-327](https://cwe.mitre.org/data/definitions/327.html)
 /// - [Common Weakness Enumeration: CWE-328](https://cwe.mitre.org/data/definitions/328.html)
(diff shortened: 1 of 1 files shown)


danparizher:
Thanks for checking, that makes sense. I changed it to explicitly list what the rule actually targets, plus `crypt.crypt` and `crypt.mksalt` when used with the weak methods (`METHOD_CRYPT`, `METHOD_MD5`, `METHOD_BLOWFISH`)

crates/ruff_linter/src/rules/flake8_bandit/rules/hashlib_insecure_hash_functions.rs
```diff
@@ -58,7 +69,6 @@ use crate::rules::flake8_bandit::helpers::string_literal;
 /// ## References
 /// - [Python documentation: `hashlib` — Secure hashes and message digests](https://docs.python.org/3/library/hashlib.html)
 /// - [Python documentation: `crypt` — Function to check Unix passwords](https://docs.python.org/3/library/crypt.html)
-/// - [Python documentation: `FIPS` - FIPS compliant hashlib implementation](https://docs.python.org/3/library/hashlib.html#hashlib.algorithms_guaranteed)
```
ntBre, line 61:
```suggestion
/// - [Python documentation: `FIPS` - FIPS compliant hashlib implementation](https://docs.python.org/3/library/hashlib.html#hashlib.algorithms_guaranteed)
```

PR #20535, opened 2025-09-23
[`pydoclint`] Fix false positive when Sphinx directives follow Raises section (`DOC502`)

### LINKED ISSUE #18959: DOC502: False positive if following text contains a colon
### Summary

Possibly related to https://github.com/astral-sh/ruff/issues/16578 but somewhat different: Text after the `Raises` block triggers a false positive for DOC502 if that text contains a colon.

In my case, the relevant part of the docstring contains Sphinx directives, properly unindented:

```
Raises:
    ExtraneousDataError:
        ``exact`` is True and there's data left over at the end of the byte
        string.

.. versionadded:: 0.7.0
    The ``init_kwargs`` argument.
```

This results in a false positive:

```
Raised exception is not explicitly raised: `. versionadded` (DOC502) [Ln 7, Col 5]
```

Note the line number is also wrong; it's line 27 of the docstring, line 6 of the `Raises` section if you include the line with "Raises" on it.

It seems like Ruff thinks `versionadded:` is part of the `Raises` block, even though it's unindented. Interestingly, removing both colons from `versionadded` makes the issue go away. Adding just one back triggers the bug again.

Playground link: https://play.ruff.rs/2f264f7b-d661-4af3-83e0-2db04e30ced8

### Version

0.12.0

danparizher:
## Summary

Fixes #18959

### DIFF
--- a/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
+++ b/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
@@ -756,13 +756,13 @@ fn parse_raises_numpy(content: &str) -> Vec<QualifiedName<'_>> {
     };
     let indentation = &dashes[..dashes.len() - dashes.trim_start().len()];
     for potential in lines {
-        // Check for Sphinx directives (lines starting with ..) - these indicate the end of the section
-        // In numpy-style, exceptions are dedented to the same level as sphinx directives
-        let trimmed = potential.trim_start();
-        if trimmed.starts_with("..") {
-            break;
-        }
         if let Some(entry) = potential.strip_prefix(indentation) {
+            // Check for Sphinx directives (lines starting with ..) - these indicate the end of the
+            // section. In numpy-style, exceptions are dedented to the same level as sphinx
+            // directives.
+            if entry.starts_with("..") {
+                break;
+            }
             if let Some(first_char) = entry.chars().next() {
                 if !first_char.is_whitespace() {
                     entries.push(QualifiedName::user_defined(entry.trim_end()));
--- a/crates/ruff_linter/resources/test/fixtures/pydoclint/DOC502_numpy.py
+++ b/crates/ruff_linter/resources/test/fixtures/pydoclint/DOC502_numpy.py
@@ -132,3 +132,18 @@ def foo():
         The ``init_kwargs`` argument.
     """
     raise ValueError
+
+# Make sure we don't bail out on a Sphinx directive in the description of one
+# of the exceptions
+def foo():
+    """First line.
+
+    Raises
+    ------
+    ValueError
+        some text
+        .. math:: e^{xception}
+    ZeroDivisionError
+        Will not be raised, DOC502
+    """
+    raise ValueError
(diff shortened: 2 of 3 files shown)


danparizher:
Note on `ruff-ecosystem` results: a net decrease in violations overall.
- Instances where a “removal/addition” pair appears at the same location.
    - Those paired changes indicate message normalization rather than behavioral regression—the false positives are being eliminated where Sphinx directives follow a Raises section
- Remaining diagnostics have cleaner messages without stray directive/markup fragments.

Net effect is reduced noise and improved precision

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -476,13 +476,45 @@ fn parse_entries(content: &str, style: Option<SectionStyle>) -> Vec<QualifiedNam
 /// ```
 fn parse_entries_google(content: &str) -> Vec<QualifiedName<'_>> {
     let mut entries: Vec<QualifiedName> = Vec::new();
-    for potential in content.lines() {
-        let Some(colon_idx) = potential.find(':') else {
-            continue;
-        };
-        let entry = potential[..colon_idx].trim();
-        entries.push(QualifiedName::user_defined(entry));
+
```
ntBre: What do you think about something more like this for this function?

```rust
fn parse_entries_google(content: &str) -> Vec<QualifiedName<'_>> {
    let mut entries: Vec<QualifiedName> = Vec::new();
    let mut lines = content.lines().peekable();
    let Some(first) = lines.peek() else {
        return entries;
    };
    let indentation = &first[..first.len() - first.trim_start().len()];
    for potential in lines {
        if let Some(entry) = potential.strip_prefix(indentation) {
            if let Some(first_char) = entry.chars().next() {
                if !first_char.is_whitespace() {
                    if let Some(colon_idx) = entry.find(':') {
                        let entry = entry[..colon_idx].trim();
                        if !entry.is_empty() {
                            entries.push(QualifiedName::user_defined(entry));
                        }
                    }
                }
            }
        }
    }
    entries
}
```

This is closer to the `parse_entries_numpy` down below and seems to work on the new test case when I tried it locally.

danparizher: This is significantly better, thank you! We will have to see what the ecosystem results will look like.

ntBre:
Could you take a look at the ecosystem results? It looks like there are still some issues. I think we may need a combination of the two approaches we've tried, especially breaking early if we find a dedented line like you had initially. We should extract some test cases from the erroneous ecosystem results too.

danparizher:
Just to clarify, for the "dedented line" breaking logic, should it:
1. Break immediately when ANY line has less indentation than the first line
2. Break only when a line has NO indentation at all
3. Break when a line is dedented to the same level as the section header

ntBre:
I think I was picturing option 1, possibly excluding completely blank lines, assuming those are allowed. I'm looking at 
> + [superset/utils/date_parser.py:173:5:](https://github.com/apache/superset/blob/78907d08cd5a3a53e25050a08328bb3b7770f3c2/superset/utils/date_parser.py#L173) DOC502 Raised exception is not explicitly raised: `- Example`

for example. The `- Example` point under a different heading is being flagged, so I assumed breaking when the indentation decreases would help:

```py
    Returns:
        str: The resulting expression for the start of the specified unit.

    Raises:
        ValueError: If the unit is not one of the valid options.

    Relation to `time_range_lookup`:
        - Handles the "start of" or "beginning of" modifiers in the first regex pattern.
        - Example: "start of this month" → `DATETRUNC(DATETIME('today'), month)`.
    """
```

But that's just an idea.

ntBre (approved):
Thank you! This looks great and shows a very clear improvement in the ecosystem check. 

I'm tempted to go ahead and land this, but I think this could also be a problem for numpy-style docstrings ([example](https://play.ruff.rs/bf640866-ee7a-4f67-96c7-3a36573b735f)). These seem a little trickier since the exceptions are already dedented to the same level as the sphinx directive, assuming I followed the docs correctly. We may just have to check for a leading `..` in that case.

Do you want to try fixing that here too, or would it be better as a separate PR?

danparizher:
I'll try to fix that here too, thanks!

PR #20541, opened 2025-09-24
[ty] handle more `enum.auto()` cases

thejchap:
## Summary

see https://github.com/astral-sh/ty/issues/1241

## Test Plan

new mdtests


crates/ty_python_semantic/src/types/enums.rs
```diff
@@ -150,14 +147,29 @@ pub(crate) fn enum_metadata<'db>(
...
+                                {
+                                    KnownClass::Str.to_instance(db)
+                                } else if Type::ClassLiteral(class)
+                                    .is_subtype_of(db, KnownClass::Bytes.to_subclass_of(db))
+                                {
+                                    KnownClass::Bytes.to_instance(db)
+                                } else if Type::ClassLiteral(class)
+                                    .is_subtype_of(db, KnownClass::Float.to_subclass_of(db))
+                                {
+                                    KnownClass::Float.to_instance(db)
```
AlexWaygood: hmm, I'm not sure we should be adding more special cases here. An enum class that uses `bytes` as a mixin isn't conceptually any different from an enum that uses `tuple` or `list` or `dict` as a mixin, or a custom dataclass or custom `NamedTuple` as a mixin. Why should we treat `bytes` and `float` specially here?

Rather than adding more special cases, I think we should have a generalised fallback logic that iterates through the MRO of the enum class to check whether it has a custom mixin that isn't `str` or `int`. (Those two make sense to special-case, I think, because of the presence of `StrEnum` and `IntEnum` in the stdlib. You can determine whether an element in the MRO is a "custom mixin" class by checking if it isn't `object` and isn't a subclass of `enum.Enum` -- if both those conditions hold, I think we can treat it as a mixin class.) If the enum class has a custom mixin and `auto()` is used to define the value of an enum member, we should probably infer the `.value` of that enum member as just being `Any`, which would still be more accurate than what we have on `main` (even if it isn't very precise!). It seems pretty hard to predict whether the `.value` property is going to give you an instance of the mixin or an integer:

```pycon
>>> class Foo: ...
>>> import enum
>>> class FooEnum(Foo, enum.Enum):
...     X = enum.auto()
...     
>>> FooEnum.X.value
1
>>> class BEnum(bytes, enum.Enum):
...     X = enum.auto()
...     
>>> BEnum.X.value
b'\x00'
>>> class TEnum(tuple, enum.Enum):
...     X = enum.auto()
...     
>>> TEnum.X.value
(1,)
```

thejchap: this makes a lot more sense - thanks - didn't feel great about the explicit special cases after going down that road...

thejchap:
@AlexWaygood i think this is ready for another look!

AlexWaygood (approved):
Thank you!!

AlexWaygood:
@woodruffw, any idea why astral-bot isn't posting or editing any ecosystem-report comments on this PR? :-)

I'd love to double-check the ecosystem impact here before merging

woodruffw:
(Just synchronizing chat streams)

Yeah, I have a theory on what's causing this -- I've opened a tracking issue on the bot!

woodruffw:
I've made a tentative fix for this, which is now in place -- I'm going to fast-forward this branch to confirm that it works.

Edit: appears to be working!

PR #20542, opened 2025-09-24
[ty] Use full project names in `good.txt`

sharkdp:
## Summary

This allows `ecosystem-analyzer` to pick up these projects as well (uses full names instead of regex patterns).

## Test Plan

- [ ] Successful run of mypy_primer
- [ ] Made sure that mypy_primer still runs all three CPython projects
- [ ] Successful run of ecosystem-analyzer
- [ ] Made sure that ecossystem-analyzer now picks up the CPython projects

### DIFF
--- a/.github/workflows/ty-ecosystem-report.yaml
+++ b/.github/workflows/ty-ecosystem-report.yaml
@@ -49,7 +49,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@27dd66d9e397d986ef9c631119ee09556eab8af9"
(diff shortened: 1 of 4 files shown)


.github/workflows/ty-ecosystem-report.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@27dd66d9e397d986ef9c631119ee09556eab8af9"
+          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@fc0f612798710b0dd69bb7528bc9b361dc60bd43"
```
sharkdp, line 52: I only bumped it in `ty-ecosystem-analyzer.yaml`, forgot to bump it here as well.

PR #20547, opened 2025-09-24
[ty] Filter overloads using variadic parameters

dhruvmanila:
## Summary

Closes: https://github.com/astral-sh/ty/issues/551

This PR adds support for step 4 of the overload call evaluation algorithm which states that:

> If the argument list is compatible with two or more overloads, determine whether one or more of the overloads has a variadic parameter (either `*args` or `**kwargs`) that maps to a corresponding argument that supplies an indeterminate number of positional or keyword arguments. If so, eliminate overloads that do not have a variadic parameter.

And, with that, the overload call evaluation algorithm has been implemented completely end to end as stated in the typing spec.

## Test Plan

Expand the overload call test suite.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/overloads.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/overloads.md
@@ -942,7 +942,7 @@ This is the step 4 of the overload call evaluation algorithm which specifies tha
 
 This is only performed if the previous step resulted in more than one matching overload.
 
-### Simple
+### Simple `*args`
 
 `overloaded.pyi`:
 
@@ -954,7 +954,7 @@ def f(x1: int) -> tuple[int]: ...
 @overload
 def f(x1: int, x2: int) -> tuple[int, int]: ...
 @overload
-def f(*x: int) -> int: ...
+def f(*args: int) -> int: ...
 ```
 
 ```py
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/call/overloads.md
```diff
@@ -931,6 +931,134 @@ def _(t: tuple[int, str] | tuple[int, str, int]) -> None:
     f(*t)  # error: [no-matching-overload]
 ```
 
+## Filtering based on variaidic arguments
+
+This is the step 4 of the overload call evaluation algorithm which specifies that:
```
AlexWaygood:
```suggestion
This is step 4 of the overload call evaluation algorithm which specifies that:
```

crates/ty_python_semantic/resources/mdtest/call/overloads.md
```diff
@@ -931,6 +931,134 @@ def _(t: tuple[int, str] | tuple[int, str, int]) -> None:
...
+def f(*args: int) -> int: ...
+```
+
+```py
+from overloaded import f
+
+def _(x1: int, x2: int, args: list[int]):
+    reveal_type(f(x1))  # revealed: tuple[int]
+    reveal_type(f(x1, x2))  # revealed: tuple[int, int]
+    reveal_type(f(*(x1, x2)))  # revealed: tuple[int, int]
```
AlexWaygood, line 966: could also test other iterables of `int`s that we know are fixed-length when splatting them, e.g.

```suggestion
    reveal_type(f(*(x1, x2)))  # revealed: tuple[int, int]
    reveal_type(f(*b"ab"))  # revealed: tuple[int, int]
```

crates/ty_python_semantic/resources/mdtest/call/overloads.md
```diff
@@ -931,6 +931,134 @@ def _(t: tuple[int, str] | tuple[int, str, int]) -> None:
...
+from overloaded import f
+
+class Foo(TypedDict):
+    x: int
+    y: int
+
+def _(foo: Foo, kwargs: dict[str, int]):
+    reveal_type(f(**foo))  # revealed: tuple[int, int]
+    reveal_type(f(**kwargs))  # revealed: tuple[int, ...]
+```
```
AlexWaygood, line 1060: Interesting -- this is what other type checkers do too, but I don't think it's actually sound unless the `TypedDict` is marked as `closed=True`. That's because you could always have a `Bar` `TypedDict` that's a subclass of `Foo` and adds additional keys:

```py
from typing import TypedDict, overload

@overload
def f(*, x: int) -> int: ...
@overload
def f(*, x: int, y: int) -> tuple[int, int]: ...
@overload
def f(**kwargs: int) -> tuple[int, ...]: ...

class Foo(TypedDict):
    x: int
    y: int

class Bar(Foo):
    z: str

def g(foo: Foo):
    reveal_type(f(**foo))

g(Bar({"x": 42, "y": 42, "z": "z"}))
```

In the example above, a `z="z"` argument is passed to the `g()` function, which isn't permitted by any overload of `f()`, but no type checker catches this.

It probably makes sense to follow what existing type checkers do here (especially since `closed=True` is a very new feature in the type system). Maybe this is just an example I should add to https://github.com/JelleZijlstra/unsoundness

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2029,6 +2090,7 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
```
```suggestion
    #[expect(clippy::too_many_arguments)]
```

AlexWaygood (approved):
Nice!

crates/ty_python_semantic/resources/mdtest/call/overloads.md
AlexWaygood: Oh, and I See @JelleZijlstra already has exactly this example :-) https://github.com/JelleZijlstra/unsoundness/blob/main/examples/typeddicts/typeddict_argument_unpacking.py

JelleZijlstra: Agree it makes sense to follow other type checkers here for now, but in the long term it would be nice to close this hole.

dhruvmanila: Oh, that's interesting! Yeah, I think what Jelle said makes sense.

crates/ty_python_semantic/resources/mdtest/call/overloads.md
dhruvmanila: Yeah, I think this makes sense although I'll put it in a relevant place as this is mainly to make sure that `*args` works and that other cases are just to verify that the overload structure is correct for non-variadic parameters.

dhruvmanila: Ok, I've added this to `call/function.md` > `Splatted arguments` > `Unknown argument length` :)

PR #20548, opened 2025-09-24
[`flake8-simplify`] Improve help message clarity (`SIM105`)

### LINKED ISSUE #20462: Suggestion: Improve SIM105 message, check for improper use of contextlib.suppress
### Summary

I recently received a [PR](https://gitlab.com/SiLA2/sila_python/-/merge_requests/76) containing a subtle bug, which was kind-of caused by ruff.

### What happened
[First commit](https://gitlab.com/StefanMa/sila_python-deletion_scheduled-74452258/-/commit/a63e34e853b701b10bf7e64807d8597edbde1089#ede45c3ac59be78ed757ad91acbe7945e3b47abe_109_113):

```python
try:
    something()
except SomeException:
    pass
```

[Ruff in CI](https://gitlab.com/StefanMa/sila_python-deletion_scheduled-74452258/-/jobs/11359074513):

```text
113 | /             try:
114 | |                 something()
115 | |             except SomeException:
116 | |                 pass
    | |____________________^
    |
help: Replace with `contextlib.suppress(SomeException)`
```

The PR author literally followed the Ruff message in the [next commit](https://gitlab.com/SiLA2/sila_python/-/commit/df38c61a6f625dec0563e73d8ba846e5dc057dd5?merge_request_iid=76), which satisfied ruff:

```python
try:
    something()
except SomeException:
    contextlib.suppress(SomeException)  # instead of `pass`
```

The correct usage is:

```python
with contextlib.suppress(SomeException):
    something()
```

The [documentation of SIM105](https://docs.astral.sh/ruff/rules/suppressible-exception/) is fine, but the user didn't look there due to the seemingly clear help message.

### Suggestions
* Change the help message of SIM105 to "Replace try-except block with `with contextlib.suppress(SomeException):`"
* Add a rule for flagging `contextlib.suppress(...)` as the only statement in an `except` block (or merge with [B022: useless-contextlib-suppress](https://docs.astral.sh/ruff/rules/useless-contextlib-suppress/))

mgiovani:
## Summary

Improve the SIM105 rule message to prevent user confusion about how to properly use `contextlib.suppress`.

The previous message "Replace with `contextlib.suppress(ValueError)`" was ambiguous and led users to incorrectly use `contextlib.suppress(ValueError)` as a statement inside except blocks instead of replacing the entire try-except-pass block with `with contextlib.suppress(ValueError):`.

This change makes the message more explicit:
- **Before**: `"Use \`contextlib.suppress({exception})\` instead of \`try\`-\`except\`-\`pass\`"`
- **After**: `"Replace \`try\`-\`except\`-\`pass\` block with \`with contextlib.suppress({exception})\`"`

The fix title is also updated to be more specific:
- **Before**: `"Replace with \`contextlib.suppress({exception})\`"`  
- **After**: `"Replace \`try\`-\`except\`-\`pass\` with \`with contextlib.suppress({exception})\`"`

Fixes #20462

## Test Plan

- ✅ All existing SIM105 tests pass with updated snapshots
- ✅ Cargo clippy passes without warnings  
- ✅ Full test suite passes
- ✅ The new messages clearly indicate that the entire try-except-pass block should be replaced with a `with` statement, preventing the misuse described in the issue

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
+++ b/crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
@@ -53,13 +53,13 @@ impl Violation for SuppressibleException {
     #[derive_message_formats]
     fn message(&self) -> String {
         let SuppressibleException { exception } = self;
-        format!("Replace `try`-`except`-`pass` block with `with contextlib.suppress({exception})`")
+        format!("Use `contextlib.suppress({exception})` instead of `try`-`except`-`pass`")
     }
 
     fn fix_title(&self) -> Option<String> {
         let SuppressibleException { exception } = self;
         Some(format!(
-            "Replace `try`-`except`-`pass` with `with contextlib.suppress({exception})`"
+            "Replace `try`-`except`-`pass` with `with contextlib.suppress({exception}): ...`"
         ))
     }
 }
--- a/crates/ruff_linter/src/rules/flake8_simplify/snapshots/ruff_linter__rules__flake8_simplify__tests__SIM105_SIM105_0.py.snap
+++ b/crates/ruff_linter/src/rules/flake8_simplify/snapshots/ruff_linter__rules__flake8_simplify__tests__SIM105_SIM105_0.py.snap
@@ -1,7 +1,7 @@
 ---
 source: crates/ruff_linter/src/rules/flake8_simplify/mod.rs
 ---
-SIM105 [*] Replace `try`-`except`-`pass` block with `with contextlib.suppress(ValueError)`
+SIM105 [*] Use `contextlib.suppress(ValueError)` instead of `try`-`except`-`pass`
  --> SIM105_0.py:6:1
   |
 5 |   # SIM105
@@ -11,7 +11,7 @@ SIM105 [*] Replace `try`-`except`-`pass` block with `with contextlib.suppress(Va
 9 | |     pass
   | |________^
   |
-help: Replace `try`-`except`-`pass` with `with contextlib.suppress(ValueError)`
+help: Replace `try`-`except`-`pass` with `with contextlib.suppress(ValueError): ...`
 1  + import contextlib
 2  | def foo():
 3  |     pass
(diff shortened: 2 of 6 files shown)


crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
```diff
@@ -53,12 +53,14 @@ impl Violation for SuppressibleException {
     #[derive_message_formats]
     fn message(&self) -> String {
         let SuppressibleException { exception } = self;
-        format!("Use `contextlib.suppress({exception})` instead of `try`-`except`-`pass`")
+        format!("Replace `try`-`except`-`pass` block with `with contextlib.suppress({exception})`")
```
ntBre: Hmm, I think the `message` might actually be okay as it was. This part wasn't included in the snippet in the issue, so I didn't realize it already mentioned try-except-pass. I think I would revert this change and keep your new `fix_title`, just to keep some variety in the phrasing since users will often see both.

crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
```diff
@@ -53,12 +53,14 @@ impl Violation for SuppressibleException {
...
     }
 
     fn fix_title(&self) -> Option<String> {
         let SuppressibleException { exception } = self;
-        Some(format!("Replace with `contextlib.suppress({exception})`"))
+        Some(format!(
+            "Replace `try`-`except`-`pass` with `with contextlib.suppress({exception})`"
```
ntBre: I don't really love the "with with," but it also makes sense. I kind of wish we could make it even more clear that you need to put the code from `try` in the context manager, but we need to keep this as a single line. Do you think something like this would be helpful or also confusing?


```suggestion
            "Replace `try`-`except`-`pass` with `with contextlib.suppress({exception}): ...`"
```

crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
mgiovani: Thank you for the feedback!
You're right. I've reverted the main message back to the original and kept only the `fix_title` improvement, which addresses the core confusion from the issue.

crates/ruff_linter/src/rules/flake8_simplify/rules/suppressible_exception.rs
mgiovani: I agree with the "with with" awkwardness, but I think it's clear.
I've also implemented your idea with the ellipsis, it's a good one to keep it as a single line

PR #20549, opened 2025-09-24
[ty] Todo-types for `os.fdopen`, `NamedTemporaryFile`, and `Path.open`

sharkdp:
## Summary

This applies the trick that we use for `builtins.open` to similar functions that have the same problem. The reason is that the problem would otherwise become even more pronounced once we add understanding of the implicit type of `self` parameters, because then something like `(base_path / "test.bin").open("rb")` also leads to a wrong return type and can result in false positives.

## Test Plan

New Markdown tests


crates/ty_python_semantic/src/types.rs
```diff
@@ -9675,6 +9699,9 @@ impl<'db> KnownBoundMethodType<'db> {
                     Some(KnownClass::Bool.to_instance(db)),
                 )))
             }
+            KnownBoundMethodType::PathOpen => {
+                Either::Right(std::iter::once(Signature::todo("`Path.open` return type")))
```
sharkdp, line 9702: This is overly aggressive, because now we replace the whole signature with `(*args, **kwargs) -> @Todo`, i.e. we also have false negatives for wrong `Path.open` calls, and return a `@Todo` type for every call to `Path.open`. But manually constructing a overload set for `Path.open` seems daunting.

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -1656,72 +1671,33 @@ impl KnownFunction {
...
+
+            KnownFunction::Fdopen => {
+                // TODO: Temporary special-casing for `os.fdopen` to avoid an excessive number of
+                // false positives in lieu of proper support for PEP-614 type aliases.
+                if let [_, Some(mode), ..] = parameter_types
+                    && is_mode_with_nontrivial_return_type(db, *mode)
+                {
+                    overload.set_return_type(todo_type!("`os.fdopen` return type"));
+                }
+            }
```
AlexWaygood, line 1754: could maybe combine these two branches


```suggestion
            KnownFunction::Open | KnownFunction::Fdopen => {
                // TODO: Temporary special-casing for `builtins.open`/`os.fdopen` to avoid an excessive number of
                // false positives in lieu of proper support for PEP-613 type aliases.
                if let [_, Some(mode), ..] = parameter_types
                    && is_mode_with_nontrivial_return_type(db, *mode)
                {
                    overload.set_return_type(todo_type!("`builtins.open`/`os.fdopen` return type"));
                }
            }
```

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -1656,72 +1671,33 @@ impl KnownFunction {
...
+
+            KnownFunction::NamedTemporaryFile => {
+                // TODO: Temporary special-casing for `tempfile.NamedTemporaryFile` to avoid an excessive number of
+                // false positives in lieu of proper support for PEP-614 type aliases.
```
AlexWaygood: I guess I originally introduced this typo in my previous PR 🙈

```suggestion
                // false positives in lieu of proper support for PEP-613 type aliases.
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -387,6 +387,9 @@ impl Display for DisplayRepresentation<'_> {
             Type::KnownBoundMethod(KnownBoundMethodType::StrStartswith(_)) => {
                 f.write_str("<method-wrapper `startswith` of `str` object>")
             }
+            Type::KnownBoundMethod(KnownBoundMethodType::PathOpen) => {
+                f.write_str("<method-wrapper `open` of `Path` object>")
```
AlexWaygood: The runtime type of `Path("foo").open` is a `types.MethodType` instance rather than a `types.MethodWrapperType` instance, since `pathlib.Path` is just a pure-Python class with regular pure-Python methods

```suggestion
                f.write_str("bound method `Path.open`")
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9579,6 +9602,7 @@ impl<'db> KnownBoundMethodType<'db> {
             | KnownBoundMethodType::PropertyDunderGet(_)
             | KnownBoundMethodType::PropertyDunderSet(_) => KnownClass::MethodWrapperType,
             KnownBoundMethodType::StrStartswith(_) => KnownClass::BuiltinFunctionType,
+            KnownBoundMethodType::PathOpen => KnownClass::Path,
```
```pycon
>>> import pathlib
>>> pathlib.Path("foo").open
<bound method Path.open of PosixPath('foo')>
>>> import types
>>> type(_) is types.MethodType
True
```

```suggestion
            KnownBoundMethodType::PathOpen => KnownClass::MethodType,
```

crates/ty_python_semantic/src/types/open_calls.rs
```diff
@@ -0,0 +1,66 @@
+use super::Type;
+use crate::Db;
+
+/// Return true, if the type passed as `mode` would require us to pick a non-trivial overload of
+/// `builtins.open` / `os.fdopen` / `Path.open`.
+pub(super) fn is_mode_with_nontrivial_return_type<'db>(db: &'db dyn Db, mode: Type<'db>) -> bool {
```
AlexWaygood: this helper is only used from `function.rs`, and it might be easier to rip out the short-term hack out again later if this helper lived in `function.rs` rather than introducing a new module? Don't have a strong opinion though

sharkdp: Right. I wanted to use it for `Path.open`, but then decided otherwise. Will move it back.

crates/ty_python_semantic/src/types.rs
sharkdp: Oops. I misinterpreted the docstring.

crates/ty_python_semantic/src/types/function.rs
sharkdp: Will leave it as-is. We'll hopefully remove it soon.

PR #20550, opened 2025-09-24
[ty] Make `FileResolver::path` return a full path

ntBre:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1242

From finding references with the LSP, `FileResolver::path` is only called once, in `UnifiedFile::path`, so I went through those references, and it looked safe to make this change in every case. Most of the references are in the various output formats, where we inherited the absolute vs relative path decision from Ruff. Two other uses are as fallbacks if converting a relativized path to a string fails. Finally, we use the path for sorting and in `UnifiedFile::relative_path`.

## Test Plan

Existing tests, with snapshots updated to show absolute paths (in the `TestDb` this just added a `/` in front of the file names). I also updated the GitLab CLI test to set the `CI_PROJECT_DIR` environment variable and ran a test in GitLab CI:

image


crates/ruff_db/src/diagnostic/render/snapshots/ruff_db__diagnostic__render__json__tests__notebook_output.snap
```diff
@@ -10,7 +10,7 @@ expression: env.render_diagnostics(&diagnostics)
       "column": 10,
       "row": 2
     },
-    "filename": "notebook.ipynb",
+    "filename": "/notebook.ipynb",
```
MichaReiser, line 13: Were those paths (and in other output formatters) relative or absolute before you migrated the renderers to the new diagnostic output? Some folks might depend on paths being relative or absolute.

ntBre: They were always absolute in Ruff (they just used `SourceFile::name`, which is still the case now just via `UnifiedFile`). JSON for example:

```
> uvx ruff@0.9.0 check try.py --output-format json | jq '.[].filename'
"/tmp/tmp.9hyS01e8LF/try.py"
```

MichaReiser: I guess I'm slightly confused why the snapshots change...

ntBre: The snapshots in `ruff_db` use the `TestDb` infrastructure shared with ty. The Ruff snapshots like [`lint__output_format_json.snap`](https://github.com/astral-sh/ruff/blob/main/crates/ruff/tests/snapshots/lint__output_format_json.snap) from the Ruff CLI tests didn't change.

I probably should have made that more clear in the summary.

This shouldn't affect anything on the Ruff side, only formats using a `Db`.

PR #20551, opened 2025-09-24
[`airflow`] Add warning to `airflow.datasets.DatasetEvent` usage (`AIR301`)

Lee-W:
## Summary

`airflow.datasets.DatasetEvent` has been removed in 3 but `AssetEvent` might be added in the future

## Test Plan

update the test fixture and reorg in the second commit

### DIFF
--- a/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
@@ -656,7 +656,9 @@ fn check_name(checker: &Checker, expr: &Expr, range: TextRange) {
         // airflow.datasets
         ["airflow", "datasets", "DatasetAliasEvent"] => Replacement::None,
         ["airflow", "datasets", "DatasetEvent"] => Replacement::Message(
-            "It has been made private in Airflow 3. An AssetEvent type will be added to the apache-airflow-task-sdk in a future version.",
+            "`DatasetEvent` has been made private in Airflow 3. \
+             Use `dict[str, Any]` for the time being. \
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
```diff
@@ -655,6 +655,9 @@ fn check_name(checker: &Checker, expr: &Expr, range: TextRange) {
         },
         // airflow.datasets
         ["airflow", "datasets", "DatasetAliasEvent"] => Replacement::None,
+        ["airflow", "datasets", "DatasetEvent"] => Replacement::Message(
+            "It has been made private in Airflow 3. An AssetEvent type will be added to the apache-airflow-task-sdk in a future version.",
```
ntBre: nit: I think it might sound a little better to use `DatasetEvent` or even `airflow.datasets.DatasetEvent` in the help message instead of starting with `It`:


```suggestion
            "`DatasetEvent` has been made private in Airflow 3. An AssetEvent type will be added to the apache-airflow-task-sdk in a future version.",
```

Lee-W:
We can suggest users to use `dict[str, Any]` for the time being. and then add the replacement at a later point. WDYT? 

I'm good with holding it as well :) it's not that urgent

ntBre:
Sure, `dict[str, Any]` sounds good to me! Or we can hold off for now, whichever you prefer :)

Lee-W:
Just updated it! Merging it early would be cool, but I'm fine if we decide to wait. We should be able to have it in the next airflow task SDK release.

ntBre:
Sounds good, I'm happy to merge this and update it again later :) Thank you!

PR #20554, opened 2025-09-24
[syntax-errors]: future-feature-not-defined (F407)

11happy:
## Summary

This PR implements https://docs.astral.sh/ruff/rules/future-feature-not-defined/ (F407) as a semantic syntax error.

## Test Plan

### DIFF
--- a/crates/ruff_linter/src/rules/pyflakes/mod.rs
+++ b/crates/ruff_linter/src/rules/pyflakes/mod.rs
@@ -2967,10 +2967,7 @@ lambda: fu
             r"
         from __future__ import *
         ",
-            &[
-                Rule::UndefinedLocalWithImportStar,
-                Rule::FutureFeatureNotDefined,
-            ],
+            &[Rule::FutureFeatureNotDefined],
         );
     }
 
--- a/crates/ruff_python_parser/src/semantic_errors.rs
+++ b/crates/ruff_python_parser/src/semantic_errors.rs
@@ -56,22 +56,6 @@ impl SemanticSyntaxChecker {
         });
     }
 
-    pub fn is_feature_name(name: &str) -> bool {
-        matches!(
-            name,
-            "nested_scopes"
-                | "generators"
-                | "division"
-                | "absolute_import"
-                | "with_statement"
-                | "print_function"
-                | "unicode_literals"
-                | "barry_as_FLUFL"
-                | "generator_stop"
-                | "annotations"
-        )
-    }
-
     fn check_stmt<Ctx: SemanticSyntaxContext>(&mut self, stmt: &ast::Stmt, ctx: &Ctx) {
         match stmt {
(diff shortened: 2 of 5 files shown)


crates/ruff_linter/src/rules/pyflakes/mod.rs
```diff
@@ -2967,7 +2967,10 @@ lambda: fu
             r"
         from __future__ import *
         ",
-            &[Rule::FutureFeatureNotDefined],
+            &[
+                Rule::UndefinedLocalWithImportStar,
+                Rule::FutureFeatureNotDefined,
+            ],
```
ntBre: I don't think this should change. We may need to preserve the `__future__` check in `statement.rs` that was previously handled by the `else if`.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -65,8 +81,26 @@ impl SemanticSyntaxChecker {
...
-                if self.seen_futures_boundary && matches!(module.as_deref(), Some("__future__")) {
-                    Self::add_error(ctx, SemanticSyntaxErrorKind::LateFutureImport, *range);
+                if matches!(module.as_deref(), Some("__future__")) {
+                    for name in names {
+                        if !Self::is_feature_name(&name.name) {
+                            // test_ok valid_future_feature
+                            // from __future__ import annotations
+
+                            // test_err invalid_future_feature
+                            // from __future__ import invalid_feature
```
ntBre, line 75: One other interesting case might be multiple invalid features. I think we'll emit one diagnostic per invalid feature, but it could be nice to see that in a test case. And maybe one case with one known and one unknown diagonstic.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -56,6 +56,22 @@ impl SemanticSyntaxChecker {
         });
     }
 
+    pub fn is_feature_name(name: &str) -> bool {
```
ntBre: nit: I would probably just make this a standalone function after the `impl SemanticSyntaxChecker` block rather than an associated function. It also doesn't need to be `pub`.

I think a name like `is_known_future_feature` might be better too, now that it's not in a `future` module.

```suggestion
    fn is_known_future_feature(name: &str) -> bool {
```

crates/ruff_linter/src/rules/pyflakes/mod.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs

ntBre (approved):
Thank you!

PR #20555, opened 2025-09-24
[ty] Add tests for interactions of `@classmethod`, `@staticmethod`, and protocol method members

AlexWaygood:
## Summary

This PR adds more tests for interactions between `@classmethod`, `@staticmethod`, and subtyping/assignability checks for protocols with method members. This partially addresses @sharkdp's post-merge review comment in https://github.com/astral-sh/ruff/pull/20165#discussion_r2348190613.

There are a lot of TODOs in these tests that are tricky to fix right now due to the fact that we currently consider function-types as not being fully static if their signature includes a `self` or `cls` argument that isn't annotated. That means that doing tests regarding the type of a method when accessed on the class object itself is currently quite tricky. I expect it to be a lot easier to resolve these TODOs after https://github.com/astral-sh/ruff/pull/20517 has landed (but I also don't think it's urgent to resolve these TODOs, since `@classmethod` and `@staticmethod` members are quite rare).

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1761,7 +1761,7 @@ class `T` has a method `m` which is assignable to the `Callable` supertype of th
 
 ```py
 from typing import Protocol
-from ty_extensions import is_subtype_of, static_assert
+from ty_extensions import is_subtype_of, is_assignable_to, static_assert
 
 class P(Protocol):
     def m(self, x: int, /) -> None: ...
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -2012,6 +2030,98 @@ static_assert(not is_assignable_to(NominalReturningSelfNotGeneric, LegacyFunctio
...
+#
+# TODO: these should pass
+static_assert(not is_assignable_to(NInstanceMethod, PClassMethod))  # error: [static-assert-error]
+static_assert(not is_assignable_to(NInstanceMethod, PStaticMethod))  # error: [static-assert-error]
+
+# A nominal type with a `@staticmethod` can satisfy a protocol with a `@classmethod`
+# if the staticmethod duck-types the same as the classmethod member
+# both when accessed on the class and when accessed on an instance of the class
+# The same also applies for a nominal type with a `@classmethod` and a protocol
+# with a `@staticmethod` member
```
sharkdp, line 2107: I can understand the reasoning, but this seems a bit surprising to me. Can I extend this reasoning to instance methods as well? Does `C` implement the `HasFInstanceMethod` protocol? All type checkers seem to think so, but it does feel wrong to me. Shouldn't there be some constraints on the implicit type of the first parameter that would prevent this?
```py
from typing import Protocol

class HasFInstanceMethod(Protocol):
    def f(self, x: int, /) -> str: ...

class C:
    @classmethod
    def f(cls, *args) -> str:
        return ""

def accept(_: HasFInstanceMethod): ...

accept(C())  # pyright, mypy, pyrefly, ty: fine!
```

sharkdp (approved):
Thank you!

crates/ty_python_semantic/resources/mdtest/protocols.md
AlexWaygood: > Does `C` implement the `HasFInstanceMethod` protocol? All type checkers seem to think so, but it does feel wrong to me. Shouldn't there be some constraints on the implicit type of the first parameter that would prevent this?

Yes, I agree with you -- and I added (currently failing) tests in this PR that assert this, too ;) they're up on lines 1791-1795.

The rule I intend to implement for ty is that (given a protocol `P`, a method member `f` and a would-be nominal subtype `N`), two things must be true in order for `N` to satisfy `P`
- the get-type of `f` when accessed on instances of `N` must be consistent with the signature of `f` when accessed on instances of `P`
- the get-type of `f` on `type[N]` must be consistent with the signature of `f` on when accessed on the class object `P`

As you note, this is a bit stricter than any other type checker currently implements, so we'll see how it goes; we might have to back out of this. But I'd like to give it a go, at least.

sharkdp: > I added (currently failing) tests in this PR that assert this, too ;) they're up on lines 1791-1795.

They looked slightly different to me, because with the `*args` signature in my comment, you can even make type checkers think that the `classmethod` signature *is* compatible with a non-classmethod protocol signature.

AlexWaygood: Oh, sorry -- I missed that. Hmm, I can see how that might be counter-intuitive, but yes, I can't immediately see a principled reason whereby we'd decide that `C` shouldn't be a subtype of `HasFInstanceMethod` there...

PR #20556, opened 2025-09-24
[syntax-errors] Reimplement `break-outside-loop` (`F701`) as a semantic syntax error

11happy:
## Summary

This PR implements https://docs.astral.sh/ruff/rules/break-outside-loop/ (F701) as a semantic syntax error.

### DIFF
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -697,6 +697,7 @@ impl SemanticSyntaxContext for Checker<'_> {
                 }
             }
             SemanticSyntaxErrorKind::FutureFeatureNotDefined(name) => {
+                // F407
                 if self.is_rule_enabled(Rule::FutureFeatureNotDefined) {
                     self.report_diagnostic(
                         pyflakes::rules::FutureFeatureNotDefined { name },
@@ -705,6 +706,7 @@ impl SemanticSyntaxContext for Checker<'_> {
                 }
             }
             SemanticSyntaxErrorKind::BreakOutsideLoop => {
+                // F701
                 if self.is_rule_enabled(Rule::BreakOutsideLoop) {
                     self.report_diagnostic(pyflakes::rules::BreakOutsideLoop, error.range);
                 }


crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -614,6 +647,23 @@ impl SemanticSyntaxChecker {
                     self.seen_futures_boundary = true;
                 }
             }
+            Stmt::For(ast::StmtFor { body, orelse, .. })
+            | Stmt::While(ast::StmtWhile { body, orelse, .. }) => {
+                self.loop_stack.push(LoopContext {
+                    inside_orelse: false,
+                });
+                for stmt in body {
+                    self.visit_stmt(stmt, ctx);
```
ntBre: We shouldn't be recursing here. The `SemanticSyntaxChecker` isn't a full `Visitor`, it only checks the single `Stmt` it's given and relies on the parent visitor (either `Checker` in Ruff or `SemanticIndexBuilder` in ty) to drive the AST traversal.

I think this also means that managing the loop state isn't really feasible in the `SemanticSyntaxChecker` itself. Instead, we'll probably need to add `in_loop_context` as a method on the `SemanticSyntaxContext` trait and implement it for Ruff and ty (and the test visitor if we want to write inline tests).

You could also consider trying to add a `SemanticSyntaxContext::current_statements()` method, which would be more general and match how Ruff currently implements this check:

https://github.com/astral-sh/ruff/blob/83f80effecf682a487abfcd101ea6bbab3471740/crates/ruff_linter/src/checkers/ast/analyze/statement.rs#L54-L61

Those are just some ideas based on a quick look, hope that helps!

11happy: Sure, Thank you for these pointers, I will try them out & keep you updated : )

11happy: done, I have added a new `current_statements` .

Thank you : )

crates/ruff_linter/src/rules/pyflakes/mod.rs
ntBre, line 1: I don't think we should delete this test case. We still need to emit F701, even if the main implementation is as a syntax error.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -2777,6 +2777,9 @@ impl SemanticSyntaxContext for SemanticIndexBuilder<'_, '_> {
             self.semantic_syntax_errors.borrow_mut().push(error);
         }
     }
+    fn in_loop_context(&self) -> bool {
```
ntBre, line 2781:
```suggestion
    
    fn in_loop_context(&self) -> bool {
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -668,6 +673,7 @@ impl SemanticSyntaxChecker {
                     self.seen_futures_boundary = true;
                 }
             }
+
```
ntBre:
```suggestion
```

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -811,12 +818,36 @@ impl SemanticSyntaxContext for Checker<'_> {
             }
         )
     }
+
+    fn in_loop_context(&self) -> bool {
+        let statements = self.current_statements();
```
ntBre: I don't think we need to track a separate `Vec` of current statements on the `Checker`. This should already be available on the `SemanticModel` field, as in the code I linked earlier.


```suggestion
        let statements = self.semantic.current_statements();
```

My earlier suggestion about adding a `current_statements` method was not for `Checker`, I meant that instead of `SemanticSyntaxContext::in_loop_context`, we could consider adding `SemanticSyntaxContext::current_statements`, which would return an iterator over `Stmt` nodes:

```rust
fn current_statements(&self) -> impl Iterator<Item = &Stmt> {
    self.semantic.current_statements()
}
```

This should be easier to implement for both Ruff and ty, and then the actual `in_loop_context` check could be shared between the two. But I think `SemanticSyntaxContext::in_loop_context` is also okay, especially if you tried this already.

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -811,12 +818,36 @@ impl SemanticSyntaxContext for Checker<'_> {
...
+        let mut cur_stmt = statements.last();
+
+        for parent in statements.iter().rev().skip(1) {
+            match *parent {
+                Stmt::For(ast::StmtFor { orelse, .. })
+                | Stmt::While(ast::StmtWhile { orelse, .. }) => {
+                    let in_orelse = orelse.contains(cur_stmt.unwrap());
```
ntBre: I think we should just pass the current statement in, like in the original F701 implementation, to avoid needing to `unwrap` here.

ntBre (changes requested):
Thanks! But I think we can make a few improvements here.

crates/ruff_linter/src/rules/pyflakes/mod.rs
11happy: done

crates/ruff_linter/src/checkers/ast/mod.rs

crates/ty_python_semantic/src/semantic_index/builder.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs

PR #20557, opened 2025-09-24
[ty] Explicitly test assignability/subtyping between unions of nominal types and protocols with method members

AlexWaygood:
(Stacked on top of https://github.com/astral-sh/ruff/pull/20555; review that PR first!)

## Summary

@sharkdp very reasonably asked in https://github.com/astral-sh/ruff/pull/20165#discussion_r2348190613 whether the `invoke_descriptor_protocol` call here would produce the right result when testing assignability between a union of nominal types and a protocol with a method member: https://github.com/astral-sh/ruff/blob/83f80effecf682a487abfcd101ea6bbab3471740/crates/ty_python_semantic/src/types/protocol_class.rs#L559-L570

This wasn't something we had explicit tests for, so this PR adds some tests for it!

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -1769,6 +1769,9 @@ class P(Protocol):
 class NominalSubtype:
     def m(self, y: int) -> None: ...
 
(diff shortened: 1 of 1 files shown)


sharkdp (approved):
Thank you!

PR #20559, opened 2025-09-24
[`isort`] Clarify dependency between `order-by-type` and `case-sensitive` settings

### LINKED ISSUE #20536: Ruff Import ordering not respecting case-insensitivity compared to flake8 `import-order-style`
### Summary

Hi there,

I'm trying to fully replace `isort` and `flake8` with `ruff`. Previously with flake8 our team had used this import order style `import-order-style = appnexus` and I'm trying to simulate the same in `ruff`. 

I cannot seem to figure out how to make the imports from the same file be case-insensitive and alphabetically sorted. The `tool.ruff.lint.isort` setting `case-sensitive` is default `false` which you would think would leave this import unchanged but it does not.

This is what the import used to look like before running `ruff check --fix`
`from pydantic import BaseModel, computed_field, Field`

After running `ruff`:

`from pydantic import BaseModel, Field, computed_field`

Here is my full config for reference and I am running it as `ruff check --fix /path/to/file.py`

```toml
[tool.ruff]
exclude = ["__pycache__", ".venv", ".vscode", "**/tests/**"]

line-length = 160
indent-width = 4

[tool.ruff.lint]
select = [
    "E", # pycodestyle errors
    "W", # pycodestyle warnings
    "F", # pyflakes
    "N", # pep8-naming
]

# run isort import sorting on top of the default linting rules 
extend-select = ["I"]

[tool.ruff.lint.per-file-ignores]
"src/netsuite/soap_helper.py" = ["E221", "E231", "E222"]
"src/netsuite/soap_mixin.py" = ["E231"]
"src/netsuite/records/customer.py" = ["E231"]
"src/job.py" = ["E231"]
"src/account.py" = ["E231"]

[tool.ruff.lint.isort]
force-sort-within-sections = true
force-wrap-aliases = true
combine-as-imports = true
lines-after-imports = 2

section-order = [
    "future",
    "standard-library",
    "third-party",
    "first-party",
    "local-folder",
]

known-first-party = ["stingray"]
known-third-party = [
    "aws",
    "aws_lambda_powertools",
    "aws_xray_sdk",
    "boto3",
    "pydantic",
    "requests",
    "yattag",
]
known-local-folder = ["config", "netsuite"]

[tool.ruff.format]
line-ending = "auto"
indent-style = "space"
quote-style = "single"
```

### Version

_No response_

ntBre:
Summary
--

Fixes #20536 by linking between the isort options `case-sensitive` and `order-by-type`. The latter takes precedence over the former, so it seems good to clarify this somewhere.

I tweaked the wording slightly, but this is otherwise based on the patch from @SkylerWittman in https://github.com/astral-sh/ruff/issues/20536#issuecomment-3326097324 (thank you!)

Test Plan
--

N/a

### DIFF
--- a/ruff.schema.json
+++ b/ruff.schema.json
@@ -1843,7 +1843,7 @@
           ]
         },
         "order-by-type": {
-          "description": "Order imports by type, which is determined by case, in addition to alphabetically.\n\nNote that this takes precedence over the [`case-sensitive`](#lint_isort_case-sensitive) setting when enabled.",
+          "description": "Order imports by type, which is determined by case, in addition to alphabetically.\n\nNote that this option takes precedence over the [`case-sensitive`](#lint_isort_case-sensitive) setting when enabled.",
           "type": [
             "boolean",
             "null"


crates/ruff_workspace/src/options.rs
```diff
@@ -2302,6 +2302,9 @@ pub struct IsortOptions {
 
     /// Order imports by type, which is determined by case, in addition to
     /// alphabetically.
+    ///
+    /// Note that this takes precedence over the
```
MichaReiser: Nit: Maybe option to make it clearer what this refers to?
```suggestion
    /// Note that this option takes precedence over the
```

Issue #19138, opened 2025-07-04
Debug assert `The node has dangling comments that need to be formatted manually. Add the special dangling comments handling to fmt_fields.`


qarmin:
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
from y import(e as#
r)
```

command
```
timeout -v 300 ruff format TEST___FILE.py
```

cause this
```
warning: Detected debug build without --no-cache.
error: Panicked while formatting /opt/BROKEN_FILES_DIR/599_IDX_0_RAND_1511437192090215324352378_minimized_389.py: This indicates a bug in Ruff. If you could open an issue at:

    https://github.com/astral-sh/ruff/issues/new?title=%5BFormatter%20panic%5D

...with the relevant file contents, the `pyproject.toml` settings, and the following stack trace, we'd be very appreciative!

panicked at crates/ruff_python_formatter/src/lib.rs:80:13:
The node has dangling comments that need to be formatted manually. Add the special dangling comments handling to `fmt_fields`.
Backtrace:    0: ruff_db::panic::install_hook::{{closure}}::{{closure}}
             at ./ruff-main/crates/ruff_db/src/panic.rs:91:34
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/alloc/src/boxed.rs:1980:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/std/src/panicking.rs:841:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/std/src/panicking.rs:699:13
   4: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/std/src/sys/backtrace.rs:168:18
   5: __rustc::rust_begin_unwind
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/std/src/panicking.rs:697:5
   6: core::panicking::panic_fmt
             at /rustc/6677875279b560442a07a08d5119b4cd6b3c5593/library/core/src/panicking.rs:75:14
   7: ruff_python_formatter::FormatNodeRule::fmt
[...]

PR #20561, opened 2025-09-24
[`ruff`] improve handling of intermixed comments inside from-imports

amyreese:
Resolves the crash when attempting to format code like:

```
from x import (a as # whatever
b)
```

And chooses to format it as:
```
from x import (
    a as b,  # whatever
)
```

Fixes issue #19138

### DIFF
--- a/crates/ruff_python_formatter/src/comments/placement.rs
+++ b/crates/ruff_python_formatter/src/comments/placement.rs
@@ -1925,7 +1925,7 @@ fn handle_bracketed_end_of_line_comment<'a>(
     CommentPlacement::Default(comment)
 }
 
-/// Attach an enclosed end-of-line comment to a [`ast::StmtImportFrom`].
+/// Attach an enclosed comment to a [`ast::StmtImportFrom`].
 ///
 /// For example, given:
 /// ```python
@@ -1936,6 +1936,33 @@ fn handle_bracketed_end_of_line_comment<'a>(
 ///
 /// The comment will be attached to the [`ast::StmtImportFrom`] node as a dangling comment, to
 /// ensure that it remains on the same line as the [`ast::StmtImportFrom`] itself.
+///
+/// If the comment's preceding node is an alias, and the comment is *before* a comma:
+/// ```python
+/// from foo import (
+///     bar as baz  # comment
+///     ,
+/// )
+/// ```
+///
+/// The comment will then be attached to the [`ast::Alias`] node as a dangling comment instead,
+/// to ensure that it retains its position before the comma.
+///
+/// Otherwise, if the comment is *after* the comma or before a following alias:
+/// ```python
+/// from foo import (
+///     bar as baz,  # comment
+/// )
+///
+/// from foo import (
+///     bar,
+///     # comment
+///     baz,
+/// )
+/// ```
+///
+/// Then it will retain the default behavior of being attached to the relevant [`ast::Alias`] node
+/// as either a leading or trailing comment.
 fn handle_import_from_comment<'a>(
     comment: DecoratedComment<'a>,
     import_from: &'a ast::StmtImportFrom,
@@ -1974,7 +2001,7 @@ fn handle_import_from_comment<'a>(
             .skip_trivia()
             .next()
         {
-            // treat comments before the comma as dangling, after as trailing (default)
+            // Treat comments before the comma as dangling, after as trailing (default)
             if let Some(AnyNodeRef::Alias(alias)) = comment.preceding_node() {
                 CommentPlacement::dangling(alias, comment)
             } else {
@@ -1986,6 +2013,27 @@ fn handle_import_from_comment<'a>(
     }
 }
 
+/// Attach an enclosed comment to the appropriate [`ast::Identifier`]  within an [`ast::Alias`].
+///
+/// For example:
+/// ```python
+/// from foo import (
+///     bar  # comment
+///     as baz,
+/// )
+/// ```
+///
+/// Will attach the comment as a trailing comment on the first name [`ast::Identifier`].
+///
+/// Whereas:
+/// ```python
+/// from foo import (
+///     bar as  # comment
+///     baz,
+/// )
+/// ```
+///
+/// Will attach the comment as a leading comment on the second name [`ast::Identifier`].
 fn handle_alias_comment<'a>(
     comment: DecoratedComment<'a>,
     alias: &'a ruff_python_ast::Alias,
--- a/crates/ruff_python_formatter/src/other/alias.rs
+++ b/crates/ruff_python_formatter/src/other/alias.rs
@@ -19,6 +19,13 @@ impl FormatNodeRule<Alias> for FormatAlias {
         write!(f, [DotDelimitedIdentifier::new(name)])?;
 
         let comments = f.context().comments().clone();
+
+        // ```python
+        // from foo import (
+        //     bar  # comment
+        //     as baz,
+        // )
+        // ```
         if comments.has_trailing(name) {
             write!(
                 f,
@@ -32,8 +39,14 @@ impl FormatNodeRule<Alias> for FormatAlias {
         }
 
         if let Some(asname) = asname {
-            write!(f, [token("as"),])?;
+            write!(f, [token("as")])?;
 
+            // ```python
+            // from foo import (
+            //     bar as  # comment
+            //     baz,
+            // )
+            // ```
             if comments.has_leading(asname) {
                 write!(
                     f,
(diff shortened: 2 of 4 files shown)


crates/ruff_python_formatter/src/other/alias.rs
```diff
@@ -17,7 +18,18 @@ impl FormatNodeRule<Alias> for FormatAlias {
...
+            let comments = f.context().comments().clone();
+            let dangling = comments.dangling(item);
+            write!(
+                f,
+                [
+                    space(),
+                    token("as"),
+                    space(),
+                    asname.format(),
+                    line_suffix(&dangling_comments(dangling), 0),
```
MichaReiser: Passing 0 for the width is incorrect here. The width is used to determine how much space the `Printer` needs to reserve for the comments. Zero means that they take zero width (or the width is ignored) and they always fit if the content before fits on the line. 

The reason `line_suffix` has a width is that Ruff excludes pragma comments from the width calculation, but all other comments need to fit into the 88-character limit, or the line is split over multiple lines. 

You can use `trailing_comments` if you want to format the comments as end-of-line comments

MichaReiser (changes requested):
Thanks for looking into this. We'll need to add tests (see `resources/fixtures`). I also suggest playing with different comment combinations. E.g. the following now flips the order of comments which is always undesired:

```py
from x import (a # more 
 as # whatever
b)
```

Gets formatted to

```py
from x import (
    a as b,  # more
    # whatever
)
```

This one doesn't flip the comment order but it places the `other` comment that used to be before `b` after `b`

```py

from x import (a
 as # whatever
 # other
b)

```

Not suggesting that you should or that it makes your life easier in any way. But an alternative fix is to update `place_comment` so that it makes comments between `as` and the first alias leading comments of the first alias

amyreese:
Working on adding some test fixtures to validate results, and considering alternative approaches.

amyreese:
In this given example putting comments in every possible position, Black will maintain placement of all the comments, rather than re-flowing any elements together and merging comments:

```
# alpha
from x import (
    # bravo
    a  # charlie
    # delta
    as  # echo
    # foxtrot
    b  # golf
    # hotel
    ,  # india
    # juliet
)  # kilo
```

Do we want to ensure that we follow Black's style and preserve all of these comments in their original positions? Or do we want to have an opinion and re-flow/merge various comments to ensure that `a as b,` is always on one line when possible?

MichaReiser:
> Do we want to ensure that we follow Black's style and preserve all of these comments in their original positions? Or do we want to have an opinion and re-flow/merge various comments to ensure that a as b, is always on one line when possible?

We've used Black's comment formatting as inspiration and we try to match it for "common" comment placements but it's not a strict requirement that we have to match black precisely (I also found Black inconsistent about when it tries to be smart/opinionated and when it's strict about preserving comment placements).

We do try to preserve own-line comments as own-line comments and end-of line comments should remain end-of-line comments. It's also important to preserve the ordering of comments. 

Trying your example with a few "similar" syntax elements (this is the formatted output):

```py
# alpha
(
    # bravo
    a  # charlie
    # delta
    +  # echo
    # foxtrot
    b,  # golf
    # hotel
    # india
    # juliet
)  # kilo

# scrambles own line and end of line comments

# alpha
with (
    # bravo
    a as (  # charlie
    # delta  # echo
        # foxtrot
        b
    ),  # golf
    # hotel
    # india
    # juliet
):  # kilo
    ...

match x:
    # alpha
    case (
        # bravo
        a  # charlie
        # delta
        as  # echo
        # foxtrot
        b,  # golf
        # hotel
        # india
        # juliet
    ):  # kilo
        ...

```

amyreese:
Updated to use `trailing_comments`. It did not seem feasible to make it fully preserve all comment positions like black does, but I ensured that comment order is at least preserved, and this provides the "opinion" of keeping `a as b` on the same line. Updated fixtures/snapshots with the example from yesterday.

MichaReiser (commented):
Does this change reflow comments of already formatted code?

Could we do something similar to the `MatchAs` formatting

https://github.com/astral-sh/ruff/blob/63b3f7adae7fcc8910baa902a2d20bfdc4a28ad9/crates/ruff_python_formatter/src/pattern/pattern_match_as.rs#L27-L44

I'm asking because the formatting of 

```py
from x import (a # more 
 as # whatever
b)


from x import (a
 as # whatever
 # other
b)
```

to 

```py
from x import (
    a as b,  # more
    # whatever
)


from x import (
    a as b,  # whatever
    # other
)
```

or

```py
# alpha
from x import (
    # bravo
    a
    
    as  # echo
    # foxtrot
    b  # golf
    # hotel
)
```

where the `foxdrot` comment now comes after `b` (it was a leading comment but it now becomes a trailing comment and the indent for `golf` is off)

```py
# alpha
from x import (
    # bravo
    a as b,  # echo
    # foxtrot
      # golf
    # hotel
)
```


The indent of `# golf` is problematic because this is now a case where the formatter needs two passes to reach convergence. We need to fix this at least and add them as test cases.

amyreese:
Hmm, I'm seeing different results from you when I test this locally:

```
$ cargo run --bin ruff_python_formatter -- --emit stdout scratch.py
...

from x import (
    a as b,  # more  # whatever
)


from x import (
    a as b,  # whatever
    # other
)

# alpha
from x import (
    # bravo
    a as b,  # echo
    # foxtrot  # golf
    # hotel
)
```

MichaReiser:
Ah, sorry. I think pulling the changes this morning failed and I then tested against the previous revision. Adding those tests might still make sense because they demonstrate other issues that your PR fixes.

amyreese:
I'm revisiting the comment placement mechanisms, and attempting to better associate comments within an alias node to the identifiers they belong with.

amyreese:
Reworked the ways comments are associated, specifically for the `Alias` and `StmtImportFrom` nodes, preserving comment placement at every position within a from-import and matching black's formatting:

```
$ cargo run --bin ruff_python_formatter -- --emit stdout crates/ruff_python_formatter/resources/test/fixtures/black/cases/import_comments.py
   Compiling ruff_db v0.0.0 (/Users/amethyst/workspace/ruff/crates/ruff_db)
   Compiling ruff_python_formatter v0.0.0 (/Users/amethyst/workspace/ruff/crates/ruff_python_formatter)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.18s
     Running `target/debug/ruff_python_formatter --emit stdout crates/ruff_python_formatter/resources/test/fixtures/black/cases/import_comments.py`
# ensure trailing comments are preserved
import x  # comment
from x import a  # comment
from x import a, b  # comment
from x import a as b  # comment
from x import a as b, b as c  # comment

# ensure intermixed end- and own-line comments are all preserved
from x import (  # one
    # two
    a  # three
    # four
    ,  # five
    # six
)  # seven

from x import (  # alpha
    # bravo
    a  # charlie
    # delta
    as  # echo
    # foxtrot
    b  # golf
    # hotel
    ,  # india
    # juliet
)  # kilo
```

Passing ruff's output black confirms no changes:

```
$ cargo run -p ruff -- format - < crates/ruff_python_formatter/resources/test/fixtures/black/cases/import_comments.py | uvx black -
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/ruff format -`
# ensure trailing comments are preserved
import x  # comment
from x import a  # comment
from x import a, b  # comment
from x import a as b  # comment
from x import a as b, b as c  # comment

# ensure intermixed end- and own-line comments are all preserved
from x import (  # one
    # two
    a  # three
    # four
    ,  # five
    # six
)  # seven

from x import (  # alpha
    # bravo
    a  # charlie
    # delta
    as  # echo
    # foxtrot
    b  # golf
    # hotel
    ,  # india
    # juliet
)  # kilo
All done! ✨ 🍰 ✨
1 file left unchanged.
```

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -1963,10 +1967,48 @@ fn handle_import_from_comment<'a>(
...
         CommentPlacement::dangling(comment.enclosing_node(), comment)
     } else {
-        CommentPlacement::Default(comment)
+        if let Some(SimpleToken {
+            kind: SimpleTokenKind::Comma,
+            ..
+        }) = SimpleTokenizer::starts_at(comment.start(), source)
+            .skip_trivia()
+            .next()
+        {
```
MichaReiser, line 2003: Nit: The `place_comments` logic tends to be complicated and its often non obvious what cases each branch is handling. That's why we tend to add inline comments that show examples of the patterns they match on (see the comments above).

Could you add such comments for this branch and to `handle_alias_comment` too? 

```
// this should be a dangling comment but only if it comes before the `,`
// ```python
// from foo import (
//      baz # comment     
//      , 
//      bar 
// )
// ```

File: crates/ruff_python_formatter/src/other/alias.rs [hunk 3]
```diff
@@ -15,9 +16,47 @@ impl FormatNodeRule<Alias> for FormatAlias {
...
+                    hard_line_break()
+                ]
+            )?;
+        } else if asname.is_some() {
+            write!(f, [space()])?;
+        }
+
         if let Some(asname) = asname {
-            write!(f, [space(), token("as"), space(), asname.format()])?;
+            write!(f, [token("as"),])?;
```
MichaReiser:
```suggestion
            write!(f, [token("as")])?;
```

amyreese:
Documented placement functions and changes, as well as in the alias formatter.  Also added more test cases and updated the logic to better match black style and merge lines/comments when no own-line comments are between the alias and the trailing comma:

Eg,
```py
from foo import (
    bar  # one
    ,  # two
```

Becomes:
```py
from foo import (
    bar,  # one  # two
)
```

Edit: to be clear, this change is also closer to the existing behavior of Ruff, so even though it changes formatting, it's no different than what Ruff already does *today*:

```
main $ cargo run -p ruff --  format --diff -
...

from foo import (
    bar  # comment
    ,  # another
    baz,
)
^D
@@ -1,5 +1,4 @@
 from foo import (
-    bar  # comment
-    ,  # another
+    bar,  # comment  # another
     baz,
 )

> [1]
```

PR #20563, opened 2025-09-25
[`airflow`]: rename `AutoImport` as `Rename` (internal)

Lee-W:
## Summary

Since we are trying to import both `AutoImport` and `SourceModuleMoved`, the previous naming was not as descriptive. Renaming it to `Rename` better reflects the intention.

## Test Plan

no functionality change


ntBre (approved):
Makes sense to me, thanks!

PR #20564, opened 2025-09-25
[`pyflakes`] Fix false positive for `__class__` in lambda expressions within class definitions (`F821`)

### LINKED ISSUE #20562: F821 false positive for `__class__` in lambda expression in class definition
### Summary

[`undefined-name` (F821)](https://docs.astral.sh/ruff/rules/undefined-name/) has a false positive for `__class__` in a lambda expression within a class definition. [Example](https://play.ruff.rs/abe91404-af96-4ccf-aebb-bd92daa872d4):
```console
$ cat >f821.py <<'# EOF'
class C:
    f = lambda self: __class__
print(C().f().__name__)
# EOF

$ python f821.py
C

$ ruff --isolated check f821.py --select F821 --output-format concise -q
f821.py:2:22: F821 Undefined name `__class__`
```

### Version

danparizher:
## Summary

Fixes #20562

### DIFF
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -2974,10 +2974,37 @@ impl<'a> Checker<'a> {
                     unreachable!("Expected Expr::Lambda");
                 };
 
+                // Here we add the implicit scope surrounding a lambda which allows code in the
+                // lambda to access `__class__` at runtime when the lambda is defined within a class.
+                // See the `ScopeKind::DunderClassCell` docs for more information.
+                let added_dunder_class_scope = if self
+                    .semantic
+                    .parent_scope(self.semantic.current_scope())
+                    .is_some_and(|parent| parent.kind.is_class())
+                {
+                    self.semantic.push_scope(ScopeKind::DunderClassCell);
+                    let binding_id = self.semantic.push_binding(
+                        TextRange::default(),
+                        BindingKind::DunderClassCell,
+                        BindingFlags::empty(),
+                    );
+                    self.semantic
+                        .current_scope_mut()
+                        .add("__class__", binding_id);
+                    true
+                } else {
+                    false
+                };
+
                 if let Some(parameters) = parameters {
                     self.visit_parameters(parameters);
                 }
                 self.visit_expr(body);
+
+                // Clean up the dunder class scope if we added it
+                if added_dunder_class_scope {
+                    self.semantic.pop_scope();
+                }
             }
         }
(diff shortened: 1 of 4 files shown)


crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -2977,7 +2977,34 @@ impl<'a> Checker<'a> {
                 if let Some(parameters) = parameters {
                     self.visit_parameters(parameters);
                 }
+
+                // Here we add the implicit scope surrounding a lambda which allows code in the
+                // lambda to access `__class__` at runtime when the lambda is defined within a class.
+                // See the `ScopeKind::DunderClassCell` docs for more information.
+                let added_dunder_class_scope = if self
```
ntBre: I was picturing this in `visit_stmt` like we did for function definitions rather than in the deferred visit. I guess lambdas aren't statements, so the analogous code would actually fit here in `visit_expr`:

https://github.com/astral-sh/ruff/blob/e4ac9e90419841a338d6909af01acc8548c8b3c3/crates/ruff_linter/src/checkers/ast/mod.rs#L1620-L1632

I think that should work because we push a `lambda` snapshot that gets used here like we do for functions. The one downside is that the popping for lambdas happens a bit farther down, unlike functions:

https://github.com/astral-sh/ruff/blob/e4ac9e90419841a338d6909af01acc8548c8b3c3/crates/ruff_linter/src/checkers/ast/mod.rs#L2062-L2063

danparizher:
Yeah, that approach makes sense to me. Thanks for the review!

ntBre:
Hmm, it looks like there are some new false positives in the ecosystem report. I was hoping it was something simple with the scope popping, which is why I pushed a commit, but that didn't end up helping. We're flagging a case like this with F401:

```py
import uuid


class C:
    uuid = lambda: str(uuid.uuid4())
```

and I don't fully understand why. We'll need to figure it out before we can merge this, though. Would you mind taking a look?

danparizher:
The issue was that the `DunderClassCell` scope was being added during lambda definition processing instead of during lambda body analysis, which caused false positives when lambdas within classes referenced module-level variables like `uuid`.

I fixed it by moving the scope logic from `visit_expr` to `visit_deferred_lambdas` (similar to how function definitions work) and changing the scope detection to check the entire scope hierarchy rather than just the current scope, ensuring `__class__` is accessible in lambdas within classes while preventing false positives for other variables.

ntBre:
Ah okay, I guess my earlier suggestion about moving it to `visit_expr` was wrong then. Thank you for debugging and sorry for the bad suggestion!

PR #20565, opened 2025-09-25
Use `Annotation::tags` instead of hardcoded rule matching in ruff server

### LINKED ISSUE #20532: Use `Annotation::tags` instead of hard coding the rules in ruff server
Instead of hard coding the rules in the ruff server:

https://github.com/astral-sh/ruff/blob/a3ec8ca9dfb484775f38712681d1174260d2bf4b/crates/ruff_server/src/lint.rs#L341-L349

Set the `unnecessary` tag on the diagnostic's primary annotation when creating the `"F401" | "F841" | "RUF059"` diagnostics

danparizher:
## Summary

Fixes #20532

## Test Plan

Tested the specific rules by creating a test file, verified the rules are still identified

### DIFF
--- a/crates/ruff_server/src/lint.rs
+++ b/crates/ruff_server/src/lint.rs
@@ -287,7 +287,7 @@ fn to_lsp_diagnostic(
         let code = code.to_string();
         (
             Some(severity(&code)),
-            tags(&code),
+            tags(diagnostic),
             Some(lsp_types::NumberOrString::String(code)),
         )
     } else {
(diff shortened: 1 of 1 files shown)


danparizher:
Thanks for the tips, I'll try these out today!

danparizher:
I was able to test it out with the config you suggested and added the tags in. Let me know if it looks good and if you get the same results!

MichaReiser:
Thank you. This is looking great.

PR #20567, opened 2025-09-25
Add 'Finding ways to help' to CONTRIBUTING.md

MichaReiser:
## Summary

I took this directly from https://github.com/astral-sh/uv/pull/15966. I only changed the URLs to point to Ruff and replaced all mentions of uv with Ruff

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,21 +7,35 @@ Welcome! We're happy to have you here. Thank you in advance for your contributio
 > This guide is for Ruff. If you're looking to contribute to ty, please see [the ty contributing
 > guide](https://github.com/astral-sh/ruff/blob/main/crates/ty/CONTRIBUTING.md).
 
-## The Basics
+## Finding ways to help
 
-Ruff welcomes contributions in the form of pull requests.
+We label issues that would be good for a first time contributor as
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -7,21 +7,35 @@ Welcome! We're happy to have you here. Thank you in advance for your contributio
...
+important to make sure we have consensus on the solution to the problem first.
+
+Outside of issues with the labels above, issues labeled as
+[`bug`](https://github.com/astral-sh/ruff/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug%22) are the
+best candidates for contribution. In contrast, issues labeled with `needs-decision` or
+`needs-design` are _not_ good candidates for contribution. Please do not open pull requests for
+issues with these labels.
+
+Please do not open pull requests for new features without prior discussion. While we appreciate
+exploration of new features, we will almost always close these pull requests immediately. Adding a
```
AlexWaygood: Maybe this is our goal moving forwards, but I just don't think it's true that this is our current practice. I'd go with slightly softer language here


```suggestion
exploration of new features, we will often close these pull requests immediately. Adding a
```

AlexWaygood (approved):

CONTRIBUTING.md
MichaReiser: Makes sense

PR #20568, opened 2025-09-25
[ty] Remove hack in protocol satisfiability check

sharkdp:


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -1901,21 +1901,23 @@ from typing_extensions import TypeVar, Self, Protocol
 from ty_extensions import is_equivalent_to, static_assert, is_assignable_to, is_subtype_of
 
 class NewStyleClassScoped[T](Protocol):
-    def method(self, input: T) -> None: ...
+    def method(self: Self, input: T) -> None: ...
```
sharkdp, line 1904: I can remove these `Self` annotations again in the type-of-self PR.

crates/ty_python_semantic/src/types/protocol_class.rs
```diff
@@ -573,16 +572,6 @@ impl<'a, 'db> ProtocolMember<'a, 'db> {
 
                 let proto_member_as_bound_method = method.bind_self(db);
```
AlexWaygood: this variable is now only used once, so it probably doesn't need to even be assigned; you can just inline it

AlexWaygood (approved):
LGTM, but I'd love to add some of the tests I added in #20569 (see https://github.com/astral-sh/ruff/pull/20569#issuecomment-3333459036)

PR #20569, opened 2025-09-25
[ty] Limit hack in protocol satisfiability to only apply to subtyping (not assignability)

AlexWaygood:
## Summary

Following https://github.com/astral-sh/ruff/commit/742f8a4ee6d0730c7eaec081c80babdabc62a8c6, we now do sometimes consider `Type::TypeVar`s to be assignable to other types, and other types to be assignable to `Type::TypeVar`s. That means that this hack in our protocol satisfiability logic is needed _less_ than it used to be: all tests (including some new ones) continue to pass if we apply the hack _only_ when the `relation` passed is `TypeRelation::Subtyping`

## Test Plan

`cargo test -p ty_python_semantic`

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -2016,16 +2016,21 @@ class NominalReturningSelfNotGeneric:
 # TODO: should pass
 static_assert(is_equivalent_to(LegacyFunctionScoped, NewStyleFunctionScoped))  # error: [static-assert-error]
 
+static_assert(is_assignable_to(NominalNewStyle, NewStyleFunctionScoped))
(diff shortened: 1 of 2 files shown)


sharkdp:
It took me quite a bit of time (and your help) to understand what was going on on the type-of-self PR, so I would be in favor of a solution that *completely* removes this hack (https://github.com/astral-sh/ruff/pull/20568). The ecosystem impact seems to be the same.

AlexWaygood:
> It took me quite a bit of time (and your help) to understand what was going on on the type-of-self PR, so I would be in favor of a solution that _completely_ removes this hack (#20568). The ecosystem impact seems to be the same.

I'm okay with that. Would you mind copying over the test additions from this version? (Explicit tests that assignability is not affected by the removal of the hack (only subtyping), and a couple of new regression tests for the fixed issue in the typing conformance suite)

PR #20571, opened 2025-09-25
[ty] Ecosystem analyzer: timing report

sharkdp:


sharkdp:
I will merge this to test it on a few branches. If it turns out to be more annoying than helpful, we can easily remove it.

PR #20577, opened 2025-09-25
[ty] Update salsa to fix hang when cycle head panics

MichaReiser:
## Summary

Update Salsa to pull in https://github.com/salsa-rs/salsa/pull/993

## Test Plan

Tested that mypy primer no longer hangs for https://github.com/astral-sh/ruff/pull/20477 when using the new version of Salsa

### DIFF
--- a/crates/ty/src/logging.rs
+++ b/crates/ty/src/logging.rs
@@ -120,8 +120,10 @@ pub(crate) fn setup_tracing(
     } else {
         match level {
             VerbosityLevel::Default => {
-                // Show warning traces
-                EnvFilter::default().add_directive(LevelFilter::WARN.into())
+                // Show warning traces for ty and ruff but not for other crates
+                EnvFilter::default()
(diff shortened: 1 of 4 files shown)


crates/ty/src/logging.rs
```diff
@@ -120,8 +120,10 @@ pub(crate) fn setup_tracing(
     } else {
         match level {
             VerbosityLevel::Default => {
-                // Show warning traces
-                EnvFilter::default().add_directive(LevelFilter::WARN.into())
+                // Show warning traces for ty and ruff but not for other crates
+                EnvFilter::default()
+                    .add_directive("ty=warn".parse().unwrap())
+                    .add_directive("ruff=warn".parse().unwrap())
```
MichaReiser, line 126: Salsa now logs a tracing message with severity `warn` if it hits a cycle that doesn't converge. I don't think we want to show that by default.

PR #7047, opened 2023-09-01
Fix placement for comments within f-strings concatenations

charliermarsh:
## Summary

Restores the dangling comment handling for f-strings, which broke with the parenthesized expression code.

Closes https://github.com/astral-sh/ruff/issues/6898.

## Test Plan

`cargo test`

No change in any of the similarity indexes or changed file counts:

| project      | similarity index  | total files       | changed files     |
|--------------|------------------:|------------------:|------------------:|
| cpython      |           0.76083 |              1789 |              1632 |
| django       |           0.99957 |              2760 |                67 |
| transformers |           0.99927 |              2587 |               468 |
| twine        |           0.99982 |                33 |                 1 |
| typeshed     |           0.99978 |              3496 |              2173 |
| warehouse    |           0.99818 |               648 |                24 |
| zulip        |           0.99942 |              1437 |                32 |

Issue #6898, opened 2023-08-26
formatter panic with comment between joined f-strings


davidszotten:
a3d4f08f moved the f-string comment handling added in f091b46 (which unfortunately didn't add enough test cases)

```
(
    f'{1}' # comment
    f'{2}'
)
```

panics with 

```
Unexpected token between nodes: `"' "`', crates/ruff_python_formatter/src/comments/placement.rs:122:17
```

cnpryer:
What's the correct way to read `\`"' "\`'`?

cnpryer:
I imagine the `"`s are just lexing noise?

https://github.com/astral-sh/ruff/blob/fbc9b5a604c98ab0e676a576a6e26be275efc364/crates/ruff_python_formatter/src/comments/placement.rs#L122-L126

charliermarsh:
I think the `SimpleTokenizer` intentionally doesn't support lexing strings, so it hits the first quote and then lexes any subsequent characters as bogus. I'm not quite familiar enough with the f-string comment handling to be sure how to solve this, but I can play around with it for a sec to get an idea.

charliermarsh:
Yeah, makes sense. I can probably restore it.

davidszotten:
i found this when trying to test #6365 after rebasing on top of the refactor in the pr description.

PR #20580, opened 2025-09-25
Remove unnecessary f-string handling in comment placement

dylwil3:
In #7047 some logic was added to comment placement in order to address a panic in #6898 . I believe that panic was caused by old behavior of `SimpleTokenizer` that is no longer present, because it does not occur if the new logic is simply removed.

This was discovered while trying to figure out the potential side-effects of #20578 - I originally was going to augment the logic to include t-strings here, but discovered that you could instead just remove it (unless I missed something?)

### DIFF
--- a/crates/ruff_python_formatter/src/comments/placement.rs
+++ b/crates/ruff_python_formatter/src/comments/placement.rs
@@ -73,20 +73,6 @@ fn handle_parenthesized_comment<'a>(
     comment: DecoratedComment<'a>,
     source: &str,
 ) -> CommentPlacement<'a> {
-    // As a special-case, ignore comments within f-strings, like:
-    // ```python
-    // (
-    //     f'{1}' # comment
-    //     f'{2}'
-    // )
-    // ```
-    // These can't be parenthesized, as they must fall between two string tokens in an implicit
-    // concatenation. But the expression ranges only include the `1` and `2` above, so we also
-    // can't lex the contents between them.
-    if comment.enclosing_node().is_expr_f_string() {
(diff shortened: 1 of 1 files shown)


MichaReiser:
It's very likely that this code is unused but it might be for a different reason than outlined in the PR and I'd prefer to fully understand why before removing this code.

>  I believe that panic was caused by old behavior of SimpleTokenizer that is no longer present, because it does not occur if the new logic is simply removed.

Can you explain what this new logic is? Reading https://github.com/astral-sh/ruff/issues/6898. My impression is that the issue is that the `SimpleTokenizer` doesn't support lexing strings. This hasn't changed. 

To me, it also seems that the comment is still true because the interpolated element's range only covers `{1}` but maybe that's fine?

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -73,20 +73,6 @@ fn handle_parenthesized_comment<'a>(
...
-    // ```python
-    // (
-    //     f'{1}' # comment
-    //     f'{2}'
-    // )
-    // ```
-    // These can't be parenthesized, as they must fall between two string tokens in an implicit
-    // concatenation. But the expression ranges only include the `1` and `2` above, so we also
-    // can't lex the contents between them.
-    if comment.enclosing_node().is_expr_f_string() {
```
MichaReiser, line 86: I can't comment there but doesn't this change result in the `SimpleTokenizer` trying to lex `' # comment` on line 136. This would be invalid because the `SimpleTokenizer` should never be pointed inside a string.

Issue #14220, opened 2024-11-09
[Feature Request] Guard against `==` for float comparison


DanielYang59:
## Feature request: Guard against `==` for float comparison

I'm not sure if this is doable from `ruff`'s side as this might need optional static typing (`mypy` for example). 

I could think of the following cases where `==` should be discouraged (or avoided altogether as it's pretty unreliable):

```python
# directly compare floats
1.0 == 1.0 + 1E8

# floats generated by evaluation
x = 1 / 3
if x == 0.333:
    do something

# sequences of floats
[1.0, ] == [1.0 + 1E8, ]

```

dylwil3:
I think that's a great idea! I think without type inference we wouldn't be able to stray far from float literals (and perhaps basic expressions built from them), so it may make sense to wait until type inference is available to implement this.

DanielYang59:
Thanks for the input, that would be hugely helpful.

Some similar scenarios came to my mind, perhaps might consider them if possible?

### Use `in` to check membership
```python
0.0 in [0, 1, 2]
```

### Use `==` to check `None`
```python
if a == None:

```

dylwil3:
Nice! The latter does not require type checking and exists already as E711: https://docs.astral.sh/ruff/rules/none-comparison/

DanielYang59:
Okay thanks I forgot about this. Appreciate that!

Anselmoo:
> 
> I could think of the following cases where `==` should be discouraged (or avoided altogether as it's pretty unreliable, I cannot think of much use cases we need such comparison):
> 

As well as inequality `!=`?

DanielYang59:
> As well as inequality !=?

Yes, I assume`a != b` is equivalent to `not (a == b)`

Anselmoo:
> > As well as inequality !=?
> 
> Yes, I assume`a != b` is equivalent to `not (a == b)`

**Further Reading & Reference:**
	•	SonarSource RSPEC-1244: [Floating point numbers should not be tested for equality](https://rules.sonarsource.com/python/RSPEC-1244/)

ntBre:
We just landed `RUF069`, which covers the direct equality comparison part of this issue. I kind of assume that will cover the majority of cases, but I think we can leave this open in case we want to consider expanding the rule or adding a different rule for sequence comparison and membership testing.

DanielYang59:
That is GREAT thanks a lot for working on that ;)

alexprengere:
[...]

PR #20585, opened 2025-09-25
[`pycodestyle`] New rule float-comparison (`E723`)

chirizxc:
## Summary

Part of https://github.com/astral-sh/ruff/issues/14220

## Test Plan

`cargo nextest run e723`


chirizxc:
We need more tests and cases where the rule can give FP and so on

chirizxc:
Also, I randomly picked CODE for this rule

ntBre:
I haven't reviewed, but if there's not a corresponding E723 rule in pycodestyle, I would default to making this a ruff rule. We have a PR in progress for RUF066, so RUF067 would be the next free code.

chirizxc:
изображение

+ [examples/topics/categorical/slope_graph.py:29:10:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/examples/topics/categorical/slope_graph.py#L29) RUF067 Comparison `df["year"] == 2000.0` should be replaced by `math.isclose(df["year"], 2000.0, rel_tol=1e-09, abs_tol=1e-09)`
+ [examples/topics/categorical/slope_graph.py:29:35:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/examples/topics/categorical/slope_graph.py#L29) RUF067 Comparison `df["year"] == 2010.0` should be replaced by `math.isclose(df["year"], 2010.0, rel_tol=1e-09, abs_tol=1e-09)`


This breaks the behavior of code and results in:
```
TypeError: cannot convert the series to <class 'float'>
````

**Comment by chirizxc** [comment]
What should we do in such a case? Should we somehow mark this fact in the documentation or what should we do, because we will not be able to check in general that such calls will not lead to some code breakages

**Comment by chirizxc** [comment]
The rule detects potentially problematic float comparisons correctly, but we should not suggest replacing only `math.isclose()` but also `numpy.isclose()`

**Comment by ntBre** [comment]
Would it help with the false positives, and maybe with the numpy cases, to apply the rule only in boolean contexts? I think we have other rules that apply only to expressions used as `if` conditions or in negated `not` expressions, for example. We even have a semantic model helper for it:

https://github.com/astral-sh/ruff/blob/5f136c2e35257e55dfbfd8afc7093aaded63a5b6/crates/ruff_python_semantic/src/model.rs#L1974

This rule is going to require type inference to get exactly right anyway, so this might be a good way to be more conservative in the meantime.

I also only gave the implementation a quick skim and could have overlooked this, but does this rule apply to comparisons like `<=` and `>` too? I would probably expect those to be allowed since something like `delta < 1e-8` should work as expected, I'm assuming that's what `math.isclose` does under the hood. Along those lines, we may want to rename the rule to `float-equality-comparison`, if it only applies to equality.

**Comment by chirizxc** [comment]
> Would it help with the false positives, and maybe with the numpy cases, to apply the rule only in boolean contexts? I think we have other rules that apply only to expressions used as `if` conditions or in negated `not` expressions, for example. We even have a semantic model helper for it:
> 
> https://github.com/astral-sh/ruff/blob/5f136c2e35257e55dfbfd8afc7093aaded63a5b6/crates/ruff_python_semantic/src/model.rs#L1974
> 
> This rule is going to require type inference to get exactly right anyway, so this might be a good way to be more conservative in the meantime.
> 
> I also only gave the implementation a quick skim and could have overlooked this, but does this rule apply to comparisons like `<=` and `>` too? I would probably expect those to be allowed since something like `delta < 1e-8` should work as expected, I'm assuming that's what `math.isclose` does under the hood. Along those lines, we may want to rename the rule to `float-equality-comparison`, if it only applies to equality.

Yes, this only applies to `==` and `!=`, just as it works in SonarQube.

File: crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+                left: locator.slice(left.range()).to_string(),
+                right: locator.slice(right.range()).to_string(),
+                operand: operand.to_string(),
+            },
+            TextRange::new(left.start(), right.end()),
+        );
+    }
+}
+
+fn has_float(expr: &Expr, semantic: &SemanticModel) -> bool {
```
ntBre, line 153: Could we use `ResolvedPythonType` for this check?

https://github.com/astral-sh/ruff/blob/b2b9d91859f2abc017e35d22f01696f141f6f5b1/crates/ruff_python_semantic/src/analyze/type_inference.rs#L68

It looks roughly similar, but I'm not sure if it covers exactly the same cases.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// **Use `math.isclose()` for scalar values:**
+/// - Comparing individual float numbers
+/// - Working with regular Python variables (not arrays)
+/// - When you need a single `True`/`False` result
+///
+/// **Use `numpy.isclose()` for array-like objects:**
+/// - Comparing `pandas` Series, `numpy` arrays, or other vectorized objects
+/// - When you need element-wise comparison of arrays
+/// - Working in data science contexts with vectorized operations
```
ntBre: Does the rule currently apply to numpy/pandas values? I think I would kind of prefer only to cover floats and only recommend the standard library `math.isclose` for now. That gives a cleaner suggestion instead of offering two alternatives in the diagnostic message and hopefully avoids some false positives.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+        .zip(&compare.ops)
+        .filter(|(_, op)| matches!(op, CmpOp::Eq | CmpOp::NotEq))
+        .filter(|((left, right), _)| has_float(left, semantic) || has_float(right, semantic))
+        .map(|((left, right), op)| (left, right, op))
+    {
+        checker.report_diagnostic(
+            FloatEqualityComparison {
+                left: locator.slice(left.range()).to_string(),
+                right: locator.slice(right.range()).to_string(),
+                operand: operand.to_string(),
```
ntBre: I think we should be able to avoid allocating a string here and just make `FloatEqualityComparison` hold `&'a str`s.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl Violation for FloatEqualityComparison {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        format!(
+            "Comparison `{} {} {}` should be replaced by `math.isclose()` or `numpy.isclose()`",
+            self.left, self.operand, self.right,
+        )
```
ntBre: Let's unpack the struct here so that the variable names render in the documentation. We extract this `format!` argument for the description in the rules table.


```suggestion
    fn message(&self) -> String {
        let FloatEqualityComparison { left, right, operand } = self;
        format!(
            "Comparison `{left} {operand} {right}` should be replaced by `math.isclose()` or `numpy.isclose()`"
        )
```

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "0.14.3")]
+pub(crate) struct FloatEqualityComparison {
+    pub left: String,
+    pub right: String,
+    pub operand: String,
+}
+
+impl Violation for FloatEqualityComparison {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
```
ntBre: There's no fix yet right? We should set this to `Never` or just delete this line since `Never` is the default.


```suggestion
```

crates/ruff_linter/src/codes.rs
```diff
@@ -1061,6 +1061,8 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Ruff, "065") => rules::ruff::rules::LoggingEagerConversion,
         (Ruff, "066") => rules::ruff::rules::PropertyWithoutReturn,
 
+        (Ruff, "067") => rules::ruff::rules::FloatEqualityComparison,
```
ntBre: Sorry for the trouble, but we landed another PR using RUF067 to avoid introducing gaps in the rule codes, so you may run into conflicts here. I'm happy to renumber the rule if you want, I know it can be a pain.

chirizxc: There's no problem with that, we just need to find out what the next free rule code.

chirizxc:
> Thank you! And sorry for the delay on the review.
> 
> I think this will be a really helpful rule, but I'm a bit wary of recommending `np.isclose`. What do you think about trying to restrict the rule to plain floats for now? We could always expand the scope later.
> 
> My other comments were just small suggestions.

Then we will severely restrict the rule, since comparing two floats is less common

chirizxc:
~Essentially, `math.isclose()` / `numpy.isclose()` should cover 100% of cases.~

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
chirizxc: I think we won't be able to avoid false positives, or we'll have to compare only floats. There is a note about this in the [SonarQube documentation](https://rules.sonarsource.com/python/RSPEC-1244/):

Whenever attempting to compare float values, it is important to consider the inherent imprecision of floating-point arithmetic.

One common solution to this problem is to use a tolerance value (also called epsilon) to define an acceptable range of difference between two floats. A tolerance value may be relative (based on the magnitude of the numbers being compared) or absolute. Note that comparing a value to 0 is a special case: as it has no magnitude, it is important to use an absolute tolerance value.

The math.isclose function allows to compare floats with a relative and absolute tolerance. One should however be careful when comparing values to 0, as by default, the absolute tolerance of `math.isclose` is 0.0 (this case is covered by rule [S6727](https://rules.sonarsource.com/python/RSPEC-6727)) . Depending on the library you’re using, equivalent functions exist, with possibly different default tolerances (e.g `numpy.isclose` or `torch.isclose` which are respectively designed to work with numpy arrays and pytorch tensors).

If precise decimal arithmetic is needed, another option is to use the Decimal class of the decimal module, which allows for exact decimal arithmetic.

chirizxc: Should we also include such a note?

ntBre:
> Then we will severely restrict the rule, since comparing two floats is less common

I think you might be convincing me. cc @amyreese since you and I talked about this a bit too.

It feels slightly awkward to me to recommend both options in the error message. If we do want to keep flagging both types of issues, which will make the rule more useful, two other options would be:
- only recommend `math.isclose` and move the numpy suggestion (and maybe pytorch too, like sonar) into the rule documentation
- try to infer the types and recommend `math.isclose` or `numpy.isclose` depending on the type

The first of these is probably more practical.

I guess a third option would be to change the main error message to something like "Unreliable floating point equality comparison" and move the replacement suggestion into the `fix_title`, making it a bit less prominent. I kind of like this idea anyway.

But I'm curious to hear other thoughts from you and Amy!

chirizxc:
> * try to infer the types and recommend `math.isclose` or `numpy.isclose` depending on the type

I think it's impossible to do that without `ty` types for now.

> * only recommend `math.isclose` and move the numpy suggestion (and maybe pytorch too, like sonar) into the rule documentation

Then the error message should be clearer, so that people look at the documentation about it

I like the third option overall, but then the error message will become less clear. I mean that people will have to refer to the documentation to understand it. (understand more precisely how to replace the current code in a more correct way)

chirizxc:
> * only recommend `math.isclose` and move the numpy suggestion (and maybe pytorch too, like sonar) into the rule documentation

Here, it would be better to quote the entire sentence from the `SonarQube` documentation, to describe cases where the library itself provides such an opportunity: 
> Depending on the library you’re using, equivalent functions exist, with possibly different default tolerances (e.g numpy.isclose or torch.isclose which are respectively designed to work with numpy arrays and pytorch tensors).

crates/ruff_linter/resources/test/fixtures/ruff/RUF070.py
```diff
@@ -0,0 +1,39 @@
...
+def foo(a, b):
+    return a == b - 0.1
+
+print(x == 3.0 == 3.0)
+print(1.0 == x == 2)
+
+assert x == float(1)
+assert (a + b) == 1.0
+assert -x == 1.0
+assert -x == 1.0
```
sobolevn:
```suggestion
assert -x == 1.0
assert -x == -1.0
```

these two checks are identical right now. let's use `UnaryOp` with a `float` in one of them.

amyreese:
I'm in favor of just having the documentation clear about `math.isclose()` being a baseline recommendation, and point to docs for ecosystem-specific alternatives.

chirizxc:
So, we choose the option to rename error message to "Unreliable floating point equality comparison" and update the documentation accordingly?

chirizxc:
@ntBre Are there any restrictions on the length of documentation for the rule🤔? It has become longer in this commit.

MichaReiser:
> @ntBre Are there any restrictions on the length of documentation for the rule🤔? It has become longer in this commit.

Not really. It should be concise, but there's no upper limit on how long the documentation is allowed to be

chirizxc:
> Thank you! This looks good to go! Can you re-code this to RUF068 to avoid leaving a gap? I'm happy to do it instead, if you prefer.

I am generally not opposed to it.

chirizxc:
https://github.com/astral-sh/ruff/pull/22114

It seems that `RUF068` is already taken.

ntBre:
> #22114
> 
> It seems that `RUF068` is already taken.

Oh yeah, I guess the diff was outdated. I was just looking at the changes in codes.rs here. `RUF069` looks like the next one then!

chirizxc:
I also slightly updated the code to find the following cases:
```py
def foo(a, b):
    return a == b - 0.1

def foo(a, b):
    return a == b - float(“2”)
```
Ecosystem report should become larger (upd. 1008 -> 1148)

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+fn has_float(expr: &Expr, semantic: &SemanticModel) -> bool {
+    match ResolvedPythonType::from(expr) {
+        ResolvedPythonType::Atom(PythonType::Number(NumberLike::Float)) => true,
+        ResolvedPythonType::Atom(PythonType::Number(NumberLike::Complex)) => true,
+        _ => {
+            match expr {
+                Expr::Call(ast::ExprCall { func, .. }) => {
+                    semantic.match_builtin_expr(func, "float")
```
dscorbett: Should this also check for `complex` calls? `complex` literals are already accounted for.

chirizxc: done i think

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,190 @@
...
+///
+/// ## How to fix
+/// For general Python code, use tolerance-based comparison functions from the
+/// standard library:
+///
+/// - Use `math.isclose()` for scalar comparisons
+/// - Use `cmath.isclose()` for complex numbers comparisons
+/// - Use `unittest.assertAlmostEqual()` in tests with `unittest`
+///
+/// Note that `math.isclose()` / `cmath.isclose()` has a default absolute tolerance of zero, so
```
ntBre:
```suggestion
/// Note that `math.isclose()` and `cmath.isclose()` have a default absolute tolerance of zero, so
```

crates/ruff_linter/src/codes.rs
```diff
@@ -1063,6 +1063,8 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Ruff, "067") => rules::ruff::rules::NonEmptyInitModule,
         (Ruff, "068") => rules::ruff::rules::DuplicateEntryInDunderAll,
 
+        (Ruff, "069") => rules::ruff::rules::FloatEqualityComparison,
+
```
ntBre:
```suggestion
        (Ruff, "069") => rules::ruff::rules::FloatEqualityComparison,

```

Not sure if the GitHub suggestion is going to work here, but I don't think we need this blank line between 68 and 69.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,190 @@
...
+                            is_numeric_expr(left) || is_numeric_expr(right)
+                        }
+                        _ => has_float(left, semantic) || has_float(right, semantic),
+                    }
+                }
+                Expr::UnaryOp(ast::ExprUnaryOp { operand, .. }) => has_float(operand, semantic),
+                Expr::Named(ast::ExprNamed { value, .. }) => has_float(value, semantic),
+                Expr::If(ast::ExprIf { body, orelse, .. }) => {
+                    has_float(body, semantic) || has_float(orelse, semantic)
+                }
```
ntBre: Are you sure we need special handling here? It looks like `ResolvedPythonType` will already handle this.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,190 @@
...
+/// - [Python documentation: `unittest.assertAlmostEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual)
+/// - [NumPy documentation: `numpy.isclose`](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html#numpy-isclose)
+/// - [NumPy documentation: `numpy.allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html#numpy-allclose)
+/// - [PyTorch documentation: `torch.isclose`](https://docs.pytorch.org/docs/stable/generated/torch.isclose.html#torch-isclose)
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "NEXT_RUFF_VERSION")]
+pub(crate) struct FloatEqualityComparison<'a> {
+    pub left: &'a str,
+    pub right: &'a str,
+    pub operand: &'a str,
```
ntBre: A few nits here. I don't think these need to be `pub` (or even `pub(crate)`?), and I think `operand` is actually the operator in the expression.


```suggestion
    left: &'a str,
    right: &'a str,
    operator: &'a str,
```

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
chirizxc: изображение
I think so

ntBre: Ah okay, I guess the `has_float` recursion is important for all three of these. Thanks for checking!

chirizxc: > Ah okay, I guess the `has_float` recursion is important for all three of these. Thanks for checking!

It seems that ultimately only `Expr::Named(ast::ExprNamed { value, .. }) => has_float(value, semantic),` is needed.

chirizxc:
> Thanks! I found a couple more very small nits in the code, and I also found some interesting cases in the ecosystem report:
> 
>     * [tests/integration/tools/test_box_zoom_tool.py:106:16:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/tests/integration/tools/test_box_zoom_tool.py#L106) RUF069 Unreliable floating point equality comparison `(results['xrstart'] + results['xrend'])/2.0 == pytest.approx(0.25)`
> 
>     * [tests/integration/tools/test_box_zoom_tool.py:107:16:](https://github.com/bokeh/bokeh/blob/829b2a75c402d0d0abd7e37ff201fbdfd949d857/tests/integration/tools/test_box_zoom_tool.py#L107) RUF069 Unreliable floating point equality comparison `(results['yrstart'] + results['yrend'])/2.0 == pytest.approx(0.75)`
> 
> 
> I think these are technically false positives because [`pytest.approx`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-approx) is another way to solve the approximate equality issue.
> 
>     * [airflow-core/tests/unit/models/test_pool.py:205:16:](https://github.com/apache/airflow/blob/0cf89fa362e72898b718fe6ff2fc8caa0b4ba45f/airflow-core/tests/unit/models/test_pool.py#L205) RUF069 Unreliable floating point equality comparison `float("inf") == pool.open_slots()`
> 
> 
> If I understand correctly, this is actually a safe comparison because two infinities can be equal, unlike NaNs.
> 
> I think we should try to suppress these two cases before landing.

It seems that everything should be fine now.

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -188,3 +191,32 @@ fn is_numeric_expr(expr: &Expr) -> bool {
...
+fn should_skip_comparison(expr: &Expr, semantic: &SemanticModel) -> bool {
+    match expr {
+        Expr::Call(ast::ExprCall {
+            func, arguments, ..
+        }) => {
+            // Skip `pytest.approx` or `math.inf`
+            if let Some(qualified_name) = semantic.resolve_qualified_name(func) {
+                if matches!(
+                    qualified_name.segments(),
+                    ["pytest", "approx"] | ["math", "inf"]
```
ntBre: I think `math.inf` is a constant, not a call, so I don't think this branch will ever be entered, unless I'm missing something:

```pycon
>>> import math
>>> math.inf
inf
>>> math.inf()
    math.inf()
    ~~~~~~~~^^
TypeError: 'float' object is not callable
```

I think the new test case with `math.inf` only works because the other side is a `float()` call. We may need to consider both `Expr::Name` and `Expr::Attribute` if we want to handle `math.inf` and `from math import inf; inf`.

But nice work finding `math.inf`, I was only thinking about `float("inf")` and `pytest.approx`!

crates/ruff_linter/src/rules/ruff/rules/float_equality_comparison.rs
```diff
@@ -0,0 +1,222 @@
...
+                ) {
+                    return true;
+                }
+            }
+
+            // Skip `float("inf")`, `float("-inf")`, `float("infinity")`
+            if semantic.match_builtin_expr(func, "float") {
```
dscorbett: This applies to 1-argument `complex` calls like `complex("inf")` too.

ntBre (approved):
Thank you! I pushed two commits slightly simplifying the `math.inf` checks (`resolve_qualified_name` already handles both names and attributes) and adding a test case for `complex("inf")`. I think this is ready to land!

Issue #20570, opened 2025-09-25
Query regarding try-consider-else documentation


LilMonk:
### Question

In the documentation https://docs.astral.sh/ruff/rules/try-consider-else/ it says to use the pattern
```python
import logging

def reciprocal(n):
    try:
        rec = 1 / n
    except ZeroDivisionError:
        logging.exception("Exception occurred")
    else:
        print(f"reciprocal of {n} is {rec}")
        return rec
```
when I use it in my code
```python
import logging

class Bar(BaseModel):
    a: int
    b: str

def create_bar(a: int, b: str) -> Bar:
    if a < 0:
        raise ValueError("a must be non-negative")
    return Bar(a=a, b=b)

def foo() -> Bar:
    log = logging.getLogger("foo")
    log.info("Creating Bar instance")
    try:
        bar = create_bar(1, "test")
    except Exception:
        log.exception("Failed to create Bar instance")
    else:
        return bar
```

But on type checkers like pylance, it gives an error
```
Function with declared return type "Bar" must return a value on all code paths
  "None" is not assignable to "Bar"
```
It is only resolved if I have a `raise` in an except block like this
```python
import logging

class Bar(BaseModel):
    a: int
    b: str

def create_bar(a: int, b: str) -> Bar:
    if a < 0:
        raise ValueError("a must be non-negative")
    return Bar(a=a, b=b)

def foo() -> Bar:
    log = logging.getLogger("foo")
    log.info("Creating Bar instance")
    try:
        bar = create_bar(1, "test")
    except Exception:
        log.exception("Failed to create Bar instance")
        raise
    else:
        return bar
```

Is my understanding of the documentation wrong, or is there anything else that I'm missing?

### Version

ruff 0.12.12

ntBre:
I think the docs could probably be made more clear by returning a default value after logging the exception:

```py
import logging

def reciprocal(n):
    try:
        rec = 1 / n
    except ZeroDivisionError:
        logging.exception("Exception occurred")
        return 0
    else:
        print(f"reciprocal of {n} is {rec}")
        return rec
```

or re-raising as you mentioned. I think code like this should also run afoul of type checkers for the same reason, though:

```py
# try.py
import logging

class Bar: ...

def create_bar(a, b) -> Bar: Bar()

def foo() -> Bar:
    log = logging.getLogger("foo")
    log.info("Creating Bar instance")
    try:
        bar = create_bar(1, "test")
        return bar
    except Exception:
[...]

PR #20587, opened 2025-09-26
Fixed documentation for try_consider_else

LilMonk:
## Summary

This PR addresses #20570 . In the example, the correct usage had a bug/issue where in the except block after logging exception, None was getting returned, which made the linters flag out the code. So adding an empty raise solves the issue.

## Test Plan

Tested it by building the doc locally.

### DIFF
--- a/crates/ruff_linter/src/rules/tryceratops/rules/try_consider_else.rs
+++ b/crates/ruff_linter/src/rules/tryceratops/rules/try_consider_else.rs
@@ -29,6 +29,7 @@ use crate::checkers::ast::Checker;
 ///         return rec
 ///     except ZeroDivisionError:
 ///         logging.exception("Exception occurred")
+///         raise
 /// ```
 ///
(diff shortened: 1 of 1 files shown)


crates/ruff_linter/src/rules/tryceratops/rules/try_consider_else.rs
```diff
@@ -41,6 +41,7 @@ use crate::checkers::ast::Checker;
 ///         rec = 1 / n
 ///     except ZeroDivisionError:
 ///         logging.exception("Exception occurred")
+///         raise
```
ntBre, line 45: I think we should add this to the example before the fix too.

LilMonk: Done.

ntBre (approved):
Thank you!

PR #18221, opened 2025-05-20
Rule/large number without separators

guillp:
## Summary

This adds a rule (with code RUF061) that automatically formats large numbers with underscore separators to make them more readable. This is as described in [PEP515](https://peps.python.org/pep-0515/), and discussed in #12787 which I opened a while back.

This rule works for:
- integers: if they are 5 digits or more, the rule will require underscores as thousands, millions, billions, etc. separators. For relatively small integers (`9999` or less) the rule does not apply since it is already readable enough.
- floats: same rules as integers, and the float part is also formatted with those same rules, but the groups are formed left to right (thousandths, millionths, etc.)
- hexadecimal notation (`0xABCD`): add underscores to form groups of 4 digits
- octal notation (`0o1234`): add underscores to form groups of 4 digits
- binary notation (`0b1010101`): add underscores to from groups of 8 bits (octets)
- scientific notation (`123e10`): the leading part is formatted with the same rules as integers. The exponent part is untouched as it should never be more than 3 chars anyway.
- positive or negative literals: the leading `+` or `-` sign is not part of `Expr::NumberLiteral` instances once parsed by ruff, so this rule does not modify them in any way, they just stay in place.

## Test Plan

A new test file `RUF061.py` is part of the PR and is executed on `cargo test`.

## TODO / to discuss
- should the size of digit groups be configurable, instead of hardcoded? Same for minimal literal size to trigger the rule. While thousands/millions/billions/etc. make sense, there may be come cultural variety around this. Groups bits 8 by 8 should also be okay in most cases. Hexadecimals and octals digits grouped 4 by 4 is the most arbitrary, probably others will have different preferences.
- idea for another rule: should the literal prefixes (`0x`/`0X`, `0o`/`0O`, `0b`/`0B`) and `e`/`E` for the scientific notation be lowercased for better readability?

Issue #14587, opened 2024-11-25
FURB157 has false negatives for non-finite float strings


dscorbett:
[`verbose-decimal-constructor` (FURB157)](https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/) has false negatives in Ruff 0.8.0 for some non-finite float strings. It recognizes the following strings case-insensitively in expressions like `Decimal(float("inf"))`, but no others:

https://github.com/astral-sh/ruff/blob/a90e404c3f010446ab8c18b4793c78834eeb65b7/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs#L157

`float` strips white space. For example, `Decimal(float(" nan "))` should be simplified to `Decimal(" nan ")`. It could be further normalized to `Decimal("NaN")`, but FURB157 does not currently normalize other float strings, so I think it should keep the string argument unchanged.

`float` allows both signs for all non-finite values. `Decimal(float("+inf"))`, `Decimal(float("+infinity"))`, `Decimal(float("+nan"))`, and `Decimal(float("-nan"))` should be simplified. FURB157 should not normalize these string arguments either, except for the last one. `Decimal(float("-nan"))` is equivalent to `Decimal("nan")`, not `Decimal("-nan")`, so for an argument of `"-nan"`, the sign character should be deleted.

dylwil3:
Decimals strike again! Thanks for finding this, we'll get it fixed 😄

Issue #14204, opened 2024-11-08
FURB157 has incorrect fixes when replacing strings with integers


dscorbett:
The fixes for [`verbose-decimal-constructor` (FURB157)](https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/) are sometimes incorrect in Ruff 0.7.3. Two of them change behavior and one introduces a syntax error.

First, negative zero should be kept as a string. `int` doesn’t have signed zeroes but `Decimal` does.
```console
$ cat furb157.py
from decimal import Decimal
print(Decimal("-0"))

$ python furb157.py
-0

$ ruff check --isolated --preview --select FURB157 furb157.py --fix
Found 1 error (1 fixed, 0 remaining).

$ cat furb157.py
from decimal import Decimal
print(Decimal(-0))

$ python furb157.py
0
```

Second, the empty string should not be normalized to zero. The empty string signals an `InvalidOperation`, which can be trapped and converted to NaN or can cause an error. “Empty” here refers to the string after normalizing away white space and underscores.
```console
$ cat furb157.py
from decimal import Decimal, ExtendedContext
print(Decimal("_", ExtendedContext))
print(Decimal(" "))

$ python furb157.py
NaN
Traceback (most recent call last):
  File "furb157.py", line 3, in <module>
    print(Decimal(" "))
          ^^^^^^^^^^^
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]

$ ruff check --isolated --preview --select FURB157 furb157.py --fix
Found 2 errors (2 fixed, 0 remaining).

$ cat furb157.py
cat furb157.py
from decimal import Decimal, ExtendedContext
print(Decimal(0, ExtendedContext))
print(Decimal(0))

$ python furb157.py
0
0
```

Third, decimal `int` literals with too many digits cause syntax errors. [The minimum maximum digit count](https://docs.python.org/3.13/library/sys.html#sys.int_info.str_digits_check_threshold) is 640. If the integer-valued string argument to `Decimal` has over that many digits, it should stay a string.
```console
$ { printf 'from decimal import Decimal\nDecimal("1'; printf 0%0.s {1..640}; printf '")\n'; } >furb157.py

$ PYTHONINTMAXSTRDIGITS=640 python furb157.py

$ ruff check --isolated --preview --select FURB157 furb157.py --fix
Found 1 error (1 fixed, 0 remaining).

[...]

Issue #13807, opened 2024-10-18
[verbose-decimal-constructor (FURB157)] doesn't trigger on `Decimal("1_000")`


monosans:
```python
Decimal(1000)  # OK
Decimal(1_000)  # OK
Decimal("1000")  # Triggers FURB157
Decimal("1_000")  # Should trigger FURB157 and become Decimal(1_000), but it doesn't
```

ruff v0.7.0

AlexWaygood:
It seems like the parsing we do here is too naive: https://github.com/astral-sh/ruff/blob/4ecfe95295eb203ba771f344f52c8e61527a4c05/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs#L77-L98

AlexWaygood:
If there are more edge cases that we don't handle, it might be worth just using `ruff_python_parser::parse_expression` rather than reimplementing all the edge cases in the linter. If this is the only edge case we don't currently handle, however, it may be best to avoid doing that and just reimplement the logic in the linter -- using the parser might be quite expensive.

dscorbett:
`Decimal` accepts more integer-valued strings than are valid as `int` literals. For example, `Decimal("1__2")` and `Decimal("_1_")` and `Decimal("١٢٣")` are valid. Should FURB157 trigger on those? If so, `ruff_python_parser::parse_expression` would not be enough.

dylwil3:
Looks like this is the official regex: 

https://github.com/python/cpython/blob/322f14eeff9e3b5853eaac3233f7580ca0214cf8/Lib/_pydecimal.py#L6059-L6077

I can try to take this sometime this weekend (but if I don't, others should feel free to go for it!)

m-aciek:
The fix in latest ruff (v0.13.1) converts `Decimal("1_000")` to `Decimal(1000)` instead of to `Decimal(1_000)`.

It becomes more problematic for bigger values, where thousand formatting really helps to read a number, like `Decimal(15_000_000)`.

Can the issue be reopened?

PR #20588, opened 2025-09-26
[`refurb`] Preserve digit separators in `Decimal` constructor (`FURB157`)

### LINKED ISSUE #20572: Fix for `FURB157` removes digit separators
The fix in latest ruff (v0.13.1) converts `Decimal("1_000")` to `Decimal(1000)` instead of to `Decimal(1_000)`.

It becomes more problematic for bigger values, where thousand formatting really helps to read a number, like `Decimal(15_000_000)`.

 _Originally posted by @m-aciek in [#13807](https://github.com/astral-sh/ruff/issues/13807#issuecomment-3332811804)_

danparizher:
## Summary

Fixes #20572

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/refurb/FURB157.py
+++ b/crates/ruff_linter/resources/test/fixtures/refurb/FURB157.py
@@ -69,3 +69,10 @@ Decimal(float("\N{space}\N{hyPHen-MINus}nan"))
 Decimal(float("\x20\N{character tabulation}\N{hyphen-minus}nan"))
 Decimal(float("   -" "nan"))
 Decimal(float("-nAn"))
+
+# Additional test cases for digit separator preservation
+# https://github.com/astral-sh/ruff/issues/20572
+Decimal("15_000_000")  # Should preserve digit separators
+Decimal("1_234_567")   # Should preserve digit separators
+Decimal("-5_000")      # Should preserve negative with separators
+Decimal("+9_999")      # Should preserve positive with separators
--- a/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
@@ -90,8 +90,10 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
             // using this regex:
             // https://github.com/python/cpython/blob/ac556a2ad1213b8bb81372fe6fb762f5fcb076de/Lib/_pydecimal.py#L6060-L6077
             // _after_ trimming whitespace from the string and removing all occurrences of "_".
-            let mut trimmed = Cow::from(str_literal.to_str().trim_whitespace());
-            if memchr::memchr(b'_', trimmed.as_bytes()).is_some() {
+            let original_str = str_literal.to_str().trim_whitespace();
+            let mut trimmed = Cow::from(original_str);
+            let has_digit_separators = memchr::memchr(b'_', trimmed.as_bytes()).is_some();
+            if has_digit_separators {
                 trimmed = Cow::from(trimmed.replace('_', ""));
             }
             // Extract the unary sign, if any.
@@ -128,7 +130,14 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
                 _ => rest,
             };
 
-            let replacement = format!("{unary}{rest}");
+            // If the original string had digit separators, preserve them in the integer literal
+            let replacement = if has_digit_separators {
+                // Convert to integer literal but preserve the original digit separators
+                original_str.to_string()
+            } else {
+                format!("{unary}{rest}")
+            };
+
             let mut diagnostic = checker.report_diagnostic(
                 VerboseDecimalConstructor {
                     replacement: replacement.clone(),
--- a/crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB157_FURB157.py.snap
+++ b/crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB157_FURB157.py.snap
@@ -167,12 +167,12 @@ FURB157 [*] Verbose expression in `Decimal` constructor
    |         ^^^^^^^
 24 | Decimal("__1____000") 
    |
-help: Replace with `1000`
+help: Replace with `1_000`
 20 | # See https://github.com/astral-sh/ruff/issues/13807
 21 | 
 22 | # Errors
    - Decimal("1_000")
-23 + Decimal(1000)
+23 + Decimal(1_000)
 24 | Decimal("__1____000") 
 25 | 
 26 | # Ok
@@ -187,12 +187,12 @@ FURB157 [*] Verbose expression in `Decimal` constructor
 25 |
 26 | # Ok
    |
-help: Replace with `1000`
+help: Replace with `__1____000`
 21 | 
 22 | # Errors
 23 | Decimal("1_000")
    - Decimal("__1____000") 
-24 + Decimal(1000) 
+24 + Decimal(__1____000) 
 25 | 
 26 | # Ok
 27 | Decimal("2e-4")
@@ -494,6 +494,7 @@ help: Replace with `"nan"`
 69 + Decimal("nan")
 70 | Decimal(float("   -" "nan"))
 71 | Decimal(float("-nAn"))
+72 | 
 
 FURB157 [*] Verbose expression in `Decimal` constructor
   --> FURB157.py:70:9
@@ -511,6 +512,8 @@ help: Replace with `"nan"`
    - Decimal(float("   -" "nan"))
 70 + Decimal("nan")
 71 | Decimal(float("-nAn"))
+72 | 
+73 | # Additional test cases for digit separator preservation
 
 FURB157 [*] Verbose expression in `Decimal` constructor
   --> FURB157.py:71:9
@@ -519,6 +522,8 @@ FURB157 [*] Verbose expression in `Decimal` constructor
 70 | Decimal(float("   -" "nan"))
 71 | Decimal(float("-nAn"))
    |         ^^^^^^^^^^^^^
+72 |
+73 | # Additional test cases for digit separator preservation
    |
 help: Replace with `"nan"`
 68 | Decimal(float("\N{space}\N{hyPHen-MINus}nan"))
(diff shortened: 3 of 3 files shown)


crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -128,7 +130,14 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
             };
 
-            let replacement = format!("{unary}{rest}");
+            // If the original string had digit separators, preserve them in the integer literal
+            let replacement = if has_digit_separators {
+                // Convert to integer literal but preserve the original digit separators
+                original_str.to_string()
+            } else {
+                format!("{unary}{rest}")
+            };
```
ntBre: I looked through some of the previous issues with this rule, and this check is not sufficient: 
- https://github.com/astral-sh/ruff/issues/13807
- https://github.com/astral-sh/ruff/issues/14204
- https://github.com/astral-sh/ruff/issues/14587

The first of these, especially [this comment](https://github.com/astral-sh/ruff/issues/13807#issuecomment-2422680644), notes a couple of cases where `has_digit_separators` would lead to an invalid `original_str`. This replacement in the snapshots below is an example of that: `Decimal(__1____000) `:

```pycon
>>> from decimal import Decimal
>>> Decimal(__1____000)
  File "<python-input-5>", line 1, in <module>
    Decimal(__1____000)
            ^^^^^^^^^^
NameError: name '__1____000' is not defined
>>> Decimal("__1____000")
Decimal('1000')
```

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -130,10 +130,12 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
 
-            // If the original string had digit separators, preserve them in the integer literal
+            // If the original string had digit separators, add proper thousand separators
             let replacement = if has_digit_separators {
-                // Convert to integer literal but preserve the original digit separators
-                original_str.to_string()
+                // Convert to integer literal with proper thousand separators
+                // Reconstruct valid separators (not necessarily the original ones)
+                let numeric_value = format!("{unary}{rest}");
+                add_thousand_separators(&numeric_value)
```
amyreese: Would it be better to just strip leading underscores, and otherwise preserve the placement and frequency, rather than assume that thousands are the only use of underscores in numeric literals?

ntBre: Leading underscores aren't the only problem. Duplicates in the middle cause a syntax error too, but not in the `Decimal` constructor:

```pycon
>>> from decimal import Decimal
>>> 1__2
  File "<python-input-2>", line 1
    1__2
     ^
SyntaxError: invalid decimal literal
>>> Decimal("1__2")
Decimal('12')
```

But I agree in principle, `add_thousand_separators` does not seem like the right approach to me.

I'm starting to think we should just mark the fix as unsafe if the value we're replacing the original with is different. The edge cases here seem pretty tricky.

danparizher: Thanks for catching that; I agree that marking it unsafe is probably the way to go to not break too much.

m-aciek: Wouldn't stripping leading underscores and replacing duplicates with a single underscore be enough?

In my opinion keeping `Decimal("10_000_000")` -> `Decimal(10_000_000)` transition as safe would be very convenient.

dscorbett: It would be enough to strip leading underscores, strip trailing underscores, strip leading zeros (unless all the digits are zeros), and collapse medial underscore sequences: `Decimal(" _-_0_01__2_ ")` would become `Decimal(-1_2)`.

m-aciek: That's the behavior I fight for ❤️ 🚀 :)

ntBre: That sounds reasonable to me, I gave up a bit too soon :) @danparizher do you want to give this a shot? We can still revert to marking the fix unsafe if the code gets too complicated, but this sounds feasible with the transformations we're already doing.

amyreese (approved):
lgtm

dscorbett:
The current state of this PR is that underscores are normalized to be thousands separators, but cf. @amyreese’s [earlier comment](https://github.com/astral-sh/ruff/pull/20588#discussion_r2393221709). If a user writes `Decimal("1__23_45_000")` that formatting was probably intentional so it should become `Decimal(1_23_45_000)`. I think normalizing the underscores more than the minimum needed to avoid a syntax error should be left for a different rule as in #18221.

MichaReiser:
What's the status of this PR? Is there something left that needs addressing or is it good to go?

ntBre:
I don't think we should normalize the digit separators to thousands separators, as @dscorbett pointed out.

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -185,6 +202,63 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
+            // and we haven't already added an underscore
+            if !result.is_empty() && !result.ends_with('_') {
+                result.push('_');
+            }
+        }
+        i += 1;
+    }
+
+    // Remove trailing underscores
+    result = result.trim_end_matches('_').to_string();
```
amyreese: I think this ends up being overly verbose and defensive, and lines 225-251 could be replaced with just:

```
let result = trimmed
    .chars()
    .dedup_by(|a, b| {*a == '_' && a == b})
    .collect::<String>();
```

Leading/trailing underscores were already trimmed by line 218, so there shouldn't be a need to both check that there are some number of digits *and* the result after trimming/deduplication is not empty.

crates/ruff_linter/resources/test/fixtures/refurb/FURB157.py
```diff
@@ -69,3 +69,10 @@
 Decimal(float("\x20\N{character tabulation}\N{hyphen-minus}nan"))
 Decimal(float("   -" "nan"))
 Decimal(float("-nAn"))
+
+# Test cases for digit separators (safe fixes)
+# https://github.com/astral-sh/ruff/issues/20572
+Decimal("15_000_000")  # Safe fix: normalizes separators, becomes Decimal(15_000_000)
+Decimal("1_234_567")   # Safe fix: normalizes separators, becomes Decimal(1_234_567)
+Decimal("-5_000")      # Safe fix: normalizes separators, becomes Decimal(-5_000)
+Decimal("+9_999")      # Safe fix: normalizes separators, becomes Decimal(+9_999)
```
amyreese, line 78: Would be nice to include test cases showing that non-thousands separators are maintained accordingly, eg `"12_34_56_78"` and `"1234_5678"`.

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -128,18 +130,33 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
+                diagnostic.set_fix(Fix::unsafe_edit(Edit::range_replacement(
+                    replacement,
+                    value.range(),
+                )));
+            } else {
+                diagnostic.set_fix(Fix::safe_edit(Edit::range_replacement(
+                    replacement,
+                    value.range(),
+                )));
+            }
```
ntBre: We should use `Fix::applicable_edit` here.

ntBre: Or wait, if we're going to preserve the original formatting, should it just be safe again?

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -185,6 +202,63 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
     }
 }
 
+/// Normalizes digit separators in a numeric string by:
+/// - Stripping leading and trailing underscores
+/// - Collapsing medial underscore sequences to single underscores
+/// - Do not force thousands separators
+fn normalize_digit_separators(original_str: &str, unary: &str, _numeric_part: &str) -> String {
```
ntBre: We should either remove `_numeric_part` if it's unused or rename it to `numeric_part`, if it's used.

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -185,6 +202,63 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
+        return format!("{unary}0");
+    }
+
+    // Extract only digits from the trimmed string
+    let digits: String = trimmed.chars().filter(char::is_ascii_digit).collect();
+
+    // If no digits found, return 0
+    if digits.is_empty() {
+        return format!("{unary}0");
+    }
```
ntBre: These `is_empty` checks don't seem right to me at all. I don't think we should return an arbitrary value for invalid input, and I think we've already validated the input in `verbose_decimal_constructor`, but I could also be missing something.

crates/ruff_linter/src/rules/refurb/rules/verbose_decimal_constructor.rs
```diff
@@ -185,6 +202,63 @@ pub(crate) fn verbose_decimal_constructor(checker: &Checker, call: &ast::ExprCal
...
+/// - Stripping leading and trailing underscores
+/// - Collapsing medial underscore sequences to single underscores
+/// - Do not force thousands separators
+fn normalize_digit_separators(original_str: &str, unary: &str, _numeric_part: &str) -> String {
+    // Remove the unary sign from the original string to work with the numeric part
+    let without_unary = if original_str.starts_with('+') || original_str.starts_with('-') {
+        &original_str[1..]
+    } else {
+        original_str
+    };
```
ntBre: We already stripped the prefix once on line 99. Can we reuse more of the work from the parent function?

ntBre:
After a bit more thought, I came up with some additional edge cases with separators and leading zeros. Hopefully everything is covered now.

PR #20591, opened 2025-09-26
[ty] Use `Top` materializations for `TypeIs` special form

AlexWaygood:
## Summary

Typeshed has many functions that return `TypeIs[SomeCovariantGeneric[Any]]`. This is somewhat unfortunate for ty, because for our purposes it's almost always better to intersect with `CovariantGeneric[object]` for the purposes of type narrowing, but it's [somewhat hard](https://github.com/python/typeshed/pull/14784) to change these in typeshed without creating false-positive diagnostics for users of other type checkers. This PR therefore introduces a workaround for the typeshed issue: under the hood, we convert the type argument passed to `TypeIs` to its top materialization before intersecting with that type.

## Test Plan

I added an mdtest that fails on `main` without this change. It's similar to the false-positive diagnostics relating to the use of [`asyncio.iscoroutine`](https://github.com/python/typeshed/blob/6af8f287bb3eae6c6d8839c69c39f07ee7057517/stdlib/asyncio/coroutines.pyi#L28) that we see going away on `homeassistant/core` in the mypy_primer report.

Some more ecosystem analysis is found in https://github.com/astral-sh/ruff/pull/20591#issuecomment-3338847554.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -1286,7 +1286,11 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
 
                     Type::unknown()
                 }
-                _ => TypeIsType::unbound(self.db(), self.infer_type_expression(arguments_slice)),
+                _ => TypeIsType::unbound(
+                    self.db(),
+                    self.infer_type_expression(arguments_slice)
+                        .top_materialization(self.db()),
+                ),
             },
             SpecialFormType::TypeGuard => {
                 self.infer_type_expression(arguments_slice);
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -1288,6 +1288,12 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                 }
                 _ => TypeIsType::unbound(
                     self.db(),
+                    // N.B. Using the top materialization here is a pragmatic decision
+                    // that makes us produce more intuitive results given how
+                    // `TypeIs` is used in the real world (in particular, in typeshed).
+                    // However, there's some debate about whether this is really
+                    // fully correct. See <https://github.com/astral-sh/ruff/pull/20591>
+                    // for more discussion.
                     self.infer_type_expression(arguments_slice)
                         .top_materialization(self.db()),
                 ),
--- a/crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
@@ -173,6 +173,11 @@ def _(d: Any):
 
 ## Narrowing
 
+```toml
+[environment]
+python-version = "3.12"
+```
+
 ```py
 from typing import Any
 from typing_extensions import TypeGuard, TypeIs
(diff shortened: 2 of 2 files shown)


AlexWaygood:
# Ecosystem analysis

Overall this PR adds about as many diagnostics as it removes, but I think it makes our narrowing behaviour for `TypeIs` more consistent with our narrowing behaviour elsewhere, and therefore easier to reason about.

There are several hits that look like true positives: we're now inferring less dynamic types which is allowing us to highlight places where the code arguably is unsound (`aiohttp`, `beartype`, `prefect`, `psycopg`, `trio`). I worry that some of these will seem pedantic, overly strict, and possibly confusing to users. But it makes our `TypeIs` narrowing behaviour more consistent with our `isinstance()` narrowing behaviour, which will be especially beneficial when https://github.com/astral-sh/ruff/pull/20368 lands.

The new diagnostic on `hydra-zen` is caused by missing support for `TypeGuard`.

We have some new false positives in `pytest`, `vision` and `xarray` that are caused by this TODO here, which causes us to incorrectly think that `Top[Callable[..., Any]]` -> `Callable[[], object]` (it should really be `Callable[Top[...], object]`, I suppose?

https://github.com/astral-sh/ruff/blob/57e1ff8294d630b0f265d875caad80e32c8ee6b6/crates/ty_python_semantic/src/types/signatures.rs#L1286-L1291

We have a new false positive on `starlette` which is caused by missing support for `functools.partial`.

carljm:
> it should really be `Callable[Top[...], object]`, I suppose?

No, it should be `Callable[Bottom[...], object]` due to contravariance of arguments.

AlexWaygood:
> > it should really be `Callable[Top[...], object]`, I suppose?
> 
> No, it should be `Callable[Bottom[...], object]` due to contravariance of arguments.

Right, yes, sorry

carljm (commented):
I don't love implicitly transforming a type explicitly written by the user :/ but I do feel like this practically improves the results, given the way `TypeIs` is currently used in typeshed. Narrowing with a gradual type is just not really very useful, if it's handled precisely (as we do).

I feel like the user intent expressed by `-> TypeIs[Foo[Any]]` is "the type is some kind of `Foo`, I don't know which kind" (this is accurately modeled by taking the top materialization), but also then "treat the contained type of `Foo` as dynamic if I end up using it" -- and this latter part is what we don't do.

I wonder what it would look like if, instead of doing this, we simplified `Foo[Any] & Foo[Bar]` to `Foo[Bar & Any]`? I think that's a correct simplification? And I think that would give the same benefits as this PR, but with fewer false positives -- it would better respect the user's intent in writing `TypeIs[Foo[Any]]`. WDYT?

AlexWaygood:
> I wonder what it would look like if, instead of doing this, we simplified `Foo[Any] & Foo[Bar]` to `Foo[Bar & Any]`? I think that's a correct simplification?

That's only a correct simplification if `Foo` is covariant, I think? I don't think it's true, for example, that `list[Any] & list[int]` can be simplified to `list[int & Any]`

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -1286,7 +1286,11 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
 
                     Type::unknown()
                 }
-                _ => TypeIsType::unbound(self.db(), self.infer_type_expression(arguments_slice)),
+                _ => TypeIsType::unbound(
+                    self.db(),
+                    self.infer_type_expression(arguments_slice)
```
AlexWaygood, line 1297:
```suggestion
                    // N.B. Using the top materialization here is a pragmatic decision
                    // that makes us produce more intuitive results given how
                    // `TypeIs` is used in the real world (in particular, in typeshed).
                    // However, there's some debate about whether this is really
                    // fully correct. See <https://github.com/astral-sh/ruff/pull/20591>
                    // for more discussion.
                    self.infer_type_expression(arguments_slice)
```

PR #20593, opened 2025-09-26
[ty] Simplify `Any | (Any & T)` to `Any`

AlexWaygood:
I think this may help reduce some of the fallout from https://github.com/astral-sh/ruff/pull/20368 by creating simpler types when we narrow unions

## Test plan

- Added mdtests
- Ran `QUICKCHECK_TESTS=1000000 cargo test --release -p ty_python_semantic -- --ignored types::property_tests::stable`

### DIFF
--- a/crates/ty_python_semantic/src/types/builder.rs
+++ b/crates/ty_python_semantic/src/types/builder.rs
@@ -504,9 +504,16 @@ impl<'db> UnionBuilder<'db> {
             if should_simplify_full && !matches!(element_type, Type::TypeAlias(_)) {
                 if ty.is_equivalent_to(self.db, element_type)
                     || ty.is_subtype_of(self.db, element_type)
+                    || ty.into_intersection().is_some_and(|intersection| {
+                        intersection.positive(self.db).contains(&element_type)
+                    })
                 {
                     return;
-                } else if element_type.is_subtype_of(self.db, ty) {
+                } else if element_type.is_subtype_of(self.db, ty)
+                    || element_type
+                        .into_intersection()
+                        .is_some_and(|intersection| intersection.positive(self.db).contains(&ty))
+                {
                     to_remove.push(index);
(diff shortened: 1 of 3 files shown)


AlexWaygood:
All the new ecosystem diagnostics look like new manifestations of https://github.com/astral-sh/ty/issues/1248. I'm not sure why we weren't emitting them before (maybe some `Todo`-type branch we have for intersections somewhere?), but they're what I would expect given our current behaviour for unannotated `dict` literals and unannotated `list` literals.

Overall I think this makes us produce simpler, more user-friendly types.

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -504,9 +504,16 @@ impl<'db> UnionBuilder<'db> {
             if should_simplify_full && !matches!(element_type, Type::TypeAlias(_)) {
                 if ty.is_equivalent_to(self.db, element_type)
                     || ty.is_subtype_of(self.db, element_type)
+                    || ty.into_intersection().is_some_and(|intersection| {
+                        intersection.positive(self.db).contains(&element_type)
+                    })
```
AlexWaygood, line 509: A more generalised solution here (that would also allow us to get rid of the separate `is_equivalent_to` call -- so might be quite a bit faster?) would be to add a third type relation, `UnionSimplification` (not sure what the best name would be really), which would sit in between `TypeRelation::Subtyping` and `TypeRelation::Assignability`. It would mostly behave the same way as subtyping (`Any | int` would not be considered simplifiable to `Any`), but `Any <: Any`, `(Any | int) <: (int | Any)` and `(Any & str) <: Any` would all be considered `true` under this type relation.

carljm:
> maybe some `Todo`-type branch we have for intersections somewhere?

We still emit Todo for mapping over an intersection with negated elements, could be that?

PR #20596, opened 2025-09-26
[ty] Apply type mappings to functions eagerly

dcreager:
`TypeMapping` is no longer cow-shaped.

Before, `TypeMapping` defined a `to_owned` method, which would make an owned copy of the type mapping. This let us apply type mappings to function literals lazily. The primary part of a function that you have to apply the type mapping to is its signature. The hypothesis was that doing this lazily would prevent us from constructing the signature of a function just to apply a type mapping; if you never ended up needed the updated function signature, that would be extraneous work.

But looking at the CI for this PR, it looks like that hypothesis is wrong! And this definitely cleans up the code quite a bit. It also means that over time we can consider replacing all of these `TypeMapping` enum variants with separate `TypeTransformer` impls.

### DIFF
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -671,10 +671,20 @@ impl<'db> FunctionLiteral<'db> {
 #[derive(PartialOrd, Ord)]
 pub struct FunctionType<'db> {
     pub(crate) literal: FunctionLiteral<'db>,
+
+    /// Contains a potentially modified signature for this function literal, in case certain operations
+    /// (like type mappings) have been applied to it.
+    ///
+    /// See also: [`FunctionLiteral::updated_signature`].
     #[returns(as_ref)]
-    pub(crate) updated_signature: Option<CallableSignature<'db>>,
+    updated_signature: Option<CallableSignature<'db>>,
+
+    /// Contains a potentially modified signature for the last overload or the implementation of this
+    /// function literal, in case certain operations (like type mappings) have been applied to it.
+    ///
+    /// See also: [`FunctionLiteral::last_definition_signature`].
     #[returns(as_ref)]
-    pub(crate) updated_last_definition_signature: Option<Signature<'db>>,
+    updated_last_definition_signature: Option<Signature<'db>>,
 }
 
 // The Salsa heap is tracked separately.
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -77,11 +77,11 @@ use crate::types::narrow::ClassInfoConstraintFunction;
 use crate::types::signatures::{CallableSignature, Signature};
 use crate::types::visitor::any_over_type;
 use crate::types::{
-    BoundMethodType, BoundTypeVarInstance, CallableType, ClassBase, ClassLiteral, ClassType,
-    DeprecatedInstance, DynamicType, FindLegacyTypeVarsVisitor, HasRelationToVisitor,
-    IsEquivalentVisitor, KnownClass, KnownInstanceType, NormalizedVisitor, SpecialFormType,
-    TrackedConstraintSet, Truthiness, Type, TypeMapping, TypeRelation, UnionBuilder, all_members,
-    binding_type, todo_type, walk_type_mapping,
+    ApplyTypeMappingVisitor, BoundMethodType, BoundTypeVarInstance, CallableType, ClassBase,
+    ClassLiteral, ClassType, DeprecatedInstance, DynamicType, FindLegacyTypeVarsVisitor,
+    HasRelationToVisitor, IsEquivalentVisitor, KnownClass, KnownInstanceType, NormalizedVisitor,
+    SpecialFormType, TrackedConstraintSet, Truthiness, Type, TypeMapping, TypeRelation,
+    UnionBuilder, all_members, binding_type, todo_type,
 };
 use crate::{Db, FxOrderSet, ModuleName, resolve_module};
 
@@ -623,11 +623,7 @@ impl<'db> FunctionLiteral<'db> {
     /// calling query is not in the same file as this function is defined in, then this will create
     /// a cross-module dependency directly on the full AST which will lead to cache
     /// over-invalidation.
-    fn signature<'a>(
-        self,
-        db: &'db dyn Db,
-        type_mappings: &'a [TypeMapping<'a, 'db>],
-    ) -> CallableSignature<'db>
+    fn signature<'a>(self, db: &'db dyn Db) -> CallableSignature<'db>
     where
         'db: 'a,
     {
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/src/types/function.rs
```diff
@@ -690,12 +677,10 @@ impl<'db> FunctionLiteral<'db> {
 #[derive(PartialOrd, Ord)]
 pub struct FunctionType<'db> {
     pub(crate) literal: FunctionLiteral<'db>,
```
dcreager, line 673: We might also consider consolidating `FunctionType` and `FunctionLiteral` into a single type now. (`FunctionLiteral` would have the extra fields that hold the updated signatures.)

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -623,11 +623,7 @@ impl<'db> FunctionLiteral<'db> {
...
     /// over-invalidation.
-    fn signature<'a>(
-        self,
-        db: &'db dyn Db,
-        type_mappings: &'a [TypeMapping<'a, 'db>],
-    ) -> CallableSignature<'db>
+    fn signature<'a>(self, db: &'db dyn Db) -> CallableSignature<'db>
     where
         'db: 'a,
     {
```
AlexWaygood:
```suggestion
    fn signature(self, db: &'db dyn Db) -> CallableSignature<'db> {
```

(Surprised Clippy didn't complain about this!)

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -660,20 +654,13 @@ impl<'db> FunctionLiteral<'db> {
...
     /// over-invalidation.
-    fn last_definition_signature<'a>(
-        self,
-        db: &'db dyn Db,
-        type_mappings: &'a [TypeMapping<'a, 'db>],
-    ) -> Signature<'db>
+    fn last_definition_signature<'a>(self, db: &'db dyn Db) -> Signature<'db>
     where
         'db: 'a,
     {
```
AlexWaygood:
```suggestion
    fn last_definition_signature(self, db: &'db dyn Db) -> Signature<'db> {
```

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -707,9 +692,6 @@ pub(super) fn walk_function_type<'db, V: super::visitor::TypeVisitor<'db> + ?Siz
     visitor: &V,
 ) {
     walk_function_literal(db, function.literal(db), visitor);
-    for mapping in function.type_mappings(db) {
-        walk_type_mapping(db, mapping, visitor);
-    }
 }
```
AlexWaygood, line 707: do we not need to walk the new fields here (`updated_signature` and `updated_last_definition_signature`)? Would it be worth adding a comment about why that's not necessary?

AlexWaygood:
> ```diff
> psycopg (https://github.com/psycopg/psycopg)
> - psycopg_pool/psycopg_pool/_acompat.py:185:12: error[invalid-return-type] Return type does not match returned value: expected `T@ensure_async`, found `object`
> - Found 699 diagnostics
> + Found 698 diagnostics
> ```

Is this ecosystem change correct? If so, can we add an mdtest making sure it doesn't regress?

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -690,12 +677,10 @@ impl<'db> FunctionLiteral<'db> {
...
-
-    /// Type mappings that should be applied to the function's parameter and return types. This
-    /// might include specializations of enclosing generic contexts (e.g. for non-generic methods
-    /// of a specialized generic class).
-    #[returns(deref)]
-    type_mappings: Box<[TypeMapping<'db, 'db>]>,
+    #[returns(as_ref)]
+    pub(crate) updated_signature: Option<CallableSignature<'db>>,
+    #[returns(as_ref)]
+    pub(crate) updated_last_definition_signature: Option<Signature<'db>>,
```
carljm: Maybe some doc comments here explaining what these fields are and when/how they become populated?

carljm (approved):
This looks great! (Modulo existing inline comments.) Thank you!

sharkdp:
Glad I looked at my notifications this morning before continuing to work on https://github.com/astral-sh/ty/issues/1261! The lazy application of type mappings to functions (and the application of type mappings to those lazy type mappings) seemed to be the root cause of that issue, and applying type-mappings eagerly would have been one of the attempts to solve that bug. And indeed, the problem is completely solved with this PR!

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./ty_main check /home/shark/playground` | 11.185 ± 0.064 | 11.117 | 11.245 | 385.74 ± 24.39 |
| `./ty_20596 check /home/shark/playground` | 0.029 ± 0.002 | 0.027 | 0.031 | 1.00 |

crates/ty_python_semantic/src/types/function.rs
sharkdp: Took me quite a bit of time to construct an example where this would have any observable effects, since you first need to apply a type mapping to the function literal and then have a type-visitor walk over it:
```py
from typing import cast
from ty_extensions import TypeOf, Unknown


class C[T]:
    def f(self, t: T) -> T | Unknown:
        raise NotImplementedError


cast(TypeOf[C[int]().f], C[int]().f)
```
I'm first applying a specialization type mapping through `C[int]().f`, and then a type visitor via `any_over_type(..., contains_unknown_or_todo, ..)` by attempting a `cast`. Without walking over the `updated_*` fields, we get a
```
Value is already of type `bound method C[int].f(t: int) -> int | Unknown`
```
diagnostic (which shouldn't happen, if `Unknown` is part of the type). After applying the following patch, the diagnostic is gone.
```diff
@@ -692,6 +692,14 @@ pub(super) fn walk_function_type<'db, V: super::visitor::TypeVisitor<'db> + ?Siz
     visitor: &V,
 ) {
     walk_function_literal(db, function.literal(db), visitor);
+    if let Some(callable_signature) = function.updated_signature(db) {
+        for signature in &callable_signature.overloads {
+            walk_signature(db, signature, visitor);
+        }
+    }
+    if let Some(signature) = function.updated_last_definition_signature(db) {
+        walk_signature(db, signature, visitor);
+    }
 }
 ```

**Comment by sharkdp** [comment]
> > ```diff
> > psycopg (https://github.com/psycopg/psycopg)
> > - psycopg_pool/psycopg_pool/_acompat.py:185:12: error[invalid-return-type] Return type does not match returned value: expected `T@ensure_async`, found `object`
> > - Found 699 diagnostics
> > + Found 698 diagnostics
> > ```
> 
> Is this ecosystem change correct? If so, can we add an mdtest making sure it doesn't regress?

```py
from typing import Any, ParamSpec, TypeVar
from inspect import isawaitable
from collections.abc import Callable, Coroutine

T = TypeVar("T")
P = ParamSpec("P")

async def ensure_async(
    f: Callable[P, T] ,
    *args: P.args,
    **kwargs: P.kwargs,
) -> T:
    rv = f(*args, **kwargs)
    if isawaitable(rv):
        rv = await rv
    return rv
```

Yes, this is a false positive that shouldn't be there. I invested quite a bit of time trying to track it down, and while I haven't fully understood what is going on, it feels like it's not worth pursuing further (given that the behavior arguably improves). First of all, it took me a while to even reproduce it, because it only happens when checking on 3.12, not on 3.13 (which I use by default for testing locally). There are a lot of `@Todo` types involved, both because we create an intersection type in the `if isawaitable(rv)` narrowing, and because `ParamSpec` is involved (which we look for via `any_over_type`, which is affected by this PR). The correct return type from the `await rv` expression should be `T@ensure_async`, but for some reason we infer `object` on `main`. And we do infer `object` on this branch, when checking with 3.13. It might also have to do with normalization, because everything works fine when the seemingly irrelevant `Callable[P, T]` is removed from union in the `f` parameter annotation.

crates/ty_python_semantic/src/types/function.rs
sharkdp: I added some doc comments and made those fields non-public. @dcreager please feel free to change them in a follow-up, if something seems wrong/missing. I'd like to merge this PR to fix https://github.com/astral-sh/ty/issues/1261.

PR #10028, opened 2024-02-18
[`ruff`] Implement compound comparison rule (`RUF029`)

tjkuson:
## Summary

Implements `non-transitive-compound-comparison` (`RUF029`) that disallows certain compound comparisons.

As [suggested](https://github.com/astral-sh/ruff/issues/8964#issuecomment-1837194529),

> Arguably we should lint against all compound Compares _except_ the following:
> 
>     * Both compares are `is`: `a is b is c`
> 
>     * Both compares are `==`: `a == b ==c`
> 
>     * Both compares are one of `<`, `<=`: `a <= b < c` (probably the most common use  of compound Compare)
> 
>     * Both compares are one of `>`, `>=`: `a >= b > c`

Not a huge fan of the name I gave it, though. Chose `RUF029` as `RUF028` has been used in a bunch of open PRs.

Closes #8964. 

## Test Plan

`cargo nextest run`

PR #20597, opened 2025-09-26
[`pylint`] Implement `bad-chained-comparison` (`PLW3601`)

### LINKED ISSUE #20531: ambiguous-chained-operator
### Context

A few days ago, I stumbled upon this snippet:

```python
False == False in [False]
```

At first, I thought that it evaluates to `False`, *which is wrong*.

That day, I learned that `in` and `not in` are actually [comparison operators](https://docs.python.org/3/reference/expressions.html#comparisons), which makes them chainable.

You can argue that this snippet is a little unobvious, but not terrible; so, I modified it:

```python
False == False in [False] == [False]
```

This lil' guy also evaluates to `True`, but in my opinion, it doesn't even look like valid Python code.

### Proposal

My feeling is that the knowledge of `in` being a comparison (so chainable) operator is more of a secret, and not something obvious. I want to suggest a rule that forbids using `in` in chained comparisons.

I think, the samples I provided are not typical code anyone would write. Strictly speaking, they are the first examples of chained `in` I have ever seen. However, if the proposed rule makes sense, I am willing to implement it.

Note that the [PLR1716](https://docs.astral.sh/ruff/rules/boolean-chained-comparison/#boolean-chained-comparison-plr1716) rule may conflict with this one, I haven't checked for that yet.

### PS

While writing this proposal, I considered another rule that would prohibit things like `a < b >= c`; more formal definition would be "if a chained comparison contains `<=` or `<`, it should only use `<=` and `<` (same with `>` and `>=`)", so:

- `a < b <= c <= d` - valid;
- `a > b > c >= d` - valid;
- `a < b >= c == d` - invalid.

I would like to hear feedback on this suggestion, thanks in advance.

kapkekes:
## Summary

Part of #970.

Implements `PLW3601` to fully match the original `W3601`.

Closes #20531, probably?

## Test Plan

`cargo nextest run` and `cargo insta test`.


ntBre:
I'll go ahead and close this for now, based on my comment on the issue (https://github.com/astral-sh/ruff/issues/20531#issuecomment-3348988228) and the reviews from the earlier PR (https://github.com/astral-sh/ruff/pull/10028).

Thank you for working on this and sorry again for not catching the earlier work sooner!

PR #20598, opened 2025-09-27
[ty] defer inference of legacy TypeVar bound/constraints/defaults

carljm:
## Summary

This allows us to handle self-referential bounds/constraints/defaults without panicking.

Handles more cases from https://github.com/astral-sh/ty/issues/256

## Test Plan

Added mdtest.


crates/ty_ide/src/goto_type_definition.rs
```diff
@@ -308,26 +308,8 @@ mod tests {
...
-        info: Source
-         --> main.py:6:1
-          |
-        4 | Alias = TypeAliasType("Alias", tuple[int, int])
-        5 |
-        6 | Alias
-          | ^^^^^
-          |
-        "#);
+        // TODO: This should jump to the definition of `Alias` above.
```
MichaReiser, line 311: Can we create an issue for this. It's otherwise hard to remember

crates/ty_python_semantic/resources/mdtest/generics/legacy/classes.md
```diff
@@ -1,5 +1,12 @@
 # Generic classes: Legacy syntax
 
+We use TypeVar defaults here, which was added in Python 3.13 for legacy typevars.
```
AlexWaygood:
```suggestion
We use TypeVar defaults here, which were added in Python 3.13 for legacy typevars.
```

AlexWaygood: (Though as we've discussed, we should really permit users to pass `default=whatever` even on lower Python versions if the user is making use of `typing_extensions.TypeVar`. Doesn't need to be fixed in this PR, but the prose could possibly be adjusted here to reflect that?)

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -33,13 +33,12 @@ reveal_type(T.__name__)  # revealed: Literal["T"]
 from typing import TypeVar
 
 T = TypeVar("T")
-# TODO: no error
 # error: [invalid-legacy-type-variable]
 U: TypeVar = TypeVar("U")
 
-# error: [invalid-legacy-type-variable] "A legacy `typing.TypeVar` must be immediately assigned to a variable"
-# error: [invalid-type-form] "Function calls are not allowed in type expressions"
-TestList = list[TypeVar("W")]
```
AlexWaygood, line 42: these assertions still pass (and still seem useful) -- what's the motivation for getting rid of them?

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -49,7 +48,7 @@ TestList = list[TypeVar("W")]
 ```py
 from typing import TypeVar
 
-# error: [invalid-legacy-type-variable] "The name of a legacy `typing.TypeVar` (`Q`) must match the name of the variable it is assigned to (`T`)"
+# error: [invalid-legacy-type-variable]
```
AlexWaygood, line 74: similarly here, what's the motivation for getting rid of the error message from the assertion?

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -66,6 +65,22 @@ T = TypeVar("T")
...
+types = (int, str)
+
+# error: [invalid-legacy-type-variable]
+T = TypeVar("T", *types)
+reveal_type(T)  # revealed: TypeVar
+
+# error: [invalid-legacy-type-variable]
+S = TypeVar("S", **{"bound": int})
+reveal_type(S)  # revealed: TypeVar
+```
```
AlexWaygood, line 105: are these mandated by the spec? They almost feel like lints to me, given that (I assume) we'll infer the type just fine for each value passed into each parameter

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -122,6 +142,20 @@ reveal_type(T.__constraints__)  # revealed: tuple[()]
...
+T = TypeVar("T", bound=TypedDict)
+```
+
+The upper bound must be a valid type expression:
+
+```py
+from typing import TypedDict
+
+# error: [invalid-type-form]
+T = TypeVar("T", bound=TypedDict)
```
AlexWaygood: I think we probably only need one copy of this assertion? 😄

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -178,6 +231,66 @@ T = TypeVar("T", covariant=cond())
...
+[environment]
+python-version = "3.10"
+```
+
+In a stub file, features from the latest supported Python version can be used on any version:
+
+```pyi
+from typing import TypeVar
+T = TypeVar("T", default=int)
+```
```
AlexWaygood: I still think it would be fine to treat stubs the same way as runtime files here -- I'd happily make the necessary changes to typeshed... but I guess other stub file authors may be relying on the behaviour of mypy and pyright here, and might get annoyed with us if we complained about this. Blegh.

Maybe it's worth saying explicitly in the prose that this isn't specified very well, but that we emulate other type checkers' behaviour here for maximum compatibility?

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -231,4 +344,94 @@ def constrained(x: T_constrained):
...
+```py
+from typing import TypeVar, Generic
+
+T = TypeVar("T", bound=list["G"])
+
+class G(Generic[T]):
+    x: T
+
+reveal_type(G[list[G]]().x)  # revealed: list[G[Unknown]]
+```
```
AlexWaygood, line 404: hrm... on line 351 you said that they _cannot_ be cyclic, but here you say that they _can_ "cyclically" refer back to their own type?

Though FWIW I think it's debatable whether this example really counts as a cycle. `G` as it appears in the bound of `list[G]` here is syntactic sugar for the default specialization of `G`, which is `G[Unknown]` -- so this type variable's upper bound is `list[G[Unknown]]`, which is not a type that refers back to `T` in any way since `T` has been fully "specialized away" from the type.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -231,4 +344,94 @@ def constrained(x: T_constrained):
...
+
+### Defaults
+
+```toml
+[environment]
+python-version = "3.13"
+```
+
+Defaults can be generic, but can only refer to earlier typevars:
```
AlexWaygood:
```suggestion
Defaults can be generic, but can only refer to typevars already defined in the same scope:
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -231,4 +344,94 @@ def constrained(x: T_constrained):
...
+reveal_type(C[int]().x)  # revealed: int
+reveal_type(C[int]().y)  # revealed: int
+
+# TODO: error
+V = TypeVar("V", default="V")
+
+class D(Generic[V]):
+    x: V
+
+reveal_type(D().x)  # revealed: V@D
```
AlexWaygood, line 438:
```suggestion
# TODO: should never leak a typevar like this in type inference
reveal_type(D().x)  # revealed: V@D
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -231,4 +344,94 @@ def constrained(x: T_constrained):
...
+_DataT = TypeVar("_DataT", bound=int, default=int)
+
+class Event(Generic[_DataT]):
+    def __init__(self, data: _DataT) -> None:
+        self.data = data
+
+def async_fire_internal(event_data: _DataT):
+    event: Event[_DataT] | None = None
+    event = Event(event_data)
+```
```
AlexWaygood, line 462: did this break in an earlier version of this PR? It seems to work fine on `main`: https://play.ty.dev/d6e5578c-31fa-4e0e-b54d-64c920e7a478

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1628,10 +1624,22 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
                 self.visit_expr(&node.value);
-                let value = self.add_standalone_assigned_expression(&node.value, node);
 
-                for target in &node.targets {
-                    self.add_unpackable_assignment(&Unpackable::Assign(node), target, value);
+                // Optimization for the common case: if there's just one target, and it's not an
+                // unpacking, and the target is a simple name, we don't need the RHS to be a
+                // standalone expression at all.
+                if let [target] = &node.targets[..]
+                    && target.is_name_expr()
```
AlexWaygood, line 1632:
```suggestion
                // Optimization for the common case: if there's just one target, and it's not an
                // unpacking, and the target is a simple name, we don't need the RHS to be a
                // standalone expression at all.
                if let [target @ ast::Expr::Name(_)] = &node.targets[..]
```

also, this seems like quite a big win from the Codspeed report -- I think it could also be safely applied for lots of non-name expressions? `x.foo = 42` and `x[0] = 42` can also definitely be excluded from being possible unpackable assignments.

crates/ty_python_semantic/src/types.rs
```diff
@@ -7948,38 +7929,96 @@ impl<'db> TypeVarInstance<'db> {
...
+                ))
+            }
+            // legacy typevar
+            DefinitionKind::Assignment(assignment) => {
+                let call_expr = assignment.value(&module).as_call_expr()?;
+                let expr = &call_expr.arguments.find_keyword("bound")?.value;
+                Some(definition_expression_type(db, definition, expr))
+            }
+            _ => None,
+        }?;
```
```suggestion
        let ty = match definition.kind(db) {
            // PEP 695 typevar
            DefinitionKind::TypeVar(typevar) => {
                let typevar_node = typevar.node(&module);
                definition_expression_type(
                    db,
                    definition,
                    typevar_node.bound.as_ref()?,
                )
            }
            // legacy typevar
            DefinitionKind::Assignment(assignment) => {
                let call_expr = assignment.value(&module).as_call_expr()?;
                let expr = &call_expr.arguments.find_keyword("bound")?.value;
                definition_expression_type(db, definition, expr)
            }
            _ => return None,
        };
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7948,38 +7929,96 @@ impl<'db> TypeVarInstance<'db> {
...
+                            .args
+                            .iter()
+                            .skip(1)
+                            .map(|arg| definition_expression_type(db, definition, arg))
+                            .collect::<Box<_>>(),
+                    ),
+                )
+            }
+            _ => None,
+        }?;
```
```suggestion
        let ty = match definition.kind(db) {
            // PEP 695 typevar
            DefinitionKind::TypeVar(typevar) => {
                let typevar_node = typevar.node(&module);
                definition_expression_type(db, definition, typevar_node.bound.as_ref()?)
                    .into_union()?
            }
            // legacy typevar
            DefinitionKind::Assignment(assignment) => {
                let call_expr = assignment.value(&module).as_call_expr()?;
                // We don't use `UnionType::from_elements` or `UnionBuilder` here,
                // because we don't want to simplify the list of constraints as we would with
                // an actual union type.
                // TODO: We probably shouldn't use `UnionType` to store these at all? TypeVar
                // constraints are not a union.
                UnionType::new(
                    db,
                    call_expr
                        .arguments
                        .args
                        .iter()
                        .skip(1)
                        .map(|arg| definition_expression_type(db, definition, arg))
                        .collect::<Box<_>>(),
                )
            }
            _ => return  None,
        };
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3991,7 +3995,37 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            self.infer_legacy_typevar(target, call_expr, definition, typevar_class)
+                        } else {
+                            self.infer_call_expression_impl(call_expr, callable_type, tcx)
+                        };
+                        self.store_expression_type(value, ty);
+                        ty
+                    } else {
+                        self.infer_expression(value, tcx)
+                    }
+                };
```
AlexWaygood, line 4029: a little less nested indentation:

```suggestion
                let value_ty = if let Some(standalone_expression) = self.index.try_expression(value)
                {
                    self.infer_standalone_expression_impl(value, standalone_expression, tcx)
                } else if let ast::Expr::Call(call_expr) = value {
                    // If the RHS is not a standalone expression, this is a simple assignment
                    // (single target, no unpackings). That means it's a valid syntactic form
                    // for a legacy TypeVar creation; check for that.
                    let callable_type = self.infer_maybe_standalone_expression(
                        call_expr.func.as_ref(),
                        TypeContext::default(),
                    );

                    let typevar_class = callable_type
                        .into_class_literal()
                        .and_then(|cls| cls.known(self.db()))
                        .filter(|cls| {
                            matches!(cls, KnownClass::TypeVar | KnownClass::ExtensionsTypeVar)
                        });

                    let ty = if let Some(typevar_class) = typevar_class {
                        self.infer_legacy_typevar(target, call_expr, definition, typevar_class)
                    } else {
                        self.infer_call_expression_impl(call_expr, callable_type, tcx)
                    };
                    self.store_expression_type(value, ty);
                    ty
                } else {
                    self.infer_expression(value, tcx)
                };
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
AlexWaygood: Oh, I see that with the new implementation of `TypeVar` definition parsing, this probably simplifies the implementation a lot. So feel free to disregard this comment.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        .infer_expression(&kwarg.value, TypeContext::default())
+                        .bool(db)
+                    {
+                        Truthiness::AlwaysTrue => covariant = true,
+                        Truthiness::AlwaysFalse => {}
+                        Truthiness::Ambiguous => {
+                            return error(
+                                &self.context,
+                                "The `covariant` parameter of `typing.TypeVar` \
+                                cannot have an ambiguous value",
```
AlexWaygood: maybe a slightly clearer error message?

```suggestion
                                "The `covariant` parameter of `typing.TypeVar` \
                                cannot have an ambiguous truthiness",
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        .infer_expression(&kwarg.value, TypeContext::default())
+                        .bool(db)
+                    {
+                        Truthiness::AlwaysTrue => contravariant = true,
+                        Truthiness::AlwaysFalse => {}
+                        Truthiness::Ambiguous => {
+                            return error(
+                                &self.context,
+                                "The `contravariant` parameter of `typing.TypeVar` \
+                                cannot have an ambiguous value",
```
```suggestion
                                cannot have an ambiguous truthiness",
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            .map(|arg| self.infer_expression(arg, TypeContext::default()))
+            .and_then(Type::into_string_literal)
+            .map(|name| name.value(db))
+        else {
+            return error(
+                &self.context,
+                "The first argument to `typing.TypeVar` must be a string literal.",
+                call_expr,
+            );
+        };
```
AlexWaygood: Hmm:

```
error[invalid-legacy-type-variable]: The first argument to `typing.TypeVar` must be a string literal.
 --> foo.py:3:5
  |
1 | import typing
2 |
3 | T = typing.TypeVar()
  |     ^^^^^^^^^^^^^^^^
  |
info: rule `invalid-legacy-type-variable` is enabled by default

Found 1 diagnostic
```

shouldn't the error message there be something like "must pass a `name` parameter to `typing.TypeVar`"? It sounds like it's complaining about the type of the first argument, but no arguments were passed

Similarly, I think it's totally fine if we don't want to support this (I don't think any other type checker does), but this isn't a _great_ set of error messages here:

```
error[invalid-legacy-type-variable]: The first argument to `typing.TypeVar` must be a string literal.
 --> foo.py:3:5
  |
1 | from typing import TypeVar
2 |
3 | T = TypeVar(name="T")
  |     ^^^^^^^^^^^^^^^^^
  |
info: rule `invalid-legacy-type-variable` is enabled by default

error[invalid-legacy-type-variable]: Unknown keyword argument `name` in `typing.TypeVar` creation
 --> foo.py:3:13
  |
1 | from typing import TypeVar
2 |
3 | T = TypeVar(name="T")
  |             ^^^^^^^^
  |
info: rule `invalid-legacy-type-variable` is enabled by default
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+
+        let Some(target_name) = target.as_name_expr().map(|n| &n.id) else {
```
```suggestion
        let ast::Expr::Name(ast::ExprName { id, .. }) = target_name else {
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            return error(
+                &self.context,
+                "A `typing.TypeVar` must be assigned to a variable",
```
```suggestion
                "A `typing.TypeVar` definition must be a simple variable assignment",
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            None,
+        )))
+    }
+
+    fn infer_assignment_deferred(&mut self, value: &ast::Expr) {
+        // Infer deferred bounds/constraints/defaults of a legacy TypeVar.
+        let Some(call_expr) = value.as_call_expr() else {
+            return;
+        };
+        let arguments = &call_expr.arguments;
```
```suggestion
        let ast::Expr::Call(ast::ExprCall { arguments, .. }) = value else {
            return;
        };
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    if !have_features_from(PythonVersion::PY312) {
+                        // We don't return here; this error is informational since this will error
+                        // at runtime, but the user's intent is plain, we may as well respect it.
+                        error(
+                            &self.context,
+                            "The `infer_variance` parameter of `typing.TypeVar` was added in Python 3.12",
+                            kwarg,
+                        );
+                    }
+                    // TODO support `infer_variance` in legacy TypeVars
```
AlexWaygood, line 4182: You're not calling `self.infer_expression()` on `kwarg.value` here, which then results in `Unknown` if you hover over `infer_variance` in the TypeVar call in the playground. This is a screenshot from a playground build on your branch:

image

```suggestion
                    // TODO support `infer_variance` in legacy TypeVars
                    self.infer_expression(&kwarg.value, TypeContext::default());
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4022,6 +4056,227 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    // at runtime, but it will likely cause fewer cascading errors if we just
+                    // ignore the unknown keyword and still understand as much of the typevar as we
+                    // can.
+                    error(
+                        &self.context,
+                        format_args!(
+                            "Unknown keyword argument `{name}` in `typing.TypeVar` creation",
+                        ),
+                        kwarg,
+                    );
```
AlexWaygood, line 4205: same here -- for an unknown keyword argument, you're never recursing into the value of the argument to infer the type at the moment

```suggestion
                    );
                    self.infer_expression(&kwarg.value, TypeContext::default());
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -6185,6 +6449,21 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 let argument_forms = vec![Some(ParameterForm::Value); call_arguments.len()];
                 self.infer_argument_types(arguments, &mut call_arguments, &argument_forms);
 
+                if class.is_known(self.db(), KnownClass::TypeVar) {
```
AlexWaygood: should this be

```suggestion
                if matches!(class.known(self.db(), Some(KnownClass::TypeVar | KnownClass::ExtensionsTypeVar)) {
```

?

crates/ty_ide/src/goto_type_definition.rs
carljm: I don't think it will be hard to remember, because I have to fix it in order to fix the cases in https://github.com/astral-sh/ty/issues/256 related to `TypeAliasType`. So in that sense we already have an issue that ensures we won't forget it.

carljm: (And I plan to do this PR next.)

crates/ty_python_semantic/resources/mdtest/generics/legacy/classes.md
carljm: > we should really permit users to pass `default=whatever` even on lower Python versions if the user is making use of `typing_extensions.TypeVar`

This PR already does that (and has tests for it, in another file below).

The thing this PR doesn't do is allow you to access the `__default__` attribute on the resulting typevar created via `typing_extensions.TypeVar`, because we don't yet track how a given TypeVar was originally created.

The change here is because some tests in this file import `typing.TypeVar`, not `typing_extensions.TypeVar`. Would you rather I change them all to import from `typing_extensions`, instead of changing the Python version for this file?

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: It seemed like an odd thing to test in this file, because it includes this diagnostic "function call not allowed in type expression", which is arguably the more basic/important error in this case, and is not an error related to TypeVars at all. So I replaced the test with a mostly-equivalent version that only tests the (same) TypeVar-related diagnostic, without the distraction of the unrelated diagnostic.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: I think I got rid of the error message with the intent to snapshot-diagnostics instead, but then decided against that because these tests make too much use of `reveal_type`, which clutters diagnostic snapshots horribly.

I can put the error message back, but it feels odd/arbitrary to inconsistently have one or two error messages tested and the majority not, for no particular reason.

I could consistently add error messages everywhere, but in general I prefer not doing that, because they are painful to update when error messages change. I'd rather use snapshot-diagnostics for testing the content of diagnostic messages.

I can add a dedicated set of snapshot-diagnostics tests for all the cases tested here?

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: Yes, it was a combination of "easier to implement without it", "no other type checker supports it", and "given the other restrictions on TypeVar argument values, it seems unlikely there's a compelling use case for this."

If we ever run into a compelling use case, it wouldn't actually be _that_ hard to add support for it later. But doesn't seem worth it right now.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: Huh, weird! Good catch :)

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: Yes, the fact that other type checkers allow this is what convinced me we should too. And to be honest, I think other type checkers are right to allow it; there's really only upside. The point of a type checker is to catch things that will error at runtime. A TypeVar definition in a stub file that uses more recent features will not error at runtime. So what's the point of being pedantic about making users use `typing_extensions.TypeVar` in a stub file?

carljm: I can extend the prose here a bit, but I'm inclined to actually make the case for the feature, not just say "other type checkers do it, so we also do it for compatibility."

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
```diff
@@ -178,6 +231,66 @@ T = TypeVar("T", covariant=cond())
...
+python-version = "3.10"
+```
+
+```py
+from typing_extensions import TypeVar
+
+T = TypeVar("T", default=int)
+# TODO: should not error, should reveal `int`
+# error: [unresolved-attribute]
+reveal_type(T.__default__)  # revealed: Unknown
```
carljm, line 316: Note that we (correctly) don't error on the TypeVar creation here. The only part of this test that is a TODO is accessing the `__default__` attribute afterwards.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: > hrm... on line 351 you said that they _cannot_ be cyclic, but here you say that they _can_ "cyclically" refer back to their own type?

I said they "cannot be generic, cyclic or otherwise", which does not say they can't be cyclic.

The bound here is not generic, because `list[G]` is a fully specialized type (as you point out below), so the above statement doesn't apply here.

> Though FWIW I think it's debatable whether this example really counts as a cycle. `G` as it appears in the bound of `list[G]` here is syntactic sugar for the default specialization of `G`, which is `G[Unknown]` -- so this type variable's upper bound is `list[G[Unknown]]`, which is not a type that refers back to `T` in any way since `T` has been fully "specialized away" from the type.

It counts as a cycle in the way that's IMO most relevant, which is that it causes a divergent Salsa query cycle and panics on current main branch :) The cycle occurs because when specializing `G`, we have to compare the specialization type against the upper bound of the typevar of `G`, which means we have to evaluate `G`, which is `G[Unknown]`, which means we have to specialize `G`... and we are in a cycle.

I think there might be an alternative fix for this specific cycle, where we special-case "unknown specialization" and don't bother checking for assignability to the upper bound (since we know that `Unknown` will be assignable to it, whatever it is). But we don't bother with that in this PR, since the more general laziness fix handles it anyway. I might try it (separately) and see if it shows a meaningful performance benefit, though.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: Yes, it broke in an earlier version of the PR. This is the bug that is fixed with the new `TypeVarInstance::original` field and the new `is_identical_to` methods.

crates/ty_python_semantic/src/semantic_index/builder.rs
carljm: I did that in an earlier version of the PR, but doing it for attribute targets breaks our implicit instance attributes handling. I might be able to do it for subscript expressions? But I suspect they are not common enough to move the needle noticeably on performance. I can experiment with that separately, but there's no need to include it in this PR.

AlexWaygood: Interesting. And, yeah, definitely not a blocker!

crates/ty_python_semantic/resources/mdtest/generics/legacy/classes.md
AlexWaygood: > Would you rather I change them all to import from `typing_extensions`, instead of changing the Python version for this file?

Yeah, that does make more sense to me -- I think it's nice to try to avoid using version blocks like this where possible, or it makes it look a little like the feature is only available on newer versions of Python, and we'll emit an error on it on older versions.

It obviously doesn't matter a huge amount, but it made me raise my eyebrow when reading through the tests :-)

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
AlexWaygood: > I could consistently add error messages everywhere, but in general I prefer not doing that, because they are painful to update when error messages change. I'd rather use snapshot-diagnostics for testing the content of diagnostic messages.
>
> I can add a dedicated set of snapshot-diagnostics tests for all the cases tested here?

That's fair. Adding a dedicated set of snapshot-diagnostics tests sounds like a good solution!

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
AlexWaygood: Meh. Yeah, it's useful, sure, but so would a number of other ad-hoc special cases that you _could_ apply to stubs if you wanted to. I feel like there's also a strong case to say that "stubs should mostly work the same way as runtime files, except for a narrow series of exceptions", and I'm not sure that this exception is really useful _enough_. There's also a strong benefit in keeping special cases to a minimum -- it reduces the cognitive overhead of understanding stub files. And `typing_extensions` is always available as an import in stub files, even if you don't depend on it, because typeshed pretends it's part of the stdlib, so I'm just really not sure it brings much benefit.

To me it feels kinda confusing that we'd complain about `typing.TypeVar(infer_variance=True)` in a `.py` file on Python <3.12 but not in a stub file. I'm not suggesting we change the behaviour, given that compatibility with other type checkers is important, but it's not a special case I'd introduce if I were starting from a blank slate, personally

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
AlexWaygood: > I said they "cannot be generic, cyclic or otherwise", which does not say they can't be cyclic.

oh, sorry. I misread that as being "three separate things that they cannot be" rather than "one thing that they cannot be, which has two subcategories"

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: Yes, thanks! I'll also add a test for this.

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: I improved the error message for the first case.

I checked the second case; mypy does not support it, but pyright and pyrefly both do. It's not hard to support, so I will go ahead and add support.

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: Done.

crates/ty_python_semantic/resources/mdtest/diagnostics/legacy_typevars.md
```diff
@@ -0,0 +1,136 @@
...
+> most one of these may be passed.
+
+```py
+from typing import TypeVar
+
+# error: [invalid-legacy-type-variable]
+T = TypeVar("T", covariant=True, contravariant=True)
+```
+
+## Variance parameters must be unambiguous
```
AlexWaygood: The same applies for `infer_variance=True`... Though we can defer adding that check until we add full support for that keyword, I suppose

carljm: I went ahead and added this check for consistency.

crates/ty_python_semantic/src/semantic_index/builder.rs
carljm: Testing the impact of this at https://github.com/astral-sh/ruff/pull/20795

crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
carljm: > The cycle occurs because when specializing `G`, we have to compare the specialization type against the upper bound of the typevar of `G`, which means we have to evaluate `G`, which is `G[Unknown]`, which means we have to specialize `G`... and we are in a cycle.

This was not actually an accurate summary of why there was a cycle here, before this PR. The correct description is this: as part of inferring the type of the symbol `T`, we had to eagerly infer the type of its upper bound, which required knowing the type of `G`, and since `G` is generic over `T`, that requires knowing the type of `T`, and right there we have a cycle.

I think this is reasonably described as a cycle between the definitions of `T` and `G`.

The "alternative fix" I described above is not actually an alternative at all -- we still need the lazy bounds to even get to the point of having independently-inferrable types for `T` and `G`. And it's not an optimization, either, because we actually already do this -- we don't check against upper bounds when applying an unknown or default specialization.

crates/ty_python_semantic/src/semantic_index/builder.rs
carljm: As I'd suspected, the impact does not appear to be detectable in either memory usage or CPU. Given that I think it makes the invariants slightly harder to understand, and requires changing a use of `infer_standalone_expression` to `infer_maybe_standalone_expression`, I'm going to close that PR.

carljm:
Thanks @AlexWaygood for the excellent review here!

PR #20602, opened 2025-09-27
[ty] Introduce `TypeRelation::Redundancy`

AlexWaygood:
## Summary

The union `T | U` can be validly simplified to `U` iff:
1. `T` is a subtype of `U` OR
2. `T` is equivalent to `U` OR
3. `U` is a union and contains a type that is equivalent to `T` OR
4. `T` is an intersection and contains a type that is equivalent to `U`

(In practice, the only situation in which 2, 3 or 4 would be true when (1) was not true would be if `T` or `U` is a dynamic type.)

Currently we achieve these simplifications in the union builder by doing something along the lines of `t.is_subtype_of(db, u) || t.is_equivalent_to_(db, u) || t.into_intersection().is_some_and(|intersection| intersection.positive(db).contains(&u)) || u.into_union().is_some_and(|union| union.elements(db).contains(&t))`. But this is both slow and misses some cases (it doesn't simplify the union `Any | (Unknown & ~None)` to `Any`, for example). We can improve the consistency and performance of our union simplifications by adding a third type relation that sits in between `TypeRelation::Subtyping` and `TypeRelation::Assignability`: `TypeRelation::UnionSimplification`.

This change leads to simpler, more user-friendly types due to the more consistent simplification. It also lead to a pretty huge performance improvement!

## Test Plan

Existing tests, plus some new ones.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/union_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/union_types.md
@@ -323,6 +323,8 @@ def f(
     d: tuple[str, ...] | tuple[int],
     e: tuple[()] | tuple[Any, ...],
     f: tuple[Any, ...] | tuple[()],
+    g: tuple[Any, ...] | tuple[Any | str, ...],
+    h: tuple[Any | str, ...] | tuple[Any, ...],
 ):
     reveal_type(a)  # revealed: tuple[int, ...]
     reveal_type(b)  # revealed: tuple[int, ...]
@@ -330,4 +332,48 @@ def f(
     reveal_type(d)  # revealed: tuple[str, ...] | tuple[int]
     reveal_type(e)  # revealed: tuple[Any, ...]
     reveal_type(f)  # revealed: tuple[Any, ...]
+    reveal_type(g)  # revealed: tuple[Any | str, ...]
+    reveal_type(h)  # revealed: tuple[Any | str, ...]
+```
+
+## Unions of other generic containers
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+from typing import Any
+
+class Bivariant[T]: ...
+
+class Covariant[T]:
+    def get(self) -> T:
+        raise NotImplementedError
+
+class Contravariant[T]:
+    def receive(self, input: T) -> None: ...
+
+class Invariant[T]:
+    mutable_attribute: T
+
+def _(
+    a: Bivariant[Any] | Bivariant[Any | str],
+    b: Bivariant[Any | str] | Bivariant[Any],
+    c: Covariant[Any] | Covariant[Any | str],
+    d: Covariant[Any | str] | Covariant[Any],
+    e: Contravariant[Any | str] | Contravariant[Any],
+    f: Contravariant[Any] | Contravariant[Any | str],
+    g: Invariant[Any] | Invariant[Any | str],
+    h: Invariant[Any | str] | Invariant[Any],
+):
+    reveal_type(a)  # revealed: Bivariant[Any]
+    reveal_type(b)  # revealed: Bivariant[Any | str]
+    reveal_type(c)  # revealed: Covariant[Any | str]
+    reveal_type(d)  # revealed: Covariant[Any | str]
+    reveal_type(e)  # revealed: Contravariant[Any]
+    reveal_type(f)  # revealed: Contravariant[Any]
+    reveal_type(g)  # revealed: Invariant[Any] | Invariant[Any | str]
+    reveal_type(h)  # revealed: Invariant[Any | str] | Invariant[Any]
 ```
--- a/crates/ty_python_semantic/resources/mdtest/union_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/union_types.md
@@ -306,3 +306,28 @@ def _(c: BC, d: BD):
     reveal_type(c)  # revealed: Literal[b""]
     reveal_type(d)  # revealed: Literal[b""]
 ```
+
+## Unions of tuples
+
+A union of a fixed-length tuple and a variable-length tuple must be collapsed to the variable-length
+element, never to the fixed-length element (`tuple[()] | tuple[Any, ...]` -> `tuple[Any, ...]`, not
+`tuple[()]`).
+
+```py
+from typing import Any
+
+def f(
+    a: tuple[()] | tuple[int, ...],
+    b: tuple[int, ...] | tuple[()],
+    c: tuple[int] | tuple[str, ...],
+    d: tuple[str, ...] | tuple[int],
+    e: tuple[()] | tuple[Any, ...],
+    f: tuple[Any, ...] | tuple[()],
+):
+    reveal_type(a)  # revealed: tuple[int, ...]
+    reveal_type(b)  # revealed: tuple[int, ...]
+    reveal_type(c)  # revealed: tuple[int] | tuple[str, ...]
+    reveal_type(d)  # revealed: tuple[str, ...] | tuple[int]
+    reveal_type(e)  # revealed: tuple[Any, ...]
+    reveal_type(f)  # revealed: tuple[Any, ...]
+```
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1413,43 +1413,9 @@ impl<'db> Type<'db> {
         }
     }
 
-    /// Return true if this type is a [subtype of] type `target`.
+    /// Return true if this type is a subtype of type `target`.
     ///
-    /// For fully static types, this means that the set of objects represented by `self` is a
-    /// subset of the objects represented by `target`.
-    ///
-    /// For gradual types, it means that the union of all possible sets of values represented by
-    /// `self` (the "top materialization" of `self`) is a subtype of the intersection of all
-    /// possible sets of values represented by `target` (the "bottom materialization" of
-    /// `target`). In other words, for all possible pairs of materializations `self'` and
-    /// `target'`, `self'` is always a subtype of `target'`.
-    ///
-    /// Note that this latter expansion of the subtyping relation to non-fully-static types is not
-    /// described in the typing spec, but the primary use of the subtyping relation is for
-    /// simplifying unions and intersections, and this expansion to gradual types is sound and
-    /// allows us to better simplify many unions and intersections. This definition does mean the
-    /// subtyping relation is not reflexive for non-fully-static types (e.g. `Any` is not a subtype
-    /// of `Any`).
-    ///
-    /// [subtype of]: https://typing.python.org/en/latest/spec/concepts.html#subtype-supertype-and-type-equivalence
-    ///
-    /// There would be an even more general definition of subtyping for gradual types, allowing a
-    /// type `S` to be a subtype of a type `T` if the top materialization of `S` (`S+`) is a
-    /// subtype of `T+`, and the bottom materialization of `S` (`S-`) is a subtype of `T-`. This
-    /// definition is attractive in that it would restore reflexivity of subtyping for all types,
-    /// and would mean that gradual equivalence of `S` and `T` could be defined simply as `S <: T
-    /// && T <: S`. It would also be sound, in that simplifying unions or intersections according
-    /// to this definition of subtyping would still result in an equivalent type.
-    ///
-    /// Unfortunately using this definition would break transitivity of subtyping when both nominal
-    /// and structural types are involved, because Liskov enforcement for nominal types is based on
-    /// assignability, so we can have class `A` with method `def meth(self) -> Any` and a subclass
-    /// `B(A)` with method `def meth(self) -> int`. In this case, `A` would be a subtype of a
-    /// protocol `P` with method `def meth(self) -> Any`, but `B` would not be a subtype of `P`,
-    /// and yet `B` is (by nominal subtyping) a subtype of `A`, so we would have `B <: A` and `A <:
-    /// P`, but not `B <: P`. Losing transitivity of subtyping is not tenable (it makes union and
-    /// intersection simplification dependent on the order in which elements are added), so we do
-    /// not use this more general definition of subtyping.
+    /// See [`TypeRelation::Subtyping`] for more details.
     pub(crate) fn is_subtype_of(self, db: &'db dyn Db, target: Type<'db>) -> bool {
         self.when_subtype_of(db, target).is_always_satisfied()
     }
@@ -1458,9 +1424,9 @@ impl<'db> Type<'db> {
         self.has_relation_to(db, target, TypeRelation::Subtyping)
     }
 
-    /// Return true if this type is [assignable to] type `target`.
+    /// Return true if this type is assignable to type `target`.
     ///
-    /// [assignable to]: https://typing.python.org/en/latest/spec/concepts.html#the-assignable-to-or-consistent-subtyping-relation
+    /// See [`TypeRelation::Assignability`] for more details.
     pub(crate) fn is_assignable_to(self, db: &'db dyn Db, target: Type<'db>) -> bool {
         self.when_assignable_to(db, target).is_always_satisfied()
     }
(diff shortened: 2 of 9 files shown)


AlexWaygood:
# Ecosystem analysis

This all looks good to me.

Here's a slightly more minimal example of what's going on with the new schemathesis diagnostics. On `main` we have this behaviour:

```py
from typing import Any

def f(x: dict[Any, Any] | dict[Any | str, Any | dict[Any, Any]]):
    reveal_type(x)            # dict[Any, Any] | dict[Any | str, Any | dict[Any, Any]]
    reveal_type(x.items())    # dict_items[Any, Any] | dict_items[Any | str, Any | dict[Any, Any]]

    for a, b in x.items():
        reveal_type(a)        # Any
        reveal_type(b)        # Any
```

And on this PR we have this behaviour:

```py
from typing import Any

def f(x: dict[Any, Any] | dict[Any | str, Any | dict[Any, Any]]):
    reveal_type(x)            # dict[Any, Any] | dict[Any | str, Any | dict[Any, Any]]
    reveal_type(x.items())    # dict_items[Any | str, Any | dict[Any, Any]]

    for a, b in x.items():
        reveal_type(a)        # Any | str
        reveal_type(b)        # Any | dict[Any, Any]
```

`dict_items[Any, Any] | dict_items[Any | str, Any | dict[Any, Any]]` is recognised as being simplifiable to `dict_items[Any | str, Any | dict[Any, Any]]` on this PR branch because `dict_items` is covariant in both its type parameters. The simplification of the union to a non-union type means that we avoid a known bug when iterating over unions (https://github.com/astral-sh/ty/issues/1089, which will be fixed by https://github.com/astral-sh/ruff/pull/20368), so in this case actually means that we preserve a stricter type at the end of the day.

The new diagnostics on `xarray` and `zulip` seem to have a similar cause.

crates/ty_python_semantic/src/types.rs
```diff
@@ -1536,10 +1543,33 @@ impl<'db> Type<'db> {
...
-                // TODO: Implement assignability and subtyping for TypedDict
-                ConstraintSet::from(relation.is_assignability())
-            }
+            // Dynamic is only a subtype of `object` and only a supertype of `Never`; both were
+            // handled above. It's always assignable, though.
+            //
+            // Union simplification sits in between subtyping and assignability. `Any <: T` only
+            // holds true if `T` is also a dynamic type or a union that contains a dynamic type.
+            // Similarly, `T <: Any` only holds true if `T` is a dynamic type or an intersection
+            // that contains a dynamic type.
```
dcreager, line 1519: Maybe this is the key bit? Can this be expressed in terms of the top/bottom materializations of the two types?

carljm:
I suspect that the relation implemented here isn't theoretically consistent, but it's nonetheless practically useful, and so we should probably document the theoretical inconsistency and do it anyway.

I think this goes back to the discussion on https://github.com/astral-sh/ruff/pull/18799 (and specifically [this comment](https://github.com/astral-sh/ruff/pull/18799#issuecomment-2993917848)). There is a more generalized definition of subtyping for gradual types (proposed by @JelleZijlstra in that PR discussion) where the rule is that `S <: T iff S- <: T- && S+ <: T+` -- in other words, every materialization of S must be a subtype of some materialization of T, and every materialization of T must be a supertype of some materialization of S.

Unfortunately we can't fully use that definition, because it breaks down (and introduces non-transitivity) in the interaction between nominal and non-fully-static structural types. It introduces cases where a nominal type X is a (nominal) subtype of a nominal type Y, but overrides a gradual attribute of Y with a more precise type, in such a way that now Y is a subtype of some structural type P, but X is not. Which gives that `X <: Y` and `Y <: P`, but `X </: P`, breaking transitivity. (One theoretical solution to this is to treat all types as structural, with nominal subtyping becoming an additional structural "tag", such that X subclassing Y doesn't necessarily imply `X <: Y`, but that's probably too big a departure from the status quo to be workable for us.)

If we _could_ use that definition of subtyping, it would remove the need to check for both equivalence and subtyping in union simplification, because under that definition of gradual subtyping, any two equivalent types (even gradual ones) are subtypes of each other (that is, it generalizes reflexivity of subtyping.)

I believe that the addition of the `t.into_intersection().is_some_and(|intersection| intersection.positive(db).contains(&u)) || u.into_union().is_some_and(|union| union.elements(db).contains(&t))` clauses to `UnionBuilder` simplification, and now the formalization of that into the new `UnionSimplification` relation, is implementing a partial / ad-hoc version of the more generalized definition of gradual subtyping, but by applying it only to unions and intersections (and not to subtyping of structural types) it avoids the non-transitivity problem.

AlexWaygood:
Okay -- I've had a go at thinking through and writing up some theoretical foundations for this! LMK what you think.

crates/ty_python_semantic/src/types.rs
AlexWaygood: I think so, yes: although `Any & T` is not a subtype of `Any`, for any type `T` it is redundant to add `Any & T` to a union that already contains `Any` as the bottom materialization of `(Any & T) | Any` will always be equivalent to the bottom materialization of `Any`, and the top materialization of `(Any & T) | Any` will always be equivalent to the top materialization of `Any`.

I wrote this up in full as the doc-comment to `TypeRelation::UnionSimplification`.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8993,16 +8990,136 @@ impl<'db> ConstructorCallError<'db> {
...
+    /// type checker should emit an error when a value of type `B` is assigned to a
+    /// variable declared as having type `A`.
+    ///
+    /// For a pair of [fully static] types `A` and `B`, the assignability relation
+    /// between `A` and `B` is the same as the subtyping relation.
+    ///
+    /// Between a pair of `C` and `D` where either `C` or `D` is not fully static, the
+    /// assignability relation may be more permissive than the subtyping relation. `D`
+    /// can be said to be assignable to `C` if *any* possibly fully static [materialization]
+    /// of `D` is a subtype of *any* possible fully static materialization of `C`.
```
JelleZijlstra: You can also express this as `Bottom[D] <: Top[C]`.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8993,16 +8990,136 @@ impl<'db> ConstructorCallError<'db> {
...
+    /// [fully static]: https://typing.python.org/en/latest/spec/glossary.html#term-fully-static-type
+    /// [materialization]: https://typing.python.org/en/latest/spec/glossary.html#term-materialize
     Assignability,
+
+    /// The "union simplification" relation.
+    ///
+    /// The union simplification relation dictates whether the union `A | B` can be
+    /// pragmatically simplified to the type `A` without downstream consequences on ty's
+    /// inference of types elsewhere.
+    ///
```
JelleZijlstra, line 9068: I think the name is a bit unfortunate because it's hard to formulate other terms from it. I'll say that `D` is union-simplifiable with regards to `C` if the relation holds between D and C.

Then, I think the formal formulation can be that `D union-simplifiable C` if `C` is equivalent to `C | D`. Equivalence, as you probably discuss somewhere, for fully static types means that the sets of values described by both types are the same; for dynamic types it means that the top and bottom materializations of both types are equivalent.

JelleZijlstra: I looked at the code a bit more and maybe "redundant with" is a good name. `C` is redundant with `D` if the union `C | D` can be safely simplified to `D`.

AlexWaygood: > I think the name is a bit unfortunate because it's hard to formulate other terms from it

Agreed... but I'm struggling to think of anything better!

crates/ty_python_semantic/src/types.rs
```diff
@@ -8993,16 +8990,136 @@ impl<'db> ConstructorCallError<'db> {
...
+    /// For a pair of [fully static] types `A` and `B`, the assignability relation
```
```suggestion
    /// For a pair of [fully static] types `A` and `B`, the union simplification relation
```

crates/ty_python_semantic/src/types.rs
AlexWaygood: > I looked at the code a bit more and maybe "redundant with" is a good name. `C` is redundant with `D` if the union `C | D` can be safely simplified to `D`.

Hmm, maybe `TypeRelation::UnionRedundancy`? And we can say `C` is union-redundant with `D` if the union `C | D` can be safely simplified to `D`

JelleZijlstra: Doesn't it also apply to intersections? That's why I wanted to avoid using the term "union".

AlexWaygood: it does, yes, fair point

JelleZijlstra (commented):
This looks correct and useful to me, besides my suggestions about the name and phrasing.

crates/ty_python_semantic/src/types.rs
dcreager: nit: I would use "some" here instead of "any" to minimize the chance of confusing it for "all"

(also a spelling nit)

```suggestion
    /// can be said to be assignable to `C` if *some* possible fully static [materialization]
    /// of `D` is a subtype of *some* possible fully static materialization of `C`.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -8993,16 +8990,136 @@ impl<'db> ConstructorCallError<'db> {
...
+    /// between `A` and `B` is the same as the subtyping relation.
+    ///
+    /// Between a pair of `C` and `D` where either `C` or `D` is not fully static, the
+    /// union simplification relation sits in between the subtyping relation and the
+    /// assignability relation. `D` can be said to be redundant in a union with `C` if either:
+    ///
+    /// 1. `D` is a subtype of `C`; or,
+    /// 2. The top materialization of the type `C | D` is equivalent to the top
+    ///    materialization of `C`, *and* the bottom materialization of `C | D` is equivalent
+    ///    to the bottom materialization of `C`.
```
dcreager: I think (1) is redundant :drum: with (2), and you can simplify the whole thing to "`C | D` is gradually equivalent to `D`, and `C & D` is gradually equivalent to `C`". That even nicely maps with the intuitive English definition. (You might still keep (1) as a separate check in the implementation, as an optimization, but that's not needed here in the theoretical definition)

crates/ty_python_semantic/src/types.rs
dcreager: Though I think the relation is reversed for intersection, right? If we use `≤R` for the new redundancy relation, and `=G` for gradual equivalence, then its

```
C ≤U D <=> (C | D) =G D
C ≤U D <=> (C & D) =G C
```

Mnemonics: the smaller thing is on the left of the relation; union makes things bigger; intersection makes things smaller

(That doesn't change Jelle's point that we don't need to include "union" in the name of the relation since it's relevant to intersections too)

AlexWaygood:
This is now stacked on top of https://github.com/astral-sh/ruff/pull/20650 and https://github.com/astral-sh/ruff/pull/20651, which I _think_ are uncontroversial bugfixes that should probably be landed first

crates/ty_python_semantic/src/types.rs
AlexWaygood: @dcreager, are you able to think of any test cases I could add for intersections (that don't already pass with the logic we have on `main`)? I'm struggling to come up with any, and I can't tell if this is because my head is going fuzzy with all the set theory or because our DNF representation of unions and intersections actually makes the question redundant (🥁) when it comes to intersections.

If it's the latter, then `UnionRedundancy` might not be such a bad name after all, and it might make the concept easier to reason about if we stated that it only applied to union simplification (not intersection simplification)

crates/ty_python_semantic/src/types.rs
AlexWaygood: Oh, great point

crates/ty_python_semantic/src/types.rs
AlexWaygood: but... just because we can't think of any new test cases doesn't mean that it's wrong to replace the existing `if x.is_subtype_of(y) || x.is_equivalent_to(y)` branches in our intersection simplification logic with the new `x.is_redundant_with(y)` method. So it does seem like it's useful (purely from a performance perspective, if nothing else) to apply it to both unions and intersections, even if (unlike with unions) this doesn't lead to any semantic improvement over what we have on `main`.

AlexWaygood:
> This is now stacked on top of #20650 and #20651, which I _think_ are uncontroversial bugfixes that should probably be landed first

Seems like #20651 wasn't so uncontroversial 😆 and my head is going fuzzy with all the type theory there.

On further reflection, I _think_ it should be okay to land this without #20651, so I've rebased this on top of `main` again as a standalone change.

AlexWaygood:
I think I addressed all review comments. This should be ready for another round of review.

crates/ty_python_semantic/resources/mdtest/union_types.md
```diff
@@ -306,3 +306,74 @@ def _(c: BC, d: BD):
...
+    a: Bivariant[Any] | Bivariant[Any | str],
+    b: Bivariant[Any | str] | Bivariant[Any],
+    c: Covariant[Any] | Covariant[Any | str],
+    d: Covariant[Any | str] | Covariant[Any],
+    e: Contravariant[Any | str] | Contravariant[Any],
+    f: Contravariant[Any] | Contravariant[Any | str],
+    g: Invariant[Any] | Invariant[Any | str],
+    h: Invariant[Any | str] | Invariant[Any],
+):
+    reveal_type(a)  # revealed: Bivariant[Any]
```
dcreager, line 371: Incidentally, it would be nice to have a version of `reveal_type` that takes in a type-form parameter, so that we don't have to jump through these hoops to get something we can pass in to `typing_extensions.reveal_type`

crates/ty_python_semantic/src/types.rs
dcreager: > re you able to think of any test cases I could add for intersections (that don't already pass with the logic we have on `main`)?

I can't think of any either. We can add those as a follow-on if we encounter any.

> So it does seem like it's useful (purely from a performance perspective, if nothing else) to apply it to both unions and intersections, even if (unlike with unions) this doesn't lead to any semantic improvement over what we have on `main`.

Agreed

dcreager (approved):

crates/ty_python_semantic/src/types.rs
```diff
@@ -8993,16 +8990,138 @@ impl<'db> ConstructorCallError<'db> {
...
+    ///
+    /// For a pair of [fully static] types `A` and `B`, the redundancy relation between `A`
+    /// and `B` is the same as the subtyping relation.
+    ///
+    /// Between a pair of `C` and `D` where either `C` or `D` is not fully static, the
+    /// redundancy relation sits in between the subtyping relation and the assignability relation.
+    /// `D` can be said to be redundant in a union with `C` if the top materialization of the type
+    /// `C | D` is equivalent to the top materialization of `C`, *and* the bottom materialization
+    /// of `C | D` is equivalent to the bottom materialization of `C`.
+    /// More concisely: `D <: C` iff `Top[C | D] == Top[C]` AND `Bottom[C | D] == Bottom[C]`.
```
sharkdp, line 9077: Another way of saying this would be: `D` can be said to be redundant in a union with `C` if `C | D` is gradually equivalent to `C`.

The top and bottom materializations of two gradual types are the same if and only if they are gradually equivalent.

And the two conditions in the end could also be reformulated as: `Top[D] <: Top[C] AND Bottom[D] <: Bottom[C]`, which would show that "redundancy" is equivalent to the "other subtyping" relation that @carljm experimented with in https://github.com/astral-sh/ruff/pull/18799 as well:

| Relation      | Conditions                                    |
|---------------|-----------------------------------------------|
| Subtyping     | `Top[D] <: Bottom[C]`                         |
| Redundancy    | `Top[D] <: Top[C] AND Bottom[D] <: Bottom[C]` |
| Assignability | `Bottom[D] <: Top[C]`                         |

This also shows that subtyping always implies redundancy, and redundancy implies assignability.

AlexWaygood: > Another way of saying this would be: `D` can be said to be redundant in a union with `C` if `C | D` is gradually equivalent to `C`.

Correct (though gradual equivalence could, I think, be defined in terms of "mutual redundancy", and if we defined redundancy using gradual equivalence, that would be a little cyclic 😆).

> This also shows that subtyping always implies redundancy, and redundancy implies assignability.

Yes, I was plannning on adding property tests to this effect

sharkdp: Reversing this logic also implies that gradual equivalence between `C` and `D` is equivalent to mutual redundancy between `C` and `D`. Which could hint at another strategy to re-implement gradual equivalence in ty.

sharkdp: Oh, I just now saw your comment. Glad we both came up with the term "mutual redundancy" :smile:

sharkdp: A quick summary of my post-merge review and discussion with Alex: Given that the redundancy operation *described* in this PR seems to be equivalent to the "S+ <: T+ && S- <: T-" subtyping from https://github.com/astral-sh/ruff/pull/18799, we should probably try to document why the same problem mentioned in that PR with transitivity of subtyping is not relevant here (probably because what has been *implemented* here does not fully match the defined redundancy relation).

PR #20603, opened 2025-09-27
[ty] Improve disambiguation of class names in diagnostics

AlexWaygood:
## Summary

Currently we'll use the fully qualified name of a class in _some_ cases when reporting instances where one class was expected but another one was found and the two classes have the same name. But there are still many cases where we won't do that -- e.g. in https://github.com/astral-sh/ruff/pull/20368, this is one of the new diagnostics:

```
error[invalid-return-type] Return type does not match returned value: expected `Iterable[Model]`, found `Iterable[Model]`
```

This PR switches the `DisplaySettings::from_possibly_ambiguous_type_pair()` function to use a `TypeVisitor` implementation, so that we can be confident it will deeply recurse into a pair of types and identify possibly ambiguous type names wherever they might be hiding.

## Test Plan

Mdtests

### DIFF
--- a/crates/ty_python_semantic/src/types/display.rs
+++ b/crates/ty_python_semantic/src/types/display.rs
@@ -2,6 +2,7 @@
 
 use std::cell::RefCell;
 use std::fmt::{self, Display, Formatter, Write};
+use std::rc::Rc;
 
 use ruff_db::display::FormatterJoinExtension;
 use ruff_python_ast::str::{Quote, TripleQuotes};
@@ -31,7 +32,7 @@ pub struct DisplaySettings<'db> {
     /// Whether rendering can be multiline
     pub multiline: bool,
     /// Class names that should be displayed fully qualified
-    pub qualified: FxHashSet<&'db str>,
+    pub qualified: Rc<FxHashSet<&'db str>>,
 }
 
 impl<'db> DisplaySettings<'db> {
@@ -62,12 +63,14 @@ impl<'db> DisplaySettings<'db> {
         collector.visit_type(db, type_2);
 
         Self {
-            qualified: collector
-                .class_names
-                .borrow()
-                .iter()
-                .filter_map(|(name, classes)| (classes.len() > 1).then_some(*name))
-                .collect(),
+            qualified: Rc::new(
+                collector
+                    .class_names
+                    .borrow()
+                    .iter()
+                    .filter_map(|(name, classes)| (classes.len() > 1).then_some(*name))
+                    .collect(),
+            ),
             ..Self::default()
         }
     }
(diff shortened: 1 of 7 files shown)


sharkdp:
```diff
- src/bokeh/model/model.py:496:16: error[invalid-return-type] Return type does not match returned value: expected `set[Model]`, found `set[Model]`
+ src/bokeh/model/model.py:496:16: error[invalid-return-type] Return type does not match returned value: expected `set[src.bokeh.model.model.Model]`, found `set[src.bokeh.model.model.Model]`
```
obviously not related to your PR, but it looks like something is wrong here?

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -15,117 +18,125 @@ use crate::types::function::{FunctionType, OverloadLiteral};
...
 
-        if class_1.name(db) == class_2.name(db) {
-            result.qualified()
-        } else {
-            result
-        }
+#[derive(Debug, Default)]
+struct AmbiguousClassCollector<'db> {
+    visited_types: RefCell<FxHashSet<Type<'db>>>,
+    class_names: RefCell<FxHashMap<&'db str, FxHashSet<ClassLiteral<'db>>>>,
```
sharkdp: Probably not worth the additional work, but instead of allocating a hash set of all discovered `ClassLiteral`s with that name, you could introduce an `Option`-like enum
```rs
enum AmbiguityState<'db> {
    Unambiguous(ClassLiteral<'db>),
    Ambiguous,
}
```
When you discover the first class literal, you start with `AmbiguityState::Unambiguous(first_literal)`. And as soon as you discover *another* class literal, you simply flip the state to `AmbiguityState::Ambiguous`. At that point, we're not interested anymore in which exact class literals contributed to the ambiguity.

sharkdp (approved):
Thank you very much. Slightly unfortunate that we gain `.clone`s in multiple places now, but this is probably not performance-critical.

AlexWaygood:
> ```diff
> - src/bokeh/model/model.py:496:16: error[invalid-return-type] Return type does not match returned value: expected `set[Model]`, found `set[Model]`
> + src/bokeh/model/model.py:496:16: error[invalid-return-type] Return type does not match returned value: expected `set[src.bokeh.model.model.Model]`, found `set[src.bokeh.model.model.Model]`
> ```
> 
> obviously not related to your PR, but it looks like something is wrong here?

Yes, definitely (multiple things, in fact...) -- but at least we now know that this specific case isn't solved by printing the fully qualified name in the diagnostics 😆

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: Ahh, that's a great suggestion, thanks!

AlexWaygood:
> Slightly unfortunate that we gain `.clone`s in multiple places now, but this is probably not performance-critical.

Yes, I didn't love this, but couldn't see an easy way to get around it. I think `DisplaySettings` should be fairly cheap to clone, though, since it uses an `Rc<HashSet>` internally rather than a `HashSet`

PR #20604, opened 2025-09-27
[`formatter`] Fix trailing newline insertion in `fmt: off` regions

### LINKED ISSUE #19492: Formatter inserts trailing newline at the end of a file even when inside a `fmt: off` region
### Summary

title is all.

you can test this bug with:
```
# fmt: off
print(1,2,3,4,5,9,13518,8315987831959831,3851758979515,31958139573175915,158315897151935,135815798151,)
```

make sure end of file does not contain newline. run `ruff format`. then newline is inserted.

### Version

ruff 0.12.4

danparizher:
## Summary

Fixes #19492

### DIFF
--- a/crates/ruff_python_formatter/src/module/mod_module.rs
+++ b/crates/ruff_python_formatter/src/module/mod_module.rs
@@ -27,14 +27,27 @@ impl FormatNodeRule<ModModule> for FormatModModule {
                 Ok(())
             }
         } else {
-            write!(
-                f,
-                [
-                    body.format().with_options(SuiteKind::TopLevel),
-                    // Trailing newline at the end of the file
-                    hard_line_break()
-                ]
-            )
+            // Check if the original source ends without a newline and has fmt: off
+            let should_add_trailing_newline = {
+                let source = f.context().source();
+                let source_ends_without_newline = !source.ends_with('\n');
+                let has_fmt_off = source.contains("# fmt: off");
+                !(source_ends_without_newline && has_fmt_off)
+            };
+
+            if should_add_trailing_newline {
+                write!(
+                    f,
+                    [
+                        body.format().with_options(SuiteKind::TopLevel),
+                        // Trailing newline at the end of the file
+                        hard_line_break()
+                    ]
+                )
+            } else {
+                // Don't add trailing newline if we're in a suppressed region
+                body.format().with_options(SuiteKind::TopLevel).fmt(f)
+            }
         }
     }
 }


danparizher:
I've implemented proper suppression state tracking by using `SuppressionKind`—the behavior works correctly with real files—it adds trailing newlines for non-suppressed regions and preserves formatting for suppressed regions.

What I couldn't figure out was that the test snapshot for Test Case 3 isn't showing the expected trailing newline, even though the actual formatter behavior is correct. Wondering if this is an issue with the fixture or intended behavior.

MichaReiser (changes requested):
This looks better, but I think is still incorrect for:

```py
def test():
	# fmt: off
	...
```

This should only disable formatting inside the `test` function but not at the module level.

crates/ruff_python_formatter/tests/snapshots/format@fmt_on_off__function_level_suppression.py.snap
MichaReiser, line 1: You'll need to create different test files or you only test whatever `fmt: off` pattern comes last

crates/ruff_python_formatter/src/module/mod_module.rs
```diff
@@ -27,14 +29,67 @@ impl FormatNodeRule<ModModule> for FormatModModule {
...
+                    for comment_range in comment_ranges.iter() {
+                        let comment_text = &source[comment_range.range()];
+
+                        // Check if this comment is at module level (not indented)
+                        let is_module_level = {
+                            let line_start = source.line_start(comment_range.start());
+                            let before_comment =
+                                &source[TextRange::new(line_start, comment_range.start())];
+                            before_comment.trim_start().is_empty()
+                                || before_comment.trim_start().starts_with('#')
```
MichaReiser: Unfortunatelly, it's more complicated than this:

```py
a = b
    # fmt: off
b = 10
```

This is a module level `fmt: off` comment

But this isn't

```py
if True:
	a = b
  # fmt: off
```


We should also support `yapf: disable` comments

https://github.com/astral-sh/ruff/blob/32be31d2f9dce7a5e23b406c5b8a338dc4d21755/crates/ruff_python_formatter/src/comments/mod.rs#L169-L171

We should also add a test for

```py
# fmt: off
```

I think the right approach is to iterate over all statements and, for each statement, get its comments (using `context.comments.leading_trailing(statement)`). Then test if the last comment is a `fmt: off` comment

crates/ruff_python_formatter/src/module/mod_module.rs
```diff
@@ -17,24 +18,48 @@ impl FormatNodeRule<ModModule> for FormatModModule {
...
+            let mut module_suppressed_at_eof = false;
+            for statement in body {
+                for comment in comments.leading_trailing(statement) {
+                    if comment.is_suppression_off_comment(source) {
+                        module_suppressed_at_eof = true;
+                    } else if comment.is_suppression_on_comment(source) {
+                        module_suppressed_at_eof = false;
+                    }
+                }
+            }
```
MichaReiser: We should avoid doing this check if the source ends with a `\n`.

crates/ruff_python_formatter/src/module/mod_module.rs
```diff
@@ -17,24 +18,48 @@ impl FormatNodeRule<ModModule> for FormatModModule {
...
         } else {
-            write!(
-                f,
-                [
-                    body.format().with_options(SuiteKind::TopLevel),
-                    // Trailing newline at the end of the file
-                    hard_line_break()
-                ]
-            )
+            let comments = f.context().comments();
```
MichaReiser: I suggest restructuring the entire code to

```rust
		body.format.with_options(SuiteKind::TopLevel).fmt(f)?;

		if source.ends_width('\n') || !has_unclosed_fmt_off(f.context()) {
			hard_line_break().fmt(f)?;
		}
```

crates/ruff_python_formatter/resources/test/fixtures/ruff/fmt_on_off/yapf_disable_module.py
```diff
@@ -0,0 +1,4 @@
+# Support for yapf: disable at module level
+# yapf: disable
+print("yapf")
+
```
MichaReiser: Is it intentional that this test case and `module_level_trailing_off` have a trailing newline?

danparizher:
I refactored the modules and extracted a helper. Feel free to commit directly if there's still some missing logic; I know we've been going back and forth for a bit on this one!

crates/ruff_python_formatter/src/module/mod_module.rs
```diff
@@ -17,24 +17,45 @@ impl FormatNodeRule<ModModule> for FormatModModule {
...
-                    body.format().with_options(SuiteKind::TopLevel),
-                    // Trailing newline at the end of the file
-                    hard_line_break()
-                ]
-            )
+            body.format().with_options(SuiteKind::TopLevel).fmt(f)?;
+
+            if source.ends_with(['\n', '\r']) || !has_unclosed_fmt_off(body, f.context()) {
+                hard_line_break().fmt(f)?;
+            }
```
MichaReiser, line 35: Hmm, I'm not sure what's happening. You have to do some debugging but I noticed that none of the tests are failing when I change this to (when I revert your changes)

```suggestion
                hard_line_break().fmt(f)?;

```

It seems we're still always inserting a trailing newline

MichaReiser (commented):
Thanks. I updated the fixture files to remove all trailing new lines to ensure they test the right thing. I also "reverted" your changes locally to verify that the tests start failing, which they don't. I'm not sure why. Can you take a look.

We should also add a test to verify that we preserve the correct number of blank lines for

```py
test
# fmt: off
# many blank lines follow


```

or 

```py
# empty file with many blank lines
# fmt: off



```

danparizher:
I've spent some time debugging but wasn't able to make much headway. Not too sure what's going on either; I've tried multiple iterations of the same solution.

Might take another look later.

MichaReiser:
I suspect that the statement formatting might insert a new line

MichaReiser:
Thanks for looking into this.

I'll close this PR as there hasn't been any progress in a while. Happy to review a re-submission that address the PR feedback

PR #20606, opened 2025-09-28
[ty] `~T` should never be assignable to `T`

AlexWaygood:
## Summary

Currently we do not emit an error on this code:

```py
from ty_extensions import Not

def f[T](x: T, y: Not[T]) -> T:
    x = y
    return x
```

But we should do! `~T` should never be assignable to `T`.

This fixes a small regression introduced in https://github.com/astral-sh/ruff/commit/14fe1228e72dca90db6b3dbb7e07f973d175ea0e#diff-8049ab5af787dba29daa389bbe2b691560c15461ef536f122b1beab112a4b48aR1443-R1446, where a branch that previously returned `false` was replaced with a branch that returns `C::always_satisfiable` -- the opposite of what it used to be! The regression occurred because we didn't have any tests for this -- so I added some tests in this PR that fail on `main`. I only spotted the problem because I was going through the code of `has_relation_to_impl` with a fine toothcomb for https://github.com/astral-sh/ruff/pull/20602 😄

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1562,7 +1562,7 @@ impl<'db> Type<'db> {
             (Type::Intersection(intersection), Type::NonInferableTypeVar(_))
                 if intersection.negative(db).contains(&target) =>
             {
-                ConstraintSet::from(true)
+                ConstraintSet::from(false)
             }
 
             // Two identical typevars must always solve to the same type, so they are always
--- a/crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
@@ -522,3 +522,17 @@ age, name = team.employees[0]
 reveal_type(age)  # revealed: Age
 reveal_type(name)  # revealed: Name
 ```
+
+## `~T` is never assignable to `T`
+
+```py
+from typing import TypeVar
+from ty_extensions import Not
+
+T = TypeVar("T")
+
+def f(x: T, y: Not[T]) -> T:
+    x = y  # error: [invalid-assignment]
+    y = x  # error: [invalid-assignment]
+    return x
+```
--- a/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
@@ -537,3 +537,14 @@ def _(x: int):
     # TODO: this should be `tuple[C, int]` as well, once we support implicit `self`
     reveal_type(C().implicit_self(x))  # revealed: tuple[Unknown, int]
 ```
+
+## `~T` is never assignable to `T`
+
+```py
+from ty_extensions import Not
+
+def f[T](x: T, y: Not[T]) -> T:
+    x = y  # error: [invalid-assignment]
+    y = x  # error: [invalid-assignment]
+    return x
+```


crates/ty_python_semantic/src/types.rs
```diff
@@ -1562,7 +1562,7 @@ impl<'db> Type<'db> {
             (Type::Intersection(intersection), Type::NonInferableTypeVar(_))
                 if intersection.negative(db).contains(&target) =>
             {
-                ConstraintSet::from(true)
+                ConstraintSet::from(false)
```
sharkdp, line 1565: I'm slightly surprised that we need these branches at all. Shouldn't the union/intersection branches take care of this, if `T <: T` is modeled correctly?

Also, is the statement "`~T` is never assignable to `T`" correct if `T` could specialize to `Any`? Or do we not have to take this case into account, for some reason?

AlexWaygood: > I'm slightly surprised that we need these branches at all. Shouldn't the union/intersection branches take care of this, if `T <: T` is modeled correctly?

Yes -- ideally we'd have the `NonInferrableTypeVar` branches below the `Union` and `Intersection` branches. Then we wouldn't need these branches at all, as you say. Unfortunately, however, value-constrained TypeVars make this quite hard, as they require special handling with regards to unions:

https://github.com/astral-sh/ruff/blob/3f640dacd42666f83d1a36d29efc507d31116060/crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md?plain=1#L227-L257

> Also, is the statement "`~T` is never assignable to `T`" correct if `T` could specialize to `Any`? Or do we not have to take this case into account, for some reason?

Hmm, interesting question. I think that also calls into doubt the validity of this pre-existing branch -- if the TypeVar could be solved to `Any`, it should not be considered a subtype of itself:

https://github.com/astral-sh/ruff/blob/3f640dacd42666f83d1a36d29efc507d31116060/crates/ty_python_semantic/src/types.rs#L1568-L1576

I wonder if this is something that @dcreager's work on constraint sets could help with? If the relation is `TypeRelation::Subtyping`, we could possibly return a constraint set from these branches that indicates that `self` is a subtype of `target` only if neither `self` nor `target` is solved to `Any`?

AlexWaygood: > I wonder if this is something that @dcreager's work on constraint sets could help with? If the relation is `TypeRelation::Subtyping`, we could possibly return a constraint set from these branches that indicates that `self` is a subtype of `target` only if neither `self` nor `target` is solved to `Any`?

But that doesn't really make sense, of course, because these are _non-inferable_ type variables -- type variables from the perspective of inside a function or class body... I think this does lead to the conclusion that `T` cannot be a subtype of `T` (where `T` is a non-inferable type variable); it can only be assignable to `T`. No matter what the bounds or constraints on a type variable are, it could _always_ be specialised to `Any`, so you can't really ever conclude that a non-inferable type variable is ever a subtype of (or disjoint from) anything, I don't think

carljm: I don't think we need to account for the possibility that a typevar is specialized to Any here. The key is that it's the same typevar, so it would have to be specialized to the _same_ Any in both usages -- which is really the same as saying it's specialized to some unknown type (which is already what we were accounting for). But since it's the _same_ type for both occurrences of T, the conclusions here still apply.

AlexWaygood: Right, I think that makes sense

AlexWaygood: so... the conclusion from Carl's comment is that the logic in this PR is correct?

dcreager: > I don't think we need to account for the possibility that a typevar is specialized to Any here. The key is that it's the same typevar, so it would have to be specialized to the _same_ Any in both usages

Right, I'd even reword this slightly to say that typevar _inference_ will never infer a gradual type for a typevar. (A typevar can be specialized to `Unknown` if we don't infer _anything_ for it, and it has no default, but I consider that different than explicitly infering a gradual type for it.)

> I'm slightly surprised that we need these branches at all. Shouldn't the union/intersection branches take care of this, if `T <: T` is modeled correctly?

The current way we're handling assignability of typevars is very brittle and dependent on these kinds of hacks, because of "rounding error" that's introduced by the early conversions of the assignability checks of union/intersection elements into a true/false value. https://github.com/astral-sh/ruff/pull/20093 is working on using constraint sets to model the assignability of a typevar, which will allow us to correctly model when e.g. each element of a union is assignable "sometimes", but in a way that combines to "always" when you merge them together. That will hopefully remove the need for these match arms completely. But having the mdtest here will be good to avoid a regression when that lands.

dcreager (approved):

PR #20608, opened 2025-09-28
[ty] Consolidate fields on `SemanticIndex`

AlexWaygood:
## Summary

This PR consolidates various fields on the `SemanticIndex` struct so that instead of having five `IndexVec` fields that are indexed by `FileScopeId`, we only have one.

The original motivation behind this PR was to see if this change would reduce memory usage at all. I'm unable to find any evidence that this makes any significant difference to ty's memory usage -- and this probably makes sense, because we only create one `SemanticIndex` struct for each file. However, I think this PR is probably worth doing anyway, as it looks like it leads to a 3% speedup on our `colour` benchmark, and speedups of 1-2% on all our other walltime benchmarks: https://codspeed.io/astral-sh/ruff/branches/alex%2Fsemantic-index-memory?runnerMode=WallTime. These reported speedups have been consistent across multiple runs on this PR, so I think they're real. I assume this PR speeds things up because we're now doing less work in `SemanticIndexBuilder::new()` (rather than having to allocate five `IndexVec`s, we now only need three for each builder), and `SemanticIndexBuilder::build()` (rather than having five different `.shrink_to_fit()` calls and three different `.collect()` calls, we now only need one of each). Conceptually, it's also quite nice to have all per-scope information consolidated in the `Scope` struct as well.

Best reviewed commit by commit.

## Test Plan

Existing tests


crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -101,10 +101,8 @@ pub(super) struct SemanticIndexBuilder<'db, 'ast> {
     semantic_checker: SemanticSyntaxChecker,
 
     // Semantic Index fields
-    scopes: IndexVec<FileScopeId, Scope>,
-    scope_ids_by_scope: IndexVec<FileScopeId, ScopeId<'db>>,
+    scopes: IndexVec<FileScopeId, ScopeBuilder<'db>>,
     place_tables: IndexVec<FileScopeId, PlaceTableBuilder>,
-    ast_ids: IndexVec<FileScopeId, AstIdsBuilder>,
     use_def_maps: IndexVec<FileScopeId, UseDefMapBuilder<'db>>,
```
dcreager, line 117: Is there something that prevents us from moving `place_tables` and `use_def_maps` into the new `ScopeBuilder` as well? (If there is, that deserves a comment here calling out why)

crates/ty_python_semantic/src/semantic_index/scope.rs
```diff
@@ -106,59 +111,161 @@ pub(crate) struct Scope {
...
-    }
-
-    pub(crate) fn node(&self) -> &NodeWithScopeKind {
+    pub(super) fn node(&self) -> &NodeWithScopeKind {
         &self.node
     }
 
-    pub(crate) fn kind(&self) -> ScopeKind {
-        self.node().scope_kind()
-    }
```
dcreager, line 138: optional nit: I think it's a minor code smell that callers now have to import `ScopeLike` to get access to some of the methods of `Scope`. You can get around that by keeping the definitions of `kind` and `parent` here in addition to the new definitions down in `ScopeLike`.

And you can change the definitions below to 

```rust
fn kind(&self) -> ScopeKind {
    Scope::kind(self)
}
```

to avoid duplicating the method body and without introducing an infinite loop

dcreager (approved):

crates/ty_python_semantic/src/semantic_index/builder.rs
AlexWaygood: It's _basically_ just that the borrow checker gets quite angry if you try to borrow the `scopes` `IndexVec` mutably and immutably at the same time -- having separate `IndexVec`s for `place_tables` and `use_def_maps` works around this, because both `place_tables` and `use_def_maps` frequently need to be borrowed mutably from `builder.rs`. There are ways of working around the borrow checker complaints if we store the use-def and place-table state on `ScopeBuilder` too, but from what I could tell they all undid the performance benefits of this change.

And yes, this is _definitely_ worth a comment -- thank you! I'll add one.

crates/ty_python_semantic/src/semantic_index/scope.rs
AlexWaygood: Yes, I was in two minds about this! I guess `ScopeLike` is really an implementation detail of the `semantic_index` submodule, so it's probably better to make it private and expose these methods publicly on `Scope`, as you suggest

AlexWaygood:
Hmm, well, after the latest rebase Codspeed is no longer claiming any speedup as a result of this, even though it pretty consistently said that there was one previously. So this PR now just seems to produce more code, for maybe not that much benefit :/ There is still this as a motivation for the change:

> Conceptually, it's also quite nice to have all per-scope information consolidated in the `Scope` struct as well.

But I don't know if it's a worthwhile change just on that basis.

MichaReiser:
My only worry with this is that it might get more challenging to satisfy the borrow checker in the future (similar to place table). But I don't feel stronlgy about this

ibraheemdev:
Did you run the detailed memory report (`TY_MEMORY_REPORT=full`) locally to see the diff? I'd be interested if this is a net-positive or net-negative to memory usage (even if minimal), because the struct of arrays representation should theoretically minimize padding, though I'm not sure if it's affected in this case because we already have an array of `Scope`s.

It can also be more cache-efficient when iterating over a single array, but the opposite can also be true if zipping over multiple.

AlexWaygood:
> Did you run the detailed memory report (`TY_MEMORY_REPORT=full`) locally to see the diff?

I ran with `TY_MEMORY_REPORT=mypy_primer` on a large project locally and couldn't see any difference in the topline numbers relative to `main`. I can run again with `TY_MEMORY_REPORT=full`

ibraheemdev:
The mypy-primer rounds the numbers to avoid excessive diffs (though maybe we should change that). Also make sure to run with `TY_MAX_PARALLELISM=1` for deterministic numbers locally.

dcreager:
If we're not seeing clear evidence of memory or CPU improvements, I would lean towards tabling this.

AlexWaygood:
Yeah, I think we've probably spent enough time looking at this now given our priorities and deadlines :-)

Thanks all!

PR #20612, opened 2025-09-28
[`flake8-annotations`] Fix return type annotations to handle shadowed builtin symbols (`ANN201`, `ANN202`, `ANN204`, `ANN205`, `ANN206`)

### LINKED ISSUE #20610: ANN2 fixes can add return types that are shadowed
### Summary

The fixes for [`missing-return-type-undocumented-public-function` (ANN201)](https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/), [`missing-return-type-private-function` (ANN202)](https://docs.astral.sh/ruff/rules/missing-return-type-private-function/), [`missing-return-type-special-method` (ANN204)](https://docs.astral.sh/ruff/rules/missing-return-type-special-method/), [`missing-return-type-static-method` (ANN205)](https://docs.astral.sh/ruff/rules/missing-return-type-static-method/), and [`missing-return-type-class-method` (ANN206)](https://docs.astral.sh/ruff/rules/missing-return-type-class-method/) add return types without checking for shadowed variables, which can cause type-checking errors. They should use [`get_or_import_builtin_symbol`](https://github.com/astral-sh/ruff/blob/0.13.2/crates/ruff_linter/src/importer/mod.rs#L285) or similar. [Example](https://play.ruff.rs/816d2f85-9113-4039-8009-35420b66fe07):
```console
$ cat >ann2.py <<'# EOF'
from collections import UserString as str
from typing import override
def foo(): return "!"
def _foo(): return "!"
class C:
    @override
    def __str__(self): return "!"
    @staticmethod
    def foo(): return "!"
    @classmethod
    def bar(cls): return "!"
# EOF

$ ruff --isolated check --select ANN201,ANN202,ANN204,ANN205,ANN206 ann2.py --unsafe-fixes --fix
Found 5 errors (5 fixed, 0 remaining).

$ mypy ann2.py
ann2.py:3: error: Incompatible return value type (got "str", expected "UserString")  [return-value]
ann2.py:4: error: Incompatible return value type (got "str", expected "UserString")  [return-value]
ann2.py:7: error: Return type "UserString" of "__str__" incompatible with return type "str" in supertype "builtins.object"  [override]
ann2.py:7: error: Incompatible return value type (got "str", expected "UserString")  [return-value]
ann2.py:9: error: Incompatible return value type (got "str", expected "UserString")  [return-value]
ann2.py:11: error: Incompatible return value type (got "str", expected "UserString")  [return-value]
Found 6 errors in 1 file (checked 1 source file)
```

### Version

danparizher:
## Summary

Fixes #20610

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
+++ b/crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
@@ -142,10 +142,7 @@ impl AutoPythonType {
                 });
                 Some((expr, vec![no_return_edit]))
             }
-            AutoPythonType::Atom(python_type) => {
-                let (expr, edits) = type_expr(python_type, checker, at)?;
-                Some((expr, edits))
-            }
+            AutoPythonType::Atom(python_type) => type_expr(python_type, checker, at),
             AutoPythonType::Union(python_types) => {
                 if target_version >= PythonVersion::PY310 {
                     // Aggregate all the individual types (e.g., `int`, `float`).
@@ -214,38 +211,14 @@ impl AutoPythonType {
 ///
 /// If the [`Expr`] relies on importing any external symbols, those imports will be returned as
 /// additional edits.
-fn type_expr(
+pub(crate) fn type_expr(
     python_type: PythonType,
     checker: &Checker,
     at: TextSize,
 ) -> Option<(Expr, Vec<Edit>)> {
     match python_type {
-        PythonType::String => {
-            let (edit, binding) = checker
-                .importer()
-                .get_or_import_builtin_symbol("str", at, checker.semantic())
-                .ok()?;
-            let expr = Expr::Name(ast::ExprName {
-                id: binding.into(),
-                range: TextRange::default(),
-                node_index: ruff_python_ast::AtomicNodeIndex::NONE,
-                ctx: ExprContext::Load,
-            });
-            Some((expr, edit.map_or_else(Vec::new, |edit| vec![edit])))
-        }
-        PythonType::Bytes => {
-            let (edit, binding) = checker
-                .importer()
-                .get_or_import_builtin_symbol("bytes", at, checker.semantic())
-                .ok()?;
-            let expr = Expr::Name(ast::ExprName {
-                id: binding.into(),
-                range: TextRange::default(),
-                node_index: ruff_python_ast::AtomicNodeIndex::NONE,
-                ctx: ExprContext::Load,
-            });
-            Some((expr, edit.map_or_else(Vec::new, |edit| vec![edit])))
-        }
+        PythonType::String => create_name_expr("str", checker, at),
+        PythonType::Bytes => create_name_expr("bytes", checker, at),
         PythonType::Number(number) => {
             let symbol = match number {
                 NumberLike::Integer => "int",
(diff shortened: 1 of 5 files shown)


crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
```diff
@@ -199,26 +211,73 @@ impl AutoPythonType {
...
+                range: TextRange::default(),
+                node_index: ruff_python_ast::AtomicNodeIndex::NONE,
+                ctx: ExprContext::Load,
+            });
+            Some((expr, edit.map_or_else(Vec::new, |edit| vec![edit])))
+        }
+        PythonType::None => {
+            let (edit, binding) = checker
+                .importer()
+                .get_or_import_builtin_symbol("None", at, checker.semantic())
```
dscorbett: `None` is a keyword: it shouldn’t be imported.

crates/ruff_linter/src/rules/flake8_annotations/snapshots/ruff_linter__rules__flake8_annotations__tests__shadowed_builtins.snap
```diff
@@ -0,0 +1,117 @@
...
+12 |     def __str__(self):
+   |         ^^^^^^^
+13 |         return "!"
+   |
+help: Add return type annotation: `str`
+9  | 
+10 | class C:
+11 |     @override
+   -     def __str__(self):
+12 +     def __str__(self) -> str:
```
ntBre: This one should also be `builtins.str` right?

crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
```diff
@@ -143,22 +143,27 @@ impl AutoPythonType {
                 Some((expr, vec![no_return_edit]))
             }
             AutoPythonType::Atom(python_type) => {
-                let expr = type_expr(python_type)?;
-                Some((expr, vec![]))
+                let (expr, edits) = type_expr(python_type, checker, at)?;
+                Some((expr, edits))
```
ntBre: nit: I think this is the same


```suggestion
                type_expr(python_type, checker, at)
```

crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
```diff
@@ -199,26 +211,69 @@ impl AutoPythonType {
...
-            NumberLike::Complex => Some(name("complex")),
-            NumberLike::Bool => Some(name("bool")),
-        },
-        PythonType::None => Some(name("None")),
+        PythonType::String => {
+            let (edit, binding) = checker
+                .importer()
+                .get_or_import_builtin_symbol("str", at, checker.semantic())
+                .ok()?;
+            let expr = Expr::Name(ast::ExprName {
```
ntBre: Can we still preserve the original `name` helper? Possibly with a generic `impl Into<Name>` argument instead of `&str`? That seemed to help with some of the repetition here. It looks like we could possibly even include the `get_or_import_builtin` call in the helper.

crates/ruff_linter/src/rules/flake8_annotations/helpers.rs
```diff
@@ -199,26 +211,69 @@ impl AutoPythonType {
...
+            });
+            Some((expr, edit.map_or_else(Vec::new, |edit| vec![edit])))
+        }
+        PythonType::None => {
+            let expr = Expr::Name(ast::ExprName {
+                id: "None".into(),
+                range: TextRange::default(),
+                node_index: ruff_python_ast::AtomicNodeIndex::NONE,
+                ctx: ExprContext::Load,
+            });
```
ntBre: I don't think this should change anything in the output, but for convenience we should be able to return an `Expr::NoneLiteral(ExprNoneLiteral::default())` here.

crates/ruff_linter/src/rules/flake8_annotations/rules/definition.rs
```diff
@@ -825,11 +826,31 @@ pub(crate) fn definition(
...
                     function.identifier(),
                 );
-                if let Some(return_type) = return_type {
-                    diagnostic.set_fix(Fix::unsafe_edit(Edit::insertion(
-                        format!(" -> {return_type}"),
-                        function.parameters.end(),
-                    )));
+                if let Some(return_type_str) = return_type {
+                    // Convert the simple return type to a proper expression that handles shadowed builtins
+                    let python_type = match return_type_str {
```
ntBre, line 831: This is a bit awkward because we just did basically the same `match` inside of `simple_magic_return_type`. I tried moving `simple_magic_return_type` into this file and making it return a `PythonType` directly, though, and then we don't have a nice way to get back to the string representation. So I think this is fine as-is.

ntBre (approved):
Thank you! I pushed one commit moving the `name` helper back into `type_expr` just to keep it more consistent with the old code. I'm also not sure what I was thinking with the `impl Into<Name>` suggestion, so I just put that back to `&str` too 😅

Issue #12800, opened 2024-08-11
[feature] Add ability to specify `--range` option multiple times


afontenot:
This feature would increase feature parity with `black`, which supports multiply specifying the equivalent `--line-ranges` option.

Accepting multiple ranges is extremely useful for programmatically checking / modifying the format of the *changed* portions of a file when changes occur. For example, on some code bases a style guide has not been correctly adhered to historically, but project maintainers want to ensure that all *new* work adheres to `ruff format` checks.

Some discussion of this happened in the original `--range` pull request https://github.com/astral-sh/ruff/pull/9733, which I'll copy here:

> The CLI only supports a single range. We can explore supporting multiple ranges in the future. The main challenge is that overlapping ranges invalidate the offset of whichever range gets formatted last (starting from the back helps but doesn't prevent it). This is especially a problem if the formatter has to extend the formatted range.

While I do agree that this adds some complexity, I think it's a valuable feature and I didn't see an open issue for it, so I decided to file one to hopefully push this forward.

MichaReiser:
Thanks for opening this issue and doing the extra work of finding the PR where it was added. 

Could you tell us a bit more about your workflow? It seems you're already calling ruff in an automated way.

afontenot:
> Could you tell us a bit more about your workflow?

Sure! I don't have this in use on any large projects (yet), but I work on some Python projects that don't have an established code style currently. I've written a small program that calls `git` to extract the precise line numbers that have been changed, and turns these into ranges that can be passed into either `black` or `ruff format`.

To make sure my contributions to these projects follow a consistent style (I use the black style on all my personal stuff), I set up a pre-commit hook that runs this tool in a `--check` form. I can also pass `--diff` to it to see code style changes suggested by `black` or `ruff` or just accept the suggested changes. This can't be done without iteratively changing one section of code at a time in `ruff`. I think project maintainers who want to make sure PRs pass a style check without reformatting their whole code base might be interested in adding a tool like this to their test suite.

Some editors allow you to pass selections to the formatter, so that's a partial workaround for this issue. However, this doesn't allow a pre-commit check, and the editor I'm currently using the most (Zed) does not support this.

dhruvmanila:
This will also be useful in implementing the upcoming (3.18) LSP feature `textDocument/rangesFormatting` which allow providing multiple ranges for a formatting request (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_rangeFormatting).

PR #20614, opened 2025-09-28
[`ruff`] Ability to specify `--range` multiple times.

cristian64:
## Summary

Line ranges that overlap (or are adjacent) will be joined. Formatting of each range is performed starting from the bottom of the file, ensuring that the following range still applies to the lines that the user intended to format.

Implements #12800.

## Test Plan

- Pre-existing unit tests ensure no regression or unexpected behavior change.
- New unit tests have been added to cover the feature enhancement.


crates/ruff/src/commands/format.rs
```diff
@@ -270,20 +270,38 @@ pub(crate) fn format_path(
...
-        FormattedSource::Formatted(formatted) => match mode {
-            FormatMode::Write => {
-                let mut writer = File::create(path).map_err(|err| {
-                    FormatCommandError::Write(Some(path.to_path_buf()), err.into())
-                })?;
-                formatted
-                    .write(&mut writer)
-                    .map_err(|err| FormatCommandError::Write(Some(path.to_path_buf()), err))?;
+    let format_result =
+        match format_source(&unformatted, source_type, Some(path), settings, ranges)? {
```
cristian64, line 298: Note for the reviewer: the only change in this block is `range` -> `ranges`; hiding whitespace changes in the diff viewer may improve readability of the diff.

crates/ruff/src/commands/format.rs
```diff
@@ -342,13 +342,14 @@ pub(crate) fn format_source(
...
-    range: Option<FormatRange>,
+    ranges: &[FormatRange],
 ) -> Result<FormattedSource, FormatCommandError> {
     match &source_kind {
         SourceKind::Python(unformatted) => {
             let options = settings.to_format_options(source_type, unformatted, path);
 
-            let formatted = if let Some(range) = range {
+            let formatted = if ranges.len() == 1 {
+                let range = &ranges[0];
```
MichaReiser, line 372: It should be fine to use the same code path for one and multiple ranges. The overhead should be minimal compared to the formatting itself.

MichaReiser (changes requested):
Thank you for looking into this. 

Unfortunately, I don't think the way the ranges are collapsed is correct and this, instead, requires deeper integration into `format_range`. 

The issue is that `format_range` can expand the formatting range. E.g. for:

```py
a = (1 + 
	3 + 2)
```

A range `2:3` will be expanded to `1:3` because Ruff can only format entire statements (or clause headers), not sub-expressions.

cristian64:
> Thank you for looking into this.
> 
> Unfortunately, I don't think the way the ranges are collapsed is correct and this, instead, requires deeper integration into `format_range`.
> 
> The issue is that `format_range` can expand the formatting range. E.g. for:
> 
> ```python
> a = (1 + 
> 	3 + 2)
> ```
> 
> A range `2:3` will be expanded to `1:3` because Ruff can only format entire statements (or clause headers), not sub-expressions.

I'm trying to digest why this would be a problem.

If the user provides ranges that overlap with these expanded ranges (which I think can be considered an edge case), the only drawback would be that Ruff would perform a redundant format operation, but the end result would be correct.

Would you have a more concrete example where things can go wrong? (And would we be happy to document the case and/or file a separate issue for it, while providing a solution for the general case?)

MichaReiser:
> Formatting of each range is performed starting from the bottom of the file, ensuring that the following range still applies to the lines that the user intended to format.

I think the assumption that ranges will not be invalidated because they're format from back to front is no longer correct if a range can be extended forward.

I think the right approach is:

* Expand every range and collapse overlapping ranges
* Find the minimal range that covers all ranges
* Expand that minimal range too
* Format the minimal range (avoids re-parsing the same file multiple times)
* Now replace the text of the expanded (and collapsed) ranges)

crates/ruff/src/commands/format.rs
cristian64: I'm not that familiar with Rust, and the use of `mut` wasn't too comforting, so I went with the option that produced a smaller diff for the sake of the reviewer.

Happy to collapse the two cases if we are okay with the solution, pending discussion of the other point you raised.

cristian64:
> > Formatting of each range is performed starting from the bottom of the file, ensuring that the following range still applies to the lines that the user intended to format.
> 
> I think the assumption that ranges will not be invalidated because they're format from back to front is no longer correct if a range can be extended forward.

If I understand correctly, the range would be invalidated, but it would sit on top of code that has just been formatted, so the format operation would be no-op, with no other side effect. (At least that was how I reasoned about this.)

MichaReiser:
I don't think this assumption is correct because you replace `unformatted` with the newly formatted content, invalidating any prevoius source range. 

What's the reaosnf or updating the unformatted code. Can't we keep using the same unformatted code and simply format different slices (in which case we should avoid reparsing the AST over and over again)

cristian64:
> I don't think this assumption is correct because you replace `unformatted` with the newly formatted content, invalidating any prevoius source range.

They would be invalidated if a range was expanded, but even then that shouldn't be a problem as the invalidated range would now sit on code that has been formatted, and a second format operation is no-op.

> What's the reaosnf or updating the unformatted code. Can't we keep using the same unformatted code and simply format different slices (in which case we should avoid reparsing the AST over and over again)

The unformatted code needs to be updated due to the potential existence of expanded ranges. I see how performance-wise it'd be better to implement this at a lower level, but it's also less trivial.

MichaReiser:
> They would be invalidated if a range was expanded, but even then that shouldn't be a problem as the invalidated range would now sit on code that has been formatted, and a second format operation is no-op.

The range could now point between Unicode characters which results in a panic. Or the range now points passed the end of the source file which is an error.

> The unformatted code needs to be updated due to the potential existence of expanded ranges. I see how performance-wise it'd be better to implement this at a lower level, but it's also less trivial.

That's fair. However, performance is a crucial aspect of our tooling.

cristian64:
> The range could now point between Unicode characters which results in a panic. 

I thought the numbers in the range were to be treated as unicode units, meaning that this shouldn't be possible.

> Or the range now points passed the end of the source file which is an error.

This case should not be possible. The range would have been joined with the previous range, or the formatting of the previous range should have hit the error earlier. (Doing it from back to front guarantees correctness.)

MichaReiser:
Oh right, you convert from `TextRange` to `FormatRange` in each iteration. This might also protect against going past the file end. 

> This case should not be possible. The range would have been joined with the previous range, or the formatting of the previous range should have hit the error earlier. (Doing it from back to front guarantees correctness.)

It's possible on the byte range level but I suspect that `to_range` protects against going past the file range?

E.g. 

```py
a = (1 + 
    3 + 2)
b = (x)
```

with 2:5 to 2:6 (3) and 2:11 to 2:8

The ranges don't overlap, so they aren't merged. Formatting 2:7 to 2:8 results in 

```
a = 1 + 3 + 2
b = (x)
```

The range 2:5 to 2:6 now incorrectly selects the `(` of `b = (x)`

cristian64:
I think there is a typo in `2:11` (should be `2:7`), but I see how it can lead to the formatting of unsolicited code blocks.

Joining of overlapping ranges must indeed be done after ranges have been potentially expanded.

cristian64:
I've added a unit test to cover the discussed case (which should be failing now). I'll see if I can look into a more robust implementation in the upcoming days.

cristian64:
I've made some progress on this. The new implementation collapses ranges _after_ they have been expanded. The unit test that was added earlier in the week is now passing, and another unit test with ranges that touch on various statements has been added.

Performance wise there should be room for improvement (avoid copies of the text; avoid parsing more than once), but the deeper in the code the harder it gets.

@MichaReiser How do you feel about this? Would this be good enough for a first iteration? Would you like to take it from here?

MichaReiser:
> @MichaReiser How do you feel about this? Would this be good enough for a first iteration? Would you like to take it from here?

Nice. I'll try to take a look at this sometime soon but it probably won't be in the next two weeks. I'm stretched extremely thin as we're working towards the ty beta. I'm sorry. Please ping me if I forget to get back to this.

crates/ruff_python_formatter/src/range.rs
```diff
@@ -117,6 +117,83 @@ pub fn format_range(
...
+        if *range == TextRange::up_to(source.text_len()) {
+            return Ok(vec![*range]);
+        }
+    }
+
+    let parsed = parse(source, ParseOptions::from(options.source_type()))?;
+    let source_code = SourceCode::new(source);
+    let comment_ranges = CommentRanges::from(parsed.tokens());
+    let comments = Comments::from_ast(parsed.syntax(), source_code, &comment_ranges);
+
```
MichaReiser, line 146: We should avoid parsing and extracting the comments here. Both are very expensive operations and we only want to perform them exactly one per file

crates/ruff_python_formatter/src/range.rs
```diff
@@ -117,6 +117,83 @@ pub fn format_range(
...
+    let mut collapsed_ranges: Vec<TextRange> = Vec::new();
+    collapsed_ranges.push(expanded_ranges[0]);
+    for range in expanded_ranges {
+        let back_range_index = collapsed_ranges.len() - 1;
+        let back_range = &collapsed_ranges[back_range_index];
+        if back_range.overlap_or_touch(&range) {
+            collapsed_ranges[back_range_index] = back_range.joined(&range);
+        } else {
+            collapsed_ranges.push(range);
+        }
```
MichaReiser, line 191: I'm not sure if simply joining here is correct. `covering_node` handles a few special cases, e.g. only formatting the clause header and `slice_range` assumes on some of those assumptions too (I think). 

So this might be a bit trickier than just joining ranges together but I'd have to take a closer look myself.

crates/ruff/src/commands/format.rs
MichaReiser: I'd very much prefer to share as much of the code because we already have very low test coverage for range formatting.

cristian64: I was leaving the removal to the end, to keep the diff a bit easier to read until we would be happy with the rest of the changes. Would you rather like this removed _now_?

crates/ruff_python_formatter/src/range.rs
cristian64: Perhaps I'm unable to see how this can be skipped. It seems the other functions require this context to be provided.

PR #20629, opened 2025-09-29
[ty] Ensure first-party search paths always appear in a sensible order

AlexWaygood:
This PR ensures that we always put `./src` before `.` in our list of first-party search paths. This better emulates the fact that at runtime, the module name of a file `src/foo.py` would almost certainly be `foo` rather than `src.foo`.

I wondered if fixing this might fix https://github.com/astral-sh/ruff/pull/20603#issuecomment-3345317444. It seems like that's not the case, but it also seems like it leads to better diagnostics because we report much more intuitive module names to the user in our error messages -- so, it's probably a good change anyway.

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/resolver.rs
+++ b/crates/ty_python_semantic/src/module_resolver/resolver.rs
@@ -9,6 +9,7 @@ specifies ty's implementation of Python's import resolution algorithm.
 */
 
 use std::borrow::Cow;
+use std::cmp::Ordering;
 use std::fmt;
 use std::iter::FusedIterator;
 use std::str::Split;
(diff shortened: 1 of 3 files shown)


AlexWaygood:
> I suspect this is the cause of [#20603 (comment)](https://github.com/astral-sh/ruff/pull/20603#issuecomment-3345317444)

Okay, well, maybe not, but it at least leads to better diagnostics 🙃

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -244,9 +245,25 @@ impl SearchPaths {
             static_paths.push(SearchPath::extra(system, path)?);
         }
 
+        // Most search paths come from user configuration, so we should just trust that the user
+        // has specified their search paths in the correct order -- if not, that's on the user.
+        // First-party search paths can be implicitly inferred by ty, however (`./src` is a common case),
+        // so for first-party search paths it makes sense to sort them into a sensible order.
+        // We make sure here that child directories always take precedence over parent directories.
```
MichaReiser: Should we move this to where we add the default search paths. I feel like we shouldn't apply this sorting when it's a user-provided list (which you're also arguing in favor of?)

AlexWaygood: Ah, we only add the implicit `src` root if there was no explicit user configuration? That makes sense; I forgot how this worked.

MichaReiser (approved):

Issue #17331, opened 2025-04-10
Ruff formatter ignoring `fmt: skip`


madduck:
### Summary

Taking this over from #8415:

I have a little snippet that inserts `ipdb` into code, and I want to keep it on a single line, so that I can easily remove it again with Vim `dd`, here is the test-case:

```Python
def main() -> None:
    import ipdb; ipdb.set_trace()  # noqa: E402,E702,I001 fmt: skip
```

This passes ruff's check alright:

```
% ruff --isolated check test.py --fix 
All checks passed!
```

But when I run the formatter on it, it seems that `fmt: skip` is being ignored:

```
% ruff --isolated format test.py --diff
--- test.py
+++ test.py
@@ -1,2 +1,4 @@
 def main() -> None:
-    import ipdb; ipdb.set_trace()  # noqa: E402 E702 I001 # fmt: skip
+    import ipdb
+
+    ipdb.set_trace()  # noqa: E402 E702 I001 # fmt: skip

1 file would be reformatted
```
I have tried `# fmt: skip` before and after the `noqa` filter, I have tried two spaces before the `#`, and I have tried without `noqa`, i.e. just `# fmt: skip`, with and without a space after the colon — the same result for each of these permutations.

If I enclose the line in `fmt:off` and `fmt:on` statements, then the formatter finds nothing to do, but that defies the purpose of the single-line import, and the per-line `fmt:skip` instruction.

### Version

0.11.4

MichaReiser:
Thanks. I thought this is the same as https://github.com/astral-sh/ruff/issues/11216 but it's only related.

madduck:
Is there anything I can do to expedite addressing of this bug?

MichaReiser:
No. I don't think so. We have all the information we need. It's just that someone needs to find time to fix this.

madduck:
> No. I don't think so. We have all the information we need. It's just that someone needs to find time to fix this.

Ok. I don't know Rust too well, nor do I have ample free time, but if you pointed me in the right direction, I could take a look…?

MichaReiser:
That's kind. I expect this fix to be somewhat involved. The best I can point you towards is https://github.com/astral-sh/ruff/blob/8c0743df97dc7afa1deab7506b322551f2e392bb/crates/ruff_python_formatter/src/comments/mod.rs#L501-L509. Then search where it is used

MichaReiser:
I tried to implement this with claude but it struggled big times. But the tests are useful:

```py
[...]

Issue #11430, opened 2024-05-15
`# noqa: E702 # fmt: skip` still formats the line of code


ngoctnq:
Input code:
```python
print(); print() # noqa # fmt: skip
```

After formatting:
```python
print()
print()  # noqa # fmt: skip
```

`# fmt: on/off` works fine though, but I want to have just one line of short boilerplate code.

I'm using the Ruff VS Code extension, which is said to ship with `ruff==0.4.1`

dhruvmanila:
I think it's because there are two statements on the same line and `fmt: skip` is being applied only for the last statement. This can be seen by adding a single-quoted string (https://play.ruff.rs/edf37109-eb34-4b5c-9867-dd5587bef085), notice that the formatter only changes the quote for the first print call and not the second one. I don't think `RUF028` will indicate here because the suppression comment is actually being used by the formatter.

I'm not exactly sure if this is the intended behavior, @konstin any thoughts here? @MichaReiser  will be able to provide more context here but he's currently on PTO until the end of next week. I can possibly look into it to confirm whether this is the intended behavior.

MichaReiser:
Oh, that's interesting and not something I have thought of when implementing suppression comments. 

> # fmt: skip comments suppress formatting for a preceding statement, case header, decorator, function definition, or class definition:

Today's behavior matches our documentation. However, Black doesn't format simple statements that end with a `fmt: skip` comment. So it is at least a Black incompatibility (and possibly not flagged by the ruff lint rule).

I'm open to changing our implementation to support this use case if it doesn't add too much complexity. I otherwise recommend documenting the deviation because it's possible to suppress the example using `fmt: off` (although it is slightly more verbose).

```python
# fmt: off
print(); print() # noqa 
# fmt: on

print( )
```

Gatsby-Lee:
> Oh, that's interesting and not something I have thought of when implementing suppression comments.
> 
> > `# fmt: skip` comments suppress formatting for a preceding statement, case header, decorator, function definition, or class definition:
> 
> Today's behavior matches our documentation. However, Black doesn't format simple statements that end with a `fmt: skip` comment. So it is at least a Black incompatibility (and possibly not flagged by the ruff lint rule).
> 
> I'm open to changing our implementation to support this use case if it doesn't add too much complexity. I otherwise recommend documenting the deviation because it's possible to suppress the example using `fmt: off` (although it is slightly more verbose).
> 
> ```python
> # fmt: off
> print(); print() # noqa 
[...]

PR #20633, opened 2025-09-29
Respect `fmt: skip` for compound statements on single line

### LINKED ISSUE #11216: [ENH] Let `fmt: skip` apply to compound statements, if they share a line.
```python
if True: x = 0  # fmt: skip

def foo(self, lorem, ipsum, dolor, sit, amet, consectetur, adipiscing, elit, sed, do): ...  # fmt: skip
```

Gets reformatted to

```python
if True:
    x = 0  # fmt: skip

def foo(
    self, lorem, ipsum, dolor, sit, amet, consectetur, adipiscing, elit, sed, do
): ...  # fmt: skip
```

Which is because `fmt: skip` only applies to the preceding `statement`. However, it would be useful to apply the `fmt: skip` to the whole [compound statement](https://docs.python.org/3/reference/compound_stmts.html) in this context. For instance, when formatting `@overload` decorated functions.

dylwil3:
Closes #11216

Essentially the approach is to implement `Format` for a new struct `FormatClause` which is just a clause header _and_ its body. We then have the information we need to see whether there is a skip suppression comment on the last child in the body and it all fits on one line.

The `ruff` test fixture (credit: mostly created by Claude) contains several lines that seem to crash `black`, see:

- https://github.com/psf/black/issues/4782
- https://github.com/psf/black/issues/4781

and another line containing a deviation that I _think_ is unintentional for `black`, see: 
- https://github.com/psf/black/issues/4783

However, it's not clear whether we want the same behavior here? In this PR we have:

```python
if True: print("this"); print("that") # fmt: skip
```

unchanged, but we _do_ add a newline between the statements here (unlike Black):

```python
print("this"); print("that") # fmt: skip
```

(see https://github.com/astral-sh/ruff/issues/11430 and https://github.com/astral-sh/ruff/issues/17331 .)

The other deviation from `black` is that we format the placement of the skip comment itself.

For review, it is simplest to go commit by commit.

### DIFF
--- a/crates/ruff_python_formatter/src/statement/clause.rs
+++ b/crates/ruff_python_formatter/src/statement/clause.rs
@@ -583,10 +583,13 @@ where
 
 impl<'ast> Format<PyFormatContext<'ast>> for FormatClause<'_, 'ast> {
     fn fmt(&self, f: &mut Formatter<PyFormatContext<'ast>>) -> FormatResult<()> {
-        if should_suppress_clause(self, f)? {
-            write_suppressed_clause(self, f)
-        } else {
-            write!(f, [self.clause_header(), self.clause_body()])
+        match should_suppress_clause(self, f)? {
+            SuppressClauseHeader::Yes {
+                last_child_in_clause,
+            } => write_suppressed_clause(self, f, last_child_in_clause),
+            SuppressClauseHeader::No => {
+                write!(f, [self.clause_header(), self.clause_body()])
+            }
         }
     }
 }
@@ -714,14 +717,14 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
     }
 }
 
-fn should_suppress_clause(
-    clause: &FormatClause<'_, '_>,
+fn should_suppress_clause<'a>(
+    clause: &FormatClause<'a, '_>,
     f: &mut Formatter<PyFormatContext<'_>>,
-) -> FormatResult<bool> {
+) -> FormatResult<SuppressClauseHeader<'a>> {
     let source = f.context().source();
 
     let Some(last_child_in_clause) = clause.header.last_child_in_clause() else {
-        return Ok(false);
+        return Ok(SuppressClauseHeader::No);
     };
 
     // Early return if we don't have a skip comment
@@ -730,7 +733,7 @@ fn should_suppress_clause(
         f.context().comments().trailing(last_child_in_clause),
         source,
     ) {
-        return Ok(false);
+        return Ok(SuppressClauseHeader::No);
     }
 
     let clause_start = clause.header.range(source)?.end();
@@ -739,16 +742,19 @@ fn should_suppress_clause(
 
     // Only applies to clauses on a single line
     if source.contains_line_break(clause_range) {
-        return Ok(false);
+        return Ok(SuppressClauseHeader::No);
     }
 
-    Ok(true)
+    Ok(SuppressClauseHeader::Yes {
+        last_child_in_clause,
+    })
 }
 
 #[cold]
 fn write_suppressed_clause(
     clause: &FormatClause,
     f: &mut Formatter<PyFormatContext<'_>>,
+    last_child_in_clause: AnyNodeRef,
 ) -> FormatResult<()> {
     if let Some((leading_comments, last_node)) = clause.leading_comments {
         leading_alternate_branch_comments(leading_comments, last_node).fmt(f)?;
@@ -759,10 +765,7 @@ fn write_suppressed_clause(
 
     let comments = f.context().comments().clone();
 
-    let last_child = header
-        .last_child_in_clause()
-        .expect("last child to exist if `should_suppress_clause` is `Ok(true)`");
-    let clause_end = last_child.end();
+    let clause_end = last_child_in_clause.end();
 
     // Write the outer comments and format the node as verbatim
     write!(
@@ -771,7 +774,7 @@ fn write_suppressed_clause(
             source_position(clause_start),
             verbatim_text(TextRange::new(clause_start, clause_end)),
             source_position(clause_end),
-            trailing_comments(comments.trailing(last_child)),
+            trailing_comments(comments.trailing(last_child_in_clause)),
             hard_line_break()
         ]
     )?;
@@ -798,4 +801,11 @@ fn write_suppressed_clause(
     }
 
     Ok(())
+}
+
+enum SuppressClauseHeader<'a> {
+    No,
+    Yes {
+        last_child_in_clause: AnyNodeRef<'a>,
+    },
 }


crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -385,13 +462,13 @@ impl<'ast> Format<PyFormatContext<'ast>> for FormatClauseHeader<'_, 'ast> {
...
 
-pub(crate) struct FormatClauseBody<'a> {
+struct FormatClauseBody<'a> {
     body: &'a Suite,
     kind: SuiteKind,
     trailing_comments: &'a [SourceComment],
 }
 
-pub(crate) fn clause_body<'a>(
+fn clause_body<'a>(
```
dylwil3, line 483: We no longer use `clause_body`, but we do use `clause_header` alone in one place - the formatting of `match` statements. Unlike match _cases_, I don't think it is legal to have an entire match statement lie on a single line, so it's okay that we don't handle that here.

MichaReiser:
> unchanged, but we do add a newline between the statements here (unlike Black):
> ```py
> print("this"); print("that") # fmt: skip
> ```

This seems unrelated to this PR as it is a pre-existing issue, right?

crates/ruff_python_formatter/src/preview.rs
```diff
@@ -27,3 +27,8 @@ pub(crate) const fn is_blank_line_before_decorated_class_in_stub_enabled(
 ) -> bool {
     context.is_preview()
 }
+
+/// Returns `true` if the [`fix_fmt_skip_in_one_liners`]() preview style is enabled.
+pub(crate) const fn is_fix_fmt_skip_in_one_liners_enabled(context: &PyFormatContext) -> bool {
+    context.is_preview()
+}
```
MichaReiser: Is the reason for gating this behind preview that we're concerned about bugs or are there cases where it can change existing formatted code?

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -36,7 +38,72 @@ pub(crate) enum ClauseHeader<'a> {
     OrElse(ElseClause<'a>),
 }
 
-impl ClauseHeader<'_> {
+impl<'a> From<ClauseHeader<'a>> for AnyNodeRef<'a> {
```
MichaReiser: Nit: I prefer to have those "helper" trait implementations after the types own `impl` block as the latter seems more important to the type

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -36,7 +38,72 @@ pub(crate) enum ClauseHeader<'a> {
...
+}
+
+impl<'a> ClauseHeader<'a> {
+    /// Returns the last child in the clause body immediately following this clause header.
+    ///
+    /// For most clauses, this is the last statement in
+    /// the primary body. For clauses like `try`, it specifically returns the last child
+    /// in the `try` body, not the `except`/`else`/`finally` clauses.
+    ///
+    /// Compare [`ruff_python_ast::AnyNodeRef::last_child_in_body`].
```
MichaReiser: I find this comment slightly confusing. Are you asking me to compare with `last_child_in_body` and spot the differences? 😆 

I assume this is the same as `last_child_in_body` but specific and limited to clause headers

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -36,7 +38,72 @@ pub(crate) enum ClauseHeader<'a> {
...
+                body.last().map(AnyNodeRef::from)
+            }
+            ClauseHeader::If(StmtIf { body, .. }) => body.last().map(AnyNodeRef::from),
+            ClauseHeader::ElifElse(ElifElseClause { body, .. }) => {
+                body.last().map(AnyNodeRef::from)
+            }
+            ClauseHeader::Try(StmtTry { body, .. }) => body.last().map(AnyNodeRef::from),
+            ClauseHeader::ExceptHandler(ExceptHandlerExceptHandler { body, .. }) => {
+                body.last().map(AnyNodeRef::from)
+            }
```
MichaReiser: Nit, similar to `last_child_in_body`
```suggestion
            ClauseHeader::Class(StmtClassDef { body, .. }) 
            | ClauseHeader::Function(StmtFunctionDef { body, .. }) 
            | ClauseHeader::If(StmtIf { body, .. }) 
            | ClauseHeader::ElifElse(ElifElseClause { body, .. }) 
            | ClauseHeader::Try(StmtTry { body, .. }) 
            | 
            ClauseHeader::ExceptHandler(ExceptHandlerExceptHandler { body, .. }) => {
                body.last().map(AnyNodeRef::from)
            }
```

Same for other branches that are over `&[Stmt]` and return the last element. It makes it easier for readers to see which branches are the same

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -502,3 +632,73 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
...
+
+    let clause_start = clause
+        .format_header
+        .header
+        .first_keyword_range(source)?
+        .start();
+
+    let clause_range = TextRange::new(clause_start, last_child_in_clause.end());
+
+    // Only applies to clauses on a single line
```
MichaReiser, line 743: I was confused for a second because I wondered: what about multiline clause headers that are suppressed by a `fmt: skip`? But that's handled by `clause_header`.

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -502,3 +632,73 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
...
+    let header = clause.format_header.header;
+
+    let range_start = header.first_keyword_range(f.context().source())?.start();
+
+    let comments = f.context().comments().clone();
+
+    let last_child = header
+        .last_child_in_clause()
+        .expect("last child to exist if `should_suppress_clause` is `Ok(true)`");
+    let range_end = last_child.end();
```
MichaReiser:
```suggestion
    let clause_start = header.first_keyword_range(f.context().source())?.start();

    let comments = f.context().comments().clone();

    let last_child = header
        .last_child_in_clause()
        .expect("last child to exist if `should_suppress_clause` is `Ok(true)`");
    let clause_end = last_child.end();
```

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -502,3 +632,73 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
...
+            source_position(range_start),
+            verbatim_text(TextRange::new(range_start, range_end)),
+            source_position(range_end),
+            trailing_comments(comments.trailing(last_child)),
+            hard_line_break()
+        ]
+    )?;
+
+    // Mark the comments for the body statements as formatted, but not the entire header node
+    // (which might include other clauses like except/else/finally for try statements).
```
MichaReiser: I think this is correct but an explanation why this is correct might be useful. 

* What happens with the leading comments of the clause headers?
* What about dangling comments (I think they're fine because a clause header with dangling comments is always guaranteed to be multiline
* What about trailing clause comments?

crates/ruff_python_formatter/src/statement/stmt_class_def.rs
```diff
@@ -132,8 +132,10 @@ impl FormatNodeRule<StmtClassDef> for FormatStmtClassDef {
 
                         Ok(())
                     }),
+                    body,
+                    SuiteKind::Class,
```
MichaReiser, line 136: Can we derive the `SuiteKind` from the `ClauseHeader` to reduce the number of arguments?

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -434,6 +511,59 @@ impl Format<PyFormatContext<'_>> for FormatClauseBody<'_> {
...
+/// Formats a clause, handling the case where the compound
+/// statement lies on a single line with `# fmt: skip` and
+/// should be suppressed.
+pub(crate) fn clause<'a, 'ast, Content>(
+    header: ClauseHeader<'a>,
+    trailing_colon_comment: &'a [SourceComment],
+    header_formatter: &'a Content,
+    body: &'a Suite,
+    kind: SuiteKind,
+    trailing_comments: &'a [SourceComment],
```
MichaReiser: It seems that `trailing_colon_comment` and `trailing_comments` are the same for all call sites. 
```suggestion
    header_formatter: &'a Content,
    trailing_colon_comments: &'a [SourceComment],
    body: &'a Suite,
    kind: SuiteKind,
```

I'd also flip the `header` and `trailing_colon` arguments as it matches the formatting order more closely:

```
<clause_header>: <trailing_colon_comments>
	<body>
``

File: crates/ruff_python_formatter/tests/snapshots/format@fmt_skip__compound_one_liners.py.snap [hunk 11]
```diff
@@ -0,0 +1,289 @@
...
+if some_very_long_condition_that_might_wrap: do_something_else_that_is_long()  # fmt: skip
+
+# Nested compound statements (outer should be preserved)
+if True:
+    for i in range(10): print(i)  # fmt: skip
+
+# Multiple statements in body (should not apply - multiline)
+if True:
+    x = 1
+    y = 2  # fmt: skip
```
MichaReiser: Change it to something that would format differently to demonstrate that this doesn't break normal `fmt: skip` behavior
```suggestion
    y  =  2  # fmt: skip
```

crates/ruff_python_formatter/tests/snapshots/format@fmt_skip__compound_one_liners.py.snap
```diff
@@ -0,0 +1,289 @@
...
+for i in range(2): print(i)
+else: print("this") # fmt: skip
+
+
+while foo(): print(i)
+else: print("this") # fmt: skip
+
+# multiple statements in body
+if True: print("this"); print("that") # fmt: skip
+```
```
MichaReiser, line 156: What about:

```py
if (
	long_condition
):  a     +     b # fmt: skip
```


I would expect that this doesn't get formatted according to how `fmt:skip`s semantic is defined.

I don't expect this to be hard to support. It might be as easy as only changing the `countains_line_break` check to only test from `:` to the end of the clause

crates/ruff_python_formatter/tests/snapshots/format@fmt_skip__compound_one_liners.py.snap
```diff
@@ -0,0 +1,289 @@
...
+
+class GenericClass(Generic[T]): pass  # fmt: skip
+
+
+# Try/except blocks
+try: risky_operation()  # fmt: skip
+except ValueError: handle_error()  # fmt: skip
+except: handle_any_error()  # fmt: skip
+else: success_case()  # fmt: skip
+finally: cleanup()  # fmt: skip
```
MichaReiser, line 208: Let's add a few more test cases with different comment placements

```suggestion
try: risky_operation()  # fmt: skip
# leading 1
except ValueError: handle_error()  # fmt: skip
# leading 2
except: handle_any_error()  # fmt: skip
# leading 3
else: success_case()  # fmt: skip
# leading 4
finally: cleanup()  # fmt: skip
# trailing
```

MichaReiser (commented):
This is great :) I suggest adding a few more tests with different comment placements. Just to make sure we handle them correctly.

crates/ruff_python_formatter/src/statement/clause.rs
dylwil3: Changed. (Maybe this is an academia thing but I'm used to "compare" sometimes meaning "see also".)

crates/ruff_python_formatter/tests/snapshots/format@fmt_skip__compound_one_liners.py.snap
dylwil3: This was tricky! It has dangling comments. 

But I think I figured it out: `clause_header` has us pass in optional leading comments and I wasn't handling those in the suppressed version of writing `clause`. (The correct transformation of dangling comments to leading comments happens inside the implementation of formatting for `StmtTry`, where they are passed in this way to `clause`.)

crates/ruff_python_formatter/src/statement/stmt_class_def.rs
dylwil3: I don't think so because I think we have to be told whether we're the last suite in the statement (but I think that's the only missing info)

crates/ruff_python_formatter/src/preview.rs
dylwil3: removed gating

dylwil3:
The CI failure here https://github.com/astral-sh/ruff/actions/runs/18664070057/job/53211233008?pr=20633 was caused by this PR uncovering a pre-existing bug. See #21065 and #21067 . Will rebase once that PR is merged and then continue working on this!

dylwil3:
Cut down a comment-related hydra head and more grew in its place. This snippet causes a panic because the indented comment is not marked as formatted. Annoyingly it is attached to `bar()` (because there is no `orelse` node).

```python
for x in it: foo()
    # comment
else: bar() # fmt: skip
```

Converting to draft while I sort it out 😄

crates/ruff_python_formatter/src/statement/clause.rs
dylwil3: Added handling for comments in headers (turns out they can happen now because of a change we made during the course of this code review - we allow multiline headers, it's only the range starting from the colon that has to be on one line). Also clarified in comment how leading and trailing comments are handled.

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -465,6 +532,58 @@ impl Format<PyFormatContext<'_>> for FormatClauseBody<'_> {
     }
 }
 
+pub(crate) struct FormatClause<'a, 'ast> {
+    format_header: FormatClauseHeader<'a, 'ast>,
+    format_body: FormatClauseBody<'a>,
```
MichaReiser: Nit: I think I'd store the data here rather than the formatters and construct the formatters on the fly in `fmt`

```suggestion
    header: ClauseHeader<'a>,
    header_formatter: &'a Content,
    leading_comments: &'a [SourceComment],
    trailing_colon_comment: &'a [SourceComment],
    body: &'a Suite,
    kind: SuiteKind,
```

Mainly because it avoids storing the `trailing_comments` twice, we only construct the formatters when necessary, and expressions like `clause.format_header.header.last_child_in_clause()` require a little less drilling (becomes `clause.header.last_child_in_clause()`)

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -587,3 +706,91 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
...
+
+    let clause_range = TextRange::new(clause_start, last_child_in_clause.end());
+
+    // Only applies to clauses on a single line
+    if source.contains_line_break(clause_range) {
+        return Ok(false);
+    }
+
+    // Recall that we already checked whether there was
+    // a skip comment
```
MichaReiser: I don't understand this comment. What are we `recalling`?

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -587,3 +706,91 @@ fn colon_range(after_keyword_or_condition: TextSize, source: &str) -> FormatResu
...
+    }
+
+    let header = clause.format_header.header;
+    let clause_start = header.first_keyword_range(f.context().source())?.start();
+
+    let comments = f.context().comments().clone();
+
+    let last_child = header
+        .last_child_in_clause()
+        .expect("last child to exist if `should_suppress_clause` is `Ok(true)`");
```
MichaReiser: You could consider changing `should_suppress_clause` to return an enum

```rust
enum SuppressClauseHeader<'a> {
	No,
	Yes { last_child: AnyNodeRef<'a> }
}
```

But I don't think either's fine really

crates/ruff_python_formatter/src/statement/clause.rs
dylwil3: I meant the line above where we did an early return, but I think the comment didn't add much so I just removed it

crates/ruff_python_formatter/src/statement/clause.rs
dylwil3: Done, good idea!

crates/ruff_python_formatter/src/statement/clause.rs
dylwil3: Done!

PR #20634, opened 2025-09-29
[`pyupgrade`] Prevent infinite loop with `I002` and `UP026`

### LINKED ISSUE #20601: [Infinite loop] I002 and UP029
### Summary

Ruff fails to converge when [`missing-required-import` (I002)](https://docs.astral.sh/ruff/rules/missing-required-import/) and [`unnecessary-builtin-import` (UP029)](https://docs.astral.sh/ruff/rules/unnecessary-builtin-import/) are selected and an import from `builtins` is required. [Example](https://play.ruff.rs/ed7201ef-57e1-4cb9-aee0-62f82047dd89):
```console
$ echo 1 | ruff --isolated check - --select I002,UP029 --diff --config 'lint.isort.required-imports=["from builtins import str"]' --unsafe-fixes

...quoting the contents of `-`, the rule codes I002, along with the `pyproject.toml` settings and executed command, we'd be very appreciative!


Would fix 100 errors.
```

### Version

IDrokin117:
## Summary
Closes #20601

Do not treat imports as unused for the rule [unnecessary-builtin-import (UP029)](https://docs.astral.sh/ruff/rules/unnecessary-builtin-import/) if they are required by `isort`([missing-required-import](https://docs.astral.sh/ruff/rules/missing-required-import/))

## Test Plan
- Added test case `i002_up029_conflict` to ensure there is no conflict


ntBre (approved):
Looks great, thank you!

PR #20635, opened 2025-09-29
[ty] Use annotated parameters as type context

ibraheemdev:
## Summary

Use the type annotation of function parameters as bidirectional type context when inferring the argument expression. For example, the following example now type-checks:

```py
class TD(TypedDict):
    x: int

def f(_: TD): ...

f({ "x": 1 })
```

Part of https://github.com/astral-sh/ty/issues/168.


ibraheemdev:
It looks like a lot of these errors are duplicated diagnostics now that we infer argument expressions multiple times for callable unions/overloads, e.g.,

```py
if flag:
    def f(_: int): ...
else:
    def f(_: int): ...

if flag:
    x = 0

# warning[possibly-unresolved-reference] Name `x` used when possibly not defined
# warning[possibly-unresolved-reference] Name `x` used when possibly not defined
f(x)
```

(In this example, multi-inference is not strictly necessary because the signatures are identical, but that's a separate performance concern).

I'm not sure there's a great way around this other than adding checks around all diagnostics that aren't type-context specific?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -9274,6 +9432,22 @@ impl GenericContextError {
...
+    /// Panic if the expression has already been inferred.
+    #[default]
+    Panic,
+
+    /// Store the intersection of all types inferred for the expression.
+    Intersect {
+        // Determines whether or not a given expression is required to be assignable to its type context
+        // despite it being inferred multiple times, i.e. whether eager diagnostics are appropriate, or a
+        // fallback type should be assumed.
+        fallback: bool,
```
ibraheemdev: The reason we even eagerly report diagnostics during the inference of `TypedDict` literals in the first place, is because a later assignability diagnostic would not have context for missing or invalid key diagnostics. Otherwise, I don't think we should be emitting diagnostics based on the type context, which acts more as a type hint (and is not necessarily required to be assignable to, e.g. when there is a fallback overload).

If we had an internal implicit `TypedDict` representation we may be able to avoid this, though we also wouldn't even need bidirectional type inference for `TypedDict` literals in that case.

sharkdp:
> It looks like a lot of these errors are duplicated diagnostics now that we infer argument expressions multiple times for callable unions/overloads […]  I'm not sure there's a great way around this other than adding checks around all diagnostics that aren't type-context specific?

Instead of adding checks for every non-type-context-specific context, can we use state on `InferContext` to keep track of whether diagnostics should be emitted or not (see the `InferContext::no_type_check` field for a similar functionality)? And then, we could circumvent the general silencing of all diagnostics for those diagnostics that should still be emitted?

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: > The reason we even eagerly report diagnostics during the inference of `TypedDict` literals in the first place, is because a later assignability diagnostic would not have context for missing or invalid key diagnostics.

Correct

> Otherwise, I don't think we should be emitting diagnostics based on the type context, which acts more as a type hint (and is not necessarily required to be assignable to, e.g. when there is a fallback overload).

:+1: 

> If we had an internal implicit `TypedDict` representation we may be able to avoid this, though we also wouldn't even need bidirectional type inference for `TypedDict` literals in that case.

Yes, although Carl pointed out that this approach would be tricky as well, because we would need to keep track of mutations (including through aliases):

```py
class Person(TypedDict):
    name: str
    age: int

def takes_person(p: Person): ...

# creates ImplicitTypedDict {"name": Literal["Alice"], "age": Literal[20] }
alice = {"name": "Alice", "age": 20}

takes_person(alice)  # okay

evil = alice
evil["age"] = None

takes_person(alice)  # not okay anymore
```

But on the other hand, this is a case that the current approach can't handle either, because `alice` would just be a `dict[Unknown | str, Unknown | str | int]`.

crates/ty_python_semantic/src/types/context.rs
```diff
@@ -153,7 +169,33 @@ impl<'db, 'ast> InferContext<'db, 'ast> {
...
+    /// type context, which may change across multiple inferences of a function
+    /// argument expression.
+    #[allow(dead_code)]
+    pub(super) fn report_bidirectional_diagnostic<'ctx>(
+        &'ctx self,
+        id: DiagnosticId,
+        severity: Severity,
+    ) -> Option<DiagnosticGuardBuilder<'ctx, 'db>> {
+        DiagnosticGuardBuilder::new(self, id, severity, true)
+    }
```
MichaReiser: I think we can wait with adding this method. `report_diagnostic` is only used for errors that can't be suppressed and shouldn't be specific to bidirectional type inference.

ibraheemdev:
I've decided to remove the idea of "lints that may be affected by bidirectional type context", as this seemed quite. Instead, the new implementation does the following for each argument:
- Performs type inference directly with the unique annotated parameter type, if there is a single binding and overload.
- Otherwise, performs type inference once without type context. This will surface any diagnostics from inference of the expression. We then perform type inference once for each overload, with its annotated parameter type, with diagnostics silenced. The intersection of all these types are taken — so the type context can potentially allow us to infer a more assignable type, but will never lead to any new diagnostics, as we don't know which overload will match during argument inference.

This does mean we get worse diagnostics for `TypedDict` arguments when there are multiple bindings/overloads involved, due to the fallback to a generic `dict` (once `TypedDict` subtyping is properly implemented, of course), but I don't think there's really a way around this, and it doesn't seem like a priority.

I also went through the ecosystem report, and most diagnostic changes seem to be correct, or the result of an unrelated known limitation (namely `TypedDict` subtyping, and the new constraint solver). This PR should be ready for review.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2512,9 +2519,17 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
         if let Some(return_ty) = self.signature.return_ty
             && let Some(call_expression_tcx) = self.call_expression_tcx.annotation
         {
-            // Ignore any specialization errors here, because the type context is only used to
-            // optionally widen the return type.
-            let _ = builder.infer(return_ty, call_expression_tcx);
+            match call_expression_tcx {
+                // A type variable is not a useful type-context for expression inference, and applying it
+                // to the return type can lead to confusing unions in nested generic calls.
+                Type::TypeVar(_) => {}
```
ibraheemdev, line 2517: This is mostly for better diagnostics. Without it, `reveal_type(identity(1))` is revealed as `T@reveal_type | int`. It also doesn't quite work for parameters annotated as `T | None` — we still might union the inferred type with `T@f` in that case.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -3661,22 +3693,29 @@ impl<'db> BindingError<'db> {
...
+        match Self::get_argument_node(node, argument_index) {
+            Some(ast::ArgOrKeyword::Arg(expr)) => expr.into(),
+            Some(ast::ArgOrKeyword::Keyword(expr)) => expr.into(),
+            None => node,
+        }
+    }
+
+    fn get_argument_node(
+        node: ast::AnyNodeRef<'_>,
+        argument_index: Option<usize>,
```
dcreager, line 3671: nit: I think you can update this helper to take in `argument_index: usize`, and use `argument_index?` up above when you call the helper. Then you only need to match on `node` below.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4924,22 +4929,155 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            .matching_overloads()
+                            .map(move |(_, overload)| (overload, binding));
+
+                        Some(Either::Right(overloads))
+                    }
+
+                    // If there is a single overload that does not match, we still infer the argument
+                    // types for better diagnostics.
+                    MatchingOverloadIndex::None => match binding.overloads() {
+                        [overload] => Some(Either::Left([(overload, binding)].into_iter())),
```
dcreager: nit: Can you use `std::iter::once` here? I find the array notation makes me assume there will be multiple elements, which is subtle since there's a tuple here, too.

crates/ty_python_semantic/src/types.rs
```diff
@@ -4197,20 +4197,26 @@ impl<'db> Type<'db> {
...
-                                .type_form()
-                                .with_annotated_type(Type::any()),
-                        ]),
-                        Some(Type::none(db)),
-                    ),
-                )
-                .into(),
+                Some(KnownFunction::AssertType) => {
+                    let val_ty =
+                        BoundTypeVarInstance::synthetic(db, "T", TypeVarVariance::Invariant);
```
dcreager, line 4199: Were there tests that were failing with the old signature? I can see that the new signature means that there's now a return value, which will have the same type as the value being checked, but I don't see any tests that rely on that behavior.

ibraheemdev: The issue that showed up in the ecosystem report was that the `f(1)` in `assert_type(f(1), int)` was inferred as `int | Any`, because the signature of `assert_type` was `(Any, ..) -> ..`. I added a test.

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: We also use `get_argument_node` in the diagnostic code with an `Option<usize>`, so I'm not sure the change is worth it.

ibraheemdev:
Deduplicating the parameter types helps the performance regression somewhat, but outside of that I think it's unavoidable.

PR #20636, opened 2025-09-29
[ty] Feature/improve slots support

BlairAllan:
## Summary

Implements `__slots__` support as requested in astral-sh/ty#1268. These changes improves ty to properly analyse code that uses `__slots__` with specific support for attrs.

Key changes:
- Attributes defined in `__slots__` are no longer unresolved
- Automatic removal of `__dict__` and `__weakref__` when not in `__slots__`
- Improved error reporting for `__slots__` violations

### Created slots.md
  - Tests slot types:
    - Tuple slots: `__slots__ `= ("name", "value")
    - Single-char string: `__slots__` = "x"
    - Multi-char string: `__slots__` = "abcd" (creates slots "a", "b", "c", "d")
    - Empty slots: `__slots__` = ()
    - Inheritance scenarios
    - Attrs like testing

### Manual testing
  - Verified error messages match expectations:
  ts.invalid = 1    # "Unresolved attribute `invalid` on type `TupleSlots`"
  sc.y = 2          #  "Unresolved attribute `y` on type `SingleChar`"
  mc.abcd = 3       #  "Unresolved attribute `abcd` on type `MultiChar`"

- Attrs Testing
  - Attrs uses tuple-style `__slots__` = ("field1", "field2") which this implementation handles perfectly


sharkdp:
Thank you for your contribution.

You recently opened (and then closed) [a similar PR](https://github.com/astral-sh/ruff/pull/20607) that explicitly contained @Claude authored commits. The changes in this PR look very similar, but the usage of AI tools is not mentioned anywhere. We are not opposed to the usage of these tools, but we are *not* interested in agent-generated PRs that no human has reviewed. We could easily generate those ourselves, if we wanted to.

Can you please clarify the current state of this PR? Are most of the changes here AI generated? If so, what kind of work did you put in as a developer to verify those changes?

Also, the current version seems to break multiple tests. Have you looked into this?

BlairAllan:
Thanks for the follow-up.

I used Claude while debugging the type inference issues around how __slots__ interacts with  __dict__ and __weakref__ was fairly tricky and left TODOS around them because theres still some unwanted behaviour.

In the latest commit, I rechecked all changes refactored, added, rewrote the tests to confirm the expected behavior and correcting my previous commits.

Now this should all be working and ready for review wit the changes matching all the usual mypy __slots__ checks and further by catching non-string items, slot defaults, builtin subclass issues, and multiple inheritance conflicts.

BlairAllan:
Thanks for the follow-up.

I used Claude while debugging the type inference issues around how slots interacts with dict and weakref was fairly tricky and left TODOS around them because theres still some unwanted behaviour.

In the latest commit, I rechecked all changes refactored, added, rewrote the tests to confirm the expected behavior and corrected my previous commits.

Now this should all be working and ready for review wit the changes matching all the usual mypy slots checks and further by catching non-string items, slot defaults, builtin subclass issues, and multiple inheritance conflicts.

sharkdp:
> I used Claude while debugging the type inference issues

Okay, but did you (or did you not) generate the bulk of this code using Claude? If the answer is yes, did you carefully review all of the code?

I'm asking because this is a big change (surprisingly big, in fact) and we don't want to invest time reviewing this if it's mostly-unreviewed LLM-generated code.

From a quick skim, we noticed that the code doesn't seem to interact with the existing `__slots__` support. Could you please comment on why that is the case?

Also, it seems that this could be split into multiple smaller PRs (do the core type inference feature first, then incrementally add more checks). If that seems reasonable to you, this would certainly help us review the changes.

BlairAllan:
> > I used Claude while debugging the type inference issues
> 
> Okay, but did you (or did you not) generate the bulk of this code using Claude? If the answer is yes, did you carefully review all of the code?
> 
> I'm asking because this is a big change (surprisingly big, in fact) and we don't want to invest time reviewing this if it's mostly-unreviewed LLM-generated code.
> 
> From a quick skim, we noticed that the code doesn't seem to interact with the existing `__slots__` support. Could you please comment on why that is the case?
> 
> Also, it seems that this could be split into multiple smaller PRs (do the core type inference feature first, then incrementally add more checks). If that seems reasonable to you, this would certainly help us review the changes.

Hi @sharkdp thank you for the feedback and sorry for the delay I've gone and split out this PR into a much smaller one ~150 lines #20996 . It could be smaller if this is still too large but I wanted to make it easier to add further constraints in the future. The initial PR just covers basic inference with several tests. 

To be transparent I use Claude to double check my code generally, but I wrote the core inference logic. Specifically when it came to adding the diagnostics I did use Claude to assist in debugging/helping me code as it required a better understanding of how ty raises these errors, which I lacked. I knew what I wanted to add to ty to improve `__slots__`  so I followed the flow of write what I know -> use Claude to find where I needed to extend/debug/add -> extensively test and read changes and verify they're correct. Just to assure you all changes, if any, made by Claude were actively reviewed and checked by myself, I wouldn't want to waste your time or the teams. 

Regarding why didn't I use the existing `__slots__` support, I added an explanation to the new PR but effectively the existing code identifies if theres something in `__slots__` but it doesn't enable us to know what the actual attribute names are, which is needed to test if a specific attribute is allowed or not.

sharkdp:
Thank you for the clarifications.

PR #20642, opened 2025-09-29
[`cli`] Add conflict between `--add-noqa` and `--diff` options

### LINKED ISSUE #20609: `warning: --fix is incompatible with --add-noqa` when using `--diff`
### Summary

It is not possible to see what an `add-noqa` operation would do before actually doing it.

```shell
$ ruff check --select D101,D102 --add-noqa --diff src/
warning: --fix is incompatible with --add-noqa.
$ 
```
IMHO: ruff either complains, or it generates fix-ups.
In the latter case, whether ruff (a) rewrites the code or (b) adds `#noqa` should be orthogonal to whether (x) changes are applied or (y) a diff is emitted; ax ay bx by should all be possible.

### Version

0.13.2

danparizher:
## Summary

Fixes #20609

### DIFF
--- a/crates/ruff/src/args.rs
+++ b/crates/ruff/src/args.rs
@@ -416,6 +416,7 @@ pub struct CheckCommand {
         conflicts_with = "stdin_filename",
         conflicts_with = "watch",
         conflicts_with = "fix",
+        conflicts_with = "diff",
     )]
     pub add_noqa: bool,
     /// See the files Ruff will be run against with the current settings.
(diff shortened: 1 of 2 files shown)


danparizher:
I'm not sure if we even want a test for conflicts like this, as I haven't seen any other tests similar to it. Let me know if we want it removed.

danparizher:
Actually, I can see that the executable in the CI has different syntax depending on the OS, so I'll just remove the test.

PR #20643, opened 2025-09-30
[ty] No union with `Unknown` for module-global symbols

sharkdp:
# not ready for review

## Summary

https://github.com/astral-sh/ty/issues/1069


sharkdp:
Let's do this without literal promotion first, and then potentially add literal promotion in a second step to evaluate it separately.

PR #20645, opened 2025-09-30
Update Salsa

MichaReiser:


MichaReiser:
Hmm, this does seem to regress perf for quiet a few projects (1-4%)

MichaReiser:
Okay, I think I now have the right amount of `inline` attributes in salsa so that this no longer regresses performance

crates/ty_python_semantic/resources/primer/bad.txt
```diff
@@ -1,2 +1,2 @@
-spark  # too many iterations (in `exported_names` query)
-steam.py  # hangs (single threaded)
+spark  # too many iterations (in `exported_names` query), `should not be able to access instance member `spark` of type variable IndexOpsLike@astype in inferable position`
+steam.py  # dependency graph cycle when querying TypeVarInstance < 'db >::lazy_default_(Id(2e007)), set cycle_fn/cycle_initial to fixpoint iterate.
```
MichaReiser, line 2: This should be easy to fix but let's do this in a separate PR

MichaReiser:
Hmm, the small performance regression (except pandas) is back. I suspect it's just due to differences in inlining, now that we switched to LTO fat? I don't think it's worth narrowing this down right now

sharkdp (approved):
Congratulations for fixing this — thank you so much!

PR #20646, opened 2025-09-30
[ty] Literal promotion refactor

sharkdp:
## Summary

Not sure if this was the original intention, but it looks to me like the previous `Type::literal_promotion_type` was more of an implementation detail for the actual operation of promoting all literals in a possibly-nested position of a type.

## Test Plan

New Markdown test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/protocols.md
+++ b/crates/ty_python_semantic/resources/mdtest/protocols.md
@@ -893,8 +893,10 @@ class LotsOfBindings(Protocol):
     match object():
         case l:  # error: [ambiguous-protocol-member]
             ...
+    # error: [ambiguous-protocol-member] "Consider adding an annotation, e.g. `m: int | str = ...`"
+    m = 1 if 1.2 > 3.4 else "a"
 
-# revealed: frozenset[Literal["Nested", "NestedProtocol", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l"]]
+# revealed: frozenset[Literal["Nested", "NestedProtocol", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m"]]
 reveal_type(get_protocol_members(LotsOfBindings))
 
 class Foo(Protocol):
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1166,21 +1166,26 @@ impl<'db> Type<'db> {
         }
     }
 
-    /// If this type is a literal, promote it to a type that this literal is an instance of.
+    /// Promote (possibly nested) literals to types that these literals are instances of.
     ///
     /// Note that this function tries to promote literals to a more user-friendly form than their
     /// fallback instance type. For example, `def _() -> int` is promoted to `Callable[[], int]`,
     /// as opposed to `FunctionType`.
-    pub(crate) fn literal_promotion_type(self, db: &'db dyn Db) -> Option<Type<'db>> {
+    pub(crate) fn promote_literals(self, db: &'db dyn Db) -> Type<'db> {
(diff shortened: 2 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -893,8 +893,10 @@ class LotsOfBindings(Protocol):
     match object():
         case l:  # error: [ambiguous-protocol-member]
             ...
+    # error: [ambiguous-protocol-member] "Consider adding an annotation, e.g. `m: int | str = ...`"
+    m = 1 if 1.2 > 3.4 else "a"
```
sharkdp, line 897: Yes, unlikely to ever appear in practice. But a nice way to demonstrate how the previous `Type::literal_promotion_type` was different from `Type::promote_literals` (this only works with the latter).

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2625,6 +2625,12 @@ pub(crate) fn report_undeclared_protocol_member(
                 SubclassOfInner::Dynamic(_) => return false,
             },
             Type::NominalInstance(instance) => instance.class(db),
+            Type::Union(union) => {
+                return union
+                    .elements(db)
+                    .iter()
+                    .all(|elem| should_give_hint(db, *elem));
+            }
```
sharkdp, line 2633: Admittedly more of a showcase for the refactor here. I'm also happy to remove this again (and the test above).

crates/ty_python_semantic/resources/mdtest/protocols.md
AlexWaygood: > unlikely to ever appear in practice

what do you mean, i write protocols like this all the time

crates/ty_python_semantic/src/types/diagnostic.rs
AlexWaygood, line 2641: I wonder if this could/should be something like 

```suggestion
class.known(db).is_none_or(|known_class| !known_class.is_singleton())
```

instead... but it probably doesn't really make a real-world difference. And off-topic for this PR.

crates/ty_python_semantic/src/types.rs
```diff
@@ -1166,21 +1166,26 @@ impl<'db> Type<'db> {
...
-            Type::BytesLiteral(_) => Some(KnownClass::Bytes.to_instance(db)),
-            Type::ModuleLiteral(_) => Some(KnownClass::ModuleType.to_instance(db)),
-            Type::EnumLiteral(literal) => Some(literal.enum_class_instance(db)),
-            Type::FunctionLiteral(literal) => Some(Type::Callable(literal.into_callable_type(db))),
-            _ => None,
+            Type::StringLiteral(_) | Type::LiteralString => KnownClass::Str.to_instance(db),
+            Type::BooleanLiteral(_) => KnownClass::Bool.to_instance(db),
+            Type::IntLiteral(_) => KnownClass::Int.to_instance(db),
+            Type::BytesLiteral(_) => KnownClass::Bytes.to_instance(db),
+            Type::ModuleLiteral(_) => KnownClass::ModuleType.to_instance(db),
```
AlexWaygood, line 1185: I'm not actually sure when promotion of module-literal types is desirable -- we might want to remove it from this method. As you discovered on your "remove `| Unknown` from module-globals PR (sorry for looking while it was still in draft 🙈), promotion of module-literal types to `types.ModuleType` can cause serious issues because each module-literal type has a distinct set of attributes available on it. This makes module-literal types somewhat different to all our other literal types

sharkdp: Was thinking the exact same thing when I changed this code. In fact, I did arrive here in the first place because promotion of module literals was causing the exact problem you are describing. I'll propose that as a separate change.

Edit: I should have read the rest of your comment before writing mine.

PR #20647, opened 2025-09-30
[ty] Clean up inherited generic contexts

dcreager:
We add an `inherited_generic_context` to the constructors of a generic class. That lets us infer specializations of the class when invoking the constructor. The constructor might itself be generic, in which case we have to merge the list of typevars that we are willing to infer in the constructor call.

Before we did that by tracking the two (and their specializations) separately, with distinct `Option` fields/parameters. This PR updates our call binding logic such that any given function call has _one_ optional generic context that we're willing to infer a specialization for. If needed, we use the existing `GenericContext::merge` method to create a new combined generic context for when the class and constructor are both generic. This simplifies the call binding code considerably, and is no more complex in the constructor call logic.

We also have a heuristic that we will promote any literals in the specialized types of a generic class, but we don't promote literals in the specialized types of the function itself. To handle this, we now track this `should_promote_literals` property within `GenericContext`. And moreover, we track this separately for each typevar, instead of a single property for the generic context as a whole, so that we can correctly merge the generic context of a constructor method (where the option should be `false`) with the inherited generic context of its containing class (where the option should be `true`).

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -5459,24 +5459,19 @@ impl<'db> Type<'db> {
                     }
                 }
 
-                let new_specialization = new_call_outcome
-                    .and_then(Result::ok)
-                    .as_ref()
-                    .and_then(Bindings::single_element)
-                    .into_iter()
-                    .flat_map(CallableBinding::matching_overloads)
-                    .next()
-                    .and_then(|(_, binding)| binding.specialization())
-                    .and_then(|specialization| specialization.restrict(db, generic_context?));
-                let init_specialization = init_call_outcome
-                    .and_then(Result::ok)
-                    .as_ref()
-                    .and_then(Bindings::single_element)
-                    .into_iter()
-                    .flat_map(CallableBinding::matching_overloads)
-                    .next()
-                    .and_then(|(_, binding)| binding.specialization())
-                    .and_then(|specialization| specialization.restrict(db, generic_context?));
+                let specialize_constructor = |outcome: Option<Bindings<'db>>| {
+                    let (_, binding) = outcome
+                        .as_ref()?
+                        .single_element()?
+                        .matching_overloads()
+                        .next()?;
+                    binding.specialization()?.restrict(db, generic_context?)
+                };
+
+                let new_specialization =
+                    specialize_constructor(new_call_outcome.and_then(Result::ok));
+                let init_specialization =
+                    specialize_constructor(init_call_outcome.and_then(Result::ok));
                 let specialization =
                     combine_specializations(db, new_specialization, init_specialization);
                 let specialized = specialization
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6772,13 +6772,11 @@ impl<'db> TypeMapping<'_, 'db> {
                 db,
                 context
                     .variables(db)
-                    .iter()
-                    .filter(|var| !var.typevar(db).is_self(db))
-                    .copied(),
+                    .filter(|var| !var.typevar(db).is_self(db)),
             ),
             TypeMapping::ReplaceSelf { new_upper_bound } => GenericContext::from_typevar_instances(
                 db,
-                context.variables(db).iter().map(|typevar| {
+                context.variables(db).map(|typevar| {
                     if typevar.typevar(db).is_self(db) {
                         BoundTypeVarInstance::synthetic_self(
                             db,
@@ -6786,7 +6784,7 @@ impl<'db> TypeMapping<'_, 'db> {
                             typevar.binding_context(db),
                         )
                     } else {
-                        *typevar
+                        typevar
                     }
                 }),
             ),
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -5481,11 +5481,6 @@ impl<'db> Type<'db> {
                     combine_specializations(db, new_specialization, init_specialization);
                 let specialized = specialization
                     .map(|specialization| {
-                        // Promote any typevars that are inferred as a literal to the corresponding
-                        // instance type.
-                        specialization.apply_type_mapping(db, &TypeMapping::PromoteLiterals)
-                    })
-                    .map(|specialization| {
                         Type::instance(
                             db,
                             generic_origin.apply_specialization(db, |_| specialization),
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -5466,10 +5466,8 @@ impl<'db> Type<'db> {
                     .into_iter()
                     .flat_map(CallableBinding::matching_overloads)
                     .next()
-                    .and_then(|(_, binding)| binding.inherited_specialization())
-                    .filter(|specialization| {
-                        Some(specialization.generic_context(db)) == generic_context
-                    });
+                    .and_then(|(_, binding)| binding.specialization())
+                    .and_then(|specialization| specialization.restrict(db, generic_context?));
                 let init_specialization = init_call_outcome
                     .and_then(Result::ok)
                     .as_ref()
@@ -5477,13 +5475,16 @@ impl<'db> Type<'db> {
                     .into_iter()
                     .flat_map(CallableBinding::matching_overloads)
                     .next()
-                    .and_then(|(_, binding)| binding.inherited_specialization())
-                    .filter(|specialization| {
-                        Some(specialization.generic_context(db)) == generic_context
-                    });
+                    .and_then(|(_, binding)| binding.specialization())
+                    .and_then(|specialization| specialization.restrict(db, generic_context?));
                 let specialization =
                     combine_specializations(db, new_specialization, init_specialization);
                 let specialized = specialization
+                    .map(|specialization| {
+                        // Promote any typevars that are inferred as a literal to the corresponding
+                        // instance type.
+                        specialization.apply_type_mapping(db, &TypeMapping::PromoteLiterals)
+                    })
                     .map(|specialization| {
                         Type::instance(
                             db,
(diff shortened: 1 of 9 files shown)


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -139,9 +139,6 @@ pub(crate) fn typing_self<'db>(
 
 /// A list of formal type variables for a generic function, class, or type alias.
 ///
-/// TODO: Handle nested generic contexts better, with actual parent links to the lexically
-/// containing context.
-///
```
dcreager, line 144: This seems moot now that we have methods for walking the lexically containing contexts when needed.

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -484,58 +479,13 @@ impl<'db> OverloadLiteral<'db> {
 #[derive(PartialOrd, Ord)]
 pub struct FunctionLiteral<'db> {
```
dcreager, line 480: `FunctionLiteral` now only contains one field, so there's an even better argument to consolidate some of the types in this file.

dcreager:
Moving this to draft while I look at the ecosystem results

dcreager:
> Moving this to draft while I look at the ecosystem results

The issue was with this part of my original PR description:

> We also have a heuristic that we will promote any literals in the specialized types of a generic class, but we don't promote literals in the specialized types of the function itself. This PR also has the nice side benefit that that heuristic is now implemented in the constructor call machinery (where is arguably a better place for it) rather than in the general call binding machinery.

It turns out that we _need_ to perform literal promotion as part of the call binding machinery; otherwise we are using the literal types (not the promoted types) when doing type checking of arguments against parameter annotations.

That complicates this PR slightly, since we now need some other mechanism to track whether we want to perform literal promotion on each typevar. The premise of this PR is that we want a single `GenericContext` to hold the entire "inference context" of a particular call site, merging together the generic context of a constructor and its containing class if needed. That means that we need to track this option on a per-typevar basis, not a per-generic-context basis.

I had considered doing that by adding a `should_promote_literals` field to `BoundTypeVarInstance`, but that makes typevar equality checks (and by extension, looking up the specialization of a particular typevar in a `Specialization`) more brittle. So instead, I've updated `GenericContext` to hold a small options struct for each typevar, and put the `should_promote_literals` field there. When accessing a constructor member, we create a copy of the class's generic context where we update that option to `true` for all of the typevars in the generic context. When that is merged with the generic context of the constructor itself (if any), the inherited typevars from the class have the option set, and the typevars from the method do not.

dcreager:
The remaining ecosystem results look correct. Or at least, they're known false positives that we're now picking up because of inferring generic class typevars more correctly.

---

`home-assistant`

[Here](https://github.com/home-assistant/core/blob/fd8ccb8d8f6749c3672168fc8bf2bfe28481e16b/homeassistant/core.py#L1566) we're calling the `Event[_DataT]` constructor, but our narrowing constraints aren't noticing that `event_data` cannot be `None`. We need to either fix the narrowing logic to eliminate `None` from the union, or rely on the better constraint solver to infer a more precise type for `_DataT`.

---

`pandas`

The `json`-related diagnostics are an interesting case where our literal promotion logic is not what the user would want. They are using literal strings as type tags in the `JSONReader` class, and when they instantiate a `JSONReader` with `typ="frame"`, they _want_ the inferred type to be `JSONReader[Literal["frame"]]`, not `JSONReader[str]`. But our literal promotion logic engages (correctly, where it should have been already doing so).

For the `timedelta` diagnostics, they are [instantiating](https://github.com/pandas-dev/pandas/blob/5cc32409652002b4de442cfa6a065553aa17c3c9/pandas/tests/scalar/timedelta/test_arithmetic.py#L181) a `numpy.timedelta64` using a [dedicated constructor overload](https://github.com/numpy/numpy/blob/e7276da07f6fc69cc2947f4dc67a87472e4b3a8b/numpy/__init__.pyi#L5208) that matches when the constructor parameter is a particular literal string. Our literal promotion is engaging before we check overloads, so we don't match that overload.

crates/ty_python_semantic/src/types.rs
```diff
@@ -5466,21 +5466,17 @@ impl<'db> Type<'db> {
...
                     .and_then(Bindings::single_element)
                     .into_iter()
                     .flat_map(CallableBinding::matching_overloads)
                     .next()
-                    .and_then(|(_, binding)| binding.inherited_specialization())
-                    .filter(|specialization| {
-                        Some(specialization.generic_context(db)) == generic_context
-                    });
+                    .and_then(|(_, binding)| binding.specialization())
+                    .and_then(|specialization| specialization.restrict(db, generic_context?));
```
AlexWaygood: I think you could improve readability here by factoring out a closure:


```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 1ab2447b5e..05dc47c3ff 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -5459,24 +5459,22 @@ impl<'db> Type<'db> {
                     }
                 }
 
-                let new_specialization = new_call_outcome
-                    .and_then(Result::ok)
-                    .as_ref()
-                    .and_then(Bindings::single_element)
-                    .into_iter()
-                    .flat_map(CallableBinding::matching_overloads)
-                    .next()
-                    .and_then(|(_, binding)| binding.specialization())
-                    .and_then(|specialization| specialization.restrict(db, generic_context?));
-                let init_specialization = init_call_outcome
-                    .and_then(Result::ok)
-                    .as_ref()
-                    .and_then(Bindings::single_element)
-                    .into_iter()
-                    .flat_map(CallableBinding::matching_overloads)
-                    .next()
-                    .and_then(|(_, binding)| binding.specialization())
-                    .and_then(|specialization| specialization.restrict(db, generic_context?));
+                let specialize_constructor = |outcome: Option<Bindings<'db>>| {
+                    let generic_context = generic_context?;
+
+                    let (_, binding) = outcome
+                        .as_ref()?
+                        .single_element()?
+                        .matching_overloads()
+                        .next()?;
+
+                    binding.specialization()?.restrict(db, generic_context)
+                };
+
+                let new_specialization =
+                    specialize_constructor(new_call_outcome.and_then(Result::ok));
+                let init_specialization =
+                    specialize_constructor(init_call_outcome.and_then(Result::ok));
                 let specialization =
                     combine_specializations(db, new_specialization, init_specialization);
                 let specialized = specialization
```

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -185,21 +201,46 @@ impl<'db> GenericContext<'db> {
...
         )
     }
 
+    pub(crate) fn variables(
+        self,
+        db: &'db dyn Db,
+    ) -> impl ExactSizeIterator<Item = BoundTypeVarInstance<'db>> + Clone {
+        self.variables_inner(db)
+            .iter()
+            .map(|(bound_typevar, _)| *bound_typevar)
```
AlexWaygood:
```suggestion
        self.variables_inner(db).keys().copied()
```

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: maybe rewrite this as a closure?


```diff
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index 77bc4578b1..0462d31c70 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -5351,19 +5351,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         collection_class: KnownClass,
     ) -> Option<Type<'db>> {
         // Extract the type variable `T` from `list[T]` in typeshed.
-        fn elt_tys(
-            collection_class: KnownClass,
-            db: &dyn Db,
-        ) -> Option<(
-            ClassLiteral<'_>,
-            impl Iterator<Item = BoundTypeVarInstance<'_>> + Clone,
-        )> {
-            let class_literal = collection_class.try_to_class_literal(db)?;
-            let generic_context = class_literal.generic_context(db)?;
-            Some((class_literal, generic_context.variables(db)))
-        }
-
-        let (class_literal, elt_tys) = elt_tys(collection_class, self.db()).unwrap_or_else(|| {
+        let elt_tys = |collection_class: KnownClass| {
+            let class_literal = collection_class.try_to_class_literal(self.db())?;
+            let generic_context = class_literal.generic_context(self.db())?;
+            Some((class_literal, generic_context.variables(self.db())))
+        };
+
+        let (class_literal, elt_tys) = elt_tys(collection_class).unwrap_or_else(|| {
             let name = collection_class.name(self.db());
             panic!("Typeshed should always have a `{name}` class in `builtins.pyi`")
         });
```

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -695,7 +629,6 @@ pub(super) fn walk_function_type<'db, V: super::visitor::TypeVisitor<'db> + ?Siz
     function: FunctionType<'db>,
     visitor: &V,
 ) {
-    walk_function_literal(db, function.literal(db), visitor);
```
AlexWaygood, line 698: we no longer need this call because `FunctionLiteral` no longer contains any non-atomic types within it that would need to be walked?

dcreager: Yep, that's right

crates/ty_python_semantic/src/types.rs
dcreager: I like

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: Done

AlexWaygood:
> 1. No successful run was found on `main` ([c91b457](https://github.com/astral-sh/ruff/commit/c91b4570448abf9871f29476fbcf822abd1afd92)) during the generation of this report, so 673167a was used instead as the comparison base. There might be some changes unrelated to this pull request in this report. [↩](#user-content-fnref-unexpected-base-5abce3b6bd8bc785d5787d8c21019003)

I suspect Codspeed _might_ have picked up some of the perf improvements from https://github.com/astral-sh/ruff/commit/c91b4570448abf9871f29476fbcf822abd1afd92 accidentally there haha

PR #20650, opened 2025-09-30
[ty] Fix subtyping of invariant generics specialized with `Any`

AlexWaygood:
## Summary

Currently we consider `list[Any]` a subtype of `list[Any]`. But we shouldn't, for the same reason that we shouldn't consider `Any` a subtype of `Any`: the two `Any`s might materialize to different types.

The reason why we make this mistake on `main` is that we (correctly) consider `list[Any]` to be equivalent to `list[Any]`, but then incorrectly assume that if `T` is equivalent to `S` then `T` will necessarily be a subtype of `S`.

## Test Plan

- Added mdtests that fail on `main`.
- Ran `QUICKCHECK_TESTS=1000000 cargo test --release -p ty_python_semantic -- --ignored types::property_tests::stable`

### DIFF
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -602,8 +602,18 @@ fn has_relation_in_invariant_position<'db>(
             base_mat,
             visitor,
         ),
-        // Subtyping between invariant type parameters without a top/bottom materialization involved
-        // is equivalence
+        // Subtyping between invariant type parameters without a top/bottom materialization necessitates
+        // checking the subtyping relation both ways: `A` must be a subtype of `B` *and* `B` must be a
+        // subtype of `A`. The same applies to assignability.
+        //
+        // For subtyping between fully static types, this is the same as equivalence. However, we cannot
+        // use `is_equivalent_to` (or `when_equivalent_to`) here, because we (correctly) understand
+        // `list[Any]` as being equivalent to `list[Any]`, but we don't want `list[Any]` to be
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -604,16 +604,10 @@ fn has_relation_in_invariant_position<'db>(
...
+        (None, None, relation) => derived_type
+            .has_relation_to_impl(db, *base_type, relation, visitor)
             .and(db, || {
-                base_type.has_relation_to_impl(
-                    db,
-                    *derived_type,
-                    TypeRelation::Assignability,
-                    visitor,
-                )
+                base_type.has_relation_to_impl(db, *derived_type, relation, visitor)
```
dcreager, line 620: This might deserve a comment why we can't use `when_equivalent_to` here. (I was confused at first, at least.) My understanding is that it's because `when_equivalent_to` specifically checks _gradual_ equivalence, which won't give the right answer for `TypeRelation::Subtyping`. (If that's right, it suggests that the old code was wrong for the `Subtyping` case?)

dcreager (approved):

crates/ty_python_semantic/src/types/generics.rs
dcreager: (Or maybe based on your Discord comment it's because it's faster? That would also deserve a comment)

AlexWaygood: Heh, no, the motivation was correctness and the speed improvement was a pleasant surprise!!

PR #13775, opened 2024-10-16
[red knot] add `Type::is_disjoint_from` and intersection simplifications

sharkdp:
## Summary

- Add `Type::is_disjoint_from` as a way to test whether two types overlap
- Add a first set of simplification rules for intersection types
  - `S & T = S` for `S <: T`
  - `S & ~T = Never` for `S <: T`
  - `~S & ~T = ~T` for `S <: T`
  - `A & ~B = A` for `A` disjoint from `B`
  - `A & B = Never` for `A` disjoint from `B`
  - `bool & ~Literal[bool] = Literal[!bool]`

resolves one item in #12694

## Open questions:

- Can we somehow leverage the (anti) symmetry between `positive` and `negative` contributions? I could imagine that there would be a way if we had `Type::Not(type)`/`Type::Negative(type)`, but with the `positive`/`negative` architecture, I'm not sure. Note that there is a certain duplication in the `add_positive`/`add_negative` functions (e.g. `S & ~T = Never` is implemented twice), but other rules are actually not perfectly symmetric: `S & T = S` vs `~S & ~T = ~T`.
- I'm not particularly proud of the way `add_positive`/`add_negative` turned out. They are long imperative-style functions with some mutability mixed in (`to_remove`). I'm happy to look into ways to improve this code *if we decide to go with this approach* of implementing a set of ad-hoc rules for simplification.
- ~~Is it useful to perform simplifications eagerly in `add_positive`/`add_negative`? (@carljm)~~ This is what I did for now.

## Test Plan

- Unit tests for `Type::is_disjoint_from`
- Observe changes in Markdown-based tests
- Unit tests for `IntersectionBuilder::build()`

PR #20651, opened 2025-09-30
[ty] Fix simplification of `T & ~T` for non-fully-static types

AlexWaygood:
## Summary

~~(Stacked on top of https://github.com/astral-sh/ruff/pull/20650; review that PR first!)~~

On `main`, we simplify `Any & ~Any` to `Any`, but we simplify `list[Any] & ~list[Any]` to `Never`. This is inconsistent! The justification for the first simplification applies to all non-fully-static types, so we should apply the same treatment to both.

`T & ~T` cannot simplify to `Never` for any non-fully-static type `T`, but it can simplify to `T`. We cannot say that `list[Any] & ~list[Any]` is so small such that the only materialization of the type is `Never`, as the first `Any` might (for example) materialize to `int` while the second one might materialize to `bool`, and `list[int] & ~list[bool]` is not equivalent to `Never` (it's equivalent to `list[int]`, since the two types are disjoint)

## Test Plan

Added mdtests that fail on `main`

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/intersection_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/intersection_types.md
@@ -907,6 +907,11 @@ def unknown(
 Dynamic types do not cancel each other out. Intersecting an unknown set of values with the negation
 of another unknown set of values is not necessarily empty, so we keep the positive contribution:
 
+```toml
+[environment]
+python-version = "3.12"
+```
+
 ```py
 from typing import Any
 from ty_extensions import Intersection, Not, Unknown
@@ -924,6 +929,50 @@ def unknown(
 ) -> None:
     reveal_type(i1)  # revealed: Unknown
     reveal_type(i2)  # revealed: Unknown
+
+class Covariant[T]:
+    def get(self) -> T:
+        raise NotImplementedError
+
+def covariant(
+    i1: Intersection[Covariant[Any], Not[Covariant[Any]]],
+    i2: Intersection[Not[Covariant[Any]], Covariant[Any]],
+) -> None:
+    reveal_type(i1)  # revealed: Covariant[Any]
+    reveal_type(i2)  # revealed: Covariant[Any]
+
+class Contravariant[T]:
+    def receive(self, input: T): ...
+
+def contravariant(
+    i1: Intersection[Contravariant[Any], Not[Contravariant[Any]]],
+    i2: Intersection[Not[Contravariant[Any]], Contravariant[Any]],
+) -> None:
+    reveal_type(i1)  # revealed: Contravariant[Any]
+    reveal_type(i2)  # revealed: Contravariant[Any]
+
+class Invariant[T]:
+    mutable_attribute: T
+
+def invariant(
+    i1: Intersection[Invariant[Any], Not[Invariant[Any]]],
+    i2: Intersection[Not[Invariant[Any]], Invariant[Any]],
+) -> None:
+    reveal_type(i1)  # revealed: Invariant[Any]
+    reveal_type(i2)  # revealed: Invariant[Any]
+
+class Bivariant[T]: ...
+
+# Because of bivariance, the specialisation here is meaningless;
+# `Bivariant[Any]` is arguably still a fully static type, even
+# though it is specialised with a gradual form! Thus self-cancellation
+# here is fine:
+def bivariant(
+    i1: Intersection[Bivariant[Any], Not[Bivariant[Any]]],
+    i2: Intersection[Not[Bivariant[Any]], Bivariant[Any]],
+) -> None:
+    reveal_type(i1)  # revealed: Never
+    reveal_type(i2)  # revealed: Never
 ```
 
 ### Mixed dynamic types
--- a/crates/ty_python_semantic/src/types/builder.rs
+++ b/crates/ty_python_semantic/src/types/builder.rs
@@ -958,6 +958,12 @@ impl<'db> InnerIntersectionBuilder<'db> {
                         self.positive.insert(Type::Never);
                         return;
                     }
+                    // `T & ~T` = `T` if `T` is not a subtype of `T`
+                    // (this can only be true if `T` is a non-fully-static type)
+                    if existing_negative.is_equivalent_to(db, new_positive) {
+                        to_remove.push(index);
+                        continue;
+                    }
                     // A & ~B = A    if A and B are disjoint
                     if existing_negative.is_disjoint_from(db, new_positive) {
                         to_remove.push(index);
@@ -1046,6 +1052,11 @@ impl<'db> InnerIntersectionBuilder<'db> {
                     if existing_positive.is_subtype_of(db, new_negative) {
                         *self = Self::default();
                         self.positive.insert(Type::Never);
+                        return;
+                    }
+                    // `T & ~T` = `T` if `T` is not a subtype of `T`
+                    // (this can only be true if `T` is a non-fully-static type)
+                    if existing_positive.is_equivalent_to(db, new_negative) {
                         return;
                     }
                     // A & ~B = A    if A and B are disjoint


sharkdp:
> On `main`, we simplify `Any & ~Any` to `Any`, but we simplify `list[Any] & ~list[Any]` to `Never`. This is inconsistent!

I agree

> The justification for the first simplification applies to all non-fully-static types, so we should apply the same treatment to both.

Which justification?

> `T & ~T` cannot simplify to `Never` for any non-fully-static type `T`, but it can simplify to `T`.

I don't think so? That's not even true for fully-static types. From my calculations, `T & ~T` should simplify to **`(Top[T] & ~Bottom[T]) & Any`**, which is only equivalent to `T` if `Bottom[T] = Never`.

* For fully static types `T`, `Top[T]` is equivalent to `Bottom[T]`, and so `Top[T] & ~Bottom[T] = Never`, which makes the whole `T & ~T` expression equivalent to `Never`, as expected.
* For `T = Any`, we have `Top[T] = object` and `Bottom[T] = Never`. And so `(Top[T] & ~Bottom[T]) & Any = (object & ~Never) & Any = Any`, as expected.
* For a gradual type like `T = Any & int` with a bottom materialization of `Never`, we indeed get `T & ~T = (int & ~Never) & Any = int & Any = T`
* But for a gradual type like `T = Any | int`, this is not true. Here, we get `T & ~T = (object & ~int) & Any = ~int & Any`, which is not equivalent to `Any | int`.

AlexWaygood:
> > The justification for the first simplification applies to all non-fully-static types, so we should apply the same treatment to both.
> 
> Which justification?

Apologies, I linked to this on Discord but should have put this in my PR description, and probably also have added better comments alongside the code changes in this PR. I was referring to @carljm's justification [here](https://github.com/astral-sh/ruff/pull/13775#discussion_r1805636567):

> I think, however, that [`Any & ~Any`] should simplify to just `Any`. Adding `Any`, `Unknown`, or `Todo` to the negative side of an intersection should always just add them to the positive side of the intersection instead. The two are equivalent: "must not be in some unknown set of values" is equivalent to "must be in some unknown set of values"; the one "unknown set of values" is just the complement of the other. Since it's an unknown set anyway, we are free to replace it with its complement.

In the same way, we can say that `list[Any]` means "must be in some unknown set of values that is no larger than `Top[list[Any]]`" and `~list[Any]` means "must not be in some unknown set of values that is no larger than `Top[list[Any]]`". Since the two unknown sets have the same upper bound and the same lower bound, we are free to replace `list[Any] & ~list[Any]` with `list[Any]` in the same way that we replace `Any & ~Any` with `Any`. The fact that some unknown set of values must be subtracted from the set of possible values doesn't contribute anything meaningful to our understanding of the type when that unknown set has the same upper and lower bound as the positive contributions to the intersection.

> From my calculations, `T & ~T` should simplify to **`(Top[T] & ~Bottom[T]) & Any`**, which is only equivalent to `T` if `Bottom[T] = Never`.

The "intersect with `Any`" part of this surprises me! Where did you derive this formulation from? But...

> But for a gradual type like `T = Any | int`, this is not true. Here, we get `T & ~T = (object & ~int) & Any = ~int & Any`, which is not equivalent to `Any | int`.

the conclusion here definitely seems correct to me. If `Any` on the r.h.s. materialized to `Never`, we'd get `(int | Any) & ~(int | Never)` -> `(int | Any) & ~int` -> `(int & ~int) | (Any & ~int)` -> `Never | (Any & ~int)` -> `Any`. And if `Any` on the r.h.s. materialized to `object`, we'd get `(int | Any) & ~(int | object)` -> `(int | Any) & ~object` -> `Never`. The union of the two is `(Any & ~int) | Never` -> `Any & ~int`.

Thinking through the formulation like that makes me wonder if we can simply say that `T & ~U` should always be equivalent to `(T & ~Top[U]) | (T & ~Bottom[U])`? That feels more appealing to me than "randomly" intersecting with `Any` 😄. Working through some other test cases:
- Fully static types: `int & ~int` -> `(int & ~Top[int]) | (int & ~Bottom[int])` -> (int & ~int) | (int & ~int)` -> `Never | Never` -> `Never`
- `Any` itself: `Any & ~Any` -> `(Any & ~Top[Any]) | (Any & ~Bottom[Any])` -> `(Any & ~object) | (Any & ~Never)` -> `Never | Any` -> `Any`
- invariant generics specialized by `Any`: `list[Any] & ~list[Any]` -> `(list[Any] & ~Top[list[Any]]) | (list[Any] & ~Bottom[list[Any]])` -> ???
- intersections with `Any` contributions: `(Any & int) & ~(Any & int)` -> `(Any & int & ~Top[Any & int]) | (Any & int & ~Bottom[Any & int])` -> `(Any & int & ~int) | (Any & int & ~Never)` -> `Never | (Any & int)` -> `Any & int`

sharkdp:
> "must not be in some unknown set of values" is equivalent to "must be in some unknown set of values"

I like these sort of phrases as a justification after the fact, but I don't think they are good tools for a proof.

> The "intersect with `Any`" part of this surprises me! Where did you derive this formulation from?

The key technique for any of these calculations is to write gradual types in the form `T = Top[T] & Any | Bottom[T]` with `Bottom[T] <: Top[T]` the two fully-static types that completely characterize the gradual type `T`. The fact that this can be done for any gradual type is non-trivial (see https://www.irif.fr/~gc/papers/set-theoretic-types-2022.pdf, p. 53, *"Given a gradual type τ, the set of all static types it materializes to forms a complete lattice, with a maximum and a minimum static type that we denote by τ⇑ and τ⇓, respectively."*).

But once we have this representation, we can start to do useful calculations. For example, we can rewrite `~T` to resemble the universal form:

```
~T = ~(Top[T] & Any | Bottom[T])
   = ~(Top[T] & Any) & ~Bottom[T]
   = (~Top[T] | ~Any) & ~Bottom[T]
   = ~Top[T] & ~Bottom[T] | ~Any & ~Bottom[T]
   = ~Top[T] | Any & ~Bottom[T]
   = ~Bottom[T] & Any | ~Top[T]
```

And from this representation, we can extract that `Top[~T] = ~Bottom[T]` and `Bottom[~T] = ~Top[T]`. In a similar way, we can prove that `Top[…]` and `Bottom[…]` distribute over intersections (and unions), i.e. `Top[A & B] = Top[A] & Top[B]`. Once we have that, we can compute:

```
Top[T & ~T] = Top[T] & Top[~T] = Top[T] & ~Bottom[T]
Bottom[T & ~T] = Bottom[T] & Bottom[~T] = Bottom[T] & ~Top[T] = Never
```

And finally we can rewrite this in the universal form

```
T & ~T = Top[T & ~T] & Any | Bottom[T & ~T]
       = (Top[T] & ~Bottom[T]) & Any | Never
       = (Top[T] & ~Bottom[T]) & Any
```

What this type shows is that `Never` is *always* a possible materialization of `T & ~T`. But it can also be as "fuzzy" as `Any` (if `T` itself is `Any`). If a gradual type has a bottom materialization of `Never`, then the expression above simplifies to `(Top[T] & ~Bottom[T]) & Any = (Top[T] & object) & Any = Top[T] & Any` which is equivalent to `T`. So whenever the bottom materialization of a type is `Never`, `T & ~T = T`.

By the way, we could also compute `T | ~T = Any | (Bottom[T] | ~Top[T])`, which demonstrates the usual duality between intersections and unions. Here, `object` is *always* a possible materialization, but it can be as fuzzy as `Any` (if `T` itself is `Any`).

sharkdp:
> Thinking through the formulation like that makes me wonder if we can simply say that `T & ~U` should always be equivalent to `(T & ~Top[U]) | (T & ~Bottom[U])`?

That doesn't look correct. Take `T = object` and `U = Any`. `T & ~U = object & ~Any = object & Any = Any`. But your expression gives `(object & Never) | (object & object) = object`.

I think it should be `T & ~U = (T & ~Top[U]) | (T & ~Bottom[U] & Any)`. The `& Any` was missing.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #20651). The diff contains 96 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #20651). The diff contains 96 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

PR #20653, opened 2025-09-30
[ty] Fix flaky constraint set rendering

dcreager:
This doesn't seem to be flaky in the sense of tests failing non-deterministically, but they are flaky in the sense of unrelated changes causing testing failures from the clauses of a constraint set being rendered in different orders. This flakiness is because we're using Salsa IDs to determine the order in which typevars appear in a constraint set BDD, and those IDs are assigned non-deterministically.

The fix is ham-fisted but effective: sort the constraints in each clause, and the clauses in each set, as part of the rendering process. Constraint sets are only rendered in our test cases, so we don't need to over-optimize this.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -1209,7 +1209,7 @@ impl<'db> SatisfiedClause<'db> {
         false
     }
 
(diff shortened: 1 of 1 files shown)


ibraheemdev (approved):

sharkdp:
Merging this, as I'm seeing it on some other PRs too.

PR #20657, opened 2025-10-01
[`ruff`] Add support for additional eager conversion patterns (`RUF065`)

### LINKED ISSUE #20583: RUF065 should not apply to the combination of `%r` and `repr`
### Summary

[`logging-eager-conversion` (RUF065)](https://docs.astral.sh/ruff/rules/logging-eager-conversion/) applies when a value is doubly converted with `%r` and `repr`. It should only apply to single conversions. [Example](https://play.ruff.rs/d6f8b38f-8d0e-4554-b377-10c2403082b9):
```console
$ cat >ruf065.py <<'# EOF'
import logging
logging.warning("%r", repr("1\n2"))
# EOF

$ python3.14 ruf065.py
WARNING:root:"'1\\n2'"

$ ruff --isolated check ruf065.py --select RUF065 --preview --output-format concise -q
ruf065.py:2:23: RUF065 Unnecessary `repr()` conversion when formatting with `%r`
```

Removing `repr`, as the rule suggests, changes the program’s behavior:
```console
$ cat >ruf065_fixed.py <<'# EOF'
import logging
logging.warning("%r", "1\n2")
# EOF

$ python3.14 ruf065_fixed.py
WARNING:root:'1\n2'
```

The combinations that are currently supported and that would be appropriate to support (not that it’s required, but you might want to) are:

| Code | Function | Currently supported | Appropriate to support | Replacement code |
| --- | --- | --- | --- | --- |
| `%s` | `str` | ✅ Yes | ✅ Yes | `%s` |
| `%r` | `repr` | ✅ Yes | ❌ No | N/A |
| `%s` | `repr` | ❌ No | ✅ Yes | `%r` |
| `%s` | `ascii` | ❌ No | ✅ Yes | `%a` |
| `%s` | `oct` | ❌ No | ✅ Yes | `%#o` |
| `%s` | `hex` | ❌ No | ✅ Yes | `%#x` |

### Version

danparizher:
## Summary

Fixes #20583

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/ruff/RUF065.py
+++ b/crates/ruff_linter/resources/test/fixtures/ruff/RUF065.py
@@ -43,3 +43,34 @@ logging.warning("Value: %r", repr(42))
 logging.error("Error: %r", repr([1, 2, 3]))
 logging.info("Debug info: %s", repr("test\nstring"))
 logging.warning("Value: %s", repr(42))
+
+# %s + ascii()
+logging.info("ASCII: %s", ascii("Hello\nWorld"))
+logging.warning("ASCII: %s", ascii("test"))
+
+# %s + oct()
+logging.info("Octal: %s", oct(42))
+logging.warning("Octal: %s", oct(255))
+
+# %s + hex()
+logging.info("Hex: %s", hex(42))
+logging.warning("Hex: %s", hex(255))
+
+# %b + bytes()
+logging.info("Bytes: %b", bytes("Hello", "utf-8"))
+logging.warning("Bytes: %b", bytes("test", "utf-8"))
+
+# Test with imported functions
+from logging import info, log
+
+info("ASCII: %s", ascii("Hello\nWorld"))
+log(logging.INFO, "ASCII: %s", ascii("test"))
+
+info("Octal: %s", oct(42))
+log(logging.INFO, "Octal: %s", oct(255))
+
+info("Hex: %s", hex(42))
+log(logging.INFO, "Hex: %s", hex(255))
+
+info("Bytes: %b", bytes("Hello", "utf-8"))
+log(logging.INFO, "Bytes: %b", bytes("test", "utf-8"))
--- a/crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
+++ b/crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
@@ -62,19 +62,24 @@ use crate::rules::flake8_logging_format::rules::{LoggingCallType, find_logging_c
 #[derive(ViolationMetadata)]
 pub(crate) struct LoggingEagerConversion {
     pub(crate) format_conversion: FormatConversion,
+    pub(crate) function_name: Option<String>,
 }
 
 impl Violation for LoggingEagerConversion {
     #[derive_message_formats]
     fn message(&self) -> String {
-        let LoggingEagerConversion { format_conversion } = self;
-        let (format_str, call_arg) = match format_conversion {
-            FormatConversion::Str => ("%s", "str()"),
-            FormatConversion::Repr => ("%r", "repr()"),
-            FormatConversion::Ascii => ("%a", "ascii()"),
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/resources/test/fixtures/ruff/RUF065.py
```diff
@@ -43,3 +43,34 @@ def str(s): return f"str = {s}"
...
+logging.info("Octal: %s", oct(42))
+logging.warning("Octal: %s", oct(255))
+
+# %s + hex()
+logging.info("Hex: %s", hex(42))
+logging.warning("Hex: %s", hex(255))
+
+# %b + bytes()
+logging.info("Bytes: %b", bytes("Hello", "utf-8"))
+logging.warning("Bytes: %b", bytes("test", "utf-8"))
```
ntBre: I'm not sure if we should be handling bytes. I might be doing something wrong, but these don't appear to work on Python 3.13:

```pycon
>>> logging.info("Bytes: %b", bytes("Hello", "utf-8"))
--- Logging error ---
  File "/usr/lib/python3.13/logging/__init__.py", line 1151, in emit
    msg = self.format(record)
  File "/usr/lib/python3.13/logging/__init__.py", line 999, in format
    return fmt.format(record)
           ~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.13/logging/__init__.py", line 712, in format
    record.message = record.getMessage()
                     ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/logging/__init__.py", line 400, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
ValueError: unsupported format character 'b' (0x62) at index 8
```

This happens even if I make the message a byte string, which otherwise works with `%` formatting:

```pycon
>>> logging.info(b"Bytes: %b", bytes("Hello", "utf-8"))
--- Logging error ---
# ...
>>> b"Bytes: %b" % bytes("Hello", "utf-8")
b'Bytes: Hello'
```

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -62,19 +62,24 @@ use crate::rules::flake8_logging_format::rules::{LoggingCallType, find_logging_c
...
-            FormatConversion::Ascii => ("%a", "ascii()"),
-            FormatConversion::Bytes => ("%b", "bytes()"),
-        };
-        format!("Unnecessary `{call_arg}` conversion when formatting with `{format_str}`")
+        let LoggingEagerConversion {
+            format_conversion,
+            function_name,
+        } = self;
+        match (format_conversion, function_name.as_deref()) {
+            (FormatConversion::Str, Some("oct")) => "Unnecessary `oct()` conversion when formatting with `%s`. Use `%#o` instead of `%s`".to_string(),
```
ntBre: Could we break these lines up? We can't insert actual newlines since the message needs to remain on a single line, but we can do something like:


```suggestion
            (FormatConversion::Str, Some("oct")) => "Unnecessary `oct()` conversion when formatting with `%s`. \
            Use `%#o` instead of `%s`".to_string(),
```

Alternatively, I think the `Use ...` parts would fit nicely in the `fix_title` function, which I think we can include even if we don't attach a fix.

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -62,19 +62,24 @@ use crate::rules::flake8_logging_format::rules::{LoggingCallType, find_logging_c
 #[derive(ViolationMetadata)]
 pub(crate) struct LoggingEagerConversion {
     pub(crate) format_conversion: FormatConversion,
+    pub(crate) function_name: Option<String>,
```
ntBre: Does `function_name` need to be a `String`? It looks like it's only ever set to a `&'static str`.

I also think it might make sense to define our own, separate enum here instead of reusing `FormatConversion`, especially if we don't support `Bytes`, as in my other comment.

Issue #20512, opened 2025-09-22
FA102 is missing `collection.abc` types


Andrej730:
### Summary

E.g. use of `Sequence[int]` is not flagged. 
See example below (command I've used is `ruff check test.py --target-version py38 --select=FA102`)

Ruff playground - https://play.ruff.rs/b295c8a4-4295-4a5a-8a2d-e99699814822
```python
from collections.abc import Sequence

# Not reported by Ruff, but results in runtime error:
# TypeError: 'ABCMeta' object is not subscriptable
def test() -> Sequence[int]:
    return [1, 2, 3]


# FA102 Missing `from __future__ import annotations`, but uses PEP 585 collection
def test2() -> list[int]:
    return [1, 2, 3]
```

### Version

ruff 0.13.1

ntBre:
This makes sense to me, but it will have to be a preview change since it expands the scope of a stable rule.

Issue #15251, opened 2025-01-04
Revisit adding all PEP585 names to UP006


MichaReiser:
I have raised a few issues with this change:

* https://github.com/astral-sh/ruff/issues/15246

  Even if the detection+fix ability is better now that these are implemented in `UP006`, something needs to be done about the fact that the newly added warnings are all duplicates of warnings and fixes that already exist in `UP035`.

  If this implementation becomes a little more polished, it probably should completely replace the implementation that's currently in `UP035`, there's really no reason to still keep that inferior implementation other than its battle-tested status. Although maybe for historic consistency, even if the *implementation* is unified towards `UP006`, the set of items that it triggers for should remain split between `UP006` (covering only builtins such as `list`) and `UP035` (covering others) like it's always been.

* https://github.com/astral-sh/ruff/issues/15245

  The fixer has bugs where it sometimes fails to kick in, and in corner cases kicks in with strange results.
  
For now I'd like to see this pull request reverted and relanded later after a more detailed discussion, because it has ongoing bugs and possibly introduces undesirable/unexpected changes to what the rules are supposed to cover.

_Originally posted by @oprypin in https://github.com/astral-sh/ruff/issues/5454#issuecomment-2569976304_

PR #15250, opened 2025-01-04
Revert "Add all PEP-585 names to UP006 rule"

MichaReiser:
Reverts astral-sh/ruff#5454

PR #5454, opened 2023-06-30
Add all PEP-585 names to UP006 rule

wookie184:
Closes #3936 

## Summary

Building on https://github.com/astral-sh/ruff/pull/4420, adds the remaining PEP-585 names.

## Test Plan

Added snapshot tests for a few cases.

PR #20659, opened 2025-10-01
[`ruff`] Extend FA102 with listed PEP 585-compatible APIs

terror:
Resolves https://github.com/astral-sh/ruff/issues/20512

This PR expands FA102’s preview coverage to flag every PEP 585-compatible API that breaks without from `from __future__ import annotations`, including `collections.abc`. The rule now treats asyncio futures, pathlib-style queues, weakref containers, shelve proxies, and the full `collections.abc` family as generics once preview mode is enabled. 

Stable behavior is unchanged; the broader matching runs behind `is_future_required_preview_generics_enabled`, letting us vet the new diagnostics before marking them as stable. 

I've also added a snapshot test that covers all of the newly supported  types.

Check out https://docs.python.org/3/library/stdtypes.html#standard-generic-classes for a list of commonly used PEP 585-compatible APIs.


crates/ruff_linter/src/preview.rs
```diff
@@ -200,6 +200,11 @@ pub(crate) const fn is_optional_as_none_in_union_enabled(settings: &LinterSettin
     settings.preview.is_enabled()
 }
 
+// flake8-future-annotations: extend FA102 to `collections.abc` generics.
```
ntBre: nit: this comment is generally the PR link. The description seems nice to have too, but the link is handy when working on stabilizations.

crates/ruff_linter/src/rules/flake8_future_annotations/mod.rs
```diff
@@ -50,6 +52,7 @@ mod tests {
             Path::new("flake8_future_annotations").join(path).as_path(),
             &settings::LinterSettings {
                 unresolved_target_version: PythonVersion::PY37.into(),
+                preview: PreviewMode::Enabled,
```
ntBre, line 69: We usually add a second test function like `fa102_preview` and only enable preview there. That way we can see the diff between the preview and non-preview behavior. This case is pretty straightforward, but I'd still prefer that.

crates/ruff_python_semantic/src/analyze/typing.rs
```diff
@@ -148,14 +148,25 @@ pub fn to_pep585_generic(expr: &Expr, semantic: &SemanticModel) -> Option<Module
...
+) -> bool {
     semantic
         .resolve_qualified_name(expr)
-        .is_some_and(|qualified_name| {
-            let [module, name] = qualified_name.segments() else {
-                return false;
-            };
-            has_pep_585_generic(module, name)
+        .is_some_and(|qualified_name| match qualified_name.segments() {
+            [module, member] if has_pep_585_generic(module, member) => true,
```
ntBre: I think it might make sense to inline `has_pep_585_generic` here and remove its comment about `as_pep_585_generic`. Unless I'm missing something, this is now the only reference to `has_pep_585_generic`.

I also found a few issues related to some of the other PEP 585 helpers here, but they seem unrelated to FA102. It looks like two of the PEP 585 helpers are only used by FA102 and two only used by UP006.

- https://github.com/astral-sh/ruff/pull/5454
- https://github.com/astral-sh/ruff/pull/15250
- https://github.com/astral-sh/ruff/issues/15251

crates/ruff_python_semantic/src/analyze/typing.rs
```diff
@@ -148,14 +148,25 @@ pub fn to_pep585_generic(expr: &Expr, semantic: &SemanticModel) -> Option<Module
...
+            [
+                "collections",
+                "abc",
+                "AsyncGenerator" | "AsyncIterable" | "AsyncIterator" | "Awaitable" | "ByteString"
+                | "Callable" | "Collection" | "Container" | "Coroutine" | "Generator" | "ItemsView"
+                | "Iterable" | "Iterator" | "KeysView" | "Mapping" | "MappingView"
+                | "MutableMapping" | "MutableSequence" | "MutableSet" | "Reversible" | "Sequence"
+                | "Set" | "ValuesView",
+            ] => include_collections_abc,
```
ntBre: Should we go ahead and add the other [PEP 585](https://peps.python.org/pep-0585/#implementation)/[standard generic aliases](https://docs.python.org/3/library/stdtypes.html#standard-generic-classes) while we're here?

That's actually how I found the related issues above, but it seems safer to do here than in the UP rule helpers.

crates/ruff_linter/src/preview.rs
terror: No worries, I'll switch that over!

crates/ruff_python_semantic/src/analyze/typing.rs
terror: Yeah I was wondering why that was separated, I'll go ahead and inline it.

crates/ruff_python_semantic/src/analyze/typing.rs
terror: Yep that makes sense to me, better to catch all cases here than having another extension to this rule later on.

terror:
@ntBre @amyreese I've re-purposed this PR to include listed PEP 585-compatible APIs [here](https://docs.python.org/3/library/stdtypes.html#standard-generic-classes). I've added a preview/non-preview test for all of these as well.

ntBre (approved):
Very nice, thank you!

PR #20660, opened 2025-10-01
[`pyupgrade`] Fix false negative for `TypeVar` with default argument in `non-pep695-generic-class` (`UP046`)

### LINKED ISSUE #20656: `non-pep695-generic-class` (`UP046`) - false negative on old `TypeVar` with `default` argument
### Summary

```py
from typing import TypeVar, Generic

T = TypeVar("T", default=int)
U = TypeVar("U")


class Foo(Generic[T]): ... # no error
class Bar(Generic[U]): ... # error
```

https://play.ruff.rs/31611e2c-077c-4291-bdeb-dd663a0f43f7

this behavior is mentioned in the [known problems section](https://docs.astral.sh/ruff/rules/non-pep695-generic-class/#known-problems) but i assume this was written before this feature was widely supported by type checkers, and i couldn't find an existing issue for this.

### Version

0.13.2

danparizher:
## Summary

Fixes #20656

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP046_0.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP046_0.py
@@ -122,7 +122,7 @@ class MixedGenerics[U]:
         return (u, t)
 
 
-# TODO(brent) default requires 3.13
+# default requires 3.13
 V = TypeVar("V", default=Any, bound=str)
 
 
--- a/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP047.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyupgrade/UP047.py
@@ -44,15 +44,15 @@ def any_str_param(s: AnyStr) -> AnyStr:
     return s
 
 
-# these cases are not handled
-
-# TODO(brent) default requires 3.13
+# default requires 3.13
 V = TypeVar("V", default=Any, bound=str)
 
 
 def default_var(v: V) -> V:
     return v
 
+
+# these cases are not handled
 
 def outer():
     def inner(t: T) -> T:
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/src/rules/pyupgrade/rules/pep695/mod.rs
```diff
@@ -133,66 +137,62 @@ impl<'a> From<&'a TypeVar<'a>> for TypeParam {
...
+                                    ctx: ast::ExprContext::Load,
+                                }),
+                            ],
+                            ctx: ast::ExprContext::Load,
+                            parenthesized: true,
+                        })))
+                    }
+                    None => None,
+                },
+                default: default.map(|expr| Box::new((*expr).clone())),
```
ntBre: tiny nit: Is this shared by all the match arms? If so we could factor it out into a shared `default` variable. I also think we can do without the explicit deref but maybe clippy complains.


```suggestion
                default: default.map(|expr| Box::new(expr.clone())),
```

crates/ruff_linter/src/rules/pyupgrade/rules/pep695/mod.rs
```diff
@@ -373,15 +373,8 @@ fn check_type_vars(vars: Vec<TypeVar<'_>>) -> Option<Vec<TypeVar<'_>>> {
...
-    // found on the type parameters
-    (vars
-        .iter()
-        .unique_by(|tvar| tvar.name)
-        .filter(|tvar| tvar.default.is_none())
-        .count()
-        == vars.len())
-    .then_some(vars)
+    // can't predict what the user wanted.
+    (vars.iter().unique_by(|tvar| tvar.name).count() == vars.len()).then_some(vars)
```
ntBre, line 387: We need to preview gate this as I mentioned on the issue. I think this might be the easiest place to do that, assuming every usage goes through this code path.

crates/ruff_linter/src/rules/pyupgrade/snapshots/ruff_linter__rules__pyupgrade__tests__UP047.py.snap
ntBre, line 1: As demonstrated here, this change also affects UP047. I think it should also affect UP040, so we should make sure we have test coverage for that as well.

PR #20661, opened 2025-10-01
[`ty`] Fix playground crashes when accessing vendored files with leading slashes

danparizher:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1290


MeGaGiGaGon:
Thanks for putting up with my silly. A couple things:
1. I think you linked the wrong issue for closing, it should be https://github.com/astral-sh/ty/issues/1290
2. Looking at the [docs for `Utf8Component`](https://docs.rs/camino/latest/camino/enum.Utf8Component.html), it looks like the only variant that can hit the `unsupported` match after this is `Prefix`. I tested a bunch of paths like `vendored:\\?\` and `vendored:C:\`, but could not trigger a crash with hitting it, so I guess it's fine? Though it may make since to not panic in there either just in-case it is reachable.

Edit: Looking at the docs more, for `Prefix` it says:
> Does not occur on Unix.

So if the playground is being hosted on a Unix machine (which I image it is), that branch would be unreachable, but it could still crash on windows.

MichaReiser:
What's the code path where we call into `VendoredPath::from` that then crashes? I wonder if we should fix this at the API boundary instead.

danparizher:
> What's the code path where we call into `VendoredPath::from` that then crashes? I wonder if we should fix this at the API boundary instead.

The crash originated from the playground’s vendored URI flow. The path is:
1) Frontend creates `vendored://...` URIs and extracts a path string (can start with “/”) for vendored files.
2) That string crosses the WASM boundary via `getVendoredFile(path)` and is passed directly to `VendoredPath::new(path)`.

https://github.com/astral-sh/ruff/blob/caf48f4bfc54bb587342bf794a4f5167d855f529/crates/ty_wasm/src/lib.rs#L605-L620

3) Inside vendored FS, we normalize via `NormalizedVendoredPath::from(&VendoredPath)`. Previously, a leading “/” became a `RootDir` component and hit the “unsupported” arm, which panicked. We now ignore `RootDir` and Windows `Prefix` components so paths are treated as relative to the archive root instead of panicking.

https://github.com/astral-sh/ruff/blob/1f1c59709da6af09bbd9c6b9030314945ca0603a/crates/ruff_db/src/vendored.rs#L393-L453

MichaReiser:
> Frontend creates vendored://... URIs and extracts a path string (can start with “/”) for vendored files.

Yes, I think we should prevent the UI from creating vendored file paths similar to https://github.com/astral-sh/ruff/pull/20666

danparizher:
Would we like to remove the logic from `crates/ruff_db/src/vendored.rs` or keep it as a fallback?

playground/ty/src/Editor/Editor.tsx
```diff
@@ -206,7 +206,8 @@ class PlaygroundServer
   private getVendoredPath(uri: Uri): string {
     // Monaco parses "vendored://stdlib/typing.pyi" as authority="stdlib", path="/typing.pyi"
     // We need to reconstruct the full path
-    return uri.authority ? `${uri.authority}${uri.path}` : uri.path;
+    const fullPath = uri.authority ? `${uri.authority}${uri.path}` : uri.path;
+    return fullPath.startsWith("/") ? fullPath.slice(1) : fullPath;
   }
```
MichaReiser, line 210: I don't think this is the right place. we need to prevent users from creating a file named `vendored://` , similar to what we've done in https://github.com/astral-sh/ruff/pull/20666

MichaReiser (approved):
Thanks

PR #20662, opened 2025-10-01
['ruff'] Fix flynt f-string quoting for mixed quote joiners

### LINKED ISSUE #19837: FLY002 fix causes syntax error with double quotes in the strings

TaKO8Ki:
## Summary

Fixes #19837

Track quote usage across the joiner and parts to choose a safe f-string quote or skip the fix when both appear.

## Test Plan

Add regression coverage to FLY002.py

### DIFF
--- a/crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
+++ b/crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
@@ -2,7 +2,7 @@ use ast::FStringFlags;
 use itertools::Itertools;
 
 use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::{self as ast, Arguments, Expr, StringFlags};
+use ruff_python_ast::{self as ast, Arguments, Expr, StringFlags, str::Quote};
 use ruff_text_size::{Ranged, TextRange};
 
 use crate::checkers::ast::Checker;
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -126,14 +128,31 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
...
+            has_double_quote |= value.contains('"');
+        }
+        f_string_elements.push(element);
     }
 
+    let quote = flags.quote_style();
+    let adjusted_quote = match quote {
+        Quote::Single if has_single_quote && !has_double_quote => quote.opposite(),
+        Quote::Double if has_double_quote && !has_single_quote => quote.opposite(),
+        _ if has_double_quote && has_single_quote => return None,
```
TaKO8Ki, line 147: If the content includes both single quote and double quote, ruff skips FLY002 for now.

I will improve this logic in a separate pull request as we discussed here.

https://github.com/astral-sh/ruff/pull/20197#pullrequestreview-3240917942

PR #20664, opened 2025-10-01
[ty] No union with `Unknown` for module-global symbols (no promotion)

sharkdp:
# not ready for review


crates/ty_python_semantic/src/place.rs
```diff
@@ -822,18 +822,37 @@ fn place_by_id<'db>(
...
+            // who need the external mutation to work can always annotate the global with the
+            // wider type. And everyone else benefits from more precise type inference.
+            let is_module_global = scope.node(db).scope_kind().is_module();
+
+            // If the visibility of the scope is private (like for a function scope), we also do
+            // not union with `Unknown`, because the symbol cannot be modified externally.
+            let scope_has_private_visibility = scope.scope(db).visibility().is_private();
+
+            // We generally trust undeclared places in stubs and do not union with `Unknown`.
+            let in_stub_file = scope.file(db).is_stub(db);
```
sharkdp, line 844: I noticed that we can now remove this exception here, but we'll evaluate that in a separate PR, to make sure it does not have any effects on the ecosystem.

AlexWaygood:
> ## Typing conformance
> 
> ```diff
> +protocols_modules.py:25:1: error[invalid-assignment] Object of type `<module '_protocols_modules1'>` is not assignable to `Options1`
> ```
> 
> This diagnostic should apparently not be there, but it looks like we also fail other tests in that file, so it seems to be a limitation that was previously hidden by `Unknown` somehow.

That's... an interesting test. Conceptually it seems to me to be the same as this, where we (correctly, I think) state that `Nominal` should not be considered assignable to `Proto`, since mutable instance attributes must be treated invariantly:

```py
from typing import Protocol, Literal, reveal_type
from ty_extensions import is_assignable_to

class Nominal:
    x: Literal[True]

class Proto(Protocol):
    x: bool

reveal_type(is_assignable_to(Nominal, Proto))  # revealed: ConstraintSet[never]
```

[Pyright](https://pyright-play.net/?pyrightVersion=1.1.405&strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAK4MYAxmADYA0UAMkjAKYgCGFAsAFBekUsDO-KADlIqNgC4uUGVAAeEug2ZsA2gBUQAV0YBdLjz6DCxMAAoiYEuQoBKKZ1nzFAIzCUDnACaNgUYGYKJlZg9lAAdJGePn5ogYqiEOJ2Dk4BcrZcQA), [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=b353f59de514ef4283a968aa09197583) and [pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA0dADIQGMSqigAddGtFRUcOHQBytVssRq6ZuvkSz5i5QG0AKpQCuMALpqNWnYOG4AFEK4IuJQAJQm6OYWVngSnlgwYHRg-pa%2BwbgRdIS5CZhJdGxpVgY0RuGR0an4YWogkiDODNBwJOSIIADEdACqLVDypCnO6KItuOhw%2BYVgvDSoDAD66M402IolnOgMYXQAtAB8dHAMlFXmlDAMzpRRYCogemsb53TA%2BAC%2Bj-WfjajjCAANxgADFoDAKGgsHgiGQQJ8gA&version=3.12) all agree with us on the class example, so I wonder why the spec appears to assert that modules should be treated differently here

[Pyrefly also agrees with us](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0AFag1wBjXFAA66aVx586NXJgCusadNFRUcOIOG4AFENwjxUAJSJpdW3XyI6eCRqwwwddDBiY4AfWIDQwd9U1wrOkIo1y8ff0CwwyVVWAtpEAAaEGS1GBJyRBB8OgBeOgAVShUYdIBfLNRRBggANxgAMWgYCjQsPCIyEFqgA&version=3.12) that the module should not be considered an instance of this protocol.

AlexWaygood:
Okay, both mypy and pyright do have internally consistent behaviour here. They both consider the module to inhabit the protocol because they do both promote the type of the module constant to `bool` rather than `Literal[True]` when viewed from external scopes.

I don't think the purpose of the test here is to assert that type checkers must promote the types of module constants when viewed from external scopes, however. I think it just wasn't considered that other type checkers might have different behaviour here, because the two established type checkers had the same behaviour when the test was written. I think this is easily addressed by adding type annotations to the global-scope attributes in `_protocols_module_1.py` [here](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/_protocols_modules1.py)

crates/ty_python_semantic/resources/mdtest/import/dunder_all.md
```diff
@@ -783,8 +783,7 @@ class A: ...
 ```py
 from subexporter import *
 
-# TODO: Should we avoid including `Unknown` for this case?
```
AlexWaygood, line 786: we could keep this TODO -- we might still want to consider infering `__all__` in the module scope as being "implicitly declared" as `list[str] | tuple[str, ...]`, since it's a very special symbol that's bestowed special semantics by Python at runtime

sharkdp:
> I think this is easily addressed by adding type annotations to the global-scope attributes in `_protocols_module_1.py` [here](https://github.com/python/typing/blob/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance/tests/_protocols_modules1.py)

Hm. This does sound very much like "adding a type annotation to get rid of a type check error". So maybe there is something that we should do about it. Do you think it's something that can wait until after this PR has been merged? (I might want to do the literal-promotion analysis anyway)

crates/ty_python_semantic/src/place.rs
```diff
@@ -822,18 +822,37 @@ fn place_by_id<'db>(
...
-                // meaning that the place cannot be rewritten from elsewhere, we do not union with `Unknown`.
-
+            // Module-level globals can be mutated externally. A `MY_CONSTANT = 1` global might
+            // be changed to `"some string"` from code outside of the module that we're looking
+            // at, and so from a gradual-guarantee perspective, it makes sense to infer a type
+            // of `Literal[1] | Unknown` for global symbols. This allows the code that does the
+            // mutation to type check correctly, and for code that uses the global, it accurately
+            // reflects the lack of knowledge about the type.
+            //
+            // External modifications (or modifications through `global` statements) that would
```
AlexWaygood:
```suggestion
            // However, external modifications (or modifications through `global` statements) that would
```

PR #20666, opened 2025-10-01
[`ty`] Reject renaming files to start with slash in Playground

TaKO8Ki:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1282

Block renaming to names starting with `/`, surfacing a clear error instead of crashing the playground.

## Test Plan

https://github.com/user-attachments/assets/94848f31-9fcf-46b7-895e-f468e5030104

### DIFF
--- a/playground/ty/src/Playground.tsx
+++ b/playground/ty/src/Playground.tsx
@@ -96,6 +96,11 @@ export default function Playground() {
     file: FileId,
     newName: string,
   ) => {
+    if (newName.startsWith("/")) {
+      setError("File names cannot start with '/'.");
+      return;
+    }
+
     const handle = files.handles[file];
     let newHandle: FileHandle | null = null;
     if (handle == null) {


playground/ty/src/Playground.tsx
```diff
@@ -96,6 +96,11 @@ export default function Playground() {
     file: FileId,
     newName: string,
   ) => {
+    if (newName.startsWith("/")) {
+      setError("File names cannot start with '/'.");
```
MichaReiser, line 100: Does the error message get cleared at some point?

TaKO8Ki: As shown in the video, the error message disappears as soon as you type anything. Actions like creating a file don’t dismiss it, and this behavior is the same for other error messages as well.

https://github.com/user-attachments/assets/adad7437-bf7b-40fb-ac75-aecfed51318d

PR #20667, opened 2025-10-01
[ty] Reformulation of public symbol inference test suite

sharkdp:
## Summary

Reformulation of the public symbol type inference test suite to use class scopes instead of module scopes. This is in preparation for an upcoming change to module-global scopes (#20664).

## Test Plan

Updated tests


crates/ty_python_semantic/resources/mdtest/boundness_declaredness/public.md
```diff
@@ -224,73 +194,59 @@ If a symbol is *undeclared*, we use the union of `Unknown` with the inferred typ
...
+        a: int = 1
 
-reveal_type(a)  # revealed: Unknown
+# error: [unresolved-attribute]
+reveal_type(Public.a)  # revealed: Unknown
 
-# Modifications allowed in this case:
-a = None
+# Modification attempts yield an error:
+# error: [unresolved-attribute]
```
sharkdp, line 248: This is the only behavioral change. It makes sense that we emit an error for class scopes. For the module-global scope, "modifying `a`" was just creating a new variable.

PR #20669, opened 2025-10-01
[ty] Enums: allow multiple aliases to point to the same member

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1293

## Test Plan

Regression test

### DIFF
--- a/crates/ty_python_semantic/src/types/enums.rs
+++ b/crates/ty_python_semantic/src/types/enums.rs
@@ -194,7 +194,10 @@ pub(crate) fn enum_metadata<'db>(
             // we don't know if it's a duplicate or not.
             if matches!(
                 value_ty,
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types/enums.rs
```diff
@@ -196,10 +196,14 @@ pub(crate) fn enum_metadata<'db>(
                 value_ty,
                 Type::IntLiteral(_) | Type::StringLiteral(_) | Type::BytesLiteral(_)
```
AlexWaygood: Shouldn't this also include `Type::BooleanLiteral(_)`? Looks like we get this wrong: https://play.ty.dev/90e8c00b-fc7b-4a5c-a051-332c5b4893ff

PR #20671, opened 2025-10-01
[ty] Public types: remove special-case for stub files

sharkdp:


sharkdp:
Ok, looks like we shouldn't do this, due to things like:

https://github.com/python/typeshed/blob/6547ec10b881acac28ed000a7c2108f8de1c6a1e/stdlib/inspect.pyi#L343

or this

https://github.com/pandas-dev/pandas-stubs/blob/a9f0bd321ef67cecd4ce0d5aeecd1a87b90ea1a2/pandas-stubs/core/series.pyi#L4007

PR #20675, opened 2025-10-01
[ty] improve base conda distinction from child conda

Danielkonge:
## Summary

#19990 didn't completely fix the base vs. child conda environment distinction, since it detected slightly different behavior than what I usually see in conda. E.g., I see something like the following:
```
(didn't yet activate conda, but base is active)
➜ printenv | grep CONDA
CONDA_PYTHON_EXE=/opt/anaconda3/bin/python
CONDA_PREFIX=/opt/anaconda3
CONDA_DEFAULT_ENV=base
CONDA_EXE=/opt/anaconda3/bin/conda
CONDA_SHLVL=1
CONDA_PROMPT_MODIFIER=(base)

(activating conda)
➜ conda activate test

(test is an active conda environment)
❯ printenv | grep CONDA
CONDA_PREFIX=/opt/anaconda3/envs/test
CONDA_PYTHON_EXE=/opt/anaconda3/bin/python
CONDA_SHLVL=2
CONDA_PREFIX_1=/opt/anaconda3
CONDA_DEFAULT_ENV=test
CONDA_PROMPT_MODIFIER=(test)
CONDA_EXE=/opt/anaconda3/bin/conda
```

But the current behavior looks for `CONDA_DEFAULT_ENV = basename(CONDA_PREFIX)` for the base environment instead of the child environment, where we actually see this equality. 

This pull request fixes that and updates the tests correspondingly.

## Test Plan

I updated the existing tests with the new behavior. Let me know if you want more tests. Note: It shouldn't be necessary to test for the case where we have `conda/envs/base`, since one should not be able to create such an environment (one with the name of `CONDA_DEFAULT_ENV`).

### DIFF
--- a/crates/ty/docs/environment.md
+++ b/crates/ty/docs/environment.md
@@ -35,12 +35,16 @@ ty also reads the following externally defined environment variables:
 
 ### `CONDA_DEFAULT_ENV`
 
-Used to determine if an active Conda environment is the base environment or not.
+Used to determine the name of the active Conda environment.
 
 ### `CONDA_PREFIX`
 
-Used to detect an activated Conda environment location.
+Used to detect the path of an active Conda environment.
 If both `VIRTUAL_ENV` and `CONDA_PREFIX` are present, `VIRTUAL_ENV` will be preferred.
+
+### `_CONDA_ROOT`
+
+Used to determine the root install path of Conda.
 
 ### `PYTHONPATH`
 
--- a/crates/ty/tests/cli/python_environment.rs
+++ b/crates/ty/tests/cli/python_environment.rs
@@ -946,6 +946,7 @@ fn defaults_to_a_new_python_version() -> anyhow::Result<()> {
 /// 2) If `CONDA_PREFIX` environment variable is set (and .filename == `CONDA_DEFAULT_ENV`)
 /// 3) If a `.venv` directory exists at the project root
 /// 4) If `CONDA_PREFIX` environment variable is set (and .filename != `CONDA_DEFAULT_ENV`)
+///    or if `_CONDA_ROOT` is set (and `_CONDA_ROOT` == `CONDA_PREFIX`)
 ///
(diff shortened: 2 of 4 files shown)


Gankra:
Hmm this implementation has seemingly been working fine for uv for quite a while? 

https://github.com/astral-sh/uv/blob/8da9df3654f255a2f2e8e83c51257c0699ece9f7/crates/uv-python/src/virtualenv.rs#L79

Did I invert something in my implementation?

Gankra:
Aha! uv just shipped this fix (and a bonus one) a couple weeks ago.

https://github.com/astral-sh/uv/pull/15679/
https://github.com/astral-sh/uv/pull/15680/

Danielkonge:
> Approving the basic idea, but some more comment/naming cleanups are in-order, ala the equivalent uv PRs. Would you be interested in doing those extra changes? If not I can.

I can try to add these changes later today.

crates/ty/docs/environment.md
```diff
@@ -35,13 +35,17 @@ ty also reads the following externally defined environment variables:
...
-Used to determine if an active Conda environment is the base environment or not.
+Used to determine the name of the active Conda environment.
 
 ### `CONDA_PREFIX`
 
-Used to detect an activated Conda environment location.
+Used to detect the path of an active Conda environment.
 If both `VIRTUAL_ENV` and `CONDA_PREFIX` are present, `VIRTUAL_ENV` will be preferred.
 
+### `_CONDA_ROOT`
```
Danielkonge, line 45: uv writes `CONDA_ROOT` instead of `_CONDA_ROOT`, but the section is "Externally-defined variables", and the actual environment variable is with the `_`, so I wrote it like this.

Danielkonge:
@Gankra: I think I added what the mentioned uv PRs covered now. Let me know if you wanted more than this? 

Also: I left it is in a separate commit for now, but should I just rebase?

Gankra (approved):
Rad, thanks so much!

PR #20676, opened 2025-10-02
[`isort`] Fix inserting required imports before future imports (`I002`)

### LINKED ISSUE #20674: I002 inserts required imports before future imports
### Summary

The fix for [`missing-required-import` (I002)](https://docs.astral.sh/ruff/rules/missing-required-import/) inserts required imports before future imports, which introduces a syntax error. [Example](https://play.ruff.rs/1ea00554-6288-47fa-a3d2-b2c71ad067d1):
```console
$ cat >i002.py <<'# EOF'
# EOF

$ ruff --isolated check i002.py --select I002 --unsafe-fixes --fix --config 'lint.isort.required-imports = ["import this"]'

$ cat i002.py
import this

$ python i002.py 2>&1 | tail -n 1
SyntaxError: from __future__ imports must occur at the beginning of the file
```

### Version

danparizher:
## Summary

Fixes #20674

### DIFF
--- a/crates/ruff_linter/src/importer/mod.rs
+++ b/crates/ruff_linter/src/importer/mod.rs
@@ -75,9 +75,16 @@ impl<'a> Importer<'a> {
             // Insert after the last top-level import.
             Insertion::end_of_statement(stmt, self.source, self.stylist).into_edit(&required_import)
         } else {
-            // Insert at the start of the file.
-            Insertion::start_of_file(self.python_ast, self.source, self.stylist)
-                .into_edit(&required_import)
+            // Check if there are any future imports that we need to respect
+            if let Some(last_future_import) = self.find_last_future_import() {
+                // Insert after the last future import
+                Insertion::end_of_statement(last_future_import, self.source, self.stylist)
+                    .into_edit(&required_import)
+            } else {
+                // Insert at the start of the file.
+                Insertion::start_of_file(self.python_ast, self.source, self.stylist)
+                    .into_edit(&required_import)
+            }
         }
     }
 
@@ -522,6 +529,17 @@ impl<'a> Importer<'a> {
         } else {
             None
         }
+    }
+
+    /// Find the last `from __future__` import statement in the AST.
+    fn find_last_future_import(&self) -> Option<&'a Stmt> {
+        self.python_ast.iter().rev().find(|stmt| {
+            if let Stmt::ImportFrom(import_from) = stmt {
+                import_from.module.as_deref() == Some("__future__")
+            } else {
+                false
+            }
+        })
     }
 
     /// Add a `from __future__ import annotations` import.
--- a/crates/ruff_linter/src/rules/isort/mod.rs
+++ b/crates/ruff_linter/src/rules/isort/mod.rs
@@ -1014,6 +1014,29 @@ mod tests {
         Ok(())
     }
 
+    #[test_case(Path::new("future_import.py"))]
+    fn required_import_with_future_import(path: &Path) -> Result<()> {
+        let snapshot = format!(
+            "required_import_with_future_import_{}",
+            path.to_string_lossy()
+        );
+        let diagnostics = test_path(
+            Path::new("isort/required_imports").join(path).as_path(),
+            &LinterSettings {
+                src: vec![test_resource_path("fixtures/isort")],
+                isort: super::settings::Settings {
+                    required_imports: BTreeSet::from_iter([NameImport::Import(
+                        ModuleNameImport::module("this".to_string()),
+                    )]),
+                    ..super::settings::Settings::default()
+                },
+                ..LinterSettings::for_rule(Rule::MissingRequiredImport)
+            },
+        )?;
+        assert_diagnostics!(snapshot, diagnostics);
+        Ok(())
+    }
+
     #[test_case(Path::new("from_first.py"))]
     fn from_first(path: &Path) -> Result<()> {
         let snapshot = format!("from_first_{}", path.to_string_lossy());
--- a/crates/ruff_linter/src/rules/isort/snapshots/ruff_linter__rules__isort__tests__required_import_existing_import.py.snap
+++ b/crates/ruff_linter/src/rules/isort/snapshots/ruff_linter__rules__isort__tests__required_import_existing_import.py.snap
@@ -4,6 +4,6 @@ source: crates/ruff_linter/src/rules/isort/mod.rs
 I002 [*] Missing required import: `from __future__ import annotations`
 --> existing_import.py:1:1
 help: Insert required import: `from __future__ import annotations`
-1 + from __future__ import annotations
-2 | from __future__ import generator_stop
+1 | from __future__ import generator_stop
+2 + from __future__ import annotations
 3 | import os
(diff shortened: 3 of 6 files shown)


crates/ruff_linter/src/importer/mod.rs
```diff
@@ -524,6 +531,17 @@ impl<'a> Importer<'a> {
...
 
+    /// Find the last `from __future__` import statement in the AST.
+    fn find_last_future_import(&self) -> Option<&'a Stmt> {
+        self.python_ast.iter().rev().find(|stmt| {
+            if let Stmt::ImportFrom(import_from) = stmt {
+                import_from.module.as_deref() == Some("__future__")
+            } else {
+                false
+            }
+        })
```
amyreese: Since future imports can only be at the beginning, iterating over all statements is inefficient, particularly on large files with many statements. Instead, only look through statements until you've found something that's *not* a future import.

Eg, something like:

```rust
fn find_last_future_import(&self) -> Option<&'a Stmt> {
  self.python_ast.iter().take_while(|stmt| {
    if let Stmt::ImportFrom(import_from) = stmt {
      import_from.module.as_deref() == Some("__future__")
    } else {
      false
    }
  }).last()
}
```

crates/ruff_linter/src/rules/isort/snapshots/ruff_linter__rules__isort__tests__required_import_existing_import.py.snap
```diff
@@ -4,6 +4,6 @@ source: crates/ruff_linter/src/rules/isort/mod.rs
 I002 [*] Missing required import: `from __future__ import annotations`
 --> existing_import.py:1:1
 help: Insert required import: `from __future__ import annotations`
-1 + from __future__ import annotations
-2 | from __future__ import generator_stop
+1 | from __future__ import generator_stop
+2 + from __future__ import annotations
```
amyreese, line 8: Do you know why these snapshots changed?

danparizher: Before the fix, the required import was being inserted at the very beginning of the file, which would cause a syntax error since future imports must come first.

Now it's inserting the required import after the existing future import, which maintains the correct order and prevents the syntax error.

crates/ruff_linter/src/importer/mod.rs
```diff
@@ -524,6 +531,20 @@ impl<'a> Importer<'a> {
...
+        self.python_ast
+            .iter()
+            .take_while(|stmt| {
+                if let Stmt::ImportFrom(import_from) = stmt {
+                    import_from.module.as_deref() == Some("__future__")
+                } else {
+                    false
+                }
+            })
+            .last()
```
ntBre: It looks like this won't properly account for docstrings:

```shell
> ./target/debug/ruff check --select I --config 'lint.isort.required-imports=["import this"]' try.py --preview --no-cache
I002 [*] Missing required import: `import this`
--> try.py:1:1
help: Insert required import: `import this`
1 | "a docstring"
2 + import this
3 | from __future__ import annotations

Found 1 error.
```

I also noticed by looking at the `Insertion::start_of_file` references that we have this `preceding_import` function, which accesses import information held on the `Importer`:

https://github.com/astral-sh/ruff/blob/4e94b22815c7979003d07a4931c9174285617c90/crates/ruff_linter/src/importer/mod.rs#L509-L515

Can we use the `runtime_imports` field referenced here instead of iterating over the statements directly? It seems functionally the same, but maybe a bit more clear what's going on.

This preceded the PR, but it looks like the `add_import` docs are a bit out of date too. This part doesn't seem to be true:

```rust
    /// Otherwise, it will be added after the most recent top-level
    /// import statement.
```

Rather, the import is added after the import preceding whatever you pass as the `at` argument. I think I002 would actually work fine if the docstring were accurate. So an alternative way to fix this would be to pass a more accurate `at` in the I002 rule code. That actually seems kind of tempting to me instead of modifying the general helper.

ntBre (changes requested):
I think we need to handle docstrings, and I also had some ideas for other places to add this check.

crates/ruff_linter/src/importer/mod.rs
danparizher: The `runtime_imports` approach would require populating it in the I002 rule, but since the rule doesn't call `visit_import()`, it would be empty. Modifying the general helper seemed cleaner since it fixes the issue for all users of `add_import`, not just I002.

```rust
.take_while(|stmt| {
    match stmt {
        Stmt::ImportFrom(import_from) => import_from.module.as_deref() == Some("__future__"),
        Stmt::Expr(stmt_expr) => matches!(stmt_expr.value.as_ref(), Expr::StringLiteral(_)),
        _ => false,
    }
})
```

crates/ruff_linter/src/importer/mod.rs
```diff
@@ -524,6 +532,32 @@ impl<'a> Importer<'a> {
...
+                }
+            })
+            .filter(|stmt| {
+                if let Stmt::ImportFrom(import_from) = stmt {
+                    import_from.module.as_deref() == Some("__future__")
+                } else {
+                    false
+                }
+            })
+            .last()
```
ntBre: I think we could take some inspiration from the `match_docstring_end` helper and just skip one statement. It's not a huge deal since something like this is a separate syntax error, but it still seems a bit more precise:

```py
"docstring"
"another string"
from __future__ import annotations  # SyntaxError: from __future__ imports must occur at the beginning of the file
```

What do you think about something like this?

```rust
    /// Find the last valid `from __future__` import statement in the AST.
    fn find_last_future_import(&self) -> Option<&'a Stmt> {
        let mut body = self.python_ast.iter().peekable();
        let _docstring = body.next_if(|stmt| is_docstring_stmt(stmt));

        body.take_while(|stmt| {
            stmt.as_import_from_stmt()
                .is_some_and(|import_from| import_from.module.as_deref() == Some("__future__"))
        })
        .last()
    }
```

crates/ruff_linter/src/rules/isort/mod.rs
```diff
@@ -1014,6 +1014,52 @@ mod tests {
...
+                    ..super::settings::Settings::default()
+                },
+                ..LinterSettings::for_rule(Rule::MissingRequiredImport)
+            },
+        )?;
+        assert_diagnostics!(snapshot, diagnostics);
+        Ok(())
+    }
+
+    #[test_case(Path::new("docstring_future_import.py"))]
```
ntBre: Can we make these two `test_case`s on the same test function? The bodies of the functions look pretty much the same to me, besides the snapshot name. I think using the test case path is enough differentiation in the filename, though.

crates/ruff_linter/src/importer/mod.rs
```diff
@@ -534,28 +534,16 @@ impl<'a> Importer<'a> {
...
-                    import_from.module.as_deref() == Some("__future__")
-                } else {
-                    false
-                }
-            })
-            .last()
+        use ruff_python_ast::helpers::is_docstring_stmt;
+
+        let mut body = self.python_ast.iter().peekable();
+        let _docstring = body.next_if(|stmt| is_docstring_stmt(stmt));
```
ntBre:
```suggestion
        let mut body = self.python_ast.iter().peekable();
        let _docstring = body.next_if(|stmt| ast::helpers::is_docstring_stmt(stmt));
```

PR #19786, opened 2025-08-06
[ty] Track when type variables are inferrable or not

dcreager:
`Type::TypeVar` now distinguishes whether the typevar in question is inferrable or not.

A typevar is _not inferrable_ inside the body of the generic class or function that binds it:

```py
def f[T](t: T) -> T:
    return t
```

The inferred type of `t` in the function body is `TypeVar(T, NotInferrable)`. This represents how e.g. assignability checks need to be valid for all possible specializations of the typevar. Most of the existing assignability/etc logic only applies to non-inferrable typevars.

Outside of the function body, the typevar is _inferrable_:

```py
f(4)
```

Here, the parameter type of `f` is `TypeVar(T, Inferrable)`. This represents how e.g. assignability doesn't need to hold for _all_ specializations; instead, we need to find the constraints under which this specific assignability check holds.

This is in support of starting to perform specialization inference _as part of_ performing the assignability check at the call site.

In the [[POPL2015][]] paper, this concept is called _monomorphic_ / _polymorphic_, but I thought _non-inferrable_ / _inferrable_ would be clearer for us.

Depends on #19784 

[POPL2015]: https://doi.org/10.1145/2676726.2676991

PR #20677, opened 2025-10-02
[ty] Don't track inferability via different `Type` variants

dcreager:
We have to track whether a typevar appears in a position where it's inferable or not. In a non-inferable position (in the body of the generic class or function that binds it), assignability must hold for every possible specialization of the typevar. In an inferable position, it only needs to hold for _some_ specialization. https://github.com/astral-sh/ruff/pull/20093 is working on using constraint sets to model assignability of typevars, and the constraint sets that we produce will be the same for inferable vs non-inferable typevars; what changes is what we _compare_ that constraint set to. (For a non-inferable typevar, the constraint set must equal the set of valid specializations; for an inferable typevar, it must not be `never`.)

When I first added support for tracking inferable vs non-inferable typevars, it seemed like it would be easiest to have separate `Type` variants for each. The alternative (which lines up with the Δ set in [POPL15](https://doi.org/10.1145/2676726.2676991)) would be to explicitly plumb through a list of inferable typevars through our type property methods. That seemed cumbersome.

In retrospect, that was the wrong decision. We've had to jump through hoops to translate types between the inferable and non-inferable variants, which has been quite brittle. Combined with the original point above, that much of the assignability logic will become more identical between inferable and non-inferable, there is less justification for the two `Type` variants. And plumbing an extra `inferable` parameter through all of these methods turns out to not be as bad as I anticipated.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -78,7 +78,7 @@ use crate::types::function::{
     FunctionDecorators, FunctionLiteral, FunctionType, KnownFunction, OverloadLiteral,
 };
 use crate::types::generics::{
-    GenericContext, LegacyGenericBase, SpecializationBuilder, bind_typevar,
+    GenericContext, InferableTypeVars, LegacyGenericBase, SpecializationBuilder, bind_typevar,
     enclosing_generic_contexts,
 };
 use crate::types::infer::nearest_enclosing_function;
@@ -5919,11 +5919,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             parenthesized: _,
         } = tuple;
 
-        // TODO: Use the list of inferable typevars from the generic context of tuple.
-        let inferable = InferableTypeVars::None;
-
         // Remove any union elements of that are unrelated to the tuple type.
         let tcx = tcx.map(|annotation| {
+            let inferable = KnownClass::Tuple
+                .try_to_class_literal(self.db())
+                .and_then(|class| class.generic_context(self.db()))
+                .map(|generic_context| generic_context.inferable_typevars(self.db()))
+                .unwrap_or(InferableTypeVars::None);
             annotation.filter_disjoint_elements(
                 self.db(),
                 KnownClass::Tuple.to_instance(self.db()),


dcreager:
ecosystem analysis:

Several projects have removed diagnostics, which look like correct better understanding of the types involved:

- `hauntsaninja/boostedblob`
- `homeassistant/core`
- `cwltool`
- `jax`
- `meson`
- `pytest`
- `schema_salad`
- `scikit-build-core`
- `strawberry`
- `werkzeug`
- `xarray`

Still looking at some of the new diagnostics. It also looks like we aren't expanding type aliases in as many places as before.

dcreager:
The remaining ecosystem additions seem related to bidi checking of arguments and overload resolution. This overlaps with some of the work @ibraheemdev is doing e.g. in https://github.com/astral-sh/ruff/pull/20796, so for now, I'm opening this up for review, and would like to address the remaining ecosystem results with him as part of that work.

dcreager:
(Note that I pulled the pure refactoring part of this PR out into a separate https://github.com/astral-sh/ruff/pull/20865, so that it's easier to review the logic changes separate from the boilerplate API changes. Please review 20865 first)

crates/ty_python_semantic/src/types.rs
```diff
@@ -1685,18 +1674,21 @@ impl<'db> Type<'db> {
             // However, there is one exception to this general rule: for any given typevar `T`,
             // `T` will always be a subtype of any union containing `T`.
             // A similar rule applies in reverse to intersection types.
-            (Type::NonInferableTypeVar(_), Type::Union(union))
-                if union.elements(db).contains(&self) =>
+            (Type::TypeVar(bound_typevar), Type::Union(union))
+                if !inferable.is_inferable(db, bound_typevar)
+                    && union.elements(db).contains(&self) =>
```
AlexWaygood, line 1723: nit: this kind of logic might read more naturally if the `is_inferable` method was a method on `BoundTypeVarInstance` rather than `Inferable`? then this could be

```suggestion
            (Type::TypeVar(bound_typevar), Type::Union(union))
                if !bound_typevar.is_inferable(db, inferable)
                    && union.elements(db).contains(&self) =>
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -2355,9 +2358,12 @@ impl<'db> Type<'db> {
 
             // Other than the special cases enumerated above, `Instance` types and typevars are
             // never subtypes of any other variants
-            (Type::NominalInstance(_) | Type::NonInferableTypeVar(_), _) => {
+            (Type::TypeVar(bound_typevar), _) => {
+                // All inferable cases should have been handled above
+                debug_assert!(!inferable.is_inferable(db, bound_typevar));
```
AlexWaygood: what happens if we fail to uphold this invariant... would that lead to very strange diagnostic messages later down the line? If so, might it be worth asserting this in release builds too? Or would that be prohibitively expensive?

```suggestion
                // All inferable cases should have been handled above
                assert!(!inferable.is_inferable(db, bound_typevar));
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -3256,7 +3264,7 @@ impl<'db> Type<'db> {
             // the bound is a final singleton class, since it can still be specialized to `Never`.
             // A constrained typevar is a singleton if all of its constraints are singletons. (Note
             // that you cannot specialize a constrained typevar to a subtype of a constraint.)
-            Type::NonInferableTypeVar(bound_typevar) => {
+            Type::TypeVar(bound_typevar) => {
```
AlexWaygood, line 3311: It looks like we previously always returned `false` here if it was an inferable `TypeVar` -- is the change in logic deliberate here?

crates/ty_python_semantic/src/types.rs
```diff
@@ -3367,7 +3373,7 @@ impl<'db> Type<'db> {
             // `Never`. A constrained typevar is single-valued if all of its constraints are
             // single-valued. (Note that you cannot specialize a constrained typevar to a subtype
             // of a constraint.)
-            Type::NonInferableTypeVar(bound_typevar) => {
+            Type::TypeVar(bound_typevar) => {
```
AlexWaygood, line 3420: same here, this looks like a behaviour change -- is that deliberate? If this fixes a bug, would it be possible to add a test that makes sure it doesn't regress?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -255,6 +309,71 @@ impl<'db> GenericContext<'db> {
...
+        cycle_initial=inferable_typevars_cycle_initial,
+        heap_size=ruff_memory_usage::heap_size,
+    )]
+    fn inferable_typevars_inner(self, db: &'db dyn Db) -> FxHashSet<BoundTypeVarIdentity<'db>> {
+        // The second inner function is because the salsa macros seem to not like nested structs
+        // and impl blocks inside the function.
+        self.inferable_typevars_innerer(db)
+    }
+
+    fn inferable_typevars_innerer(self, db: &'db dyn Db) -> FxHashSet<BoundTypeVarIdentity<'db>> {
```
AlexWaygood: lol. Is it worth nesting these methods inside the `inferable_typevars` method, so that it's obvious to readers that they're implementation details of that method?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -255,6 +309,71 @@ impl<'db> GenericContext<'db> {
...
+                    }
+                }
+            }
+        }
+
+        let visitor = CollectTypeVars {
+            typevars: RefCell::new(FxHashSet::default()),
+            seen_types: RefCell::new(FxIndexSet::default()),
+        };
+        for bound_typevar in self.variables(db) {
```
AlexWaygood: hmm, same comment as https://github.com/astral-sh/ruff/pull/20822#discussion_r2424885000 ;)

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -617,6 +603,15 @@ impl<'db> Signature<'db> {
         inferable: InferableTypeVars<'_, 'db>,
         visitor: &IsEquivalentVisitor<'db>,
     ) -> ConstraintSet<'db> {
+        // The typevars in self and other should also be considered inferable when checking whether
+        // two signatures are equivalent.
+        let self_inferable =
+            (self.generic_context).map(|generic_context| generic_context.inferable_typevars(db));
+        let other_inferable =
+            (other.generic_context).map(|generic_context| generic_context.inferable_typevars(db));
```
AlexWaygood, line 611: (obviously not blocking, just curious) -- are the parentheses around `self.generic_context` to try to get rustfmt to format this the way you'd like it to be formatted?

crates/ty_python_semantic/src/types/generics.rs
ibraheemdev: Do you remember what the issue was here? Might be worth opening a Salsa issue, the macros should be relatively transparent.

dcreager:
The performance analysis keeps bouncing back and forth between being a regression and an improvement as I rebase to pick up more changes from `main`. :angry:

crates/ty_python_semantic/src/types.rs
dcreager: Should not be too expensive — this was previously a case that the compiler rejected as not representable, so it shouldn't occur in practice. I'll change to a full assert as you suggest

crates/ty_python_semantic/src/types.rs
dcreager: Ooh I like that

AlexWaygood:
> The performance analysis keeps bouncing back and forth between being a regression and an improvement as I rebase to pick up more changes from `main`. 😠

the `DateType` benchmark does move quite a lot in general because it's a small codebase that has some _very_ expensive subtype checks in it. (We originally added it because an early version of one of my protocol PRs would have meant that we never finished type-checking `DateType` 🙃)

So `DateType` is just very sensitive to small changes in efficiency of subtype checks between the specific types that are being used in that codebase. It looks like this PR is still a performance improvement on hydra-zen, which is a much bigger codebase!

crates/ty_python_semantic/src/types.rs
dcreager: Done

crates/ty_python_semantic/src/types.rs
dcreager: This reverts back to [what we had](https://github.com/astral-sh/ruff/pull/19786/files#diff-8049ab5af787dba29daa389bbe2b691560c15461ef536f122b1beab112a4b48aL2386) before #19786, which introduced the `NonInferableTypeVar` variant. I can't think of any way that an inferable typevar would affect whether we consider something a singleton — they should only appear in function/callable signatures, and we don't have to look at the signature to determine whether a function or callable is a singleton. Realistically, the `false` branch from before should have included a `debug_assert`.

crates/ty_python_semantic/src/types.rs
dcreager: Ditto above

crates/ty_python_semantic/src/types/generics.rs
dcreager: whopps, done

crates/ty_python_semantic/src/types/generics.rs
dcreager: Putting `innerer` inside of `inner` didn't work, but I don't remember the specific error. Put all of it inside of the outer function works, though, since the type and `impl` block don't need to be inside the salsa-tracked function that way.

crates/ty_python_semantic/src/types/signatures.rs
dcreager: Yep

crates/ty_python_semantic/src/types.rs
AlexWaygood: Thanks!

Issue #20196, opened 2025-09-01
Ruff should warn on all CPython SyntaxWarnings


JelleZijlstra:
### Summary

CPython emits a SyntaxWarning for certain code, but SyntaxWarnings have some issues that can make them difficult to reliably detect in CI (current discussion in [https://discuss.python.org/t/pep-765-disallow-return-break-continue-that-exit-a-finally-block/71348](discuss.python.org) about PEP 765). So it would be nice if Ruff could guarantee that it emits an error on any code for which CPython would produce a SyntaxWarning.

Here is a [playground link](https://play.ruff.rs/0437f5dd-3ea3-463c-b27e-71814287282e) covering all current SyntaxWarnings I could find in the CPython source code. Ruff already warns for most of them, though the logic may not always match exactly.
 
* Invalid numeric literals (e.g., `123or 1`) [lexer.c](https://github.com/python/cpython/blob/2a54acf3c3d9f388c3d878a17ea804a801affca9/Parser/lexer/lexer.c#L346)
  * Not implemented in Ruff. See [this thread](https://discuss.python.org/t/whats-going-on-with-syntaxwarning-invalid-decimal-literal/52694) and linked discussions for background.
* Invalid escape sequences in strings (e.g., `"\99"`) [string_parser.c](https://github.com/python/cpython/blob/2a54acf3c3d9f388c3d878a17ea804a801affca9/Parser/string_parser.c#L47)
  * Ruff W605. Might need auditing of the code to verify this matches the CPython logic exactly.
* Backslash next to `{` or `}` in f-strings or t-strings (e.g., `f"\{1}"`) ([lexer.c](https://github.com/python/cpython/blob/2a54acf3c3d9f388c3d878a17ea804a801affca9/Parser/lexer/lexer.c#L1578))
  * Also W605, but uses a different code path in CPython. 
* Control flow in a `finally` block (see PEP 765) ([ast_preprocess.c](https://github.com/python/cpython/blob/2a54acf3c3d9f388c3d878a17ea804a801affca9/Python/ast_preprocess.c#L71))
  * Ruff B012. 
* `is` / `is not` with a literal (`() is ()`) [codegen.c](https://github.com/python/cpython/blob/2a54acf3c3d9f388c3d878a17ea804a801affca9/Python/codegen.c#L1836)
  * Ruff F632.
[...]

PR #20681, opened 2025-10-02
Add `B012` and `PYI057` as default rules

ntBre:
## Summary

@AlexWaygood and I discussed this a bit on [Discord](https://discord.com/channels/1039017663004942429/1343691651243315312/1422892554868756552), and it seemed relevant for the Python 3.14 release next week, so I figured I'd put up a PR.

This adds [jump-statement-in-finally (B012)](https://docs.astral.sh/ruff/rules/jump-statement-in-finally/#jump-statement-in-finally-b012) and [byte-string-usage (PYI057)](https://docs.astral.sh/ruff/rules/byte-string-usage/#byte-string-usage-pyi057) to the default rule set. `B012` is especially relevant for Python 3.14 because it will become a `SyntaxWarning` and may be upgraded to a `SyntaxError` in the future. `PYI057`, on the other hand, covers APIs that now won't be removed until 3.17, but Ruff could help to make this more visible to avoid it being pushed back again.

I'm realizing now that I didn't attempt to gate the new defaults based on the target Python version as Alex suggested, but I can look into that if we want. It doesn't immediately seem straightforward to do, based on our current usage of `DEFAULT_SELECTORS`, though. `B012` seems serious enough to me always to be enabled, but the nice stdlib replacement for `PYI057` was only added in Python 3.12.

## Test Plan

Existing tests, plus a new CLI test with the new rules

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -51,10 +51,12 @@ If left unspecified, Ruff's default configuration is equivalent to:
     target-version = "py39"
 
     [tool.ruff.lint]
-    # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
+    # Enable Pyflakes (`F`), a subset of the pycodestyle (`E`) codes,
+	# and a couple of other codes corresponding to CPython deprecations
+	# and syntax warnings by default.
     # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
     # McCabe complexity (`C901`) by default.
-    select = ["E4", "E7", "E9", "F"]
+    select = ["E4", "E7", "E9", "F", "PYI057", "B012"]
     ignore = []
 
     # Allow fix for all enabled rules (when `--fix`) is provided.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -52,8 +52,8 @@ If left unspecified, Ruff's default configuration is equivalent to:
 
     [tool.ruff.lint]
     # Enable Pyflakes (`F`), a subset of the pycodestyle (`E`) codes,
-	# and a couple of other codes corresponding to CPython deprecations
-	# and syntax warnings by default.
+    # and a couple of other codes corresponding to CPython deprecations
+    # and syntax warnings by default.
     # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
     # McCabe complexity (`C901`) by default.
     select = ["E4", "E7", "E9", "F", "PYI057", "B012"]
--- a/README.md
+++ b/README.md
@@ -314,9 +314,7 @@ Rust as a first-party feature.
 By default, Ruff enables Flake8's `F` rules, along with a subset of the `E`
 rules, omitting any stylistic rules that overlap with the use of a formatter,
 like `ruff format` or [Black](https://github.com/psf/black). Ruff also enables
-[`B012`](https://docs.astral.sh/ruff/rules/jump-statement-in-finally/), which
-corresponds to a `SyntaxWarning`, and
-[`PYI057`](https://docs.astral.sh/ruff/rules/byte-string-usage/), which flags
+`B012`, which corresponds to a `SyntaxWarning`, and `PYI057`, which flags
 usage of some long-deprecated APIs in the standard library.
 
 If you're just getting started with Ruff, **the default rule set is a great place to start**: it
--- a/README.md
+++ b/README.md
@@ -311,9 +311,13 @@ for more on the linting and formatting commands, respectively.
 isort, pyupgrade, and others. Regardless of the rule's origin, Ruff re-implements every rule in
 Rust as a first-party feature.
 
-By default, Ruff enables Flake8's `F` rules, along with a subset of the `E` rules, omitting any
-stylistic rules that overlap with the use of a formatter, like `ruff format` or
-[Black](https://github.com/psf/black).
+By default, Ruff enables Flake8's `F` rules, along with a subset of the `E`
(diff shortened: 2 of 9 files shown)


docs/configuration.md
```diff
@@ -51,10 +51,12 @@ If left unspecified, Ruff's default configuration is equivalent to:
     target-version = "py39"
 
     [tool.ruff.lint]
-    # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
+    # Enable Pyflakes (`F`), a subset of the pycodestyle (`E`) codes,
+	# and a couple of other codes corresponding to CPython deprecations
+	# and syntax warnings by default.
```
AlexWaygood: (Use spaces rather than tabs ;)

```suggestion
    # Enable Pyflakes (`F`), a subset of the pycodestyle (`E`) codes,
    # and a couple of other codes corresponding to CPython deprecations
    # and syntax warnings by default.
```

ntBre: Wow how did you see the tabs instead of spaces?? I thought for sure GitHub's suggestion rendering was failing me until I checked locally. Thanks!

AlexWaygood: > Wow how did you see the tabs instead of spaces??

I'll never reveal my secrets ;)

AlexWaygood:
> `B012` seems serious enough to me always to be enabled

agreed; I think this is almost always a mistake, whatever Python version you're using

> but the nice stdlib replacement for `PYI057` was only added in Python 3.12.

Hmm, yeah. _Ideally_ I think this would only be enabled by default if your code targets Python 3.12+. But most users can probably use a `bytes | bytearray` union on lower Python versions instead of an ABC. And if they have typing_extensions as a dependency, they can use the backport of `collections.abc.Buffer` in typing_extensions too. So I think it _should_ be okay if it's made the default regardless of target Python version.

It might be worth adding some more examples to https://docs.astral.sh/ruff/rules/byte-string-usage/#example for what you can do if you support Python <3.12? (It also seems like there's a typo in https://docs.astral.sh/ruff/rules/byte-string-usage/#why-is-this-bad -- it says `Python \<3.12` rather than `Python <3.12`.)

MichaReiser:
I'm a bit worried about making our default rules more complicated, especially if the selection becomes dynamic

> I'm realizing now that I didn't attempt to gate the new defaults based on the target Python version as Alex suggested, but I can look into that if we want.

Could `B012` become a syntax rule? I guess the main reason why it shouldn't is because there are reasons why someone would want to disable it (on a per-file or per-line basis).

ntBre:
I can definitely expand the examples, I think I already caught the typo but I'll double check that too.

> Could B012 become a syntax rule? I guess the main reason why it shouldn't is because there are reasons why someone would want to disable it (on a per-file or per-line basis).

It's tempting to make it a semantic syntax error, but it would be a bit more aggressive than CPython itself. The [PEP](https://peps.python.org/pep-0765/#emit-syntaxerror-in-cpython) explicitly rejected or at least deferred plans to upgrade the warning to an error.

I know what you mean about expanding the defaults. I'm also happy to split the two rules into separate PRs and/or defer both until a later release after more discussion. I just wanted to have it ready in case we wanted to include these in the 3.14 release next week.

AlexWaygood:
> I'm a bit worried about making our default rules more complicated, especially if the selection becomes dynamic

Yeah, I can see that making the default selection dynamic might be pretty confusing to users. As I said above, I think it's probably okay if PYI057 is enabled by default regardless of Python version. Of the two, I also think there's an even stronger case for enabling B012 by default, as well, so considering the two rules separately might indeed make sense here.

> Could `B012` become a syntax rule? I guess the main reason why it shouldn't is because there are reasons why someone would want to disable it (on a per-file or per-line basis).

Yeah, I think it's pretty important that B012 should be suppressible:
- CPython will still parse and execute the code at runtime (albeit with a warning)
- The vast majority of the time, code that uses this pattern isn't working as intended, and there's always a way to rewrite the code in a clearer, more explicit way. But not _every_ instance of this pattern is _necessarily_ a bug; sometimes it might actually be working in the desired way! Even in these instances, I think the lint is beneficial, since you could rewrite the code in a clearer way, but that doesn't mean that it should be unsuppressible.
- What if your code has lots of instances of this pattern? I think it would be reasonable to add `noqa` comments to all occurences (using `--add-noqa`) as part of the initial PR upgrading your Ruff version, and then slowly remove the `noqa` comments in followup PRs.
- There's no particular reason to make B012 non-suppressible while other rules that cover `SyntaxWarning`s at runtime (such as [F632](https://docs.astral.sh/ruff/rules/is-literal/)) are still suppressible.

MichaReiser:
It makes me wonder if we should have a single rule that flags any syntax warnings in your code (maybe dependent on python version). 

But that is a separate discussion

AlexWaygood:
yeah, I do agree that we should try to ensure that we catch all pattenrs that CPython flags at compile-time with `SyntaxWarning`s, that rules which flag syntax warnings such as `B012` and `F632` should probably all be enabled by default, and that these rules should probably all be in a common category. It's definitely a similar problem to the syntax-errors project, and I've chatted with @ntBre about it before. But, I also agree that this is a separate, broader discussion.

dylwil3:
> It makes me wonder if we should have a single rule that flags any syntax warnings in your code (maybe dependent on python version).
> 

see https://github.com/astral-sh/ruff/issues/20196

ntBre:
It sounds like we may want to hold off on this for now and take a more general approach to syntax warnings, which sounds good to me. I think we can still call out these rules in the release post without enabling them by default and hopefully still help with the deprecation.

PR #20682, opened 2025-10-02
[syntax-errors] Alternative `match` patterns bind different names

11happy:
## Summary

This PR implements semantic syntax error where alternative patterns bind different names

## Test Plan


11happy:
test `/ruff/crates/ty_python_semantic/resources/mdtest/import/star.md` is failing as now two syntax errors are emitted at same line line 295 & 366, Can we add multiple syntax errors on the same line ?

Thank you

crates/ruff_python_parser/tests/snapshots/invalid_syntax@irrefutable_case_pattern.py.snap
```diff
@@ -427,3 +427,12 @@ Module(
    |                         ^^^ Syntax Error: name capture `var` makes remaining patterns unreachable
 12 |     case 2: ...
    |
+
+
+   |
+ 9 |     case 2: ...
+10 | match x:
+11 |     case enum.variant | var: ...  # or pattern with irrefutable part
+   |          ^^^^^^^^^^^^^^^^^^ Syntax Error: alternative pattern bind different names: expected ``, found `var`
```
ntBre: It looks like something is going wrong with the error message here (```expected ``, ...```).

crates/ruff_python_parser/tests/snapshots/invalid_syntax@statements__match__star_pattern_usage.py.snap
```diff
@@ -613,3 +613,25 @@ Module(
...
    |
+
+
+## Semantic Syntax Errors
+
+   |
+ 7 |     case *foo:
+ 8 |         pass
+ 9 |     case *foo | 1:
+   |          ^^^^^^^^ Syntax Error: alternative pattern bind different names: expected `foo`, found ``
```
ntBre: These also show the empty placeholders in the error message.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1748,12 +1762,43 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
...
+                    for current_name in all_names.iter().skip(1) {
+                        if current_name != first_name {
+                            // test_err alternate_binded_pattern
+                            // match x:
+                            //     case [a] | [b]: ...
+                            //     case [a] | []: ...
+                            //     case (x, y) | (x,): ...
+                            //     case [a, _] | [a, b]: ...
+                            //     case (x, (y | z)): ...
+                            //     case [a] | [b] | [c]: ...
```
ntBre: I think some other interesting cases involve more nested structure. That's the reason we have to do this check in a visitor:

```py
match x:
    case [a] | [C(x)]: ...
```

We could also add some `test_ok` cases, like `case [a] | [C(a)]: ...` which should work. You might be able to adapt some of the cases above from `duplicate_match_class_attr` if you want even deeper nesting.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1748,12 +1762,43 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
                 // test_ok multiple_assignment_in_case_pattern
                 // match 2:
                 //     case Class(x) | [x] | x: ...
+
+                let mut all_names: Vec<Vec<String>> = Vec::new();
```
ntBre: What do you think about something more like this?

```rust
                let mut previous_names: Option<FxHashSet<&ast::name::Name>> = None;
                for pattern in patterns {
                    let mut visitor = Self {
                        names: FxHashSet::default(),
                        ctx: self.ctx,
                    };
                    visitor.visit_pattern(pattern);
                    match previous_names {
                        Some(ref previous_names) => {
                            for missing_name in previous_names.symmetric_difference(&visitor.names)
                            {
                                    self.ctx,
                                    SemanticSyntaxErrorKind::AlternateBindedPattern(
                                        missing_name.to_string(),
                                        missing_name.to_string(),
                                    ),
                                    pattern.range(),
                                );
                            }
                        }
                        None => previous_names = Some(visitor.names),
                    }
                }
```

The nice thing about this is that we can avoid allocating additional `Vec`s, but the error message might be a bit worse. I'm just passing the same `missing_name` twice here, obviously we'd have to do something different. But I think an error message either like Python itself:

```pycon
>>> match 42:
...     case [x] | [C(x), C(y)]: ...
...
  File "<python-input-12>", line 2
    case [x] | [C(x), C(y)]: ...
         ^^^^^^^^^^^^^^^^^^
SyntaxError: alternative patterns bind different names
```

or like Rust:

```
error[E0408]: variable `y` is not bound in all patterns
 --> try.rs:4:9
  |
4 |         x | y => todo!(),
  |         ^   - variable not in all patterns
  |         |
  |         pattern doesn't bind `y`

error[E0408]: variable `x` is not bound in all patterns
 --> try.rs:4:13
  |
4 |         x | y => todo!(),
  |         -   ^ pattern doesn't bind `x`
  |         |
  |         variable not in all patterns
```

where we either mention zero or one variable name instead of both sets would be fine.

ntBre:
> test `/ruff/crates/ty_python_semantic/resources/mdtest/import/star.md` is failing as now two syntax errors are emitted at same line line 295 & 366, Can we add multiple syntax errors on the same line ?

Yes, I think you can stack the comments, although you may need to move the existing one to the previous line like this:

```
    # error: [invalid-syntax] "name capture `E` makes remaining patterns unreachable"
    # error: [invalid-syntax] "your error here"
    case E | F:  
        ...
```

crates/ruff_python_parser/tests/snapshots/invalid_syntax@statements__match__star_pattern_usage.py.snap
11happy: done , now I am mentioning only a single variable as rust does.

crates/ruff_python_parser/src/semantic_errors.rs
11happy: yes I added some more deeper nested cases & test_ok as well.

crates/ruff_python_parser/src/semantic_errors.rs
11happy: yeah , I agree this approach is much better & have updated accordingly & changed the diagnostic accordingly following similar to rust.

11happy:
> > test `/ruff/crates/ty_python_semantic/resources/mdtest/import/star.md` is failing as now two syntax errors are emitted at same line line 295 & 366, Can we add multiple syntax errors on the same line ?
> 
> Yes, I think you can stack the comments, although you may need to move the existing one to the previous line like this:
> 
> ```
>     # error: [invalid-syntax] "name capture `E` makes remaining patterns unreachable"
>     # error: [invalid-syntax] "your error here"
>     case E | F:  
>         ...
> ```

Sure, have added accordingly.
Thank you : )

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1511,7 +1508,7 @@ pub enum SemanticSyntaxErrorKind {
     /// Represents the use of a `__future__` feature that is not defined.
     FutureFeatureNotDefined(String),
 
-    AlternateBindedPattern(String, String),
+    AlternateBindedPattern(String),
```
ntBre: nit: Could we add some docs here? Just for more context, we don't currently show any of these docs anywhere, but we want to keep  them up to date in case we display them to users like rule docs in the future. That's also why most of the early ones are more elaborate with examples  and such.

crates/ruff_python_parser/src/semantic_errors.rs
ntBre: After looking at this again, I think I'd actually prefer the simpler error that Python emits for now. Using `symmetric_difference` like I suggested doesn't result in a very nice error message in some cases.

I think ideally we would emit something very similar to the rustc diagnostic, but we don't have access to an actual `Diagnostic` type in the `SemanticSyntaxChecker` at the moment. I might open a separate issue to follow up on that because it seems useful more generally.

For reference, I took a look at how Rust emits this [diagnostic](https://github.com/rust-lang/rust/blob/28d0a4a205f9e511ad2f51ee79a4aa19a704a455/compiler/rustc_resolve/src/late.rs#L3714-L3723) today. Something like you had initially actually looks closer, but again, we can't really make use of the additional information to render a nice diagnostic currently.

11happy: sure,

crates/ruff_python_parser/src/semantic_errors.rs

11happy:
rebased : )

crates/ruff_python_parser/src/semantic_errors.rs
ntBre: That looks great, thank you!

ntBre (approved):
Thank you!

I pushed two small commits, one using the full match pattern's range like CPython does. I think the narrower range could be helpful in the future, but the whole thing seems okay for now. And one renaming the error to `DifferentMatchPatternBindings`. I still don't totally love the name, but it's not user-facing, so we can always iterate on it if needed.

henryiii:
This change seems to now cause a SyntaxError on the following valid code that doesn't use different names:

```
invalid-syntax: alternative patterns bind different names
   --> src/sp_repo_review/checks/ruff.py:127:17
    |
125 |           match ruff:
126 |               case (
127 | /                 {"lint": {"select": x} | {"extend-select": x}}
128 | |                 | {"select": x}
129 | |                 | {"extend-select": x}
    | |______________________________________^
130 |               ):
131 |                   return cls.code in x or "ALL" in x
    |
```

See https://github.com/scientific-python/cookie/pull/663

PR #20684, opened 2025-10-02
[ty] Fix file root matching for `/`

BurntSushi:
Previously, we would always add `/{*filepath}` as our wildcard to match
descendant paths. But when the root is just `/` (as it can be in tests,
weird environments or in the ty playground), this causes a double `/`
and inhibits most descendant matches.

The regression test added in this commit fails without this fix.
Specifically, it panics because it can't find a file root for
`/project`.

This also enables the `log` feature on `tracing`, which means we get
better debug messages in the JavaScript console when running the
playground.

Fixes https://github.com/astral-sh/ty/issues/1277

### DIFF
--- a/crates/ruff_db/src/files/file_root.rs
+++ b/crates/ruff_db/src/files/file_root.rs
@@ -95,7 +95,10 @@ impl FileRoots {
         self.by_path.insert(route.clone(), root).unwrap();
 
         // Insert a path that matches all subdirectories and files
-        route.push_str("/{*filepath}");
+        if !route.ends_with("/") {
+            route.push('/');
+        }
+        route.push_str("{*filepath}");
 
         self.by_path.insert(route, root).unwrap();
         self.roots.push(root);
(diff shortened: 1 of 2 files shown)


ibraheemdev (approved):

Cargo.toml
```diff
@@ -177,7 +177,7 @@ test-case = { version = "3.3.1" }
 thiserror = { version = "2.0.0" }
 tikv-jemallocator = { version = "0.6.0" }
 toml = { version = "0.9.0" }
-tracing = { version = "0.1.40" }
+tracing = { version = "0.1.40", features = ["log"] }
```
MichaReiser: Can we only enable this in ty_wasm? 

It makes it a bit easier to reason about the change

BurntSushi: I think it makes sense to enable this "globally," but I get wanting to do it narrowly. I've switched the PR to enabling `log` only for `ty_wasm`.

PR #20685, opened 2025-10-02
Bump 0.13.3

dylwil3:


ntBre:
Ah you're working on it too :smile: I think we need one --bench flag for each additional argument, but I'm waiting on my CI to run too: https://github.com/astral-sh/ruff/pull/20686

PR #20686, opened 2025-10-02
Update benchmarking CI for cargo-codspeed v4

ntBre:
Summary
--

A new codspeed [release](https://github.com/CodSpeedHQ/codspeed-rust/releases/tag/v4.0.0) came out, and our CI was failing. It looks like the
previously-positional list of benchmarks now corresponds to a `--bench` [flag](https://github.com/CodSpeedHQ/codspeed-rust/blob/7159cf86b9432243565e488c827c1663b5385fde/crates/cargo-codspeed/src/app.rs#L39-L43).

Test Plan
--

CI on this PR


crates/ruff/tests/lint.rs
dylwil3, line 1: should this and the below be part of the diff? did `rustfmt` change or something?

ntBre: I need to drop that commit, I just wanted to make sure the benchmarks would actually run by making some random changes

PR #20687, opened 2025-10-02
[`fastapi`] Fix false positives for path parameters that FastAPI doesn't recognize (`FAST003`)

### LINKED ISSUE #20680: FAST003 false positives for path parameters that FastAPI doesn’t recognize
### Summary

[`fast-api-unused-path-parameter` (FAST003)](https://docs.astral.sh/ruff/rules/fast-api-unused-path-parameter/) detects path parameter patterns that FastAPI doesn’t, in which case it is not really a path parameter so the rule doesn’t apply and the fix doesn’t help. FAST003 should only apply to [parameters as detected by FastAPI](https://github.com/Kludex/starlette/blob/0.48.0/starlette/routing.py#L121). [For example](https://play.ruff.rs/9515c75f-3dc9-4252-8782-f031bc5fdb0d), parameters (including convertors) must be ASCII and can’t include spaces. Note that these strings aren’t errors to FastAPI: they are valid, albeit unusual, paths without parameters.
```console
$ cat >fast003.py <<'# EOF'
from fastapi import FastAPI
app = FastAPI()

@app.get("/f1/{x: str}")
async def f1():
    return locals()

@app.get("/f2/{用户身份}")
async def f2():
    return locals()

from starlette.convertors import Convertor, register_url_convertor
class ComplexConvertor(Convertor[complex]):
    regex = r"[+-]?[0-9]+(\.[0-9]+)?[+-][0-9]+(\.[0-9]+)?[Jj]"
    def convert(self, value: str) -> complex: return complex(value)
    def to_string(self, value: complex) -> str: return str(value).lstrip("(").rstrip(")")
register_url_convertor("ℂ", ComplexConvertor())
@app.get("/f3/{complex_number:ℂ}")
async def f3():
    return locals()
# EOF

$ pip install -q 'fastapi[standard]==0.118.0'

$ fastapi dev fast003.py

$ curl 'http://127.0.0.1:8000/f1/%7Bx:%20str%7D'
{}

$ curl 'http://127.0.0.1:8000/f2/%7B%E7%94%A8%E6%88%B7%E8%BA%AB%E4%BB%BD%7D'
{}

$ curl 'http://127.0.0.1:8000/f3/%7Bcomplex_number:%E2%84%82%7D'
{}

$ ruff --isolated check fast003.py --select FAST003 --unsafe-fixes --fix 
Found 3 errors (3 fixed, 0 remaining).

$ curl 'http://127.0.0.1:8000/f1/%7Bx:%20str%7D'
{"detail":[{"type":"missing","loc":["query","x"],"msg":"Field required","input":null}]}

$ curl 'http://127.0.0.1:8000/f1/%7Bx:%20str%7D?x=foo'
{"x":"foo"}

$ curl 'http://127.0.0.1:8000/f2/%7B%E7%94%A8%E6%88%B7%E8%BA%AB%E4%BB%BD%7D'
{"detail":[{"type":"missing","loc":["path","用户身份"],"msg":"Field required","input":null}]}

$ curl 'http://127.0.0.1:8000/f2/%7B%E7%94%A8%E6%88%B7%E8%BA%AB%E4%BB%BD%7D?%E7%94%A8%E6%88%B7%E8%BA%AB%E4%BB%BD=foo'
{"detail":[{"type":"missing","loc":["path","用户身份"],"msg":"Field required","input":null}]}

$ curl 'http://127.0.0.1:8000/f3/%7Bcomplex_number:%E2%84%82%7D'
{"detail":[{"type":"missing","loc":["query","complex_number"],"msg":"Field required","input":null}]}

$ curl 'http://127.0.0.1:8000/f3/%7Bcomplex_number:%E2%84%82%7D?complex_number=foo'
{"complex_number":"foo"}
```

### Version

danparizher:
## Summary

Fixes #20680

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/fastapi/FAST003.py
+++ b/crates/ruff_linter/resources/test/fixtures/fastapi/FAST003.py
@@ -227,3 +227,32 @@ async def read_thing(query: str):
 @app.get("/things/{ thing_id : str }")
 async def read_thing(query: str):
     return {"query": query}
+
+
+# https://github.com/astral-sh/ruff/issues/20680
+# These should NOT trigger FAST003 because FastAPI doesn't recognize them as path parameters
+
+# Non-ASCII characters in parameter name
+@app.get("/f1/{用户身份}")
+async def f1():
+    return locals()
+
+# Space in parameter name  
+@app.get("/f2/{x: str}")
+async def f2():
+    return locals()
+
+# Non-ASCII converter
+@app.get("/f3/{complex_number:ℂ}")
+async def f3():
+    return locals()
+
+# Mixed non-ASCII characters
+@app.get("/f4/{用户_id}")
+async def f4():
+    return locals()
+
+# Space in parameter name with converter
+@app.get("/f5/{param: int}")
+async def f5():
+    return locals()
--- a/crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
+++ b/crates/ruff_linter/src/rules/fastapi/rules/fastapi_unused_path_parameter.rs
@@ -164,9 +164,9 @@ pub(crate) fn fastapi_unused_path_parameter(
     }
 
     // Check if any of the path parameters are not in the function signature.
-    for (path_param, range) in path_params {
-        // Ignore invalid identifiers (e.g., `user-id`, as opposed to `user_id`)
-        if !is_identifier(path_param) {
+    for (path_param, param_content, range) in path_params {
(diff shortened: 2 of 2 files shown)


dscorbett:
`is_fastapi_valid_path_parameter` doesn’t exactly match [the regex FastAPI uses](https://github.com/Kludex/starlette/blob/0.48.0/starlette/routing.py#L121).

ntBre:
It makes sense to me just to use the same regex. It looks like they just iterate over the regex matches to extract parameters too, which might help us simplify `PathParamIterator` as well.

Issue #19016, opened 2025-06-28
Cleanup Ruff's CLI tests


MichaReiser:
Ruff's CLI integration tests (`crates/ruff/tests`) contain a lot of shared boilerplate code (shared between test functions but also test files). We should try to extract those into a more reusable testing infra. I think ty's CLI test infra works pretty well and we should try if we can use the same structure for ruff. 

https://github.com/astral-sh/ruff/blob/1f27d53fd576e92a7d2cd4abc1761d6ad836d85b/crates/ty/tests/cli/main.rs#L600

Renkai:
Hi, I tried to help on this with PR:https://github.com/astral-sh/ruff/pull/20689
Not sure if it's the right direction. If I get yes I will keep working until the file lint.rs are all rewritten in the new style.

statxc:
I think this is not fully completed. @MichaReiser Please review this issue again and let me know if it needs to update more. I am ready to work on this

RafaelJohn9:
Hello here 👋 ,

I'm guessing the issue is not open to contributions anymore ?

PR #20689, opened 2025-10-03
Refactor Rust lint test structure to use RuffTestFixture

Renkai:
Part of: https://github.com/astral-sh/ruff/issues/19016, for lint.rs


MichaReiser:
Thanks for working on this. Yes, this is what I had in mind.

Renkai:
Hi @MichaReiser , not sure why I can't modify the reviewers label, PTAL

Renkai:
Can't help on modify the windows insta snapshots very well, because I don't own a windows computer

MichaReiser:
> Can't help on modify the windows insta snapshots very well, because I don't own a windows computer

You can edit snapshots by hand. It's a bit pain but it should be okay if it's only that `/` is replaced with `TMP/`. 

For the other changes. It looks like there's a missing filter somewhere where the base path isn't filtered out anymore.

Renkai:
> > Can't help on modify the windows insta snapshots very well, because I don't own a windows computer
> 
> You can edit snapshots by hand. It's a bit pain but it should be okay if it's only that `/` is replaced with `TMP/`.
> 
> For the other changes. It looks like there's a missing filter somewhere where the base path isn't filtered out anymore.

I fixed the filter problem, but for some tests, windows and linux have different behavior, the linux tests expect `ruff.toml` but windows expect `[TMP]/ruff.toml`. I doubt the `fs::relativize_path` method does not work well in some cases for windows.

https://github.com/Renkai/ruff/blob/2d3458f242188f9a3cdda46ada0aefacafd7f406/crates/ruff_workspace/src/configuration.rs#L1605

Renkai:
In the new pattern, there are four functions have different behavior between windows and linux
- line_too_long_width_override
- mixed_levels
- precedence
- top_level_options
I reverted them to the main branch version. It's most likely caused by `fs::relativize_path` can't work well in some cases, I recommend to create a new issue for them.

crates/ruff/tests/test_fixture.rs
```diff
@@ -0,0 +1,201 @@
+//! Test fixture utilities for ruff CLI tests
+//!
+//! The core concept is borrowed from ty/tests/cli/main.rs and can be extended
+//! with more functionality from there in the future if needed.
+
+#![allow(dead_code)]
```
MichaReiser: Let's remove the `allow` dead code and remove any unusede code instead

crates/ruff/tests/test_fixture.rs
```diff
@@ -0,0 +1,201 @@
...
+/// Creates a regex filter for replacing temporary directory paths in snapshots
+pub(crate) fn tempdir_filter(path: impl AsRef<Path>) -> String {
+    let path_str = path.as_ref().to_str().unwrap();
+
+    // Strip UNC prefix if present to normalize the path
+    let normalized_path = if let Some(stripped) = path_str.strip_prefix(r"\\?\") {
+        stripped
+    } else {
+        path_str
+    };
```
MichaReiser: What's the reason why you aren't using dunce here (you can add it as a dev-only dependency)

crates/ruff/tests/cli/main.rs
```diff
@@ -0,0 +1,201 @@
...
+        path_str
+    };
+
+    // Escape the normalized path
+    let escaped_path = escape(normalized_path);
+    // Replace literal backslashes in the escaped pattern with a character class that matches both
+    let cross_platform_pattern = escaped_path.replace(r"\\", r"[\\/]");
+
+    // Always add optional UNC prefix to match both \\?\C:\... and C:\... patterns
+    // This ensures compatibility regardless of whether input or output contains UNC prefix
```
MichaReiser: Rather than handling UNC paths here, I suggest to do the same as ty and strip any UNC prefix in `new`

crates/ruff/tests/test_fixture.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// - Environment isolation for consistent test behavior
+    pub(crate) fn new() -> Result<Self> {
+        let temp_dir = TempDir::new()?;
+
+        // Canonicalize the tempdir path because macOS uses symlinks for tempdirs
+        // and that doesn't play well with our snapshot filtering.
+        let project_dir = temp_dir
+            .path()
+            .canonicalize()
+            .context("Failed to canonicalize project path")?;
```
MichaReiser: Let's strip the UNC prefix here using dunce (you can add it as a dev dependency)

crates/ruff/tests/test_fixture.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///
+    /// * `path` - The relative path for the file
+    /// * `content` - The content to write to the file
+    pub(crate) fn write_file(&self, path: impl AsRef<Path>, content: &str) -> Result<()> {
+        let path = path.as_ref();
+        let file_path = self.project_dir.join(path);
+
+        Self::ensure_parent_directory(&file_path)?;
+
+        let content = content.trim_start_matches('\n');
```
MichaReiser: Can we use `dedent` here instead of trimming start new lines only?

crates/ruff/tests/test_fixture.rs
Renkai: The current functions are actually all used, not sure why the clippy believe they are dead, add some unit tests to make sure they are used now.

crates/ruff/tests/test_fixture.rs
Renkai: Nice crate, I didn't know it exists before.

PR #20692, opened 2025-10-03
[ty] Don't consider bound `self` parameter when comparing bound methods

### LINKED ISSUE #1301: Bound `self` should not be considered when checking assignability of bound methods
This seems related to #1169 and #1172, but slightly different and more general. It does not require overloads, nor having other parameters or a return value that use `typing.Self`.

We are currently not performing assignability/subtyping checks on bound methods correctly [[playground](https://play.ty.dev/11d587b4-4f0c-450a-a180-52b2c35edc91)]:

```py
# These succeed, as they should, since return types are in covariant position.
static_assert(is_subtype_of(Callable[[], int], Callable[[], object]))
static_assert(is_subtype_of(Callable[[], Iterator[int]], Callable[[], Iterator[object]]))

class I:
    def method1(self) -> int: ...
    def method2(self) -> Iterator[int]: ...

class O:
    def method1(self) -> object: ...
    def method2(self) -> Iterator[object]: ...

# These fail
reveal_type(I().method1)
reveal_type(O().method1)
static_assert(is_subtype_of(TypeOf[I().method1], TypeOf[O().method1]))
static_assert(is_subtype_of(TypeOf[I().method2], TypeOf[O().method2]))
```

Some printf-debugging shows that we are checking the `self` parameter when doing the assignability check. In this case, since parameters are in a contravariant position, we are deciding that `Self: int` is not a subtype of `Self: object`, and therefore causing the overall subtyping check to fail.

In a bound method, the bound `self` parameter has been curried away, and so shouldn't be considered at all when determining subtyping of the bound method.

dcreager:
In a bound method, the bound `self` parameter is no longer available for callers to provide an argument for. (That's the whole point of binding it!) That means we should not consider it when checking subtyping/assignability of bound methods.

We were already doing this correctly when comparing a bound method to other callable types, since we would coerce the bound method into its equivalent `Callable` type before performing the check, and our conversion into a `Callable` type correctly removes the bound `self` parameter from the signature.

But we weren't doing the same when comparing a bound method to another bound method — instead we were comparing the _unbound_ function signatures, and also comparing the bound `self` instance types. We can get the correct behavior (and eliminate having to duplicate logic) by coercing both bound methods into `Callable` types for these checks, too.

Closes https://github.com/astral-sh/ty/issues/1301

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of.md
@@ -1963,6 +1963,10 @@ static_assert(not is_subtype_of(type[A], Callable[[int], A]))
 
 ### Bound methods
 
+The bound `self` or `cls` parameter of a bound method has been curried away, and is no longer
+available as a parameter when calling it. It therefore is not considered when checking subtyping
+involving the bound method.
+
 ```py
 from typing import Callable
 from ty_extensions import TypeOf, static_assert, is_subtype_of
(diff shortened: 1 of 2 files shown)


dcreager:
@AlexWaygood convinced me synchronously that this is wrong, and so is https://github.com/astral-sh/ty/issues/1301! The bound methods of two different classes really are different sets of objects, and are correctly not subtypes of each other. It was a different issue that I was encountering on #20677.

PR #20697, opened 2025-10-04
[`flake8-simplify`] Skip `SIM911` when unknown arguments are present

### LINKED ISSUE #18778: [flake8-simplify] SIM911 false positive when methods have args
### Summary

`SIM911` has a false positive if you give the methods args, since they are removed when it is transformed into `.items()`, which changes the behavior of the code to no longer raise a `TypeError`. If there is a situation in where the code thinks the variable is a `dict`, but it's actually the `dict` class, that could actually show up in non-constructed code. I was unable to find such a situation though. [playground](https://play.ruff.rs/fb6f3473-53a1-401b-88ea-a943232e4b44)
```
PS ~\Desktop\New_folder>Get-Content issue.py
```
```py
d = {}
for country, stars in zip(d.keys(*x), d.values("hello")):...
```
```
PS ~\Desktop\New_folder>uvx ruff check issue.py --isolated --select SIM --fix --diff
```
```diff
--- issue.py
+++ issue.py
@@ -1,2 +1,2 @@
 d = {}
-for country, stars in zip(d.keys(*x), d.values("hello")):...
+for country, stars in d.items():...

Would fix 1 error.
```

### Version

ruff 0.12.0 (87f0feb 2025-06-17) + playground

TaKO8Ki:
## Summary

Fixes #18778

Prevent SIM911 from triggering when zip() is called on .keys()/.values() that take any positional or keyword arguments, so Ruff
never suggests the lossy rewrite.

## Test Plan

Added a test case to SIM911.py.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_simplify/rules/zip_dict_keys_and_values.rs
+++ b/crates/ruff_linter/src/rules/flake8_simplify/rules/zip_dict_keys_and_values.rs
@@ -78,13 +78,18 @@ pub(crate) fn zip_dict_keys_and_values(checker: &Checker, expr: &ast::ExprCall) 
     let [arg1, arg2] = &args[..] else {
         return;
     };
-    let Some((var1, attr1)) = get_var_attr(arg1) else {
+    let Some((var1, attr1, args1)) = get_var_attr_args(arg1) else {
         return;
     };
-    let Some((var2, attr2)) = get_var_attr(arg2) else {
+    let Some((var2, attr2, args2)) = get_var_attr_args(arg2) else {
         return;
     };
-    if var1.id != var2.id || attr1 != "keys" || attr2 != "values" {
+    if var1.id != var2.id
+        || attr1 != "keys"
+        || attr2 != "values"
+        || !args1.is_empty()
+        || !args2.is_empty()
+    {
         return;
     }
     if !checker.semantic().match_builtin_expr(func, "zip") {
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/flake8_simplify/rules/zip_dict_keys_and_values.rs
```diff
@@ -78,13 +78,18 @@ pub(crate) fn zip_dict_keys_and_values(checker: &Checker, expr: &ast::ExprCall)
...
+    let Some((var2, attr2, args2)) = get_var_attr_args(arg2) else {
         return;
     };
-    if var1.id != var2.id || attr1 != "keys" || attr2 != "values" {
+    if var1.id != var2.id
+        || attr1 != "keys"
+        || attr2 != "values"
+        || !args1.is_empty()
+        || !args2.is_empty()
+    {
```
TaKO8Ki, line 92: I initially thought I have to modify SIM911  to skip only the autofix, but I found a similar implementation here and aligned with that instead. If you’d prefer we go with it, I can update the implementation accordingly.

https://github.com/astral-sh/ruff/blob/00f702e9fff0837b3e1aa41ea65ec300665f9ace/crates/ruff_linter/src/rules/flake8_simplify/rules/key_in_dict.rs#L68

crates/ruff_linter/src/rules/flake8_simplify/rules/zip_dict_keys_and_values.rs
```diff
@@ -122,8 +127,11 @@ pub(crate) fn zip_dict_keys_and_values(checker: &Checker, expr: &ast::ExprCall)
     )));
 }
 
-fn get_var_attr(expr: &Expr) -> Option<(&ExprName, &Identifier)> {
-    let Expr::Call(ast::ExprCall { func, .. }) = expr else {
+fn get_var_attr_args(expr: &Expr) -> Option<(&ExprName, &Identifier, &Arguments)> {
+    let Expr::Call(ast::ExprCall {
+        func, arguments, ..
+    }) = expr
+    else {
```
ntBre: Instead of adding `arguments` to the return type, would it be easier just to return `None` if the arguments aren't empty here? I think that would have the same effect.

crates/ruff_linter/src/rules/flake8_simplify/rules/zip_dict_keys_and_values.rs
ntBre: Either way works, I guess it makes sense to be consistent with the other SIM rule.

PR #20702, opened 2025-10-05
[ruff] Add rules for checking slashes in `path()` for Django

jvacek:
## Summary
The MR adds two new rules: DJ014 and DJ015. These two rules enforce the slash consistency when making django paths.

There was a discussion in Discord about whether the upstream flake8-django package was abandoned yet or not. See here: https://discord.com/channels/1039017663004942429/1039017663512449056/1422941869091979284

I am still wondering if it makes sense to add these under `DJ` directly, and not `XDJ` or something like that. Happy to make a backport package for flake8 for this as well.
## Test Plan
- Added fixtures and snapshots


jvacek:
Can anyone help me figure out what's up with the output in the comment above? I think it might be incorrectly removing parts of the ouptut as all of the `<foo:bar>` bits in the message are being removed, [but I do have them in the snapshots](https://github.com/astral-sh/ruff/pull/20702/files#diff-aa1602dea69d60eff8f64b144ae25343690f74b4d98de60449468e173aebdd98R142)

e.g., the third line has this
```
+ analytics/urls.py:29:14: DJ014 [*] URL route `stats/remote//installation` is missing a trailing slash
```
but should be showing this instead
```
+ analytics/urls.py:29:14: DJ014 [*] URL route `stats/remote/<int:remote_server_id>/installation` is missing a trailing slash
```

ntBre:
> Can anyone help me figure out what's up with the output in the comment above? I think it might be incorrectly removing parts of the ouptut as all of the `<foo:bar>` bits in the message are being removed, [but I do have them in the snapshots](https://github.com/astral-sh/ruff/pull/20702/files#diff-aa1602dea69d60eff8f64b144ae25343690f74b4d98de60449468e173aebdd98R142)

I haven't looked super closely yet, but the `<foo:bar>` bits look kind of like markdown links. They might be getting stripped out somewhere in the ecosystem report generation if similar cases are working in the snapshot tests.

jvacek:
@ntBre Should I add some explicit note about these not being in the upstream package?

ntBre:
I think it's okay, but I'd have to double check what we've done for similar cases in the past.

jvacek:
I'm gonna pull this down to Draft until I can get the CI to run

jvacek:
I removed the following check from the rules as I noticed that a lot of our `urls.py` files don't import anything from django.

This is a deviation from the other _django_ rules, but I believe the path checks will be sufficient for short-circuiting in non-django files, and the one less check is also an improvement for Django files. Curious what you think about this?

https://github.com/astral-sh/ruff/pull/20702/commits/cfe53bb43528be191e9d3358e0110fdbe2de7364#diff-f0f08bc80e806c09c11295057e6c8c9361be949dd66335a9e3ec9b49443d8a0fL63-L65

crates/ruff/tests/cli/snapshots/cli__lint__requires_python_no_tool.snap
```diff
@@ -12,7 +12,6 @@ info:
     - UP007
     - test.py
     - "-"
-snapshot_kind: text
```
jvacek, line 15: I'm not sure what would cause these to be removed, is this intended?

ntBre: Yeah, newer versions of cargo-insta don't add this line, but it will linger around until someone updates a snapshot! So this is expected

jvacek:
@ntBre need anything else from me here still?

ntBre:
I don't think so, I just need some time to review :)

UnknownPlatypus:
Should these rule be added in the DJ1xx range in case the pylint plugin regain some traction to avoid conflicting rule codes ?

ntBre:
There was some discussion on Discord about that. It looks like the upstream's last commit was January 8th, 2024, so almost two years ago. That seemed long enough to me to add these in the DJ group, but I don't feel too strongly and am happy to reconsider if others do!

UnknownPlatypus:
I think it's fine to add to the DJ group too (otherwise it will add friction and confusion for people looking for Django related rules) but I would add new rules as code `DJ100` and `DJ101` instead of `DJ014` and `DJ015` to be extra-careful and maybe better convey these are not in the initial flake8 implementation. 
Maybe it's not necessary, feel free to ignore.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+/// urlpatterns = [
+///     path("help/", views.help_view),
+///     path("about/", views.about_view),
+/// ]
+/// ```
+///
+/// ## References
+/// - [Django documentation: URL dispatcher](https://docs.djangoproject.com/en/stable/topics/http/urls/)
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "v0.14.1")]
```
ntBre:
```suggestion
#[violation_metadata(preview_since = "0.14.4")]
```

We stopped using the `v` prefix a while back. Maybe I should try to add some validation in this macro, it's very new!

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+pub(crate) struct DjangoURLPathWithLeadingSlash {
```
ntBre, line 45: I think the Rust convention is actually to use `Url`, even if it seems a bit weird.

```suggestion
pub(crate) struct DjangoUrlPathWithLeadingSlash {
```

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl AlwaysFixableViolation for DjangoURLPathWithLeadingSlash {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let DjangoURLPathWithLeadingSlash { url_pattern } = self;
+        format!("URL route `{url_pattern}` has an unnecessary leading slash")
+    }
+
+    fn fix_title(&self) -> String {
+        "Remove leading slash".to_string()
```
ntBre, line 57: nit:


```suggestion
        "Remove the leading slash".to_string()
```

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+
+            // Check if it matches any additional configured path functions
+            let qualified_name_str = segments.join(".");
+            checker
+                .settings()
+                .flake8_django
+                .additional_path_functions
+                .iter()
+                .any(|path| path == &qualified_name_str)
+        });
```
ntBre, line 83: I found a couple of other rules where we do something like this:

https://github.com/astral-sh/ruff/blob/f70034f0476cffc07ed9d30e4922e7b9db54ee6e/crates/ruff_linter/src/rules/flake8_bandit/rules/unsafe_markup_use.rs#L124-L134

Basically it might be better to convert the `additional_path_functions` to `QualifiedName`s too instead of converting the `qualified_name` back to a string.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    if !is_path_function {
+        return;
+    }
+
+    // Get the first argument (the route pattern)
+    let Some(route_arg) = call.arguments.args.first() else {
```
ntBre, line 90: If this can also be provided as a keyword we should use:

https://github.com/astral-sh/ruff/blob/f70034f0476cffc07ed9d30e4922e7b9db54ee6e/crates/ruff_python_ast/src/nodes.rs#L3325-L3327

Or we could use `find_positional` just above it, if it's positional-only.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+/// urlpatterns = [
+///     path("help/", views.help_view),
+///     path("about/", views.about_view),
+/// ]
+/// ```
+///
+/// ## References
+/// - [Django documentation: URL dispatcher](https://docs.djangoproject.com/en/stable/topics/http/urls/)
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "v0.14.4")]
```
ntBre, line 44:
```suggestion
#[violation_metadata(preview_since = "0.14.4")]
```

We dropped the v prefix a while back. Maybe I should try to enforce this in the macro, it's new!

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+        return;
+    }
+
+    // Get the first argument (the route pattern)
+    let Some(route_arg) = call.arguments.args.first() else {
+        return;
+    };
+
+    // Check if it's a string literal
+    if let Expr::StringLiteral(ast::ExprStringLiteral { value, .. }) = route_arg {
```
ntBre, line 95: We could use another let-else here to avoid some nesting, but I'm happy either way.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+            // Determine the quote style to find the insertion point for removal
+            let string_content = checker.locator().slice(route_arg.range());
+            let quote_len =
+                if string_content.starts_with("'''") || string_content.starts_with("\"\"\"") {
+                    3
+                } else if string_content.starts_with('\'') || string_content.starts_with('"') {
+                    1
+                } else {
+                    return; // Invalid string format
+                };
```
ntBre, line 122: I think we could use something like this here:

```suggestion
            let quote_len = value.first_literal_flags().quote_str().text_len();
```

crates/ruff_linter/src/rules/flake8_django/rules/url_path_with_leading_slash.rs
```diff
@@ -0,0 +1,131 @@
...
+
+            // Remove the leading slash (after the opening quote(s))
+            let removal_start = route_arg.range().start() + TextSize::new(quote_len);
+            let removal_end = removal_start + TextSize::new(1); // Remove one character (the slash)
+
+            diagnostic.set_fix(Fix::safe_edit(Edit::deletion(removal_start, removal_end)));
```
ntBre, line 128: Should this be an unsafe edit? From the rule description and what I found in the docs, it seems like this could change program behavior, making it unsafe. But maybe I'm being too pedantic, especially if the original behavior was probably unintentional.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+///     path("help/", views.help_view),
+///     path("about/", views.about_view),
+/// ]
+/// ```
+///
+/// ## References
+/// - [Django documentation: URL dispatcher](https://docs.djangoproject.com/en/stable/topics/http/urls/)
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "v0.14.4")]
+pub(crate) struct DjangoURLPathWithoutTrailingSlash {
```
ntBre, line 45:
```suggestion
#[violation_metadata(preview_since = "0.14.4")]
pub(crate) struct DjangoUrlPathWithoutTrailingSlash {
```

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+
+            // Check if it matches any additional configured path functions
+            let qualified_name_str = segments.join(".");
+            checker
+                .settings()
+                .flake8_django
+                .additional_path_functions
+                .iter()
+                .any(|path| path == &qualified_name_str)
+        });
```
ntBre, line 83: It might make sense to factor this out into a helper function if it's used in both rules.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    if !is_path_function {
+        return;
+    }
+
+    // Get the first argument (the route pattern)
+    let Some(route_arg) = call.arguments.args.first() else {
+        return;
+    };
```
ntBre, line 92: Same as  above, I think we can use one of our argument helper functions here.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+        // Determine the quote style to find the insertion point for the slash
+        // (just before the closing quotes)
+        let string_content = checker.locator().slice(route_arg.range());
+        let quote_len = if string_content.ends_with("'''") || string_content.ends_with("\"\"\"") {
+            3
+        } else if string_content.ends_with('\'') || string_content.ends_with('"') {
+            1
+        } else {
+            return; // Invalid string format
+        };
```
ntBre, line 125: Same suggestion as above again, using the quote flags.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        // Insert "/" just before the closing quote(s)
+        let insertion_point = route_arg.range().end() - TextSize::new(quote_len);
```
ntBre, line 128: I'm pretty sure this is impossible to hit given the checks above, but we might want to use a `saturating_sub` or `checked_sub` here just in case the `route_arg.range().end()` somehow ends up less than `quote_len`.

I've caused a panic like this myself recently 😅

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+        diagnostic.set_fix(Fix::safe_edit(Edit::insertion(
```
ntBre, line 129: This one sounded safer to me in the docs, but just flagging to double check the safety of both rules.

crates/ruff_linter/src/rules/flake8_django/rules/url_path_without_trailing_slash.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl AlwaysFixableViolation for DjangoURLPathWithoutTrailingSlash {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let DjangoURLPathWithoutTrailingSlash { url_pattern } = self;
+        format!("URL route `{url_pattern}` is missing a trailing slash")
+    }
+
+    fn fix_title(&self) -> String {
+        "Add trailing slash".to_string()
```
ntBre, line 57:
```suggestion
        "Add a trailing slash".to_string()
```

crates/ruff_workspace/src/options.rs
```diff
@@ -900,6 +900,10 @@ pub struct LintCommonOptions {
     #[option_group]
     pub flake8_copyright: Option<Flake8CopyrightOptions>,
 
+    /// Options for the `flake8-django` plugin.
+    #[option_group]
+    pub flake8_django: Option<Flake8DjangoOptions>,
```
ntBre, line 905: It's easy to miss this, but I think we're not supposed to add new settings here:

https://github.com/astral-sh/ruff/blob/f70034f0476cffc07ed9d30e4922e7b9db54ee6e/crates/ruff_workspace/src/options.rs#L867-L870

We should add this here instead:

https://github.com/astral-sh/ruff/blob/f70034f0476cffc07ed9d30e4922e7b9db54ee6e/crates/ruff_workspace/src/options.rs#L499-L505

crates/ruff_workspace/src/options.rs
```diff
@@ -1431,6 +1435,39 @@ impl Flake8CopyrightOptions {
...
+    /// Django's path function, you can add `"mytools.path"` to this list.
+    #[option(
+        default = "[]",
+        value_type = "list[str]",
+        example = r#"
+            # Allow checking URL patterns from custom path functions
+            additional-path-functions = ["mytools.path", "myapp.urls.custom_path"]
+        "#
+    )]
+    pub additional_path_functions: Option<Vec<String>>,
```
ntBre, line 1460:
```suggestion
    pub extend_path_functions: Option<Vec<String>>,
```

I think `extend` is the more typical prefix for these settings. [extend-immutable-calls](https://docs.astral.sh/ruff/settings/#lint_flake8-bugbear_extend-immutable-calls) and [extend-allowed-calls](https://docs.astral.sh/ruff/settings/#lint_flake8-boolean-trap_extend-allowed-calls) are a couple of examples I looked at.

crates/ruff_linter/resources/test/fixtures/flake8_django/DJ100.py
```diff
@@ -0,0 +1,46 @@
...
+    path("bad", views.bad_view),  # DJ100
+    path("also-good/", views.also_good_view),
+    path("also-bad", views.also_bad_view),  # DJ100
+]
+
+# Error - missing trail slash and argument should stay in message
+urlpatterns_params_bad = [
+    path("bad/<slug:slug>", views.bad_view),  # DJ100
+    path("<slug:slug>", views.bad_view),  # DJ100
+]
```
ntBre, line 46: Could we add some tests like these:

```pycon
>>> "/path\
... /".endswith("/")
True
>>> "\
... /path".startswith("/")
True
```

I think these could potentially cause problems with the string indexing we're doing in the fixes. One blunt solution would be just to skip fixes when there's a backslash in the route string.

MichaReiser:
Thank you for working on this. I'll close this PR due to inactivity. Feel free to resubmit it once the changes are addressed and we'll happily take another look

PR #20704, opened 2025-10-05
Fix false negatives in `Truthiness::from_expr` for lambdas, generators, and f-strings

### LINKED ISSUE #20703: `Truthiness:from_expr` has false negatives
### Summary

[`Truthiness:from_expr`](https://github.com/astral-sh/ruff/blob/0.13.3/crates/ruff_python_ast/src/helpers.rs#L1217) has false negatives for certain expressions that are guaranteed to be truthy: generator expressions, lambda expressions, f-string interpolations containing byte strings (even if they are empty), and f-string interpolations containing debug text. This affects [`subprocess-popen-with-shell-equals-true` (S602)](https://docs.astral.sh/ruff/rules/subprocess-popen-with-shell-equals-true/), [`call-with-shell-equals-true` (S604)](https://docs.astral.sh/ruff/rules/call-with-shell-equals-true/), [`unix-command-wildcard-injection` (S609)](https://docs.astral.sh/ruff/rules/unix-command-wildcard-injection/), and [`expr-or-true` (SIM222)](https://docs.astral.sh/ruff/rules/expr-or-true/), at least. [Example](https://play.ruff.rs/3e39aaa8-8e77-4c80-866d-3cec57cfa924):

```console
$ cat >truthiness_fn.py <<'# EOF'
import subprocess
subprocess.run("echo | xxd -p", shell=(x for x in ()))
print(bool(dict(shell=lambda: 0)["shell"]))
print(subprocess.check_output("echo chmod 777 truthiness_*.py", shell=f"{b""}"))
print(f"{""=}" or 0)
# EOF

$ python truthiness_fn.py
0a
True
b'chmod 777 truthiness_fn.py\n'
""=''

$ ruff --isolated check truthiness_fn.py --select S602,S604,S609,SIM222
All checks passed!
```

### Version

ruff 0.13.3 (188c0dce2 2025-10-02)

danparizher:
## Summary

Fixes #20703

### DIFF
--- a/crates/ruff_python_ast/src/helpers.rs
+++ b/crates/ruff_python_ast/src/helpers.rs
@@ -1402,8 +1402,7 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
             f_string.elements.iter().all(|element| match element {
                 InterpolatedStringElement::Literal(string_literal) => !string_literal.is_empty(),
                 InterpolatedStringElement::Interpolation(f_string) => {
-                    f_string.debug_text.is_some()
-                        || inner(&f_string.expression)
+                    f_string.debug_text.is_some() || inner(&f_string.expression)
                 }
             })
         }


crates/ruff_python_ast/src/helpers.rs
dylwil3, line 1: Thank you! This looks good. Instead of these unit tests, could you pick some of the affected rules and add snapshot tests?

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1397,7 +1401,11 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
         ast::FStringPart::FString(f_string) => {
             f_string.elements.iter().all(|element| match element {
                 InterpolatedStringElement::Literal(string_literal) => !string_literal.is_empty(),
-                InterpolatedStringElement::Interpolation(f_string) => inner(&f_string.expression),
+                InterpolatedStringElement::Interpolation(f_string) => {
+                    f_string.debug_text.is_some()
+                        || !matches!(f_string.conversion, ast::ConversionFlag::None)
```
dylwil3: I'm not sure I understand this line. Which of the deductions in `inner` is incorrect for string, ascii, or `repr` conversion flags?

danparizher: When an f-string interpolation has a conversion flag like `!s`, `!a`, or `!r`, Python's `str()`, `ascii()`, and `repr()` functions always produce a non-empty string representation of the object.

This means that all the complex expressions in the `inner` function (like `BoolOp`, `BinOp`, `Call`, `Attribute`, etc.) that normally return `false` because we can't determine if they're empty become guaranteed to be non-empty when conversion flags are applied.

The conversion flags essentially override the uncertainty by ensuring the result is always a meaningful string representation, which is why the condition correctly returns `true` whenever any conversion flag is present.

dscorbett: All three conversions can produce empty strings.
```console
$ cat >counterexamples.py <<'# EOF'
from configparser import Error
error = Error()
print(len(f"{error!s}"))
print(len(f"{error!a}"))
print(len(f"{error!r}"))
# EOF

$ python counterexamples.py
0
0
0
```

danparizher: Thanks for this; I was wrong about that then. In that case, we should have the condition removed.

dylwil3: I don't think those counterexamples are relevant here, as far as I can tell. The function in question returns `true` when we can be _certain_ that the f-string is nonempty, and `false` otherwise. It already returns false for `Name` expressions, so it would already be okay for `counterexamples.py` with or without worrying about conversion flags.

My claim is that the expressions for which we return `true` would be unaffected by the presence of these flags (insofar as their boolean value).

PR #20715, opened 2025-10-06
[ty] Enforce that `typing_extensions` must come from a stdlib search path

AlexWaygood:
## Summary

I _believe_ this fixes the panic in https://github.com/astral-sh/ty/issues/1289. It's what mypy does, anyway, and it seems like reasonable behaviour. (I don't want to close that issue, though, because I think we should do other things as well, such as improve our documentation around these options.)

## Test Plan

The following steps reproduce the panic on `main`:

1. `git clone https://github.com/copdips/ty-issues-1289`
2. `uv venv -p3.12`
3. `uv sync`
4. `cd backend`
5. `cargo run --manifest-path path/to/ruff/clone -p ty check`

If I then checkout this branch on my Ruff clone, the panic then no longer manifests when executing the final command.

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/resolver.rs
+++ b/crates/ty_python_semantic/src/module_resolver/resolver.rs
@@ -666,11 +666,15 @@ struct ModuleNameIngredient<'db> {
 /// Returns `true` if the module name refers to a standard library module which can't be shadowed
 /// by a first-party module.
 ///
-/// This includes "builtin" modules, which can never be shadowed at runtime either, as well as the
-/// `types` module, which tends to be imported early in Python startup, so can't be consistently
-/// shadowed, and is important to type checking.
+/// This includes "builtin" modules, which can never be shadowed at runtime either, as well as
+/// certain other modules that are involved in an import cycle with `builtins` (`types`,
+/// `typing_extensions`, etc.). This latter set of modules cannot be allowed to be shadowed by
+/// first-party or "extra-path" modules, or we risk panics in unexpected places due to being
+/// unable to resolve builtin symbols. This is similar behaviour other other type checkers such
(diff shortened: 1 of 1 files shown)


AlexWaygood:
We _could_ consider just copying mypy completely here, and making all these modules non-shadowable: https://github.com/python/mypy/blob/3807423e9d98e678bf16b13ec8b4f909fe181908/mypy/build.py#L104-L117.

The underlying issue here is really one of documentation, though: the docs for the `extra-paths` option should mention the `python` option; possibly we should add aliases for the `python` option to make it clear that you should use it to point to your virtual environment; possibly we should rename `extra-paths` to something like `non-environment-paths` so that it sounds less "inviting" to users. Given that I think we'll need to make at least some of those changes anyway, I think it probably makes sense to add modules to the non-shadowable denylist as and when issues like this come up, rather than overly hastily assuming that they'll all cause issues for us just because they cause issues for (some) other type checkers.

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -666,11 +666,15 @@ struct ModuleNameIngredient<'db> {
...
 /// by a first-party module.
 ///
-/// This includes "builtin" modules, which can never be shadowed at runtime either, as well as the
-/// `types` module, which tends to be imported early in Python startup, so can't be consistently
-/// shadowed, and is important to type checking.
+/// This includes "builtin" modules, which can never be shadowed at runtime either, as well as
+/// certain other modules that are involved in an import cycle with `builtins` (`types`,
+/// `typing_extensions`, etc.). This latter set of modules cannot be allowed to be shadowed by
+/// first-party or "extra-path" modules, or we risk panics in unexpected places due to being
+/// unable to resolve builtin symbols. This is similar behaviour other other type checkers such
```
AlexWaygood:
```suggestion
/// unable to resolve builtin symbols. This is similar behaviour to other type checkers such
```

PR #20718, opened 2025-10-06
[ty] Add `--venv` as an alias to `--python`

AlexWaygood:
## Summary

This PR adds `--venv` as an alias to `--python` on the ty CLI. The motivation is that if you're just skimming the online docs for the CLI interface, this may make it easier to spot at a glance that this option should be used to specify the location of your virtual environment. (See the changes to the generated `cli.md` file in this PR.)

We shouldn't rename the option to `--venv` because it can also be used to point to a system environment. It's tempting to add a `--venv-path` alias, which would match the name of one of pyright's options, but pyright's `--venv-path` option has different semantics, so that would probably cause more confusion rather than clarifying anything.

Helps with https://github.com/astral-sh/ty/issues/1289


MichaReiser (approved):
I don't love it but sounds reasonable (Do I remember it correctly that this is something you always wanted to do but I was grumpy and said no? 😆 )

AlexWaygood:
> (Do I remember it correctly that this is something you always wanted to do but I was grumpy and said no? 😆 )

Hehe, that sounds about right 😆

PR #20723, opened 2025-10-06
[ty] type-of-`self` in methods

sharkdp:


AlexWaygood:
Reopening to get a fresh ecosystem report now https://github.com/astral-sh/ruff/pull/20814 has landed

AlexWaygood:
> Reopening to get a fresh ecosystem report now #20814 has landed

Nice, we're down from 472 new `invalid-super-argument` diagnostics to only 4 🥳

AlexWaygood:
> Nice, we're down from 472 new `invalid-super-argument` diagnostics to only 4 🥳

...And now only 1 (which is a true positive), following https://github.com/astral-sh/ruff/pull/20843

PR #20725, opened 2025-10-06
Update default and latest Python versions for 3.14

### LINKED ISSUE #19467: Change `latest` Python version to Python 3.14
To finalize our stabilization of support for Python 3.14, we should address this `TODO`:

https://github.com/astral-sh/ruff/blob/5cace28c3ecc29f050aa5a57c8659e3154e3211e/crates/ruff_python_ast/src/python_version.rs#L58-L61

ntBre:
Summary
--

Closes #19467 and also removes the warning about using Python 3.14 without
preview enabled.

I also bumped `PythonVersion::default` to 3.9 because it reaches EOL this month,
but we could also defer that for now if we wanted.

The first three commits are related to the `latest` bump to 3.14; the fourth commit
bumps the default to 3.10.

Note that this PR also bumps the default Python version for ty to 3.10 because 
there was a test asserting that it stays in sync with `ast::PythonVersion`.

Test Plan
--

Existing tests

I spot-checked the ecosystem report, and I believe these are all expected. Inbits doesn't specify a target Python version, so I guess we're applying the default. UP007, UP035, and UP045 all use the new default value to emit new diagnostics.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_18.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_18.py
@@ -14,6 +14,6 @@ class Bar:
     ]
 
 
-# OK: Allow named expressions in annotations.
+# This is no longer allowed on Python 3.14+
 x: (y := 1)
 print(y)
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_26.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_26.py
@@ -13,16 +13,16 @@ CStr2: TypeAlias = Union["C", str]  # always okay
 
 # References to a class from inside the class:
 class C:
-    other: C = ...  # valid in a `.pyi` stub file, not in a `.py` runtime file
+    other: C = ...  # valid in a `.pyi` stub file, and in a `.py` runtime file with deferred annotations
     other2: "C" = ...  # always okay
-    def from_str(self, s: str) -> C: ...  # valid in a `.pyi` stub file, not in a `.py` runtime file
+    def from_str(self, s: str) -> C: ...  # valid in a `.pyi` stub file, and in a `.py` runtime file with deferred annotations
     def from_str2(self, s: str) -> "C": ...  # always okay
 
 # Circular references:
 class A:
-    foo: B  # valid in a `.pyi` stub file, not in a `.py` runtime file
+    foo: B  # valid in a `.pyi` stub file, and in a `.py` runtime file with deferred annotations
     foo2: "B"  # always okay
-    bar: dict[str, B]  # valid in a `.pyi` stub file, not in a `.py` runtime file
+    bar: dict[str, B]  # valid in a `.pyi` stub file, and in a `.py` runtime file with deferred annotations
     bar2: dict[str, "A"]  # always okay
 
 class B:
--- a/crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__F821_F821_18.py.snap
+++ b/crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__F821_F821_18.py.snap
@@ -4,7 +4,7 @@ source: crates/ruff_linter/src/rules/pyflakes/mod.rs
 F821 Undefined name `y`
   --> F821_18.py:19:7
    |
-17 | # OK: Allow named expressions in annotations.
+17 | # This is no longer allowed on Python 3.14+
 18 | x: (y := 1)
 19 | print(y)
    |       ^


crates/ruff_linter/src/rules/flake8_pyi/snapshots/ruff_linter__rules__flake8_pyi__tests__PYI055_PYI055.py.snap
```diff
@@ -1,6 +1,196 @@
 ---
 source: crates/ruff_linter/src/rules/flake8_pyi/mod.rs
 ---
+PYI055 [*] Multiple `type` members in a union. Combine them into one, e.g., `type[int | str | complex]`.
+ --> PYI055.py:4:4
+  |
+2 | from typing import Union
+3 |
+4 | s: builtins.type[int] | builtins.type[str] | builtins.type[complex]
```
ntBre, line 9: I think these are correct. The check disabling these on earlier Python versions was added in https://github.com/astral-sh/ruff/pull/6457:

https://github.com/astral-sh/ruff/blob/70f51e964855da881c40f87d3eeebf93520042bf/crates/ruff_linter/src/rules/flake8_pyi/rules/unnecessary_type_union.rs#L64-L67

ntBre (commented):
I talked with @dylwil3 on Discord, and we think the removed diagnostics are related to pre-existing issues. He tested at least some of them and found that the diagnostics are also suppressed on earlier Python versions with `from __future__ import annotations`, so I'm planning to open a follow-up issue and not block the release, if that sounds good to everyone. The rules that look suspicious to me (and I'll include in the follow-up issue) are:

- FAST001
- FAST003
- A003
- PTH210
- RUF001

In general, these look like cases where we no longer visit annotations because they aren't evaluated at runtime, but it seems like a worse result not to visit these annotations since it prevents us from emitting helpful diagnostics.

On the other side, the rules with changes that look correct to me are:
- PYI055
- TC001-TC004
- F821
- UP037
- UP006
- UP007
- UP036
- RUF058

Somewhere in between I'd put TC010. String union members are allowed on 3.14, but we probably still want to emit _some_ diagnostic for `"int" | str`; this doesn't seem to [trigger](https://play.ruff.rs/d397295d-b8c5-40b5-aca4-c3cad92c9474) any of the TC rules unless I'm missing some additional configuration.

One remaining question is whether it's better to accept these snapshots or pin them to 3.13. I think it's probably easier to pin for now and then remove the pins while addressing the follow-up issue, but I'm curious what others think.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -5479,7 +5479,7 @@ mod tests {
         Program::get(&db)
             .set_python_version_with_source(&mut db)
             .to(PythonVersionWithSource {
-                version: PythonVersion::latest_preview(),
+                version: PythonVersion::latest(),
```
sharkdp: Is the change here necessary? It seems to me like we would want to run this test on 3.15 preview soon?

sharkdp (approved):
The ty changes look good to me, thank you!

@AlexWaygood Are there any interactions with typeshed here that we should worry about? Going to a higher default version should be fine, I think?

AlexWaygood:
> @AlexWaygood Are there any interactions with typeshed here that we should worry about? Going to a higher default version should be fine, I think?

Yes, should be fine!

crates/ty/src/python_version.rs
AlexWaygood, line 1: should we not add a `Py314` variant to the `PythonVersion` enum in this file? We should definitely let ty users specify `--python-version=3.14`!

crates/ty/src/python_version.rs
```diff
@@ -7,9 +7,9 @@ pub enum PythonVersion {
     Py37,
     #[value(name = "3.8")]
     Py38,
-    #[default]
     #[value(name = "3.9")]
     Py39,
+    #[default]
```
MichaReiser, line 12: We also need to update the default value in the ruff settings documentation [here](https://github.com/astral-sh/ruff/blob/6b7a9dc2f29b2d8565c32c22189295049e3814a8/crates/ruff_workspace/src/options.rs#L330)

crates/ruff_python_parser/tests/fixtures.rs
```diff
@@ -40,7 +40,7 @@ fn inline_err() {
 fn test_valid_syntax(input_path: &Path) {
     let source = fs::read_to_string(input_path).expect("Expected test file to exist");
     let options = extract_options(&source).unwrap_or_else(|| {
-        ParseOptions::from(Mode::Module).with_target_version(PythonVersion::latest_preview())
+        ParseOptions::from(Mode::Module).with_target_version(PythonVersion::latest())
```
MichaReiser: I think it makes sense to keep `latest_preview` and use it here as we, very intentionally, want to use the latest preview version for parser tests.

MichaReiser:
I agree that we shouldn't block the release if those are pre-existing issues, and creating follow-up issues sounds great. I do think that it's somewhat important to look into what's happening. From a quick investigation, it seems that 

```rust
return dbg!(semantic.resolve_name(response_mode_name_expr))
            == dbg!(semantic.resolve_name(return_value_name_expr));
```

now returns `false` because `resolve_name` for the return value returns None. I haven't investigated further. @dylwil3 is this something you could take a look at. Unless @ntBre want's to fix it before the release ;)

Regarding the tests. For tests that regress, I'd prefer to use `test_path_with_settings_diff` so that it clearly demonstrates the undesired difference. The bugfix PR should then remove all the diffs.

The stringified union example is a bit funny (but also pre-existing). 

```py
x: "int" | str
```

is reported by TC010 but the following isn't, because the rule isn't in a runtime context:

```py

x: "int" | str
```

I'd say this is also a pre-existing issue that doesn't need to block the 3.14 release

MichaReiser (approved):

ntBre:
Thanks all for the reviews!

I think I've addressed all the feedback and will draft the follow-up issue while CI runs.

PR #20726, opened 2025-10-06
Add missing import to example doc

ShaharNaveh:
Feel free to close this PR if you think it's noise


carljm:
Thanks for the PR, but I don't think this helps the example communicate better. It's a snippet, not necessarily a complete module, and I think the intent of `secrets.randbelow(2)` is already sufficiently clear.

Issue #5734, opened 2023-07-13
Implement flake8-mock-spec's rules


sauloperez:
I want to enforce the use of the `autospec` argument on Python's unittest mocks in my team's codebase and so after a quick googling I found https://pypi.org/project/flake8-mock-spec/. Specifically, the reule `TMS021` is what we need.

It would be really good to give support for that plugin, as we are actively migrating to Ruff.

evanrittenhouse:
I can implement this.

neob91-close:
I'd love so see this get implemented!

kirschem-fernride:
Any progress on this?

Tenzer:
I have tried to start some work on this and I've made an incomplete draft PR in #20728. I'm hoping I can get a bit of feedback on what I'm doing looks correct, as this is my first time contributing to Ruff.

PR #20728, opened 2025-10-06
[ruff] Add flake8-mock-spec rules

Tenzer:
## Summary

This adds rules from [flake8-mock-spec](https://pypi.org/project/flake8-mock-spec/), as requested in #5734.

## Test Plan

This isn't currently tested, as it isn't finished yet.

Feed back would be appreciated about the current direction of this before I go further.

### DIFF
--- a/crates/ruff_linter/src/codes.rs
+++ b/crates/ruff_linter/src/codes.rs
@@ -1184,6 +1184,16 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Flake8Logging, "014") => (RuleGroup::Stable, rules::flake8_logging::rules::ExcInfoOutsideExceptHandler),
         (Flake8Logging, "015") => (RuleGroup::Stable, rules::flake8_logging::rules::RootLoggerCall),
 
+        // flake8-mock-spec
+        (Flake8MockSpec, "010") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::Mock),
+        (Flake8MockSpec, "011") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::MagicMock),
+        (Flake8MockSpec, "012") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::NonCallableMock),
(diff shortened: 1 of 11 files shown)


crates/ruff_linter/src/codes.rs
```diff
@@ -1184,6 +1184,16 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Flake8Logging, "014") => (RuleGroup::Stable, rules::flake8_logging::rules::ExcInfoOutsideExceptHandler),
         (Flake8Logging, "015") => (RuleGroup::Stable, rules::flake8_logging::rules::RootLoggerCall),
 
+        // flake8-mock-spec
+        (Flake8MockSpec, "010") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::Mock),
+        (Flake8MockSpec, "011") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::MagicMock),
+        (Flake8MockSpec, "012") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::NonCallableMock),
+        (Flake8MockSpec, "013") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::AsyncMock),
+        (Flake8MockSpec, "014") => (RuleGroup::Preview, rules::flake8_mock_spec::rules::ThreadingMock),
```
Tenzer, line 1192: This rule doesn't exist in flake8-mock-spec - probably because `ThreadingMock` only was added in Python 3.13 - so I have taken the liberty of using the next rule number for it.

MichaReiser:
Thank you for working on this. I think we should first come to a decision on the issue before working on a PR.

Tenzer:
It's not clear to me what outstanding decision there is to be made from the issue. As far as I can see there hasn't been any activity towards this for the last two years, so I thought it might be easiest to try to work a bit towards it and gather feedback on that - hence this PR.

MichaReiser:
The outstanding decision is whether we want to add the new mockspec plugin (does it fit into the rules we think should be part of Ruff. Every rule also adds a fair amount of maintenance burden)

Tenzer:
Ah, okay. Is there anything I can do to help getting a decision made on that?

PR #20730, opened 2025-10-07
[ty] Truncate type display for long unions in some situations

InvalidPathException:
## Summary

Fixes [astral-sh/ty#1307](https://github.com/astral-sh/ty/issues/1307)

Unions with length <= 5 are unaffected to minimize test churn
Unions with length > 5 will only display the first 3 elements + "... omitted x union elements"
Here "length" is defined as the number of elements after condensation to literals

Before:
```reveal_type(x)  # revealed: Literal[1, 2] | A | B | C | D | E | F | G```
```reveal_type(x)  # revealed: Result1A | Result1B | Result2A | Result2B | Result3 | Result4```
After:
```reveal_type(x)  # revealed: Literal[1, 2] | A | B | ... omitted 5 union elements```
```reveal_type(x)  # revealed: Result1A | Result1B | Result2A | ... omitted 3 union elements```

This formatting is consistent with `crates/ty_python_semantic/src/types/call/bind.rs` line 2992

## Test Plan

Cosmetic only, covered and verified by changes in mdtest

### CODE REVIEW

File: crates/ty_python_semantic/resources/mdtest/exception/control_flow.md [hunk 1]
```diff
@@ -522,7 +522,7 @@ try:
         reveal_type(x)  # revealed: Literal[2]
     reveal_type(x)  # revealed: Literal[2]
 except:
-    reveal_type(x)  # revealed: Literal[1, 2] | A | B | C | D | E | F | G
+    reveal_type(x)  # revealed: Literal[1, 2] | A | B | ... omitted 5 union elements
```
AlexWaygood: So, I actually think we should probably never truncate unions for `reveal_type` calls. A `reveal_type` call is the user explicitly asking us to tell them what the type of something is; it would be annoying if we only then displayed part of the type.

I think it's probably good for the default to be to truncate unions (as you already have it), but I think we need to still have an option where we can force unions to be displayed in full, without truncation. And we need to invoke that option for specific diagnostics such as `reveal_type`. We should be able to do this by extending our existing `DisplaySettings` struct, I think?

crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
```diff
@@ -86,7 +86,7 @@ error[call-non-callable]: Object of type `Literal[5]` is not callable
    |         ^^^^
    |
 info: Union variant `Literal[5]` is incompatible with this call site
-info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | (def f4[T](x: T@f4) -> int) | Literal[5] | (Overload[() -> None, (x: str) -> str]) | (Overload[() -> None, (x: str, y: str) -> str]) | PossiblyNotCallable`
+info: Attempted to call union type `(def f1() -> int) | (def f2(name: str) -> int) | (def f3(a: int, b: int) -> int) | ... omitted 5 union elements`
```
AlexWaygood: the problem here is that only two elements of this union are not callable (`Literal[5]` and `PossiblyNotCallable`), and they are both truncated away, so it's no longer clear what the problem is. I suppose the ideal display here would be something like

```suggestion
info: Attempted to call union type `(def f1() -> int) | <...3 union elements omitted...> | Literal[5] | <...3 union elements omitted...>`
```

Which indicates that we possibly need to have some way of filtering out union elements that do not satisfy a certain predicate...

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1212,19 +1215,41 @@ impl Display for DisplayUnionType<'_> {
             .filter(|element| is_condensable(*element))
             .collect::<Vec<_>>();
 
+        let total_entries =
+            usize::from(!condensed_types.is_empty()) + elements.len() - condensed_types.len();
+
+        if total_entries == 0 {
+            return Ok(());
+        }
```
AlexWaygood: I'm not sure this is possible? If a union has 0 elements, we should eagerly normalize the union into `Type::Never` when constructing the type

```suggestion
        assert_ne!(total_entries, 0);
```

AlexWaygood (changes requested):
Thanks! I think there's some subtleties here that we need to think through a bit

crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f…_-_Try_to_cover_all_pos…_-_Cover_non-keyword_re…_(707b284610419a54).snap
carljm: The previous line does specifically mention `Literal[5]`, so I'm not sure it's critical that the union display be that smart. In general I think when there's an error that applies to only some elements of the union, the diagnostic itself should call out the specific problematic union element(s), we shouldn't rely on union display for that. Showing the union is just additional context.

AlexWaygood: That's fair

crates/ty_python_semantic/src/types/display.rs
InvalidPathException: resolved

AlexWaygood (approved):
Looks excellent, thank you!! Will wait until tomorrow before merging in case anybody else has thoughts

AlexWaygood:
And this had exactly the effect on https://github.com/astral-sh/ruff/pull/20368 that I was hoping for. Some of the new diagnostics on that PR branch are so much better now -- thank you!

image

flying-sheep:
How do I disable this via config or CLI?

AlexWaygood:
> How do I disable this via config or CLI?

There's no way currently, but note that we always display unions without truncation if you call `reveal_type` on an object. Note also that `reveal_type` returns the object passed in, so you can use it in nested expressions: `reveal_type(x)[reveal_type(y)]` etc.

Issue #11753, opened 2024-06-05
Comprehension / generator, opt-in feature to preserve line breaks


Martin19037:
Before I begin, here is some context:

I would love to incorporate ruff as a pre-commit formatter for one of our projects. We developed the habit of almost always separating comprehension target-expression, for-in statement(s), and optional if-expression(s) into separate lines, except for very short and trivial comprehensions. Naturally, this makes identifying these separate comprehension parts very easy at a glance, at the cost of a few more lines. Also, we have increased our line length somewhat.

In its current form, ruff will always collapse comprehensions to a single line if:
- it is short enough to fit within the line length
- there are no "magic trailing commas" in any of the comprehension parts

We do use trailing commas ourselves, so many of our hand-formatted comprehensions that contain trailing commas in any form remain more or less unchanged after running ruff format. The ones that do not, usually get collapsed to a single line mess (also due to our increased line length).

One toy-example for this would be:
```python
[
    i + n
    for i in range(5)
    for n in range(5)
    if i < 3
]
```

which would be collapsed to a single line:
```python
[i + n for i in range(5) for n in range(5) if i < 3]
```

which is not as easy to recognize at a glance as the original example.

Would it be possible to have an opt-in feature to always preserve line breaks in comprehensions and generators? The behavior would be the same as if there were a trailing comma anywhere in the comprehension, and each comprehension-part is given a separate line, maybe even expand the whole comprehension into separate lines if there is at least one line break in it.

MichaReiser:
Thanks for the detailed write up. I can understand that nested comprehensions can get hard to read.

> Would it be possible to have an opt-in feature to always preserve line breaks in comprehensions and generators? 

Respecting line breaks would set a new precedence for ruff/black. There are other formatters that rely on line breaks. For example, prettier doesn't collapse an object expression if it has a line break after the last property's value:

```javascript
let a = { 
	a: 1,
  	b: 2
}
```

Whereas the following will be collapsed because it doesn't have a line break after `b`

```javascript

let a = { 
	a: 1,
  	b: 2 }
```

I could see a similar rule where the comprehension wouldn't collapse if there's a line break before the closing `]`, but I'm very hesitant of adding it because:

* It makes formatting changes non-reversible. A comprehension never collapses after the formatter expanded it once
* It's non obvious why the formatter doesn't collapse some comprehensions but does collapse others
* It will lead to less consistent code because users can now overrule the formatter. 

[...]

PR #16902, opened 2025-03-21
Implement flat width limits for comprehensions

aronhoff:
## Summary

Currently ruff formats comprehension expressions on a single line if they fit. This can squish a lot of complexity on the same line, making it harder to comprehend. There is also no way to force the comprehension to be expanded, like magic trailing commas for function calls. There has been a discussion of this in #11753.

One of the most viable suggestions in that issue (in my eyes at least) was to configure a specific width limit for the flat form of these expressions.
- If the expression fits in this width limit, it can be printed on one line (subject to all other usual limitations) .
- If the expression exceeds this limit, it is printed expanded. The specific limit no longer applies.

This pull request is a proof of concept implementation. I wanted to gather some feedback before finishing it up. Currently, flat width limits are implemented for:
- list comprehensions
- dict comprehensions
- set comprehensions
- generator expressions
- ternary expressions (inline ifs)

Please let me know if you have any feedback, for the approach or for the code! I am new to this codebase and to rust in general.

### Implementation details

The basis of the change is introducing a tag in `ruff_formatter`, `StartWidthLimitedBlock`/`EndWidthLimitedBlock` that can put an alternative line width limit on the formatter stack. This alternative line width limit is effectively `min(current_width_limit, current_cursor_position + expression_flat_width_limit)`. The alternative limit is removed from the stack at the end of the tag.

Beyond that, in `ruff_python_formatter`, this tag is only applied if the expression in question will end up being printed flat. If the expression ends up being printed in the expanded form, the tags are not emitted.

Out of scope for this PR, but I expect the formatting tag introduced here could be useful for simplifying the formatting of docstring code blocks.

## Test Plan

I added a few snapshot tests to verify that the approach works. This would likely need to be expanded.

PR #20731, opened 2025-10-07
Add support for optional multi-line comprehensions (#11753)

kevind-kizen:
Issue reference: https://github.com/astral-sh/ruff/issues/11753

This PR implements a new formatter option `comprehension-line-break` that allows users to preserve multi-line formatting in comprehensions, when they choose to do so, addressing #11753.

## Motivation

Currently, the ruff formatter automatically collapses multi-line comprehensions to a single line when they fit within the configured line width. This can reduce readability for complex comprehensions that were intentionally formatted across multiple lines for clarity. This change gives users control over this behavior.

### New Configuration Option

  Added comprehension-line-break with two modes:
  - "auto" (default): Current behavior - collapses comprehensions to single line if they fit
  - "preserve": Preserves multi-line formatting from source code

  Configuration in pyproject.toml:

```
[tool.ruff.format]
comprehension-line-break = "preserve"
```

Example Behavior

Input:
```
{
    obj["key"]: obj["value"]
    for obj
    in get_fields()
    if obj["key"] != "name"
}
```

Output with "auto" (default):
```
{obj["key"]: obj["value"] for obj in get_fields() if obj["key"] != "name"}
```

Output with "preserve":
```
{
    obj["key"]: obj["value"]
    for obj in get_fields()
    if obj["key"] != "name"
}
```

## Changes Made

  1. Added ComprehensionLineBreak enum in crates/ruff_python_formatter/src/options.rs
    - Implements Auto and Preserve modes
    - Includes serde serialization and JSON schema support
  2. Updated configuration system:
    - Added field to PyFormatOptions, FormatterSettings, and FormatConfiguration
    - Wired up configuration through workspace options
    - Updated JSON schema generation
  3. Modified all comprehension formatters to respect the new option:
    - Dictionary comprehensions (expr_dict_comp.rs)
    - List comprehensions (expr_list_comp.rs)
    - Set comprehensions (expr_set_comp.rs)
    - Generator expressions (expr_generator.rs)
  4. Implementation approach:
    - Added is_comprehension_multiline() helper to detect multi-line comprehensions in source
    - When preserve mode is active and source is multi-line, uses group().should_expand(true) to force expansion
    - Maintains backward compatibility with default auto mode
  5. Testing:
    - Added comprehensive test fixture covering all comprehension types
    - Tests both auto and preserve modes
    - All existing tests continue to pass

  Compatibility

  - Backward compatible: Default behavior unchanged
  - Black compatibility: auto mode maintains Black-compatible formatting
  - Performance: Minimal overhead - only checks source formatting when preserve mode is enabled

  Test Plan

  - Added test fixture: comprehension_line_break.py with corresponding snapshot tests
  - Tested with various comprehension types: dict, list, set, and generator expressions
  - Verified behavior with comments, nested comprehensions, and long expressions
  - All existing formatter tests pass
  - Ran ruff binary locally in my codebase

I know that this issue has not reached consensus on the philosophy, but wanted to drop an implementation draft to get the ball rolling as I believe this is a major readability flaw to always collapse comprehensions.

### DIFF
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/comprehension_line_break.options.json
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/comprehension_line_break.options.json
@@ -1,0 +1,8 @@
+[
+  {
+    "comprehension_line_break": "auto"
+  },
+  {
+    "comprehension_line_break": "preserve"
+  }
+]
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/comprehension_line_break.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/comprehension_line_break.py
@@ -1,0 +1,87 @@
+# Test file for comprehension-line-break option
+# Tests both "auto" (default) and "preserve" modes
+
+# Dictionary comprehensions
+dict_comp = {
+    obj["key"]: obj["value"]
+    for obj
+    in get_fields()
+    if obj["key"] != "name"
+}
+
+# This one should fit on one line in auto mode
+short_dict = {
+    k: v
+    for k, v in items()
+}
+
+# List comprehensions
+list_comp = [
+    item * 2
+    for item
+    in range(10)
+    if item % 2 == 0
+]
+
+# Set comprehensions
+set_comp = {
+    value.upper()
+    for value
+    in collection
+    if value
+}
+
+# Generator expressions
+gen_expr = (
+    x ** 2
+    for x
+    in numbers
+    if x > 0
+)
+
+# Nested comprehensions
+nested = [
+    [
+        (i, j)
+        for j
+        in range(3)
+    ]
+    for i
+    in range(5)
+]
+
+# Complex multiline with multiple clauses
+complex_comp = [
+    process(item)
+    for sublist
+    in data
+    for item
+    in sublist
+    if validate(item)
+    if not skip(item)
+]
+
+# Single-line comprehensions should remain single-line
+single_line_dict = {k: v for k, v in items()}
+single_line_list = [x for x in range(10)]
+single_line_set = {x for x in collection}
+single_line_gen = (x for x in numbers)
+
+# Comprehensions with comments
+commented = {
+    # Process key
+    key.lower():
+    # Process value
+    value.strip()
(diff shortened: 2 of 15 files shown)


MichaReiser:
Thanks for working on this but I'll close it as I consider presering line breaks as the absolute last resort because it defeats the purpose of a formatter -- to ensure code is consistently formatted. 

https://github.com/astral-sh/ruff/issues/11753#issuecomment-3015015216 suggests an alternative that I want to try first before going down the path of just preserving line breaks as is

kevind-kizen:
Fair enough -- not my project. I understand the philosophical goal of consistent formatting, but I also believe that if the consistent formatting is less readable, then that's suboptimal. I understand your formatter values standard opinionated formatting.

A mutli-line comprehension makes it super easy to visually scan each component of the comprehension, IMO. 

I see the alternative proposed and something like that would certainly be more deterministic, which is valuable. I'll add some comments on the issue to re-open debate, but it's obviously your call -- I would really like to see some solution to this as my new organization uses the ruff formatter.

aronhoff:
@kevind-kizen I implemented an idea similar to the rustfmt solution linked in #16902, except I used fixed width values rather than percentages and I didn't know how I could decide on the numbers in the comprehension width policy, so I left it configurable. I couldn't continue on that PR without guidance on what would be acceptable. Might be useful background if you want to work on it further.

kevind-kizen:
I can explore that direction, but the main criteria I typically use for
deciding to use a multi-line comprehension has more to do with the logical
complexity rather than simply line lengths.

I find that there are many cases where single line comprehensions are
appropriate, even with slightly longer line lengths, such as the case of
descriptive variable names. And there are some short comprehensions that
have a little extra complexity, for example function calls where I want
clarity on what function is being called.

I tend to prefer them in most cases.

I can provide some examples.

kevind-kizen:
I think that simply allowing the preservation of line breaks when the
developer makes the decision to break the lines along the comprehension
keywords (which is what I did in the PR) is very similar in philosophy to
the magic-trailing-commas configuration option.

kevind-kizen:
I am also curious about other developers choices — I might run an analysis
of popular GitHub python projects to get you some stats and try to better
understand the most common patterns.

PR #20732, opened 2025-10-07
[ty] Add support for functional `TypedDict` syntax

ibraheemdev:
## Summary

Adds support for the functional `TypedDict` syntax, e.g.
```py
Person = TypedDict("Person", { "name": str })

person: Person = { "name": "..." }
```

Part of https://github.com/astral-sh/ty/issues/154.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
@@ -1,6 +1,6 @@
 # Unsupported special types
 
-We do not understand the functional syntax for creating `NamedTuple`s, `TypedDict`s or `Enum`s yet.
+We do not understand the functional syntax for creating `NamedTuple`s or `Enum`s yet.
 But we also do not emit false positives when these are used in type expressions.
 
 ```py
--- a/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
@@ -1,7 +1,7 @@
 # Unsupported special types
 
-We do not understand the functional syntax for creating `NamedTuple`s or `Enum`s yet.
-But we also do not emit false positives when these are used in type expressions.
+We do not understand the functional syntax for creating `NamedTuple`s or `Enum`s yet. But we also do
+not emit false positives when these are used in type expressions.
 
 ```py
 import collections
--- a/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
@@ -10,9 +10,8 @@ import typing
 
 MyEnum = enum.Enum("MyEnum", ["foo", "bar", "baz"])
 MyIntEnum = enum.IntEnum("MyIntEnum", ["foo", "bar", "baz"])
-MyTypedDict = typing.TypedDict("MyTypedDict", {"foo": int})
 MyNamedTuple1 = typing.NamedTuple("MyNamedTuple1", [("foo", int)])
 MyNamedTuple2 = collections.namedtuple("MyNamedTuple2", ["foo"])
 
-def f(a: MyEnum, b: MyTypedDict, c: MyNamedTuple1, d: MyNamedTuple2): ...
+def f(a: MyEnum, c: MyNamedTuple1, d: MyNamedTuple2): ...
 ```
--- a/crates/ty_python_semantic/resources/mdtest/typed_dict.md
+++ b/crates/ty_python_semantic/resources/mdtest/typed_dict.md
@@ -13,6 +13,8 @@ from typing import TypedDict
 class Person(TypedDict):
     name: str
     age: int | None
+
+reveal_type(Person)  # revealed: <class 'Person'>
 ```
 
 New inhabitants can be created from dict literals. When accessing keys, the correct types should be
@@ -21,6 +23,7 @@ inferred based on the `TypedDict` definition:
 ```py
 alice: Person = {"name": "Alice", "age": 30}
 
+reveal_type(alice)  # revealed: Person
 reveal_type(alice["name"])  # revealed: str
 reveal_type(alice["age"])  # revealed: int | None
 
@@ -84,6 +87,91 @@ alice["extra"] = True
 # error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "extra""
 bob["extra"] = True
 ```
+
+## Functional
+
+You can also define a `TypedDict` using the functional syntax:
+
+```py
+from typing import TypedDict
+from typing_extensions import Required, NotRequired
+
+Person = TypedDict("Person", {"name": Required[str], "age": int | None})
+
+reveal_type(Person)  # revealed: typing.TypedDict
+```
+
+New inhabitants can be created from dict literals. When accessing keys, the correct types should be
+inferred based on the `TypedDict` definition:
+
+```py
+alice: Person = {"name": "Alice", "age": 30}
+
+reveal_type(alice)  # revealed: Person
+reveal_type(alice["name"])  # revealed: str
+reveal_type(alice["age"])  # revealed: int | None
+
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "non_existing""
+reveal_type(alice["non_existing"])  # revealed: Unknown
+```
+
+Inhabitants can also be created through a constructor call:
+
+```py
+bob = Person(name="Bob", age=25)
+
+reveal_type(bob["name"])  # revealed: str
+reveal_type(bob["age"])  # revealed: int | None
+
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "non_existing""
+reveal_type(bob["non_existing"])  # revealed: Unknown
+```
+
+Methods that are available on `dict`s are also available on `TypedDict`s:
+
+```py
+bob.update(age=26)
+```
+
+The construction of a `TypedDict` is checked for type correctness:
+
+```py
+# error: [invalid-argument-type] "Invalid argument to key "name" with declared type `str` on TypedDict `Person`"
+eve1a: Person = {"name": b"Eve", "age": None}
+# error: [invalid-argument-type] "Invalid argument to key "name" with declared type `str` on TypedDict `Person`"
+eve1b = Person(name=b"Eve", age=None)
+
+# error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `Person` constructor"
+eve2a: Person = {"age": 22}
+# error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `Person` constructor"
+eve2b = Person(age=22)
+
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "extra""
+eve3a: Person = {"name": "Eve", "age": 25, "extra": True}
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "extra""
+eve3b = Person(name="Eve", age=25, extra=True)
+```
+
+Assignments to keys are also validated:
+
+```py
+# error: [invalid-assignment] "Invalid assignment to key "name" with declared type `str` on TypedDict `Person`: value of type `None`"
+alice["name"] = None
+
+# error: [invalid-assignment] "Invalid assignment to key "name" with declared type `str` on TypedDict `Person`: value of type `None`"
+bob["name"] = None
+```
+
+Assignments to non-existing keys are disallowed:
+
+```py
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "extra""
+alice["extra"] = True
+
+# error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "extra""
+bob["extra"] = True
+```
+
 
 ## Nested `TypedDict`
 
@@ -277,6 +365,29 @@ Extra fields are still not allowed, even with `total=False`:
 invalid_extra = OptionalPerson(name="George", extra=True)
 ```
 
+`total` can also be set with the functional syntax:
+
+```py
+from typing import TypedDict
+
+OptionalPerson2 = TypedDict("OptionalPerson2", {"name": str, "age": int | None}, total=False)
+
+charlie = OptionalPerson2()
+david = OptionalPerson2(name="David")
+emily = OptionalPerson2(age=30)
+frank = OptionalPerson2(name="Frank", age=25)
+
+# TODO: we could emit an error here, because these fields are not guaranteed to exist
+reveal_type(charlie["name"])  # revealed: str
+reveal_type(david["age"])  # revealed: int | None
+
+# error: [invalid-argument-type] "Invalid argument to key "name" with declared type `str` on TypedDict `OptionalPerson2`"
+invalid = OptionalPerson2(name=123)
+
+# error: [invalid-key] "Invalid key access on TypedDict `OptionalPerson2`: Unknown key "extra""
+invalid_extra = OptionalPerson2(name="George", extra=True)
+```
+
 ## `Required` and `NotRequired`
 
 You can have fine-grained control over keys using `Required` and `NotRequired` qualifiers. These
@@ -329,6 +440,42 @@ Type validation still applies to all fields when provided:
 ```py
 # error: [invalid-argument-type] "Invalid argument to key "id" with declared type `int` on TypedDict `Message`"
 invalid_type = Message(id="not-an-int", content="Hello")
+```
+
+`Required`/`NotRequired` can also be set with the functional syntax:
+
+```py
+from typing_extensions import TypedDict, Required, NotRequired
+
+# total=False by default, but id is explicitly Required
+Message2 = TypedDict("Message2", {"id": Required[int], "content": str, "timestamp": NotRequired[str]}, total=False)
+
+# total=True by default, but content is explicitly NotRequired
+User2 = TypedDict("User2", {"name": str, "email": Required[str], "bio": NotRequired[str]})
+
+# Valid Message2 constructions
+msg1 = Message2(id=1)  # id required, content optional
+msg2 = Message2(id=2, content="Hello")  # both provided
+msg3 = Message2(id=3, timestamp="2024-01-01")  # id required, timestamp optional
+
+# Valid User2 constructions
+user1 = User2(name="Alice", email="alice@example.com")  # required fields
+user2 = User2(name="Bob", email="bob@example.com", bio="Developer")  # with optional bio
+
+reveal_type(msg1["id"])  # revealed: int
+reveal_type(msg1["content"])  # revealed: str
+reveal_type(user1["name"])  # revealed: str
+reveal_type(user1["bio"])  # revealed: str
+
+# error: [missing-typed-dict-key] "Missing required key 'id' in TypedDict `Message2` constructor"
+invalid_msg = Message2(content="Hello")  # Missing required id
+
+# error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `User2` constructor"
+# error: [missing-typed-dict-key] "Missing required key 'email' in TypedDict `User2` constructor"
+invalid_user = User2(bio="No name provided")  # Missing required name and email
+
+# error: [invalid-argument-type] "Invalid argument to key "id" with declared type `int` on TypedDict `Message2`"
+invalid_type = Message2(id="not-an-int", content="Hello")
 ```
 
 ## Structural assignability
@@ -388,6 +535,8 @@ reveal_type(alice["name"])  # revealed: str
 
 ### Reading
 
+`class.py`:
+
 ```py
 from typing import TypedDict, Final, Literal, Any
 
@@ -419,7 +568,40 @@ def _(person: Person, literal_key: Literal["age"], union_of_keys: Literal["age",
     reveal_type(person[unknown_key])  # revealed: Unknown
 ```
 
+`functional.py`:
+
+```py
+from typing import TypedDict, Final, Literal, Any
+
+Person = TypedDict("Person", {"name": str, "age": int | None})
+
+NAME_FINAL: Final = "name"
+AGE_FINAL: Final[Literal["age"]] = "age"
+
+def _(person: Person, literal_key: Literal["age"], union_of_keys: Literal["age", "name"], str_key: str, unknown_key: Any) -> None:
+    reveal_type(person["name"])  # revealed: str
+    reveal_type(person["age"])  # revealed: int | None
+
+    reveal_type(person[NAME_FINAL])  # revealed: str
+    reveal_type(person[AGE_FINAL])  # revealed: int | None
+
+    reveal_type(person[literal_key])  # revealed: int | None
+
+    reveal_type(person[union_of_keys])  # revealed: int | None | str
+
+    # error: [invalid-key] "Invalid key access on TypedDict `Person`: Unknown key "non_existing""
+    reveal_type(person["non_existing"])  # revealed: Unknown
+
+    # error: [invalid-key] "TypedDict `Person` cannot be indexed with a key of type `str`"
+    reveal_type(person[str_key])  # revealed: Unknown
+
+    # No error here:
+    reveal_type(person[unknown_key])  # revealed: Unknown
+```
+
 ### Writing
+
+`class.py`:
 
 ```py
 from typing_extensions import TypedDict, Final, Literal, LiteralString, Any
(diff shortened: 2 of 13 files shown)


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5625,6 +5592,165 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        if let Some(Type::SpecialForm(SpecialFormType::TypedDict)) = tcx.annotation {
+            let mut typed_dict_items = FxOrderMap::default();
+
+            for item in items {
+                let Some(Type::StringLiteral(key)) =
+                    self.infer_optional_expression(item.key.as_ref(), TypeContext::default())
+                else {
+                    // Emit a diagnostic here? We seem to support non-string literals.
+                    unimplemented!()
+                };
```
ibraheemdev: Should we be erroring here? pyright doesn't seem to support string constants as `TypedDict` keys, but we currently ignore non-literal keys when type-checking.

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_types.md
```diff
@@ -10,9 +10,8 @@ import typing
 
 MyEnum = enum.Enum("MyEnum", ["foo", "bar", "baz"])
 MyIntEnum = enum.IntEnum("MyIntEnum", ["foo", "bar", "baz"])
-MyTypedDict = typing.TypedDict("MyTypedDict", {"foo": int})
 MyNamedTuple1 = typing.NamedTuple("MyNamedTuple1", [("foo", int)])
 MyNamedTuple2 = collections.namedtuple("MyNamedTuple2", ["foo"])
 
-def f(a: MyEnum, b: MyTypedDict, c: MyNamedTuple1, d: MyNamedTuple2): ...
+def f(a: MyEnum, c: MyNamedTuple1, d: MyNamedTuple2): ...
```
sharkdp, line 16: We could also update the text above and remove `TypedDict`s from that enumeration.

crates/ty_python_semantic/src/types.rs
```diff
@@ -4746,7 +4764,7 @@ impl<'db> Type<'db> {
                             Parameter::positional_only(Some(Name::new_static("typename")))
                                 .with_annotated_type(KnownClass::Str.to_instance(db)),
                             Parameter::positional_only(Some(Name::new_static("fields")))
-                                .with_annotated_type(KnownClass::Dict.to_instance(db))
+                                .with_annotated_type(Type::SpecialForm(SpecialFormType::TypedDict))
```
sharkdp: This is necessary because we now infer the dict literal that is being passed to the `typing.TypedDict` constructor as a special form type?

crates/ty_python_semantic/src/types.rs
```diff
@@ -4833,11 +4853,18 @@ impl<'db> Type<'db> {
...
-            // TODO: some `SpecialForm`s are callable (e.g. TypedDicts)
-            Type::SpecialForm(_) => CallableBinding::not_callable(self).into(),
-
             Type::EnumLiteral(enum_literal) => enum_literal.enum_class_instance(db).bindings(db),
 
+            Type::KnownInstance(KnownInstanceType::TypedDictType(typed_dict)) => Binding::single(
+                self,
+                Signature::new(
+                    Parameters::new([Parameter::keyword_variadic(Name::new_static("kwargs"))
+                        .with_annotated_type(Type::any())]),
```
sharkdp: Using `**kwargs: Any` here implies that we won't get completions for parameters for these constructor calls. This also doesn't seem to work for class-based `TypedDict`s at the moment, so it's certainly something that we can implement later.

crates/ty_python_semantic/src/types.rs
```diff
@@ -9166,6 +9204,14 @@ impl Truthiness {
         }
     }
 
+    pub(crate) fn unwrap_or(self, value: bool) -> bool {
```
sharkdp, line 9239: This seems helpful. I was thinking about giving it another name (e.g. `disambiguate`), or just providing `fn as_bool(self) -> Option<bool>`, which would turn the callsite into `a.to_bool().unwrap_or(b)`, but I think I like your version best.

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -37,25 +45,54 @@ impl Default for TypedDictParams {
...
-#[derive(Debug, Copy, Clone, PartialEq, Eq, salsa::Update, Hash, get_size2::GetSize)]
-pub struct TypedDictType<'db> {
+#[derive(
+    Copy, Clone, Debug, Eq, PartialEq, Hash, salsa::Update, PartialOrd, Ord, get_size2::GetSize,
+)]
+pub enum TypedDictType<'db> {
     /// A reference to the class (inheriting from `typing.TypedDict`) that specifies the
     /// schema of this `TypedDict`.
-    defining_class: ClassType<'db>,
+    FromClass(ClassType<'db>),
```
sharkdp: I think we (I) used the term "class-based `TypedDict`" before, so maybe
```suggestion
    ClassBased(ClassType<'db>),
```
but I don't feel strongly.

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -37,25 +45,54 @@ impl Default for TypedDictParams {
...
+    pub(crate) fn from_class(class: ClassType<'db>) -> Self {
+        TypedDictType::FromClass(class)
     }
 
-    pub(crate) fn defining_class(self) -> ClassType<'db> {
-        self.defining_class
+    /// Returns an incomplete `TypedDictType` from its items.
+    ///
+    /// This is used to instantiate a `TypedDictType` from the dictionary literal passed to a
+    /// `TypedDict` constructor.
```
sharkdp: Maybe
```suggestion
    /// `typing.TypedDict` constructor (functional form for creating `TypedDict`s).
```

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -37,25 +45,54 @@ impl Default for TypedDictParams {
...
 impl<'db> TypedDictType<'db> {
-    pub(crate) fn new(defining_class: ClassType<'db>) -> Self {
-        Self { defining_class }
+    pub(crate) fn from_class(class: ClassType<'db>) -> Self {
+        TypedDictType::FromClass(class)
     }
 
-    pub(crate) fn defining_class(self) -> ClassType<'db> {
-        self.defining_class
+    /// Returns an incomplete `TypedDictType` from its items.
```
sharkdp: Is "incomplete" the same as "nameless"?
```suggestion
    /// Returns an incomplete (nameless) `TypedDictType` from its items.
```

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -37,25 +45,54 @@ impl Default for TypedDictParams {
...
-    pub(crate) fn items(self, db: &'db dyn Db) -> FxOrderMap<Name, Field<'db>> {
-        let (class_literal, specialization) = self.defining_class.class_literal(db);
-        class_literal.fields(db, specialization, CodeGeneratorKind::TypedDict)
+    /// Return the meta-type of this `TypedDict` type.
+    pub(super) fn to_meta_type(self, db: &'db dyn Db) -> Type<'db> {
+        // `TypedDict` instances are instances of `dict` at runtime, but its important that we
+        // understand a more specific meta type in order to correctly handle `__getitem__`.
+        match self {
+            TypedDictType::FromClass(class) => SubclassOfType::from(db, class),
+            TypedDictType::Synthesized(_) => KnownClass::TypedDictFallback.to_class_literal(db),
```
sharkdp: Hm, I'm wondering why this doesn't cause problems for dunder-calls on synthesized `TypedDict`s?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5625,6 +5592,165 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        dict: &ast::ExprDict,
+        tcx: TypeContext<'db>,
+    ) -> Option<Type<'db>> {
+        let ast::ExprDict {
+            range: _,
+            node_index: _,
+            items,
+        } = dict;
+
+        // Evaluate the dictionary literal passed to the `TypedDict` constructor.
```
sharkdp: It confused me a bit that this `infer_typed_dict_expression` function is now used for two rather distinct tasks:
* Inferring an incomplete synthesized `TypedDict` *schema* from a dict like `{"name": str, "age": int}` that is being passed to `typing.TypedDict`
* Inferring a `TypedDict` *inhabitant* type for a dict like `{"name": "Alice", "age": 25}` in the context of a `TypedDict` annotation.

Maybe those could be two separate functions? The only thing they share seems to be the `ast::ExprDict` unpacking above?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5625,6 +5592,165 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+
+                typed_dict_items.insert(ast::name::Name::new(key.value(self.db())), field);
+            }
+
+            // Create an incomplete synthesized `TypedDictType`, to be completed by the `TypedDict`
+            // constructor binding.
+            return Some(Type::TypedDict(TypedDictType::from_items(
+                self.db(),
+                typed_dict_items,
+            )));
```
sharkdp: `Type::TypedDict` refers to an *inhabitant* of a `TypedDict` type (not to its specification), so it feels strange/wrong to me to return this variant here, even if this type is not externally observable, since it will be passed to the constructor immediately. Introducing a completely new type variant to hold on to those nameless schemas seems like overkill, but maybe we could use a new `KnownInstanceType` variant?

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: From https://typing.python.org/en/latest/spec/typeddict.html#use-of-final-values-and-literal-types:

> Type checkers are only expected to support actual string literals, not final names or literal types, for specifying keys in a TypedDict type definition

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5625,6 +5592,165 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        .map(|_| Type::TypedDict(typed_dict))
+    }
+
+    fn infer_typed_dict_field_type_expression(
+        &mut self,
+        expr: &ast::Expr,
+    ) -> TypeAndQualifiers<'db> {
+        let ty = match expr {
+            ast::Expr::Subscript(subscript @ ast::ExprSubscript { value, slice, .. }) => {
+                let value_ty = self.infer_expression(value, TypeContext::default());
```
sharkdp: There's probably a reason we can't use `infer_annotation_expression` here (which would already handle `Required`, etc.)? I'm not sure I understand it, though. Are we forced to infer this as a normal expression?

sharkdp (approved):
This is fantastic — thank you very much.

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: "Only expected to support" is not the same as "must emit an error for anything else", though...

In general, we defer a lot of operations to type-check time that other type checkers attempt to do during (their equivalent of) semantic indexing. That puts limitations on other type checkers that we don't necessarily have for things like this, which is why language like this appears in the spec about what type checkers are "expected" to support

ibraheemdev: I changed the code to ignore these for now. I think eventually it makes sense to at least warn if we don't support them.

crates/ty_python_semantic/src/types.rs
ibraheemdev: Yes, if it was inferred as a dictionary literal we would not have access to its items. `TypedDict` works here as a type annotation because it is not otherwise allowed in type-form expressions. I added a comment.

crates/ty_python_semantic/src/types.rs
ibraheemdev: Added a TODO comment.

crates/ty_python_semantic/src/types/typed_dict.rs
ibraheemdev: I copied the `FromClass` and `Synthesized` naming from the `Protocol` enum, but also don't feel strongly.

crates/ty_python_semantic/src/types/typed_dict.rs
ibraheemdev: `KnownInstanceType::TypedDictType` is special-cased in `Type::bindings`, but I think it would otherwise go through `instance_fallback` to an instance of `KnownClass::TypedDictFallback`, not `meta_type` to its class literal.

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: I tried that at first, the problem is that we check this type against the type annotation of `SpecialFormType::TypedDict`, which `KnownInstanceType::TypedDictType` is not currently assignable to (falling back to an instance of `TypedDictFallback`). Maybe it's worth changing that, because it is quite counterintuitive?

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: This code is copied from `infer_annotation_expression`, but restricted to `Required`, `NotRequired`, and `ReadOnly` qualifiers. `infer_annotation_expression` has handling for other expressions and qualifiers as well, which I'm not sure we want to allow here?

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Hmm... it seems to me that it's rather the annotation which is wrong? `typing.TypedDict` is [not actually allowed](https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions) in type annotations. And if it were, I would rather expect it to be accepting inhabitants of `TypedDict`s, instead of dict literals that describe the schema of a `TypedDict`?

If I understand correctly, you want to specify *some* type in the annotation in order to later recognize it in the type context when you infer the type for the schema dict literal. Could we make up a new type for just for this purpose? Some new `SpecialForm` variant maybe? You would then be free to design assignability rules. Maybe @AlexWaygood has some ideas...

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: I see. Class-based `TypedDict`s would have the same limitation. We probably don't show an error if you use something like `name: Final[str]` at the moment.

Given that we don't emit errors for this case here either, I'm inclined to say I'd rather use `infer_annotation_expression` and maybe later pass down a "context" enum or a bitset parameter that would describe what kind of qualifiers are legal in the annotation that is currently being inferred?

I'll leave it up to you, though, if you want to change it or not.

crates/ty_python_semantic/src/types.rs
```diff
@@ -6836,6 +6867,9 @@ pub enum KnownInstanceType<'db> {
     /// A single instance of `typing.TypeAliasType` (PEP 695 type alias)
     TypeAliasType(TypeAliasType<'db>),
 
+    /// A single instance of `typing.TypedDict`.
+    TypedDictType(TypedDictType<'db>),
```
AlexWaygood: Hmm, I haven't read through this whole PR yet so I'm not sure what the correct doc-comment here should be, but I don't think this can be an accurate description of what `TypedDictType` represents... `TypedDict` is a function at runtime, so it is impossible to create an "instance of `TypedDict`". Calling `TypedDict` "as a function" at runtime creates a new class in exactly the same way as inheriting from `TypedDict` at runtime:

```pycon
>>> import typing
>>> type(typing.TypedDict)
<class 'function'>
>>> X = typing.TypedDict("X", {})
>>> X
<class '__main__.X'>
>>> type(X)
<class 'typing._TypedDictMeta'>
>>> class Y(typing.TypedDict): ...
... 
>>> Y
<class '__main__.Y'>
>>> type(Y)
<class 'typing._TypedDictMeta'>
```

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: @carljm suggesting using `typing.TypedDict` here *because* it isn't allowed in type-form expressions, so it will not conflict with real type annotations. We are also inferring the dictionary literal as if it is a `TypedDict` instance, so if we implemented the assignability rules for `typing.TypedDict`, this should work fine.

That said, you're right that this is somewhat of a hack, and we could instead use a special type for the annotation (and even a special type for the inferred dictionary literal, but that might be overkill).

sharkdp: > @carljm suggesting using `typing.TypedDict` here _because_ it isn't allowed in type-form expressions

I see. I was definitely quite confused by this hack in the beginning, but if there are no observable side-effects, I'm also fine with documenting it, and keeping it as is.

crates/ty_python_semantic/src/types.rs
AlexWaygood: It seems incorrect to me that we infer different types for `X` and `Y` (and different types for `X.__class__` and `Y.__class__`) here on your PR branch. They're both class objects in exactly the same way at runtime:

```py
from typing import TypedDict, reveal_type

X = TypedDict("X", {})

reveal_type(X)  # revealed: typing.TypedDict
reveal_type(X.__class__)  # revealed: TypedDictFallback

class Y(TypedDict): ...

reveal_type(Y)  # revealed: <class 'Y'>
reveal_type(Y.__class__)  # revealed: type
```

We previously experimented with making the `Type::ClassLiteral()` variant internally wrap an enum, so that we could express the fact that not all class-literal objects at runtime are created via class statements. https://github.com/astral-sh/ruff/pull/19998 turned out not to be the correct approach for `NewType`s (calls to `NewType` don't actually create classes!), but I think it could be the right approach here, and for the functional `enum.Enum` syntax, and for `collections.namedtuple()` calls, and for the functional `typing.NamedTuple` syntax, and for three-argument calls to `type()`, since those all _do_ actually create classes at runtime.

Having said all that, this would all complicate your approach (possibly by quite a lot), so I'm okay if you don't want to try making `Type::ClassLiteral` an enum. I only mention it because I think we'll have the same issue of "class objects created via function calls" for all these other things we'll need to support in the long term, too. So it may be worth digging in and trying to pull off that refactor at _some_ point.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: It's definitely really cool that we can use bidirectional inference here to control how type inference of a specific sub-expression of a call-expression works. I have no concrete things in mind, but I could certainly imagine that we could use this for other things as well (instead of detecting the right callable type, the precise parameter, etc.)

crates/ty_python_semantic/src/types.rs
ibraheemdev: Hmm, I see. Maybe I misread [this comment](https://github.com/astral-sh/ruff/pull/20523#issuecomment-3330619430), which I thought meant that we *don't* need to support definitionless classes for functional `TypedDict`s.

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: I also find this all a _little_ confusing in places, I'm afraid 😬

I actually feel more strongly about not returning a `Type::TypedDict` variant for the dictionary literal, though. I'm okay with (ab)using `SpecialForm::TypedDictType` for the type annotation if we document the hack properly, but I think I'd much prefer returning a `KnownInstanceType::TypedDictSchema` (feel free to bikeshed the name) for the dictionary literal that describes the schema. The dictionary literal _describes_ a `TypedDict`'s shape rather than _inhabiting_ a `TypedDict` -- reusing the `Type::TypedDict` variant for this purpose feels conceptually wrong, and error-prone

crates/ty_python_semantic/src/types.rs
carljm: Sorry, I think that was just my oversight in that comment, not thinking about the fact that functional TypedDict does actually create a class at runtime.

AlexWaygood: I can't speak for Carl, but I think possibly what he was saying there was that it turned out that `NewType` calls don't actually create new class objects afterall, so actually `NewType` _is_ different from functional `NamedTuple`/`enum`/`TypedDict`/classes created using three-argument `type()`. Meaning that in fact, https://github.com/astral-sh/ruff/pull/20126 doesn't help us here, because unlike #19998, it doesn't propose turning `ClassLiteral` into an enum

AlexWaygood: oops, I posted my reply before I saw Carl's!

ibraheemdev: I see, that makes a lot of sense. I would prefer to get this PR merged even if it is slightly incorrect (assuming it doesn't generate a lot of ecosystem false positives), but I'm happy to continue this work to support the `ClassLiteral` refactor. It probably makes sense to do that with functional `TypedDict`s first, rather than implement a new feature along with the refactor.

AlexWaygood: That SGTM, breaking it up into chunks will make it easier to review too!

crates/ty_python_semantic/resources/mdtest/typed_dict.md
```diff
@@ -97,10 +97,18 @@ from typing import TypedDict
...
-
 reveal_type(Person)  # revealed: typing.TypedDict
 ```
 
+The `TypedDict` schema must be passed directly as the second argument:
+
+```py
+fields = {"name": str}
+
+# error: [invalid-argument-type] "Argument is incorrect: Expected `_TypedDictSchema`, found `dict[Unknown | str, Unknown | <class 'str'>]`"
```
ibraheemdev, line 108: I'm not sure how we should display `TypedDictSchema` in diagnostics, given that it is an internal type. Ideally we would have a special diagnostic for this specific case, but I believe the `TypedDictSchema` type might still show up in the IDE in other cases.

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: I refactored this to use a new `SpecialForm::TypedDictSchema` and `KnownInstanceType::TypedDictSchema` type. Hopefully that makes it clearer.

ibraheemdev:
The main limitation of this PR based on the ecosystem report looks to be inheriting from a `TypedDict` created with the functional syntax:
```
error: [invalid-base] Invalid class base with type `typing.TypedDict` 
```

The errors are a bit unfortunate, but I'm not sure there's an easy way to avoid these false positives before the `ClassLiteral` refactor.

AlexWaygood:
> The main limitation of this PR based on the ecosystem report looks to be inheriting from a `TypedDict` created with the functional syntax:
> 
> ```
> error: [invalid-base] Invalid class base with type `typing.TypedDict` 
> ```
> 
> The errors are a bit unfortunate, but I'm not sure there's an easy way to avoid these false positives before the `ClassLiteral` refactor.

Could you add a branch to `ClassBase::try_from_type` similar to this one, to silence the false positives?

https://github.com/astral-sh/ruff/blob/7a347c43705125eadf3c4d4bad045f59d0f205c3/crates/ty_python_semantic/src/types/class_base.rs#L260-L264

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -1096,7 +1098,52 @@ impl<'db> Bindings<'db> {
...
+                            .collect::<FxOrderMap<_, _>>();
+
+                        overload.set_return_type(Type::KnownInstance(
+                            KnownInstanceType::TypedDictType(SynthesizedTypedDictType::new(
+                                db,
+                                Name::new(name.value(db)),
+                                params,
+                                items,
+                            )),
+                        ));
```
AlexWaygood, line 1146: I personally would not call the result of a `TypedDict("X", {"foo": int})` call a "synthesized" `TypedDict`. I don't think we've laid down a definition anywhere, but to me a "synthesized" type is one that we "spin out of thin air" during type checking, that doesn't actually correspond to a definition anywhere in the user's source code. But functional `TypedDict`s _do_ have definitions in source code -- it's just that the definitions are formed of assignments to function calls rather than class statements

Maybe rename the struct from `SynthesizedTypedDictType` to `FunctionalTypedDictType`?

crates/ty_python_semantic/src/types.rs
```diff
@@ -6486,8 +6542,9 @@ impl<'db> Type<'db> {
                 Protocol::Synthesized(_) => None,
             },
 
-            Type::TypedDict(typed_dict) => {
-                Some(TypeDefinition::Class(typed_dict.defining_class().definition(db)))
+            Type::TypedDict(typed_dict) => match typed_dict {
+                TypedDictType::FromClass(class) => Some(TypeDefinition::Class(class.definition(db))),
+                TypedDictType::Synthesized(_) => None,
```
AlexWaygood: Could we not add a `TypeDefinition::Assignment` variant to the `TypeDefinition` enum, and return that variant from this arm? That doesn't necessarily need to be done in this PR, but maybe we should add a TODO here and/or a followup issue about it? It would be nice for go-to-declaration to work with functional `TypedDict`s

crates/ty_python_semantic/src/types.rs
```diff
@@ -6836,6 +6893,13 @@ pub enum KnownInstanceType<'db> {
     /// A single instance of `typing.TypeAliasType` (PEP 695 type alias)
     TypeAliasType(TypeAliasType<'db>),
 
+    /// A single instance of `typing.TypedDict`.
```
```suggestion
    /// A single class object created using the `typing.TypedDict` functional syntax
```

AlexWaygood (approved):
Thanks for taking on board the feedback -- this looks great!

AlexWaygood:
> ```diff
> + static_frame/test/unit/test_type_clinic.py:197:39: error[invalid-argument-type] Argument is incorrect: Expected `_TypedDictSchema`, found `dict[str, <class 'int'> | <class 'float'> | <class 'str'>]`
> ```

It would be nice if we could special-case this kind of diagnostic to say something like "Expected a dictionary literal with string-literal keys and types as values", rather than exposing the internal `_TypedDictSchema` type here. But we can defer that; this seems fine for an initial implementation

crates/ty_python_semantic/src/types.rs
ibraheemdev: Opened https://github.com/astral-sh/ty/issues/1322.

ibraheemdev:
It looks like there's a panic with recursive `TypeDict` definitions in the typing conformance suite:
```py
RecursiveMovie = TypedDict(
    "RecursiveMovie", {"title": Required[str], "predecessor": NotRequired["RecursiveMovie"]}
)
```

carljm:
> It looks like there's a panic with recursive `TypeDict` definitions in the typing conformance suite

Shoot, I should have thought of this in advance. I think it means that we will have to make inference of the actual dict spec lazy, similar to how it is for class typed-dicts by default (since all our understanding of class bodies is lazy).

It may be that the best way to do this is similar to what I've done for TypeVar definitions in https://github.com/astral-sh/ruff/pull/20598, where we more fully special-case the entire assignment statement. This would probably eliminate the need for some of the machinery added in this PR around integrating with the full call-binding machinery.

PR #20733, opened 2025-10-07
[`pylint`] Implement `stop-iteration-return` (`PLR1708`)

fatelei:
## Summary

implement pylint rule stop-iteration-return / R1708

### DIFF
--- a/crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
+++ b/crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
@@ -49,32 +49,6 @@ impl Violation for StopIterationReturn {
     }
 }
 
-/// Check if a statement list contains any yield statements
-fn contains_yield_statement(body: &[ast::Stmt]) -> bool {
-    struct YieldFinder {
-        found: bool,
-    }
-
-    impl Visitor<'_> for YieldFinder {
-        fn visit_expr(&mut self, expr: &ast::Expr) {
-            if matches!(expr, ast::Expr::Yield(_) | ast::Expr::YieldFrom(_)) {
-                self.found = true;
-            } else {
-                walk_expr(self, expr);
-            }
-        }
-    }
-
-    let mut finder = YieldFinder { found: false };
-    for stmt in body {
-        walk_stmt(&mut finder, stmt);
-        if finder.found {
-            return true;
-        }
-    }
-    false
-}
-
 /// PLR1708
 pub(crate) fn stop_iteration_return(checker: &Checker, raise_stmt: &ast::StmtRaise) {
     // Fast-path: only continue if this is `raise StopIteration` (with or without args)
@@ -108,6 +82,32 @@ fn in_generator_context(checker: &Checker) -> bool {
             if contains_yield_statement(&function_def.body) {
                 return true;
             }
+        }
+    }
+    false
+}
+
+/// Check if a statement list contains any yield statements
+fn contains_yield_statement(body: &[ast::Stmt]) -> bool {
+    struct YieldFinder {
+        found: bool,
+    }
+
+    impl Visitor<'_> for YieldFinder {
+        fn visit_expr(&mut self, expr: &ast::Expr) {
+            if matches!(expr, ast::Expr::Yield(_) | ast::Expr::YieldFrom(_)) {
+                self.found = true;
+            } else {
+                walk_expr(self, expr);
+            }
+        }
+    }
+
+    let mut finder = YieldFinder { found: false };
+    for stmt in body {
+        walk_stmt(&mut finder, stmt);
+        if finder.found {
+            return true;
         }
     }
     false


crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::checkers::ast::Checker;
+
+/// ## What it does
+/// Checks for explicit `raise StopIteration` in generator functions.
+///
+/// ## Why is this bad?
+/// Raising `StopIteration` with a value in a generator function was deprecated
+/// in Python 3.5 and removed in later versions. This pattern breaks the
+/// abstraction between generators and iterators and can cause unexpected
+/// behavior.
```
ntBre: I tried this out on Python 3.7, which is the earliest version we support, and it raises a runtime error even then, so I think we should try to emphasize that a  bit more than just saying it was deprecated and will be removed. 

I also don't think it matters if the exception has a value, at least on 3.13:

```pycon
Python 3.13.7 (main, Aug 15 2025, 12:34:02) [GCC 15.2.1 20250813] on linux
>>> def my_generator():
...     yield 1
...     yield 2
...     raise StopIteration
...
>>> for _ in my_generator(): ...
...
Ellipsis
Ellipsis
  File "<python-input-0>", line 4, in my_generator
    raise StopIteration
StopIteration

The above exception was the direct cause of the following exception:

    for _ in my_generator(): ...
             ~~~~~~~~~~~~^^
RuntimeError: generator raised StopIteration
>>>
```

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,97 @@
...
+/// - [PEP 479](https://peps.python.org/pep-0479/)
+/// - [Python documentation](https://docs.python.org/3/library/exceptions.html#StopIteration)
+#[derive(ViolationMetadata)]
+pub(crate) struct StopIterationReturn;
+
+impl Violation for StopIterationReturn {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Explicit `raise StopIteration` in generator; use `return` instead".to_string()
+    }
```
ntBre, line 49: The `use return instead` part might be nice as the `fix_title`. I believe it will display even if the rule doesn't actually have a fix.


```suggestion
    fn message(&self) -> String {
        "Explicit `raise StopIteration` in generator".to_string()
    }
    
    fn fix_title(&self) -> Option<String> {
        "Use `return` instead".to_string()
    }
```

crates/ruff_linter/src/codes.rs
```diff
@@ -284,6 +284,7 @@ pub fn code_to_rule(linter: Linter, code: &str) -> Option<(RuleGroup, Rule)> {
         (Pylint, "R1702") => (RuleGroup::Preview, rules::pylint::rules::TooManyNestedBlocks),
         (Pylint, "R1704") => (RuleGroup::Stable, rules::pylint::rules::RedefinedArgumentFromLocal),
         (Pylint, "R1706") => (RuleGroup::Removed, rules::pylint::rules::AndOrTernary),
+        (Pylint, "R1708") => (RuleGroup::Stable, rules::pylint::rules::StopIterationReturn),
```
ntBre: New rules have to be added in `RuleGroup::Preview`.

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// PLR1708
+pub(crate) fn stop_iteration_return(checker: &Checker, raise_stmt: &ast::StmtRaise) {
+    // Check if we're in a generator function
+    if !checker.semantic().current_scope().kind.is_generator() {
+        return;
+    }
```
ntBre, line 68: I could be wrong, but I'm pretty sure this scope kind is for when you're inside of a generator expression, not inside the definition of a generator function. I'm not seeing any new snapshot files to confirm that, though. You'll need to run the tests locally and accept any new snapshots for the new tests you added to work.

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        // Check if it's calling StopIteration
+        if let ast::Expr::Name(ast::ExprName {
+            id,
+            ctx: _,
+            range: _,
+            node_index: _,
+        }) = func.as_ref()
+        {
+            if id == "StopIteration" {
```
ntBre: I think we'll need to use [`has_builtin_binding`](https://github.com/astral-sh/ruff/blob/d2e8d765705956694ac7bb23f4de179352b3c3ed/crates/ruff_python_semantic/src/model.rs#L273) here. That will make sure `StopIteration` hasn't been shadowed or anything like that.

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut in_generator_function = false;
+    for scope in checker.semantic().current_scopes() {
+        if let ruff_python_semantic::ScopeKind::Function(function_def) = scope.kind {
+            // Check if this function contains any yield statements by traversing its body
+            if contains_yield_statement(&function_def.body) {
+                in_generator_function = true;
+                break;
+            }
+        }
+    }
```
amyreese: Since this is likely the more "expensive" portion of the check (iterating over every statement in the parent function's body), I'd suggest extracting this logic into a function, and call that function from inside the matching branches below, so that we only check for a generator function context if the statement actually matches `raise StopIteration`.

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,135 @@
...
+            ctx: _,
+            range: _,
+            node_index: _,
+        }) = func.as_ref()
+        {
+            if id == "StopIteration" && checker.semantic().has_builtin_binding("StopIteration") {
+                // It's a StopIteration being raised with arguments
+                checker.report_diagnostic(StopIterationReturn, raise_stmt.range());
+            }
+        }
```
amyreese: I think this nested if block could actually be replaced with:

```rust
if checker.semantic().match_builtin_expr(&func, "StopIteration")
```

See the `blocking-input` rule for an example: https://github.com/astral-sh/ruff/blob/511710e1ef930050d6a58d73251ddf85558ecc93/crates/ruff_linter/src/rules/flake8_async/rules/blocking_input.rs#L46

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,135 @@
...
+        id,
+        ctx: _,
+        range: _,
+        node_index: _,
+    }) = exc.as_ref()
+    {
+        // Check if it's just `raise StopIteration` (without arguments)
+        if id == "StopIteration" && checker.semantic().has_builtin_binding("StopIteration") {
+            checker.report_diagnostic(StopIterationReturn, raise_stmt.range());
+        }
```
amyreese: Same for `match_builtin_expr`

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,135 @@
...
+        return;
+    };
+
+    // Check if it's a StopIteration exception (could be with or without a value)
+    if let ast::Expr::Call(ast::ExprCall {
+        func,
+        arguments: _,
+        range: _,
+        node_index: _,
+    }) = exc.as_ref()
```
amyreese: Since the two arms of this if-else share their internal logic, it might be possible to combine them into a match statement to reduce duplication and nesting. Something like:

```rust
match &raise_stmt.exc.as_ref() {
  Some(ast::Expr::Call(...) | ast::Expr::Name(...)) => {
    ...
  },
  _ => {}
}
```

crates/ty/docs/environment.md
ntBre, line 45: I think we should revert changes to this file, unless this is just a GitHub rendering bug.

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::checkers::ast::Checker;
+
+/// ## What it does
+/// Checks for explicit `raise StopIteration` in generator functions.
+///
+/// ## Why is this bad?
+/// Raising `StopIteration` in a generator function causes a `RuntimeError`
+/// at runtime (even in Python 3.7+, the earliest version we support). This
+/// breaks the abstraction between generators and iterators and will crash
+/// the program when the generator is iterated over.
```
ntBre: The first part here feels a bit redundant, and then I think the fact that it raises a runtime error is a bit more severe than breaking an abstraction. I think we could just say something like:


```suggestion
/// Raising `StopIteration` in a generator function causes a `RuntimeError`
/// when the generator is iterated over.
```

crates/ty/docs/environment.md
amyreese: I think this is github due to the branch being based on a two-week old rev. When I did a local rebase, it didn't show in the results.

crates/ruff_linter/resources/test/fixtures/pylint/stop_iteration_return.py
```diff
@@ -0,0 +1,129 @@
...
+    
+    yield from inner_gen()
+
+
+def generator_in_class():
+    class MyClass:
+        def generator_method(self):
+            yield 1
+            raise StopIteration("method")  # Should trigger
+
```
amyreese: There's also some extra indentation and whitespace in here. I suggest running `ruff format` on this fixture and then re-running the tests and accept the snapshot changes.

ntBre (approved):
Thank you! This looks good to me now if @amyreese is happy with it.

I pushed two commits fixing the merge conflicts from one of my PRs earlier and fixing a small nit about the function order.

amyreese (approved):

PR #20740, opened 2025-10-07
[ty] Make `infer_method_information` less confusing

sharkdp:
## Summary

`infer_method_information` was previously calling `ClassLiteral::to_class_type`, which uses the default-specialization of a generic class. This specialized `ClassType` was later only used if the class was non-generic, making the specialization irrelevant. The implementation was still a bit confusing, so this PR proposes a way to avoid turning the class literal into a `ClassType`.

### DIFF
--- a/crates/ty_python_semantic/src/types/signatures.rs
+++ b/crates/ty_python_semantic/src/types/signatures.rs
@@ -26,7 +26,7 @@ use crate::types::function::FunctionType;
 use crate::types::generics::{GenericContext, typing_self, walk_generic_context};
 use crate::types::infer::nearest_enclosing_class;
 use crate::types::{
-    ApplyTypeMappingVisitor, BindingContext, BoundTypeVarInstance, ClassType,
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -57,12 +58,22 @@ fn infer_method_information<'db>(
...
-
-    Some(MethodInformation { method, class })
+        .inner_type()
+    {
+        Type::ClassLiteral(class_literal) => {
+            Some((class_literal, class_literal.generic_context(db).is_some()))
+        }
+        Type::GenericAlias(alias) => Some((alias.origin(db), true)),
+        _ => None,
+    }?;
```
AlexWaygood: nit

```suggestion
    {
        Type::ClassLiteral(class_literal) => {
            (class_literal, class_literal.generic_context(db).is_some())
        }
        Type::GenericAlias(alias) => (alias.origin(db), true),
        _ => return None,
    };
```

carljm:
Thank you!

PR #20743, opened 2025-10-07
[ty] Fix tiny mistake in protocol tests

sharkdp:


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -964,11 +964,15 @@ class Foo(Protocol):
 
     def __init__(self) -> None:
         self.x = 42  # fine
+        self.y = "foo"  # fine
+
         self.a = 56  # TODO: should emit diagnostic
         self.b: int = 128  # TODO: should emit diagnostic
 
     def non_init_method(self) -> None:
-        self.y = 64  # fine
```
sharkdp, line 971: The `y` attribute is annotated as `y: str`, so this is not fine.

MichaReiser (approved):

crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -964,11 +964,15 @@ class Foo(Protocol):
 
     def __init__(self) -> None:
         self.x = 42  # fine
+        self.y = "foo"  # fine
+
```
AlexWaygood: It doesn't matter much, but I think the point here was to demonstrate that, just like other classes, it's okay for the sole assignment of the `y` attribute to be outside `__init__`. The only difference is that, unlike other classes, it does _have_ to have a declaration in the class body (implicit instance attributes in protocols are not allowed)

```suggestion
```

AlexWaygood (approved):

PR #20745, opened 2025-10-07
[ty] Improve diagnostics for bad `@overload` definitions

AlexWaygood:
## Summary

This PR attempts to address some of the confusion in https://github.com/astral-sh/ty/issues/1316 by improving our diagnostics around bad `@overload` definitions:
- If a series of `@overload`-decorated functions is not followed by a non-`@overload`-decorated function, the diagnostic message is improved. Specifically, the diagnostic message no longer uses the term "implementation function", which is a little jargon-y for users (and was misunderstood in <https://github.com/astral-sh/ty/issues/1316>). Instead, it now uses the term "non-`@overload`-decorated function"
- A new warning-level diagnostic is added that is triggered if you provide a nontrivial body for an `@overload`-decorated function. While there's nothing _wrong_ with this in principle, the body will be completely discarded at runtime, so it strongly suggests that the user has misunderstood the purpose of the `@overload` decorator. I think it will probably help users quite a bit to have this pointed out to them.

On this file:

```py
from typing import overload

@overload
def foo(x: int) -> int:
    return x

@overload
def foo(x: str) -> None:
    """Docstring"""
    pass
    print("oh no, a string")
```

We now emit these diagnostics:

image

## Test Plan

Mdtests and snapshots


AlexWaygood:
The two added diagnostics both look like true positives!

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Invalid_-_Overload_without_an_…_-_Regular_modules_(5c8e81664d1c7470).snap
```diff
@@ -17,43 +17,47 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/overloads.md
...
-error[invalid-overload]: Overloaded non-stub function `func` must have an implementation
+error[invalid-overload]: Overloads for function `func` must be followed by a non-`@overload`-decorated implementation function
  --> src/mdtest_snippet.py:7:5
   |
 5 | @overload
-6 | # error: [invalid-overload] "Overloaded non-stub function `func` must have an implementation"
+6 | # error: [invalid-overload] "Overloads for function `func` must be followed by a non-`@overload`-decorated implementation function"
 7 | def func(x: str) -> str: ...
-  |     ^^^^
+  |     ^^^^ Attempting to call `func` will raise `TypeError` at runtime
```
sharkdp: This is also part of the concise message, and I think it it's more confusing than helpful? We could add a "will raise `TypeError` at runtime" message to a lot of lints…

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Invalid_-_Overload_without_an_…_-_Regular_modules_(5c8e81664d1c7470).snap
```diff
@@ -17,43 +17,47 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/overloads.md
...
 8 |
 9 | class Foo:
   |
+info: Overloaded functions without implementations are only permitted in stub files, on protocols, or for abstract methods
+info: See https://docs.python.org/3/library/typing.html#typing.overload for more details
```
sharkdp: :+1:

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -958,6 +959,62 @@ declare_lint! {
...
+    /// a misunderstanding of how the `@overload` decorator works.
+    ///
+    /// ## Example
+    ///
+    /// ```py
+    /// from typing import overload
+    ///
+    /// @overload
+    /// def foo(x: int) -> int:
+    ///     return x + 1
```
sharkdp: Maybe
```suggestion
    ///     return x + 1  # will never be executed
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -958,6 +959,62 @@ declare_lint! {
...
+    ///
+    /// @overload
+    /// def foo(x: str) -> str:
+    ///     return "Oh no, got a string"
```
sharkdp:
```suggestion
    ///     return "Oh no, got a string"  # will never be executed
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2169,6 +2181,37 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    "Useless body for `@overload`-decorated function `{}`",
+                    &function.name
+                ));
+                diagnostic.set_primary_message("This statement will never be executed");
+                diagnostic.info(
+                    "`@overload`-decorated functions are solely for type checkers \
+                    and must be overridden at runtime by a non-`@overload`-decorated implementation",
+                );
+                diagnostic.help("Consider replacing this function body with `...` or `pass`");
+                break;
```
sharkdp, line 2212: There's probably an alternative implementation here where you would collect all remaining statements and use the full range for emitting the diagnostic, but I don't think it's worth the additional complexity.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2169,6 +2181,37 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    continue;
+                };
+                let mut diagnostic = builder.into_diagnostic(format_args!(
+                    "Useless body for `@overload`-decorated function `{}`",
+                    &function.name
+                ));
+                diagnostic.set_primary_message("This statement will never be executed");
+                diagnostic.info(
+                    "`@overload`-decorated functions are solely for type checkers \
+                    and must be overridden at runtime by a non-`@overload`-decorated implementation",
```
carljm: Nit: "overridden" doesn't feel like the right term here. That term is usually applied to subclass overrides, not to subsequent shadowing definitions. In general I don't think it's necessarily clear to the reader what "overridden at runtime" means.

I would probably just say something like "... are solely to provide a signature for type checkers; their bodies are unused"

crates/ty_python_semantic/resources/mdtest/snapshots/overloads.md_-_Overloads_-_Invalid_-_Overload_without_an_…_-_Regular_modules_(5c8e81664d1c7470).snap
AlexWaygood: > We could add a "will raise `TypeError` at runtime" message to a lot of lints…

heh, and I like to, when it's true! I think it's useful to be clear when something is a theoretical soundness complaint that might or might not cause issues, and when something will actually, definitely, 100% of the time cause a `TypeError` to be raised (assuming our inference is correct, of course). Subclassing a `@final` class is the former category; this is the latter category! There's no way the user could possibly want this behaviour

AlexWaygood: Good point about it being part of the concise message, though. That's too long for a supposedly single-line diagnostic.

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: Diagnostics with very long ranges can also be rendered oddly. And can interact in surprising ways with suppression comments.

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: switched to "overwritten", which I think is more precise, and avoids the confusion about "overidden" usually being applied to subclass overrides

PR #20746, opened 2025-10-07
[ty] Add an evaluation for completions

BurntSushi:
This is still early days, but I hope the framework introduced here makes
it very easy to add new truth data. Truth data should be seen as a form
of regression test for non-ideal ranking of completion suggestions.

I think it would help to read `crates/ty_completion_eval/README.md`
first to get an idea of what you're reviewing.


crates/ty_completion_eval/truth/type-var-typing-over-ast/main.py
```diff
@@ -0,0 +1,12 @@
...
+# even though there is also an `ast.TypeVar`. Getting this one
+# right seems tricky, and probably requires module-specific
+# heuristics.
+#
+# ref: https://github.com/astral-sh/ty/issues/1274#issuecomment-3345884227
+TypeVa<CURSOR: typing.TypeVar>
+
+# This is a similar case of `ctypes.cast` being preferred over
+# `typing.cast`. Maybe `typing` should just get a slightly higher
+# weight than most other stdlib modules?
```
sharkdp, line 11: A slightly more principled way of doing this might be to get [a list of standard library modules](https://github.com/python/typeshed/blob/main/stdlib/VERSIONS), and then do some statistics across the ecosystem to see which modules are often imported from. We'd probably not want to break alphabetical order completely, but if some modules are used a lot more (orders of magnitude more), this might be an argument for putting them in an extra section on top of the other completions, or similar?

For example, comparing `ast` to `typing` across all mypy_primer projects, we can see that the latter is imported 370x more often:
```
▶ rg 'from (typing|ast) import' --only-matching --no-filename | sort | uniq -c
     76 from ast import
  28146 from typing import
```

AlexWaygood: > For example, comparing `ast` to `typing` across all mypy_primer projects, we can see that the latter is imported 370x more often:

although, mypy_primer _is_ somewhat intentionally biased towards projects with type annotations, so that we can see how well type checkers do on reasonably well-typed code 😄

though there are a lot of exceptions to that rule, to ensure well-rounded coverage for various edge cases

sharkdp: Fair, but ty users will also work on sample of python projects that is biased towards more `typing` usage ;-)

AlexWaygood: well, maybe, but that's less true for autocompletion users than for our other users!

BurntSushi: I think using some kind of background frequency distribution to influence ranking is a plausibly good idea. At least worth a try. It's hard to know how often it will produce sub-optimal results though. We could perhaps limit it to _just_ standard library modules to keep it restricted in scope. (And reading @sharkdp's comments again, I think this is perhaps what is meant.)

.github/workflows/ci.yaml
```diff
@@ -707,6 +707,24 @@ jobs:
       - run: cargo binstall --no-confirm cargo-shear
       - run: cargo shear
 
+  ty-completion-evaluation:
+    name: "ty completion evaluation"
+    runs-on: depot-ubuntu-22.04-16
+    needs: determine_changes
+    if: ${{ needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main' }}
```
MichaReiser: Should we limit this task to only run on ty changes
```suggestion
    if: ${{ needs.determine_changes.outputs.ty == 'true' || github.ref == 'refs/heads/main' }}
```

crates/ty_completion_eval/src/main.rs
```diff
@@ -0,0 +1,588 @@
...
+         of the ruff repository",
+        truth = truth.as_std_path().display(),
+    );
+    // The temporary directory at which we copy our truth
+    // data to. We do this because we can't use the truth
+    // data as-is with its `<CURSOR>` annotations (and perhaps
+    // any other future annotations we add).
+    let tmpdir = std::env::var_os("TMPDIR")
+        .and_then(|path| SystemPathBuf::from_path_buf(PathBuf::from(path)).ok())
+        .unwrap_or_else(|| SystemPathBuf::from("/tmp"));
```
MichaReiser: Could we use the `tempdir` crate here, which works cross platform and cleans up once it's done

MichaReiser (approved):
This is great. I didn't review the CLI code in detail.

crates/ty_completion_eval/truth/higher-level-symbols-preferred/uv.lock
AlexWaygood, line 1: why is numpy listed in this uv.lock file? It doesn't seem to be listed as a dependency in `higher-level-symbols-preferred/pyproject.toml`.

Was this accidentally copied and pasted from the example that does include numpy as a dependency, by any chance? ;)

crates/ty_completion_eval/truth/internal-typeshed-hidden/main.py
```diff
@@ -0,0 +1,9 @@
+# This is a case where a symbol from an internal module appears
+# before the desired symbol from `typing`.
+#
+# We use a slightly different example than the one reported in
+# the issue to capture the deficiency via ranking. That is,
+# in ty#1274, the (current) top suggestion is the correct one.
```
AlexWaygood: GitHub is adding a weird link here to a non-existent webpage :/

crates/ty_completion_eval/truth/internal-typeshed-hidden/main.py
```diff
@@ -0,0 +1,9 @@
...
+#
+# ref: https://github.com/astral-sh/ty/issues/1274#issuecomment-3345923575
+AwaiGen<CURSOR: typing.AwaitableGenerator>
```
AlexWaygood: `typing.AwaitableGenerator` is a bit of a weird one because it doesn't actually exist at runtime -- and it's decorated with `@type_check_only`, so we should actually suppress it from autocompletions altogether (https://github.com/astral-sh/ty/issues/816).

`NoneType` might be quite a good one here? It exists in `_typeshed/__init__.pyi` and `types.pyi`, but (assuming you're using Python 3.10+), you'd never want it to be imported from the `_typeshed` module (which doesn't exist at runtime). You'd always want it imported from `types` (which does).

crates/ty_completion_eval/src/main.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// task printed to stdout.
+    ///
+    /// If the expected answer is found in the completion list, then
+    /// it is marked with an `*` along with its rank.
+    ShowOne(ShowOneCommand),
+}
+
+#[derive(Debug, clap::Parser)]
+struct AllCommand {
+    /// The mean recriporcal rank threshold that the evaluation must
```
AlexWaygood:
```suggestion
    /// The mean recriprocal rank threshold that the evaluation must
```

AlexWaygood (approved):
Very cool! I also haven't reviewed the CLI code in depth

crates/ty_completion_eval/src/main.rs
sharkdp: image

Almost, but not quite :smile:

AlexWaygood: Dang

crates/ty_completion_eval/src/main.rs
```diff
@@ -0,0 +1,588 @@
...
+                }
+            }
+            #[allow(clippy::cast_precision_loss)]
+            let mrr = precision_sum / (sources.len() as f64);
+            if let Some(ref mut wtr) = results_wtr {
+                wtr.flush()?;
+            }
+
+            let mut out = std::io::stdout().lock();
+            writeln!(out, "mean reciprocal rank: {mrr}")?;
```
sharkdp: The inner physicist in me always gets slightly nervous if statistical measures are presented to full numerical precision (*"mean reciprocal rank: 0.20409790112917506"*) :smile:. The actual precision of the MRR measure could be estimated by computing the standard deviation alongside the mean, but that obviously seems like overkill here. Given that individual reciprocal ranks can have values like 1/100 = 0.01, and given that we only have ~10 tasks, three or four decimal digits seem more than enough, so maybe
```suggestion
            writeln!(out, "mean reciprocal rank: {mrr:.4}")?;
```
(Due to the large spread in reciprocal ranks and the small number of tasks, the actual standard deviation is around 0.34 at the moment, higher than the MRR. That would mean that even one digit is enough)

crates/ty_completion_eval/src/main.rs
```diff
@@ -0,0 +1,588 @@
...
+                            &task.name,
+                            &task.cursor.path.as_str(),
+                            task.cursor.index,
+                            rank,
+                        ))?;
+                    }
+                }
+            }
+            #[allow(clippy::cast_precision_loss)]
+            let mrr = precision_sum / (sources.len() as f64);
```
sharkdp: This seems wrong to me? Shouldn't we be normalizing by the total number of tasks, instead of the number of source files? If I split a file with multiple tasks across several files, the MRR should stay the same?

crates/ty_completion_eval/src/main.rs
```diff
@@ -0,0 +1,588 @@
...
+                wtr.flush()?;
+            }
+
+            let mut out = std::io::stdout().lock();
+            writeln!(out, "mean reciprocal rank: {mrr}")?;
+            if mrr < threshold {
+                writeln!(out, "MRR does not exceed threshold of {threshold}")?;
+                Ok(ExitCode::FAILURE)
+            } else {
+                writeln!(out, "MRR exceeds threshold of {threshold}")?;
```
sharkdp: If someone is not familiar with the MRR measure, this message could read like a failure, so maybe:
```suggestion
                writeln!(out, "Success: MRR exceeds threshold of {threshold}")?;
```
(and similarly, *"Failure: MRR does not exceed threshold …"* above)?

sharkdp (approved):
Very nice! Since Micha and Alex didn't review the CLI tool, I did the opposite, and only reviewed the actual MRR computation.

crates/ty_completion_eval/src/main.rs
BurntSushi: I did it this way intentionally to avoid clean-up so that `uv sync` doesn't have to do as much work on repeated runs. It's also easier to debug when something goes wrong. But maybe that is a premature optimization.

I've updated this to use `tempfile` and added a flag that lets you opt into keeping the temporary directory around after the eval is done.

crates/ty_completion_eval/truth/higher-level-symbols-preferred/uv.lock
BurntSushi: Oh nice catch. And yes, it was a copy-and-paste error!

But actually, this also revealed that I didn't have `uv.lock` files in each of the project directories and we probably should. So I fixed that as well.

crates/ty_completion_eval/truth/internal-typeshed-hidden/main.py
BurntSushi: Where do you see the link rendered? Anyway, I tried to fix this.

crates/ty_completion_eval/truth/internal-typeshed-hidden/main.py
BurntSushi: Yes, perfect! Thank you for catching this.

crates/ty_completion_eval/src/main.rs
BurntSushi: Thanks! I'll do `.4` for now. And yeah, definitely plan to add more tasks. Hopefully lots more.

The thing driving the higher standard deviation here is that completion requests (aside from auto-import) don't take what has been typed into account. Instead, that work is done by the client. I haven't yet decided whether non-auto-import completions should do this on the server side (in which case, the eval as it exists is correct), or if we should continue relying on the client (in which case, the eval should mimic the client).

crates/ty_completion_eval/truth/internal-typeshed-hidden/main.py
AlexWaygood: Right here, on the PR view (and I think it would also be rendered if I was just browsing the file on GitHub):

---

image

---

if I click on that link, it takes me to a 404 page

crates/ty_completion_eval/src/main.rs
BurntSushi: Also, we have 16 evaluation tasks right now. Each truth directory can contain 1 or more tasks.

crates/ty_completion_eval/src/main.rs
BurntSushi: Nice catch. My original eval had one task per source. But I refactored to permit each source project to be able to have multiple tasks. And of course, forgot to fix this.

crates/ty_completion_eval/src/main.rs
BurntSushi: Yup that's fair. I also added "minimum threshold" to make it clear that we want to be above the set threshold.

PR #20749, opened 2025-10-07
[ty] Infer better specializations of unions with `None` (etc)

dcreager:
This PR adds a specialization inference special case that lets us handle the following examples better:

```py
def f[T](t: T | None) -> T: ...
def g[T](t: T | int | None) -> T | int: ...

def _(x: str | None):
    reveal_type(f(x))  # revealed: str (previously str | None)

def _(y: str | int | None):
    reveal_type(g(x))  # revealed: str | int (previously str | int | None)
```

We already have a special case for when the formal is a union where one element is a typevar, but it maps the entire actual type to the typevar (as you can see in the "previously" results above).

The new special case kicks in when the actual is also a union. Now, we filter out any actual union elements that are already subtypes of the formal, and only bind whatever types remain to the typevar. (The `| None` pattern appears quite often in the ecosystem results, but it's more general and works with any number of non-typevar union elements.)

The new constraint solver should handle this case as well, but it's worth adding this heuristic now with the old solver because it eliminates some false positives from the ecosystem report, and makes the ecosystem report less noisy on the other constraint solver PRs.

### DIFF
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -17,7 +17,7 @@ use crate::types::{
     ApplyTypeMappingVisitor, BoundTypeVarInstance, ClassLiteral, FindLegacyTypeVarsVisitor,
     HasRelationToVisitor, IsEquivalentVisitor, KnownClass, KnownInstanceType, MaterializationKind,
     NormalizedVisitor, Type, TypeMapping, TypeRelation, TypeVarBoundOrConstraints, TypeVarInstance,
-    TypeVarKind, TypeVarVariance, UnionType, binding_type, declaration_type,
+    TypeVarKind, TypeVarVariance, UnionBuilder, UnionType, binding_type, declaration_type,
 };
 use crate::{Db, FxOrderMap, FxOrderSet};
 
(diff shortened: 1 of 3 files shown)


dcreager:
Lots of false positives removed in the ecosystem report. The only additions involve splatting a `**kwargs` argument, which is covered by https://github.com/astral-sh/ty/issues/1248.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1138,25 +1138,56 @@ impl<'db> SpecializationBuilder<'db> {
...
+                        builder.add(element)
                     });
-                let bound_typevar = bound_typevars.next();
-                let additional_bound_typevars = bound_typevars.next();
-                if let (Some(bound_typevar), None) = (bound_typevar, additional_bound_typevars) {
+                let Some(remaining_actual) = actual_non_subtypes.try_build() else {
+                    return Ok(());
+                };
+                self.add_type_mapping(formal_bound_typevar, remaining_actual);
+            }
```
AlexWaygood: slightly simpler might be to do this? it would also mean you wouldn't need the `UnionBuilder` import up at the top

```diff
diff --git a/crates/ty_python_semantic/src/types/generics.rs b/crates/ty_python_semantic/src/types/generics.rs
index 14388f0518..e53bd41333 100644
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -1170,14 +1170,11 @@ impl<'db> SpecializationBuilder<'db> {
                 if (actual_union.elements(self.db).iter()).any(|ty| ty.is_type_var()) {
                     return Ok(());
                 }
-                let actual_non_subtypes = (actual_union.elements(self.db).iter().copied())
-                    .filter(|ty| !ty.is_subtype_of(self.db, formal))
-                    .fold(UnionBuilder::new(self.db), |builder, element| {
-                        builder.add(element)
-                    });
-                let Some(remaining_actual) = actual_non_subtypes.try_build() else {
+                let remaining_actual =
+                    actual_union.filter(self.db, |ty| !ty.is_subtype_of(self.db, formal));
+                if remaining_actual.is_never() {
                     return Ok(());
-                };
+                }
                 self.add_type_mapping(formal_bound_typevar, remaining_actual);
             }
```

AlexWaygood (approved):
Nice!

crates/ty_python_semantic/src/types/generics.rs
dcreager: Yeah I had thought about that, but was feeling mildly icky about using `Never` in that way. I guess for unions it's okay, since that really is equivalent to "no elements in the union".

AlexWaygood: I don't have a strong opinion FWIW, but your current spelling _did_ make me do a "Huh, don't we have a `UnionType::filter()` method yet... oh! We do" 😄

Issue #20747, opened 2025-10-07
Fix several rules to handle deferred annotations


ntBre:
Several rules exhibit very different behavior in the presence of deferred annotations, whether from an explicit `from __future__ import annotations` or on Python 3.14. These include:

- FAST001
- FAST003
- A003
- PTH210
- RUF001

You can see the effect of switching the Python version from 3.13 to 3.14 in this [commit](https://github.com/astral-sh/ruff/pull/20725/commits/095acaff3356c9941262b72d4e911514d6dd28bb) from #20725.

In addition to the tests modified in the commit above, each of these rules also has a new `assert_diagnostics_diff` test:

- [FAST001 and FAST003](https://github.com/astral-sh/ruff/blob/a658f4181d4e3ee85815e9f84c88d77c51160186/crates/ruff_linter/src/rules/fastapi/mod.rs#L34)
- [A003](https://github.com/astral-sh/ruff/blob/a658f4181d4e3ee85815e9f84c88d77c51160186/crates/ruff_linter/src/rules/flake8_builtins/mod.rs#L68)
- [PTH210](https://github.com/astral-sh/ruff/blob/a658f4181d4e3ee85815e9f84c88d77c51160186/crates/ruff_linter/src/rules/flake8_use_pathlib/mod.rs#L87)
- [RUF001](https://github.com/astral-sh/ruff/blob/a658f4181d4e3ee85815e9f84c88d77c51160186/crates/ruff_linter/src/rules/ruff/mod.rs#L249)

To fix this issue, we should remove all of these diffs.

Micha and Dylan observed that `semantic.resolve_name` seems to be returning the wrong results now (https://github.com/astral-sh/ruff/pull/20725#issuecomment-3376143411), and I think they suggested possibly making these checks deferred so that we can resolve these lookups differently.

Unless the changes end up being quite small, I think it could make sense to tackle these with one PR per rule.

11happy:
Hello @ntBre , I am interested to work on this , Can I work on `A003` ? 
Thank you

MichaReiser:
We have to check with @dylwil3, as he's currenlty working on this

dylwil3:
The more the merrier - feel free to work on `A003` @11happy !

11happy:
While Debugging for `A003` I noticed that for this case
```
def str(self):
        pass

    def method_usage(self) -> str:
        pass
```
in python 3.14 zero references were there you see can the debug prints below
```
binding: str
[...]

PR #20750, opened 2025-10-07
[`fastapi`] Handle deferred annotations correctly for `FAST001` and `FAST003`

dylwil3:
Progress towards #20747


crates/ruff_linter/src/checkers/ast/analyze/deferred_scopes.rs
```diff
@@ -264,6 +266,12 @@ pub(crate) fn deferred_scopes(checker: &Checker) {
             ]) {
                 pep8_naming::rules::invalid_first_argument_name(checker, scope);
             }
+            if checker.is_rule_enabled(Rule::FastApiRedundantResponseModel) {
```
MichaReiser, line 269: Does this change how the rule handles non-deferred type annotations?

dylwil3: Unfortunately yes, it seems. For example if you shadow the class later in the file. Back to the drawing board...

(Incidentally, while investigating this I discovered a different issue with the stable behavior https://github.com/astral-sh/ruff/issues/20945)

dylwil3:
Gonna close this as stale and others can pick it up - it turned out to be fairly confusing and I haven't had time to revisit it.

PR #20756, opened 2025-10-07
[ty] Disambiguate classes that live in different modules but have the same fully qualified names

AlexWaygood:
## Summary

Even disambiguating classes using their fully qualified names is not enough for some diagnostics. We've seen real-world examples in the ecosystem (and https://github.com/astral-sh/ruff/pull/20368 introduces some more!) where two types can be different, but can still have the same fully qualified name. In these cases, our disambiguation machinery needs to print the file path and line number of the class in order to disambiguate classes with similar names in our diagnostics.

Helps with https://github.com/astral-sh/ty/issues/1306

## Test Plan

Mdtests


crates/ty_python_semantic/src/types/display.rs
```diff
@@ -222,13 +266,13 @@ struct ClassDisplay<'db> {
...
-        let module_ast = parsed_module(self.db, file).load(self.db);
-        let index = semantic_index(self.db, file);
-        let file_scope_id = body_scope.file_scope_id(self.db);
+impl<'db> ClassLiteral<'db> {
+    fn class_parents(self, db: &'db dyn Db) -> Vec<String> {
+        let body_scope = self.body_scope(db);
+        let file = body_scope.file(db);
+        let module_ast = parsed_module(db, file).load(db);
+        let index = semantic_index(db, file);
+        let file_scope_id = body_scope.file_scope_id(db);
```
MichaReiser, line 290: Why can't we use `self.db` here anymore?

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -266,17 +310,35 @@ impl ClassDisplay<'_> {
...
+            let path = match path {
+                // Normalize path separator to `/` for system paths, for easier testing on Windows.
+                // `/` is a valid path separator on Windows, it's just not the default separator,
+                // so this seems fine.
+                FilePath::System(path) => path
+                    .strip_prefix(self.db.system().current_directory())
+                    .unwrap_or(path)
+                    .as_std_path()
+                    .to_slash()
+                    .expect("Path should be valid UTF-8"),
```
MichaReiser: I don't think we should do this. We intentionally use platform specific path representations anywhere where we display a path. I don't see why we should deviate from that here and it results in an inconsistent experience (I don't think easier testing is a good enough reason)

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -92,10 +114,32 @@ impl<'db> AmbiguousClassCollector<'db> {
...
-                if let AmbiguityState::Unambiguous(existing) = value
-                    && *existing != class
-                {
-                    *value = AmbiguityState::Ambiguous;
+                match value {
+                    AmbiguityState::Unambiguous(existing) => {
+                        if *existing != class {
+                            let class_parents = class.class_parents(db);
+                            if existing.class_parents(db) == class_parents {
+                                *value = AmbiguityState::RequiresFileAndLineNumber;
```
MichaReiser: I'm not sure I fully understand this code but what if you have something like this

```
class B:
	...

class A(B):
	...
```

both are defined in a pyi and in the py file and there's a diagnostic involving `A`. Would it mean that we don't show the file and line number for `A` because it has different parents?

MichaReiser (commented):
Lovely :)

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: Because it's now a method on `ClassLiteral` (it used to be a method on `ClassDisplay`), and `ClassLiteral` doesn't have a `self.db` field. It's now a method on `ClassLiteral` rather than `ClassDisplay` because it makes it easier to reuse the method from other places in this module.

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: Yeah, fair enough. I guess I have to go write some complicated regexes to get mdtest to work with these paths that are rendered differently on differnt platforms. So be it :)

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: "class parents" is possibly a slightly confusing term here. (I didn't add or name that method, in my defence ;)

For a class `A` in a module `foo.bar`, the "parents" of `A` will be `["foo", "bar"]`. For a class `B` defined in the scope of a method `m` of a class `D` in a module `foo.bar`, the "parents" of `B` will be `["foo", "bar", "D", "<locals of function 'm'>"]`.

On `main`, we currently only disambiguate classes in our diagnostics if there are two classes with the same unqualified name appearing in the same diagnostic (e.g. "expected `A`, got `A`" is a terrible diagnostic message -- "expected `foo.bar.A`, got `baz.A`" is much better). This PR goes a step further by also comparing the _fully_ qualified names of the classes: if using the fully qualified names would also not be sufficient to disambiguate the classes (because they have the same fully qualified names, as well as the same unqualified names!), we also display the file and line number when displaying the class.

AlexWaygood: I renamed the method and added some docs

PR #20759, opened 2025-10-08
[ty] Use 3.14 as the default version

sharkdp:
## Summary

Bump the latest supported Python version of ty to 3.14 and updates some references from 3.13 to 3.14.

This also fixes a bug with `dataclasses.field` on 3.14 (which adds a new keyword-only parameter to that function, breaking our previously naive matching on the parameter structure of that function).

## Test Plan

A `ty check` on a file with template strings (without any further configuration) doesn't raise errors anymore.

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -1798,12 +1798,11 @@ Meta.<CURSOR>
             // whether we're in release mode or not. These differences
             // aren't really relevant for completion tests AFAIK, so
             // just redact them. ---AG
-            filters => [(r"(?m)\s*__(annotations|new)__.+$", "")]},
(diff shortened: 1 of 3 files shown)


MichaReiser (approved):

sharkdp:
Looks like we're breaking some dataclass tests in the conformance suite. Looking into it.

sharkdp:
```diff
+qualifiers_annotated.py:36:42: error[invalid-syntax] named expression cannot be used within a type annotation
+qualifiers_annotated.py:40:28: error[invalid-syntax] await expression cannot be used within a type annotation
```

These are syntax errors now. I'll open a PR upstream (Edit: https://github.com/python/typing/pull/2093).

PR #20764, opened 2025-10-08
[ty] Prefer declared base class attribute over inferred attribute on subclass

sharkdp:
## Summary

When accessing an (instance) attribute on a given class, we were previously traversing its MRO, and building a union of types (if the attribute was available on multiple classes in the MRO) until we found a *definitely bound* symbol. The idea was that possibly unbound symbols in a subclass might only partially shadow the underlying base class attribute.

This behavior was problematic for two reasons:
* if the attribute was definitely bound on a class (e.g. `self.x = None`), we would have stopped iterating, even if there might be a `x: str | None` declaration in a base class (the bug reported in https://github.com/astral-sh/ty/issues/1067).
* if the attribute originated from an implicit instance attribute assignment (e.g. `self.x = 1` in method `Sub.foo`), we might stop looking and miss another implicit instance attribute assignment in a base class method (e.g. `self.x = 2` in method `Base.bar`).

With this fix, we still iterate the MRO of the class, but we only stop iterating if we find a *definitely declared* symbol. In this case, we only return the declared attribute type. Otherwise, we keep building a union of inferred attribute types.

The implementation here seemed to be the easiest fix for https://github.com/astral-sh/ty/issues/1067 that also kept the ecosystem impact low (the changes that I see all look correct). However, as the Markdown tests show, there are other things to fix in this area. For example, we should do a similar thing for *class attributes*. This is more involved, though (affects many different areas and probably involves a change to our descriptor protocol implementation), so I'd like to postpone this to a follow-up.

closes https://github.com/astral-sh/ty/issues/1067

## Test Plan

Updated Markdown tests, including a regression test for https://github.com/astral-sh/ty/issues/1067.

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -2891,7 +2891,7 @@ impl<'db> ClassLiteral<'db> {
                         if boundness == Boundness::Bound {
                             if is_declared {
                                 // We found a definitely-declared attribute. Discard possibly collected
-                                // inferred types from superclasses and return the declared type.
+                                // inferred types from subclasses and return the declared type.
                                 return member;
                             }
 


crates/ty_python_semantic/src/place.rs
```diff
@@ -232,13 +233,9 @@ pub(crate) fn place<'db>(
...
 
-/// Infer the public type of a class symbol (its type as seen from outside its scope) in the given
+/// Infer the public type of a class member/symbol (its type as seen from outside its scope) in the given
 /// `scope`.
-pub(crate) fn class_symbol<'db>(
-    db: &'db dyn Db,
-    scope: ScopeId<'db>,
-    name: &str,
-) -> PlaceAndQualifiers<'db> {
+pub(crate) fn class_member<'db>(db: &'db dyn Db, scope: ScopeId<'db>, name: &str) -> Member<'db> {
```
sharkdp, line 238: We could think about moving this to the newly created `member` module...

crates/ty_python_semantic/src/member.rs
```diff
@@ -0,0 +1,66 @@
+use crate::{
+    place::{Place, PlaceAndQualifiers},
+    types::Type,
+};
+
+/// The return type of certain member-lookup operations. Contains information
+/// about the type, type qualifiers, boundness/declaredness, and additional
+/// metadata (e.g. whether or not the member was declared)
+#[derive(Debug, Clone, PartialEq, Eq, salsa::Update, get_size2::GetSize)]
+pub(crate) struct Member<'db> {
```
sharkdp, line 10: I don't *love* the fact that we now have this `Member(PlaceAndQualifiers(Place(Type)))` onion structure, but each single layer has its own dedicated purpose and is used in a lot of places.

I considered adding `is_declared` to `PlaceAndQualifiers` somehow, as this flag might be relevant in other areas as well. This would be a huge refactoring, but more importantly, I think that we probably want to add new flags to `Member` soon (or use `bitflags`) in order to convey even more metadata (e.g. is this an implicit instance attribute? is the attribute a data/non-data descriptor?).

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2874,27 +2879,32 @@ impl<'db> ClassLiteral<'db> {
...
+                        is_declared,
                     } = class.own_instance_member(db, name)
                     {
-                        // TODO: We could raise a diagnostic here if there are conflicting type qualifiers
-                        union_qualifiers |= qualifiers;
-
                         if boundness == Boundness::Bound {
-                            if union.is_empty() {
-                                // Short-circuit, no need to allocate inside the union builder
+                            if is_declared {
```
sharkdp, line 2892: This (and the few lines below) is the key change. Everything else is just refactoring (and recording whether or not a particular `Place` is declared or inferred).

crates/ty_python_semantic/src/member.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl Default for Member<'_> {
+    fn default() -> Self {
+        Member::undeclared(PlaceAndQualifiers::default())
+    }
+}
+
+impl<'db> Member<'db> {
+    pub(crate) fn undeclared(inner: PlaceAndQualifiers<'db>) -> Self {
```
sharkdp: Maybe `non_declared` or `inferred` is clearer?

AlexWaygood:
> ```diff
> pip (https://github.com/pypa/pip)
> + src/pip/_internal/cli/index_command.py:133:9: error[invalid-assignment] Object of type `Any` is not assignable to attribute `prompting` on type `Unknown | MultiDomainBasicAuth | None`
> + src/pip/_internal/cli/index_command.py:134:9: error[invalid-assignment] Object of type `Any` is not assignable to attribute `keyring_provider` on type `Unknown | MultiDomainBasicAuth | None`
> - Found 460 diagnostics
> + Found 462 diagnostics
> ```

That's a weird diagnostic. Shouldn't `Any` be assignable to everything?

sharkdp:
> That's a weird diagnostic. Shouldn't `Any` be assignable to everything?

I stumbled over the exact same thing and assumed it was just me. But we should definitely improve the diagnostic message here\*. The problem is not the type of the object that we're trying to assign. The problem is in the type of the object that we're acessing the attribute *on* (`None` doesn't have a `prompting` attribute).

> Object of type `Any` is not assignable to **attribute `prompting` on type `Unknown | MultiDomainBasicAuth | None`**

\* ideally by adding a subdiagnostic that explains the underlying reason why the assignment to the attribute *on that union type* has failed (because the assignment to that attribute on one particular element failed).

AlexWaygood:
Ahhh, I see. Thanks!

In that case, the type of the object we're trying to assign to the `prompting` attribute seems irrelevant? Shouldn't we just say something like this in the diagnostic?

```
Attribute `prompting` may not be settable on object of type `Unknown | MultiDomainBasicAuth | None`
```

sharkdp:
> Shouldn't we just say something like this in the diagnostic

That would be one way to improve the diagnostic, yes. The other thing that I proposed would also split up that union type and add to yours: "… because it can not be set on union element `None`", or similar.

AlexWaygood:
yes -- both sound good ;) but not for this PR, of course.

crates/ty_python_semantic/src/member.rs
sharkdp: Changed to `inferred`, also added some doc comments.

crates/ty_python_semantic/src/member.rs
carljm: This feels like it intersects with discussions we've had a long time ago already (and several TODO comments we have) about how the `Boundness` part of `Place` currently conflates both "boundness" and "declaredness" in a way that isn't always consistent. I think we always vaguely had the intention of cleaning that up, but it never became a priority. it seems like this new flag is providing the same metadata that would be provided by that refactor, but at a new outer layer, mostly in order to avoid the need for a big refactor of `Place`?

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2874,27 +2879,32 @@ impl<'db> ClassLiteral<'db> {
...
+                                // We found a definitely-declared attribute. Discard possibly collected
+                                // inferred types from superclasses and return the declared type.
```
carljm: Don't we walk _up_ the MRO? So how could it be that we find a definitely-declared attribute and have to discard possibly collected inferred types from _superclasses_? Should this say _subclasses_ instead, or am I missing something?

sharkdp: > Should this say _subclasses_ instead

Yes, absolutely.

crates/ty_python_semantic/src/member.rs
sharkdp: I opened https://github.com/astral-sh/ty/issues/1341 to track this.

PR #20765, opened 2025-10-08
Update `lint.flake8-type-checking.quoted-annotations` docs

CarrotManMatt:
Update `lint.flake8-type-checking.quoted-annotations` docs to mention Python 3.14.


crates/ruff_workspace/src/options.rs
```diff
@@ -2167,7 +2167,8 @@ pub struct Flake8TypeCheckingOptions {
     /// ```
     ///
     /// Note that this setting has no effect when `from __future__ import annotations`
-    /// is present, as `__future__` annotations are always treated equivalently
+    /// is present or when minimum supported Python version >=3.14, as `__future__`
+    /// annotations and PEP 649/PEP 749 annotations are always treated equivalently
```
ntBre: I think we might want to leave the original text and append something like

> Similarly, this setting has no effect on Python versions after 3.14 because  these annotations are also deferred.

It doesn't have a practical difference in most cases, but the 3.14 semantics are a bit different from quoted annotations, so it seems better not to group them together completely. 

My understanding based on PEP 649 is that `from __future__ import annotations` literally turns annotations into strings (requiring them to be `eval`ed or similar if you need to evaluate them), whereas 3.14 leaves them as normal expressions.

CarrotManMatt: Thanks for the review!💚

Yep, agreed that keeping the wording separate reflects the difference in outcomes from using PEP 649 annotations vs `from __future__ import annotations`.
PEP 649 annotations aren't exactly left as normal expressions, they're held unevaluated internally until a view of them is explicitly requested with one of the `annotationlib.Format`s (one of which is complete evaluation to normal expressions). But that's unnecessary nit-picking that doesn't need to be mentioned in the docs.

In your opinion, is it worth calling the 3.14+ annotations `PEP 649 annotations` or `PEP 649/PEP 749 annotations`? (As PEP 749 was just the confirmation for how to internally implement the annotation structure defined in PEP 649.)

CarrotManMatt:
I'd appreciate if you could perform the docs generation for me after I make the requested change.

MichaReiser:
Thank you

PR #20767, opened 2025-10-08
[ty] Add tests for instance attributes in class hierarchies

sharkdp:
## Summary

This adds a couple of new test cases related to https://github.com/astral-sh/ty/issues/1067 and beyond that. For now, they are just documenting the current (problematic) behavior. Since the topic has some subtleties, I'd like to merge this prior to the actual bugfix(es) in order to evaluate the changes in an easier way.


sharkdp:
I guess we can discuss the desired behavior in the upcoming PR(s).

PR #20768, opened 2025-10-08
Remove parentheses around multiple exception types on Python 3.14+

ntBre:
Summary
--

This PR implements the black preview style from https://github.com/psf/black/pull/4720. As of Python 3.14, you're allowed to omit the parentheses around groups of exceptions, as long as there's no `as` binding:

**3.13**

```pycon
Python 3.13.4 (main, Jun  4 2025, 17:37:06) [Clang 20.1.4 ] on linux
>>> try: ...
... except (Exception, BaseException): ...
...
Ellipsis
>>> try: ...
... except Exception, BaseException: ...
...
  File "<python-input-1>", line 2
    except Exception, BaseException: ...
           ^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized
```

**3.14**

```pycon
Python 3.14.0rc2 (main, Sep  2 2025, 14:20:56) [Clang 20.1.4 ] on linux
>>> try: ...
... except Exception, BaseException: ...
...
Ellipsis
>>> try: ...
... except (Exception, BaseException): ...
...
Ellipsis
>>> try: ...
... except Exception, BaseException as e: ...
...
  File "<python-input-2>", line 2
    except Exception, BaseException as e: ...
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized when using 'as'
```

I think this ended up being pretty straightforward, at least once Micha showed me where to start :)

Test Plan
--

New tests

At first I thought we were deviating from black in how we handle comments within the exception type tuple, but I think this applies to how we format all tuples, not specifically with the new preview style.

### DIFF
--- a/crates/ruff_python_formatter/src/preview.rs
+++ b/crates/ruff_python_formatter/src/preview.rs
@@ -28,7 +28,9 @@ pub(crate) const fn is_blank_line_before_decorated_class_in_stub_enabled(
     context.is_preview()
 }
 
-/// Returns `true` if the [`remove_parens_around_except_types`](TODO) preview style is enabled.
+/// Returns `true` if the
+/// [`remove_parens_around_except_types`](https://github.com/astral-sh/ruff/pull/20768) preview
+/// style is enabled.
 pub(crate) const fn is_remove_parens_around_except_types_enabled(
     context: &PyFormatContext,
 ) -> bool {


MichaReiser (approved):
Looks great. 

I just realized that we haven't updated our black tests in a while. It's something I normally do before working on preview styles because we then "inherit" the black tests

Would you mind updating the black tests in a separate PR that we can merge first to get more test coverage?

Updating is normally very smooth but you might run into problems if Black changed their test structure. All that's needed is to run 

https://github.com/astral-sh/ruff/blob/a9efdea113a4f676ef85d89ed2dc285f7087478d/crates/ruff_python_formatter/resources/test/fixtures/import_black_tests.py

I believe it takes a path to black's checkout as an argument

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__remove_except_types_parens.py.snap
MichaReiser, line 1: Nice. The most satisfying feeling ever when working on the formatter :)

PR #20771, opened 2025-10-08
[ty] infer `Self` return type annotation for class methods

Glyphack:
## Summary

Infer annotated `Self` return type correctly for class methods.

Example:

```py
from typing import Type, Self

class C:
    @classmethod
    def make_instance(cls: Type[Self]) -> Self:
        return cls()

reveal_type(C.make_instance())  # Should be C but is Unknown
reveal_type(C.make_instance)  # revealed: bound method <class 'C'>.make_instance() -> C
```

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/cls.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/cls.md
@@ -1,0 +1,20 @@
+# `cls`
+
+```toml
+[environment]
+python-version = "3.13"
+```
+
+## Methods
+
+```py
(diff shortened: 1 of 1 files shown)


Glyphack:
I thought this was a separate issue than not supporting `type[Self]` but after playing around with it I realized that's the root cause.
I think this playground link shows how:
https://play.ty.dev/55f8e04b-148c-4ea0-8916-e302591678bb

Maybe this PR I can only add the generic context when encountering `type[Self]`

Glyphack:
This is fixed already.

PR #20772, opened 2025-10-08
implement unused variable export and class member (fixes #20766)

### LINKED ISSUE #20766: Rules for Undocumented Public Variable / Constant and Class Member (attribute
### Summary

[PEP 257 – Docstring Conventions (attribute docstrings)](https://peps.python.org/pep-0257/) defines a syntax for variable/member docstrings. The syntax is controversial, but it is supported by tools (e.g. VSCode).

Currently there is no lint to enforce global variables / constants to either be made private (underscore) or be documented.

So I'm imagining two new rules, very similar to the existing pydocstyle rules. The first one I think would be great, the second one is a bit difficult due to the various other ways that exist to document class members.

# 1. undocumented-public-module-variable
 ## What it does
 Checks for undocumented public module-level variable definitions.

 ## Why is this bad?
 Public module-level variables should be documented via docstrings to
 describe their purpose and expected values.

 According to PEP 257, docstrings for module-level variables should be
 placed immediately after the variable assignment.

 If the codebase adheres to a standard format for module-level variable
 docstrings, follow that format for consistency.

 ## Example
 ```python
 DEFAULT_TIMEOUT: int = 30
 MAX_CONNECTIONS: int = 100
 ```

 Use instead:
 ```python
 DEFAULT_TIMEOUT: int = 30
 """The default timeout duration in seconds."""

 MAX_CONNECTIONS: int = 100
 """The maximum number of concurrent connections allowed."""
 ```

# 2. undocumented-public-class-attribute
 ## What it does
 Checks for undocumented public class attribute definitions.

 ## Why is this bad?
 Public class attributes should be documented via docstrings to describe
 their purpose and expected values.

 According to PEP 257, docstrings for class attributes should be placed
 immediately after the attribute assignment.

 If the codebase adheres to a standard format for attribute docstrings,
 follow that format for consistency.

 ## Example
 ```python
 class Config:
     timeout: int = 30
     max_retries: int = 3
 ```

 Use instead:
 ```python
 class Config:
     timeout: int = 30
     """The timeout duration in seconds."""

     max_retries: int = 3
     """The maximum number of retry attempts."""
 ```

phiresky:
## Summary

This PR implements two new rules as described in #20766.

Main questions are:
1. where to put visibility check code and __all__ parsing
2. where to put rules (new group or the existing group where it very clearly would fit in)

Example:

```python
# Should trigger D1002 - undocumented public module variable (annotated)
public_var: int = 1

# Should NOT trigger - has docstring
documented_var: str = "hello"
"""This variable is documented."""

# Should NOT trigger - private variable
_private_var: int = 2

class MyClass:
    # Should trigger D1001 - undocumented public class attribute (annotated)
    class_var: int = 10

    # Should NOT trigger - has docstring
    documented_class_var: str = "test"
    """This class variable is documented."""
```

## Test Plan

Tested with snapshots.

### DIFF
--- a/crates/ruff_python_semantic/src/binding.rs
+++ b/crates/ruff_python_semantic/src/binding.rs
@@ -304,7 +304,7 @@ impl<'a> Binding<'a> {
             _ => None,
         }
     }
-    pub fn visibility(&self, source: &str) -> Visibility{
+    pub fn visibility(&self, source: &str) -> Visibility {
         visibility::binding_visibility(self.name(source))
     }
 }


ntBre:
Thank you for working on this! However, I think we should try to reach a decision on whether we want to include this rule in Ruff before moving onto an implementation. I probably should have mentioned this explicitly on the issue, but that's why I added the `needs-decision` label (and typically add that label to any new rule request).

Every rule adds some maintenance burden, and in my experience the docstring rules are especially error prone. At a minimum we should wait a little while and see if others are interested in this rule. I'll also approve the workflow run so we can get a sense for the ecosystem impact, but that can be another useful factor.

I searched a bit for previous issues and only really found https://github.com/astral-sh/ruff/issues/12434. I think there could be some overlap with the attribute-related pydoclint rules that haven't been implemented yet.

PR #20776, opened 2025-10-08
[ty] Type-context aware literal promotion

ibraheemdev:
## Summary

Avoid literal promotion when a literal type annotation is provided, e.g.,
```py
x: list[Literal[1]] = [1]
```

Resolves https://github.com/astral-sh/ty/issues/1198. This does not fix issue https://github.com/astral-sh/ty/issues/1284, but it does make it more relevant because after this change, it is possible to directly instantiate a generic type with a literal specialization.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -222,6 +222,13 @@ reveal_type(o)  # revealed: list[tuple[Literal[1], ...]]
 
 p: list[tuple[int, ...]] = [(1, 1, 1)]
 reveal_type(p)  # revealed: list[tuple[int, ...]]
+
+# literal promotion occurs based on assignability, an exact match is not required
+q: list[int | Literal[1]] = [1]
+reveal_type(q)  # revealed: list[int]
+
+r: list[Literal[1, 2, 3, 4]] = [1, 2]
+reveal_type(r)  # revealed: list[Literal[1, 2, 3, 4]]
 ```
 
 ## PEP-604 annotations are supported


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1122,15 +1133,19 @@ impl<'db> SpecializationBuilder<'db> {
...
+    ) -> Specialization<'db> {
         let types = (generic_context.variables_inner(self.db).iter()).map(|(variable, options)| {
             let mut ty = self.types.get(variable).copied();
 
             // When inferring a specialization for a generic class typevar from a constructor call,
             // promote any typevars that are inferred as a literal to the corresponding instance
             // type.
             if options.should_promote_literals {
-                ty = ty.map(|ty| ty.promote_literals(self.db));
+                ty = ty.map(|ty| ty.promote_literals(self.db, tcx));
```
ibraheemdev: Hmm.. this is incorrect, because we're promoting the type of each parameter, not the type of the generic class, so the type annotation specialization does not match. We might need to delay the promotion step for this to work.

crates/ty_python_semantic/src/types.rs
```diff
@@ -845,9 +847,25 @@ impl<'db> Type<'db> {
...
-        self,
+        &self,
+        db: &'db dyn Db,
         known_class: KnownClass,
+    ) -> Option<Specialization<'db>> {
+        let class_literal = known_class.try_to_class_literal(db).unwrap_or_else(|| {
+            panic!(
+                "typeshed should always have a `{}` class in `builtins.py`",
+                known_class.display(db)
+            )
```
sharkdp: So far, I think we only required `object` to exist. It looks like this requires a custom typeshed to have all known classes?

crates/ty_python_semantic/src/types.rs
```diff
@@ -1169,22 +1192,35 @@ impl<'db> Type<'db> {
...
-            _ => self,
+            _ => return self,
+        };
+
+        // Avoid literal promotion if it leads to an unassignable type.
+        if tcx
+            .annotation
+            .is_none_or(|annotation| promoted.is_assignable_to(db, annotation))
+        {
+            return promoted;
```
sharkdp, line 1215: An interesting consequence here is that we would still perform promotion if something else in the annotation makes the promoted type assignable. For example:
```py
from typing import Any, Literal

xs: list[Any | Literal[1]] = [1, 1]

reveal_type(xs)  # list[Any | int]
```
Probably nothing to worry about, though...

AlexWaygood:
> ```diff
> + tests/core/test_scope.py:127:23: error[invalid-argument-type] Argument to bound method `set` is incorrect: Expected `Box[str]`, found `Box[str]`
> + tests/core/test_scope.py:148:19: error[invalid-argument-type] Argument to bound method `set` is incorrect: Expected `Box[str]`, found `Box[str]`
> + tests/core/test_scope.py:159:19: error[invalid-argument-type] Argument to bound method `set` is incorrect: Expected `Box[str]`, found `Box[str]`
> ```

These look quite odd -- any idea what's causing these diagnostics? `Box[str]` must refer to the same `Box[str]` across any single diagnostic in these instances, or we'd automatically use the fully qualified name of `Box` in the diagnostic to disambiguate it (e.g. https://github.com/astral-sh/ruff/pull/20756#issuecomment-3379038485)

This also doesn't look correct -- `dict[str, str]` is one of the union elements in the first type here, so it seems like it should be assignable to that union:

```diff
tornado (https://github.com/tornadoweb/tornado)
+ tornado/test/auth_test.py:563:69: error[invalid-argument-type] Argument to function `url_concat` is incorrect: Expected `None | dict[str, str] | list[tuple[str, str]] | tuple[tuple[str, str], ...]`, found `dict[str, str]`
```

AlexWaygood:
> ```diff
> - colour/graph/conversion.py:1121:30: error[invalid-argument-type] Argument is incorrect: Expected `str`, found `Unknown | str | ((sd: @Todo | SpectralDistribution | MultiSpectralDistributions, cmfs: MultiSpectralDistributions | None = None, illuminant: SpectralDistribution | None = None, k: @Todo | None = None, method: str = str, **kwargs: Any) -> @Todo) | ... omitted 60 union elements`
> + colour/graph/conversion.py:1121:30: error[invalid-argument-type] Argument is incorrect: Expected `str`, found `Unknown | Literal["Spectral Distribution", "CIE XYZ", "Luminous Flux", "Luminous Efficiency", "Luminous Efficacy", "Luminance", "Lightness", "Whiteness", "Yellowness", "CIE xy", "Colorimetric Purity", "Complementary Wavelength", "Dominant Wavelength", "Excitation Purity", "Wavelength", "CIE xyY", "CIE Lab", "CIE Luv", "CIE Luv uv", "CIE UCS", "CIE UCS uv", "CIE UVW", "DIN99", "hdr-CIELAB", "Hunter Lab", "Hunter Rdab", "ICaCb", "ICtCp", "IgPgTg", "IPT", "IPT Ragoo 2021", "Jzazbz", "hdr-IPT", "OSA UCS", "Oklab", "ProLab", "sUCS", "Yrg", "CIE 1931", "CIE 1960 UCS", "CIE 1976 UCS", "RGB", "Scene-Referred RGB", "HSV", "HSL", "HCL", "IHLS", "CMY", "CMYK", "RGB Luminance", "Prismatic", "Output-Referred RGB", "YCbCr", "YcCbcCrc", "YCoCg", "sRGB", "Hexadecimal", "CSS Color 3", "Munsell Colour", "Munsell Value", "CRI", "CQS", "CCT", "Mired", "ATD95", "CIECAM02", "CIECAM02 JMh", "CAM16", "CAM16 JMh", "CIECAM16", "CIECAM16 JMh", "Hellwig 2022", "Hellwig 2022 JMh", "Kim 2009", "Hunt", "LLAB", "Nayatani95", "RLAB", "sCAM", "sCAM JMh", "ZCAM", "CAM02LCD", "CAM02SCD", "CAM02UCS", "CAM16LCD", "CAM16SCD", "CAM16UCS"] | (def sd_to_XYZ(sd: @Todo | SpectralDistribution | MultiSpectralDistributions, cmfs: MultiSpectralDistributions | None = None, illuminant: SpectralDistribution | None = None, k: @Todo | None = None, method: str = Literal["ASTM E308"], **kwargs: Any) -> @Todo) | ... omitted 117 union elements`
> ```

Possibly not something for this PR, but we should look into why we're creating these huge unions in `colour` (the size of which apparently doubles with this PR!). It obviously really slows us down when we get unions that big. We might want to look into whether it's really correct/beneficial to be inferring these huge unions on that codebase. Even if it is, maybe there are some places where we handle unions in an inefficient way that we can optimise. (Operations on unions are inherently at least quadratic, but we might be able to add caching in some places, and there still might be some places where our handling is accidentally worse than quadratic unnecessarily.)

(The reason why we added `colour` as a benchmark originally is that both pyright and mypy used to have truly pathological performance on that codebase. I guess we're starting to see why that might be the case...!)

crates/ty_python_semantic/src/types/generics.rs
sharkdp: I think I'd need an example to understand this better..

crates/ty_python_semantic/src/types.rs
AlexWaygood: we also require `tuple`, here: <https://github.com/astral-sh/ruff/blob/f0d0b5790079cc38ec0f4b34d771a056f1cd1481/crates/ty_python_semantic/src/types/tuple.rs#L200-L217>

But yes, in general I think we should try to impose as few restrictions as possible on users of custom typesheds, and should try our hardest to gracefully fallback to `Unknown` in situations when we try to lookup classes in typeshed

crates/ty_python_semantic/resources/mdtest/type_compendium/tuple.md
```diff
@@ -525,10 +525,6 @@ from typing import Literal
 reveal_type(list((1, 2, 3)))  # revealed: list[int]
 reveal_type(list(((1, 2, 3),)))  # revealed: list[tuple[int, int, int]]
 
-# TODO: we could bidirectionally infer that the user does not want literals to be promoted here,
-# and avoid this diagnostic
-#
-# error: [invalid-assignment] "`list[int]` is not assignable to `list[Literal[1, 2, 3]]`"
```
AlexWaygood, line 531: there are some interesting examples above this for unannotated `frozenset`s (which don't need to be fixed in this PR, I'm just mentioning it because it might be something you'd be interested in fixing in future work). `frozenset`, like `tuple`, is covariant, so for an unannotated `x = frozenset((1, 2, 3))` call, we would ideally infer the type as `frozenset[Literal[1, 2, 3]]`, the same as we would infer `(1, 2, 3)` as `tuple[Literal[1], Literal[2], Literal[3]]` rather than `tuple[int, int, int]`. Unlike with invariant generics, `Literal` types don't really cause any ergonomic issues when they appear in specializations of covariant types.

But we need to be careful: `list(frozenset((1, 2, 3)))` should still be inferred as `list[int]` rather than `list[Literal[1, 2, 3]]` if there's no declared type! And `[frozenset((1, 2))]` should be `list[frozenset[int]]`

crates/ty_python_semantic/src/types.rs
AlexWaygood: Hrm, I think I _would_ prefer `list[Any | Literal[1]]` to be the revealed type there... but I do agree that this seems unlikely to come up in the real world, so it doesn't need to block this PR

crates/ty_python_semantic/src/types.rs
```diff
@@ -5170,11 +5218,13 @@ impl<'db> Type<'db> {
         db: &'db dyn Db,
         name: &str,
         mut argument_types: CallArguments<'_, 'db>,
+        tcx: TypeContext<'db>,
```
AlexWaygood, line 5216: It seems like we currently have only one use of `try_call_dunder` that passes a `tcx` which isn't `TypeContext::default()`, but all callsites become more verbose because of the new parameter. Is it worth maybe having two methods?

```rs
    fn try_call_dunder(
        self,
        db: &'db dyn Db,
        name: &str,
        mut argument_types: CallArguments<'_, 'db>,
    ) -> Result<Bindings<'db>, CallDunderError<'db>> {
        self.try_call_dunder_with_type_context(
	        self,
	        db: &'db dyn Db,
	        name: &str,
	        mut argument_types: CallArguments<'_, 'db>,
            TypeContext::default()
	    )
    }

    fn try_call_dunder_with_type_context(
        self,
        db: &'db dyn Db,
        name: &str,
        mut argument_types: CallArguments<'_, 'db>,
        tcx: TypeContext<'db>,
    ) -> Result<Bindings<'db>, CallDunderError<'db>> {
        // Has the body of the method called `try_call_dunder` on your branch
    }
```

Then most callsites could use the `try_call_dunder()` method and not have to worry about the `tcx` argument.

Or are we anticipating that we'll want to eventually plumb the type context through many more of these `try_call_dunder` calls in due course?

crates/ty_python_semantic/src/types.rs
```diff
@@ -6279,14 +6366,16 @@ impl<'db> Type<'db> {
         self,
         db: &'db dyn Db,
         type_mapping: &TypeMapping<'a, 'db>,
+        tcx: TypeContext<'db>,
```
AlexWaygood, line 6364: same question here -- it seems like the vast majority of `apply_type_mapping()` calls currently just pass in `TypeContext::default()`, so I wonder if it would be more ergonomic to have separate `apply_type_mapping` and `apply_type_mapping_with_context` methods, where the former would just be a thin wrapper around the latter?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -130,13 +130,9 @@ type IntList = list[int]
 m: IntList = [1, 2, 3]
 reveal_type(m)  # revealed: list[int]
 
-# TODO: this should type-check and avoid literal promotion
-# error: [invalid-assignment] "Object of type `list[Unknown | int]` is not assignable to `list[Literal[1, 2, 3]]`"
 n: list[typing.Literal[1, 2, 3]] = [1, 2, 3]
```
T-256, line 133: is that mean `x: list[Literal[1, 2, 3]] = [2]` declaration is fine?

crates/ty_python_semantic/src/types/generics.rs
ibraheemdev: Nevermind, I just needed to map the type variable instances, instead of relying on the order of the parameters.

crates/ty_python_semantic/src/types.rs
ibraheemdev: We also require this for `list`, `set`, and `dict`, which are the same classes that this method is used for: <https://github.com/astral-sh/ruff/blob/db91ac7dce312b24dc7d7adbf0f2857f6771fccd/crates/ty_python_semantic/src/types/infer/builder.rs#L5589-L5592>

It would be easy to fallback and ignore the type context though, if that's preferable.

crates/ty_python_semantic/src/types.rs
ibraheemdev: I think passing `TypeContext::default()` is a useful cue that we are ignoring a potential type context. There are probably cases where we currently have type context but are not making use of it.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
ibraheemdev: Yes, that works fine, as it's equivalent to `x: list[Literal[1] | Literal[2] | Literal[3]] = [2]`.

ibraheemdev:
> These look quite odd -- any idea what's causing these diagnostics?

I fixed these, we were passing incorrect `TypeContext` to generic constructors in some cases, the ecosystem report is much better now.

The diagnostics are interesting though — we currently reinfer the RHS of annotated assignments without any type-context before displaying the diagnostics. This was originally to avoid unioning the inferred type with the type annotation, but it now also means that we always perform literal promotion (as there is no type-context to prevent it). For example:
```py
x: list[Literal[3]] = [2]
```
```
error[invalid-assignment]: Object of type `list[Unknown | int]` is not assignable to `list[Literal[3]]`
 --> x.py:3:1
  |
1 | from typing import Literal
2 |
3 | x: list[Literal[3]] = [2]
  | ^
```
The type in the diagnostic is now *less* precise because of the reinference. I don't think the "`X[T]` is not assignable to `X[T]`" error is ever possible without a bug in ty, but this example also seems problematic. I'm not sure if we should avoid literal promotion entirely during reinference (which could lead to large unions in diagnostics), or just remove the reinference step entirely.

crates/ty_python_semantic/src/types.rs
AlexWaygood: > It would be easy to fallback and ignore the type context though, if that's preferable.

I think that would be better, if it's easy. Even if it's unlikely that users would actually try to use custom typesheds that didn't include these classes, it makes it easier to write tests that have custom typesheds in them if we keep the requirements here to an absolute minimum

crates/ty_python_semantic/src/types.rs
AlexWaygood: Okay, that's fine then 👍

AlexWaygood:
> The diagnostics are interesting though — we currently reinfer the RHS of annotated assignments without any type-context before displaying the diagnostics. This was originally to avoid unioning the inferred type with the type annotation, but it now also means that we always perform literal promotion (as there is no type-context to prevent it).

Those diagnostics don't seem _too_ bad to me tbh — I've seen worse from mypy/pyright for this kind of thing 😆 I wouldn't spend too much time on that right now, personally

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
T-256: it's worth to add such example to mdtests, At first, I thought all values in `Literal[` part of annotation must be as same as declared values.

ibraheemdev:
The regression on `colour` seemed to be because of a bug in the literal promotion code. This change now has no performance impact.

Issue #20774, opened 2025-10-08
Formatter can reuse nested quotes before 3.12


ntBre:
### Summary

While updating our black compatibility tests for https://github.com/astral-sh/ruff/pull/20768, I ran into a failure related to this test case:

```py
f"{f'''{'nested'} inner'''} outer"
```

This is valid on Python versions before 3.12 (when quote reuse became allowed in f-strings), but the formatted result converts the innermost quotes to double quotes, making it invalid:

```py
f"{f'''{"nested"} inner'''} outer"
```

I was hoping to put up a quick PR fixing this, but I'm still working on it and figured I might as well open an issue. I bisected the issue to #13906, but that's not too helpful because this PR is when we first introduced f-string formatting, as Micha pointed out on Discord.

Something interesting I've found besides the bug itself is that our main `format` test doesn't really catch this. `fstring.py` actually has more unsupported syntax errors before formatting than after, which I think prevents our error fingerprinting from working as intended. Our test runner produces a snapshot with the invalid quotes above, without panicking from the new unsupported syntax error, like the black test did.

I've played a bit with more information we could add to the fingerprint, but unfortunately I haven't found a good solution there yet either.

MichaReiser:
Could we list the syntax errors at the end of the test snapshot? We could even use diagnostics to render them nicely

PR #16523, opened 2025-03-05
[syntax-errors] Parenthesized context managers before Python 3.9

ntBre:
Summary
--

WIP currently I just added all of the valid cases from Alex's comment here https://github.com/astral-sh/ruff/pull/16106#issuecomment-2653505671

Test Plan
--
Inline tests

PR #20777, opened 2025-10-08
Render unsupported syntax errors in formatter tests

ntBre:
## Summary

Based on the suggestion in https://github.com/astral-sh/ruff/issues/20774#issuecomment-3383153511, I added rendering of unsupported syntax errors in our `format` test. 

In support of this, I added a `DummyFileResolver` type to `ruff_db` to pass to `DisplayDiagnostics::new` (first commit). Another option would obviously be implementing this directly in the fixtures, but we'd have to import a `NotebookIndex` somehow; either by depending directly on `ruff_notebook` or re-exporting it from `ruff_db`. I thought it might be convenient elsewhere to have a dummy resolver, for example in the parser, where we currently have a separate rendering pipeline [copied](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_parser/tests/fixtures.rs#L321) from our old rendering code in `ruff_linter`. I also briefly tried implementing a `TestDb` in the formatter since I noticed the `ruff_python_formatter::db` module, but that was turning into a lot more code than the dummy resolver.

We could also push this a bit further if we wanted. I didn't add the new snapshots to the black compatibility tests or to the preview snapshots, for example. I thought it was kind of noisy enough (and helpful enough) already, though. We could also use a shorter diagnostic format, but the full output seems most useful once we accept this initial large batch of changes.

## Test Plan

I went through the baseline snapshots pretty quickly, but they all looked reasonable to me, with one exception I noted below. I also tested that the case from #20774 produces a new unsupported syntax error.

### DIFF
--- a/crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
+++ b/crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
@@ -712,6 +712,8 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{'aa'}" }'
 f'{1=: "abcd {'aa'}}'
+# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+# spec, which is valid even before 3.12.
 f'{x:a{z:hy "user"}} \'\'\''
 
 # Changing the outer quotes is fine because the format-spec is in a nested expression.
@@ -1534,6 +1536,8 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{"aa"}" }'
 f'{1=: "abcd {'aa'}}'
+# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+# spec, which is valid even before 3.12.
 f"{x:a{z:hy \"user\"}} '''"
 
 # Changing the outer quotes is fine because the format-spec is in a nested expression.
@@ -2361,6 +2365,8 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{"aa"}" }'
 f'{1=: "abcd {'aa'}}'
+# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+# spec, which is valid even before 3.12.
 f"{x:a{z:hy \"user\"}} '''"
 
 # Changing the outer quotes is fine because the format-spec is in a nested expression.
@@ -2413,26 +2419,26 @@ print(f"{ {}, 1 }")
 
 ### Unsupported Syntax Errors
 error[invalid-syntax]: Cannot use an escape sequence (backslash) in f-strings on Python 3.10 (syntax was added in Python 3.12)
-   --> fstring.py:762:19
+   --> fstring.py:764:19
     |
-760 | f'{1: abcd "{"aa"}" }'
-761 | f'{1=: "abcd {'aa'}}'
-762 | f"{x:a{z:hy \"user\"}} '''"
+762 | # FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+763 | # spec, which is valid even before 3.12.
+764 | f"{x:a{z:hy \"user\"}} '''"
     |                   ^
-763 |
-764 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
+765 |
+766 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
     |
 warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.
 
 error[invalid-syntax]: Cannot use an escape sequence (backslash) in f-strings on Python 3.10 (syntax was added in Python 3.12)
-   --> fstring.py:762:13
+   --> fstring.py:764:13
     |
-760 | f'{1: abcd "{"aa"}" }'
-761 | f'{1=: "abcd {'aa'}}'
-762 | f"{x:a{z:hy \"user\"}} '''"
+762 | # FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+763 | # spec, which is valid even before 3.12.
+764 | f"{x:a{z:hy \"user\"}} '''"
     |             ^
-763 |
-764 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
+765 |
+766 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
     |
 warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.
 
@@ -2448,25 +2454,25 @@ error[invalid-syntax]: Cannot reuse outer quote character in f-strings on Python
 warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.
 
 error[invalid-syntax]: Cannot reuse outer quote character in f-strings on Python 3.10 (syntax was added in Python 3.12)
-   --> fstring.py:771:14
+   --> fstring.py:773:14
     |
-769 | f'{1=: "abcd \'\'}'  # Don't change the outer quotes, or it results in a syntax error
-770 | f"{1=: abcd \'\'}"  # Changing the quotes here is fine because the inner quotes aren't the opposite quotes
-771 | f"{1=: abcd \"\"}"  # Changing the quotes here is fine because the inner quotes are escaped
+771 | f'{1=: "abcd \'\'}'  # Don't change the outer quotes, or it results in a syntax error
+772 | f"{1=: abcd \'\'}"  # Changing the quotes here is fine because the inner quotes aren't the opposite quotes
+773 | f"{1=: abcd \"\"}"  # Changing the quotes here is fine because the inner quotes are escaped
     |              ^
-772 | # Don't change the quotes in the following cases:
-773 | f'{x=:hy "user"} \'\'\''
+774 | # Don't change the quotes in the following cases:
+775 | f'{x=:hy "user"} \'\'\''
     |
 warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.
 
 error[invalid-syntax]: Cannot reuse outer quote character in f-strings on Python 3.10 (syntax was added in Python 3.12)
-   --> fstring.py:762:14
+   --> fstring.py:764:14
     |
-760 | f'{1: abcd "{"aa"}" }'
-761 | f'{1=: "abcd {'aa'}}'
-762 | f"{x:a{z:hy \"user\"}} '''"
+762 | # FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+763 | # spec, which is valid even before 3.12.
+764 | f"{x:a{z:hy \"user\"}} '''"
     |              ^
-763 |
-764 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
+765 |
+766 | # Changing the outer quotes is fine because the format-spec is in a nested expression.
     |
 warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/fstring.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/fstring.py
@@ -706,6 +706,8 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{'aa'}" }'
 f'{1=: "abcd {'aa'}}'
+# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
+# spec, which is valid even before 3.12.
 f'{x:a{z:hy "user"}} \'\'\''
 
 # Changing the outer quotes is fine because the format-spec is in a nested expression.


crates/ruff_python_formatter/tests/snapshots/format@expression__binary.py.snap
```diff
@@ -959,3 +959,39 @@ aaaaaaaaaaa = (
...
+    |
+
+error[invalid-syntax]
+   --> binary.py:496:5
+    |
+494 |   # wrapped in parentheses.
+495 |   (
+496 | /     t"hellooooooooooooooooooooooo \
+497 | |         worlddddddddddddddddddddddddddddddddd"
+    | |______________________________________________^ Cannot use t-strings on Python 3.10 (syntax was added in Python 3.14)
```
ntBre: I think this might be an unrelated bug. I would expect this range/underline to look more like:
```
invalid-syntax: Cannot use t-strings on Python 3.13 (syntax was added in Python 3.14)
 --> try.py:1:15
  |
1 |   aaaaaaaaaaa = t"hellooooooooooooooooooooooo \
  |  _______________^
2 | |         worlddddddddddddddddddddddddddddddddd" + (aaaaaaaaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbb)
  | |______________________________________________^
  |
```

which is what we produce for the unformatted file when I tried it separately.

crates/ruff_db/src/diagnostic/render.rs
```diff
@@ -1170,6 +1170,33 @@ pub fn ceil_char_boundary(text: &str, offset: TextSize) -> TextSize {
...
+    }
+
+    fn is_notebook(&self, _file: &UnifiedFile) -> bool {
+        false
+    }
+
+    fn current_directory(&self) -> &Path {
+        static CWD: std::sync::LazyLock<std::path::PathBuf> =
+            std::sync::LazyLock::new(|| std::path::PathBuf::from("."));
+        &CWD
```
MichaReiser: I don't think this is necessary. You can just return `Path::new(".")`

crates/ruff_python_formatter/tests/snapshots/format@expression__binary.py.snap
```diff
@@ -959,3 +959,39 @@ aaaaaaaaaaa = (
     + (aaaaaaaaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbb)
 )
 ```
+
+
+## Unsupported Syntax Errors
+error[invalid-syntax]
```
MichaReiser: It took me a while to find the message and it seems to be different from what I see in the playground where the primary annotation message comes right after the [invalid-syntax]. 

I haven't checked the CLI but ideally, the rendering would be the same.

crates/ruff_python_formatter/tests/snapshots/format@expression__binary.py.snap
MichaReiser: This looks the same when using `ruff check`. 

```
invalid-syntax: Cannot use t-strings on Python 3.10 (syntax was added in Python 3.14)
   --> scratch.py:496:5
    |
494 |   # wrapped in parentheses.
495 |   (
496 | /     t"hellooooooooooooooooooooooo \
497 | |         worlddddddddddddddddddddddddddddddddd"
    | |______________________________________________^
498 |       + (aaaaaaaaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbb)
499 |   )
    |
```

The ranges look correct in the playground. So this might be diagnostic rendering specific

crates/ruff_python_formatter/tests/snapshots/format@statement__type_alias.py.snap
```diff
@@ -404,3 +403,916 @@ type short_bound_long_default[
     )
 ] = int
 ```
+
+
+## Unsupported Syntax Errors
```
MichaReiser: Let's bump the python version for this test to be at least 3.10.

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
```diff
@@ -807,6 +807,301 @@ with (
 ```
 
 
+### Unsupported Syntax Errors
+error[invalid-syntax]
```
MichaReiser: Hmm, this is somewhat annoying. What we really want here is to not have new syntax errors. Showing all makes it sort of difficult to spot any accidential syntax errors (when it introduced a new syntax error).

That made me think about fingerprinting again and maybe there's a way to improve the fingerprint. 

The formatter shouldn't insert or remove statements. That means, we should be able to match diagnostics to their corresponding statement (include that in the fingerprint). Maybe something like this: 

* If there are any syntax errors
* Traverse the entire AST and build a Vec that contains the range of every statement (in sorted order)
* For each syntax error, search for the statement (index) with the smallest range that fully contains the syntax error's range
* Use that index as part of the fingerprint

We could also just traverse the AST for every syntax error instead of building a vec first

crates/ruff_db/src/diagnostic/render.rs
ntBre: Ah wow 🤦 It looks like I can simplify the version in the linter too:

https://github.com/astral-sh/ruff/blob/4917e080ef9bd13d4e72aec68d8f1c2d37088aa6/crates/ruff_linter/src/fs.rs#L12-L15

I assumed `Path::new` returned a `Path` and not a `&Path`.

crates/ruff_python_formatter/tests/snapshots/format@expression__binary.py.snap
ntBre: Oh right, that's the difference between `Diagnostic::invalid_syntax` and the linter's `create_syntax_error_diagnostic`. I can just copy the latter over here.

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
ntBre: I guess my thinking was that this looks a lot more annoying here than it will in the future, unless we often add many new inputs with unsupported syntax at once. And we can avoid some of those by splitting up the test cases into version-specific groups as you mention above, but that might be bad for other reasons I haven't run into yet.

I can definitely look more into the fingerprinting though, that was my first thought when running into this yesterday. I think I may have briefly explored something vaguely along these lines even. I was hoping that we could attach the `node_index` for the AST node to the syntax error when constructing it, at least in tests, but it looks like we just use `NONE` for the node index in the parser (and often emit syntax errors before we have an AST node anyway).

MichaReiser: Yeah, `node_index` is only initialized in ty. But it's roughly the same as just walking the tree in source order (`SourceOrderVisitor`) and assigning ids.

ntBre: The good news is that this seems to work quite well. I think we can probably revert the unsupported syntax error snapshots? It still seems useful to render the new syntax errors as full diagnostics in the `Formatted code introduced new unsupported syntax errors` message, though, so I think we can keep the resolver.

The bad news is that this uncovered a few more issues in the f-string tests. Not sure about others since that panic aborts the test run.

ntBre: Actually I guess I looked at this too quickly right before standup. All but one of the new panics is kind of a false positive. In cases like this:

```py
f"foo {'\'bar\''}"
```

we know it's okay to reuse a quote since that's allowed on the same Python version as escapes. But the fingerprints are different: we lose a "Cannot use an escape sequence" error and gain a "Cannot reuse outer quote character" error.

So I think we should keep the snapshots but restrict them to the new errors detected by the new fingerprinting scheme and remove this hard panic. Maybe that's what you had in mind initially.

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
```diff
@@ -807,6 +807,29 @@ with (
...
+error[invalid-syntax]: Cannot use parentheses within a `with` statement on Python 3.8 (syntax was added in Python 3.9)
+   --> with.py:359:6
+    |
+357 |     pass
+358 |
+359 | with (
+    |      ^
+360 |     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+361 | ):
+    |
```
ntBre, line 831: These also appear to be false positives. They seem somewhat related to our discussion [here](https://github.com/astral-sh/ruff/pull/16523#discussion_r1984657793), but I thought that only applied to tuples. These cases are fine on 3.8 besides the runtime errors:

```pycon
Python 3.8.20 (default, Oct  2 2024, 16:34:12)
[Clang 18.1.8 ] on linux
>>> with (
...   1 + 2
... ): ...
...
AttributeError: __enter__
>>> if True:
...   with (
...     1
...     if True
...     else 2
...   ):
...     pass
...
  File "<stdin>", line 2, in <module>
AttributeError: __enter__
```

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
MichaReiser: Can you say why the fingerprints are different? I would have expected them to be the same because the syntax error belongs to the same statement.

If the fingerprints are different, we could then consider a `# invalid-syntax: allow` pragma comment to allowlist the few cases where we intentionally want to allow them.

ntBre: We hash the syntax error kind, which has the same variant (`Pep701FString`) but a different inner variant (`Pep701FString(Backslash)` before vs `Pep701FString(NestedQuote)` after).

I can try playing with the pragma comment.

ntBre: I feel like there may be a nicer approach here, but I got something working by checking the preceding line for a pragma comment. It works well for the f-string cases. The remaining syntax error snapshots look like bugs that I plan to follow up on.

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -537,22 +577,107 @@ source_type                = {source_type:?}"#,
...
+    fn new(parsed: &Parsed<Mod>, source: &str) -> Self {
+        let index = LineIndex::from_source_text(source);
+        let mut visitor = Self {
+            nodes: Vec::new(),
+            pragma_comments: Indexer::from_tokens(parsed.tokens(), source)
+                .comment_ranges()
+                .into_iter()
+                .filter(|comment_range| source[*comment_range].contains("invalid-syntax: allow"))
+                .map(|range| index.line_index(range.start()))
+                .collect(),
```
MichaReiser: You can use `CommentRanges::from(parsed.tokens())` instead of `Indexer` since you don't need any of the other `Indexer` metadata.

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -537,22 +577,107 @@ source_type                = {source_type:?}"#,
...
+        visitor
+    }
+
+    /// Return the index of the statement node that contains `range`, if it exists, and if it is not
+    /// ignored by a pragma comment.
+    fn node_id(&self, range: TextRange) -> Option<usize> {
+        let position = self
+            .nodes
+            .iter()
+            .position(|node| node.contains_range(range))?;
```
MichaReiser: I don't think this is correct, or at least, it's not as narrow as it could be

```py
if True:
	a = 10
```

The range of the `if` statement contains both the range of `True` but also of `a = 10`. 

What we want is to find the node with the narrowest range that fully encloses `range`. Roughly:

* Find the first range that contains `range.start` (you could even use a binary search)
* Now find the last range that still contains `range.end`, this is the most narrow range

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -537,22 +577,107 @@ source_type                = {source_type:?}"#,
...
+    }
+}
+
 /// Collects the unsupported syntax errors and assigns a unique hash to each error.
 fn collect_unsupported_syntax_errors(
-    errors: &[UnsupportedSyntaxError],
+    parsed: &Parsed<Mod>,
+    source: &str,
 ) -> FxHashMap<u64, UnsupportedSyntaxError> {
+    let visitor = StmtVisitor::new(parsed, source);
```
MichaReiser: Let's only build this when `parsed.unsupported_syntax_errors` isn't empty

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -537,22 +577,107 @@ source_type                = {source_type:?}"#,
...
+
     let mut collected = FxHashMap::default();
 
-    for error in errors {
-        let mut error_fingerprint = fingerprint_unsupported_syntax_error(error, 0);
+    for error in parsed.unsupported_syntax_errors() {
+        let Some(node_id) = visitor.node_id(error.range) else {
+            continue;
```
MichaReiser: I think we should panic here or at least use `None` as node id instead of omitting the syntax error.

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -422,29 +464,25 @@ fn ensure_unchanged_ast(
...
+        input_path.file_name().unwrap().to_string_lossy(),
+        formatted_code,
+    )
+    .finish();
+    let diagnostics = formatted_unsupported_syntax_errors
+        .values()
+        .map(|error| {
+            let mut diag = Diagnostic::new(DiagnosticId::InvalidSyntax, Severity::Error, error);
+            let span = Span::from(file.clone()).with_range(error.range());
+            diag.annotate(Annotation::primary(span));
```
MichaReiser, line 489: Should we add a note saying that, if expected, they can be suppressed with `invalid-syntax: allow`?

crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
```diff
@@ -2409,3 +2415,38 @@ print(
 # Regression tests for https://github.com/astral-sh/ruff/issues/15536
 print(f"{ {}, 1 }")
 ```
+
+
+### Unsupported Syntax Errors
+error[invalid-syntax]: Cannot use an escape sequence (backslash) in f-strings on Python 3.10 (syntax was added in Python 3.12)
```
MichaReiser, line 2421: Are these only the syntax errors that weren't suppressed with `invalid-syntax` allowed? I'm asking because it's not entirely clear to me when I'm supposed to use `invalid-syntax` vs when it's okay to just tolerate the invalid syntax errors. 

I think I'd either keep panicking in the presence of an invalid syntax error that isn't explicitly suppressed with an `invalid-syntax` pragma comment or render them:

The risk I see with the pragma comments is that it might unintentionally suppress more syntax errors than I intended (e.g. in the `with` test case, I might allow `invalid-syntax` errors to suppress the diagnostics for 3.8, but then introduce an error for 3.9). On the other hand, it makes it very clear that unsupported syntax errors need to be fixed before a PR is merged (or they need to be explicitly suppressed).

If we go with rendering the errors, then I think we should add a short paragraph right below Unsupported syntax errors, saying that these need to be fixed unless they're pre-existing syntax errors already present in the source (and not introduced by the formatter).

ntBre: Yes, we only render the leftover syntax errors that weren't present initially and weren't suppressed with the pragma comment. With the pragma comments, I was thinking we ideally wouldn't tolerate any snapshots, but I didn't want to fix these bugs in the same PR :sweat_smile: So I see what you mean, we should go back to panicking instead of snapshotting once these are resolved.

I think we may also be able to print the fingerprint for new errors so that you could include that in the pragma comment, to make them a bit more specific. Or maybe part of the error message would be easier to use, like in mdtests.

But I think I would still lean toward just rendering all of the new errors too and not having the pragma comments. That's a bit simpler and we didn't have many errors to render once we updated the fingerprint filtering anyway.

crates/ruff_python_formatter/tests/fixtures.rs
ntBre: This one needs to be optional because `node_id` also filters out the pragma comments. Maybe that's a strange API in itself. But I can also restore the `unwrap` inside of `node_id` for finding the enclosing node. That's what I had before I added in the comment check.

crates/ruff_python_formatter/tests/fixtures.rs
ntBre: I was having some trouble with the binary search, so I took a more literal approach:

```rust
        let (position, node_range) = self
            .nodes
            .iter()
            .enumerate()
            .filter(|(_, node)| node.contains_range(range))
            .min_by_key(|(_, node)| node.len())
            .unwrap();
```

crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
MichaReiser: > But I think I would still lean toward just rendering all of the new errors too and not having the pragma comments. That's a bit simpler and we didn't have many errors to render once we updated the fingerprint filtering anyway.

Sounds good to me. Sorry for misdirecting you on the pragma comments

ntBre: No worries, it was interesting to try out! I'll keep my cleanup commits from this morning and then revert them all, just in case we ever want to resurrect them.

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
ntBre: I've looked into this a bit, and I actually don't think we can fix this easily. As noted in the thread linked above, we decided to be a bit more aggressive here and flag tuple cases as syntax errors since they will always cause an error at runtime. But since we parse these two cases the same way:

```py
with (a): ...  # ok, even at runtime, if `a` implements the protocol
with (a,): ... # runtime error on 3.8, tuple definitely doesn't implement the protocol
```

we can't differentiate between a single `WithItem` passed to the context manager, which is okay on 3.8, and a single-item tuple, which we intentionally wanted to flag.

Since 3.8 is EOL and we haven't gotten any user reports, I think we should just accept these snapshots in line with our previous discussion.

crates/ruff_python_formatter/tests/fixtures.rs
```diff
@@ -537,22 +590,64 @@ source_type                = {source_type:?}"#,
...
+
+    /// Return the index of the statement node that contains `range`.
+    fn node_id(&self, range: TextRange) -> usize {
+        self.nodes
+            .iter()
+            .enumerate()
+            .filter(|(_, node)| node.contains_range(range))
+            .min_by_key(|(_, node)| node.len())
+            .expect("Expected an enclosing node in the AST")
+            .0
```
MichaReiser, line 617: I'm fine with this as I don't think it matters much. But if you're interested. Here's an example using `partition_point` to find the first node that starts at or before a given offset.

https://github.com/astral-sh/ruff/blob/6ef72e71edab6436626bff527d2e69a022978c7b/crates/ruff_python_trivia/src/comment_ranges.rs#L39-L41

crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
MichaReiser: I'm not sure if all those errors are now okay or if some of those are new errors introduced by the formatter. If any of them falls into the latter category, could you add a `FIXME` to the fixture file above the relevant line?

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
MichaReiser: I think we could fix it by tracking in `try_parse_parenthesized_with_items` whether there's a trailing comma (peek for a trailing comma inside the `parse_comma_separated_list` callback) and then return it as part of `try_parse_parenthesized_with_items`'s result? 

But I also agree, I don't think this is worth spending much time on.

crates/ruff_python_formatter/tests/fixtures.rs
ntBre: Ah okay, thank you! I found the examples above that using `binary_search_by`, but `partition_point`  is good to know about.

I think I'll just leave it as-is for now if it doesn't matter too much.

crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
ntBre: I thought the errors for line 762 were related to https://github.com/astral-sh/ruff/issues/20774, but it turns out that they're false positives with our syntax error detection instead. I'll add the FIXME and open a separate issue.

crates/ruff_python_formatter/tests/snapshots/format@statement__with.py.snap
ntBre: I'm looking at this quickly now. `parse_comma_separated_list` already returns whether there was a trailing comma, so this might be an easy fix.

PR #20778, opened 2025-10-08
[`flake8-datetimez`] Clarify docs for several rules

### LINKED ISSUE #19384: [docs] DTZ011: Docs confuse `date` with `datetime`
### Summary

Currently, [the docs](https://docs.astral.sh/ruff/rules/call-date-today/#call-date-today-dtz011) says:

> Python datetime objects can be naive or timezone-aware. While an aware object represents a specific moment in time, a naive object does not contain enough information to unambiguously locate itself relative to other datetime objects. Since this can lead to errors, it is recommended to always use timezone-aware objects.
> 
> datetime.date.today returns a naive datetime object. Instead, use datetime.datetime.now(tz=...).date() to create a timezone-aware object.

This is mentioning `datetime` objects multiple times, but this rule is about `date` objects. `date` objects can never be tz-aware (and are also not considered naïve, because that concept makes no sense for them). Overall I find the rule problematic conceptually, but the documentation should at least be accurate and don't mention `datetime` objects, or at least mention them in relation to `date` objects.

### Version

_No response_

ageorgou:
## Summary

Resolves #19384.

- Distinguishes more clearly between `date` and `datetime` objects.
- Uniformly links to the relevant Python docs from rules in this category.

I've tried to be clearer, but there's still a contradiction in the rules as written: we say "use timezone-aware objects", but `date`s are inherently timezone-naive.

Also, the full docs don't always match the error message: for instance, in [DTZ012](https://docs.astral.sh/ruff/rules/call-date-fromtimestamp/), the example says to use:
```python
datetime.datetime.fromtimestamp(946684800, tz=datetime.UTC)
```
while `fix_title` returns "Use `datetime.datetime.fromtimestamp(ts, tz=...)**.date()**` instead".
I have left this as it was for now.

## Test Plan
Ran `mkdocs` locally and inspected result.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_datetimez/rules/call_date_fromtimestamp.rs
+++ b/crates/ruff_linter/src/rules/flake8_datetimez/rules/call_date_fromtimestamp.rs
@@ -18,8 +18,8 @@ use crate::checkers::ast::Checker;
 /// always use timezone-aware objects.
 ///
 /// `datetime.date.fromtimestamp(ts)` returns a naive date object.
-/// Instead, use `datetime.datetime.fromtimestamp(ts, tz=...)` to create a
-/// timezone-aware datetime object.
+/// Instead, use `datetime.datetime.fromtimestamp(ts, tz=...).date()` to
+/// create a timezone-aware datetime object and retrieve its date component.
 ///
 /// ## Example
 /// ```python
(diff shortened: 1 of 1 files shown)


ageorgou:
Updated, let me know if that's what you had in mind.

PR #20783, opened 2025-10-09
 [ty] Support `dataclass_transform` for base class models 

sharkdp:
## Summary

Support `dataclass_transform` when used on a (base) class.

## Typing conformance

* The changes in `dataclasses_transform_class.py` look good, just a few mistakes due to missing `alias` support.
* I didn't look closely at the changes in `dataclasses_transform_converter.py` since we don't support `converter` yet. 

## Ecosystem impact

The impact looks huge, but it's concentrated on a single project (ibis). Their setup looks more or less like this:

* the real `Annotatable`: https://github.com/ibis-project/ibis/blob/d7083c2c96e12bb7b2a1e643a52b4725f4303fcb/ibis/common/grounds.py#L100-L101
* the real `DataType`: https://github.com/ibis-project/ibis/blob/d7083c2c96e12bb7b2a1e643a52b4725f4303fcb/ibis/expr/datatypes/core.py#L161-L179
* the real `Array`: https://github.com/ibis-project/ibis/blob/d7083c2c96e12bb7b2a1e643a52b4725f4303fcb/ibis/expr/datatypes/core.py#L1003-L1006

```py
from typing import dataclass_transform

@dataclass_transform()
class Annotatable:
    pass

class DataType(Annotatable):
    nullable: bool = True

class Array[T](DataType):
    value_type: T
```

They expect something like `Array([1, 2])` to work, but ty, pyright, mypy, and pyrefly would all expect there to be a first argument for the `nullable` field on `DataType`. I don't really understand on what grounds they expect the `nullable` field to be excluded from the signature, but this seems to be the main reason for the new diagnostics here. Not sure if related, but it looks like their typing setup is not really complete (https://github.com/ibis-project/ibis/issues/6844#issuecomment-1868274770, this thread also mentions `dataclass_transform`).

## Test Plan

Update pre-existing tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
@@ -122,9 +122,6 @@ class CustomerModel(ModelBase):
     id: int
     name: str
 
-# TODO: this is not supported yet
-# error: [unknown-argument]
-# error: [unknown-argument]
 CustomerModel(id=1, name="Test")
 ```
 
@@ -216,11 +213,7 @@ class OrderedModelBase: ...
 class TestWithBase(OrderedModelBase):
     inner: int
 
-# TODO: No errors here, should reveal `bool`
-# error: [too-many-positional-arguments]
-# error: [too-many-positional-arguments]
-# error: [unsupported-operator]
-reveal_type(TestWithBase(1) < TestWithBase(2))  # revealed: Unknown
+reveal_type(TestWithBase(1) < TestWithBase(2))  # revealed: bool
 ```
 
 ### `kw_only_default`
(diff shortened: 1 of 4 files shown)


sharkdp:
Looks like we should solve https://github.com/astral-sh/ty/issues/1068 first

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -437,19 +426,15 @@ class DefaultFrozenModel:
...
+f.name = "new"  # error: [invalid-assignment]
 
 class Mutable(DefaultFrozenModel, frozen=False):
     name: str
 
-# TODO: no error here
-# error: [unknown-argument]
 m = Mutable(name="test")
-m.name = "new"  # No error
+# TODO: This should not be an error
```
sharkdp, line 436: This is tracked under *"Support overwriting of \*_default parameters …"* in https://github.com/astral-sh/ty/issues/1327

PR #20792, opened 2025-10-09
[ty] allow any string `Literal` type expression as a key when constructing a `TypedDict`

mtshiba:
## Summary

This PR allows any string `Literal` type expression to be used as a key when constructing a `TypedDict`.

The pattern of reusing constant key strings when constructing a `TypedDict` seems to be common in several libraries (e.g., [home-assistant/core](https://github.com/home-assistant/core/blob/f6fb4c8d5a26d496ca0ebe790776bc208386e83d/homeassistant/components/environment_canada/weather.py#L227C1-L241C14)).
Currently, we only accept string literal keys when constructing a `TypedDict`, but any expression that can be inferred to be of type string `Literal` is also acceptable. In fact, [the spec](https://typing.python.org/en/latest/spec/typeddict.html#use-of-final-values-and-literal-types) explicitly states that `Literal` type expressions and `Final` names ​​should be allowed.

## Test Plan

New tests in `typed_dict.md`

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -5565,14 +5565,9 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             }
         }
 
-        validate_typed_dict_dict_literal(
-            self.db(),
-            &self.context,
-            typed_dict,
-            dict,
-            dict.into(),
-            |expr| self.expression_type(expr),
-        )
+        validate_typed_dict_dict_literal(&self.context, typed_dict, dict, dict.into(), |expr| {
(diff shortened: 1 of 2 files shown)


ibraheemdev:
> In fact, [the spec](https://typing.python.org/en/latest/spec/typeddict.html#use-of-final-values-and-literal-types) explicitly states that Literal type expressions and Final names ​​should be allowed.

FWIW I think the spec only states this for `TypedDict` *operations*, not definitions, but I agree that we should support this anyways as it's common in the ecosystem.

I think we can also update our bidirectional inference code, [which makes the same assumption](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types/infer/builder.rs#L5860).

PR #20793, opened 2025-10-09
[`flake8-logging-format`] Avoid dropping implicitly concatenated pieces in the `G004` fix

pieterh-oai:
## Summary

The original autofix for G004 was quietly dropping everything but the f-string components of any implicit concatenation sequence; this addresses that.

Side note: It looks like `f_strings` is a bit risky to use (since it implicitly skips non-f-string parts); use iter and include implicitly concatenated pieces. We should consider if it's worth having (convenience vs. bit risky). 

## Test Plan

```
cargo test -p ruff_linter
```

Backtest (run new testcases against previous implementation):
```
git checkout HEAD^ crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
cargo test -p ruff_linter

```

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_logging_format/G004_implicit_concat.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_logging_format/G004_implicit_concat.py
@@ -1,0 +1,8 @@
+import logging
+
+variablename = "value"
+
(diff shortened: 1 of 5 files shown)


pieterh-oai:
Thanks for the quick review @ntBre !

PR #18708, opened 2025-06-16
Disallow newlines in format specifiers of single quoted f- or t-strings

MichaReiser:
## Summary

This PR disallows newlines in format specifiers of single-quoted f- and t-strings. This aligns ruff's behavior with CPython's upstream behavior. 

I think we should wait with merging this PR until after the next release so that the formatter can fix-up invalid format specifiers. 

Part of https://github.com/astral-sh/ruff/issues/18632

Fixes https://github.com/astral-sh/ruff/issues/18730

Issue #11820, opened 2024-06-10
formatter deviation: line break not occurring on assignments where LHS goes over line-length


jonathan-hill-visasq:
I'm seeing a case where black is expanding an assignment but ruff is not - even re-writing the black formatted result.

To keep it concise, this example uses an ultra-short line-length of 14:

Input
```
long_variable_name = 1234
```

Black:
```
long_variable_name = (
    1234
)
```

Ruff:
```
long_variable_name = 1234

```

This seems to occur only when the name of the variable assigned to, is longer than line-length (so expanding the line won't bring it within the allowed length).

I can't find this in the documented deviations or other issues at present, and this doesn't seem to be a duplicate of another issue.

MichaReiser:
Yes, I can confirm that #8940 would fix the deviation. The problem is, that the PR has a few regressions around call expression formatting, which is why the PR was abandoned. So fixing this requires some more work and is also something that I don't think we can fix outside of a new formatter style guide as it would change how already formatted code gets formatted.

PR #20794, opened 2025-10-09
Update Black tests

ntBre:
Summary
--

```shell
git clone git@github.com:psf/black.git ../other/black
crates/ruff_python_formatter/resources/test/fixtures/import_black_tests.py ../other/black
```

Then ran our tests and accepted the snapshots

I had to make a small fix to our tuple normalization logic for `del` statements
in the second commit, otherwise the tests were panicking at a changed AST. I
think the new implementation is closer to the intention described in the nearby
comment anyway, though.

The first commit adds the new Python, settings, and `.expect` files, the next three commits make some small
fixes to help get the tests running, and then the fifth commit accepts all but one of the new snapshots. The last commit includes the new unsupported syntax error for one f-string example, tracked in #20774.

Test Plan
--

Newly imported tests. I went through all of the new snapshots and added review comments below. I think they're all expected, except a few cases I wasn't 100% sure about.

### DIFF
--- a/crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fstring.py.snap
+++ b/crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fstring.py.snap
@@ -72,3 +72,17 @@ f'Hello \'{tricky + "example"}\''
 f"Tried directories {str(rootdirs)} \
 but none started with prefix {parentdir_prefix}"
 ```
+
+## New Unsupported Syntax Errors
+
+error[invalid-syntax]: Cannot reuse outer quote character in f-strings on Python 3.10 (syntax was added in Python 3.12)
+ --> fstring.py:6:9
+  |
+4 | f"some f-string with {a} {few():.2f} {formatted.values!r}"
+5 | f"some f-string with {a} {few(''):.2f} {formatted.values!r}"
+6 | f"{f'''{"nested"} inner'''} outer"
+  |         ^
+7 | f'"{f"{nested} inner"}" outer'
+8 | f"space between opening braces: { {a for a in (1, 2, 3)} }"
+  |
+warning: Only accept new syntax errors if they are also present in the input. The formatter should not introduce syntax errors.


crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__remove_lone_list_item_parens.py.snap
```diff
@@ -0,0 +1,283 @@
...
+items = [
+    (
+        {"key1": "val1", "key2": "val2", "key3": "val3"}
+        if some_var == "long strings"
+        else {"key": "val"}
+    )  # comment
+]  # comment
+```
+
+## Black Differences
```
ntBre, line 90: These look like known deviations since we don't implement the `remove_lone_list_item_parens` style  yet.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__type_param_defaults.py.snap
```diff
@@ -25,14 +24,16 @@ def trailing_comma1[T=int,](a: str):
 
 def trailing_comma2[T=int](a: str,):
     pass
+
+def weird_syntax[T=lambda: 42, **P=lambda: 43, *Ts=lambda: 44](): pass
```
ntBre, line 28: This is the only new code in this file, I think we've gotten closer to black here, if I'm interpreting the diff correctly.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__cantfit.py.snap
```diff
@@ -0,0 +1,204 @@
...
++this_is_a_ridiculously_long_name_and_nobody_in_their_right_mind_would_use_one_like_it = 1  # with a comment
+ this_is_a_ridiculously_long_name_and_nobody_in_their_right_mind_would_use_one_like_it = [
+     1,
+     2,
+     3,
+ ]
+-this_is_a_ridiculously_long_name_and_nobody_in_their_right_mind_would_use_one_like_it = (
+-    function()
+-)
++this_is_a_ridiculously_long_name_and_nobody_in_their_right_mind_would_use_one_like_it = function()
```
ntBre, line 68: These all look like improvements to me, but I'm slightly confused about this case and the one above without comments. Maybe we have a slightly different line-length metric from black, I'm not sure why they wrap these.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fmtskip10.py.snap
```diff
@@ -0,0 +1,76 @@
...
+
+## Black Differences
+
+```diff
+--- Black
++++ Ruff
+@@ -1,8 +1,14 @@
+-def foo(): return "mock"  # fmt: skip
+-if True: print("yay")  # fmt: skip
+-for i in range(10): print(i)  # fmt: skip
```
ntBre, line 26: These are related to https://github.com/astral-sh/ruff/issues/11216 and https://github.com/astral-sh/ruff/pull/20633, Dylan's PR will probably clean them up!

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fmtskip10.py.snap
```diff
@@ -0,0 +1,76 @@
...
++    print(i)  # fmt: skip
+ 
+-j =     1 # fmt: skip
+-while j < 10: j += 1  # fmt: skip
++j =     1  # fmt: skip
++while j < 10:
++    j += 1  # fmt: skip
+ 
+-b = [c for c in "A very long string that would normally generate some kind of collapse, since it is this long"] # fmt: skip
++b = [c for c in "A very long string that would normally generate some kind of collapse, since it is this long"]  # fmt: skip
```
ntBre, line 43: I'm less sure about this case and the `j =     1` case above. It looks like we're adding a space before the `# fmt: skip`. It seems like we should probably skip that too.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fstring.py.snap
```diff
@@ -72,3 +72,17 @@ f'Hello \'{tricky + "example"}\''
 f"Tried directories {str(rootdirs)} \
 but none started with prefix {parentdir_prefix}"
 ```
+
+## New Unsupported Syntax Errors
+
+error[invalid-syntax]: Cannot reuse outer quote character in f-strings on Python 3.10 (syntax was added in Python 3.12)
```
ntBre, line 78: Bug tracked in https://github.com/astral-sh/ruff/issues/20774.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fstring_quotations.py.snap
```diff
@@ -0,0 +1,126 @@
...
+
+```diff
+--- Black
++++ Ruff
+@@ -20,7 +20,7 @@
+     + f'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"{"b"}"'
+ )
+ 
+-a = f'bbbbbbb"{"b"}"' "aaaaaaaa"
++a = f'bbbbbbb"{"b"}"aaaaaaaa'
```
ntBre, line 54: Seems intentional, we merge implicitly concatenated strings.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__generics_wrapping.py.snap
```diff
@@ -0,0 +1,561 @@
...
+ 
+ 
+ def func[
+     T,  # comment but it's long so it doesn't just move to the end of the line
+     U,  # comment comment comm comm ent ent
+-    Z: int,  # comment ent ent comm comm comment
++    Z: # comment ent ent comm comm comment
++    int,
+ ]():
+     pass
```
ntBre, line 212: I think our formatting looks better for most of these, but this one looks a bit weird splitting the `Z:` and the `int` annotation. That's what the input looks like though, so we've preserved more of the original formatting.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__long_strings__type_annotations.py.snap
```diff
@@ -0,0 +1,114 @@
...
++++ Ruff
+@@ -21,6 +21,6 @@
+ 
+ 
+ def func(
+-    argument: "int |" "str",
+-) -> Set["int |" " str"]:
++    argument: ("int |str"),
++) -> Set["int | str"]:
+     pass
```
ntBre, line 51: The implicit concatenation here seems nice, maybe a bit weird that we preserve the parens though.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__pep604_union_types_line_breaks.py.snap
ntBre, line 1: Should we revert this? We had a commit to fix the typos, but they get restored by our script. Maybe we should fix the typos upstream?

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__pep_701.py.snap
```diff
@@ -80,8 +80,9 @@ x = f"a{2+2:=^{foo(x+y**2):something else}}b"
 x = f"a{2+2:=^{foo(x+y**2):something else}one more}b"
 f'{(abc:=10)}'
 
-f"This is a really long string, but just make sure that you reflow fstrings {
-    2+2:d}"
+f"""This is a really long string, but just make sure that you reflow fstrings {
+    2+2:d
+}"""
```
ntBre, line 85: This test case was changed in #18708, should we revert this too?

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__prefer_rhs_split_reformatted.py.snap
```diff
@@ -0,0 +1,123 @@
...
+         a=1,
+-        b=(
+-            2 if some_kind_of_data is not None else some_other_kind_of_data
+-        ),  # some explanation of why this is actually necessary
++        b=2
++        if some_kind_of_data is not None
++        else some_other_kind_of_data,  # some explanation of why this is actually necessary
+         c=3,
+     )
+ )
```
ntBre, line 48: As noted in the comment above, this test case was added as part of a fix for https://github.com/psf/black/issues/1187. Before that, black had the same style as ruff. I can't quite tell if this is a known/expected deviation.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_comments7.py.snap
```diff
@@ -157,7 +156,12 @@ square = Square(4) # type: Optional[Square]
 ```diff
 --- Black
 +++ Ruff
-@@ -34,13 +34,9 @@
+@@ -29,17 +29,14 @@
+     MyLovelyCompanyTeamProjectComponent as component,  # DRY
+ )
+ 
++
+ result = 1  # look ma, no comment migration xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
ntBre, line 164: I'm confused about this one. The .py file didn't change for this, and it looks like our diff output just changed a bit, or we're adding a blank line for some reason.

If I actually run the formatter on this file, we're not adding a blank line, so this seems like some kind of diff artifact.

And diffing 0.13.3 (my system ruff) against this PR's version shows nothing too:

```shell
> diff \
<(cargo run -p ruff -- format --no-cache --check --preview  crates/ruff_python_formatter/resources/test/fixtures/black/cases/preview_comments7.py)  \
<(ruff format --no-cache --check --preview  crates/ruff_python_formatter/resources/test/fixtures/black/cases/preview_comments7.py)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/ruff format --no-cache --check --preview crates/ruff_python_formatter/resources/test/fixtures/black/cases/preview_comments7.py`
```

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_import_line_collapse.py.snap
ntBre, line 1: These are related to the `always_one_newline_after_import` preview style tracked in the style guide issue.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_long_dict_values.py.snap
ntBre, line 1: I _think_ these are expected. The quoting changes are definitely expected, the .py file was modified upstream in https://github.com/psf/black/pull/4377, and this file is a test for `wrap_long_dict_values_in_parens`, which we intentionally don't support yet.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_long_strings.py.snap
ntBre, line 1: These also seem expected to me, I think the changes are just a shift from the extra "very" added to one of the strings.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_long_strings__regression.py.snap
```diff
@@ -1171,13 +1171,21 @@ s = f'Lorem Ipsum is simply dummy text of the printing and typesetting industry:
...
  # Regression test for https://github.com/psf/black/issues/3506.
- s = f"With single quote: '  {my_dict['foo']} With double quote: \"  {my_dict['bar']}"
+ # Regressed again by https://github.com/psf/black/pull/4498
+-s = (
+-    "With single quote: ' "
+-    f" {my_dict['foo']}"
+-    ' With double quote: " '
+-    f' {my_dict["bar"]}'
+-)
++s = f"With single quote: '  {my_dict['foo']} With double quote: \"  {my_dict['bar']}"
```
ntBre, line 1184: I think this is expected and in line with the `string_processing` preview style.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_multiline_strings.py.snap
ntBre, line 1: These should be expected in line with `multiline_string_handling` being a non-goal.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_remove_multiline_lone_list_item_parens.py.snap
ntBre, line 1: These seem expected, more `lone_list_item` changes.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_wrap_comprehension_in.py.snap
ntBre, line 1: Expected, `wrap_comprehension_in` feature that we plan to implement.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__remove_except_types_parens.py.snap
ntBre, line 1: These should be fixed in https://github.com/astral-sh/ruff/pull/20768.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__cantfit.py.snap
ntBre: After looking around a bit more, I think this is related to https://github.com/astral-sh/ruff/issues/11820.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__fmtskip10.py.snap
MichaReiser: Yes, we format the `fmt: skip` comment. I don't think this is something that we need to change.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__generics_wrapping.py.snap
MichaReiser: I think it's okay. It's not like Ruff makes it worse. It's just that the user put the comment there.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__long_strings__type_annotations.py.snap
MichaReiser: Removing parens would set a new precedence but we could

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__pep_701.py.snap
MichaReiser: We can change it or leave it as is if the test framework can handle it just fine.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__prefer_rhs_split_reformatted.py.snap
MichaReiser: I don't think this is intentional but I'm not sure it's worth fixing

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__preview_comments7.py.snap
MichaReiser: The `.expect` file changed. I expect this to be because black changed the number of empty lines after imports.

We'll have to be cautious about that change because I suspect it breaks isort compatibility.

crates/ruff_python_formatter/tests/snapshots/black_compatibility@cases__type_param_defaults.py.snap
```diff
@@ -44,35 +45,18 @@ def trailing_comma2[T=int](a: str,):
...
+ def simple[T = something_that_is_long](
+@@ -27,9 +27,7 @@
  
-@@ -31,7 +31,7 @@
+ def trailing_comma1[
+     T = int,
+-](
+-    a: str,
+-):
++](a: str):
```
MichaReiser, line 56: Huh, we ignore the trailing comma

ntBre: Oh, this might actually be okay. This is the input:

```py
def trailing_comma1[T=int,](a: str):
    pass
```

So we respect the trailing comma in the type parameters, and there's no trailing comma in the function parameters.

PR #20795, opened 2025-10-09
[ty] avoid standalone expressions for simple subscript targets

carljm:
## Summary

We don't need the RHS of a simple assignment statement of the form `x[0] = ...` to be a standalone expression, if there is just one target and no unpacking. Eliminate these standalone expressions to reduce Salsa memory usage.

## Test Plan

Existing tests.


carljm:
No performance or memory impact, and I think this makes the code invariants slightly more confusing, so not going to merge it.

PR #20796, opened 2025-10-10
[ty] Use declared variable types as bidirectional type context

ibraheemdev:
## Summary

Use the declared type of variables as type context for the RHS of assignment expressions, e.g.,
```py
x: list[int | str]
x = [1]
reveal_type(x)  # revealed: list[int | str]
```

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -1345,7 +1345,16 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         true
     }
 
-    fn add_binding(&mut self, node: AnyNodeRef, binding: Definition<'db>, ty: Type<'db>) {
+    /// Add a binding for the given definition.
+    ///
+    /// Returns the result of the `infer_value_ty` closure, which is called with the declared type
+    /// as type context.
+    fn add_binding(
+        &mut self,
+        node: AnyNodeRef,
+        binding: Definition<'db>,
+        infer_value_ty: impl FnOnce(&mut Self, TypeContext<'db>) -> Type<'db>,
+    ) -> Type<'db> {
         /// Arbitrary `__getitem__`/`__setitem__` methods on a class do not
         /// necessarily guarantee that the passed-in value for `__setitem__` is stored and
         /// can be retrieved unmodified via `__getitem__`. Therefore, we currently only
@@ -1390,7 +1399,6 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         let file_scope_id = binding.file_scope(db);
         let place_table = self.index.place_table(file_scope_id);
         let use_def = self.index.use_def_map(file_scope_id);
-        let mut bound_ty = ty;
 
         let global_use_def_map = self.index.use_def_map(FileScopeId::global());
         let place_id = binding.place(self.db());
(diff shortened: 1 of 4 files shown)


ibraheemdev:
This is more complicated for class attributes as may need to infer the RHS multiple times during attribute resolution. This is similar to what we do for function overloads, but the difference is that the RHS of an assignment statement is a standalone expression, so I'm not exactly sure how to approach this. We may be able to collect the possible type contexts beforehand and infer the intersection as the single type for the expression, but this needs more work.

I'm going to cut down this PR to just handle simple assignment statements, and implement class attributes in a followup PR.

carljm:
Took a look at some of the ecosystem hits here, just documenting what I found:

* aioredis is https://github.com/astral-sh/ty/issues/1248
* bandersnatch seems like it might be an issue with this PR? Even with `Unknown | None` as the type context, I don't see why we shouldn't narrow away the `None` from the conditional assignment
* similarly with cki-lib, it looks like this may be causing us to fail to narrow types in a problematic way

IIRC our approach to type context is generally just to union it in in the specialization builder? It seems like this approach may need some refinement in cases where the type context is a union, so we avoid unioning in any elements of the context union that are definitely not part of the type of the expression we are inferring?

carljm:
I guess the issue isn't really new to this PR, it already exists: https://play.ty.dev/6d7b5e71-848a-4091-90ea-b56f548a5eff

I think in that case it's going to be important that we can at least narrow away the `None`.

ibraheemdev:
https://github.com/astral-sh/ruff/pull/20528 adds a `filter_disjoint_elements` operation that we might be able to use here (though I'm not sure if will work as is for `T | None`).

dcreager:
> I think in that case it's going to be important that we can at least narrow away the `None`.

Discussed this a bit with @ibraheemdev just now, and I think a rewording of this that I strongly agree with is that we should reveal the same types for `x` and `y` in: [[playground](https://play.ty.dev/c7c679db-2465-4cc9-8e04-0f7e25833214)]

```py
def id[T](x: T) -> T:
    return x

def _(i: int):
    x: int | None = id(i)
    y: int | None = i
    reveal_type(x)
    reveal_type(y)  # revealed: int
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -1332,7 +1332,16 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+    /// Add a binding for the given definition.
+    ///
+    /// Returns the result of the `infer_value_ty` closure, which is called with the declared type
+    /// as type context.
+    fn add_binding(
+        &mut self,
+        node: AnyNodeRef,
+        binding: Definition<'db>,
+        infer_value_ty: impl FnOnce(&mut Self, TypeContext<'db>) -> Type<'db>,
+    ) -> Type<'db> {
```
dcreager, line 1357: nit: Most of the callers below don't look to be using the type context, and provide a closure that just immediately returns the previous `ty` parameter. You can reduce the churn on the diff by keeping `add_binding` with its existing signature, and calling this new variant e.g. `add_binding_with_type_context`.

```rust
fn add_binding(&mut self, node: AnyNodeRef, binding: Definition<'db>, ty: Type<'db>) {
    self.add_binding_with_context(node, binding, |_, _| ty);
}
```

ibraheemdev:
The diff is between this PR and https://github.com/astral-sh/ruff/pull/20875. The ecosystem report looks good, mostly positive changes and a couple regressions due to other known limitations (i.e. https://github.com/astral-sh/ty/issues/1248).

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: This has come up a few times, and I think I would prefer to keep the signature as is for now, because I think it's useful to know when there is a potential type context available that is being ignored, and in fact there are a couple of cases where we probably could be using the type context, but I've held off for now.

dcreager (approved):

PR #20734, opened 2025-10-07
[`ruff`] Use `DiagnosticTag` for more pyupgrade rules

TaKO8Ki:
## Summary

Fixes a part of #20590

Rolling out DiagnosticTag to every existing rule in one go would complicate the review, so in this pull request I’ve limited the changes to **pyupgrade**. (UP019, UP021, UP005, UP026, UP023)

If you’d prefer I include more changes in one go, I can do that—please let me know.

Issue #20590, opened 2025-09-26
Consider using `Deprecated` and `Unnecessary` tags for more rules


MichaReiser:
With https://github.com/astral-sh/ruff/pull/20565 landed, it's now very easy for rules to set the `Unnecessary` or `Deprecated` tags which changes how violations are displayed in editors. We should consider setting those tags for more rules.

PR #20801, opened 2025-10-10
[`ruff`] Use DiagnosticTag for more pyflakes and panda rules

TaKO8Ki:
## Summary

Fixes a part of #20590 and follow up to #20734

Rolling out DiagnosticTag to every existing rule in one go would complicate the review, so in this pull request I’ve limited the changes to flake8 and numpy. (RUF100, PD011, F811, F842, RUF029)

### DIFF
--- a/crates/ruff_linter/src/rules/pandas_vet/rules/subscript.rs
+++ b/crates/ruff_linter/src/rules/pandas_vet/rules/subscript.rs
@@ -165,16 +165,17 @@ pub(crate) fn subscript(checker: &Checker, value: &Expr, expr: &Expr) {
     match attr.as_str() {
         // PD007
         "ix" if checker.is_rule_enabled(Rule::PandasUseOfDotIx) => {
-            checker.report_diagnostic(PandasUseOfDotIx, range)
+            let mut diagnostic = checker.report_diagnostic(PandasUseOfDotIx, range);
+            diagnostic.add_primary_tag(ruff_db::diagnostic::DiagnosticTag::Deprecated);
         }
         // PD008
         "at" if checker.is_rule_enabled(Rule::PandasUseOfDotAt) => {
-            checker.report_diagnostic(PandasUseOfDotAt, range)
+            checker.report_diagnostic(PandasUseOfDotAt, range);
         }
         // PD009
         "iat" if checker.is_rule_enabled(Rule::PandasUseOfDotIat) => {
-            checker.report_diagnostic(PandasUseOfDotIat, range)
+            checker.report_diagnostic(PandasUseOfDotIat, range);
         }
-        _ => return,
-    };
+        _ => (),
+    }
 }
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/pandas_vet/rules/subscript.rs
```diff
@@ -165,16 +165,17 @@ pub(crate) fn subscript(checker: &Checker, value: &Expr, expr: &Expr) {
     match attr.as_str() {
         // PD007
         "ix" if checker.is_rule_enabled(Rule::PandasUseOfDotIx) => {
-            checker.report_diagnostic(PandasUseOfDotIx, range)
+            let mut diagnostic = checker.report_diagnostic(PandasUseOfDotIx, range);
```
MichaReiser: Hmm. I'm not sure about using deprecated here. The method isn't deprecated according to [pandas documentation](https://docs.astral.sh/ruff/rules/pandas-use-of-dot-values/); they only advise against it. To me, that makes this rule similar to other rules that recommend one pattern over another because it's less error prone

crates/ruff_linter/src/rules/ruff/rules/unused_async.rs
```diff
@@ -189,11 +189,12 @@ pub(crate) fn unused_async(
...
 
     if !found_await_or_async {
-        checker.report_diagnostic(
+        let mut diagnostic = checker.report_diagnostic(
             UnusedAsync {
                 name: name.to_string(),
             },
             function_def.identifier(),
         );
+        diagnostic.add_primary_tag(ruff_db::diagnostic::DiagnosticTag::Unnecessary);
```
MichaReiser: I think using the unnecessary tag for this rule is problematic because it makes the function appear as unused when this is inaccurate (it's not the async keyword that's rendered as unused).

crates/ruff_linter/src/rules/pyflakes/rules/unused_annotation.rs
```diff
@@ -46,6 +46,7 @@ pub(crate) fn unused_annotation(checker: &Checker, scope: &Scope) {
             None
         }
     }) {
-        checker.report_diagnostic(UnusedAnnotation { name }, range);
+        let mut diagnostic = checker.report_diagnostic(UnusedAnnotation { name }, range);
```
MichaReiser: Is my understanding correct that `range` is the range for the entire binding? If so, I think we shoulnd't use the unnecessary tag because it makes the whole binding appear as unnecessary, when it's just the annotation that's unnecessary.

TaKO8Ki: You're right. I will remove the tag here.

crates/ruff_linter/src/rules/pandas_vet/rules/subscript.rs
TaKO8Ki: This PandasUseOfDotIx rule is other one PD007 https://docs.astral.sh/ruff/rules/pandas-use-of-dot-ix. It says:

> The .ix method is deprecated as its behavior is ambiguous.

So it seems that the method is deprecated?

TaKO8Ki:
@MichaReiser Thank you for the review. I have addressed your comments.

MichaReiser:
Thanks

PR #20802, opened 2025-10-10
[ty] Treat `Callable`s as bound-method descriptors in special cases

sharkdp:
## Summary

Treat `Callable`s as bound-method descriptors if `Callable` is the return type of a decorator that is applied to a function definition. See the [rendered version of the new test file](https://github.com/astral-sh/ruff/blob/david/callables-as-descriptors/crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md) for the full description of this new heuristic.

I could imagine that we want to treat `Callable`s as bound-method descriptors in other cases as well, but this seems like a step in the right direction. I am planning to add other "use cases" from https://github.com/astral-sh/ty/issues/491 to this test suite.

partially addresses https://github.com/astral-sh/ty/issues/491
closes https://github.com/astral-sh/ty/issues/1333

## Ecosystem impact

All positive

* 2961 removed `unsupported-operator` diagnostics on `sympy`, which was one of the main motivations for implementing this change
* 37 removed `missing-argument` diagnostics, and no added call-error diagnostics, which is an indicator that this heuristic shouldn't cause many false positives
* A few removed `possibly-missing-attribute` diagnostics when accessing attributes like `__name__` on decorated functions. The two added `unused-ignore-comment` diagnostics are also cases of this.
* One new `invalid-assignment` diagnostic on `dd-trace-py`, which looks suspicious, but only because our `invalid-assignment` diagnostics are not great. This is actually a "Implicit shadowing of function" diagnostic that hides behind the `invalid-assignment` diagnostic, because a module-global function is being patched through a `module.func` attribute assignment.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
@@ -7,10 +7,10 @@ python-version = "3.14"
 
 ## Introduction
 
-Some common callable objects (functions, lambdas) are also bound-method descriptors. That is, they
-have a `__get__` method which returns a bound-method object that binds the receiver instance to the
-first argument. The bound-method object therefore has a different signature, lacking the first
-argument:
+Some common callable objects (all functions, including lambdas) are also bound-method descriptors.
+That is, they have a `__get__` method which returns a bound-method object that binds the receiver
+instance to the first argument. The bound-method object therefore has a different signature, lacking
+the first argument:
 
 ```py
 from ty_extensions import CallableTypeOf


crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
```diff
@@ -0,0 +1,190 @@
+# Callables as descriptors?
+
+```toml
+[environment]
+python-version = "3.14"
+```
+
+## Introduction
```
sharkdp, line 8: Part of the text here is a copy from @carljm's post here: https://discuss.python.org/t/when-should-we-assume-callable-types-are-method-descriptors/92938

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -579,7 +579,13 @@ impl<'db> Signature<'db> {
...
 
     pub(crate) fn bind_self(&self, db: &'db dyn Db, self_type: Option<Type<'db>>) -> Self {
-        let mut parameters = Parameters::new(self.parameters().iter().skip(1).cloned());
+        let mut parameters = self.parameters.iter().cloned().peekable();
+
+        if parameters.peek().is_some_and(Parameter::is_positional) {
+            parameters.next();
+        }
+
+        let mut parameters = Parameters::new(parameters);
```
sharkdp, line 590: This is the fix to https://github.com/astral-sh/ty/issues/1333

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
```diff
@@ -0,0 +1,190 @@
...
+        return self.post_process(x * x)
+
+def square_then(c: Callable[[float], int]) -> Callable[[float], int]:
+    return SquareCalculator(c)
+
+class Calculator:
+    square_then_round = square_then(round)
+
+reveal_type(Calculator().square_then_round(3.14))  # revealed: Unknown | int
+```
```
sharkdp, line 190: The argument here is a bit weak. I had trouble coming up with a better example.

It still feels wrong to me to generally assume that returned `Callable` types are function-like if a function-like callable is passed into a call (would we try to detect that in every possible parameter position? or only apply the heuristic for single-argument calls?).

crates/ty_python_semantic/src/types.rs
```diff
@@ -1014,6 +1014,13 @@ impl<'db> Type<'db> {
         }
     }
 
+    pub(crate) const fn unwrap_as_callable_type(self) -> Option<CallableType<'db>> {
```
sharkdp, line 1017: According to our usual naming logic, this function should be called `into_callable`. But that function already exists (and does something else, namely try to *turn* a `Type` into its equivalent `Callable` type). I would propose to rename the existing `Type::into_callable` in order to restore our naming scheme here.

As an aside, I always try `ty.as_*` first (e.g. `ty.as_class_literal()`) before remembering that they all start with `.into_`. "into" sounds like a transformation, but it's really just an unwrapping.

AlexWaygood: IIRC Clippy complains if you name it with `as_*`, because the convention is that `as_*` funcitons are meant to take `&self`, whereas our various `into_*` functions consume `self`? I might be misremembering, though

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -579,7 +579,13 @@ impl<'db> Signature<'db> {
     }
 
     pub(crate) fn bind_self(&self, db: &'db dyn Db, self_type: Option<Type<'db>>) -> Self {
-        let mut parameters = Parameters::new(self.parameters().iter().skip(1).cloned());
+        let mut parameters = self.parameters.iter().cloned().peekable();
+
+        if parameters.peek().is_some_and(Parameter::is_positional) {
+            parameters.next();
+        }
```
AlexWaygood, line 588: shouldn't positional-or-keyword parameters be treated the same way as positional-only parameters here?

Also: for something like this:

```py
class F:
    def method(*args: *tuple[F, int, *tuple[str, ...]]): ...
```

the signature of `F().method` should arguably be `(*args: *tuple[int, *tuple[str, ...]]) -> Unknown`: `bind_self` should pop the first element off the mixed tuple's prefix.

We don't need to worry about such edge cases now, but we could add a TODO comment possibly.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2175,7 +2175,26 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 .try_call(self.db(), &CallArguments::positional([inferred_ty]))
                 .map(|bindings| bindings.return_type(self.db()))
             {
-                Ok(return_ty) => return_ty,
+                Ok(return_ty) => {
+                    let is_input_function_like = inferred_ty
+                        .into_callable(self.db())
+                        .and_then(super::super::Type::unwrap_as_callable_type)
```
```suggestion
                        .and_then(Type::unwrap_as_callable_type)
```

sharkdp: huh, weird. I made that change locally, but apparently the change didn't make it into the final commit.

crates/ty_python_semantic/src/types/signatures.rs
AlexWaygood: (edited the above snippet because I forgot that it needs to be `*args: *tuple[...]` unless you want to say "every element of `*args` needs to have the fancy tuple type")

sharkdp: > shouldn't positional-or-keyword parameters be treated the same way as positional-only parameters here?

`is_positional` includes positional-or-keyword parameters as well



> We don't need to worry about such edge cases now, but we could add a TODO comment possibly.

will do

AlexWaygood: > `is_positional` includes positional-or-keyword parameters as well

argh, sorry. I looked it up to check it, opened `signatures.rs`, did CMD+F for `is_positional`, jumped to the definition for `is_positional_only` (because it was above the definition for `is_positional`) and didn't notice that I was reading the wrong method definition...

sharkdp: No worries. I also checked again because I wasn't sure :smile:

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -581,6 +581,9 @@ impl<'db> Signature<'db> {
     pub(crate) fn bind_self(&self, db: &'db dyn Db, self_type: Option<Type<'db>>) -> Self {
         let mut parameters = self.parameters.iter().cloned().peekable();
 
+        // TODO: Theoretically, for a signature like `f(*args: tuple[MyClass, int, *tuple[str, ...]])` with
```
AlexWaygood: Same mistake [I originally made](https://github.com/astral-sh/ruff/pull/20802#discussion_r2426595738) 😄

This needs to be either a starred tuple or `Unpack[tuple[MyClass, int, *tuple[str, ...]]]`. `*args: tuple[int, ...]` means "in the method body, `*args` will be a tuple of unknown length in which each element of the tuple has type `tuple[int, ...]`". `*args: *tuple[int, ...]` means "in the method body, the variable `args` will be a tuple of unknown length in which each element has type `int`"

```suggestion
        // TODO: Theoretically, for a signature like `f(*args: *tuple[MyClass, int, *tuple[str, ...]])` with
```

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -581,6 +581,9 @@ impl<'db> Signature<'db> {
...
+        // a non-positional first parameter, we should also "skip the first parameter" by modifying the tuple
```
AlexWaygood:
```suggestion
        // a variadic first parameter, we should also "skip the first parameter" by modifying the tuple
```

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
```diff
@@ -0,0 +1,190 @@
+# Callables as descriptors?
+
+```toml
+[environment]
+python-version = "3.14"
+```
+
+## Introduction
+
+Some common callable objects (functions, lambdas) are also bound-method descriptors. That is, they
```
AlexWaygood: `lambda` expressions create instances of `types.FunctionType` just like functions defined with `def` statements, so arguably there's no difference here

```suggestion
Some common callable objects (most commonly, functions!) are also bound-method descriptors. That is, they
```

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
```diff
@@ -0,0 +1,190 @@
...
+class C1:
+    def method(self: C1, x: int) -> str:
+        return str(x)
+
+def _(
+    accessed_on_class: CallableTypeOf[C1.method],
+    accessed_on_instance: CallableTypeOf[C1().method],
+):
+    reveal_type(accessed_on_class)  # revealed: (self: C1, x: int) -> str
+    reveal_type(accessed_on_instance)  # revealed:        (x: int) -> str
```
AlexWaygood, line 28: off-topic: a `reveal_typeform` function would be pretty useful for this kind of mdtest (and for playing around in the playground). Rather than having to create the function, I could just do `reveal_typeform(CallableTypeOf[C1.method])` and it would print the signature in a diagnostic

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+Other callable objects (`staticmethod` objects, instances of classes with a `__call__` method but no
+dedicated `__get__` method) are *not* bound-method descriptors. If accessed as class attributes via
+an instance, they are simply themselves:
```
AlexWaygood, line 33: `staticmethod` _does_ actually have a `__get__` method at runtime -- it returns the underlying function object (whether it's accessed on the class or the instance). On Python <=3.9, `staticmethod`s are not directly callable; it's only by calling `__get__` on the staticmethod that you retrieve the original function object and get something that you can actually call:

```pycon
% uvx python3.9                             
Python 3.9.6 (default, Apr 30 2025, 02:07:17) 
[Clang 17.0.0 (clang-1700.0.13.5)] on darwin
>>> class Foo:
...     @staticmethod
...     def bar(): return 42
... 
>>> original_staticmethod = Foo.__dict__['bar']
>>> original_staticmethod
<staticmethod object at 0x100575370>
>>> original_staticmethod()
TypeError: 'staticmethod' object is not callable
>>> Foo.bar
<function Foo.bar at 0x1005fb4c0>
>>> Foo.bar()
42
```

On Python 3.10, `staticmethod.__call__` was added, so `staticmethod` objects are now directly callable. But it's still the case that staticmethods are descriptors that return the underlying function object when you access them, on either the class object or an instance of the class.

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
AlexWaygood: The most common example where a non-function callable is returned by a decorator is `functools._lru_cache_wrapper`:

```pycon
% py
Python 3.13.1 (main, Jan  3 2025, 12:04:03) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> import functools
>>> class F:
...     @functools.lru_cache
...     def whatever(self): ...
...     
>>> F.whatever
<functools._lru_cache_wrapper object at 0x101549640>
```

But ([somewhat](https://github.com/python/typeshed/issues/11280) [famously](https://github.com/python/typeshed/issues/6347)), `_lru_cache_wrapper` has the same descriptor-binding behaviour that `types.FunctionType` has. So you could use that as an argument in favour of assuming that all objects returned from decorators have this method-binding behaviour if they have `__call__` attributes, I suppose...

crates/ty_python_semantic/src/types.rs
AlexWaygood: > I would propose to rename the existing `Type::into_callable` in order to restore our naming scheme here.

That makes sense to me. Maybe `try_upcast_to_callable`?

crates/ty_python_semantic/src/types/signatures.rs
sharkdp: Oh, I didn't see your edit.

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
sharkdp: Right, I still found it helpful for them to be listed here (this is from Carl's post). They might have a common supertype in `FunctionType`, but they are created by completely different syntactical constructs.

AlexWaygood: Maybe something like this?

```suggestion
Some common callable objects (all functions, including lambdas) are also bound-method descriptors. That is, they
```

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
sharkdp: > a reveal_typeform function would be pretty useful

Yes!!

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
sharkdp: > `staticmethod` _does_ actually have a `__get__` method at runtime -- it returns the underlying function object

Right, I know. The text says "Other callable objects (`staticmethod` objects, […]) are *not* bound-method descriptors", where "bound-method descriptor" would imply the sort of `self`-binding behavior that we know from normal functions. But the text could certainly be clarified further, if you think that'd be valuable.

> On Python 3.10, `staticmethod.__call__` was added, so `staticmethod` objects are now directly callable.

TIL!

AlexWaygood: > The text says "Other callable objects (`staticmethod` objects, […]) are _not_ bound-method descriptors"

Hmmmm, but it also says:

> If accessed as class attributes via
an instance, they are simply themselves

Which I don't think is accurate for staticmethods — they _are_ descriptors, just not in the same way as instance methods

crates/ty_python_semantic/resources/mdtest/call/callables_as_descriptors.md
sharkdp: Or if Python had explicitly specialized calls to generic functions, we could *finally* add C++'s [`std::declval`](https://en.cppreference.com/w/cpp/utility/declval.html):
```py
def declval[T]() -> T:
    raise RuntimeError("can not make up a `T` out of thin air")

reveal_type(declval[CallableTypeOf[C1.method]]())
```

Beautiful!

crates/ty_python_semantic/src/types.rs
sharkdp: Will do that as a follow-up

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2175,7 +2175,26 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 .try_call(self.db(), &CallArguments::positional([inferred_ty]))
                 .map(|bindings| bindings.return_type(self.db()))
             {
-                Ok(return_ty) => return_ty,
+                Ok(return_ty) => {
+                    let is_input_function_like = inferred_ty
+                        .into_callable(self.db())
+                        .and_then(Type::unwrap_as_callable_type)
```
AlexWaygood, line 2181: Note that `unwrap_as_callable_type` will return `None` if `into_callable` returns `Some(Type::Union)`, even if each element of the union returned was a `Callable` type

sharkdp: Oh, that's a good catch! Will look into this tomorrow (post-merge, I don't think it's critical).

AlexWaygood:
It looks like this PR might have caused the ty fuzzer job to take significantly longer in CI. In PRs yesterday, it took around 5 minutes to complete, but following this PR it seems to be taking around 18 minutes to complete. It looks like this change causes pathological performance issues on seeds 32 and 56, for some reason.

It doesn't seem to be causing issues on any of our benchmarks, and mypy_primer still seems to be finishing in good time, so the problematic pattern is probably something that doesn't actually come up much in practice? So a quick fix might be to just add 32 and 56 to this set here: https://github.com/astral-sh/ruff/blob/1ed9b215b981d0283022185f988957d22a02b918/python/py-fuzzer/fuzz.py#L155-L156

We should probably try to pin down what exactly the problematic pattern _is_, though, in the medium/long term

Issue #20724, opened 2025-10-06
FURB188 fix should be marked unsafe except with known `str`s


dscorbett:
### Summary

The fix for [`slice-to-remove-prefix-or-suffix` (FURB188)](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix/) should be marked unsafe unless it is statically known that the full string and its affix are both `str`s. If it is statically known that they are instances of unrelated classes, the rule should be suppressed. Otherwise, the fix can change the program’s behavior. They have to be exactly `str` for the fix to be safe, not subclasses of `str`. [Example](https://play.ruff.rs/e827d415-18c2-424a-86e4-a49c3c731b10):
```console
$ cat >furb188.py <<'# EOF'
class Seq:
    def __init__(self, inner):
        self.inner = inner
    def startswith(self, prefix):
        return tuple(self.inner[:len(prefix)]) == tuple(prefix)
    def __getitem__(self, item):
        return type(self)(self.inner[item])
seq = Seq(("1", "2", "3", "4", "5"))
try:
    if seq.startswith("123"):
        seq = seq[3:]
    print(seq.inner)
except AttributeError as e:
    print(e)

text = "12345"
prefix = ("123",)
try:
    if text.startswith(prefix):
        text = text[len(prefix):]
    print(text)
except TypeError as e:
    print(e)
# EOF

$ python furb188.py
('4', '5')
2345

$ ruff --isolated check furb188.py --select FURB188 --fix
Found 2 errors (2 fixed, 0 remaining).

$ python furb188.py
'Seq' object has no attribute 'removeprefix'
removeprefix() argument must be str, not tuple
```

### Version

ruff 0.13.3 (188c0dce2 2025-10-02)

PR #20805, opened 2025-10-10
Implement safety checks for affix removal in slice_to_remove_prefix_or_suffix rule

tinkerer-shubh:
## Summary
closes: https://github.com/astral-sh/ruff/issues/20724

This pull request implements **safety checks for affix removal** in the `slice_to_remove_prefix_or_suffix` rule.

It introduces a new `FixSafety` enum to determine whether an affix removal is *safe* or *unsafe*. Diagnostic fixes now use these safety checks when applying replacements for prefix and suffix removals.  
This prevents unsafe modifications that could alter runtime behavior — for instance, when dealing with custom sequence types that may not behave like built‑in `str` or `bytes`.

Additionally, helper functions have been added to:
- Classify expression types
- Assess safety based on semantic analysis

## Test Plan

The changes were tested by:
- Adding new test cases to `crates/ruff_linter/resources/test/fixtures/refurb/FURB188.py` to cover the new functionality and verify correct affix removal behavior.
- Updating snapshot tests in  
  `crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB188_FURB188.py.snap`  
  to reflect the new diagnostics and unsafe fix annotations.

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
@@ -1,9 +1,8 @@
-use rustc_hash::FxHashSet;
-
+use ruff_diagnostics::Applicability;
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::{self as ast, PythonVersion};
-use ruff_python_semantic::analyze::typing::find_binding_value;
-use ruff_python_semantic::{BindingId, SemanticModel};
+use ruff_python_semantic::analyze::typing::{find_binding_value, is_bytes, is_string};
+use ruff_python_semantic::{Binding, SemanticModel};
 use ruff_text_size::Ranged;
 
 use crate::Locator;
@@ -81,8 +80,7 @@ pub(crate) fn slice_to_remove_affix_expr(checker: &Checker, if_expr: &ast::ExprI
             let kind = removal_data.affix_query.kind;
             let text = removal_data.text;
 
-            let Some(fix_safety) = remove_affix_fix_safety(&removal_data, checker.semantic())
-            else {
+            let Some(applicability) = affix_applicability(&removal_data, checker.semantic()) else {
                 return;
             };
 
@@ -97,10 +95,7 @@ pub(crate) fn slice_to_remove_affix_expr(checker: &Checker, if_expr: &ast::ExprI
                 generate_removeaffix_expr(text, &removal_data.affix_query, checker.locator());
 
             let edit = Edit::replacement(replacement, if_expr.start(), if_expr.end());
-            diagnostic.set_fix(match fix_safety {
-                FixSafety::Safe => Fix::safe_edit(edit),
-                FixSafety::Unsafe => Fix::unsafe_edit(edit),
-            });
+            diagnostic.set_fix(Fix::applicable_edit(edit, applicability));
         }
     }
 }
@@ -115,8 +110,7 @@ pub(crate) fn slice_to_remove_affix_stmt(checker: &Checker, if_stmt: &ast::StmtI
             let kind = removal_data.affix_query.kind;
             let text = removal_data.text;
 
-            let Some(fix_safety) = remove_affix_fix_safety(&removal_data, checker.semantic())
-            else {
+            let Some(applicability) = affix_applicability(&removal_data, checker.semantic()) else {
                 return;
             };
 
@@ -135,10 +129,7 @@ pub(crate) fn slice_to_remove_affix_stmt(checker: &Checker, if_stmt: &ast::StmtI
             );
 
             let edit = Edit::replacement(replacement, if_stmt.start(), if_stmt.end());
-            diagnostic.set_fix(match fix_safety {
-                FixSafety::Safe => Fix::safe_edit(edit),
-                FixSafety::Unsafe => Fix::unsafe_edit(edit),
-            });
+            diagnostic.set_fix(Fix::applicable_edit(edit, applicability));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
```diff
@@ -500,3 +513,84 @@ struct RemoveAffixData<'a> {
...
+        (TypeKnowledge::KnownNonStr, _) | (_, TypeKnowledge::KnownNonStr) => None,
+        (TypeKnowledge::KnownStr, TypeKnowledge::KnownStr) => Some(FixSafety::Safe),
+        (TypeKnowledge::KnownBytes, TypeKnowledge::KnownBytes) => Some(FixSafety::Safe),
+        (TypeKnowledge::KnownStr, TypeKnowledge::KnownBytes)
+        | (TypeKnowledge::KnownBytes, TypeKnowledge::KnownStr) => None,
+        _ => Some(FixSafety::Unsafe),
+    }
+}
+
+fn classify_expr_type(expr: &ast::Expr, semantic: &SemanticModel) -> TypeKnowledge {
```
ntBre: I don't think we need to re-implement a check like this. We have existing helpers for `str` and `bytes` here:

https://github.com/astral-sh/ruff/blob/bbd3856de8f83c31db84ef8a104d67ae104732ee/crates/ruff_python_semantic/src/analyze/typing.rs#L1065-L1073

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
```diff
@@ -500,3 +513,84 @@ struct RemoveAffixData<'a> {
     /// Contains the prefix or suffix used in `text.startswith(prefix)` or `text.endswith(suffix)`
     affix_query: AffixQuery<'a>,
 }
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum FixSafety {
+    Safe,
+    Unsafe,
+}
```
ntBre: We can reuse our existing `Applicability` type for this:

https://github.com/astral-sh/ruff/blob/bbd3856de8f83c31db84ef8a104d67ae104732ee/crates/ruff_diagnostics/src/fix.rs#L14

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
```diff
@@ -121,11 +134,11 @@ pub(crate) fn slice_to_remove_affix_stmt(checker: &Checker, if_stmt: &ast::StmtI
...
-            diagnostic.set_fix(Fix::safe_edit(Edit::replacement(
-                replacement,
-                if_stmt.start(),
-                if_stmt.end(),
-            )));
+            let edit = Edit::replacement(replacement, if_stmt.start(), if_stmt.end());
+            diagnostic.set_fix(match fix_safety {
+                FixSafety::Safe => Fix::safe_edit(edit),
+                FixSafety::Unsafe => Fix::unsafe_edit(edit),
+            });
```
ntBre: Once we use `Applicability` we can also use `Fix::applicable_edit`.

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB188_FURB188.py.snap
```diff
@@ -21,6 +22,7 @@ help: Use removesuffix instead of assignment conditional upon endswith.
 8  | 
 9  |     return filename
 10 | 
+note: This is an unsafe fix and may change runtime behavior
```
ntBre: We should double-check some of these. Hopefully using the `typing` helpers I mentioned below will help because this seems like a case where we could know that they are both strings. `filename` is annotated in the function signature, and the suffix is a string literal.

dscorbett: Although the annotation says that `isinstance(filename, str)`, it is not guaranteed that `type(filename) is str`. `filename` could be an instance of a subclass of `str` for which FURB188’s assumptions don’t hold. Example:
```console
$ cat >furb188.py <<'# EOF'
class StringX(str):
    def __getitem__(self, key, /):
        return type(self)(super().__getitem__(key))

    def parenthesize(self):
        return type(self)(f"({self})")

def remove_extension_via_slice(filename: str) -> str:
    if filename.endswith(".txt"):
        filename = filename[:-4]
    return filename

print(remove_extension_via_slice(StringX("foo.txt")).parenthesize())
# EOF

$ python furb188.py
(foo)

$ ruff --isolated check furb188.py --select FURB188 --fix

$ python furb188.py 2>&1 | tail -n 1
AttributeError: 'str' object has no attribute 'parenthesize'
```

ntBre: I could be wrong, but my initial thought is that this is overly defensive. I think many of our checks that rely on type annotations, like the `typing` helpers, would be susceptible to this kind of thing.

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
tinkerer-shubh: ahh yess. thanks sm! addressed in https://github.com/astral-sh/ruff/pull/20805/commits/a5694f1ef0f794627090cbce94a479b1147ab210

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
tinkerer-shubh: addressed in https://github.com/astral-sh/ruff/commit/a5694f1ef0f794627090cbce94a479b1147ab210

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
tinkerer-shubh: addressed in https://github.com/astral-sh/ruff/commit/a5694f1ef0f794627090cbce94a479b1147ab210

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB188_FURB188.py.snap
tinkerer-shubh: agreed. I did revisit the PR with this feedback in mind and it does make sense. thanks!

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
```diff
@@ -514,83 +505,89 @@ struct RemoveAffixData<'a> {
...
     }
+
+    false
 }
 
-fn classify_expr_type(expr: &ast::Expr, semantic: &SemanticModel) -> TypeKnowledge {
-    let mut seen = FxHashSet::default();
-    classify_expr_type_inner(expr, semantic, &mut seen)
+/// Check if an expression is deterministically a string literal or a variable bound to a string.
+fn is_expr_string(expr: &ast::Expr, semantic: &SemanticModel) -> bool {
```
ntBre, line 556: Did you try using the existing helpers I recommended in https://github.com/astral-sh/ruff/pull/20805#discussion_r2421775622? This is quite a bit different from what I had in mind, but  I'd be curious to know if you ran into problems with that route.

crates/ruff_linter/src/rules/refurb/rules/slice_to_remove_prefix_or_suffix.rs
```diff
@@ -514,83 +505,89 @@ struct RemoveAffixData<'a> {
...
-
-    match (text_knowledge, affix_knowledge) {
-        (TypeKnowledge::KnownNonStr, _) | (_, TypeKnowledge::KnownNonStr) => None,
-        (TypeKnowledge::KnownStr, TypeKnowledge::KnownStr) => Some(FixSafety::Safe),
-        (TypeKnowledge::KnownBytes, TypeKnowledge::KnownBytes) => Some(FixSafety::Safe),
-        (TypeKnowledge::KnownStr, TypeKnowledge::KnownBytes)
-        | (TypeKnowledge::KnownBytes, TypeKnowledge::KnownStr) => None,
-        _ => Some(FixSafety::Unsafe),
+/// Check if an expression is a tuple or a variable bound to a tuple.
+fn is_tuple_affix(expr: &ast::Expr, semantic: &SemanticModel) -> bool {
```
ntBre, line 538: I'm surprised that we need to special-case tuples. Shouldn't we treat any definite non-strings and non-bytes the same way?

A small variant of your `tuple_affix_should_be_suppressed` test case doesn't get suppressed, as one example:

```py
def object_affix_should_be_suppressed(text: str) -> str:
    prefix = object()
    if text.startswith(prefix):
        text = text[len(prefix):]
    return text
```

I don't think we have a good way to infer a static but non-specific type to handle  this part of the issue:

> If it is statically known that they are instances of unrelated classes, the rule should be suppressed.

I think I'd be fine with checking that the affix and suffix are both bytes or str and then marking the fix  unsafe if that's not true. We could also suppress the rule entirely in that case. There's precedent for that in another rule:

https://github.com/astral-sh/ruff/blob/b562fe030d61e65e595fbf393bd06299cab87cde/crates/ruff_linter/src/rules/pylint/rules/bad_str_strip_call.rs#L76-L96

I think we need basically the same thing here.

PR #20807, opened 2025-10-10
[ty] Add some completion ranking improvements

BurntSushi:
This PR contains a few improvements for completions:

* We now take the text that has been typed into consideration when
generating completions in all cases. Previously, we only did this for
auto-import completions.
* The keyword values `None`, `True` and `False` are now included in
scoped completions.
* Completions are not shown within comments.
* Completions are not shown within strings.
* Completions are _still_ shown within the interpolated segments of
f-strings and t-strings.

We add some new evaluation tasks to capture some of these cases. (And
lots of tests for the string detection logic.)

**Note:** I did make some changes to the `TokenFlags` representation.
I'm open to other ideas. But I wanted to flag this since `TokenFlags`
is used within the representation of a `Token`, which is a fundamental
data type inside of Ruff. Thankfully, even though `TokenFlags` gets
bigger, `Token` itself does not (because of padding/alignment).

### DIFF
--- a/crates/ty_completion_eval/completion-evaluation-tasks.csv
+++ b/crates/ty_completion_eval/completion-evaluation-tasks.csv
@@ -7,7 +7,7 @@ import-deprioritizes-sunder,main.py,0,1
 internal-typeshed-hidden,main.py,0,4
 none-completion,main.py,0,11
 numpy-array,main.py,0,
-numpy-array,main.py,1,
+numpy-array,main.py,1,1
 object-attr-instance-methods,main.py,0,1
 object-attr-instance-methods,main.py,1,1
 raise-uses-base-exception,main.py,0,2


crates/ty_ide/src/completion.rs
```diff
@@ -235,19 +249,61 @@ pub fn completion<'db>(
...
+        ("False", KnownClass::Bool),
+    ];
+    for (name, known) in keywords {
+        if !query.is_match_symbol_name(name) {
+            continue;
+        }
+        completions.push(Completion {
+            name: ast::name::Name::new(name),
+            insert: None,
+            ty: Some(known.to_class_literal(db)),
```
AlexWaygood: `True` is an instance of the `bool` class; it is not the `bool` class object itself. So IIUC what's going on here, this should be

```suggestion
            ty: Some(known.to_instance(db)),
```

(the same applies for the other two keywords)

But also, a more precise type for `True` than `bool` is `Literal[True]` (`Type::BooleanLiteral(true)` in our representation). None is just an instance of `NoneType`, though; there's no more precise type there.

BurntSushi: Derp. Fixed!

crates/ty_ide/src/completion.rs
```diff
@@ -235,19 +249,61 @@ pub fn completion<'db>(
...
+///
+/// Note that at present, these should only be added to "scoped"
+/// completions. i.e., This will include `None`, `True`, `False`, etc.
+fn add_keyword_value_completions<'db>(
+    db: &'db dyn Db,
+    query: &QueryPattern,
+    completions: &mut Vec<Completion<'db>>,
+) {
+    let keywords = [
+        ("None", KnownClass::NoneType.to_instance(db)),
```
AlexWaygood: (sorry for the back-and-forth on this) -- I now remember that we actually have a helper function for getting "instance of `NoneType`" specifically, since that's quite common in ty:

```suggestion
        ("None", Type::none(db)),
```

crates/ruff_python_parser/src/token.rs
```diff
@@ -729,25 +735,27 @@ impl fmt::Display for TokenKind {
 
 bitflags! {
     #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-    pub(crate) struct TokenFlags: u8 {
+    pub struct TokenFlags: u16 {
         /// The token is a string with double quotes (`"`).
         const DOUBLE_QUOTES = 1 << 0;
+        /// The token is a string with single quotes (`'`).
+        const SINGLE_QUOTES = 1 << 1;
```
AlexWaygood: This has a _bit_ of a bad smell for me, because for all the `*Flags` structs in the `ruff_python_ast` crate, we simply use the absence of the `DOUBLE_QUOTES` flag to determine if a string uses single quotes. This change reduces the symmetry between this bitflag and all those bitflags:

https://github.com/astral-sh/ruff/blob/dc64c086336148c57db14060c86f448351710b2e/crates/ruff_python_ast/src/nodes.rs#L809-L833

https://github.com/astral-sh/ruff/blob/dc64c086336148c57db14060c86f448351710b2e/crates/ruff_python_ast/src/nodes.rs#L1398-L1431

https://github.com/astral-sh/ruff/blob/dc64c086336148c57db14060c86f448351710b2e/crates/ruff_python_ast/src/nodes.rs#L1826-L1850

https://github.com/astral-sh/ruff/blob/dc64c086336148c57db14060c86f448351710b2e/crates/ruff_python_ast/src/nodes.rs#L2021-L2075

It seems what you're looking for is a way to determine whether a token with `TokenKind::Unknown` represents an unclosed string. So maybe we could just have an `IS_STRING` (or `ANY_STRING`?) flag that is always set when the lexer starts lexing a string? Something like this?

```diff
--- a/crates/ruff_python_parser/src/lexer.rs
+++ b/crates/ruff_python_parser/src/lexer.rs
@@ -760,10 +760,10 @@ impl<'src> Lexer<'src> {
         #[cfg(debug_assertions)]
         debug_assert_eq!(self.cursor.previous(), quote);
 
+        self.current_flags |= TokenFlags::IS_STRING;
+
         if quote == '"' {
             self.current_flags |= TokenFlags::DOUBLE_QUOTES;
-        } else if quote == '\'' {
-            self.current_flags |= TokenFlags::SINGLE_QUOTES;
         }
 
         if self.cursor.eat_char2(quote, quote) {
@@ -924,10 +924,10 @@ impl<'src> Lexer<'src> {
         #[cfg(debug_assertions)]
         debug_assert_eq!(self.cursor.previous(), quote);
 
+        self.current_flags |= TokenFlags::IS_STRING;
+
         if quote == '"' {
             self.current_flags |= TokenFlags::DOUBLE_QUOTES;
-        } else if quote == '\'' {
-            self.current_flags |= TokenFlags::SINGLE_QUOTES;
         }
 
         // If the next two characters are also the quote character, then we have a triple-quoted
diff --git a/crates/ruff_python_parser/src/token.rs b/crates/ruff_python_parser/src/token.rs
index c26254f23f..329c65812c 100644
--- a/crates/ruff_python_parser/src/token.rs
+++ b/crates/ruff_python_parser/src/token.rs
@@ -736,10 +736,12 @@ impl fmt::Display for TokenKind {
 bitflags! {
     #[derive(Clone, Copy, Debug, PartialEq, Eq)]
     pub struct TokenFlags: u16 {
+        /// The token is probably a string.
+        /// Note that this will not only be set for string tokens, but also for
+        /// tokens with [`TokenKind::Unknown`] that the lexer suspects to be strings.
+        const IS_STRING = 1 << 0;
         /// The token is a string with double quotes (`"`).
-        const DOUBLE_QUOTES = 1 << 0;
-        /// The token is a string with single quotes (`'`).
-        const SINGLE_QUOTES = 1 << 1;
+        const DOUBLE_QUOTES = 1 << 1;
         /// The token is a triple-quoted string i.e., it starts and ends with three consecutive
         /// quote characters (`"""` or `'''`).
         const TRIPLE_QUOTED_STRING = 1 << 2;
@@ -844,12 +846,12 @@ impl TokenFlags {
         self.intersects(TokenFlags::RAW_STRING)
     }
 
-    /// Returns `true` when the flags contain an explicit double or single quotes.
+    /// Returns `true` if this token is probably a string.
     ///
-    /// This is useful for determining whether an "unknown" token corresponds
-    /// to the start of a string.
-    pub const fn has_quotes(self) -> bool {
-        self.intersects(TokenFlags::DOUBLE_QUOTES) || self.intersects(TokenFlags::SINGLE_QUOTES)
+    /// Note that this may be set for tokens with [`TokenKind::Unknown`]
+    /// as well as for tokens with [`TokenKind::String`].
+    pub const fn is_string(self) -> bool {
+        self.intersects(TokenFlags::IS_STRING)
     }
 }
 
diff --git a/crates/ty_ide/src/completion.rs b/crates/ty_ide/src/completion.rs
index 2322389222..2a8be276b4 100644
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -863,10 +863,10 @@ fn is_in_unclosed_string(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
     for (i, t) in tokens.iter().enumerate().rev() {
         match t.kind() {
             TokenKind::Unknown => {
-                // It's possible to lex an unknown token with single/double
-                // quotes. If we got here, then we assume it is the opening
-                // of an unclosed string.
-                if t.flags().has_quotes() {
+                // If the file is currently invalid syntax, the token kind may be
+                // `TokenKind::Unknown`, so also check the flags to see if the
+                // lexer suspects that it's dealing with a string.
+                if t.flags().is_string() {
                     return true;
                 }
                 // Otherwise, we permit any kind of unknown token to be
```

crates/ty_ide/src/completion.rs
```diff
@@ -729,6 +792,137 @@ fn import_tokens(tokens: &[Token]) -> Option<(&Token, &Token)> {
...
+    if !matches!(last.kind(), TokenKind::Name) {
+        return None;
+    }
+    // This one's weird, but if the cursor is beyond
+    // what is in the closest `Name` token, then it's
+    // likely we can't infer anything about what has
+    // been typed. This likely means there is whtiespace
+    // or something that isn't represented in the token
+    // stream. So just give up.
+    if last.range().end() < offset {
```
```suggestion
    if last.end() < offset {
```

crates/ty_ide/src/completion.rs
```diff
@@ -729,6 +792,137 @@ fn import_tokens(tokens: &[Token]) -> Option<(&Token, &Token)> {
...
+        return None;
+    }
+    Some(source.as_str()[last.range()].to_string())
```
```suggestion
    Some(source[last.range()].to_string())
```

crates/ty_ide/src/completion.rs
```diff
@@ -2790,12 +2953,7 @@ f = Foo()
         // but we instead fall back to scope based completions. Since
         // we're inside a string, we should avoid giving completions at
         // all.
```
AlexWaygood: you can delete the TODO comment here -- it's now done :-)

crates/ruff_python_parser/src/token.rs
MichaReiser: Hmm, I think my preferred solution would just be to always lex strings as strings, even if they're unclosed. But I'd have to play with it to see if this breaks any assumptions in Ruff

MichaReiser: I'll give this a try Tomorrow.

sharkdp:
I planned to take a look at this, but it looks like we have two reviews already.

crates/ruff_python_parser/src/token.rs
MichaReiser: See https://github.com/astral-sh/ruff/pull/20848

I'll rebase this PR once https://github.com/astral-sh/ruff/pull/20848 lands

MichaReiser:
Hmm, I think the evaluation results differ between mac and linux?

crates/ty_python_semantic/src/types.rs
```diff
@@ -67,7 +67,7 @@ use crate::types::visitor::any_over_type;
 use crate::unpack::EvaluationMode;
 pub use crate::util::diagnostics::add_inferred_python_version_hint_to_diagnostic;
 use crate::{Db, FxOrderSet, Module, Program};
-pub(crate) use class::{ClassLiteral, ClassType, GenericAlias, KnownClass};
+pub use class::{ClassLiteral, ClassType, GenericAlias, KnownClass};
```
```suggestion
pub(crate) use class::{ClassLiteral, ClassType, GenericAlias, KnownClass};
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -4556,7 +4556,7 @@ impl KnownClass {
     /// the class. If this class is generic, this will use the default specialization.
     ///
     /// If the class cannot be found in typeshed, a debug-level log message will be emitted stating this.
-    pub(crate) fn to_instance(self, db: &dyn Db) -> Type<'_> {
+    pub fn to_instance(self, db: &dyn Db) -> Type<'_> {
```
```suggestion
    pub(crate) fn to_instance(self, db: &dyn Db) -> Type<'_> {
```

MichaReiser:
Sorry @BurntSushi, I'll squash-merge the commits because @AlexWaygood and I don't have as good a commit hygiene as you.

Issue #20800, opened 2025-10-10
FAST002 fix is invalid with `...`


CoderJoshDK:
### Summary

FAST002 converts parameters into their `Annotated` counterparts. If the source code contains an ellipses type (literal `...`) then the fix will generate incorrect semantics.

Take the following signature:

```py
@router.get("/available")
async def get_available(
    ghl_account_id: str = Query(..., description="GHL account identifier"),
) -> SystemsAvailableResponse: ...
```
The use of `...` in the default for `Query` is valid syntax (should be picked up in a lint that you shouldn't do this, but it is "valid"). 
When trying to run fix, you get:

```py
@router.get("/available")
async def get_available(
    ghl_account_id: Annotated[str, Query(description="GHL account identifier")] = ...,
) -> SystemsAvailableResponse: ...
```

When in reality, there is no default value and ruff should convert it to:

```py
@router.get("/available")
async def get_available(
    ghl_account_id: Annotated[str, Query(description="GHL account identifier")],
) -> SystemsAvailableResponse: ...
```

### Version

ruff 0.14.0

CoderJoshDK:
Minimum required repo:

```sh
uv add "fastapi[standard]"
```

```py
# main.py
from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/")
async def root(name: str = Query(..., description="Your name!")):
    return {"message": f"Hello {name}"}
```

```sh
uv run fastapi dev main.py
```

If you then visit the docs page, you will find that "name" is marked as required. If you then run the fix:

```py
from typing import Annotated

from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/")
async def root(name: Annotated[str, Query(description="Your name!")] = ...):
    return {"message": f"Hello {name}"}
```

And visit docs now; you find name is no longer required. Implying a default. But executing the endpoint will cause an error
```
ValueError: [TypeError("'ellipsis' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

So I would consider this fix as incorrect, seeing that it changes the behavior. To match behavior after fix, drop the `...`.

ntBre:
Thanks for the report, this does seem like a bug. At first I thought this was a general type error, but as you said, the ellipsis has special meaning to fastapi, and I think we should handle that in the fix.

PR #20810, opened 2025-10-11
[`fastapi`] Handle ellipsis defaults in FAST002 autofix (`FAST002`)

hengky-kurniawan-1:
## Summary

Implement handling of ellipsis (`...`) defaults in the `FAST002` autofix to correctly differentiate between required and optional parameters in FastAPI route definitions.

Previously, the autofix did not properly handle cases where parameters used `...` as a default value (to indicate required parameters). This could lead to incorrect transformations when applying the autofix.

This change updates the `FAST002` autofix logic to:
- Correctly recognize `...` as a valid FastAPI required default.
- Preserve the semantics of required parameters while still applying other autofix improvements.
- Avoid incorrectly substituting or removing ellipsis defaults.

Fixes https://github.com/astral-sh/ruff/issues/20800

## Test Plan

Added a new test fixture at:
```crates/ruff_linter/resources/test/fixtures/fastapi/FAST002_2.py```

### DIFF
--- a/crates/ruff_linter/src/rules/fastapi/snapshots/ruff_linter__rules__fastapi__tests__fast-api-non-annotated-dependency_FAST002_2.py.snap
+++ b/crates/ruff_linter/src/rules/fastapi/snapshots/ruff_linter__rules__fastapi__tests__fast-api-non-annotated-dependency_FAST002_2.py.snap
@@ -240,7 +240,7 @@ FAST002 [*] FastAPI dependency without `Annotated`
 80 |     # First param should be fixed with default preserved  
 81 |     optional_param: str = Query("default", description="Optional"),
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-82 |     # Second param should be fixed
+82 |     # Second param should not be fixed because of the preceding default
 83 |     required_param: str = Query(..., description="Required"),
    |
 help: Replace with `typing.Annotated`
@@ -257,7 +257,7 @@ help: Replace with `typing.Annotated`
 81 |     # First param should be fixed with default preserved  
    -     optional_param: str = Query("default", description="Optional"),
 82 +     optional_param: Annotated[str, Query(description="Optional")] = "default",
-83 |     # Second param should be fixed
+83 |     # Second param should not be fixed because of the preceding default
 84 |     required_param: str = Query(..., description="Required"),
 85 |     # Third param should be fixed with default preserved
 note: This is an unsafe fix and may change runtime behavior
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/ruff_linter/resources/test/fixtures/fastapi/FAST002_2.py [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+) -> str:
+    return "authenticated"
+
+
+@app.get("/test10")
+async def test_security_with_ellipsis(
+    # This should become: token: Annotated[str, Security(oauth2_scheme)]
+    token: str = Security(oauth2_scheme),
+) -> str:
+    return "secured"
```
ntBre: I think we could probably omit these cases since `Depends` and `Security` don't take ellipsis arguments. Is that right?

The names feel slightly misleading to me since there aren't ellipses in these examples, and we have other tests for `Depends` and `Security` in FAST002_1.py.

crates/ruff_linter/resources/test/fixtures/fastapi/FAST002_2.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@app.get("/test11")
+async def test_mixed_parameters(
+    # First param should be fixed
+    required_param: str = Query(..., description="Required"),
+    # Second param should be fixed with default preserved  
+    optional_param: str = Query("default", description="Optional"),
+) -> str:
+    return f"{required_param}-{optional_param}"
```
ntBre: Do we have a test case where we bail out  because we removed a default? I think swapping the order of these two parameters to trigger that behavior might be more interesting, unless I missed another case like that.

ntBre (commented):
Thank you! This looks great. I just had a couple of small ideas/suggestions about the tests.

ntBre:
It looks like the linux test failure was just a transient issue. The only real CI problem is the formatting.

hengky-kurniawan-1:
Thanks for the review! I’ll update it soon.

crates/ruff_linter/resources/test/fixtures/fastapi/FAST002_2.py
hengky-kurniawan-1: Interesting thought! I’ll update it to include that case soon.

crates/ruff_linter/resources/test/fixtures/fastapi/FAST002_2.py
hengky-kurniawan-1: Oh right, I’ll remove those cases.

crates/ruff_linter/src/rules/fastapi/snapshots/ruff_linter__rules__fastapi__tests__fast-api-non-annotated-dependency_FAST002_2.py.snap
```diff
@@ -0,0 +1,303 @@
...
+
+FAST002 FastAPI dependency without `Annotated`
+  --> FAST002_2.py:83:5
+   |
+81 |     optional_param: str = Query("default", description="Optional"),
+82 |     # Second param should be fixed
+83 |     required_param: str = Query(..., description="Required"),
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+84 |     # Third param should be fixed with default preserved
+85 |     another_optional_param: int = Query(42, description="Another optional"),
```
ntBre: It looks like this comment is just outdated. The new comment in the .py file looks perfect.

hengky-kurniawan-1: Nice catch.
I’ve updated the snapshot accordingly. Thanks!

PR #18473, opened 2025-06-04
Infer type of self as typing.Self in method body

Glyphack:
Part of #159

Add support for inferring type of self parameter in methods as `typing.Self`. This PR is different than https://github.com/astral-sh/ruff/pull/18007. This PR assigns the type when a symbol is resolved to function parameter. The other PR assigns the type during a method call.

PR #20812, opened 2025-10-11
[ty] Infer type of self as typing.Self in method body (with updated salsa, micha playground)

MichaReiser:
This is https://github.com/astral-sh/ruff/pull/18473 but my own version where I play with an updated Salsa version.

I don't plan on merging this PR. It's just to get perf numbers, see if mypy primer passes etc.


MichaReiser:
Setuptools now panics with too many iterations for `setuptools/_vendor/jaraco/collections/__init__.py`. This seems correct as we now end up with a pretty ridiculous type. The other project also fail because of too many iteration errors. This seems no longer salsa-related but is an issue with our convergence functions.

PR #20814, opened 2025-10-11
[ty] Fix false-positive diagnostics on `super()` calls

AlexWaygood:
## Summary

This PR fixes a variety of false-positive `invalid-super-argument` diagnostics that we can emit on `main`. These aren't _that_ common in the ecosystem at the moment, but they will become much more common when we infer a good type for `self` in method bodies. https://github.com/astral-sh/ruff/pull/20723#issuecomment-3372209543 indicates that if we added an improved type of `self` right now, it would lead to nearly 500 new `invalid-super-argument` diagnostics being added, and I'd guess that nearly all of those are false positives.

There were two problems with our current logic: this `match` was not an exhaustive `match`, and was returning `None` for a large number of types. In reality, nearly all objects are valid as the second argument to a `super(x, y)` call as long as they are an instance or subclass of the first argument. (The same is true if the arguments are provided implicitly by the interpreter, as is usually the case.) This PR makes the `match` exhaustive, and significantly reduces the number of types which we disallow as the second argument to `super()`: we now only disallow "purely structural" types such as `Callable` types and synthesized protocols:

https://github.com/astral-sh/ruff/blob/7064c38e53714edd76a9c2f80fed67b38a0b7c54/crates/ty_python_semantic/src/types.rs#L11332-L11359

The second problem relates to this `is_subclass_of` call here:

https://github.com/astral-sh/ruff/blob/7064c38e53714edd76a9c2f80fed67b38a0b7c54/crates/ty_python_semantic/src/types.rs#L11450-L11454

With our current implementation of `is_subclass_of`, `list[Unknown]` is not recognised as being a "subclass of" `Sequence[int]`, for example, because `Sequence[int]` does not literally appear in the MRO of `list[Unknown]` (`Sequence[Unknown]` does). We've [discussed](https://github.com/astral-sh/ty/issues/512#issuecomment-2923625966) [elsewhere](https://github.com/astral-sh/ty/issues/492#issuecomment-2906000565) that this is a bit of a footgun since it doesn't really correspond to the subclass relationship between class objects at runtime, but for now I just work around the footgun :-)

Fixes https://github.com/astral-sh/ty/issues/987, fixes https://github.com/astral-sh/ty/issues/697

## Test plan

Mdtests and snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -11242,20 +11242,32 @@ impl<'db> EnumLiteralType<'db> {
     }
 }
 
+/// Enumeration of ways in which a `super()` call can cause us to emit a diagnostic.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum BoundSuperError<'db> {
+    /// The second argument to `super()` (which may have been implicitly provided by
+    /// the Python interpreter) has an abstract or structural type.
+    /// It's impossible to determine whether a `Callable` type or a synthesized protocol
+    /// type is an instance or subclass of the pivot class, so these are rejected.
     AbstractOwnerType {
         owner_type: Type<'db>,
-        error_context: Option<TypeVarInstance<'db>>,
-    },
-    InvalidPivotClassType {
         pivot_class: Type<'db>,
+        /// If `owner_type` is a type variable, this contains the type variable instance
+        typevar_context: Option<TypeVarInstance<'db>>,
     },
+    /// The first argument to `super()` (which may have been implicitly provided by
+    /// the Python interpreter) is not a valid class type.
+    InvalidPivotClassType { pivot_class: Type<'db> },
+    /// The second argument to `super()` was not a subclass or instance of the first argument.
+    /// (Note that both arguments may have been implicitly provided by the Python interpreter.)
     FailingConditionCheck {
         pivot_class: Type<'db>,
         owner: Type<'db>,
-        error_context: Option<TypeVarInstance<'db>>,
+        /// If `owner_type` is a type variable, this contains the type variable instance
+        typevar_context: Option<TypeVarInstance<'db>>,
     },
+    /// It was a single-argument `super()` call, but we were unable to determine
+    /// the implicit arguments provided by the Python interpreter.
     UnavailableImplicitArguments,
 }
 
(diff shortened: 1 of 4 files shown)


AlexWaygood:
With this PR, the `super()` implementation becomes large enough that we should move it to a separate `types::bound_super` submodule. I haven't done that in this PR, to keep the diff somewhat readable, but I can do it in a followup.

crates/ty_python_semantic/src/types.rs
```diff
@@ -11279,6 +11312,38 @@ impl BoundSuperError<'_> {
             }
         }
     }
+
+    fn add_context(db: &'db dyn Db, diagnostic: &mut Diagnostic, type_var: TypeVarInstance<'db>) {
+        match type_var.bound_or_constraints(db) {
+            Some(TypeVarBoundOrConstraints::UpperBound(bound)) => diagnostic.info(format_args!(
+                "Type variable `{}` has upper bound `{}`, \
+                which is not an instance or subclass",
```
sharkdp: should this (and similar messages below) specify "which is not an instance or a subclass of …", if possible?

crates/ty_python_semantic/src/types.rs
```diff
@@ -11401,16 +11429,168 @@ impl<'db> BoundSuperType<'db> {
...
+            Type::Union(union) => {
+                return Ok(UnionType::from_elements(
+                    db,
+                    union
+                        .elements(db)
+                        .iter()
+                        .copied()
+                        .map(delegate_to)
+                        .collect::<Result<Vec<_>, _>>()?,
+                ));
```
sharkdp: A version that avoids the additional allocation would be:
```suggestion
                return Ok(union
                    .elements(db)
                    .iter()
                    .copied()
                    .try_fold(UnionBuilder::new(db), |builder, element| {
                        delegate_to(element).map(|ty| builder.add(ty))
                    })?
                    .build());
```

AlexWaygood: Oh, nice! This branch was unchanged from `main`, but this seems like a better way of doing this for sure

AlexWaygood:
> I mostly reviewed the implementation and trust you on the semantics here (or to pull someone in for a second review if you are unsure about parts of it).

I checked all the `reveal_type` calls against the actual type at runtime that you get in the REPL.

Some parts of this I'm slightly unsure about are:
- _Should_ we be banning abstract/structural types as the second argument to `super()`? I don't really know what our behaviour should be if we should allow them. Anyway, we banned them before this PR (but with a worse error message), so this PR didn't really change the status quo there. If others disagree with it and have a good suggestion for what our behaviour should be there, I can make a followup PR and rip it out
- Some of the error messages could probably still be improved (but again, I think they're better than they were before this PR)
- It's not sound in general to upcast a `TypedDict` to a `dict` type like I'm doing there, but I _think_ it's okay in this instance? Anyway, that's an extreme edge case that probably nobody will actually come across in reality

I think we can easily iterate on any of these, which is why I merged despite being unsure about them 😄

Issue #13708, opened 2024-10-10
Extremely verbose `mdtest` output if a revealed type is not what is expected


AlexWaygood:
If you have a `# revealed` assertion on a line in an `mdtest`, and the revealed type is not what you are asserting, the output is something like this:

```
  /src/test.py
    line 16: unmatched assertion: revealed: tuple[Literal[A], Literal[B], Literal[C], Literal[A], Literal[object]]
    line 16: unexpected error: [undefined-reveal] "`reveal_type` used without importing it; this is allowed for debugging convenience but will fail at runtime"
    line 16: unexpected error: [revealed-type] "Revealed type is `tuple[Literal[A], Literal[X], Literal[Y], Literal[O], Literal[object]]`"
```

Really I only made a single mistake here: the revealed type is not what I asserted it would be. However, for this single mistake, I have to pay the "fine" of three (long) lines of `mdtest` output!

AlexWaygood:
(cc. @carljm)

AlexWaygood:
That sounds reasonable to me, too... though I'm still not sure I _fully_ understand why we don't simply suppress all the `unimported-reveal` diagnostics in `mdtest`s. If a contributor writes this in a test:

```py
reveal_type(2)
```

And the test comes back at them with a message telling them

```
    line 16: unexpected error: [undefined-reveal] "`reveal_type` used without importing it; this is allowed for debugging convenience but will fail at runtime"
```

they'll surely think that they need to import `reveal_type` in order to make the test to pass, but that's not true (the error will go away as soon as they add a `# revealed` assertion).

In https://github.com/astral-sh/ruff/pull/13695#discussion_r1795629361 you mentioned that you wanted to emit these diagnostics to be emitted in `mdtest` so that you'd be able to test the diagnostics emitted by `reveal_type` itself, but I'm not sure I see why the tests for `reveal_type` itself _have_ to use `mdtest`. ISTM that a simpler way of doing it would be to have it so that the `unimported-reveal` diagnostics are never emitted in the context of `mdtest`, and the (small number of) tests for `reveal_type` itself are not written using `mdtest`.

carljm:
[...]

PR #20820, opened 2025-10-12
[ty] Filter out `revealed-type` and `undefined-reveal` diagnostics from mdtest snapshots

AlexWaygood:
## Summary

As we discussed in ty planning on Friday, `revealed-type` and `undefined-reveal` diagnostics often make mdtest snapshots very noisy. They're almost never relevant to the reason why you've enabled snapshotting for a specific mdtest. The `undefined-reveal` diagnostics can be avoided by explicitly importing `reveal_type` (which is not something you'd usually do in an mdtest), but the `revealed-type` diagnostics can only be avoided by having a separate mdtest entirely that enables snapshotting and avoids having any `reveal_type` calls in it.

This PR suppresses `revealed-type` and `undefined-reveal` diagnostics from the diagnostics snapshotted during an mdtest run. If we do want to capture a snapshot that shows how these diagnostics are displayed, we can still do that as an end-to-end CLI test in the `ty` crate.

## Test Plan

Snapshots!

### DIFF
--- a/crates/ty_test/src/lib.rs
+++ b/crates/ty_test/src/lib.rs
@@ -16,7 +16,7 @@ use ruff_source_file::{LineIndex, OneIndexed};
 use std::backtrace::BacktraceStatus;
 use std::fmt::Write;
 use ty_python_semantic::pull_types::pull_types;
-use ty_python_semantic::types::check_types;
+use ty_python_semantic::types::{UNDEFINED_REVEAL, check_types};
 use ty_python_semantic::{
     Module, Program, ProgramSettings, PythonEnvironment, PythonPlatform, PythonVersionSource,
     PythonVersionWithSource, SearchPath, SearchPathSettings, SysPrefixPathOrigin, list_modules,
@@ -383,7 +383,7 @@ fn run_test(
             if test.should_snapshot_diagnostics() {
                 snapshot_diagnostics.extend(diagnostics.into_iter().filter(|diagnostic| {
                     diagnostic.id() != DiagnosticId::RevealedType
-                        && !diagnostic.id().is_lint_named("undefined-reveal")
+                        && !diagnostic.id().is_lint_named(&UNDEFINED_REVEAL.name())
                 }));
             }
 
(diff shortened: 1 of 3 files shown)


MichaReiser:
I think I'd prefer if we disable the rule instead of filtering out the diagnostics here:

https://github.com/astral-sh/ruff/blob/dc64c086336148c57db14060c86f448351710b2e/crates/ty_test/src/db.rs#L29

We could then (in the future, not in this PR), add a `rule_selection` option to mdtests, to re-enable the rule when necessary

AlexWaygood:
> I think I'd prefer if we disable the rule instead of filtering out the diagnostics here:

That would disable these rules for mdtest as a whole, rather than just the snapshots, though, right? There's a fair amount of logic in mdtest that's specifically there so that we don't have to do that. I suggested just suppressing these rules instead in mdtest [just over a year ago](https://github.com/astral-sh/ruff/issues/13708#issuecomment-2406098565), but @carljm was against it, which is why the approach I'm taking here attempts to specifically filter the diagnostics out of mdtest _snapshots_ without disabling these diagnostics for regular mdtest.

crates/ty_test/src/lib.rs
```diff
@@ -377,8 +377,14 @@ fn run_test(
...
                 }),
             };
+
+            // Filter out `revealed-type` and `undefined-reveal` diagnostics from snapshots,
+            // since they make snapshots very noisy!
             if test.should_snapshot_diagnostics() {
-                snapshot_diagnostics.extend(diagnostics);
+                snapshot_diagnostics.extend(diagnostics.into_iter().filter(|diagnostic| {
+                    diagnostic.id() != DiagnosticId::RevealedType
+                        && !diagnostic.id().is_lint_named("undefined-reveal")
```
MichaReiser: Can't we use `UNDEFINED_REVEAL::name()` here to avoid hardcoding the name

AlexWaygood: It requires making `UNDEFINED_REVEAL` `pub` rather than `pub(crate)`, which would make it the first `pub` diagnostic in `types/diagnostic.rs`. But I'm okay with that if you are 😄

MichaReiser: I'm fine with that

PR #18327, opened 2025-05-26
[ty] Don't check classes and overloads if the file isn't first-party code

AlexWaygood:

PR #20822, opened 2025-10-12
[ty] Diagnostic for generic classes that reference typevars in enclosing scope

dcreager:
Generic classes are not allowed to bind or reference a typevar from an enclosing scope:

```py
def f[T](x: T, y: T) -> None:
    class Ok[S]: ...
    # error: [invalid-generic-class]
    class Bad1[T]: ...
    # error: [invalid-generic-class]
    class Bad2(Iterable[T]): ...

class C[T]:
    class Ok1[S]: ...
    # error: [invalid-generic-class]
    class Bad1[T]: ...
    # error: [invalid-generic-class]
    class Bad2(Iterable[T]): ...
```

It does not matter if the class uses PEP 695 or legacy syntax. It does not matter if the enclosing scope is a generic class or function. The generic class cannot even _reference_ an enclosing typevar in its base class list.

This PR adds diagnostics for these cases.

In addition, the PR adds better fallback behavior for generic classes that violate this rule: any enclosing typevars are not included in the class's generic context. (That ensures that we don't inadvertently try to infer specializations for those typevars in places where we shouldn't.) The `dulwich` ecosystem project has [examples of this](https://github.com/jelmer/dulwich/blob/d912eaaffd60b4978ff92b91a8300e2cd234e0fc/dulwich/config.py#L251) that were causing new false positives on #20677.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -855,7 +855,9 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             }
 
             let scope = class.body_scope(self.db()).scope(self.db());
-            if let Some(parent) = scope.parent() {
+            if self.context.is_lint_enabled(&INVALID_GENERIC_CLASS)
+                && let Some(parent) = scope.parent()
+            {
                 for self_typevar in class.typevars_referenced_in_definition(self.db()) {
                     let self_typevar_name = self_typevar.typevar(self.db()).name(self.db());
                     for enclosing in enclosing_generic_contexts(self.db(), self.index, parent) {
--- a/crates/ty_python_semantic/src/types/context.rs
+++ b/crates/ty_python_semantic/src/types/context.rs
@@ -100,6 +100,10 @@ impl<'db, 'ast> InferContext<'db, 'ast> {
         self.diagnostics.get_mut().extend(other);
     }
 
+    pub(super) fn is_lint_enabled(&self, lint: &'static LintMetadata) -> bool {
+        LintDiagnosticGuardBuilder::severity_and_source(self, lint).is_some()
+    }
+
     /// Optionally return a builder for a lint diagnostic guard.
     ///
     /// If the current context believes a diagnostic should be reported for
@@ -396,11 +400,10 @@ pub(super) struct LintDiagnosticGuardBuilder<'db, 'ctx> {
 }
 
 impl<'db, 'ctx> LintDiagnosticGuardBuilder<'db, 'ctx> {
-    fn new(
+    fn severity_and_source(
         ctx: &'ctx InferContext<'db, 'ctx>,
         lint: &'static LintMetadata,
-        range: TextRange,
-    ) -> Option<LintDiagnosticGuardBuilder<'db, 'ctx>> {
+    ) -> Option<(Severity, LintSource)> {
         // The comment below was copied from the original
         // implementation of diagnostic reporting. The code
         // has been refactored, but this still kind of looked
@@ -429,14 +432,25 @@ impl<'db, 'ctx> LintDiagnosticGuardBuilder<'db, 'ctx> {
         if ctx.is_in_multi_inference() {
             return None;
         }
-        let id = DiagnosticId::Lint(lint.name());
+
+        Some((severity, source))
+    }
+
+    fn new(
+        ctx: &'ctx InferContext<'db, 'ctx>,
+        lint: &'static LintMetadata,
+        range: TextRange,
+    ) -> Option<LintDiagnosticGuardBuilder<'db, 'ctx>> {
+        let (severity, source) = Self::severity_and_source(ctx, lint)?;
 
         let suppressions = suppressions(ctx.db(), ctx.file());
+        let lint_id = LintId::of(lint);
         if let Some(suppression) = suppressions.find_suppression(range, lint_id) {
             ctx.diagnostics.borrow_mut().mark_used(suppression.id());
             return None;
         }
 
+        let id = DiagnosticId::Lint(lint.name());
         let primary_span = Span::from(ctx.file()).with_range(range);
         Some(LintDiagnosticGuardBuilder {
             ctx,


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -317,15 +317,30 @@ impl<'db> GenericContext<'db> {
     /// list.
     pub(crate) fn from_base_classes(
         db: &'db dyn Db,
+        definition: impl FnOnce() -> Definition<'db>,
```
dcreager: See the comment below; this is passed in as a callback so that we don't have to query the class's definition unless it actually references legacy typevars in its base class list. Passing the definition in directly was causing some of the salsa query dependency tracking tests to fail.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1458,13 +1460,69 @@ impl<'db> ClassLiteral<'db> {
...
+    /// any typevars bound in its generic context, as well as any typevars mentioned in its base
+    /// class list. (This is used to ensure that classes do not bind or reference typevars from
+    /// enclosing generic contexts.)
+    pub(crate) fn typevars_referenced_in_definition(
+        self,
+        db: &'db dyn Db,
+    ) -> FxOrderSet<BoundTypeVarInstance<'db>> {
+        struct CollectTypeVars<'db> {
+            typevars: RefCell<FxOrderSet<BoundTypeVarInstance<'db>>>,
+            seen_types: RefCell<FxIndexSet<NonAtomicType<'db>>>,
```
AlexWaygood: nit: I think these can both be `FxIndexSet`s, and `FxIndexSet` is a simpler type than `FxOrderSet`. They're both ordered according to insertion order; IIRC the only difference is that `FxOrderSet` is also hashable, but that doesn't seem to be necessary here.

```suggestion
            typevars: RefCell<FxIndexSet<BoundTypeVarInstance<'db>>>,
            seen_types: RefCell<FxIndexSet<NonAtomicType<'db>>>,
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1458,13 +1460,69 @@ impl<'db> ClassLiteral<'db> {
...
+                        walk_non_atomic_type(db, non_atomic_type, self);
+                    }
+                }
+            }
+        }
+
+        let visitor = CollectTypeVars {
+            typevars: RefCell::new(FxOrderSet::default()),
+            seen_types: RefCell::new(FxIndexSet::default()),
+        };
```
AlexWaygood: nit: if you added `#[derive(Default)]` to the `CollectTypeVars` struct, this could just be:

```suggestion
        let visitor = CollectTypeVars::default();
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -854,7 +856,47 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 }
             }
 
-            // (6) Check that a dataclass does not have more than one `KW_ONLY`.
+            let scope = class.body_scope(self.db()).scope(self.db());
+            if let Some(parent) = scope.parent() {
```
AlexWaygood: nit: the level of nesting here is huge. Could everything under this `if let` maybe be factored out into a standalone method/function, so that we could use early-return style to reduce the indentation?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -854,7 +856,47 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                                                .into(),
+                                        )
+                                        .message(
+                                            format_args!(
+                                                "Type variable `{self_typevar_name}` is bound \
+                                                in this enclosing scope",
+                                            ),
+                                        ),
+                                    );
+                                }
```
AlexWaygood: `full_range` here will be a huge range if the class has lots of methods in it. Additionally, it can be a bit hard to tell the two highlighted ranges apart from each other currently:

image

What about something like this?

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 894ef87710..aa52a3c0d1 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8190,6 +8190,13 @@ impl<'db> BindingContext<'db> {
             BindingContext::Synthetic => None,
         }
     }
+
+    fn into_definition(self) -> Option<Definition<'db>> {
+        match self {
+            BindingContext::Definition(definition) => Some(definition),
+            BindingContext::Synthetic => None,
+        }
+    }
 }
 
 /// A type variable that has been bound to a generic context, and which can be specialized to a
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index dcbb4a5118..0ccbb658e6 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -866,28 +866,28 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                         {
                             if let Some(builder) = self
                                 .context
-                                .report_lint(&INVALID_GENERIC_CLASS, &class_node.name)
+                                .report_lint(&INVALID_GENERIC_CLASS, class.header_range(self.db()))
                             {
                                 let mut diagnostic = builder.into_diagnostic(format_args!(
                                     "Generic class `{}` must not reference type variables \
                                     bound in an enclosing scope",
                                     class_node.name,
                                 ));
-                                if let BindingContext::Definition(other_definition) =
-                                    other_typevar.binding_context(self.db())
+                                diagnostic.set_primary_message(format_args!(
+                                    "`{self_typevar_name}` referenced in class definition here"
+                                ));
+                                if let Some(class) = other_typevar
+                                    .binding_context(self.db())
+                                    .into_definition()
+                                    .and_then(|definition| {
+                                        binding_type(self.db(), definition).into_class_literal()
+                                    })
                                 {
                                     diagnostic.annotate(
-                                        Annotation::primary(
-                                            other_definition
-                                                .full_range(self.db(), self.module())
-                                                .into(),
-                                        )
-                                        .message(
-                                            format_args!(
-                                                "Type variable `{self_typevar_name}` is bound \
-                                                in this enclosing scope",
-                                            ),
-                                        ),
+                                        Annotation::secondary(class.header_span(self.db()))
+                                        .message(format_args!(
+                                            "Type variable `{self_typevar_name}` bound in this enclosing scope",
+                                        )),
                                     );
                                 }
                             }
```

which renders like this:

image

crates/ty_python_semantic/resources/mdtest/generics/scoping.md
```diff
@@ -260,27 +260,31 @@ class C[T]:
...
 ```py
 from typing import Iterable
 
 def f[T](x: T, y: T) -> None:
     class Ok[S]: ...
-    # TODO: error for reuse of typevar
+    # error: [invalid-generic-class] "Generic class `Bad1` must not reference type variables bound in an enclosing scope"
     class Bad1[T]: ...
-    # TODO: error for reuse of typevar
+    # error: [invalid-generic-class] "Generic class `Bad2` must not reference type variables bound in an enclosing scope"
```
AlexWaygood: nit: since you've enabled snapshots, I think I'd avoid asserting the precise error messages here. It can be annoying to update loads of mdtest assertions whenever you change an error message

crates/ty_python_semantic/src/types/generics.rs
MichaReiser: That's probably worth an inline comment as it's non-obvious for future readers. 

However, I do think what you did here is cheating the test rather than fixing the underlying issue. The problem is that calling `definition` adds a dependency on the ast node. The way you cheated the test is by early-returning in the case we're testing, a class with no type vars! But the same problem still exists for classes with type variables. 

The more proper fix here is probably to make whatever calls `from_base_classes` to be behind a salsa query (at least, in positions where we can reach this call across file boundaries).

crates/ty_python_semantic/resources/mdtest/generics/scoping.md
dcreager: Done

crates/ty_python_semantic/src/types/class.rs
dcreager: TIL that difference between `IndexSet` and `OrderSet`! Done

crates/ty_python_semantic/src/types/class.rs
dcreager: Done

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: TIL `header_range`, thank you! Done. Similarly, I used the signature span if the enclosing generic context is from a function.

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: Done. I moved this into a new `report_foo` function alongside the others in diagnostic.rs

crates/ty_python_semantic/resources/mdtest/snapshots/scoping.md_-_Scoping_rules_for_ty…_-_Nested_formal_typeva…_-_Generic_class_within…_(3259718bf20b45a2).snap
```diff
@@ -0,0 +1,63 @@
...
+ --> src/mdtest_snippet.py:3:5
+  |
+1 | from typing import Iterable
+2 |
+3 | def f[T](x: T, y: T) -> None:
+  |     ------------------------ Type variable `T` is bound in this enclosing scope
+4 |     class Ok[S]: ...
+5 |     # error: [invalid-generic-class]
+6 |     class Bad1[T]: ...
+  |           ^^^^ `T` referenced in class definition here
```
AlexWaygood: heh, this seems like a pre-existing bug in `ClassLiteral::header_span()` that it doesn't cover the type parameters

crates/ty_python_semantic/src/types/generics.rs
dcreager: > The more proper fix here is probably to make whatever calls `from_base_classes` to be behind a salsa query (at least, in positions where we can reach this call across file boundaries).

That was an easier real fix than I expected! Done

crates/ty_python_semantic/resources/mdtest/snapshots/scoping.md_-_Scoping_rules_for_ty…_-_Nested_formal_typeva…_-_Generic_class_within…_(3259718bf20b45a2).snap
dcreager: Yep I saw that too

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3055,6 +3055,37 @@ pub(crate) fn report_cannot_pop_required_field_on_typed_dict<'db>(
...
+                Type::FunctionLiteral(function) => function.spans(db).map(|spans| spans.signature),
+                _ => None,
+            };
+            if let Some(span) = span {
+                diagnostic.annotate(Annotation::secondary(span).message(format_args!(
+                    "Type variable `{typevar_name}` is bound in this enclosing scope",
+                )));
+            }
+        }
+    }
```
AlexWaygood: you could use early-return style to reduce the indentation here even more, now that it's a standalone function:

```suggestion
    let Some(builder) = context.report_lint(&INVALID_GENERIC_CLASS, class.header_range(db)) else {
        return;
    };
    let mut diagnostic = builder.into_diagnostic(format_args!(
        "Generic class `{}` must not reference type variables bound in an enclosing scope",
        class_node.name,
    ));
    diagnostic.set_primary_message(format_args!(
        "`{typevar_name}` referenced in class definition here"
    ));
    let Some(other_definition) = other_typevar.binding_context(db).definition() else {
        return;
    };
    let span = match binding_type(db, other_definition) {
        Type::ClassLiteral(class) => Some(class.header_span(db)),
        Type::FunctionLiteral(function) => function.spans(db).map(|spans| spans.signature),
        _ => return,
    };
    if let Some(span) = span {
        diagnostic.annotate(Annotation::secondary(span).message(format_args!(
            "Type variable `{typevar_name}` is bound in this enclosing scope",
        )));
    }
```

dcreager: Mucho gusto

AlexWaygood:
> I'm surprised by the performance regression. This doesn't seem to add that much complexity. It might be worth taking a look at what's happening.

It looks like this may have been fixed by adding the Salsa caching, though there is now a memory-usage regression being reported by mypy_primer

dcreager:
> though there is now a memory-usage regression being reported by mypy_primer

That is unsurprising, since we'll have a cached result for the newly cached function for every non-generic class. (And technically for every generic class that is generic via inheritance using legacy typevars) It's not a huge regression, so I would lean towards accepting it.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1452,19 +1458,78 @@ impl<'db> ClassLiteral<'db> {
...
+                    }
+                }
+            }
+        }
+
+        let visitor = CollectTypeVars::default();
+        if let Some(generic_context) = self.generic_context(db) {
+            for bound_typevar in generic_context.variables(db) {
+                visitor.visit_bound_type_var_type(db, bound_typevar);
+            }
```
```suggestion
            walk_generic_context(db, generic_context, &visitor);
```

(requires importing `walk_generic_context` from `generics.rs`)

dcreager: I could've sworn I tried that! Ah well, done

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1452,19 +1460,76 @@ impl<'db> ClassLiteral<'db> {
         })
     }
 
+    #[salsa::tracked(
+        cycle_fn=generic_context_cycle_recover,
+        cycle_initial=generic_context_cycle_initial,
+        heap_size=ruff_memory_usage::heap_size,
+    )]
```
MichaReiser, line 1467: Regarding the memory regression. Would it be possible to only cache the cases where we have a type var? Or is this already what this is doing :)

dcreager: I don't think we can without walking the base class in the same way

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -854,7 +854,27 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-            // (6) Check that a dataclass does not have more than one `KW_ONLY`.
+            let scope = class.body_scope(self.db()).scope(self.db());
+            if let Some(parent) = scope.parent() {
+                for self_typevar in class.typevars_referenced_in_definition(self.db()) {
+                    let self_typevar_name = self_typevar.typevar(self.db()).name(self.db());
+                    for enclosing in enclosing_generic_contexts(self.db(), self.index, parent) {
+                        if let Some(other_typevar) =
+                            enclosing.binds_named_typevar(self.db(), self_typevar_name)
+                        {
+                            report_rebound_typevar(
```
MichaReiser, line 867: Should we move the check whether the rule is enabled further up (to avoid the entire scope walking?)

It might require extracting the following into a `lint_severity(name: &LintName) -> Option<Severity>` helper

https://github.com/astral-sh/ruff/blob/2ce3aba458d0ee6760639485a9857c4c89af63ac/crates/ty_python_semantic/src/types/context.rs#L404-L431

AlexWaygood: Could also experiment with reviving https://github.com/astral-sh/ruff/pull/18327?

dcreager: Done

PR #20825, opened 2025-10-12
[`pyupgrade`] Extend `UP019` to detect `typing_extensions.Text` (`UP019`)

### LINKED ISSUE #20819: UP019 should detect `typing_extensions.Text`
### Summary

[`typing-text-str-alias` (UP019)](https://docs.astral.sh/ruff/rules/typing-text-str-alias/) detects `typing.Text`. It would also be useful for it to detect `typing_extensions.Text`. [Example](https://play.ruff.rs/c9b704ed-1e1e-45aa-8e5a-de32fd3280ac):
```console
$ cat >up019.py <<'# EOF'
from typing_extensions import Text
foo: Text = "bar"
# EOF

$ ruff --isolated check up019.py --select UP019
All checks passed!
```

### Version

danparizher:
## Summary

Fixes #20819

### DIFF
--- a/crates/ruff_linter/src/preview.rs
+++ b/crates/ruff_linter/src/preview.rs
@@ -265,3 +265,7 @@ pub(crate) const fn is_fix_read_whole_file_enabled(settings: &LinterSettings) ->
 pub(crate) const fn is_fix_write_whole_file_enabled(settings: &LinterSettings) -> bool {
     settings.preview.is_enabled()
 }
+
+pub(crate) const fn is_typing_extensions_str_alias_enabled(settings: &LinterSettings) -> bool {
+    settings.preview.is_enabled()
+}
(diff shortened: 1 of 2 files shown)


crates/ruff_linter/src/rules/pyupgrade/snapshots/ruff_linter__rules__pyupgrade__tests__UP019.py.snap
```diff
@@ -66,3 +66,54 @@ help: Replace with `str`
    - def print_fourth_word(word: Goodbye) -> None:
 19 + def print_fourth_word(word: str) -> None:
 20 |     print(word)
+21 |     
+22 | 
+
+UP019 [*] `typing.Text` is deprecated, use `str`
```
ntBre: It might also make sense to record the module we found so that we can report `typing_extensions.Text` here.

PR #20841, opened 2025-10-13
[ty] Do not consider function-like property for `Callable` type relations

### LINKED ISSUE #1342: Type relations for function-like callables are too strict
Assignability, subtyping, and maybe also equivalence are currently too strict for function-like callables. For example, it should be possible to assign `Mock()` to a function-like callable, but we currently get:

```py
from unittest.mock import Mock

@dataclass(order=True)
class Data:
    value: int

def test_t():
    d = Data(1)
    # error: Object of type `Mock` is not assignable to
    # attribute `__lt__` of type `(self: Data, other: Data) -> bool`
    d.__lt__ = Mock()
```
https://play.ty.dev/8148ed96-48fa-4706-9e50-2ff61f0f5a05

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1342

## Ecosystem analysis

This looks good:

```diff
bokeh (https://github.com/bokeh/bokeh)
- src/bokeh/core/templates.py:112:1: error[invalid-assignment] Object of type `dict[str, () -> Unknown]` is not assignable to `dict[str, () -> Template]`
```

## Test Plan

Regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
@@ -1204,9 +1204,9 @@ python-version = "3.12"
 from dataclasses import dataclass
 from typing import Callable
 from types import FunctionType
-from ty_extensions import CallableTypeOf, TypeOf, static_assert, is_subtype_of, is_assignable_to
+from ty_extensions import CallableTypeOf, TypeOf, static_assert, is_subtype_of, is_assignable_to, is_equivalent_to
 
-@dataclass
+@dataclass(order=True)
 class C:
     x: int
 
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -1233,8 +1233,20 @@ static_assert(not is_assignable_to(EquivalentPureCallableType, DunderInitType))
 static_assert(is_subtype_of(DunderInitType, EquivalentFunctionLikeCallableType))
 static_assert(is_assignable_to(DunderInitType, EquivalentFunctionLikeCallableType))
 
-static_assert(not is_subtype_of(EquivalentFunctionLikeCallableType, DunderInitType))
-static_assert(not is_assignable_to(EquivalentFunctionLikeCallableType, DunderInitType))
```
sharkdp, line 1237: To be honest, I don't understand the reasoning behind these tests at all (I wrote them myself).

AlexWaygood: > Are there any important reasons why we should _not_ consider function-like callable types subtypes of equivalent non-function like callable types?

We should consider function-like `Callable`s subtypes of non-function-like `Callable`s. But I don't think we should ever consider non-function-like `Callable`s subtypes of function-like `Callable`s. I think that's what the logic on `main` is _trying_ to accomplish?

Non-function-like `Callable`s cannot be subtypes of function-like `Callable`s, because function-like `Callable`s have attributes from `FunctionType` available to them, such as `__name__`, `__qualname__` and `__kwdefaults__`. These aren't available on non-function-like `Callable`s.

This issue makes me wonder if we should reconsider having the concept of "function-like `Callable`s" in the first place -- an alternative way of representing this concept would just be to use `types.FunctionType & <some Callable type>`

PR #20842, opened 2025-10-13
[ty] Treat functions, methods, and dynamic types as function-like `Callable`s

sharkdp:
## Summary

Treat functions, methods, and dynamic types as function-like `Callable`s

closes https://github.com/astral-sh/ty/issues/1342
closes https://github.com/astral-sh/ty/issues/1344

## Ecosystem analysis

All removed diagnostics look like cases of https://github.com/astral-sh/ty/issues/1344

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
@@ -476,7 +476,7 @@ static_assert(not is_equivalent_to(Callable[[int, str], None], Callable[[int, st
 static_assert(not is_equivalent_to(Callable[..., None], Callable[[], None]))
 ```
 
-A function with no explicit return type should be gradual equivalent to a function-like callable
+A function with no explicit return type should be gradually equivalent to a function-like callable
 with a return type of `Any`.
 
 ```py
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1402,10 +1402,7 @@ impl<'db> Type<'db> {
         match self {
             Type::Callable(_) => Some(self),
 
-            Type::Dynamic(_) => Some(CallableType::single_function_like(
-                db,
-                Signature::dynamic(self),
-            )),
+            Type::Dynamic(_) => Some(CallableType::function_like(db, Signature::dynamic(self))),
 
             Type::FunctionLiteral(function_literal) => {
                 Some(Type::Callable(function_literal.into_callable_type(db)))
@@ -9869,16 +9866,6 @@ impl<'db> CallableType<'db> {
             db,
             CallableSignature::single(signature),
             false,
-        ))
-    }
-
-    /// Create a callable type with a single non-overloaded signature. Mark the callable
-    /// as function-like.
-    pub(crate) fn single_function_like(db: &'db dyn Db, signature: Signature<'db>) -> Type<'db> {
-        Type::Callable(CallableType::new(
-            db,
-            CallableSignature::single(signature),
-            true,
         ))
     }
 


crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
AlexWaygood, line 1: do we really not already have tests like these somewhere...?

crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
```diff
@@ -476,14 +476,17 @@ static_assert(not is_equivalent_to(Callable[[int, str], None], Callable[[int, st
 static_assert(not is_equivalent_to(Callable[..., None], Callable[[], None]))
 ```
 
-A function with no explicit return type should be gradual equivalent to a callable with a return
-type of `Any`.
+A function with no explicit return type should be gradual equivalent to a function-like callable
```
AlexWaygood:
```suggestion
A function with no explicit return type should be gradually equivalent to a function-like callable
```

or

```suggestion
A function with no explicit return type should be gradual-equivalent to a function-like callable
```

crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
```diff
@@ -506,12 +512,27 @@ def variadic_without_annotation(*args, **kwargs):
...
+gradually equivalent to `Callable[..., Any]`, because the latter is not a function-like callable
+type:
+
+```py
+static_assert(not is_equivalent_to(CallableTypeOf[variadic_without_annotation], Callable[..., Any]))
+static_assert(not is_equivalent_to(CallableTypeOf[variadic_with_annotation], Callable[..., Any]))
+```
+
+A function with either `*args` or `**kwargs` (and not both) is is not equivalent to a callable with
+`...` as the parameter type.
```
AlexWaygood, line 535: 👍

crates/ty_python_semantic/src/types.rs
```diff
@@ -9771,7 +9774,7 @@ impl<'db> BoundMethodType<'db> {
                     .iter()
                     .map(|signature| signature.bind_self(db, Some(self_instance))),
             ),
-            false,
+            true,
```
AlexWaygood, line 9774: strictly speaking, if we're doing it this way rather than with intersections, I suppose we should _also_ have a `method_like: bool` flag for `BoundMethodType`s -- methods not only have all the attributes that functions have, but they also have attributes such as `__closure__` that functions do not have

(Not blocking for this PR, though, of course)

crates/ty_python_semantic/src/types.rs
```diff
@@ -9869,6 +9872,16 @@ impl<'db> CallableType<'db> {
...
 
+    /// Create a callable type with a single non-overloaded signature. Mark the callable
+    /// as function-like.
+    pub(crate) fn single_function_like(db: &'db dyn Db, signature: Signature<'db>) -> Type<'db> {
+        Type::Callable(CallableType::new(
+            db,
+            CallableSignature::single(signature),
+            true,
+        ))
+    }
```
AlexWaygood: This looks identical to the `CallableType::function_like` method immediately below it?

crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
sharkdp: :man_shrugging:

crates/ty_python_semantic/src/types.rs
sharkdp: I very much like the `Callable[…] & FunctionType` / `Callable[…] & BoundMethodType` approach, but I'd like to move on with this for now, to get to [what I'm actually trying to achieve](https://github.com/astral-sh/ruff/pull/20802) :smile:

crates/ty_python_semantic/src/types.rs
sharkdp: Oh, haha. Thanks!

crates/ty_python_semantic/src/types.rs
AlexWaygood: yeah ikik, that's why I said this is non-blocking 😄

PR #20843, opened 2025-10-13
[ty] Fix further issues in `super()` inference logic

AlexWaygood:
## Summary

Some small followups to https://github.com/astral-sh/ruff/pull/20814, after studying the latest ecosystem report on https://github.com/astral-sh/ruff/pull/20723 and studying the code some more:
- Allow `type[]` types as the first argument to `super()`: there are several ecosystem projects that do that
- Disallow `GenericAlias` instances as the first argument to `super()`: this fails at runtime
- Add a hint if the user has annotated `self` with a `TypeVar` that doesn't have an upper bound (they probably want to add an upper bound to the `TypeVar`)

## Test Plan

Snapshots and mdtests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/class/super.md
+++ b/crates/ty_python_semantic/resources/mdtest/class/super.md
@@ -127,7 +127,7 @@ def _(x: object, y: SomeTypedDict, z: Callable[[int, str], bool]):
 reveal_type(super(list[int], []))
 ```
 
-`super(pivot_class, owner)` can be done from inside methods, just like single-argument `super()`:
+`super(pivot_class, owner)` can be called from inside methods, just like single-argument `super()`:
 
 ```py
 class Super:


crates/ty_python_semantic/resources/mdtest/class/super.md
```diff
@@ -116,6 +118,26 @@ def _(x: object, y: SomeTypedDict, z: Callable[[int, str], bool]):
...
+
+# The first argument to `super()` must be an actual class object;
+# instances of `GenericAlias` are not accepted at runtime:
+#
+# error: [invalid-super-argument]
+# revealed: Unknown
+reveal_type(super(list[int], []))
+```
+
+`super(pivot_class, owner)` can be done from inside methods, just like single-argument `super()`:
```
AlexWaygood:
```suggestion
`super(pivot_class, owner)` can be called from inside methods, just like single-argument `super()`:
```

Issue #19751, opened 2025-08-05
Panic `f-string: unexpected token TStringMiddle`


qarmin:
Quite similar to https://github.com/astral-sh/ruff/issues/18860

File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
_='''{s''T'{: '''
```

command
```
timeout -v 300 ruff check TEST___FILE.py --select ALL --preview --output-format concise --no-cache --fix --unsafe-fixes --isolated
```

cause this
```
All checks passed!

warning: `incorrect-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible. Ignoring `incorrect-blank-line-before-class`.
warning: `multi-line-summary-first-line` (D212) and `multi-line-summary-second-line` (D213) are incompatible. Ignoring `multi-line-summary-second-line`.
error: Panicked while linting /opt/BROKEN_FILES_DIR/917_IDX_0_RAND_244014174919259028981136_minimized_366.py: This indicates a bug in Ruff. If you could open an issue at:

    https://github.com/astral-sh/ruff/issues/new?title=%5BLinter%20panic%5D

...with the relevant file contents, the `pyproject.toml` settings, and the following stack trace, we'd be very appreciative!

panicked at crates/ruff_python_parser/src/parser/expression.rs:1695:25:
internal error: entered unreachable code: f-string: unexpected token `TStringMiddle` at 12..13
Backtrace:    0: ruff_db::panic::install_hook::{{closure}}::{{closure}}
             at ./ruff-main/crates/ruff_db/src/panic.rs:91:34
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/f34ba774c78ea32b7c40598b8ad23e75cdac42a6/library/alloc/src/boxed.rs:1985:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/f34ba774c78ea32b7c40598b8ad23e75cdac42a6/library/std/src/panicking.rs:841:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/f34ba774c78ea32b7c40598b8ad23e75cdac42a6/library/std/src/panicking.rs:706:13
   4: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/f34ba774c78ea32b7c40598b8ad23e75cdac42a6/library/std/src/sys/backtrace.rs:174:18
[...]

PR #20848, opened 2025-10-13
Improved error recovery for unclosed strings (including f- and t-strings)

MichaReiser:
## Summary

This PR improves our lexer to preserve `STRING` tokens instead of converting them to `Unknown` if a string literal misses its closing quotes. 
Instead of converting to `UNKNOWN`, it sets a flag on the string literal that allows upstream tools to check if it's an unclosed string literal.

The benefit of preserving string literals is that it gives us much better error recovery because the parser now recognizes those literals. 
That means, ty will correctly infer the literal type for `a = "unclosed` to be `Literal["unclosed"]`. 

Unfortunately, preserving the kind for unclosed string literals regressed the f-string's and t-string's recovery mechanism. So, I went ahead and improved that too. 

There are a few improvements:

* Preserve the F-STRING middle even if it's unclosed (e.g. `f"unclosed`) instead of parsing this as `f""` 
* Better recovery for missing `}`. E.g., the parser now matches the quotes for `f"{ab"` instead of assuming that the closing quotes start a new string
* Better recovery for `r` format specifiers if the `}` is missing: `f"{ab:r"` now parses the `r` as the raw conversion flag rather than `r"` the start of a raw string literal

Fixes https://github.com/astral-sh/ruff/issues/19751
Fixes https://github.com/astral-sh/ruff/issues/20849

## Review

You probably want to skip the first commit :) It updates all snapshots to now include the `unclosed: <UNCLOSED>`  flag.

## Test Plan

Reviewed and updated the snapshot tests. I also reviewed all usages of `TokenKind::String` to find cases where the missing closing quote could now cause issues. 

This change should have no impact on AST-based lint rules or the formatter because they both only run when there are no parse errors.

### DIFF
--- a/crates/ruff_python_parser/src/parser/expression.rs
+++ b/crates/ruff_python_parser/src/parser/expression.rs
@@ -1721,13 +1721,6 @@ impl<'src> Parser<'src> {
         let start = self.node_start();
         self.bump(TokenKind::Lbrace);
 
-        // Before newline is incorrect too for multiline strings. We need to
-        // check if the quotes on the first line are only followed by whitespace
-
-        // TODO: What about multiline? Should we move the assertion that it's before a newline here?
-        // The check in `re_lex_string_token_in_interpolation_element` whether it's positioned
-        // at a newline is also incorrect if the string is multiline as it doesn't check
-        // the end of the first line.
         self.tokens
             .re_lex_string_token_in_interpolation_element(string_kind);
 
@@ -1786,6 +1779,8 @@ impl<'src> Parser<'src> {
         };
 
         let conversion = if self.eat(TokenKind::Exclamation) {
+            // Ensure that the `r` is lexed as a `r` name token instead of a raw string
+            // in `f{abc!r"` (note the missing `}`).
             self.tokens.re_lex_raw_string_in_format_spec();
 
             let conversion_flag_range = self.current_token_range();
--- a/crates/ruff_python_parser/tests/fixtures.rs
+++ b/crates/ruff_python_parser/tests/fixtures.rs
@@ -276,8 +276,10 @@ fn extract_options(source: &str) -> Option<ParseOptions> {
 #[ignore]
 #[expect(clippy::print_stdout)]
 fn parser_quick_test() {
-    let source = r#"
-f"{""#;
+    let source = "\
+f'{'
+f'{foo!r'
+";
 
     let parsed = parse_unchecked(source, ParseOptions::from(Mode::Module));
 


crates/ruff_linter/resources/test/fixtures/flake8_implicit_str_concat/ISC_syntax_error.py
```diff
@@ -1,10 +1,10 @@
-# The lexer doesn't emit a string token if it's unterminated
+# The lexer emits a string token if it's unterminated
 "a" "b
 "a" "b" "c
 "a" """b
 c""" "d
 
-# For f-strings, the `FStringRanges` won't contain the range for
+# This is also true for
 # unterminated f-strings.
```
MichaReiser, line 8: I know this looks silly, but keeping the comment over two lines reduces the snapshot changes.

crates/ruff_python_parser/src/token.rs
```diff
@@ -729,7 +729,7 @@ impl fmt::Display for TokenKind {
 
 bitflags! {
     #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-    pub(crate) struct TokenFlags: u8 {
+    pub(crate) struct TokenFlags: u16 {
```
MichaReiser, line 732: While this increases the size of `TokenFlags`, it doesn't increase the size of `Token`. Which is why I didn't bother with any fancy encoding (e.g. it's unclosed if `RAW_STRING_UPPERCASE` and `RAW_STRING_LOWERCASE` are set)

crates/ruff_python_parser/src/lexer.rs
```diff
@@ -2739,6 +2876,164 @@ t"{(lambda x:{x})}"
         }
     }
 
+    #[test]
+    fn lex_fstring_unclsoed() {
```
AlexWaygood:
```suggestion
    fn lex_fstring_unclosed() {
```

MichaReiser: Ah come on, my spelling is obviously better

crates/ruff_python_parser/src/lexer.rs
```diff
@@ -995,7 +995,7 @@ impl<'src> Lexer<'src> {
                     self.current_flags |= TokenFlags::UNCLOSED_STRING;
 
                     self.push_error(LexicalError::new(
-                        LexicalErrorType::StringError,
```
MichaReiser, line 982: This error was just wrong. `a = "string<EOF` reported an unexpected string error rather than unclosed string literal

PR #20850, opened 2025-10-13
[ty] Do not bind self to non-positional parameters

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1333

## Test Plan

Regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/self.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/self.md
@@ -473,4 +473,18 @@ class C:
 reveal_type(generic_context(C.f))  # revealed: None
 ```
 
+## Non-positional first parameters
+
+This makes sure that we don't bind `self` if it's not a positional parameter:
+
+```py
+from ty_extensions import CallableTypeOf
+
+class C:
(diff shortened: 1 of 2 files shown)


AlexWaygood:
> ## `mypy_primer` results
> No ecosystem changes detected ✅ No memory usage changes detected ✅

Well, it still makes sense as a standalone change!

sharkdp:
> > No ecosystem changes detected ✅ No memory usage changes detected ✅
> 
> Well, it still makes sense as a standalone change!

I didn't quite understand why you thought most of the changes came from the fix, but I agree that it can be a standalone PR ;-)

(the fix *was* necessary in order to remove 1139 diagnostics that would otherwise be caused in combination with https://github.com/astral-sh/ruff/pull/20802)

AlexWaygood:
> I didn't quite understand why you thought most of the changes came from the fix

Yes, I obviously misunderstood. Sorry!

sharkdp:
> > I didn't quite understand why you thought most of the changes came from the fix
> 
> Yes, I obviously misunderstood. Sorry!

No worries, good suggestion nevertheless

PR #20855, opened 2025-10-14
[ty] Support implicit imports of submodules in `__init__.pyi`

Gankra:
This is a second take at the implicit imports approach, allowing `from . import submodule` in an `__init__.pyi` to create the `mypackage.submodule` attribute everyhere.

This implementation operates inside of the available_submodule_attributes subsystem instead of as a re-export rule.

The upside of this is we are no longer purely syntactic, and absolute from imports that happen to target submodules work (an intentional discussed deviation from pyright which demands a relative from import). Also we don't re-export functions or classes.

The downside(?) of this is star imports no longer see these attributes (this may be either good or bad. I believe it's not a huge lift to make it work with star imports but it's some non-trivial reworking).

I've also intentionally made `import mypackage.submodule` not trigger this rule although it's trivial to change that.

I've tried to cover as many relevant cases as possible for discussion in the new test file I've added (there are some random overlaps with existing tests but trying to add them piecemeal felt confusing and weird, so I just made a dedicated file for this extension to the rules).

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
@@ -570,7 +570,8 @@ reveal_type(Z)  # revealed: int
 ## `from` Import of Non-Submodule
 
 A from import that terminates in a non-submodule should not expose the intermediate submodules as
-attributes.
+attributes. This is an arbitrary decision but on balance probably safe and correct, as otherwise it
+would be hard for a stub author to be intentional about the submodules being exposed as attributes.
 
 `mypackage/__init__.pyi`:
 
--- a/crates/ty_python_semantic/src/semantic_index.rs
+++ b/crates/ty_python_semantic/src/semantic_index.rs
@@ -285,7 +285,7 @@ pub(crate) struct SemanticIndex<'db> {
     imported_modules: Arc<FxHashSet<ModuleName>>,
 
     /// `from...import` statements within this file that might import a submodule.
-    maybe_imported_modules: Arc<FxHashSet<MaybeModuleImport>>,
+    maybe_imported_modules: FxHashSet<MaybeModuleImport>,
 
     /// Flags about the global scope (code usage impacting inference)
     has_future_annotations: bool,
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -1266,6 +1266,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
         self.scopes_by_node.shrink_to_fit();
         self.generator_functions.shrink_to_fit();
         self.enclosing_snapshots.shrink_to_fit();
+        self.maybe_imported_modules.shrink_to_fit();
 
         SemanticIndex {
             place_tables,
@@ -1278,7 +1279,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
             scopes_by_node: self.scopes_by_node,
             use_def_maps,
             imported_modules: Arc::new(self.imported_modules),
-            maybe_imported_modules: Arc::new(self.maybe_imported_modules),
+            maybe_imported_modules: self.maybe_imported_modules,
             has_future_annotations: self.has_future_annotations,
             enclosing_snapshots: self.enclosing_snapshots,
             semantic_syntax_errors: self.semantic_syntax_errors.into_inner(),


crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -230,6 +274,9 @@ pub(crate) struct SemanticIndex<'db> {
     /// The set of modules that are imported anywhere within this file.
     imported_modules: Arc<FxHashSet<ModuleName>>,
 
+    /// The set of modules that are imported anywhere within this file.
+    maybe_imported_modules: Arc<FxHashSet<(u32, Option<String>, String)>>,
+
```
MichaReiser: How is this different from `imported_modules`? 

I also suggest using a named struct here. It's non obvious what the semantic meaning of the tuple fields are.


It might be worth using `Name` instead of `String`, to reduce the amount of heap allocations (or use `ModuleName`?)

crates/ty_python_semantic/src/types.rs
```diff
@@ -9867,6 +9870,14 @@ impl<'db> ModuleLiteralType<'db> {
             .into_iter()
             .flat_map(|file| imported_modules(db, file))
             .filter_map(|submodule_name| submodule_name.relative_to(self.module(db).name(db)))
+            .chain(
+                maybe_imported_relative_submodules_of_stub_package(db, self.module(db))
+                    .iter()
+                    // Throw out the result if the import is not a module (e.g. it's a function/class)
+                    // So that `from a.b import myfunction` is not considered an import of `a.b`
+                    .filter(|submodule| self.resolve_submodule(db, submodule).is_some())
```
MichaReiser: Should we move this check into `maybe_imported_relative_submodules_of_stub_package`, given that `maybe_imported_relative_submodules_of_stub_package` is cached? Is it worth caching `maybe_imported_relative_submodules_of_stub_package`?

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+
+This document covers ty-specific extensions to the
+[standard import conventions](https://typing.python.org/en/latest/spec/distributing.html#import-conventions).
+
+It's a common idiom for a package's `__init__.py(i)` to include several imports like
+`from . import mysubmodule`, with the intent that the `mypackage.mysubmodule` attribute should work
+for anyone who only imports `mypackage`.
+
+In the context of a `.py` we handle this well through our general attempts to faithfully implement
+import side-effects. However for `.pyi` files we are expected to apply
```
AlexWaygood, line 11: nit: it's not really a side effect! Just the direct impact of the import system working as designed: importing any object `X` (whether that object is a class, function, module, or anything else) into a module `M` stores `X` in the global namespace of `M` and makes it available as an attribute on `M` when `M` is imported.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+- Extensions should ideally be kept narrow to continue to encourage explicit API design
+- Extensions should be easy to explain, document, and understand
+- Extensions should ideally still be a subset of runtime behaviour (if it works in a stub, it works
+    at runtime)
+- Extensions should ideally not make `.pyi` files more permissive than `.py` files (if it works in a
+    stub, it works in an impl)
+
+To that end we define the following extension:
+
+> If an `__init__.pyi` for `mypackage` contains a `from...import` targetting a submodule, then that
```
AlexWaygood:
```suggestion
> If an `__init__.pyi` for `mypackage` contains a `from...import` targeting a submodule, then that
```

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+> submodule should be available as an attribute of `mypackage`.
+
+## Relative From Import of Direct Submodule in Init
```
AlexWaygood: nit: I'd spell these as

```suggestion
## Relative `from` Import of Direct Submodule in `__init__`
```

throughout

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+```
+
+`main.py`:
+
+```py
+import mypackage
+
+# error: "has no attribute `imported`"
+reveal_type(mypackage.imported.X)  # revealed: Unknown
+```
```
AlexWaygood: I think we _should_ try to support this, at least in non-stub files (but possibly not in stubs). This is exactly what the original examples in https://github.com/astral-sh/ty/issues/133 were about, before that issue got kinda hijacked by examples that were really about exemptions we should carve out for the stub file re-export convention 😄

We don't need to add support for this in this PR, but it might be good to add a TODO here, I think?

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Star Import Unaffected
+
+Even if the `__init__` idiom is in effect, star imports do not pick it up.
```
AlexWaygood: Maybe add "But we could consider improving our handling in the future", or something, to make it clear that there's no reason they _shouldn't_, this just reflects our current implementation?

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+```
+
+`main.py`:
+
+```py
+import mypackage
+
+# error: "has no attribute `imported`"
+reveal_type(mypackage.imported.X)  # revealed: Unknown
+```
```
AlexWaygood: Again, I think it would be good to at least support this for non-stub files eventually; it's probably worth adding a TODO here to make it clear that this isn't the desired end state

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,426 @@
...
+
+## From Import of Other Package's Submodule
+
+`from mypackage import submodule` from outside the package is not modeled as a side-effect on
+`mypackage`, even in the importing file.
```
AlexWaygood: hmmmm, we could consider changing this in the future, too

crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -89,6 +89,50 @@ pub(crate) fn imported_modules<'db>(db: &'db dyn Db, file: File) -> Arc<FxHashSe
...
+/// This set only considers `from...import` statements (but it could also include `import`).
+/// It also only returns a non-empty result for `__init__.pyi` files.
+///
+/// This function specifically implements the rule that if an `__init__.pyi` file
+/// contains a `from...import` that imports a relative submodule of the package,
+/// that relative submodule should be available as an attribute of the package.
+///
+/// While we endeavour to accurately model import side-effects for `.py` files, we intentionally
+/// limit them for `.pyi` files to encourage more intentional API design. The standard escape
+/// hatch for this is the `import x as x` idiom, but in practice some other idioms are popular.
```
AlexWaygood: Explicitly listing the exported symbols in `__all__` is another standard escape hatch here too

crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -89,6 +89,50 @@ pub(crate) fn imported_modules<'db>(db: &'db dyn Db, file: File) -> Arc<FxHashSe
...
+pub(crate) fn maybe_imported_relative_submodules_of_stub_package<'db>(
+    db: &'db dyn Db,
+    importing_module: Module<'db>,
+) -> Vec<ModuleName> {
+    let Some(file) = importing_module.file(db) else {
+        return Vec::new();
+    };
+    if !file.is_package_stub(db) {
+        return Vec::new();
+    }
```
AlexWaygood: nit: we usually return `Box<[]>` for Salsa queries like this, just to store a tiny bit less data in the Salsa cache (and since it's immutable anyway, so a `Vec` doesn't really get you anything)

```suggestion
) -> Box<[ModuleName]> {
    let Some(file) = importing_module.file(db) else {
        return Box::default();
    };
    if !file.is_package_stub(db) {
        return Box::default();
    }
```

Gankra:
Analysis of ecosystem failures:

# croniter 

This case would be fixed by introducing a rule that non-matching asname aliases disable the submodule init rule. I briefly discussed this with Alex and he said something to the effect of "well adding the submodule even if there's an alias is the runtime behaviour" and then I forgot to add a test for it mentioning it as a distinct case.

**EDIT: I pushed this change with tests**

Failure:

```text
paasta (https://github.com/yelp/paasta)
paasta_tools/cli/cmds/validate.py
[error] call-non-callable - :890:12 - Object of type `<module 'croniter.croniter'>` is not callable 
```

In [validate.py](https://github.com/yelp/paasta/blob/8c21e65cf7fd7581e6bef492fe9910c0bf5cc1ea/paasta_tools/cli/cmds/validate.py#L890):



In [typeshed's `croniter/__init__.pyi`](https://github.com/python/typeshed/blob/main/stubs/croniter/croniter/__init__.pyi) we
have the `croniter.croniter` module now masking the `croniter` class in the `croniter` package:

```py
from croniter import croniter
...
iter = croniter(cron_schedule, starting_from)
```

```py
# import that we now consider as introducing the submodule attr
from . import croniter as croniter_m
# explicit re-export of the class (that now gets shadowed by our "submodule attrs win ties" rule)
from .croniter import croniter as croniter
```


# scikit

This case is harder for me to wrap my head around, largely because I can't give a coherent explanation for why it
worked before. As in, I was under the impression that `from xyz import lobpcg` was not a valid re-export of the lobpcg
function, and that listing it in `__all__` would only affect star imports (of which there appear to be none).

Failure:

```
scikit-learn (https://github.com/scikit-learn/scikit-learn)
sklearn/manifold/_spectral_embedding.py
[error] call-non-callable - :415:28 - Object of type `<module 'scipy.sparse.linalg._eigen.lobpcg'>` is not callable
[error] call-non-callable - :449:32 - Object of type `<module 'scipy.sparse.linalg._eigen.lobpcg'>` is not callable
```

In `_spectral_embedding.py`:

```py
from scipy.sparse.linalg import lobpcg
...
_, diffusion_map = lobpcg(laplacian, X, M=M, tol=tol, largest=False)
...
```

In [scipy-stubs' `sparse/linalg/__init__.pyi`](https://github.com/scipy/scipy-stubs/blob/master/scipy-stubs/sparse/linalg/__init__.pyi):

```py
# im
from ._eigen import lobpcg

__all__ = [
    ...
    "lobpcg",
    ...
]
```

In [scipy-stubs' `sparse/linalg/_eigen/__init__.pyi` ](https://github.com/scipy/scipy-stubs/blob/master/scipy-stubs/sparse/linalg/_eigen/__init__.pyi)

```py
# importing the lobpcg function from the submodule of the same name
from .lobpcg import lobpcg

__all__ = [..., "lobpcg", ...]
```

AlexWaygood:
> As in, I was under the impression that `from xyz import lobpcg` was not a valid re-export of the lobpcg
> function, and that listing it in `__all__` would only affect star imports (of which there appear to be none).

Ah, no, `__all__` does also impact which symbols are considered explicitly re-exported from a stub file as well as impacting `*` imports. If you have an imported symbol in a stub file that does not have a redundant alias but _is_ listed in `__all__`, that still counts as an explicit re-export of that symbol

Gankra:
Ah ok I had a nagging feeling you *did* say that but I was like "no Aria you always want `__all__` to do more than it does". 

Ok then my guess is that the issue is `from ._eigen import lobpcg` (in [`linalg/__init__.pyi`](https://github.com/scipy/scipy-stubs/blob/master/scipy-stubs/sparse/linalg/__init__.pyi) is an unfortunate syntactic ambiguity here. 

It is simultaneously a valid import of the `scipy.sparse.linalg._eigen.lobpcg` *module* and of the *function* it re-exports. And presumably tie-break logic kicks in and we are sad? This is an extra fun case because it's even in the syntactic form pyright wants!

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
Gankra: That sounds like a side-effect to me? As in, state has been mutated.

crates/ty_python_semantic/src/types.rs
Gankra: Yes good call, and done!

Gankra: (This fixed most of the perf regressions)

crates/ty_python_semantic/src/semantic_index.rs
Gankra: I originally had this as ModuleName but it created some odd error-handling contortions in the code that get undone anyway because some of the APIs we invoke take a `str` and then themselves parse it as a ModuleName.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
AlexWaygood: But that mutation of state is the _primary_ thing an `import` statement in Python does, so I don't see it as a _side_ effect 😄 it's the intended purpose, not something that happens by accident as a result of an operation that's intended to do something else

Gankra:
I believe the re-export version of this PR avoids the linalg issue by virtue of it not adding anything to the special set of submodule attributes that gets higher priority (not necessarily an argument for it being a better approach, just noting it).

Gankra:
I've put up a reduced test case for the scipy issue (as reduced as I could get it).

Also I think addressed all review comments.

crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -230,6 +280,9 @@ pub(crate) struct SemanticIndex<'db> {
     /// The set of modules that are imported anywhere within this file.
     imported_modules: Arc<FxHashSet<ModuleName>>,
 
+    /// The set of modules that are imported anywhere within this file.
+    maybe_imported_modules: Arc<FxHashSet<MaybeModuleImport>>,
```
MichaReiser: Could you update the comment so that it explains how this field is different from `imported_modules`? 

I assume we can't really unify the two (I'm asking because I want to avoid that we store redundant information in the semantic index, which is one of the main contributor to ty's memory usage)

MichaReiser: I don't think this needs to be an `Arc`, as we have no `#[salsa::tracked]` query that returns the maybe imported modules directly

```suggestion
    maybe_imported_modules: FxHashSet<MaybeModuleImport>,
```

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1278,6 +1282,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
             scopes_by_node: self.scopes_by_node,
             use_def_maps,
             imported_modules: Arc::new(self.imported_modules),
+            maybe_imported_modules: Arc::new(self.maybe_imported_modules),
```
MichaReiser: Let's call `shrink_to_fit` before passing the modules to `SemanticIndex`

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1562,6 +1567,15 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                         (&alias.name.id, false)
                     };
 
+                    // If there's no alias or a redundant alias, record this as a potential import of a submodule
+                    if alias.asname.is_none() || is_reexported {
```
MichaReiser, line 1568: Would it be sufficient to only collect the import statements at the module level (skip if we're inside a function or class)? In which case, I even wonder if this has to be inside `semantic_index` or if it could just be its own salsa query that traverses the top-level statements of a module without traversing into inner blocks.

crates/ty_python_semantic/src/semantic_index.rs
Gankra: Whether we can unify the two I believe hinges on whether we actually want `from...import` and `import` to ever be treated differently for the purposes of import/re-export idioms.

Gankra: Actually I think I could probably unify them still...

crates/ty_python_semantic/src/semantic_index/builder.rs
Gankra: I was going off of the fact that we apply no such logic when handling `imported_modules`. I agree anything but top-level feels a biiit dubious (and I believe pyright agrees with that).

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
Gankra: Agreed that supporting this would be pretty natural, although I'm a bit nervous about making this extension/idiom "too broad" and undermining the whole idea of re-exports having to be explicit.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
Gankra: Given the rest of our handling of non-stubs I agree that this should work for non-stubs (and *only* non-stubs).
(Definitely out of scope for this PR though)

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
Gankra: Agreed, I think fixing this could be in-scope for this PR without too much work.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
AlexWaygood: > Agreed that supporting this would be pretty natural, although I'm a bit nervous about making this extension/idiom "too broad" and undermining the whole idea of re-exports having to be explicit.

yeah, I do agree with that hesitancy for stub files!

For runtime files, though, I think we should just model the runtime semantics as closely as is reasonable. Failing to be explicit about re-exports in a runtime file is at best something a linter should be concerned with, not a type checker, arguably. In runtime files, a type checker should focus on things that are actually going to cause runtime errors (or are unsound in some type-theoretic way). Things that are in poor taste because they rely on implicit behaviour don't really fall into that bucket IMO.

Gankra: I should add some tests for imports at non-global scope (inside functions), which, are a more muddy situation.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
AlexWaygood: I think this might reveal an interesting difference in perspective that stems from how dynamic Python is (so much happens at runtime; so little happens at compile time compared to, e.g., Rust).

In Python, I think the statement `from x import y` in a module `z` is best conceptualised as an explicit command to the interpreter that says "At runtime, at this point in the global namespace, lookup the symbol `y` in the `x` module/package and store a reference to it in the global namespace of `z`". All other consequences of the `import` statement, such as it then being available for use elsewhere in the module `z`, or being available as an attribute on the module `z`, are downstream from that. `from x import y` just desugars to something like this, which is executed at runtime just like all other statements or expressions in the module's global namespace:

```py
try:
    y = __import__("x.y").y
except ImportError:  # not a submodule; look it up as an attribute instead
    y = __import__("x").y
```

If `y` is loaded into the global namespace of an `z/__init__.py` file, it will be available from all code in that file and as an attribute on the `z` package, just like any other symbol that's loaded into the global namespace of that file. So I don't really view the state mutation here as fundamentally different to what happens with any other import in Python, or as a "side effect" of the import system -- this is just how imports work in Python, and it's the _primary_ effect of them IMO

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,607 @@
...
+X: int = 42
+```
+
+`main.py`:
+
+```py
+import mypackage
+
+# error: "has no attribute `imported`"
+reveal_type(mypackage.imported.X)  # revealed: Unknown
```
AlexWaygood: I still think it would be great to add a TODO comment here (https://github.com/astral-sh/ruff/pull/20855#discussion_r2429085956)

Since this is what the original examples in github.com/astral-sh/ty/issues/133 were about, I think this PR arguably doesn't actually close that issue -- rather than being about implicit imports of submodules in `__init__.py` files (what that issue was originally about), to my mind this PR is really about carving out exemptions to the explicit re-export convention for stub files in `__init__.pyi` files. It's still a good and important change for us to make! But something slightly different to what github.com/astral-sh/ty/issues/133 was originally about.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,607 @@
...
```
AlexWaygood: I still think a TODO comment here would be great to make it clear that this isn't asserting our intended behaviour, just capturing our current behaviour (https://github.com/astral-sh/ruff/pull/20855#discussion_r2429093216)

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,607 @@
...
+
+`main.py`:
+
+```py
+import mypackage
+from mypackage import imported
+
+reveal_type(imported.X)  # revealed: int
+# error: "has no attribute `imported`"
+reveal_type(mypackage.imported.X)  # revealed: Unknown
```
AlexWaygood: I know you signalled this in the prose description, but I think a TODO comment next to this assertion would also be great -- it can be confusing for people working on features if existing "assertions" start "breaking" when they improve semantics. Having a TODO comment right next to the assertion is really helpful for that.

MichaReiser:
What's the status of this PR? Are you waiting on any more input/feedback? Is it blocked on anything?

Gankra:
Nice, the optimization I wanted to try ended up actually being a fix for the `from .x import x` issue by virtue of it dropping support for that form altogether. Now only `from . import x` or `from mypackage import x` actually work. I noticed in practice that the users of this idiom are in fact only `from . import x` and not the other form, so, this works well!

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -0,0 +1,823 @@
...
+from mypackage import *
+
+reveal_type(imported.X)  # revealed: int
+reveal_type(Z)  # revealed: int
+```
+
+## `from` Import of Non-Submodule
+
+A from import that terminates in a non-submodule should not expose the intermediate submodules as
+attributes.
```
AlexWaygood: (this could also be changed in the future, as discussed, but no need to do that in this PR of course!)

crates/ty_python_semantic/src/semantic_index/builder.rs
AlexWaygood: this looks like it's still unaddressed

PR #20856, opened 2025-10-14
[ty] Skip eagerly evaluated scopes for attribute storing

Glyphack:
## Summary

Fix https://github.com/astral-sh/ty/issues/664

This PR adds support for storing attributes in comprehension scopes (any eager scope.)

For example in the following code we infer type of `z` correctly:

```py
class C:
    def __init__(self):
        [None for self.z in range(1)]
reveal_type(C().z) # previously [unresolved-attribute] but now shows Unknown | int
```

The fix works by adjusting the following logics:

To identify if an attriute is an assignment to self or cls we need to check the scope is a method. To allow comprehension scopes here we skip any eager scope in the check.
Also at this stage the code checks if self or the first method argument is shadowed by another binding that eager scope to prevent this:

```py
class D:
    g: int

class C:
    def __init__(self):
        [[None for self.g in range(1)] for self in [D()]]
reveal_type(C().g) # [unresolved-attribute]
```

When determining scopes that attributes might be defined after collecting all the methods of the class the code also returns any decendant scope that is eager and only has eager parents until the method scope.

When checking reachability of a attribute definition if the attribute is defined in an eager scope we use the reachability of the first non eager scope which must be a method. This allows attributes to be marked as reachable and be seen.

There are also which I didn't add support for:

```py
class C:
    def __init__(self):
        def f():
            [None for self.z in range(1)]
        f()

reveal_type(C().z) # [unresolved-attribute]
```

In the above example we will not even return the comprehension scope as an attribute scope because there is a non eager scope (`f` function) between the comprehension and the `__init__` method


crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -184,29 +184,34 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
         self.current_scope_info().file_scope_id
     }
 
-    /// Returns the scope ID of the surrounding class body scope if the current scope
+    /// Returns the scope ID of the method scope if the current scope
     /// is a method inside a class body. Returns `None` otherwise, e.g. if the current
     /// scope is a function body outside of a class, or if the current scope is not a
     /// function body.
     fn is_method_of_class(&self) -> Option<FileScopeId> {
```
Glyphack: This function is also used at https://github.com/Glyphack/ruff/blob/8d53802bc7e9566a9087fd3e3a460345e40a7e6d/crates/ty_python_semantic/src/semantic_index/builder.rs#L1681 and I think this new behavior should be there as well.
I'm not happy with the name though, this now returns either method or an eager child scope of the class.

Glyphack:
There is no ecosystem change detected by mypy primer? With the amount of custom logic added I expected to have more effect. Do you think I should simplify the logic even more since there's not much impact?

Other solution could be only support children eager scopes and don't go into decendants. This would turn a lot of loops into if statements.

sharkdp:
> There is no ecosystem change detected by mypy primer? With the amount of custom logic added I expected to have more effect. Do you think I should simplify the logic even more since there's not much impact?

I would expect the impact to be bigger when combined with type-of-`self`-in-method bodies, no? Maybe we could test that opening a draft PR that merges both. Not strictly necessary though.

> Other solution could be only support children eager scopes and don't go into decendants. This would turn a lot of loops into if statements.

I don't really see a reason not to support that if you already implemented it correctly.

sharkdp:
Thank you for working on this? Could you please also add some tests with generic methods like described [here](https://github.com/astral-sh/ty/issues/664#issuecomment-3380666555)? I'd like to see that we correctly skip the type params scope when detecting methods in classes.

Glyphack:
Thanks. I didn't think of the impact there, I was still hoping to reduce the false positives like [this](https://play.ty.dev/5f89a8f7-2d3e-456e-82e5-c6be2c9fbe62). So let's keep the current implementation.

I went over the tests again added more examples of what is supported.

sharkdp:
I merged this branch into the type-of-self PR and compared it with the type-of-self PR alone. This seems to remove 4 diagnostics from the ecosystem:

type-of-self:

image

type-of-self + this:

image

I did this to evaluate whether I need to prioritize this, or if it can land separately afterwards. After seeing these results, I'm going with the latter.

But we'll come back to this eventually.

carljm:
Just rebased this (and updated some tests) to get an updated ecosystem report.

crates/ty_python_semantic/resources/mdtest/attributes.md
```diff
@@ -401,35 +396,75 @@ class D:
...
+c_instance = C()
+
+# This attribute is in the function f and is not reachable
+# error: [unresolved-attribute]
+reveal_type(c_instance.a)  # revealed: Unknown
+
+# TODO: Even though g method is called and is reachable we do not record this attribute assignment
+# error: [unresolved-attribute]
+reveal_type(c_instance.b)  # revealed: Unknown
+```
```
carljm, line 454: It's not totally clear to me if this is the right behavior or not. We can't guarantee that these attributes will ever be initialized. We can definitely say that they _might_ be.

For reference, this behavior [matches pyright](https://pyright-play.net/?pyrightVersion=1.1.405&strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMogCmAboQIYA2A%2BvAoQFD0DGFZAzm1AMIBc9UAqABNCwKFSqokMCQAo2hCsACUfQeuGiowWav4aDAYiiEQ4EDygBtAK4oibMBVJCAtGRgwQSAEY2YhAC6%2BgbqVgB0kdq4UApK4WSYKNYAjIHBIRoiYmi6aqHqxqbmlrb2hI7OhG4eXr7%2BQZkFEVHAMXHA4T5JqemMBbnKjESklDSIhLJcuglDI%2BTUtJPTyl1zJAvjdFMzTMpAA). [Mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=fe330769b4d8a148d2c801648eca02fb) finds both attributes and gives them type `Any`. [Pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN2UYANxiooAfSbEYAHXRyAxlFRw4dAMKI5dHXUwwwdceNYQGxgBRwYUMAEot6Xc70G6YCw%2B0ufAYjowlNSUiHQA2gCu6IJwuFAimAC0qAwMlBDYEQwwALrePs5hhMXuvHTWtoSonE5hAIw5efJOBfqGbJ6OBS7%2BgcGhkdHwcQnJqemZ2U3dhcWEpZTlNmCE2DXhDU35Lh12cnKCImKSzDAW6p5Ve0NHElJnF3ar14eid6fnlwrXIAA0IFloHASORECB-ABVBjQMykdxRBTQ3DoOD7LBuMC8GgpcToCI0bCBCz4UKsBh2OiJAB85TSXV0ggYEUoTjAMhAADl8YSQnRgPgAL7suR-EBkQRgKCkQgMWhQCj%2BAAKpAlUvKGBwBDoCmRkDYzJSEGRhDk-gAyjAYHQABapYhwRAAekd4oMUsIvDYjpg6EdmFwCjgjp16D1BqRvoWdFQQlQ0FQ2Fg2t1EH1lENyLouGIEeBcjIDGtyMSIkocCNTgAvHR2QBmQh1ABMwvQIAF-1QiIgIgAYtAYBQ0Fg8EQyG2gA) finds both attributes and gives them type `int`. No type checker attempts to distinguish these cases based on trying to determine whether the nested function is ever called; this is quite complex to do in general and I doubt we should attempt it either.

Overall I think the current behavior in this PR is fine and this case is probably rare (and it doesn't show up in the ecosystem report.) I don't think we need TODOs to revisit it; we'll revisit it only if there are real user reports.

crates/ty_python_semantic/resources/mdtest/attributes.md
```diff
@@ -401,35 +396,75 @@ class D:
...
+            [... for self.b in [1]]
+        g()
+
+c_instance = C()
+
+# This attribute is in the function f and is not reachable
+# error: [unresolved-attribute]
+reveal_type(c_instance.a)  # revealed: Unknown
+
+# TODO: Even though g method is called and is reachable we do not record this attribute assignment
```
carljm: I don't think we need a TODO specifically here, because I don't think we will ever attempt to distinguish `f()` / `self.a` from `g()` / `self.b` here, based on call-graph analysis.
```suggestion
```

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -188,29 +188,34 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
         self.current_scope_info().file_scope_id
     }
 
-    /// Returns the scope ID of the surrounding class body scope if the current scope
-    /// is a method inside a class body. Returns `None` otherwise, e.g. if the current
-    /// scope is a function body outside of a class, or if the current scope is not a
+    /// Returns the scope ID of the method scope if the current scope
+    /// is a method inside a class body or current scope is in eagerly executed scope in a method.
```
carljm:
```suggestion
    /// Returns the scope ID of the current scope if the current scope
    /// is a method inside a class body or an eagerly executed scope inside a method.
```

crates/ty_python_semantic/src/semantic_index.rs
```diff
@@ -1195,6 +1225,124 @@ def f(a: str, /, b: str, c: int = 1, *args, d: int = 2, **kwargs):
         }
     }
 
+    /// Test case to validate that comprehensions inside a method are correctly marked as attribute
+    /// scopes
+    #[test]
+    fn comprehension_in_method_instance_attribute() {
```
carljm: It's not clear to me that this test and the below test couldn't be equally effective as mdtests (and/or aren't already effectively duplicated by mdtests in this PR.) But these seem harder to maintain. Do you feel like they are important?

Glyphack: Applied.
I had to change `semantic_index.rs` to track attributes and adjust the reachability analysis in `class.rs`. During development I used this test to ensure the `semantic_index` logic was working fine and then updated the other file at that point I didn't know how to validate it with an mdtest. So now that we have the full implementation we can get rid of this.

crates/ty_python_semantic/resources/mdtest/attributes.md
Glyphack: I haven't worked with call graph analysis and didn't think about it's complexity. If an issue was reported for this I would follow it up.

carljm:
Thank you!

PR #20858, opened 2025-10-14
[ty] Handle decorators which return unions of `Callable`s

sharkdp:
## Summary

If a function is decorated with a decorator that returns a union of `Callable`s, also treat it as a union of function-like `Callable`s.

Labeling as `internal`, since the previous change has not been released yet.

## Test Plan

New regression test.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -2211,6 +2211,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                             ))),
                             Type::Union(union) => union
                                 .try_map(db, |element| into_function_like_callable(db, *element)),
+                            // Intersections are currently not handled here because that would require
+                            // the decorator to be explicitly annotated as returning an intersection.
                             _ => None,
                         }
                     }


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2199,19 +2199,35 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            Type::Union(union) => union
+                                .try_map(db, |element| into_function_like_callable(db, *element)),
+                            _ => None,
+                        }
+                    }
+
                     let is_input_function_like = inferred_ty
                         .try_upcast_to_callable(self.db())
                         .and_then(Type::as_callable)
                         .is_some_and(|callable| callable.is_function_like(self.db()));
```
sharkdp, line 2223: Initially, I also expanded this "is the input function-like" check to handle unions. @AlexWaygood, I think that was also your original idea. I can do this, but I'm struggling to come up with a scenario where this would be relevant. You would need to have a union of function literals that is being `@decorated`. While it may be possible to construct this somehow, it doesn't feel like it could come up in practice?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2199,19 +2199,35 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    ) -> Option<Type<'d>> {
+                        match ty {
+                            Type::Callable(callable) => Some(Type::Callable(CallableType::new(
+                                db,
+                                callable.signatures(db),
+                                true,
+                            ))),
+                            Type::Union(union) => union
+                                .try_map(db, |element| into_function_like_callable(db, *element)),
+                            _ => None,
```
sharkdp, line 2216: I chose to not handle intersections, because that would require an explicit `Intersection` return type on the decorator.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2199,19 +2199,35 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 .map(|bindings| bindings.return_type(self.db()))
             {
                 Ok(return_ty) => {
+                    fn into_function_like_callable<'d>(
```
AlexWaygood, line 2202: Surely this should be

```suggestion
                    fn as_function_like_callable<'d>(
```

According to your new naming convention ;)

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: That makes sense... Maybe it's worth a comment in the code?

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Well... here, an actual transformation is taking place ;-)

PR #20859, opened 2025-10-14
[ty] Document when a rule was added

MichaReiser:
## Summary

Replace the placeholder `1.0.0` version in the lint definitions with the version when they were first added to ty. 

Update the `rules.md` generation script to include the version in the documentation.

Closes https://github.com/astral-sh/ty/issues/828


MichaReiser:
> As for the actual versions that we put in, I think I would probably prefer them to point to our first stable release, once that's out?

Yes, I think we can consider doing this once we have a stable release (or we update all versions to 0.0.1 with the beta released).

PR #20861, opened 2025-10-14
[ty] Treat `ClassVar`-qualified `Callable`s as bound-method descriptors

sharkdp:


sharkdp:
The ecosystem impact is the same as on https://github.com/astral-sh/ruff/pull/20860 (and this PR includes the same changes). So the impact of also treating `ClassVar`s as bound-method descriptors is zero.

Issue #20845, opened 2025-10-13
Size of ruff-wasm-web - increase by 57% over the past 9 months


AbdealiLoKo:
### Question

Hi, I've been actively following and using the experimental version of ruff-wasm-web 
https://www.npmjs.com/package/@astral-sh/ruff-wasm-web

I noticed a large change in the wasm bundle size recently:
[0.9.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.9.0) -> 8.84 MB
[0.9.10](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.9.10) -> 9.06 MB
[0.10.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.10.0) -> 9.07 MB
[0.11.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.11.0) -> 9.06 MB
[0.11.13](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.11.13) -> 9.17 MB
[0.12.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.12.0) -> 9.23 MB
[0.12.4](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.12.4) -> 9.30 MB
[0.12.5](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.12.5) -> **12.8 MB**
[0.12.12](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.12.12) -> 13.3 MB
[0.13.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.13.0) -> 13.4 MB
[0.13.4](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.13.4) -> 13.9 MB
[0.14.0](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web/v/0.14.0) -> 13.9 MB

8.84 MB -> 13.9 MB is nearly a 57% increase in bundle size

Just checking to see if this is expected ? For a browser, 8 MB is already considered large, and 14MB seemed to be surprising

Are there some suggestions or so to reduce the file size for the browser based wasm ?

ntBre:
I only checked a few of these, but this roughly seems to match the trend in our release  binary size. We may need to take a look at this. Thanks for the report!

| Release | Size (MiB) |
[...]

PR #9031, opened 2023-12-07
release: switch to Cargo's default

BurntSushi:
This sets `lto = "thin"` instead of using "fat" LTO, and sets `codegen-units = 16`. These are the defaults for Cargo's `release` profile, and I think it may give us faster iteration times, especially when benchmarking. The point of this PR is to see what kind of impact this has on benchmarks. It is expected that benchmarks may regress to some extent.

I did some quick ad hoc experiments to quantify this change in compile times. Namely, I ran:

    cargo build --profile release -p ruff_cli

Then I ran

    touch crates/ruff_python_formatter/src/expression/string/docstring.rs

(because that's where i've been working lately) and re-ran

    cargo build --profile release -p ruff_cli

This last command is what I timed, since it reflects how much time one has to wait between making a change and getting a compiled artifact.

Here are my results:

* With status quo `release` profile, build takes 77s
* with `release` but `lto = "thin"`, build takes 41s
* with `release`, but `lto = false`, build takes 19s
* with `release`, but `lto = false` **and** `codegen-units = 16`, build takes 7s
* with `release`, but `lto = "thin"` **and** `codegen-units = 16`, build takes 16s (i believe this is the default `release` configuration)

This PR represents the last option. It's not the fastest to compile, but it's nearly a whole minute faster! The idea is that with `codegen-units = 16`, we still make use of parallelism, but keep _some_ level of LTO on to try and re-gain what we lose by increasing the number of codegen units.

PR #20863, opened 2025-10-14
Enable lto=fat

MichaReiser:
## Summary

This PR switches from thin to fat LTO optimization.

The main motivation is that using lto=fat fixes the binary size increase that we've seen after switching to inventory for [ingredient registration](https://github.com/astral-sh/ruff/issues/20845#issuecomment-3401604238). The regression is mainly due to symbols not being removed when using lto=thin, but they're successfully removed when using lto=fat (from 19MB to 17MB).

Using lto=fat also results in a significant performance improvement, which I think is worth it on its own.

Unfortunately, using lto=flat does have the downside that release builds take significantly longer. One of the main motivations for using `lto=thin` when we switched from `fat` to `thin` in https://github.com/astral-sh/ruff/pull/9031 was to improve performance. To mitigate this, I changed the `--profiling` profile to only use `lto=thin`, similar to what we do in uv (where we disable lto entirely). This PR is also likely to make the mypy primer and benchmark jobs slower (or when running mypy primer locally), because of a significant increase in compile time.

This setup now mirrors uv's (with the exception that `profiling` use `lto=fat`).

I do feel a bit bad about making this change while @BurntSushi is out, because I know he feels the most vocal about this.

Clean build timing:

* fat: 2m 02s
* thin: 1m 00s

Fixes https://github.com/astral-sh/ruff/issues/20845

Relevant discussions:

* https://github.com/astral-sh/uv/pull/5904
* https://github.com/astral-sh/ruff/pull/9031
* https://github.com/astral-sh/uv/pull/5955

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -308,7 +308,7 @@ opt-level = 1
 # But compile times with `lto = fat` are completely untenable.
 #
 # This setup does risk that we are measuring something in benchmarks
-# that we are shipping, but in order to make those two the same, we'd
+# that we aren't shipping, but in order to make those two the same, we'd
 # either need to make compile times way worse for development, or take
 # a hit to binary size and a slight hit to runtime performance in our
 # release builds.


MichaReiser:
I'm not sure why that one instrumented benchmark regresses. But a 10% improvement on our walltime benchmarks is very convincing

Cargo.toml
```diff
@@ -298,11 +296,30 @@ opt-level = 3
...
+# building the release, where as `profiling` is meant for building ty/ruff
+# for running benchmarks.
+#
+# The main differences here are to avoid stripping debug information
+# and disabling fat lto. This does result in a mismatch between our release
+# configuration and our benchmarking configuration, which is unfortunate.
+# But compile times with `lto = fat` are completely untenable.
+#
+# This setup does risk that we are measuring something in benchmarks
+# that we are shipping, but in order to make those two the same, we'd
```
ntBre:
```suggestion
# that we aren't shipping, but in order to make those two the same, we'd
```

BurntSushi:
Yeah I think I've slowly become okay with this sort of change. I even recently did the same for ripgrep. I still don't like that we have different compilation settings for profiling versus release, but I haven't been bitten too hard by it yet.

PR #20865, opened 2025-10-14
[ty] Add (unused) `inferable` parameter to type property methods

dcreager:
A large part of the diff on #20677 just involves threading a new `inferable` parameter through all of the type property methods. In the interests of making that PR easier to review, I've pulled that bit out into here, so that it can be reviewed in isolation. This should be a pure refactoring, with no logic changes or behavioral changes.

### DIFF
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -1,4 +1,4 @@
-use crate::types::constraints::ConstraintSet;
+use std::marker::PhantomData;
 
 use itertools::Itertools;
 use ruff_python_ast as ast;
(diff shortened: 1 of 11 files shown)


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -140,6 +141,16 @@ pub(crate) fn typing_self<'db>(
...
 }
 
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum InferableTypeVars<'a, 'db> {
+    None,
+    // TODO: This variant isn't used, and only exists so that we can include the 'a and 'db in the
+    // type definition. They will be used soon when we start creating real InferableTypeVars
+    // instances.
+    #[expect(unused)]
+    Unused(PhantomData<&'a &'db ()>),
```
sharkdp, line 151: :+1:

I used the same trick in a refactoring yesterday :smile:

MichaReiser: What's the benefit of using `PhantomData` over `()`? Ah, never mind, we need the lifetime

dcreager: > I used the same trick in a refactoring yesterday

Blast, no fist-bump reaction emoji here! Please know that I am giving you a fist bump.

PR #20867, opened 2025-10-14
Fix syntax error false positives for escapes and quotes in f-strings

### LINKED ISSUE #20844: Unsupported syntax error false positive for escapes in f-string format specifier
This string from our formatter test suite:

https://github.com/astral-sh/ruff/blob/e26dce5543687e84e40f0d57c8ba4949275b18a2/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/fstring.py#L709-L711

gets [formatted](https://play.ruff.rs/5520ae38-9395-48ed-87cd-67e7002255d9) to

```py
f"{x:a{z:hy \"user\"}} '''"
```

which is reported as [invalid](https://play.ruff.rs/154cf36b-de49-45a1-9d27-c3bf5032c591) by Ruff. However, this is allowed even before Python 3.12:

```pycon
Python 3.7.9 (default, Aug 23 2020, 00:57:53)
[Clang 10.0.1 ] on linux
>>> f"{x:a{z:hy \"user\"}} '''"
```

even if GitHub's syntax highlighting doesn't like it. The `NameError` means this parsed successfully, unlike escapes in the expression part:

```pycon
>>> f"{'\x123'}"
  File "<stdin>", line 1
SyntaxError: f-string expression part cannot include a backslash
```

It seems that my logic for detecting these was too naive after all:

https://github.com/astral-sh/ruff/blob/e26dce5543687e84e40f0d57c8ba4949275b18a2/crates/ruff_python_parser/src/parser/expression.rs#L1595-L1612

This example from the formatter emits both the backslash and reused quote character diagnostics. We should double-check the comment semantics too, but I think that part may still be correct.

ntBre:
Summary
--

Fixes #20844 by refining the unsupported syntax error check for [PEP 701]
f-strings before Python 3.12 to allow backslash escapes and escaped outer quotes
in the format spec part of f-strings. These are only disallowed within the
f-string expression part on earlier versions. Using the examples from the PR:

```pycon
>>> f"{1:\x64}"
'1'
>>> f"{1:\"d\"}"
ValueError: Invalid format specifier '"d"' for object of type 'int'
```

Note that the second case is a runtime error, but this is actually avoidable if
you override `__format__`, so despite being pretty weird, this could actually be
a valid use case.

```pycon
>>> class C:
...     def __format__(*args, **kwargs): return "<C>"
...
>>> f"{C():\"d\"}"
'<C>'
```

At first I thought narrowing the range we check to exclude the format spec would
only work for escapes, but it turns out that cases like `f"{1:""}"` are already
covered by an existing `ParseError`, so we can just narrow the range of both our
escape and quote checks.

Our comment check also seems to be working correctly because it's based on the
actual tokens. A case like [this](https://play.ruff.rs/9f1c2ff2-cd8e-4ad7-9f40-56c0a524209f):

```python
f"""{1:# }"""
```

doesn't include a comment token, instead the `#` is part of an
`InterpolatedStringLiteralElement`.

Test Plan
--

New inline parser tests

[PEP 701]: https://peps.python.org/pep-0701/

### DIFF
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/fstring.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/fstring.py
@@ -706,8 +706,6 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{'aa'}" }'
 f'{1=: "abcd {'aa'}}'
-# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
-# spec, which is valid even before 3.12.
 f'{x:a{z:hy "user"}} \'\'\''
 
 # Changing the outer quotes is fine because the format-spec is in a nested expression.
--- a/crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
+++ b/crates/ruff_python_formatter/tests/snapshots/format@expression__fstring.py.snap
@@ -712,8 +712,6 @@ f'{1:hy "user"}'
 f'{1: abcd "{1}" }'
 f'{1: abcd "{'aa'}" }'
 f'{1=: "abcd {'aa'}}'
-# FIXME(brent) This should not be a syntax error on output. The escaped quotes are in the format
-# spec, which is valid even before 3.12.
 f'{x:a{z:hy "user"}} \'\'\''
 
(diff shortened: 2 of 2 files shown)


MichaReiser (approved):
Thank you

Issue #15642, opened 2025-01-21
Extend PEP 695 generic class and function rules


ntBre:
Follow up tasks from #15565

- [ ] Rename private type variables in generics (don't make inline generics like `[_T]`)
- [ ] Allow multiple base classes for generic classes
- [ ] Handle generic methods
- [ ] Handle `default` kwargs for Python >= 3.13

Related: #4617, #12542

ntBre:
I started working on renaming private type variables, but I think it's a bit more complicated than I thought. At first I was just going to do the rename as part of `DisplayTypeVar::fmt`, but I realized that you have to be pretty careful not to change the meaning of other instances of the type you're renaming. For example,


```python
class T: ...

_T = TypeVar("_T")

def generic(var: _T) -> _T:
    x: T = T()
    return var
```

You can't even just look at the signature or the parent scopes, you also have to look at the body of the class or function.

Even if you restrict the fix to variables not in the parent scope, you then have to update the name of the type in the body of the class or function too:

```python
_G = TypeVar("_G")

class SafePrivate(Generic[_G]):
    var1: _G
```

This was the test case I was writing when I realized that I needed to rename the inner `_G` too, not just the one in the class signature. I found the [Renamer](https://github.com/astral-sh/ruff/blob/05abd642a8ddb3f3b6e7c78592995c6ccfca2520/crates/ruff_linter/src/renamer.rs#L12) API, which sounded very promising, but I'm running into some panics like this: `assertion failed: start.raw <= end.raw`, which I assume is because of the overlapping `Edit::replacement` from the existing UP046 code and the edits from `Renamer::rename`.

Would I need a separate rule to do the `rename` in a separate pass from the initial replacement? Or is there another approach I'm missing?

AlexWaygood:
> Would I need a separate rule to do the `rename` in a separate pass from the initial replacement? Or is there another approach I'm missing?

This might be quite a good solution. It would allow us to use the `Renamer` API (which I agree is the easiest, most robust way to do this!), and would also allow us to complain about e.g. handwritten PEP-695 classes that use TypeVars with leading underscores, which I think is a bit of a stylistic antipattern since they're private to the class by default. For this, though, I think it does make sense to have one rule that would rename type parameters in type aliases, functions and classes, rather than a separate rule for each. (That doesn't mean that they all have to be tackled in one PR, however!)

AlexWaygood:
[...]

PR #20868, opened 2025-10-14
[syntax-errors] Annotated name cannot be global

11happy:
## Summary

This PR implements a new semantic syntax error where annotated name can't be global
example
```
x: int = 1

def f():
    global x
    x: str = "foo"  # SyntaxError: annotated name 'x' can't be global
 ```

## Test Plan

I have written tests as directed in #17412

### DIFF
--- a/crates/ruff_python_parser/src/semantic_errors.rs
+++ b/crates/ruff_python_parser/src/semantic_errors.rs
@@ -302,7 +302,7 @@ impl SemanticSyntaxChecker {
                 if let Expr::Name(ast::ExprName { id, .. }) = target.as_ref() {
                     if let Some(global_stmt) = ctx.global(id.as_str()) {
                         let global_start = global_stmt.start();
-                        if ctx.in_function_scope() || target.start() < global_start {
+                        if !ctx.in_module_scope() || target.start() < global_start {
                             Self::add_error(
                                 ctx,
                                 SemanticSyntaxErrorKind::AnnotatedGlobal(id.to_string()),
--- a/crates/ruff_linter/resources/test/fixtures/semantic_errors/annotated_global.py
+++ b/crates/ruff_linter/resources/test/fixtures/semantic_errors/annotated_global.py
@@ -28,3 +28,11 @@ f: int = 1  # okay
 
 g: int = 1
 global g  # error
+
+class C:
+    x: str
+    global x  # error
+
+class D:
+    global x  # error
+    x: str
--- a/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__semantic_syntax_error_annotated_global.py_3.14.snap
+++ b/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__semantic_syntax_error_annotated_global.py_3.14.snap
@@ -54,3 +54,21 @@ invalid-syntax: annotated name `g` can't be global
    | ^
 30 | global g  # error
    |
+
+invalid-syntax: annotated name `x` can't be global
+  --> resources/test/fixtures/semantic_errors/annotated_global.py:33:5
+   |
+32 | class C:
+33 |     x: str
+   |     ^
+34 |     global x  # error
+   |
+
+invalid-syntax: annotated name `x` can't be global
+  --> resources/test/fixtures/semantic_errors/annotated_global.py:38:5
+   |
+36 | class D:
+37 |     global x  # error
+38 |     x: str
+   |     ^
+   |

### CODE REVIEW

File: crates/ruff_linter/src/linter.rs [hunk 1]
```diff
@@ -1060,6 +1060,28 @@ mod tests {
...
+        b: int = 1
+        def outer1():
+            def inner1():
+                global b
+                b: str = 'nested'
+        
+        c: int = 1
+        def f2():
+            global c
+            c: list[str] = []
```
amyreese: Should we also add a test case like:

```
d: int = 1
def f3():
    global d
    d: str
```

to ensure we catch annotations without explicit assignments, and:

```
e: int = 1
def f3():
    e: str = "whatever"
```

to prove that this doesn't trigger on shadowed variables that aren't actually marked with `global`?

11happy: yes, these are great additions to the testcases, I have added them, Thank you : )

amyreese (approved):

ntBre:
Actually, I thought of one more edge case, and it looks like we have a false positive:

```py
global f
f: int
```

This is actually allowed in module scope, for whatever reason, but the opposite order is not:

```py
f: int
global f
```

```shell
$ python <<EOF
global f
f: int
EOF
$ python <<EOF
f: int
global f
EOF
  File "<stdin>", line 2
SyntaxError: annotated name 'f' can't be global
```

11happy:
> Thanks for adding the new test case, but as I mentioned in my previous comment, this case is actually allowed by CPython, so the new snapshot is showing a false positive. We'll need to fix that and the merge conflicts before this can land.

Done, Thank you for the patience with this PR
: )

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -297,6 +299,15 @@ impl SemanticSyntaxChecker {
                     };
                     visitor.visit_expr(annotation);
                 }
+                if let Expr::Name(ast::ExprName { id, .. }) = target.as_ref() {
+                    if ctx.global(id.as_str()).is_some() && ctx.in_function_scope() {
```
ntBre: I pushed another test case that should still raise an error in the module scope. This `in_function_scope` check is too restrictive, we still miss the new case I added:

```shell
❯ python <<EOF
g: int = 1
global g
EOF
  File "<stdin>", line 2
SyntaxError: annotated name 'g' can't be global
```

I also noticed that ty actually has a false positive on the `f` case: https://play.ty.dev/d6285858-e5cf-429e-a983-291a243db217.

So we may want to fix that as well, if we can find a nice solution on the Ruff side.

For Ruff, we may be able to compare the range returned by `ctx.global` to the range of the `ExprName`. I think that's what we do for the other `global` error:

https://github.com/astral-sh/ruff/blob/e8540d9b085980bef22cc8e1d4c6e75531cf140c/crates/ruff_python_parser/src/semantic_errors.rs#L858-L860

11happy: with the current fix, all tests are working as expected. however, I wanted to ask about an edge case:

does order matter in function scope for the `AnnotatedGlobal` error? Currently, Ruff emits `AnnotatedGlobal` for this case:
```python
a: str = "hel"
def foo():
    a: str = "foo"
    global a
```

But ty emits a different error  https://play.ty.dev/dc80ad75-cb1c-4f30-99f2-4fa555b191c7

also, since most of the errors in #17412 are now addressed, I would appreciate any suggestions for related issues or areas I could work on next. I would also be grateful if you could review some of my other open PRs when you have time. 
Thank you : )

ntBre: No I don't believe order matters in functions, only in the module scope:

```pycon
>>> def foo():
...     a: str
...     global a
...
  File "<python-input-1>", line 3
    global a
    ^^^^^^^^
SyntaxError: annotated name 'a' can't be global
>>> def bar():
...     global a
...     a: str
...
  File "<python-input-2>", line 3
    a: str
    ^^^^^^
SyntaxError: annotated name 'a' can't be global
```

at least the module scope is the only strange case I've found so far.

> But ty emits a different error [play.ty.dev/dc80ad75-cb1c-4f30-99f2-4fa555b191c7](https://play.ty.dev/dc80ad75-cb1c-4f30-99f2-4fa555b191c7)

I think we can file a follow-up issue for ty, but I would expect the fix to be here:

https://github.com/astral-sh/ruff/blob/0f373603ebd007196af4f211e14d421024a6b947/crates/ty_python_semantic/src/semantic_index/builder.rs#L2275-L2282

Assuming both errors can be emitted in this part of the code, I would probably prioritize the `AnnotatedGlobal` error here instead of `LoadBeforeGlobalDeclaration` to match CPython.

> also, since most of the errors in https://github.com/astral-sh/ruff/issues/17412 are now addressed, I would appreciate any suggestions for related issues or areas I could work on next. 

In general, anything with the `help wanted` or `bug` labels would be great :) We don't have too many big tracking issue like #17412 open at the moment. A couple of issues that I've opened that came to mind are https://github.com/astral-sh/ruff/issues/15642 and https://github.com/astral-sh/ruff/issues/17203 if those look interesting to you.

> I would also be grateful if you could review some of my other open PRs when you have time.

Will do! They're all in my inbox, just been busy with other tasks recently.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -297,6 +299,18 @@ impl SemanticSyntaxChecker {
                     };
                     visitor.visit_expr(annotation);
                 }
+                if let Expr::Name(ast::ExprName { id, .. }) = target.as_ref() {
+                    if let Some(global_stmt) = ctx.global(id.as_str()) {
+                        let global_start = global_stmt.start();
+                        if ctx.in_function_scope() || target.start() < global_start {
```
ntBre: Could you check how CPython and your code handle the class scope? I think we probably want to check `!ctx.in_module_scope()` here instead of specifically checking `in_functon_scope`. I'm picturing an example like this:

```py
class C:
    x: str
    global x

class D:
    global x
    x: str
```

It looks like these should both be errors:

```pycon
>>> class C:
...     x: str
...     global x
...
  File "<python-input-3>", line 3
    global x
    ^^^^^^^^
SyntaxError: annotated name 'x' can't be global
>>> class D:
...     global x
...     x: str
...
  File "<python-input-4>", line 3
    x: str
    ^^^^^^
SyntaxError: annotated name 'x' can't be global
```

11happy: done , Yes, these were earlier not being detected, I have updated the check with `!ctx.in_module_scope()` .

PR #20869, opened 2025-10-14
[syntax-errors]: implement F702 as semantic syntax error

11happy:
## Summary

This PR implements `F702` https://docs.astral.sh/ruff/rules/continue-outside-loop/ as semantic syntax error.

## Test Plan

Tests are already previously written in F702


ntBre (approved):
Great, thank you!

PR #20806, opened 2025-10-10
[ty] use type context more aggressively to infer values ​​when constructing a `TypedDict`

mtshiba:
## Summary

Based on @ibraheemdev's comment on #20792:

> I think we can also update our bidirectional inference code, [which makes the same assumption](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types/infer/builder.rs?rgh-link-date=2025-10-09T21%3A30%3A31Z#L5860).

This PR also adds more test cases for how `TypedDict` annotations affect generic call inference.

## Test Plan

New tests in `typed_dict.md`

PR #20871, opened 2025-10-14
[ty] refactor `Place`

mtshiba:
## Summary

Part of astral-sh/ty#1341

The following changes will be made to `Place`.

* Introduce `TypeOrigin`
* `Place::Type` -> `Place::Defined`
* `Place::Unbound` -> `Place::Undefined`
* `Boundness` -> `Definedness`

`TypeOrigin::Declared`+`Definedness::PossiblyUndefined` are patterns that weren't considered before, but this PR doesn't address them yet, only refactors.

## Test Plan

N/A

### DIFF
--- a/crates/ty_python_semantic/src/types/member.rs
+++ b/crates/ty_python_semantic/src/types/member.rs
@@ -4,8 +4,47 @@ use crate::place::{
     place_from_bindings,
 };
 use crate::semantic_index::{place_table, scope::ScopeId, use_def_map};
+use crate::types::Type;
 
-pub(super) type Member<'db> = PlaceAndQualifiers<'db>;
+/// The return type of certain member-lookup operations. Contains information
+/// about the type, type qualifiers, boundness/declaredness.
+#[derive(Debug, Clone, PartialEq, Eq, salsa::Update, get_size2::GetSize, Default)]
+pub(super) struct Member<'db> {
+    /// Type, qualifiers, and boundness information of this member
+    pub(super) inner: PlaceAndQualifiers<'db>,
+}
+
+impl<'db> Member<'db> {
+    pub(super) fn unbound() -> Self {
+        Self {
+            inner: PlaceAndQualifiers::unbound(),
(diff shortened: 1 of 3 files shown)


mtshiba:
Hmm, the diff in mypy_primer is unexpected.
There was no diff in the initial commit, and it appears that the diff was caused by a merge commit that should have been unrelated.

It seems that this also occurred in #20792 and #20806.
I don't know the cause, but perhaps there is non-deterministic inference being made regarding `dataclasses.field` or overloading?

AlexWaygood:
> but perhaps there is non-deterministic inference being made regarding `dataclasses.field` or overloading?

Yeah, there does seem to be something nondeterministic happening there. It's quite annoying; if you feel like tracking it down, that would be very helpful 😄

crates/ty_python_semantic/src/types/member.rs
```diff
@@ -1,103 +1,35 @@
...
-    /// Map a type transformation function over the type of this member.
-    #[must_use]
-    pub(super) fn map_type(self, f: impl FnOnce(Type<'db>) -> Type<'db>) -> Self {
-        Self {
-            inner: self.inner.map_type(f),
-            is_declared: self.is_declared,
-        }
-    }
-}
+pub(super) type Member<'db> = PlaceAndQualifiers<'db>;
```
sharkdp: I think I'm inclined to keep `Member` as a struct, even if it only consists of `inner: pub(super) inner: PlaceAndQualifiers<'db>` for now. I am quite confident that we will add other attributes to this soon.

But great to see that the new `Place` makes the `is_declared` flag superfluous.

sharkdp (approved):
Thank you for doing this! Just one comment.

If someone is unhappy with the newly proposed names here, we can always easily change them later. I think we should aim to get this refactor in as quick as possible, or it will collect a lot of conflicts.

sharkdp:
Thank you!

PR #20875, opened 2025-10-15
[ty] Avoid unnecessarily widening generic specializations

ibraheemdev:
## Summary

Ignore the type context when specializing a generic call if it leads to an unnecessarily wide return type. For example, [the example mentioned here](https://github.com/astral-sh/ruff/pull/20796#issuecomment-3403319536) works as expected after this change:
```py
def id[T](x: T) -> T:
    return x

def _(i: int):
    x: int | None = id(i)
    y: int | None = i
    reveal_type(x)  # revealed: int
    reveal_type(y)  # revealed: int
```

I also added extended our usage of `filter_disjoint_elements` to tuple and typed-dict inference, which resolves https://github.com/astral-sh/ty/issues/1266.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
@@ -11,7 +11,7 @@ class Member:
     role: str = field(default="user")
     tag: str | None = field(default=None, init=False)
 
-# revealed: (self: Member, name: str, role: str = str) -> None
+# revealed: (self: Member, name: str, role: str = Literal["user"]) -> None
 reveal_type(Member.__init__)
 
 alice = Member(name="Alice", role="admin")
@@ -37,7 +37,7 @@ class Data:
     content: list[int] = field(default_factory=list)
     timestamp: datetime = field(default_factory=datetime.now, init=False)
 
-# revealed: (self: Data, content: list[int] = list[int]) -> None
+# revealed: (self: Data, content: list[int] = Unknown) -> None
 reveal_type(Data.__init__)
 
 data = Data([1, 2, 3])
@@ -64,7 +64,7 @@ class Person:
     role: str = field(default="user", kw_only=True)
 
 # TODO: this would ideally show a default value of `None` for `age`
-# revealed: (self: Person, name: str, *, age: int | None = int | None, role: str = str) -> None
+# revealed: (self: Person, name: str, *, age: int | None = None, role: str = Literal["user"]) -> None
 reveal_type(Person.__init__)
 
 alice = Person(role="admin", name="Alice")


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -416,13 +415,13 @@ a = f("a")
...
-reveal_type(b)  # revealed: list[int | Literal["a"]]
+reveal_type(b)  # revealed: list[Literal["a"] | int]
 
 c: list[int | str] = f("a")
-reveal_type(c)  # revealed: list[int | str]
+reveal_type(c)  # revealed: list[str | int]
 
 d: list[int | tuple[int, int]] = f((1, 2))
-reveal_type(d)  # revealed: list[int | tuple[int, int]]
+reveal_type(d)  # revealed: list[tuple[int, int] | int]
```
ibraheemdev, line 425: The churn here could be avoided if we performed a *third* specialization, inferring the call expression annotation first before the argument types, but I'm not sure it's worth it.

ibraheemdev:
The ecosystem results look good. There are some regressions around `functools.reduce` because we aren't able to correctly infer the type variables in a `Callable`, so we choose the overly narrow type without widening based on the type context. However, those should be fixed by the new trait solver, we shouldn't need the type context for that to work.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2563,10 +2548,42 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+            .return_ty
+            .apply_specialization(self.db, isolated_specialization);
+
+        // TODO: Ideally we would infer the annotated type _before_ the arguments if this call is part of an
+        // annotated assignment, to closer match the order of any unions written in the type annotation.
+        if let Some(call_expression_tcx) = self.call_expression_tcx.annotation {
+            match call_expression_tcx {
+                // A type variable is not a useful type-context for expression inference, and applying it
+                // to the return type can lead to confusing unions in nested generic calls.
+                Type::TypeVar(_) => {}
```
dcreager: You can use `Type::is_type_var` here. Or alternatively, do we want to update this to skip the type context if it contains a typevar anywhere inside of it? In that case you'd use `Type::has_typevar`

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2563,10 +2548,42 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
             }
         }
 
-        self.specialization = self
-            .signature
-            .generic_context
-            .map(|gc| builder.build(gc, *self.call_expression_tcx));
+        // Build the specialization once without type context.
+        let isolated_specialization = builder.build(gc, *self.call_expression_tcx);
```
dcreager: Should this use `TypeContext::default`? The comment says "without the type context"

crates/ty_python_semantic/src/types.rs
```diff
@@ -1196,15 +1196,20 @@ impl<'db> Type<'db> {
         if yes { self.negate(db) } else { *self }
     }
 
-    /// Remove the union elements that are not related to `target`.
-    pub(crate) fn filter_disjoint_elements(self, db: &'db dyn Db, target: Type<'db>) -> Type<'db> {
+    /// Filters union elements based on the provided predicate.
+    pub(crate) fn filter(self, db: &'db dyn Db, f: impl FnMut(&Type<'db>) -> bool) -> Type<'db> {
```
dcreager: nit: I would call this `filter_union`

crates/ty_python_semantic/src/types.rs
```diff
@@ -1196,15 +1196,20 @@ impl<'db> Type<'db> {
         if yes { self.negate(db) } else { *self }
     }
 
-    /// Remove the union elements that are not related to `target`.
-    pub(crate) fn filter_disjoint_elements(self, db: &'db dyn Db, target: Type<'db>) -> Type<'db> {
+    /// Filters union elements based on the provided predicate.
```
dcreager:
```suggestion
    /// If this type is a union, filters the union elements based on the provided predicate.
    /// Otherwise returns the type unchanged.
```

(probably needs to be reflowed to satisfy pre-commit)

crates/ty_python_semantic/src/types.rs
```diff
@@ -10861,9 +10866,9 @@ impl<'db> UnionType<'db> {
     pub(crate) fn filter(
         self,
         db: &'db dyn Db,
-        filter_fn: impl FnMut(&&Type<'db>) -> bool,
+        mut f: impl FnMut(&Type<'db>) -> bool,
     ) -> Type<'db> {
-        Self::from_elements(db, self.elements(db).iter().filter(filter_fn))
+        Self::from_elements(db, self.elements(db).iter().filter(|ty| f(ty)))
```
dcreager, line 11203: Can we make the callback take in `Type<'db>` instead of `&Type<'db>`? I find that much more ergonomic, and I think it just requires a `.copied()` before the `filter`. (Ditto above in `Type::filter`)

AlexWaygood: Ibraheem's current signature matches the signature for `UnionType::map()`. I don't have a strong objection to changing _that_ signature, but I think it's nice to keep the signature of `map()` consistent with the signature of `filter()`. (And IIRC, I made the callback passed to `UnionType::map()` take `&Type<'db>` because it seemed generally to make most callsites more ergonomic than if it took `Type<'db>`. Due to the fact that `some_union.elements(db).iter()` calls return `Iterator<Item = &Type<'db>>`s rather than `Iterator<Item = Type<'db>>`s)

dcreager: :+1: I agree with keeping them consistent, but feel strongly enough about avoiding `&Type` when we can that I might tackle that as a separate follow on. You know, in my copious spare time.

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: I'll use `is_type_var` for now. We still want to use the type context as much as possible, we just want to avoid ever adding a type-mapping to a typevar, e.g. `dict[TypedDict(..), T]` is still useful. I'll address this in a follow up PR.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2490,20 +2490,23 @@ struct ArgumentTypeChecker<'a, 'db> {
...
     call_expression_tcx: &'a TypeContext<'db>,
+    return_ty: Type<'db>,
     errors: &'a mut Vec<BindingError<'db>>,
 
     inferable_typevars: InferableTypeVars<'db, 'db>,
     specialization: Option<Specialization<'db>>,
 }
 
 impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
+    #[allow(clippy::too_many_arguments)]
```
```suggestion
    #[expect(clippy::too_many_arguments)]
```

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: I updated the comment, this is specifically talking about inferring the return type using the type context. We still need the type context to avoid eager literal promotion (regardless of whether or not we choose the inference with or without the type context).

crates/ty_python_semantic/src/types.rs
ibraheemdev: There are some `Type` methods that take `&self`, so we should probably address them all at once. Only changing `filter` would mean `.filter(Type::is_typed_dict)` no longer works, for example.

AlexWaygood: I think the last time we discussed it, we decided that it was pretty borderline but that it was _probably_ slightly more efficient for methods like that to take `&self` rather than `self`, despite the fact that `Type` is `Copy`, given how big `Type` is

ibraheemdev:
I updated the heuristic to ignore the type context if the specialization is already assignable to the type context. This seems more consistent than the previous subtyping check, and also means we don't have to perform a second specialization if the first one is valid.

This is somewhat related to https://github.com/astral-sh/ty/issues/136, but I think that is a decision we have to make separately and more related to preferring the declared type itself, not the inference that uses the declared type. For example, in https://github.com/astral-sh/ruff/pull/20796, we want re-assignments to be narrowed while still accounting for the type context. This seems to remove all the false positives that showed up in the ecosystem report as https://github.com/astral-sh/ruff/pull/20796, and removes a number of existing false positives as well.

The regression with `functools.reduce` remains as mentioned before, and there seems to also be a regression with static method constructors, but I believe in that case we aren't able to perform literal promotion due to an unrelated issue.

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: Actually, I'm not even sure we need this check anymore, given we'll prefer the first inference over the one unioned with a typevar. I suppose it might make for better diagnostics for invalid assignments?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
dcreager: There is a TODO below indicating where we'd make the change if we wanted to fix this, so I'm 👍 to this. Though maybe add a TODO comment here, too, calling out that we realize that we reordered the union, and that we know how we'd fix it?

PR #20878, opened 2025-10-15
[`pyflakes`] Revert to stable behavior if imports for module lie in alternate branches for `F401`

### LINKED ISSUE #20839: F401: false positive causes NameError
### Summary

ruff 0.14.0 introduced a false-positive for F401 unused-import:

```python
  try:
      import asdfasfd.asdf.sadf
  except ImportError:
      import argparse
      import asdfasfd  # W: `asdfasfd` imported but unused
      asdfasfd.asdf = argparse.Namespace()
```

autofix causes:
```python
  File "foo.py", line 6, in <module>                                                
    asdfasfd.asdf = argparse.Namespace()       
    ^^^^^^^^                                               
NameError: name 'asdfasfd' is not defined   
```

### Version

0.14.0

dylwil3:
Closes #20839

### DIFF
--- a/crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__f401_type_checking.snap
+++ b/crates/ruff_linter/src/rules/pyflakes/snapshots/ruff_linter__rules__pyflakes__tests__f401_type_checking.snap
@@ -1,0 +1,66 @@
+---
+source: crates/ruff_linter/src/rules/pyflakes/mod.rs
+---
+F401 [*] `mlflow.pyfunc.loaders.chat_agent` imported but unused
+ --> f401_preview_submodule.py:2:8
+  |
+2 | import mlflow.pyfunc.loaders.chat_agent
+  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+3 | import mlflow.pyfunc.loaders.chat_model
+4 | import mlflow.pyfunc.loaders.code_model
+  |
+help: Remove unused import: `mlflow.pyfunc.loaders.chat_agent`
+1 | 
+  - import mlflow.pyfunc.loaders.chat_agent
+2 | import mlflow.pyfunc.loaders.chat_model
+3 | import mlflow.pyfunc.loaders.code_model
+4 | from mlflow.utils.pydantic_utils import IS_PYDANTIC_V2_OR_NEWER
+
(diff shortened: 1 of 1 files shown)


dylwil3:
The ecosystem hits look correct - these two are debatable, but keeping the stable behavior seems fine for now:

- [providers/google/src/airflow/providers/google/common/utils/id_token_credentials.py:43:12:](https://github.com/apache/airflow/blob/7939692b266a20b340f133f998612f083c585b6c/providers/google/src/airflow/providers/google/common/utils/id_token_credentials.py#L43)
- [mlflow/pyfunc/loaders/__init__.py:1:8:](https://github.com/mlflow/mlflow/blob/2749d7690b0ed57e0a26af4475021a380e25fdb3/mlflow/pyfunc/loaders/__init__.py#L1)

crates/ruff_linter/src/rules/pyflakes/rules/unused_import.rs
```diff
@@ -911,6 +915,12 @@ fn has_simple_shadowed_bindings(scope: &Scope, id: BindingId, semantic: &Semanti
         if i > 0 && shadowed_binding.is_used() {
             return false;
         }
+        if shadowed_binding
+            .source
+            .is_none_or(|node_id| !semantic.same_branch(node_id, binding_node))
+        {
+            return false;
+        }
         matches!(
```
MichaReiser, line 947: Nit: Maybe add some spaces to make it clearer that these are separate checks (and the comment only applies to the first)
```suggestion
        if i > 0 && shadowed_binding.is_used() {
            return false;
        }
        
        if shadowed_binding
            .source
            .is_none_or(|node_id| !semantic.same_branch(node_id, binding_node))
        {
            return false;
        }
        
        matches!(
```

MichaReiser:
The examples with `if TYPE_CHECKING` are a bit unfortunate. Do you think it would be possible to instead test if the import is in an alternate branch than the other import? 

E.g. the following should be flagged

```py

if True:
	import a.b
	if other:
		import a.b
```

but not

```py
if True:
	import a.b
else:
	if other:
		import a.b
```

So, I guess, we'd have to check if the first import ancestor chain is a prefix of the second import's ancestor chain

MichaReiser:
Did this new approach improve the ecosystem results?

dylwil3:
> Did this new approach improve the ecosystem results?

Yes! The things that remain are variations on the `try`/`except` pattern in the original issue.

PR #20879, opened 2025-10-15
[`pandas-vet`] Fix false positive for `.values` on NumPy `NamedTuples` (`PD011`)

### LINKED ISSUE #20874: False positve PD011: using `.values` on a NumPy NamedTuple, not a Pandas Series or Index object
### Summary

I suspect this situation is infrequent.

```python
...
import pandas

if TYPE_CHECKING:
    from numpy.lib._arraysetops_impl import UniqueInverseResult
...
        unique: UniqueInverseResult[numpy.uint64] = numpy.unique_inverse(arrayAnalyzed[slicerArcCode])

        state.arrayArcCodes = unique.values
```

[Snippet from this module](https://github.com/hunterhogan/mapFolding/blob/2ac46a4c15ee06b5a58ea674891ee747766e843d/mapFolding/algorithms/matrixMeandersNumPyndas.py), if you want it.

[`numpy.unique_inverse`](https://github.com/numpy/numpy/blob/0532af47d6a815298b7841de00bdbc547104b237/numpy/lib/_arraysetops_impl.py#L577-L584) returns a NamedTuple with a 'values' field.

`numpy.unique_all` and `numpy.unique_counts` [do the same](https://github.com/numpy/numpy/blob/0532af47d6a815298b7841de00bdbc547104b237/numpy/lib/_arraysetops_impl.py#L416-L430).

I'm not a NumPy expert, but I'm not aware of other functions that do the same thing, and I did a quick search of NumPy for 'from typing import NamedTuple' and that was the only module.

<details><summary>Diagnostic</summary>
<p>

```json
[{
	"resource": "/c:/apps/mapFolding/mapFolding/algorithms/matrixMeandersNumPyndas.py",
	"owner": "Ruff",
	"code": {
		"value": "PD011",
		"target": {
			"$mid": 1,
			"path": "/ruff/rules/pandas-use-of-dot-values",
			"scheme": "https",
			"authority": "docs.astral.sh"
		}
	},
	"severity": 4,
	"message": "Use `.to_numpy()` instead of `.values`",
	"source": "Ruff",
	"startLineNumber": 44,
	"startColumn": 26,
	"endLineNumber": 44,
	"endColumn": 39,
	"origin": "extHost2"
}]
```

</p>
</details> 

<details><summary>Environment</summary>
<p>

>Version: 1.106.0-insider (user setup)
Commit: 359fa23b39bb37e5157e99c2bc5fc81088dd8431
Date: 2025-10-14T19:15:26.327Z
Electron: 37.6.0
ElectronBuildId: 12506819
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26100

> charliermarsh.ruff
Version
2025.28.0
Last Updated
6 days ago

</p>
</details> 

### Version

2025.28.0

danparizher:
## Summary

Fixes #20874 - False positive PD011: using `.values` on a NumPy NamedTuple, not a Pandas Series or Index object

## Problem Analysis

The PD011 rule (`pandas-use-of-dot-values`) was incorrectly flagging `.values` usage on non-pandas objects, particularly:

1. **NumPy NamedTuples** returned by functions like `numpy.unique_inverse()`, `numpy.unique_all()`, and `numpy.unique_counts()`
2. **Simple non-pandas objects** like integers, strings, or other data types that happen to have a `.values` attribute

The original issue reported a false positive where `unique.values` was flagged, where `unique` was a `UniqueInverseResult` NamedTuple from NumPy, not a pandas object.

### General Solution Strategy

Rather than creating NumPy-specific logic, I enhanced the existing `test_expression` helper function in `crates/ruff_linter/src/rules/pandas_vet/helpers.rs` to be more intelligent about determining whether a binding comes from pandas-related sources.

### Implementation Details

1. **Enhanced `test_expression` Function**:
   - Added logic to check if variable bindings come from pandas-related sources
   - Uses `find_binding_value()` to trace back to the original assignment
   - Calls new `is_pandas_related_value()` function to determine pandas relevance

2. **New `is_pandas_related_value()` Function**:
   - Analyzes expressions to determine if they originate from pandas
   - Handles literals (numbers, strings, etc.) - returns `false` (not pandas-related)
   - Handles pandas imports - returns `true` (pandas-related)
   - Handles pandas method calls and function calls - returns `true`
   - Handles unknown expressions - returns `false` (conservative approach)

## Ecosystem Impact Analysis

The ecosystem bot reported **-17 violations** across 2 projects:

### apache/superset (-14 violations)

- **12 PD011 violations removed**: These were false positives where `.values` was accessed on non-pandas objects in pandas postprocessing modules
- **1 PD013 violation removed**: `.stack` usage that was incorrectly flagged
- **1 PD008 violation removed**: `.at` usage that was incorrectly flagged

### bokeh/bokeh (-3 violations)

- **1 PD011 violation removed**: `.values` usage on non-pandas object
- **1 PD013 violation removed**: `.stack` usage that was incorrectly flagged  
- **1 PD010 violation removed**: `.pivot` usage that was incorrectly flagged

### DIFF
--- a/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_typed_unique_inverse.snap
+++ b/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_typed_unique_inverse.snap
@@ -1,4 +1,0 @@
----
-source: crates/ruff_linter/src/rules/pandas_vet/mod.rs
----
-
--- a/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_all.snap
+++ b/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_all.snap
@@ -1,4 +1,0 @@
----
-source: crates/ruff_linter/src/rules/pandas_vet/mod.rs
----
-
--- a/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_counts.snap
+++ b/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_counts.snap
@@ -1,4 +1,0 @@
----
-source: crates/ruff_linter/src/rules/pandas_vet/mod.rs
----
-
--- a/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_inverse.snap
+++ b/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_numpy_unique_inverse.snap
@@ -1,4 +1,0 @@
----
-source: crates/ruff_linter/src/rules/pandas_vet/mod.rs
----
-
--- a/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_simple_non_pandas.snap
+++ b/crates/ruff_linter/src/rules/pandas_vet/snapshots/ruff_linter__rules__pandas_vet__tests__PD011_pass_simple_non_pandas.snap
@@ -1,4 +1,0 @@
----
-source: crates/ruff_linter/src/rules/pandas_vet/mod.rs
----
-


crates/ruff_linter/src/rules/pandas_vet/rules/attr.rs
```diff
@@ -43,6 +43,38 @@ impl Violation for PandasUseOfDotValues {
     }
 }
 
+/// Check if a binding comes from a NumPy function that returns a `NamedTuple` with a `.values` field.
```
ntBre: I don't think this is specific to numpy. A very simple [example](https://play.ruff.rs/90cd0ead-1783-400e-abf0-abe203f94b28) like this exhibits the same behavior:

```py
import pandas

p = 1
p.values
```

We also already have the `test_expression` helper that tries to filter out non-pandas bindings. It might be worth trying to improve that, which would be shared by the other pandas rules.

ntBre (changes requested):

MichaReiser:
Hi @danparizher. Thanks for working on this. 

If you want to help us as reviewers. What I'd find very useful is if you add a few more details to the PR summary explaining why you took this specific approach. What constraints did you consider? What are its up or downsides? 

I'd also find it very helpful if you could look at the ecosystem changes and summarize the result. This gives me confidence that there are no unexpected surprises and that this PR is ready to review. 

I understand that this will take more time on your side, but it will allow us to review your PRs more quickly.

danparizher:
Hi @MichaReiser, thanks for the feedback. I've updated the description for this PR and will continue to do so for future PRs!

crates/ruff_linter/src/rules/pandas_vet/helpers.rs
```diff
@@ -61,3 +78,46 @@ pub(super) fn test_expression(expr: &Expr, semantic: &SemanticModel) -> Resoluti
...
+        // Direct pandas module access
+        Expr::Name(name) => {
+            if let Some(binding_id) = semantic.resolve_name(name) {
+                let binding = semantic.binding(binding_id);
+                if let BindingKind::Import(import) = &binding.kind {
+                    return matches!(import.qualified_name().segments(), ["pandas"]);
+                }
+            }
+            false
+        }
```
ntBre: This looks a lot like the first part of the `match` in `test_expression`. Should we just recurse in `test_expression` for names, attributes, and calls? (And add `match` arms for attributes and calls like you have here)

crates/ruff_linter/src/rules/pandas_vet/mod.rs
```diff
@@ -269,6 +269,53 @@ mod tests {
     ",
         "PD011_pass_node_name"
     )]
+    #[test_case(
```
ntBre: Could we use the `paths` test runner just below this instead of `contents`? I think  it's a lot easier to read and modify files than these snippets.

You can probably put these all in one file if you isolate them in separate functions like:

```py
def f():
	import pandas as pd
	import numpy as np
	unique = np.unique_inverse([1, 2, 3, 2, 1])
	result = unique.values
```

at least I've seen that pattern in other cases.

ntBre (changes requested):
I made a couple of inline comments, but then I started looking at the ecosystem check, and it looks like we've gone too far in the opposite direction. The first  two cases I clicked on:

- [superset/charts/client_processing.py:110:14:](https://github.com/apache/superset/blob/4ddc3f14ed32c99805d124814f5d3e76f1a9922d/superset/charts/client_processing.py#L110) PD013 `.melt` is preferred to `.stack`; provides same functionality
- [superset/utils/csv.py:79:21:](https://github.com/apache/superset/blob/4ddc3f14ed32c99805d124814f5d3e76f1a9922d/superset/utils/csv.py#L79) PD008 Use `.loc` instead of `.at`. If speed is important, use NumPy.

seem like clear false negatives, at least to me. I guess we're having trouble detecting  them because the function arguments have been shadowed, but the parameters themselves are even annotated as pandas dataframes.

This doesn't seem like an improvement to me and seems to be a step toward making the rules less useful, as Micha mentioned on the issue.

I would probably lean toward closing this, especially since the user closed the issue as well.

danparizher:
I pushed a change with your feedback. Feel free to close if it makes more sense, since the original issue was also closed.

ntBre:
Thanks again for your work here and sorry for the delay in getting back to it. I think I will go ahead and close this for now based on the discussion above.

PR #20880, opened 2025-10-15
[`flake8-todos`] Extend TD003 to allow Jira-style issue IDs on the same line (`TD003`)

### LINKED ISSUE #20809: Extend TD003 to allow "Jira-style" issue IDs on the same line
### Summary

Rule: https://docs.astral.sh/ruff/rules/missing-todo-link/#missing-todo-link-td003

Extension to https://github.com/astral-sh/ruff/issues/8061, where this was changed for "GitHub-style" issue numbers, to allow them on the same line as the `TODO` comment.

### Current behavior
For issue _numbers_, this is allowed:
`# TODO: Move this part to the other place #6877`

but this is not:
`# TODO: Move this part to the other place RFFU-6877`

### Outcome
TD003 should accept issue IDs of the form `<PREFIX>-<NUMBER>` on the same line as the TODO comment, i.e.:

`# TODO: Move this part to the other place RFFU-6877`

danparizher:
## Summary

Fixes #20809. Tightens regex pattern and adds test coverage based on ecosystem analysis.

## Ecosystem Analysis

**51 TD003 violations removed** from real-world projects:

- Apache Airflow: 50 violations
- Apache Superset: 1 violation

These represent legitimate Jira-style issue IDs previously flagged as missing links.

### Pattern Specificity Fix

- **Before**: `[A-Z]+-\d+` (accepted single-letter keys like `A-1`)
- **After**: `[A-Z]{2,}-\d+` (requires minimum 2 uppercase letters)
- **Rationale**: Aligns with Jira project key requirements (2-10 uppercase letters)

### Test Coverage

Added test cases:

- ✅ Valid: `ABC-123`, `PROJ-456`, `RFFU-6877`
- ❌ Invalid: `A-1`, `abc-123`, `ABC123`, `123-ABC`

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
+++ b/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
@@ -244,6 +244,7 @@ static ISSUE_LINK_TODO_LINE_REGEX_SET: LazyLock<RegexSet> = LazyLock::new(|| {
     RegexSet::new([
         r"\s*(http|https)://.*", // issue link
         r"\s*#\d+.*",            // issue code - like "#003"
+        r"\s*[A-Z]+-\d+.*",      // Jira-style issue code - like "RFFU-6877"
     ])
     .unwrap()
 });
--- a/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
+++ b/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
@@ -248,7 +248,9 @@ static ISSUE_LINK_TODO_LINE_REGEX_SET: LazyLock<RegexSet> = LazyLock::new(|| {
     RegexSet::new([
         r"\s*(http|https)://.*", // issue link
         r"\s*#\d+.*",            // issue code - like "#003"
-        r"\s*[A-Z]{2,}-\d+.*",   // Jira-style issue code - like "RFFU-6877"
+        r"\s*[A-Z]{2,}-\d+\s*$", // Jira-style issue code at end of line - like "RFFU-6877"
+        r"\s*[A-Z]{2,}-\d+\s",   // Jira-style issue code followed by space - like "RFFU-6877 "
+        r"\s*\([A-Z]{2,}-\d+\)", // Jira-style issue code in parentheses - like "(RFFU-6877)"
     ])
     .unwrap()
 });
--- a/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
+++ b/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
@@ -249,8 +249,8 @@ static ISSUE_LINK_TODO_LINE_REGEX_SET: LazyLock<RegexSet> = LazyLock::new(|| {
         r"\s*(http|https)://.*", // issue link
         r"\s*#\d+.*",            // issue code - like "#003"
         r"\s*[A-Z]{2,}-\d+\s*$", // Jira-style issue code at end of line - like "RFFU-6877"
-        r"\s*[A-Z]{2,}-\d+\s",   // Jira-style issue code followed by space - like "RFFU-6877 "
         r"\s*\([A-Z]{2,}-\d+\)", // Jira-style issue code in parentheses - like "(RFFU-6877)"
+        r"\s*[A-Z]{2,}-\d+\s*:", // Jira-style issue code followed by colon - like "RFFU-6877:"
     ])
     .unwrap()
 });
--- a/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
+++ b/crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
@@ -100,6 +100,10 @@ impl Violation for MissingTodoAuthor {
 ///
 /// # TODO(charlie): this comment has an issue code (matches the regex `[A-Z]+\-?\d+`)
 /// # SIXCHR-003
+///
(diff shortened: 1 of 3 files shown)


MichaReiser:
Could you please take a look at the ecosystem changes and summarize your findings?

MichaReiser:
- [superset/tasks/async_queries.py:107:15:](https://github.com/apache/superset/blob/0b3fe3d60ccbb5d80370af3baafcd74879d34de6/superset/tasks/async_queries.py#L107) TD003 Missing issue link for this TODO

This is now a false negative

crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
```diff
@@ -244,6 +248,7 @@ static ISSUE_LINK_TODO_LINE_REGEX_SET: LazyLock<RegexSet> = LazyLock::new(|| {
     RegexSet::new([
         r"\s*(http|https)://.*", // issue link
         r"\s*#\d+.*",            // issue code - like "#003"
+        r"\s*[A-Z]{2,}-\d+.*",   // Jira-style issue code - like "RFFU-6877"
```
MichaReiser: Looking through the ecosystem reports, I think we should make this more strict. E.g. any of:

* Comes directly after todo (`TODO RFFU-6887` or `TODO: RFFU-6887)
* In parentheses `(RFFU-6877)`
* At the end of the line


@simonpercivall how do you format your Jira style issue codes that are easy to detect but don't lead to accidential false-negatives?

simonpercivall: Personally wouldn't be unhappy with requiring parentheses if at the end of the line.

We generally have just put it as the last item on the line, but with no (or arbitrary) separation. Consistency via a rule would just be helpful.

crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
```diff
@@ -244,6 +248,9 @@ static ISSUE_LINK_TODO_LINE_REGEX_SET: LazyLock<RegexSet> = LazyLock::new(|| {
     RegexSet::new([
         r"\s*(http|https)://.*", // issue link
         r"\s*#\d+.*",            // issue code - like "#003"
+        r"\s*[A-Z]{2,}-\d+\s*$", // Jira-style issue code at end of line - like "RFFU-6877"
+        r"\s*[A-Z]{2,}-\d+\s",   // Jira-style issue code followed by space - like "RFFU-6877 "
```
MichaReiser: This one still seems to permissive and makes the other regex patterns mostly redundant (and we still see the false positives). Can we limit the pattern to only be allowed right after the TODO or if followed by a `:`?

crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
MichaReiser: @danparizher can you explain the your reasoning for why you only allow the codes at the end of a line, in parentheses, or after a colon but e.g. not when it's directly after a `TODO`?

danparizher: @MichaReiser The regex patterns are intentionally restrictive to prevent false positives. If we matched issue codes directly after "TODO" (like `# TODO PROJ-123`), we'd incorrectly flag legitimate descriptions that happen to mention issue numbers, such as `# TODO: Fix bug ABC-123` or `# TODO: Working on PROJ-123 and PROJ-456`.

By requiring issue codes the way they are currently implemented, we ensure they're deliberately placed as references rather than accidentally mentioned in descriptive text. This follows the principle that it's better to miss some edge cases than to generate incorrect warnings.

crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
```diff
@@ -100,6 +100,10 @@ impl Violation for MissingTodoAuthor {
 ///
 /// # TODO(charlie): this comment has an issue code (matches the regex `[A-Z]+\-?\d+`)
 /// # SIXCHR-003
+///
+/// # TODO(charlie): PROJ-123 this comment has a Jira-style issue ID on the same line
```
MichaReiser, line 104: Looking at the tests, won't this still trigger a violation as it is the same as 

```
# TODO: PROJ-123 Another Jira-style example
```

crates/ruff_linter/src/rules/flake8_todos/rules/todos.rs
MichaReiser: I'm not sure how a code like `TODO PROJ-123` or `TODO: PROJ-123` would be an issue here. 

At the same time, I start to worry that explaining the behavior of this rule is becoming increasingly difficult.

@simonpercivall are there any other tools we could take inspiration from how to handle JIRA comments. E.g. does VS Code (or is there an extension) that allows you to click on JIRA issue codes? Do you know what formats it detects?

gcomneno:
I added snapshot fixtures for TD003 covering Jira-style issue keys on the same line as TODO, based on the formats discussed here (immediately after TODO, with/without a colon), plus guardrails (key mentioned in descriptive text, multiple keys, lowercase/invalid key). This should help tighten the regex without regressions.

Commit: https://github.com/gcomneno/ruff/commit/0f26e06706af5143cace1321925873e66ee50c1f

rgant:
It would be very nice if this change could be merged. I am switching from GitHub issues to Jira in my project and need this to make the TODO comments read similar to when I used GH issues.

PR #20881, opened 2025-10-15
[`flake8-pyi`] Fix PYI034 to not trigger on metaclasses (`PYI034`)

### LINKED ISSUE #20781: `PYI034` should not trigger on a metaclass
### Summary

This is what happens if I follow this rule in a metaclass in Python 3.14

Image

### Version

0.14.0

danparizher:
## Summary

Fixes #20781

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_pyi/rules/non_self_return_type.rs
+++ b/crates/ruff_linter/src/rules/flake8_pyi/rules/non_self_return_type.rs
@@ -50,15 +50,19 @@ use ruff_text_size::Ranged;
 /// 1. `__aiter__` methods that return `AsyncIterator`, despite the class
 ///    inheriting directly from `AsyncIterator`.
 ///
-/// Metaclasses are an exception for `__new__`. As mentioned in [PEP 673],
-/// `Self` is disallowed in metaclasses. The heuristic used to identify a
-/// metaclass-like `__new__` method signature is that it has:
+/// The rule attempts to avoid flagging methods on metaclasses, since
+/// [PEP 673] specifies that `Self` is disallowed in metaclasses. Ruff can
+/// detect a class as being a metaclass if it inherits from a stdlib
+/// metaclass such as `builtins.type` or `abc.ABCMeta`, and additionally
+/// infers that a class may be a metaclass if it has a `__new__` method
+/// with a similar signature to `type.__new__`. The heuristic used to
+/// identify a metaclass-like `__new__` method signature is that it:
 ///
-/// 1. Exactly 5 parameters (including `cls`)
-/// 1. Second parameter annotated with `str`
-/// 1. Third parameter annotated with a `tuple` type
-/// 1. Fourth parameter annotated with a `dict` type
-/// 1. Fifth parameter is keyword-variadic (`**kwargs`)
+/// 1. Has exactly 5 parameters (including `cls`)
+/// 1. Has a second parameter annotated with `str`
+/// 1. Has a third parameter annotated with a `tuple` type
+/// 1. Has a fourth parameter annotated with a `dict` type
+/// 1. Has a fifth parameter is keyword-variadic (`**kwargs`)
 ///
 /// For example, the following class would be detected as a metaclass, disabling
 /// the rule:


crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI034.pyi
```diff
@@ -252,3 +252,10 @@ from some_module import PotentialTypeVar
 class Generic5(list[PotentialTypeVar]):
     def __new__(cls: type[Generic5]) -> Generic5: ...
     def __enter__(self: Generic5) -> Generic5: ...
+
+
+# Test case based on issue #20781 - metaclass with complex __new__ signature
+class MetaclassInWhichSelfCannotBeUsed5(type):
+    def __new__(
+        cls, name: str, bases: tuple[type[Any], ...], attrs: dict[str, Any], **kwargs: Any
+    ) -> MetaclassInWhichSelfCannotBeUsed5: ...
```
AlexWaygood: This test passes on `main`. Are you able to add a test that does not pass on `main`, but does pass on this PR branch?

AlexWaygood (changes requested):
Thanks!

crates/ruff_linter/resources/test/fixtures/flake8_pyi/PYI034.pyi
danparizher: I've updated the test case to use `type(Protocol)` which triggers `IsMetaclass::Maybe`. The fix skips `IsMetaclass::Yes` and `IsMetaclass::Maybe` cases as required by PEP 673.

AlexWaygood: Thank you! Before we land this, I'd still like to wait for a bit more information on the issue to see if this is sufficient to fix the problem the user was experiencing

AlexWaygood: It sounds like the consensus on the issue is that we should implement option (2) I outlined in https://github.com/astral-sh/ruff/issues/20781#issuecomment-3410351921. @danparizher, are you interested in updating this PR to include that change? I think we should probably work the heuristic into the `ruff_python_semantic::analyze::class::is_metaclass()` function. If we see that the class (or any of its known superclasses) has a `__new__` method that matches that heuristic, we should declare that the class is "maybe a metaclass" (but not "definitely a metaclass").

AlexWaygood (approved):
Thanks, I pushed a few edits. The implementation LGTM now.

@ntBre, I think you mentioned on the issue that it might be good to document this heuristic? That does seem like it might be worthwhile. I'll punt this back to you now, if that's okay 😄

ntBre:
Thanks @AlexWaygood, this looks great! And you did all the hard work, I basically just copied your docstring into the rule docs :smile:

crates/ruff_linter/src/rules/flake8_pyi/rules/non_self_return_type.rs
```diff
@@ -50,6 +50,25 @@ use ruff_text_size::Ranged;
...
 ///
+/// Metaclasses are an exception for `__new__`. As mentioned in [PEP 673],
+/// `Self` is disallowed in metaclasses. The heuristic used to identify a
+/// metaclass-like `__new__` method signature is that it has:
+///
+/// 1. Exactly 5 parameters (including `cls`)
+/// 1. Second parameter annotated with `str`
+/// 1. Third parameter annotated with a `tuple` type
+/// 1. Fourth parameter annotated with a `dict` type
+/// 1. Fifth parameter is keyword-variadic (`**kwargs`)
```
AlexWaygood:
```suggestion
/// The rule attempts to avoid flagging methods on metaclasses, since
/// [PEP 673] specifies that `Self` is disallowed in metaclasses. Ruff can
/// detect a class as being a metaclass if it inherits from a stdlib
/// metaclass such as `builtins.type` or `abc.ABCMeta`, and additionally
/// infers that a class may be a metaclass if it has a `__new__` method
/// with a similar signature to `type.__new__`. The heuristic used to
/// identify a metaclass-like `__new__` method signature is that it:
///
/// 1. Has exactly 5 parameters (including `cls`)
/// 1. Has a second parameter annotated with `str`
/// 1. Has a third parameter annotated with a `tuple` type
/// 1. Has a fourth parameter annotated with a `dict` type
/// 1. Has a fifth parameter is keyword-variadic (`**kwargs`)
```

AlexWaygood:
> Thanks @AlexWaygood, this looks great! And you did all the hard work, I basically just copied your docstring into the rule docs 😄

Ah I have to share credit with @danparizher for that docstring, he wrote half of it!!

PR #20882, opened 2025-10-15
[ty] Fix match pattern value narrowing to use equality semantics

ericmarkmartin:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1349.

Fix match statement value patterns to use equality comparison semantics instead of incorrectly narrowing to literal types directly. Value patterns use equality for matching, and equality can be overridden, so we can't always narrow to the matched literal.

## Test Plan

Updated match.md with corrected expected types and an additional example with explanation

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/match.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/match.md
@@ -128,17 +128,13 @@ from typing import Literal
 class C:
     pass
 
-def _(x: Literal["foo", "bar", 42, b"foo"] | bool | complex):
+def _(x: Literal["foo", b"bar"] | int):
     match x:
-        case "foo" if reveal_type(x):  # revealed: Literal["foo"] | int | float | complex
+        case "foo" if reveal_type(x):  # revealed: Literal["foo"] | int
             pass
-        case 42 if reveal_type(x):  # revealed: int | float | complex
+        case b"bar" if reveal_type(x):  # revealed: Literal[b"bar"] | int
             pass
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/narrow/match.md
```diff
@@ -105,15 +134,15 @@ x = get_object()
 reveal_type(x)  # revealed: object
 
 match x:
-    case "foo" if reveal_type(x):  # revealed: Literal["foo"]
+    case "foo" if reveal_type(x):  # revealed: object
```
sharkdp: I think we should update these tests here to see that `x` in the guard expression *can* be a narrowed type.

crates/ty_python_semantic/resources/mdtest/narrow/match.md
```diff
@@ -131,13 +160,14 @@ reveal_type(x)  # revealed: object
 
 match x:
     case "foo" | 42 | None:
-        reveal_type(x)  # revealed: Literal["foo", 42] | None
+        reveal_type(x)  # revealed: object
```
sharkdp: Similar here. Can we maybe update the type of `x` (to a large union) to restore the spirit of this test.

crates/ty_python_semantic/resources/mdtest/narrow/match.md
```diff
@@ -153,13 +183,13 @@ x = get_object()
 reveal_type(x)  # revealed: object
 
 match x:
-    case "foo" | 42 | None if reveal_type(x):  # revealed: Literal["foo", 42] | None
+    case "foo" | 42 | None if reveal_type(x):  # revealed: object
```
sharkdp: Same here.

PR #20887, opened 2025-10-15
Remove `release` CI job

MichaReiser:
## Summary

We added the release CI job to catch issues with conditionally compiled code (`debug`, `cfg(debug_assertions)` back in April 2024 (https://github.com/astral-sh/ruff/pull/11182). 

But, we've since then also added `cargo-test-linux-release` (https://github.com/astral-sh/ruff/pull/14604), which not only compiles with the release profile but also runs all tests

https://github.com/astral-sh/ruff/blob/3d2ece2f42f24b0c18968d3a624d37eeb37525ff/.github/workflows/ci.yaml#L300-L331

That's why I think we can safely remove the release CI Job, as it's covered by cargo-test-linux-release

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -361,6 +361,37 @@ jobs:
           cargo nextest run --all-features --profile ci
           cargo test --all-features --doc
 
+  cargo-test-macos:
+    name: "cargo test (macos)"
+    runs-on: depot-windows-2022-16
+    needs: determine_changes
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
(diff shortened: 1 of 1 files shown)


sharkdp:
> That's why I think we can safely remove the release CI Job, as it's covered by cargo-test-linux-release

Looks like the release job was running on macOS though, so we lose *some* coverage with this change. That might be acceptable if we have other macOS builds, though.

MichaReiser:
> Looks like the release job was running on macOS though, so we lose some coverage with this change. That might be acceptable if we have other macOS builds, though.

As far as I understand, the reason it ran on macOS was that those used to be the fastest runners available. We can introduce a new CI job that runs all tests on macos if we want, that should still be much cheaper than another release job

.github/workflows/ci.yaml
```diff
@@ -361,6 +361,37 @@ jobs:
           cargo nextest run --all-features --profile ci
           cargo test --all-features --doc
 
+  cargo-test-macos:
+    name: "cargo test (macos)"
+    runs-on: depot-windows-2022-16
```
AlexWaygood, line 366: the MacOS job runs on Windows?

MichaReiser: yes, it's faster that way

PR #20888, opened 2025-10-15
[ty] Support dataclass-transform `field_specifiers`

sharkdp:
## Summary

Add support for the `field_specifiers` parameter on `dataclass_transform` decorator calls.

closes https://github.com/astral-sh/ty/issues/1068

## Conformance test results

All true positives :heavy_check_mark: 

## Ecosystem analysis

* `trio`: this is the kind of change that I would expect from this PR. The code makes use of a dataclass `Outcome` with a `_unwrapped: bool = attr.ib(default=False, eq=False, init=False)` field that is excluded from the `__init__` signature, so we now see a bunch of constructor-call-related errors going away.
* `home-assistant/core`: They have a `domain: str = attr.ib(init=False, repr=False)` field and then use
  ```py
    @domain.default
    def _domain_default(self) -> str:
        # …
  ```
  This accesses the `default` attribute on `dataclasses.Field[…]` with a type of `default: _T | Literal[_MISSING_TYPE.MISSING]`, so we get those "Object of type `_MISSING_TYPE` is not callable" errors. I don't really understand how that is supposed to work. Even if `_MISSING_TYPE` would be absent from that union, what does this try to call? pyright also issues an error and it doesn't seem to work at runtime? So this looks like a true positive?
* `attrs`: Similar here. There are some new diagnostics on code that tries to access `.validator` on a field. This *does* work at runtime, but I'm not sure how that is supposed to type-check (without a [custom plugin](https://github.com/python/mypy/blob/2c6c3959356674262d9b2c2dc43a33486e807a9c/mypy/plugins/attrs.py#L575-L602)). pyright errors on this as well.
* A handful of new false positives because we don't support `alias` yet

## Test Plan

Updated tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -638,6 +638,11 @@ impl<'db> Bindings<'db> {
                         let kw_only = if Program::get(db).python_version(db) >= PythonVersion::PY310
                         {
                             match kw_only {
+                                // We are more conservative here when turning the type for `kw_only`
+                                // into a bool, because a field specifier in a stub might use
+                                // `kw_only: bool = ...` and the truthiness of `...` is always true.
+                                // This is different from `init` above because may need to fall back
+                                // to `kw_only_default`, whereas `init_default` does not exist.
                                 Some(Type::BooleanLiteral(yes)) => Some(yes),
                                 _ => None,
                             }


crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -596,6 +598,48 @@ impl<'db> Bindings<'db> {
                         }
                     }
 
+                    function @ Type::FunctionLiteral(function_type)
+                        if dataclass_field_specifiers.contains(&function)
+                            || function_type.is_known(db, KnownFunction::Field) =>
+                    {
+                        let default = overload.parameter_type_by_name("default").unwrap_or(None);
```
sharkdp: the code inside this branch has just moved

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4509,10 +4518,39 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                infer_definition_types(db, class_definition)
+                    .declaration_type(class_definition)
+                    .inner_type()
+                    .as_class_literal()?
+                    .dataclass_params(db)
+                    .map(|params| SmallVec::from(params.field_specifiers(db)))
+            }
+
+            if let Some(specifiers) = field_specifiers(self.db(), self.index, self.scope()) {
+                self.dataclass_field_specifiers = specifiers;
```
sharkdp, line 4561: This is the key change. When we see an annotated assignment like the one for `id` in `C` …
```py
def fancy_field(*, init: bool = True, kw_only: bool = False) -> Any: ...
@dataclass_transform(field_specifiers=(fancy_field,))

def fancy_model[T](cls: type[T]) -> type[T]:
    # ...
    return cls

@fancy_model
class Person:
    id: int = fancy_field(init=False)
```
… we temporarily store a list of "active" field-specifiers (for this class: `[fancy_field]`) in the inference builder. Later, when inferring the call expression on the right-hand side of that assignment, we can pass down this list of field-specifiers into the call resolution logic in order to specialize calls to `fancy_field`.

sharkdp:
Support for metaclass-based models is actually missing here (and we generally don't support baseclass-based models yet). I think it makes sense to do this [in a second step](https://github.com/astral-sh/ruff/pull/20896), actually.

crates/ty_python_semantic/src/types.rs
```diff
@@ -618,12 +619,12 @@ impl<'db> PropertyInstanceType<'db> {
...
+    /// For the precise meaning of the fields, see [1].
     ///
     /// [1]: https://docs.python.org/3/library/dataclasses.html
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-    pub struct DataclassParams: u16 {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+    pub struct DataclassFlags: u16 {
         const INIT = 0b0000_0000_0001;
         const REPR = 0b0000_0000_0010;
         const EQ = 0b0000_0000_0100;
```
AlexWaygood: nit: it's a pre-existing issue, but most of our bitflags in this repo use bitshifts rather than octal to define the flags, which I find more readable, e.g.

https://github.com/astral-sh/ruff/blob/98d27c412810e157f8a65ea75726d66676628225/crates/ruff_python_ast/src/nodes.rs#L812-L823

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -596,6 +598,48 @@ impl<'db> Bindings<'db> {
                         }
                     }
 
+                    function @ Type::FunctionLiteral(function_type)
+                        if dataclass_field_specifiers.contains(&function)
```
sharkdp, line 602: Hm, this is probably too naive for overloaded functions (which is common for field specifiers)

crates/ty_python_semantic/src/types.rs
```diff
@@ -634,51 +635,79 @@ bitflags! {
...
+impl<'db> DataclassParams<'db> {
+    fn default_params(db: &'db dyn Db) -> Self {
+        Self::from_flags(db, DataclassFlags::default())
+    }
+
+    fn from_flags(db: &'db dyn Db, flags: DataclassFlags) -> Self {
+        let dataclasses_field = known_module_symbol(db, KnownModule::Dataclasses, "field")
+            .place
+            .ignore_possibly_unbound()
+            .unwrap_or_else(|| Type::none(db));
```
AlexWaygood: Why are we falling back to the Python `None` type here -- wouldn't we normally fall back to `Unknown` for this kind of thing? Seems probably worth a comment?

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -596,6 +598,48 @@ impl<'db> Bindings<'db> {
...
+                            || function_type.is_known(db, KnownFunction::Field) =>
+                    {
+                        let default = overload.parameter_type_by_name("default").unwrap_or(None);
+                        let default_factory = overload
+                            .parameter_type_by_name("default_factory")
+                            .unwrap_or(None);
+                        let init = overload.parameter_type_by_name("init").unwrap_or(None);
+                        let kw_only = overload.parameter_type_by_name("kw_only").unwrap_or(None);
```
AlexWaygood: I guess all these `.unwrap_or(None)` calls could also be `.unwrap_or_default()` calls, but IDK if that's actually clearer

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -272,6 +273,10 @@ pub(super) struct TypeInferenceBuilder<'db, 'ast> {
 
     /// `true` if all places in this expression are definitely bound
     all_definitely_bound: bool,
+
+    /// A list of `dataclass_transform` field specifiers that are "active" (when inferring
+    /// the right hand side of an annotated assignment in a class that is a dataclass).
+    dataclass_field_specifiers: SmallVec<[Type<'db>; 2]>,
```
AlexWaygood: why `2`, specifically?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4509,10 +4518,39 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+
+                let class_node = enclosing_scope.node().as_class()?;
+
+                let class_definition = index.expect_single_definition(class_node);
+                infer_definition_types(db, class_definition)
+                    .declaration_type(class_definition)
+                    .inner_type()
+                    .as_class_literal()?
+                    .dataclass_params(db)
+                    .map(|params| SmallVec::from(params.field_specifiers(db)))
```
AlexWaygood, line 4550: (`enclosing_scope.node().as_class()?` looks very cheap; I don't think it's any faster to explicitly check `enclosing_scope.kind()` before doing that call)

```suggestion
                let enclosing_scope = index.scope(scope.file_scope_id(db));
                let class_node = enclosing_scope.node().as_class()?;
                let class_definition = index.expect_single_definition(class_node);
                binding_type(db, class_definition)
                    .as_class_literal()?
                    .dataclass_params(db)
                    .map(|params| SmallVec::from(params.field_specifiers(db)))
```

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: nice solution!

...Though I wonder what happens if you have nested classes with different dataclass_transform functions... does such a concept even make sense? It might be an interesting test, anyway... curious what other type checkers do for this kind of thing...

```py
def fancy_field(*, init: bool = True, kw_only: bool = False) -> Any: ...

def other_fancy_field(*, init: bool = True, kw_only: bool = False) -> Any: ...

@dataclass_transform(field_specifiers=(fancy_field,))
def fancy_model[T](cls: type[T]) -> type[T]:
    # ...
    return cls

@dataclass_transform(field_specifiers=(other_fancy_field,))
def other_fancy_model[T](cls: type[T]) -> type[T]:
    # ...
    return cls

@fancy_model
class Person:
    id: int = fancy_field(init=False)

    @other_fancy_model
    class Whatever:
        id: int = other_fancy_field(init=False)
        is_this_legal: str = fancy_field(init=False)
```

crates/ty_python_semantic/src/types/call/bind.rs
sharkdp: ... no, that's fine

crates/ty_python_semantic/src/types.rs
sharkdp: Just plain wrong. I was using `Type::none(db)` as the fallback representation for missing field specifiers on this branch initially. That's my only explanation for why I possibly wrote that ;-)

crates/ty_python_semantic/src/types/call/bind.rs
sharkdp: This is pre-existing code, and I think I like the explicit form better?

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Maybe 3 would be better, so we can still handle Pydantic models inline? Otherwise, 1 also seems reasonable if we only want to be able to handle standard dataclasses inline.

* Standard dataclasses use 1 field specifier
* attrs uses 2 field specifiers
* pydantic uses 3 field specifiers
* SQLAlchemy uses 9 field specifiers, but I certainly don't want to reserve space for 9 `Type`s in every type inference builder, just in case.

AlexWaygood: 1, 2, or 3 all seem like reasonable choices, in that case! But I'd love for this to be captured in a comment, so other people reading through the source code don't have the same question as me ;)

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Added a test case for this. Did not yet look into how all kinds of libraries handle this, but we implement the behavior that makes the most sense to me: models only react to field-specifiers of their own dataclass transformer.

It works because nested classes will use their own `TypeInferenceBuilder`.

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -478,18 +478,154 @@ class Person:
...
+reveal_type(Person.__init__)  # revealed: (self: Person, name: str, *, age: int | None) -> None
+
+alice = Person("Alice", age=30)
+
+reveal_type(alice.id)  # revealed: int
+reveal_type(alice.name)  # revealed: str
+reveal_type(alice.age)  # revealed: int | None
+```
+
+### For base-class-based transformers
```
AlexWaygood, line 518: [insert "all your base are belong to us" joke]

crates/ty_python_semantic/src/types.rs
```diff
@@ -618,67 +621,95 @@ impl<'db> PropertyInstanceType<'db> {
...
+    }
+
+    fn from_flags(db: &'db dyn Db, flags: DataclassFlags) -> Self {
+        let dataclasses_field = known_module_symbol(db, KnownModule::Dataclasses, "field")
+            .place
+            .ignore_possibly_undefined()
+            .unwrap_or_else(Type::unknown);
+
+        Self::new(db, flags, vec![dataclasses_field].into_boxed_slice())
+    }
```
AlexWaygood, line 702: I am sort-of morbidly curious about what _does_ happen to our inference if you try to define a dataclass, with `dataclasses.field` for one of the fields, and you're using a custom typeshed that doesn't have `dataclasses.field` in it. That might be an interesting test, but it's almost certainly not worth spending time on it right now 😄

PR #20892, opened 2025-10-15
Auto-accept snapshot changes as part of typeshed-sync PRs

AlexWaygood:
## Summary

Some of our ty snapshots (in particular, several in the `ty_ide` crate that relate to autocompletions) are very sensitive to changes in typeshed. If typeshed changes the line number a class or function is defined on, that often leads to the snapshots in `ty_ide` needing to be updated manually before a typeshed-sync PR can be merged, which is a bit of a pain.

This PR adds a step to the `sync_typeshed.yaml` workflow that checks for changes to snapshots and automatically accepts them all before the PR is made. This should hopefully reduce the number of changes we need to make manually to sync-typeshed PRs. (We'll still need to review these PRs carefully to check that all the changes to the snapshots are _desirable_, but I still think that this will probably save us time overall.) The step is allowed to fail; if checking for changes to snapshots produces a nonzero exit code, the sync-typeshed PR should still be filed.

The `cargo insta` docs are here: https://insta.rs/docs/cli/

## Test Plan

I:
1. Ran `git checkout dc64c086336148c57db14060c86f448351710b2e -- crates/ty_python_semantic/resources/mdtest/snapshots`. This reverted all the snapshot changes from https://github.com/astral-sh/ruff/pull/20820, a PR which resulted in lots of snapshot changes -- multiple snapshot changes per mdtest file, in some cases.
2. Ran `cargo insta test --accept`. I observed that this command reverted all changes from step (1).

### DIFF
--- a/.github/workflows/sync_typeshed.yaml
+++ b/.github/workflows/sync_typeshed.yaml
@@ -29,7 +29,12 @@ on:
     - cron: "0 0 1,15 * *"
 
 env:
-  FORCE_COLOR: 1
+  # Don't set this flag globally for the workflow: it does strange things
+  # to the snapshots in the `cargo insta test --accept` step in the MacOS job.
+  #
+  # FORCE_COLOR: 1
+
+  CARGO_TERM_COLOR: always
   GH_TOKEN: ${{ github.token }}
 
   # The name of the upstream branch that the first worker creates,
@@ -88,6 +93,8 @@ jobs:
           git commit -m "Sync typeshed. Source commit: https://github.com/python/typeshed/commit/$(git -C ../typeshed rev-parse HEAD)" --allow-empty
       - name: Sync Linux docstrings
         if: ${{ success() }}
+        env:
+          FORCE_COLOR: 1
         run: |
           cd ruff
           ./scripts/codemod_docstrings.sh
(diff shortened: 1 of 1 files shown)


AlexWaygood:
I don't know why the Ruff ecosystem job is repeatedly failing on this PR (I tried rerunning it), but it seems unrelated.

AlexWaygood:
I kicked off a test run here using this branch: https://github.com/astral-sh/ruff/actions/runs/18533177734

AlexWaygood:
> I kicked off a test run here using this branch: [astral-sh/ruff/actions/runs/18533177734](https://github.com/astral-sh/ruff/actions/runs/18533177734)

It took a few iterations but I think it all looks good now:
- Workflow run: https://github.com/astral-sh/ruff/actions/runs/18535425113
- PR: https://github.com/astral-sh/ruff/pull/20899

AlexWaygood:
It doesn't look this worked, but I don't really understand why :/

The typeshed sync workflow last night ran the full test suite, and I can see snapshot tests failing as part of the run. But no snapshot changes were accepted afterwards; it just says

> error: test run failed

https://github.com/astral-sh/ruff/actions/runs/19381406943/job/55460969374

I've tried running the same command locally with the same environment variables set, and it always results in snapshots being updated at the end of the run. Not sure what's up here.

PR #20893, opened 2025-10-15
Update parser snapshots

AlexWaygood:
## Summary

#20867 had a merge race with #20848, which means that the parser snapshot tests are now failing on `main` and all PRs (see e.g. https://github.com/astral-sh/ruff/actions/runs/18530391729)

## Test Plan

`cargo insta test -p ruff_python_parser --accept`


ntBre:
Thank you! Just noticed and was starting a branch too!

ntBre (approved):

PR #20894, opened 2025-10-15
Don't use codspeed or depot runners in CI jobs on forks

AlexWaygood:
## Summary

Currently I get a GitHub notification every time I sync my Ruff fork with `astral-sh/ruff`. The notification informs me that [the `ci` workflow failed](https://github.com/AlexWaygood/ruff/actions/runs/18496851594), because it tried to run various jobs on Depot or codspeed runners that my fork doesn't have access to. This is kinda annoying, and -- while I don't really _need_ to sync my fork on regular basis, since I have the ability to create branches directly on the upstream repo -- I'm guessing this also occurs for all our third-party contributors who have forks of Ruff.

The solution to this is not to run CI jobs on forks if they require a Depot or Codspeed runner. Unfortunately there doesn't appear to be a way to say "run this workflow on all pushes to `main`, but skip the whole workflow if it's a fork" -- you have to include the `if github.repository == 'astral-sh/ruff'` check in each separate CI job inside the workflow.

Another solution would be to say that contributors who have forks should simply disable GitHub Actions altogether on their forks. But running CI from a fork can be pretty useful, especially if you're a third-party contributor, as a way to test changes without opening a PR upstream (which can result in notifications for the upstream maintainers, even if the PR is opened in draft mode).

## Test Plan

CI on this PR

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -237,9 +237,9 @@ jobs:
 
   cargo-test-linux:
     name: "cargo test (linux)"
-    runs-on: depot-ubuntu-22.04-16
+    runs-on: ${{ github.repository == 'astral-sh/ruff' && 'depot-ubuntu-22.04-16' || 'ubuntu-latest' }}
     needs: determine_changes
-    if: ${{ github.repository == 'astral-sh/ruff' && !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -239,7 +239,7 @@ jobs:
     name: "cargo test (linux)"
     runs-on: depot-ubuntu-22.04-16
     needs: determine_changes
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
+    if: ${{ github.repository == 'astral-sh/ruff' && !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
```
MichaReiser: I think I slightly prefer `github.event.pull_request.head.repo.fork == false` as it's more explicit. 

An alternative is to disable all CI jobs for forks by adding the same condition at the very top of the file

AlexWaygood: > An alternative is to disable all CI jobs for forks by adding the same condition at the very top of the file

hmm, where exactly would I put this condition at the top of the file? I'd love to do that if it's possible, but I didn't think it was

AlexWaygood: > I think I slightly prefer `github.event.pull_request.head.repo.fork == false` as it's more explicit.

I don't think this would fix the issue I'm trying to fix, because the specific issue I'm trying to fix is that these jobs are being run on pushes to the `main` branch of the `AlexWaygood/ruff` repo. I.e., that would be a `push` event rather than a `pull_request` event.

But I'd also like these jobs that can't be run on forks disabled if I make a pull request against my own fork (which I don't do that often, but have done in the past, for testing purposes!) 😄

The expression I'm using here is basically identical to the example GitHub has in its docs for `if` conditions, FWIW: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-only-run-job-for-specific-repository. And it's also what we already use elsewhere in this workflow file: https://github.com/astral-sh/ruff/blob/4b7f184ab7848fee4f6641600b4f52a06e772e7a/.github/workflows/ci.yaml#L1015

MichaReiser: Well... 

I guess the ideal would be to use a different runner based on whether this is a fork or not. But I don't know how easily this is possible (and it won't work with codspeed).

AlexWaygood: > I guess the ideal would be to use a different runner based on whether this is a fork or not. But I don't know how easily this is possible (and it won't work with codspeed).

With Claude's help, I think I figured this out! (Except for Codspeed, of course)

PR #20897, opened 2025-10-15
[ty] add legacy namespace package support

Gankra:
Detect legacy namespace packages and treat them like namespace packages when looking them up as the *parent* of the module we're interested in. In all other cases treat them like a regular package.

(This PR is coauthored by @MichaReiser in a shared coding session)

Fixes https://github.com/astral-sh/ty/issues/838

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/resolver.rs
+++ b/crates/ty_python_semantic/src/module_resolver/resolver.rs
@@ -982,8 +982,9 @@ where
         if is_regular_package {
             // This is the only place where we need to consider the existence of legacy namespace
             // packages, as we are explicitly searching for the *parent* package of the module
-            // we actually want. Here, such a package should be treated as a namespace package.
-            // In all other contexts it acts like a normal package and needs no special handling.
+            // we actually want. Here, such a package should be treated as a PEP-420 ("modern")
+            // namespace package. In all other contexts it acts like a normal package and needs
+            // no special handling.
             in_namespace_package = is_legacy_namespace_package(&package_path, resolver_state);
         } else if package_path.is_directory(resolver_state)
             // Pure modules hide namespace packages with the same name
@@ -1023,9 +1024,9 @@ where
 
 /// Determines whether a package is a legacy namespace package.
 ///
-/// Before PEP 420 introduce proper namespace packages, the ecosystem developed
+/// Before PEP 420 introduced implicit namespace packages, the ecosystem developed
 /// its own form of namespace packages. These legacy namespace packages continue to persist
-/// in modern codebases because they work with ancient pythons and if it ain't broke don't fix it.
+/// in modern codebases because they work with ancient Pythons and if it ain't broke, don't fix it.
 ///
 /// A legacy namespace package is distinguished by having an `__init__.py` that contains an
 /// expression to the effect of:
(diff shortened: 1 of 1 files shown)


Gankra:
I'm now reasonably happy with the shape of the implementation, main place to quibble is how much elbow grease to put into the visitor that tries to find the idiom.

MichaReiser:
For added context: This makes airflow work out-of-the box (whereas it wasn't even possible to set up the import paths correctly before) and fixes about 5k unresolved import errors

MichaReiser (approved):
Thanks for finishing this up. Looks good to me but I'll leave this open for a little longer in case someone else wants to take a look

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -977,7 +980,11 @@ where
         let is_regular_package = package_path.is_regular_package(resolver_state);
 
         if is_regular_package {
-            in_namespace_package = false;
+            // This is the only place where we need to consider the existence of legacy namespace
+            // packages, as we are explicitly searching for the *parent* package of the module
+            // we actually want. Here, such a package should be treated as a namespace package.
```
AlexWaygood: to better distinguish the two kinds of namespace packages?

```suggestion
            // we actually want. Here, such a package should be treated as a PEP-420 ("modern") namespace package.
```

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1014,6 +1021,61 @@ where
     })
 }
 
+/// Determines whether a package is a legacy namespace package.
+///
+/// Before PEP 420 introduce proper namespace packages, the ecosystem developed
```
AlexWaygood:
```suggestion
/// Before PEP 420 introduced implicit namespace packages, the ecosystem developed
```

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1014,6 +1021,61 @@ where
...
+/// its own form of namespace packages. These legacy namespace packages continue to persist
+/// in modern codebases because they work with ancient pythons and if it ain't broke don't fix it.
```
AlexWaygood:
```suggestion
/// in modern codebases because they work with ancient Pythons and if it ain't broke don't, fix it.
```

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1014,6 +1021,61 @@ where
...
+/// expression to the effect of:
+///
+/// ```python
+/// __path__ = __import__("pkgutil").extend_path(__path__, __name__)
+/// ```
+///
+/// The resulting package simultaneously has properties of both regular packages and namespace ones:
+///
+/// * Like regular packages, `__init__.py` is defined and can contain items other than submodules
+/// * Like namespace packages, multiple copies of the package may exist with different submodules
```
AlexWaygood:
```suggestion
/// * Like implicit namespace packages, multiple copies of the package may exist with different submodules,
///   and these submodules will be merged into one namespace at runtime by the interpreter
```

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -1014,6 +1021,61 @@ where
     })
 }
 
+/// Determines whether a package is a legacy namespace package.
```
AlexWaygood, line 1025: this comment is fantastic, thanks for writing it up!!

Issue #20901, opened 2025-10-15
Consider using `Diagnostic`s for semantic syntax errors


ntBre:
While reviewing #20682 today, I  wanted to suggest emitting a diagnostic about alternate `match` patterns binding different variables just like rustc does:

```
error[E0408]: variable `a` is not bound in all patterns
 --> src/main.rs:4:5
  |
4 |     [] | [a, b, c] | [d, e, f]=> {}
  |     ^^    -          ^^^^^^^^^ pattern doesn't bind `a`
  |     |     |
  |     |     variable not in all patterns
  |     pattern doesn't bind `a`
```

However, we don't have direct access to the `Diagnostic` in the `SemanticSyntaxChecker`. Instead, we can currently only `report_semantic_error` to the parent visitor. I don't think we can simply replace `report_semantic_error` with `report_diagnostic`, for example, because Ruff's implementation of `report_semantic_error` needs to convert some semantic errors into rule violations, unlike ty's implementation.

Another wrinkle is that Ruff and ty use a different helper function to convert errors into diagnostics, [`create_syntax_error_diagnostic`](https://github.com/astral-sh/ruff/blob/583ac216c447436db942d6257b60e8d875e9584c/crates/ruff_linter/src/message/mod.rs#L36) vs [`Diagnostic::invalid_syntax`](https://github.com/astral-sh/ruff/blob/583ac216c447436db942d6257b60e8d875e9584c/crates/ruff_db/src/diagnostic/mod.rs#L90). The main difference is where the error message ends up:

```
> ruff check try.py
invalid-syntax: name capture `x` makes remaining patterns unreachable
 --> try.py:2:10
  |
1 | match 42:
2 |     case x: ...
  |          ^
3 |     case y: ...
  |

Found 1 error.
> uvx ty check try.py
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
Checking ------------------------------------------------------------ 1/1 files                                                                                                 error[invalid-syntax]
 --> try.py:2:10
  |
1 | match 42:
2 |     case x: ...
  |          ^ name capture `x` makes remaining patterns unreachable
3 |     case y: ...
  |

Found 1 diagnostic
```

This is a bit of a separate issue, and I kind of think we should standardize on the Ruff version anyway.

[...]

PR #20903, opened 2025-10-15
Standardize syntax error construction

ntBre:
Summary
--

This PR unifies the two different ways Ruff and ty construct syntax errors. Ruff has been storing the primary message in the diagnostic itself, while ty attached the message to the primary annotation:

```
> ruff check try.py
invalid-syntax: name capture `x` makes remaining patterns unreachable
 --> try.py:2:10
  |
1 | match 42:
2 |     case x: ...
  |          ^
3 |     case y: ...
  |

Found 1 error.
> uvx ty check try.py
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
Checking ------------------------------------------------------------ 1/1 files                                                                                                 
error[invalid-syntax]
 --> try.py:2:10
  |
1 | match 42:
2 |     case x: ...
  |          ^ name capture `x` makes remaining patterns unreachable
3 |     case y: ...
  |

Found 1 diagnostic
```

I think there are benefits to both approaches, and I do like ty's version, but I feel like we should pick one (and it might help with #20901 eventually). I slightly prefer Ruff's version, so I went with that. Hopefully this isn't too controversial, but I'm happy to close this if it is.

Note that this shouldn't change any other diagnostic formats in ty because [`Diagnostic::primary_message`](https://github.com/astral-sh/ruff/blob/98d27c412810e157f8a65ea75726d66676628225/crates/ruff_db/src/diagnostic/mod.rs#L177) was already falling back to the primary annotation message if the diagnostic message was empty. As a result, I think this change will partially resolve the FIXME therein.

Test Plan
--

Existing tests with updated snapshots

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/semantic_syntax_erro…_-_Semantic_syntax_erro…_-_`async`_comprehensio…_-_Python_3.10_(96aa8ec77d46553d).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/semantic_syntax_erro…_-_Semantic_syntax_erro…_-_`async`_comprehensio…_-_Python_3.10_(96aa8ec77d46553d).snap
@@ -33,13 +33,13 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/diagnostics/semantic_syn
 # Diagnostics
 
 ```
-error[invalid-syntax]
+error[invalid-syntax]: cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax was added in 3.11)
  --> src/mdtest_snippet.py:6:19
   |
 4 | async def f():
 5 |     # error: 19 [invalid-syntax] "cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax…
 6 |     return {n: [x async for x in elements(n)] for n in range(3)}
-  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax was added in 3.11)
+  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
 7 | async def test():
 8 |     # error: [not-iterable] "Object of type `range` is not async-iterable"
   |
--- a/crates/ruff/src/diagnostics.rs
+++ b/crates/ruff/src/diagnostics.rs
@@ -13,7 +13,6 @@ use log::{debug, warn};
 use ruff_db::diagnostic::Diagnostic;
 use ruff_linter::codes::Rule;
 use ruff_linter::linter::{FixTable, FixerResult, LinterResult, ParseSource, lint_fix, lint_only};
-use ruff_linter::message::create_syntax_error_diagnostic;
 use ruff_linter::package::PackageRoot;
 use ruff_linter::pyproject_toml::lint_pyproject_toml;
 use ruff_linter::settings::types::UnsafeFixes;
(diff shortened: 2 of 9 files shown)


AlexWaygood:
> Merging #20903 will **improve performances by 6.51%**

yeah, that's just a merge race with https://github.com/astral-sh/ruff/pull/20377, judging by the footnote that Codspeed put at the bottom of that comment 😆

ntBre:
Yes good call, I don't think this should affect performance :laughing:

crates/ty_python_semantic/resources/mdtest/snapshots/semantic_syntax_erro…_-_Semantic_syntax_erro…_-_`async`_comprehensio…_-_Python_3.10_(96aa8ec77d46553d).snap
```diff
@@ -33,13 +33,13 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/diagnostics/semantic_syn
...
+error[invalid-syntax]: cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax was added in 3.11)
  --> src/mdtest_snippet.py:6:19
   |
 4 | async def f():
 5 |     # error: 19 [invalid-syntax] "cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax…
 6 |     return {n: [x async for x in elements(n)] for n in range(3)}
-  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax was added in 3.11)
+  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
 7 | async def test():
 8 |     # error: [not-iterable] "Object of type `range` is not async-iterable"
```
AlexWaygood, line 44: Something like this might be nice here (but not blocking):

```
error[invalid-syntax]: cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10
 --> src/mdtest_snippet.py:6:19
  |
4 | async def f():
5 |     # error: 19 [invalid-syntax] "cannot use an asynchronous comprehension inside of a synchronous comprehension on Python 3.10 (syntax…
6 |     return {n: [x async for x in elements(n)] for n in range(3)}
  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ Syntax was added in 3.11
7 | async def test():
8 |     # error: [not-iterable] "Object of type `range` is not async-iterable"
```

AlexWaygood (approved):

crates/ty_python_semantic/resources/mdtest/snapshots/semantic_syntax_erro…_-_Semantic_syntax_erro…_-_`async`_comprehensio…_-_Python_3.10_(96aa8ec77d46553d).snap
ntBre: I was thinking about that too, or an info sub-diagnostic. I think either option will need something like https://github.com/astral-sh/ruff/issues/20901, though.

ntBre: (or just a separate constructor from the very generic `invalid_syntax` for unsupported syntax errors)

PR #20904, opened 2025-10-15
[ty] mdtests with external dependencies

sharkdp:
## Summary

This PR adds the possibility to write mdtests that specify external dependencies in a `project` section of TOML blocks. For example, here is a test that makes sure that we understand Pydantic's dataclass-transform setup:

````markdown
```toml
[environment]
python-version = "3.12"
python-platform = "linux"

[project]
dependencies = ["pydantic==2.12.2"]
```

```py
from pydantic import BaseModel

class User(BaseModel):
    id: int
    name: str

user = User(id=1, name="Alice")
reveal_type(user.id)  # revealed: int
reveal_type(user.name)  # revealed: str

# error: [missing-argument] "No argument provided for required parameter `name`"
invalid_user = User(id=2)
```
````

## How?

Using the `python-version` and the `dependencies` fields from the Markdown section, we generate a `pyproject.toml` file, write it to a temporary directory, and use `uv sync` to install the dependencies into a virtual environment. We then copy the Python source files from that venv's `site-packages` folder to a corresponding directory structure in the in-memory filesystem. Finally, we configure the search paths accordingly, and run the mdtest as usual.

I fully understand that there are valid concerns here:
* Doesn't this require network access? (yes, it does)
* Is this fast enough? (`uv` caching makes this almost unnoticeable, actually)
* Is this deterministic? ~~(probably not, package resolution can depend on the platform you're on)~~ (yes, hopefully)

For this reason, this first version is opt-in, locally. ~~We don't even run these tests in CI (even though they worked fine in a previous iteration of this PR).~~ You need to set `MDTEST_EXTERNAL=1`, or use the new `-e/--enable-external` command line option of the `mdtest.py` runner. For example:
```bash
# Skip mdtests with external dependencies (default):
uv run crates/ty_python_semantic/mdtest.py

# Run all mdtests, including those with external dependencies:
uv run crates/ty_python_semantic/mdtest.py -e

# Only run the `pydantic` tests. Use `-e` to make sure it is not skipped:
uv run crates/ty_python_semantic/mdtest.py -e pydantic
```

## Why?

I believe that this can be a useful addition to our testing strategy, which lies somewhere between ecosystem tests and normal mdtests. Ecosystem tests cover much more code, but they have the disadvantage that we only see second- or third-order effects via diagnostic diffs. If we unexpectedly gain or lose type coverage somewhere, we might not even notice (assuming the gradual guarantee holds, and ecosystem code is mostly correct). Another disadvantage of ecosystem checks is that they only test checked-in code that is usually correct. However, we also want to test what happens on wrong code, like the code that is momentarily written in an editor, before fixing it. On the other end of the spectrum we have normal mdtests, which have the disadvantage that they do not reflect the reality of complex real-world code. We experience this whenever we're surprised by an ecosystem report on a PR.

That said, these tests should not be seen as a replacement for either of these things. For example, we should still strive to write detailed self-contained mdtests for user-reported issues. But we might use this new layer for regression tests, or simply as a debugging tool. It can also serve as a tool to document our support for popular third-party libraries.

## Test Plan

* I've been locally using this for a couple of weeks now.
* `uv run crates/ty_python_semantic/mdtest.py -e`

### DIFF
--- a/crates/ty_test/src/external_dependencies.rs
+++ b/crates/ty_test/src/external_dependencies.rs
@@ -20,11 +20,15 @@ pub(crate) fn setup_venv(
         .tempdir()
         .context("Failed to create temporary directory for mdtest virtual environment")?;
 
-    let temp_path = SystemPath::from_std_path(temp_dir.path())
+    // Canonicalize here to fix problems with `.strip_prefix()` later on Windows
+    let temp_dir_path = dunce::canonicalize(temp_dir.path())
+        .context("Failed to canonicalize temporary directory path")?;
+
+    let temp_path = SystemPath::from_std_path(&temp_dir_path)
         .ok_or_else(|| {
             anyhow!(
                 "Temporary directory path is not valid UTF-8: {}",
-                temp_dir.path().display()
+                temp_dir_path.display()
             )
         })?
         .to_path_buf();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4557,6 +4557,7 @@ dependencies = [
  "anyhow",
  "camino",
  "colored 3.0.0",
+ "dunce",
  "insta",
  "memchr",
  "path-slash",
--- a/crates/ty_test/Cargo.toml
+++ b/crates/ty_test/Cargo.toml
@@ -25,6 +25,7 @@ ty_vendored = { workspace = true }
 
 anyhow = { workspace = true }
 camino = { workspace = true }
+dunce = { workspace = true }
 colored = { workspace = true }
 insta = { workspace = true, features = ["filters"] }
 memchr = { workspace = true }


crates/ty_test/src/external_dependencies.rs
```diff
@@ -0,0 +1,212 @@
...
+    venv_path: &SystemPath,
+    dest_venv_path: &SystemPath,
+    python_version: PythonVersion,
+) -> Result<()> {
+    use std::fs;
+
+    // Determine the site-packages path within the venv
+    let (site_packages_path, python_dir_name) = if cfg!(target_os = "windows") {
+        let sp_path = venv_path.join("Lib").join("site-packages");
+        (sp_path, String::new()) // Windows doesn't need the python version in the path
```
sharkdp: I have not tested this, this part was written by Claude. Happy to remove it or leave it in, just in case it's correct.

crates/ty_python_semantic/resources/mdtest/external/attrs.md
```diff
@@ -0,0 +1,77 @@
+# attrs
+
+```toml
+[environment]
+python-version = "3.13"
```
MichaReiser, line 5: We could combine it (and require) a `python-platform` and pass the platform to uv sync's `--python-platform`. This should give us predictable output and should be reproducible across platforms for as long as the dependencies build on all platforms (I think)

crates/ty_test/src/external_dependencies.rs
MichaReiser: Could we reuse some of the logic from here https://github.com/astral-sh/ruff/blob/9cb5483e70166ec4f7b07bb45b692b9132442c1a/crates/ty_test/src/lib.rs#L291-L297

or maybe better, use ty's virtual environment detection https://github.com/astral-sh/ruff/blob/9cb5483e70166ec4f7b07bb45b692b9132442c1a/crates/ty_python_semantic/src/site_packages.rs#L394

crates/ty_test/src/external_dependencies.rs
```diff
@@ -0,0 +1,212 @@
...
+    db.create_directory_all(&dest_site_packages)
+        .context("Failed to create site-packages directory in database")?;
+
+    // Recursively copy all files from site-packages
+    copy_directory_recursive(db, &site_packages_path, &dest_site_packages)?;
+
+    Ok(())
+}
+
+fn copy_directory_recursive(db: &mut Db, src: &SystemPath, dest: &SystemPath) -> Result<()> {
```
MichaReiser, line 132: It might be worth adding a method like this to `MemoryFileSystem`, given that we have the same/something similar in 

https://github.com/astral-sh/ruff/blob/9cb5483e70166ec4f7b07bb45b692b9132442c1a/crates/ruff_benchmark/src/real_world_projects.rs#L315-L374

MichaReiser (approved):
This is great.

I think we should run the tests in CI or we'll keep regressing those tests.

crates/ty_test/src/external_dependencies.rs
MichaReiser: This would also be useful in `ty_wasm` if we want to support external dependencies in the playground (I think)

sharkdp: I spent an hour trying to unify these two use cases, but ultimately failed. Either it resulted in more code overall, or I ran into `sync` problems that I didn't understand :disappointed:

crates/ty_test/src/external_dependencies.rs
sharkdp: Thank you, reusing `site_packages.rs` functionality makes this much cleaner now.

sharkdp:
> I think we should run the tests in CI or we'll keep regressing those tests.

I mainly didn't want to force this on anyone given that it's still relatively experimental. I was planning to enable it in CI later, but doing it now is fine with me, too.

MichaReiser:
I haven't tested it but this is what we use in most tempdir tests:

https://github.com/astral-sh/ruff/blob/646ad154c37fbec184805b9b36626b7e8782a16c/crates/ty/tests/file_watching.rs#L379-L386

sharkdp:
> I haven't tested it but this is what we use in most tempdir tests:

There was a problem with a UNC prefix, which is why I used `dunce::canonicalize` now. Seems to work.

PR #20905, opened 2025-10-15
Move TOML indent size config

esadek:
Move TOML indent size configuration in `.editorconfig`. No need to have separate `indent_size = 4` configurations.


MichaReiser (approved):
Thank you

PR #20907, opened 2025-10-16
[`flake8-django`] Apply `DJ001` to annotated fields

### LINKED ISSUE #20906: DJ001 not raised when field has type annotation
### Summary

```python
from django.db import models


class TestModel(models.Model):
    charfield1 = models.CharField(null=True) # Raises DJ001
    charfield2: models.CharField[str, str] = models.CharField(null=True) # No error
```

```
$ ruff check --select=DJ001 test.py
DJ001 Avoid using `null=True` on string-based fields such as `CharField`
 --> test.py:5:18
  |
4 | class TestModel(models.Model):
5 |     charfield1 = models.CharField(null=True)
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
6 |     charfield2: models.CharField[str, str] = models.CharField(null=True)
  |

Found 1 error.
```

I discovered this when using https://github.com/typeddjango/django-stubs with pyright, which doesn't detect the types correctly without explicit annotations.

### Version

ruff 0.14.0

danparizher:
## Summary

Fixes #20906

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_django/DJ001.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_django/DJ001.py
@@ -48,37 +48,7 @@ class CorrectModel(models.Model):
     urlfieldu = models.URLField(max_length=255, null=True, blank=True, unique=True)
 
 
-class IncorrectModelWithAnnotations(models.Model):
-    charfield: models.CharField[str, str] = models.CharField(max_length=255, null=True)
-    textfield: models.TextField[str, str] = models.TextField(max_length=255, null=True)
-    slugfield: models.SlugField[str, str] = models.SlugField(max_length=255, null=True)
-    emailfield: models.EmailField[str, str] = models.EmailField(max_length=255, null=True)
-    filepathfield: models.FilePathField[str, str] = models.FilePathField(max_length=255, null=True)
-    urlfield: models.URLField[str, str] = models.URLField(max_length=255, null=True)
-
-
 class IncorrectModelWithSimpleAnnotations(models.Model):
     charfield: models.CharField = models.CharField(max_length=255, null=True)
     textfield: models.TextField = models.TextField(max_length=255, null=True)
     slugfield: models.SlugField = models.SlugField(max_length=255, null=True)
-    emailfield: models.EmailField = models.EmailField(max_length=255, null=True)
-    filepathfield: models.FilePathField = models.FilePathField(max_length=255, null=True)
-    urlfield: models.URLField = models.URLField(max_length=255, null=True)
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/flake8_django/rules/nullable_model_string_field.rs
```diff
@@ -60,16 +60,30 @@ pub(crate) fn nullable_model_string_field(checker: &Checker, body: &[Stmt]) {
...
+                    checker.report_diagnostic(
+                        DjangoNullableModelStringField {
+                            field_name: field_name.to_string(),
+                        },
+                        value.range(),
+                    );
+                }
+            }
+            _ => continue,
         }
```
MichaReiser: We should avoid duplicating that much code. All we need is to extract value. Something like this should work

```suggestion
        let value = match statement {
            Stmt::Assign(ast::StmtAssign { value, .. }) => value,
            Stmt::AnnAssign(ast::StmtAnnAssign {
                value: Some(value), ..
            }) => value,
            _ => continue,
        };
        
        
        if let Some(field_name) = is_nullable_field(value, checker.semantic()) {
            checker.report_diagnostic(
                DjangoNullableModelStringField {
                    field_name: field_name.to_string(),
                },
                value.range(),
            );
        }
```

crates/ruff_linter/resources/test/fixtures/flake8_django/DJ001.py
```diff
@@ -46,3 +46,39 @@ class CorrectModel(models.Model):
...
+    urlfield: models.URLField[str, str] = models.URLField(max_length=255, null=False, blank=True)
+
+    charfieldu: models.CharField[str, str] = models.CharField(max_length=255, null=True, blank=True, unique=True)
+    textfieldu: models.TextField[str, str] = models.TextField(max_length=255, null=True, blank=True, unique=True)
+    slugfieldu: models.SlugField[str, str] = models.SlugField(max_length=255, null=True, blank=True, unique=True)
+    emailfieldu: models.EmailField[str, str] = models.EmailField(max_length=255, null=True, blank=True, unique=True)
+    filepathfieldu: models.FilePathField[str, str] = models.FilePathField(
+        max_length=255, null=True, blank=True, unique=True
+    )
+    urlfieldu: models.URLField[str, str] = models.URLField(max_length=255, null=True, blank=True, unique=True)
```
MichaReiser: Do we need all those tests? They seem redundant to me, considering that we ignore the annotation.

ntBre (approved):
Looks good to me, if Micha's happy with it!

PR #20908, opened 2025-10-16
[ty] Check typeshed VERSIONS for parent modules when reporting failed stdlib imports

Gankra:
This is a drive-by improvement that I stumbled backwards into while looking into 

* https://github.com/astral-sh/ty/issues/296

I was writing some simple tests for "thing not in old version of stdlib" diagnostics and checked what was added in 3.14, and saw `compression.zstd` and to my surprise discovered that `import compression.zstd` and `from compression import zstd` had completely different quality diagnostics.

This is because `compression` and `compression.zstd` were *both* introduced in 3.14, and so per VERSIONS policy only an entry for `compression` was added, and so we don't actually have any definite info on `compression.zstd` and give up on producing a diagnostic. However the `from compression import zstd` form fails on looking up `compression` and we *do* have an exact match for that, so it gets a better diagnostic!

(aside: I have now learned about the VERSIONS format and I *really* wish they would just enumerate all the submodules but, oh well!)

The fix is, when handling an import failure, if we fail to find an exact match *we requery with the parent module*. In cases like `compression.zstd` this lets us at least identify that, hey, not even `compression` exists, and luckily that fixes the whole issue. In cases where the parent module and submodule were introduced at different times then we may discover that the parent module is in-range and that's fine, we don't produce the richer stdlib diagnostic.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/basic.md_-_Structures_-_Attempting_to_import…_(dba22bd97137ee38).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/basic.md_-_Structures_-_Attempting_to_import…_(dba22bd97137ee38).snap
@@ -12,10 +12,10 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/import/basic.md
 ## mdtest_snippet.py
 
 ```
-1 | import compression.zstd # error: [unresolved-import]
+1 | import compression.zstd  # error: [unresolved-import]
 2 | from compression import zstd  # error: [unresolved-import]
-3 | import compression.fakebutwhocansay # error: [unresolved-import]
-4 | from compression import fakebutwhocansay # error: [unresolved-import]
+3 | import compression.fakebutwhocansay  # error: [unresolved-import]
+4 | from compression import fakebutwhocansay  # error: [unresolved-import]
 ```
 
 # Diagnostics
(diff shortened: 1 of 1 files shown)


Gankra:
(Broken into two commits so you can see the before and after on the snapshot)

Gankra:
Can't help but wonder if we could statically precompute a Complete VERSIONS map...

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4786,25 +4786,36 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-                            "The stdlib module `{module_name}` is only available on Python {version_range}",
-                            version_range = version_range.diagnostic_display(),
-                        ));
-                        add_inferred_python_version_hint_to_diagnostic(
-                            self.db(),
-                            &mut diagnostic,
-                            "resolving modules",
-                        );
-                        return;
+                loop {
```
MichaReiser: Nit: Use `module_name.ancestors()`?

crates/ty_python_semantic/resources/mdtest/diagnostics/newer_python.md
AlexWaygood, line 1: Ah nice -- can you move these existing tests to this file so that all our tests for this feature are in the same place?

https://github.com/astral-sh/ruff/blob/5fb142374daec49d53bcc00126bf485e82c015a6/crates/ty_python_semantic/resources/mdtest/import/basic.md?plain=1#L180-L207

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: +1

AlexWaygood (approved):
Ahh thank you!!

crates/ty_python_semantic/resources/mdtest/diagnostics/newer_python.md
Gankra: I did this but I completely misread the direction you wanted this to go so instead my new test file is deleted and I added a new test to `basic.md`

crates/ty_python_semantic/src/types/infer/builder.rs
Gankra: rip `loop` 😢

crates/ty_python_semantic/resources/mdtest/diagnostics/newer_python.md
AlexWaygood: Haha, I don't mind either way, I'd just like all the tests for the feature to be together!

PR #20909, opened 2025-10-16
[ty] Add version hint for failed stdlib accesses

Gankra:
This is the ultra-minimal implementation of

* https://github.com/astral-sh/ty/issues/296

that was previously discussed as a good starting point. In particular we don't actually bother trying to figure out the exact python versions, but we still mention "hey btw for No Reason At All... you're on python 3.10" when you try to access something that has a definition rooted in the stdlib that we believe exists sometimes.

### DIFF
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -3165,19 +3165,25 @@ pub(super) fn hint_if_stdlib_attribute_exists_on_other_versions(
     let Some(file) = module.file(db) else {
         return;
     };
-    // Must be a stdlib module
     let Some(search_path) = module.search_path(db) else {
         return;
     };
     if !search_path.is_standard_library() {
         return;
     }
-    // We must be aware that this is a real symbol on *some* version
+
+    // We populate place_table entries for stdlib items across all known versions and platforms,
+    // so if this lookup succeeds then we know that this lookup *could* succeed with possible
+    // configuration changes.
     let symbol_table = place_table(db, global_scope(db, file));
     if symbol_table.symbol_by_name(attr).is_none() {
         return;
     }
 
+    // For now, we just mention the current version they're on, and hope that's enough of a nudge.
+    // TODO: determine what version they need to be on
+    // TODO: also mention the platform we're assuming
+    // TODO: determine what platform they need to be on
     add_inferred_python_version_hint_to_diagnostic(
         db,
         &mut diagnostic,


crates/ty_python_semantic/resources/mdtest/snapshots/super.md_-_Super_-_Basic_Usage_-_Explicit_Super_Objec…_(b753048091f275c0).snap
```diff
@@ -128,6 +128,7 @@ error[unresolved-attribute]: Type `<super: <class 'C'>, C>` has no attribute `c`
 20 |
 21 | super(B, C()).a
    |
+info: Python 3.12 was assumed when accessing `c` because it was specified on the command line
 info: rule `unresolved-attribute` is enabled by default
```
Gankra: These diagnostics for accesses to `super()` are very sad and maybe we should special case them? I'm not sure if there's a more general rule that excludes super and similar objects.

crates/ty/tests/cli/python_environment.rs
```diff
@@ -875,12 +882,20 @@ fn defaults_to_a_new_python_version() -> anyhow::Result<()> {
...
       | ^^^^^^^^^^
       |
+    info: Python 3.10 was assumed when accessing `grantpt`
+     --> ty.toml:3:18
+      |
+    2 | [environment]
+    3 | python-version = "3.10"
+      |                  ^^^^^^ Python 3.10 assumed due to this configuration setting
+    4 | python-platform = "linux"
+      |
```
Gankra, line 980: Oh boy these are really verbose when it's not "from the CLI"

crates/ty_python_semantic/resources/mdtest/snapshots/super.md_-_Super_-_Basic_Usage_-_Explicit_Super_Objec…_(b753048091f275c0).snap
Gankra: These firing on `super` is so sad, I'm not sure if there's a good General rule that would exclude this and similar types.

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3140,6 +3141,56 @@ pub(super) fn hint_if_stdlib_submodule_exists_on_other_versions(
     add_inferred_python_version_hint_to_diagnostic(db, &mut diagnostic, "resolving modules");
 }
 
+/// This function receives an unresolved `foo.bar` attribute access,
+/// where `foo` can be resolved to have a type but that type does not
+/// have a `bar` attribute.
+///
+/// If the type of `foo` has a definition that originates in the
+/// stsnfstf library and `foo.bar` *does* exist as an attribute on *other*
```
MichaReiser:
```suggestion
/// stsnfstf library and `foo.bar` *does* exist as an attribute on *other*
```

The what library? `stsnfstf`

crates/ty/tests/cli/python_environment.rs
MichaReiser: Yeah. Rustc hides very verbose diagnostic details unless ran with `-v`. I added the infrastructure for it in https://github.com/astral-sh/ruff/pull/20912 but I also think it's sort of fine here since those diagnostics should be rare

sharkdp:
To be honest, I'm not sure if this initial step is a strict improvement. Sure, it seems helpful in the rare case where you are accessing an attribute that is not available on the Python version that you have configured, *but is available under the exact name on a higher version*.

But in the vast majority of `unresolved-attribute` cases (e.g. a misspelled attribute name, or an incomplete attribute name, because I'm still typing it), it just adds additional noise?

```py
"test.exe".trimsuffix(".exe")  # no, it's called "replacesuffix"
```

```
error[unresolved-attribute]: Type `Literal["test.exe"]` has no attribute `trimsuffix`
 --> /home/shark/playground/test.py:4:1
  |
4 | "test.exe".trimsuffix(".exe")
  | ^^^^^^^^^^^^^^^^^^^^^
  |
info: Python 3.14 was assumed when accessing `trimsuffix` because it was specified on the command line
info: rule `unresolved-attribute` is enabled by default
```

Note that the diagnostic change here does not show up in the ecosystem diff because `info` hints are not part of the concise message, but we have thousands of `unresolved-attribute` diagnostics [across the ecosystem](https://ecosystem.ty.dev/), most of which will be more verbose with this change.

AlexWaygood (commented):
Yeah, I agree with @sharkdp that in its current state this could just add a lot of noise and not be that helpful :-(

But I think it's pretty easy to make a few adjustments to this so that it's less noisy, but still pretty useful. Something like this, where we do actually check that the exact symbol exists in the symbol table _somewhere_ (the definition just isn't visible on this Python version), could work pretty well:

```diff
diff --git a/crates/ty_python_semantic/src/semantic_index/place.rs b/crates/ty_python_semantic/src/semantic_index/place.rs
index 38e3d92816..75d3eaac65 100644
--- a/crates/ty_python_semantic/src/semantic_index/place.rs
+++ b/crates/ty_python_semantic/src/semantic_index/place.rs
@@ -181,7 +181,6 @@ impl PlaceTable {
     }
 
     /// Looks up a symbol by its name and returns a reference to it, if it exists.
-    #[cfg(test)]
     pub(crate) fn symbol_by_name(&self, name: &str) -> Option<&Symbol> {
         self.symbols.symbol_id(name).map(|id| self.symbol(id))
     }
diff --git a/crates/ty_python_semantic/src/types/diagnostic.rs b/crates/ty_python_semantic/src/types/diagnostic.rs
index f07d934dad..cb6bf8c4e7 100644
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -9,6 +9,7 @@ use crate::lint::{Level, LintRegistryBuilder, LintStatus};
 use crate::module_resolver::file_to_module;
 use crate::semantic_index::definition::{Definition, DefinitionKind};
 use crate::semantic_index::place::{PlaceTable, ScopedPlaceId};
+use crate::semantic_index::{global_scope, place_table};
 use crate::suppression::FileSuppressionId;
 use crate::types::call::CallError;
 use crate::types::class::{DisjointBase, DisjointBaseKind, Field};
@@ -3155,6 +3156,16 @@ pub(super) fn hint_if_stdlib_attribute_exists_on_other_versions(
     value_type: &Type,
     attr: &Identifier,
 ) {
+    let Type::ModuleLiteral(module) = *value_type else {
+        return;
+    };
+    let Some(file) = module.module(db).file(db) else {
+        return;
+    };
+    let symbol_table = place_table(db, global_scope(db, file));
+    if symbol_table.symbol_by_name(attr).is_none() {
+        return;
+    }
     let Some(definition) = value_type.definition(db) else {
         return;
     };
```

This edit would mean that the suggestions would only fire on `unresolved-attribute` errors involving module-literal types. But I think probably most "oops, I tried to access this attribute that only exists on a new Python version" errors involve module-literal objects. So I think even this limited hint would help a lot of users and be a great first step.

Gankra:
Heh, I considered that implementation originally and was like "wait I can do better, and modules are their own definitions so this handles those too!"

Having seen the results, I agree it's a safer minimum.

Gankra:
Pushed up the new implementation.

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3140,6 +3141,50 @@ pub(super) fn hint_if_stdlib_submodule_exists_on_other_versions(
...
+    // as this covers the most important cases while not being too noisy
+    // about basic typos or special types like `super(C, self)`
+    let Type::ModuleLiteral(module_ty) = value_type else {
+        return;
+    };
+    let module = module_ty.module(db);
+    let Some(file) = module.file(db) else {
+        return;
+    };
+    // Must be a stdlib module
```
```suggestion
    // Version-dependent definitions are much more common in the stdlib
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3140,6 +3141,50 @@ pub(super) fn hint_if_stdlib_submodule_exists_on_other_versions(
...
+    let Some(search_path) = module.search_path(db) else {
+        return;
+    };
+    if !search_path.is_standard_library() {
+        return;
+    }
+    // We must be aware that this is a real symbol on *some* version
```
AlexWaygood: it could also be in the symbol table for the global scope, but without a visible definition, if it's a platform-specific definition (the definition only exists in amn `if sys.platform == "linux"` branch, but their configured platform is `"win32"`). We could consider also highlighting the user's configured platform as well as their configured Python version. Doesn't need to be done in this PR, though

AlexWaygood (approved):
Thank you, this is great!

PR #20910, opened 2025-10-16
[ty] Consider `type_check_only` when ranking completions

decorator-factory:
## Summary

This pull request enhances `Type`s with information on whether they are "for type checking only", and changes autocompletion ranking so that 

This is the first step in solving https://github.com/astral-sh/ty/issues/816

- [x] TODO: I only implemented `@type_check_only` for classes, right now I'm figuring out how to do it for functions. It's a bit puzzling why this decorator supports functions in the first place. The only place I found it used successfully is `scipy-stubs`, where "imaginary methods" are used to make certain classes match certain protocols.

## Test Plan

Added a `import_type_check_only_lowers_ranking` test case, which used to fail.

### DIFF
--- a/crates/ty_completion_eval/truth/import-deprioritizes-type_check_only/main.py
+++ b/crates/ty_completion_eval/truth/import-deprioritizes-type_check_only/main.py
@@ -6,7 +6,7 @@ from module import Class
 Class.meth_<CURSOR:meth_azurous>
 
 # TODO: bound methods don't preserve type-check-only-ness, this is a bug
-# Class().meth_[CURSOR:meth_azurous]
+Class().meth_<CURSOR:meth_azurous>
 
 # TODO: auto-imports don't take type-check-only-ness into account, this is a bug
-# UniquePrefixA[CURSOR:module.UniquePrefixAzurous]
+UniquePrefixA<CURSOR:module.UniquePrefixAzurous>
--- a/crates/ty_completion_eval/completion-evaluation-tasks.csv
+++ b/crates/ty_completion_eval/completion-evaluation-tasks.csv
@@ -7,6 +7,8 @@ import-deprioritizes-sunder,main.py,0,1
 import-deprioritizes-type_check_only,main.py,0,1
 import-deprioritizes-type_check_only,main.py,1,1
 import-deprioritizes-type_check_only,main.py,2,1
+import-deprioritizes-type_check_only,main.py,3,2
+import-deprioritizes-type_check_only,main.py,4,3
 internal-typeshed-hidden,main.py,0,4
 none-completion,main.py,0,11
 numpy-array,main.py,0,


AlexWaygood:
> TODO: I only implemented `@type_check_only` for classes, right now I'm figuring out how to do it for functions. It's a bit puzzling why this decorator supports functions in the first place. The only place I found it used successfully is `scipy-stubs`, where "imaginary methods" are used to make certain classes match certain protocols.

Fine to defer functions for now, IMO! I think the vast majority of uses are with classes. Let's get an MVP in first that we can iterate on in followups

crates/ty_ide/src/completion.rs
```diff
@@ -839,16 +846,22 @@ fn is_in_string(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
     })
 }
 
-/// Order completions lexicographically, with these exceptions:
+/// Order completions according to the following rules:
 ///
-/// 1) A `_[^_]` prefix sorts last and
-/// 2) A `__` prefix sorts last except before (1)
+/// 1) Names with no underscore prefix
+/// 2) Names starting with `_` but not dunders
```
decorator-factory, line 851: Maybe it's out of scope for this PR, but perhaps deprecated items should also ranked  last? (and crossed out in the autocompletion list)

Image

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -1363,7 +1368,7 @@ impl KnownFunction {
             .then_some(candidate)
     }
 
-    /// Return `true` if `self` is defined in `module` at runtime.
+    /// Return `true` if `self` is defined in `module`
```
decorator-factory, line 1370: Was it an important detail that it's really defined in the module at runtime? I am not sure, but `type_check_only` should be the only exception, probably

crates/ty_python_semantic/src/types.rs
```diff
@@ -899,6 +899,22 @@ impl<'db> Type<'db> {
         matches!(self, Type::Dynamic(_))
     }
 
+    /// Is a value of this type only usable in typing contexts?
+    pub(crate) fn is_type_check_only(&self, db: &'db dyn Db) -> bool {
+        match self {
+            Type::ClassLiteral(class_literal) => class_literal.type_check_only(db),
+            Type::FunctionLiteral(f) => {
+                if f.is_known(db, KnownFunction::TypeCheckOnly) {
```
decorator-factory: If there a way to avoid this somehow? Maybe `typing.type_check_only` could get the `TYPE_CHECK_ONLY` decorator flag when it's "inserted" into the typing module? I wasn't able to find where that happens

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2610,6 +2611,14 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 continue;
             }
 
+            if decorator_ty
+                .as_function_literal()
+                .is_some_and(|function| function.is_known(self.db(), KnownFunction::TypeCheckOnly))
+            {
+                type_check_only = true;
+                continue;
+            }
```
MichaReiser, line 2627: Given that this is only used in the LSP context. Could we infer this lazilzy (we can cache it with a salsa query) instead of storing it on every class literal?

So the way this would work is that you make it a method on `ClassLiteral` that then iterates over all decorators too see if any of them are `type_check_only`.

crates/ty_ide/src/completion.rs
MichaReiser: Yeah, agree. I think that sounds like a great change for another PR

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: > Given that this is only used in the LSP context.

We'll want to use it in other contexts in the future, though, e.g. we'll want to emit a diagnostic if a `type_check_only` symbol is imported in a `.py` file outside of an `if TYPE_CHECKING` block

crates/ty_ide/src/completion.rs
```diff
@@ -339,6 +345,7 @@ fn add_unimported_completions<'db>(
             module_name: Some(symbol.module.name(db)),
             import: import_action.import().cloned(),
             builtin: false,
+            is_type_check_only: false,
```
AlexWaygood, line 349: This seems to mean that completions that would require an auto-import are always considered "not type-check-only". Here's a screenshot from a local playground build using your branch (to build the playground locally, use `cd playground && npm start --workspace ty-playground` from the Ruff repo root):

image

`module.py` has these contents, so I would expect `Foo` to be ranked below `Foooo`:

```py
from typing import type_check_only

@type_check_only
class Foo: ...

class Foooo: ...
```

since that's what I get in other situations on your branch (which is great!):

image

crates/ty_python_semantic/src/types.rs
AlexWaygood: This should do it:

```diff
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index 820da99a89..19d78133b3 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -2205,6 +2205,11 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         let known_function =
             KnownFunction::try_from_definition_and_name(self.db(), definition, name);
 
+        // `type_check_only` is itself not available at runtime
+        if known_function == Some(KnownFunction::TypeCheckOnly) {
+            function_decorators |= FunctionDecorators::TYPE_CHECK_ONLY;
+        }
+
         let body_scope = self
             .index
             .node_scope(NodeWithScopeRef::Function(function))
```

crates/ty_ide/src/completion.rs
AlexWaygood: if getting this right is difficult for auto-import completions, it's fine to defer it for now, but we should add a TODO comment here if so

decorator-factory: I think let's defer it for now :+1:
I added some TODOs in the integration test.

Right now deprecation and type-check-only-ness are stored on types, not symbols, which makes this difficult. (we'd have to eagerly infer types to produce autoimport suggestions if I understand it correctly)

decorator-factory:
> I've one question about whether we can compute the is_type_check_only flag lazily rather than storing it on every ClassLiteral

I think it is nicely symmetrical with storing `FunctionDecorators` on function types :smiley_cat: . It also allows pre-populating this flag in things we know to be type-check only like `@typing.type_check_only` itself. (the diff Alex suggested above)

This is something `ty` would have to check pretty much every time you use a class in a runtime context (like if you do `list(something)`), so I don't know which way is better for performance there

crates/ty_ide/src/completion.rs
AlexWaygood: Hmm, so we'd have to infer the types of every symbol that could possibly be imported in order to provide autocomplete suggestions for a single `x = Fo<CURSOR>` situation... that would indeed undermine the "only lazily infer exactly what we need" architecture of ty 😆

MichaReiser: > situation... that would indeed undermine the "only lazily infer exactly what we need" architecture of ty 😆

Yeah, let's wait for @BurntSushi on this, but we can definitely infer the type after we've done some filtering

MichaReiser:
@decorator-factory I'm not sure what the state of this PR is. Would you mind requesting re-review when it's ready for another round of feedback or you can put your PR in draft mode and change it to "ready for review" once it's ready for another round. 

Thank you

decorator-factory:
I think it's ready to merge in the current state. As Alex said, the main use case for `@type_check_only` is classes/protocols that only exist at runtime, so we can fix the `TODO` tests for methods in a later PR

crates/ty_completion_eval/truth/import-deprioritizes-type_check_only/main.py
```diff
@@ -0,0 +1,12 @@
...
+
+from module import Class
+
+Class.meth_<CURSOR:meth_azurous>
+
+# TODO: bound methods don't preserve type-check-only-ness, this is a bug
+# Class().meth_[CURSOR:meth_azurous]
+
+# TODO: auto-imports don't take type-check-only-ness into account, this is a bug
+# UniquePrefixA[CURSOR:module.UniquePrefixAzurous]
```
AlexWaygood: I _believe_ the idea with the tests in this crate is that for autocomplete suggestions that don't achieve a great result currently, that's simply reflected in a lower score in `crates/ty_completion_eval/completion-evaluation-tasks.csv`. So that implies that we should uncomment these lines, and then when we improve our ranking in the future that will simply result in us having to update the score in that CSV file to reflect the fact that the ranking has improved.

decorator-factory: I see, that's how it seems to work in the `numpy-array` and `scope-existing-over-new-import` cases indeed. Fixed in the latest commit

BurntSushi: @AlexWaygood Yup that's exactly right! More broadly, it may be the case that some rankings can't improve or can't improve without regressing others.

crates/ty_ide/src/completion.rs
BurntSushi: My loose plan at this point was to basically never infer the type of unimported symbols, by virtue of there being so many. It's true that we could try doing that after we've done filtering, but I fear this could still have a negative overall effect where we end needing to type check a substantial portion of third party code just by virtue of using completions. And I of course worry about the latency impact this will have on auto-import completions.

This is why the way auto-import works currently is to basically bypass any type checking and just pick up symbols straight from the AST. The way I would try to tackle `type_check_only` support here is to see if we can add it to that AST extraction bit and _not_ try to bring in type checking machinery for it.

PR #20911, opened 2025-10-16
Avoid printing out entire search path if long

hauntsaninja:
We have environments with a thousand entries in our search path.
This overwhelms ty output


MichaReiser:
What a coincidence that we worked on this at the exact same time... https://github.com/astral-sh/ruff/pull/20912

I'll close this as my PR retains the option to list all search paths by opting in with `-v`.

hauntsaninja:
Oh funny! I didn't even see the issue in the ty repo, was just testing out a22 :-) Thanks!

PR #20914, opened 2025-10-16
[ty] Do not assume that `field`s have a default value

sharkdp:
## Summary

fixes https://github.com/astral-sh/ty/issues/1366

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1636,14 +1636,16 @@ impl<'db> Type<'db> {
             (Type::KnownInstance(KnownInstanceType::Field(field)), right)
                 if relation.is_assignability() =>
             {
-                field.default_type(db).has_relation_to_impl(
-                    db,
-                    right,
-                    inferable,
-                    relation,
-                    relation_visitor,
-                    disjointness_visitor,
-                )
+                field.default_type(db).when_none_or(|default_type| {
+                    default_type.has_relation_to_impl(
+                        db,
+                        right,
+                        inferable,
+                        relation,
+                        relation_visitor,
+                        disjointness_visitor,
+                    )
+                })
             }
 
             // Dynamic is only a subtype of `object` and only a supertype of `Never`; both were
@@ -7499,7 +7501,9 @@ fn walk_known_instance_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
             // Nothing to visit
         }
         KnownInstanceType::Field(field) => {
-            visitor.visit_type(db, field.default_type(db));
+            if let Some(default_ty) = field.default_type(db) {
+                visitor.visit_type(db, default_ty);
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -1636,14 +1636,16 @@ impl<'db> Type<'db> {
...
+                    default_type.has_relation_to_impl(
+                        db,
+                        right,
+                        inferable,
+                        relation,
+                        relation_visitor,
+                        disjointness_visitor,
+                    )
+                })
             }
```
AlexWaygood, line 1649: is this correct? This seems to imply that a `Field` instance without a default is assignable to literally _anything_?

Can we make the special case here slightly narrower, like this?

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 8fa72656e4..d538308d80 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1634,10 +1634,12 @@ impl<'db> Type<'db> {
             // This allows field definitions like `name: str = field(default="")` in dataclasses
             // to pass the assignability check of the inferred type to the declared type.
             (Type::KnownInstance(KnownInstanceType::Field(field)), right)
-                if relation.is_assignability() =>
+                if relation.is_assignability() && field.default_type(db).is_some() =>
             {
-                field.default_type(db).when_none_or(|default_type| {
-                    default_type.has_relation_to_impl(
+                field
+                    .default_type(db)
+                    .expect("Checked in branch arm")
+                    .has_relation_to_impl(
                         db,
                         right,
                         inferable,
@@ -1645,7 +1647,6 @@ impl<'db> Type<'db> {
                         relation_visitor,
                         disjointness_visitor,
                     )
-                })
             }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7599,9 +7603,13 @@ impl<'db> KnownInstanceType<'db> {
...
                     KnownInstanceType::Deprecated(_) => f.write_str("warnings.deprecated"),
                     KnownInstanceType::Field(field) => {
-                        f.write_str("dataclasses.Field[")?;
-                        field.default_type(self.db).display(self.db).fmt(f)?;
-                        f.write_str("]")
+                        f.write_str("dataclasses.Field")?;
+                        if let Some(default_ty) = field.default_type(self.db) {
+                            f.write_str("[")?;
+                            default_ty.display(self.db).fmt(f)?;
+                            f.write_str("]")?;
```
```suggestion
                            write!(f, "[{}]", default_ty.display(self.db))?;
```

crates/ty_python_semantic/src/types.rs
sharkdp: > This seems to imply that a `Field` instance without a default is assignable to literally _anything_?

Yes, I think we want that. Consider something like the following, where the `field` call is really only there to exclude the `id` field from the `__init__` signature. This field does not have a default value.
```py
id: int = field(init=False)
```

AlexWaygood: Ah I see. Fair enough! Do we have tests that will fail if we made the bad refactor I proposed here?

sharkdp: Yes, all kinds of tests fail if we do this.

AlexWaygood: thanks!

PR #20915, opened 2025-10-16
[ty] Use field-specifier return type as the default type for the field

sharkdp:
## Summary

`dataclasses.field` and field-specifier functions of commonly used libraries like `pydantic`, `attrs`, and `SQLAlchemy` all return the default type for the field (or `Any`) instead of an actual `Field` instance, even if this is not what happens at runtime. Let's make use of this fact and assume that *all* field specifiers return the type of the default value of the field.

For standard dataclasses, this leads to more or less the same outcome (see test diff for details), but this change is important for 3rd party dataclass-transformers.

## Test Plan

Tested the consequences of this change on the field-specifiers branch as well.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
@@ -11,7 +11,7 @@ class Member:
     role: str = field(default="user")
     tag: str | None = field(default=None, init=False)
 
-# revealed: (self: Member, name: str, role: str = Literal["user"]) -> None
+# revealed: (self: Member, name: str, role: str = str) -> None
 reveal_type(Member.__init__)
 
 alice = Member(name="Alice", role="admin")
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
```diff
@@ -11,7 +11,7 @@ class Member:
     role: str = field(default="user")
     tag: str | None = field(default=None, init=False)
 
-# revealed: (self: Member, name: str, role: str = Literal["user"]) -> None
+# revealed: (self: Member, name: str, role: str = str) -> None
```
sharkdp, line 14: This happens due to literal promotion, I guess.

crates/ty_python_semantic/resources/mdtest/dataclasses/fields.md
```diff
@@ -37,7 +37,7 @@ class Data:
     content: list[int] = field(default_factory=list)
     timestamp: datetime = field(default_factory=datetime.now, init=False)
 
-# revealed: (self: Data, content: list[int] = list[Unknown]) -> None
+# revealed: (self: Data, content: list[int] = list[int]) -> None
```
sharkdp, line 40: This shows that our previous approach, calling `default_factory` manually, was apparently too naive.

PR #20917, opened 2025-10-16
[ty] Fix panic 'missing root' when handling completion request

MichaReiser:
## Summary

Fixes a panic in `list_modules_in` (which is called by the completions request) when the file root for editable installs or a custom typeshed path was missing.

Funnily enough, there were existing tests covering this already. Unfortunately, they also used their custom logic to register the file roots, so that the test passed, but the same request failed in production. 

I used this PR as an opportunity to improve our file-root related logging and the information we include if a file root is missing.

Fixes https://github.com/astral-sh/ty/issues/1363

## Test Plan

Updated existing tests, added new test

### DIFF
--- a/crates/ruff_db/src/files.rs
+++ b/crates/ruff_db/src/files.rs
@@ -193,6 +193,17 @@ impl Files {
         roots.at(&absolute)
     }
 
+    /// The same as [`Self::root`] but panics if no root is found.
+    #[track_caller]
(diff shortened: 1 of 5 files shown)


Danielkonge:
Just a note: I checked out this branch and tried it out, and it indeed fixes the problem I kept encountering in the linked issue.

PR #20922, opened 2025-10-16
[ty] Infer type for implicit `self` parameters in method bodies

sharkdp:
## Summary

Infer a type of `Self` for unannotated `self` parameters in methods of classes.

part of https://github.com/astral-sh/ty/issues/159

closes https://github.com/astral-sh/ty/issues/1081

## Conformance tests changes

```diff
+enums_member_values.py:85:9: error[invalid-assignment] Object of type `int` is not assignable to attribute `_value_` of type `str`
```
A true positive :heavy_check_mark: 

```diff
-generics_self_advanced.py:35:9: error[type-assertion-failure] Argument does not have asserted type `Self@method2`
-generics_self_basic.py:14:9: error[type-assertion-failure] Argument does not have asserted type `Self@set_scale
```

Two false positives going away :heavy_check_mark: 

```diff
+generics_syntax_infer_variance.py:82:9: error[invalid-assignment] Cannot assign to final attribute `x` on type `Self@__init__`
```

This looks like a true positive to me, even if it's not marked with `# E` :heavy_check_mark: 

```diff
+protocols_explicit.py:56:9: error[invalid-assignment] Object of type `tuple[int, int, str]` is not assignable to attribute `rgb` of type `tuple[int, int, int]`
```

True positive :heavy_check_mark: 

```
+protocols_explicit.py:85:9: error[invalid-attribute-access] Cannot assign to ClassVar `cm1` from an instance of type `Self@__init__`
```

This looks like a true positive to me, even if it's not marked with `# E`. But this is consistent with our understanding of `ClassVar`, I think. :heavy_check_mark: 

```py
+qualifiers_final_annotation.py:52:9: error[invalid-assignment] Cannot assign to final attribute `ID4` on type `Self@__init__`
+qualifiers_final_annotation.py:65:9: error[invalid-assignment] Cannot assign to final attribute `ID7` on type `Self@method1`
```

New true positives :heavy_check_mark: 

```py
+qualifiers_final_annotation.py:52:9: error[invalid-assignment] Cannot assign to final attribute `ID4` on type `Self@__init__`
+qualifiers_final_annotation.py:57:13: error[invalid-assignment] Cannot assign to final attribute `ID6` on type `Self@__init__`
+qualifiers_final_annotation.py:59:13: error[invalid-assignment] Cannot assign to final attribute `ID6` on type `Self@__init__`
```

This is a new false positive, but that's a pre-existing issue on main (if you annotate with `Self`): https://play.ty.dev/3ee1c56d-7e13-43bb-811a-7a81e236e6ab  :x: => reported as https://github.com/astral-sh/ty/issues/1409

## Ecosystem

* There are 5931 new `unresolved-attribute` and 3292 new `possibly-missing-attribute` attribute errors, way too many to look at all of them. I randomly sampled 15 of these errors and found: 
  * 13 instances where there was simply no such attribute that we could plausibly see. Sometimes [I didn't find it anywhere](https://github.com/internetarchive/openlibrary/blob/8644d886c6579a5f49faadc4cd1ba9992e603d7e/openlibrary/plugins/openlibrary/tests/test_listapi.py#L33). Sometimes it was set externally on the object. Sometimes there was some [`setattr` dynamicness going on](https://github.com/pypa/setuptools/blob/a49f6b927d83b97630b4fb030de8035ed32436fd/setuptools/wheel.py#L88-L94). I would consider all of them to be true positives.
  * 1 instance where [attribute was set on `obj` in `__new__`](https://github.com/sympy/sympy/blob/9e87b44fd43572b9c4cc95ec569a2f4b81d56499/sympy/tensor/array/array_comprehension.py#L45C1-L45C36), which we don't support yet
  * 1 instance [where the attribute was defined via `__slots__`
](https://github.com/spack/spack/blob/e250ec0fc81130b708a8abe1894f0cc926880210/lib/spack/spack/vendor/pyrsistent/_pdeque.py#L48C5-L48C14)
* I see 44 instances [of the false positive above](https://github.com/astral-sh/ty/issues/1409) with `Final` instance attributes being set in `__init__`. I don't think this should block this PR.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -3,7 +3,7 @@ use std::{iter, mem};
 use itertools::{Either, Itertools};
 use ruff_db::diagnostic::{Annotation, DiagnosticId, Severity};
 use ruff_db::files::File;
-use ruff_db::parsed::{ParsedModuleRef, parsed_module};
+use ruff_db::parsed::ParsedModuleRef;
 use ruff_python_ast::visitor::{Visitor, walk_expr};
 use ruff_python_ast::{self as ast, AnyNodeRef, ExprContext, PythonVersion};
 use ruff_python_stdlib::builtins::version_builtin_was_added;
@@ -83,7 +83,7 @@ use crate::types::generics::{
     GenericContext, InferableTypeVars, LegacyGenericBase, SpecializationBuilder, bind_typevar,
     enclosing_generic_contexts, typing_self,
 };
-use crate::types::infer::{nearest_enclosing_class, nearest_enclosing_function};
+use crate::types::infer::nearest_enclosing_function;
 use crate::types::instance::SliceLiteral;
 use crate::types::mro::MroErrorKind;
 use crate::types::signatures::Signature;
@@ -2543,45 +2543,57 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         parameter: &ast::Parameter,
     ) -> Option<Type<'db>> {
         let db = self.db();
-        let scope_id = self.scope();
-        let file = scope_id.file(db);
-        let function_scope = scope_id.scope(db);
-        let method = function_scope.node().as_function()?;
+        let file = self.file();
 
-        let parent_scope_id = function_scope.parent()?;
-        let parent_scope = self.index.scope(parent_scope_id);
-        parent_scope.node().as_class()?;
+        let function_scope_id = self.scope();
+        let function_scope = function_scope_id.scope(db);
+        let function = function_scope.node().as_function()?;
 
-        let method_definition = self.index.expect_single_definition(method);
+        let parent_file_scope_id = function_scope.parent()?;
+        let mut parent_scope_id = parent_file_scope_id.to_scope_id(db, file);
+
+        // Skip type parameter scopes, if the method itself is generic.
+        if parent_scope_id.is_annotation(db) {
+            let parent_scope = parent_scope_id.scope(db);
+            parent_scope_id = parent_scope.parent()?.to_scope_id(db, file);
+        }
+
+        // Return early if this is not a method inside a class.
+        let class = parent_scope_id.scope(db).node().as_class()?;
+
+        let method_definition = self.index.expect_single_definition(function);
         let DefinitionKind::Function(function_definition) = method_definition.kind(db) else {
             return None;
         };
 
-        let func_type = infer_definition_types(db, method_definition)
-            .declaration_type(method_definition)
-            .inner_type()
-            .as_function_literal()?;
-
-        let module = parsed_module(db, file).load(db);
         if function_definition
-            .node(&module)
+            .node(self.module())
             .parameters
             .index(parameter.name())
-            .is_some_and(|index| index != 0)
+            .is_none_or(|index| index != 0)
         {
             return None;
         }
 
-        if func_type.is_classmethod(db) {
+        let method = infer_definition_types(db, method_definition)
+            .declaration_type(method_definition)
+            .inner_type()
+            .as_function_literal()?;
+
+        if method.is_classmethod(db) {
             // TODO: set the type for `cls` argument
             return None;
-        } else if func_type.is_staticmethod(db) {
+        } else if method.is_staticmethod(db) {
             return None;
         }
 
-        let class = nearest_enclosing_class(db, self.index, scope_id).unwrap();
+        let class_definition = self.index.expect_single_definition(class);
+        let class_literal = infer_definition_types(db, class_definition)
+            .declaration_type(class_definition)
+            .inner_type()
+            .as_class_literal()?;
 
-        typing_self(db, self.scope(), Some(method_definition), class)
+        typing_self(db, self.scope(), Some(method_definition), class_literal)
     }
 
     /// Set initial declared/inferred types for a `*args` variadic positional parameter.
--- a/crates/ty_python_semantic/resources/mdtest/annotations/self.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/self.md
@@ -56,7 +56,7 @@ In instance methods, the first parameter (regardless of its name) is assumed to 
 
 ```toml
 [environment]
-python-version = "3.11"
+python-version = "3.12"
 ```
 
 ```py
@@ -68,11 +68,26 @@ class A:
 
         return self
 
-    def a_method(self) -> int:
-        def first_arg_is_not_self(a: int) -> int:
+    def implicit_self_generic[T](self, x: T) -> T:
+        reveal_type(self)  # revealed: Self@implicit_self_generic
+
+        return x
+
+    def method_a(self) -> None:
+        def first_param_is_not_self(a: int):
             reveal_type(a)  # revealed: int
-            return a
-        return first_arg_is_not_self(1)
+            reveal_type(self)  # revealed: Self@method_a
+
+        def first_param_is_not_self_unannotated(a):
+            reveal_type(a)  # revealed: Unknown
+            reveal_type(self)  # revealed: Self@method_a
+
+        def first_param_is_also_not_self(self) -> None:
+            reveal_type(self)  # revealed: Unknown
+
+        def first_param_is_explicit_self(this: Self) -> None:
+            reveal_type(this)  # revealed: Self@method_a
+            reveal_type(self)  # revealed: Self@method_a
 
     @classmethod
     def a_classmethod(cls) -> Self:


crates/ty_python_semantic/resources/mdtest/attributes.md
```diff
@@ -1839,6 +1846,7 @@ def external_getattribute(name) -> int:
 
 class ThisFails:
     def __init__(self):
+        # error: [invalid-assignment] "Implicit shadowing of function `__getattribute__`"
```
sharkdp, line 1849: Probably something we should look at or create an issue for.

python/py-fuzzer/fuzz.py
```diff
@@ -139,7 +139,7 @@ def print_description(self, index: int, num_seeds: int) -> None:
                 case Executable.TY:
                     panic_message = f"The following code triggers a {new}ty panic:"
                 case _ as unreachable:
-                    assert_never(unreachable)  # ty: ignore[type-assertion-failure]
```
sharkdp, line 142: FYI @AlexWaygood :smile:

crates/ty_python_semantic/resources/mdtest/type_qualifiers/final.md
```diff
@@ -88,6 +88,8 @@ class C:
         self.FINAL_C: Final[int] = 1
         self.FINAL_D: Final = 1
         self.FINAL_E: Final
+        # TODO: Should not be an error
+        # error: [invalid-assignment] "Cannot assign to final attribute `FINAL_E` on type `Self@__init__`"
```
sharkdp, line 92: This (and below) was really only "working" because we didn't infer a type of `self`, not because we ever added support for this. As far as I remember, no other type checker supports this either and we decided at the time that it was not important to implement.

python/py-fuzzer/fuzz.py
AlexWaygood: Let's go!

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -206,7 +206,7 @@ enum ReduceResult<'db> {
 //
 // For now (until we solve https://github.com/astral-sh/ty/issues/957), keep this number
 // below 200, which is the salsa fixpoint iteration limit.
-const MAX_UNION_LITERALS: usize = 199;
+const MAX_UNION_LITERALS: usize = 190;
```
MichaReiser, line 209: Just curious, what's the reason for reducing the count here?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2535,6 +2537,53 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
     }
 
+    /// Special case for unannotated `cls` and `self` arguments to class methods and instance methods.
+    fn special_first_method_parameter_type(
+        &mut self,
+        parameter: &ast::Parameter,
+    ) -> Option<Type<'db>> {
+        let db = self.db();
+        let scope_id = self.scope();
+        let file = scope_id.file(db);
```
MichaReiser: Nit: Can we use `self.file()` here? All scopes should belong to the same file

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2535,6 +2537,53 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        let function_scope = scope_id.scope(db);
+        let method = function_scope.node().as_function()?;
+
+        let parent_scope_id = function_scope.parent()?;
+        let parent_scope = self.index.scope(parent_scope_id);
+        parent_scope.node().as_class()?;
```
MichaReiser: I'm not sure if this is still the case or if I remember it incorrectly. 

For generic classes, isn't there a nested scope for the generic between the class and its members?

MichaReiser: Nit: I'd find an explicit return here easier to read
```suggestion
        if !parent_scope.kind().is_class() {
        	return;
      	}
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2535,6 +2537,53 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            .declaration_type(method_definition)
+            .inner_type()
+            .as_function_literal()?;
+
+        let module = parsed_module(db, file).load(db);
+        if function_definition
+            .node(&module)
+            .parameters
+            .index(parameter.name())
+            .is_some_and(|index| index != 0)
```
MichaReiser: Should we also return here if the function has no parameter named `parameter.name` (which should never happen?)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2535,6 +2537,53 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        }
+
+        if func_type.is_classmethod(db) {
+            // TODO: set the type for `cls` argument
+            return None;
+        } else if func_type.is_staticmethod(db) {
+            return None;
+        }
+
+        let class = nearest_enclosing_class(db, self.index, scope_id).unwrap();
```
MichaReiser: Nit: Not sure if it's worth it but we already check that the parent is a class on the line above. Could we infer the type directly instead of walking the ancestor chain again?

MichaReiser (approved):
Congrats!

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: I think it's the other way around — the generic scope contains the class scope, which then contains the class members

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2535,6 +2537,53 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        let DefinitionKind::Function(function_definition) = method_definition.kind(db) else {
+            return None;
+        };
+
+        let func_type = infer_definition_types(db, method_definition)
+            .declaration_type(method_definition)
+            .inner_type()
+            .as_function_literal()?;
+
+        let module = parsed_module(db, file).load(db);
```
dcreager: Can this be `self.module()`?

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: Or alternatively, can we pass the parameter index into `infer_parameter_definition`, and only call this method for the first parameter? Then we don't have to look up by name, and we don't have to verify that what we found was parameter 0.

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: Or rather, the check above will only pass when the function containing this parameter is an immediate child of a class. It won't work for nested functions. _Explicit_ `Self` is allowed in a nested function ([spec](https://typing.python.org/en/latest/spec/generics.html#valid-locations-for-self), `HasNestedFunction.foo.nested`). But is _implicit_ `Self` allowed there? i.e., should this work?

```py
class C:
    def f(self):
        def nested(self):
            reveal_type(self)  # revealed: Self@C
```

If so, then the call above should use `nearest_enclosing_class`. If not, then I agree with Micha that here we just use the immediately enclosing class that we can get from above.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Yes, but this is still a good catch. Because there can be a generic scope between class body and function body if the *method* itself is generic. We made that error in a few other places, and I thought I had them all in my mind (and we're fixing them in https://github.com/astral-sh/ruff/pull/20856), but didn't remember that we introduce this error again here.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: > can we pass the parameter index into infer_parameter_definition

I see no way of doing that, unfortunately. The AST doesn't store the index of the parameter.

crates/ty_python_semantic/src/types/builder.rs
sharkdp: Sorry, this was discussed on the original PR. It was necessary to prevent ecosystem panics. The true fix for this is https://github.com/astral-sh/ty/issues/957.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Thanks, I added better tests and removed the `nearest_enclosing_class` call.

crates/ty_python_semantic/src/types/infer/builder.rs
dcreager: It looks like [`element_offset`](https://doc.rust-lang.org/std/primitive.slice.html#method.element_offset) is [stabilizing soon](https://github.com/rust-lang/rust/issues/126769#issuecomment-3418757954) — so in some number of months we'll be able to get the parameter index from the function's parameter slice without threading it through explicitly!

PR #20900, opened 2025-10-15
[ty] display variance on hover over type variables

mtshiba:
## Summary

Edit: please merge #20924 first

Closes astral-sh/ty#1348

The variance of a type variable is displayed (next to the type) when hovered over.
They are also displayed in the type parameters list.

This is an enhancement to the language server, but also includes some internal additions to `ty_python_semantic` (such as calculating variances for type aliases).

## Test Plan

New tests in `ty_ide/hover/tests`.

PR #20924, opened 2025-10-16
[ty] impl `VarianceInferable` for `KnownInstanceType`

mtshiba:
## Summary

Derived from #20900

Implement `VarianceInferable` for `KnownInstanceType` (especially for `KnownInstanceType::TypeAliasType`).

The variance of a type alias matches its value type. In normal usage, type aliases are expanded to value types, so the variance of a type alias can be obtained without implementing this. However, for example, if we want to display the variance when hovering over a type alias, we need to be able to obtain the variance of the type alias itself (cf. #20900).

## Test Plan

I couldn't come up with a way to test this in mdtest, so I'm testing it in a test submodule at the end of `types.rs`.
I also added a test to `mdtest/generics/pep695/variance.md`, but it passes without the changes in this PR.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -10775,6 +10775,7 @@ impl<'db> PEP695TypeAliasType<'db> {
         semantic_index(db, scope.file(db)).expect_single_definition(type_alias_stmt_node)
     }
 
+    /// The RHS type of a PEP-695 style type alias with specialization applied.
     #[salsa::tracked(cycle_fn=value_type_cycle_recover, cycle_initial=value_type_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
     pub(crate) fn value_type(self, db: &'db dyn Db) -> Type<'db> {
         let value_type = self.raw_value_type(db);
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -10783,6 +10790,15 @@ impl<'db> PEP695TypeAliasType<'db> {
         }
     }
 
+    pub(crate) fn raw_value_type(self, db: &'db dyn Db) -> Type<'db> {
```
sharkdp, line 10804: Can we please document what `raw_value_type` does compared to `value_type`?

More importantly, `value_type` has fixpoint handling, but `raw_value_type` does not. Does this lead to problems with recursive type aliases?

mtshiba: `raw_value_type` is called by `KnownInstanceType::variance_of` except within the queried `value_type`.

[As already explained](https://github.com/astral-sh/ruff/pull/20924#issue-3522711333), `KnownInstanceType::variance_of` is not normally called. In current usage, the only time it may be called is during a hover request.
Clearly, the type inference functions that `raw_value_type` depends on never call a hover request. Therefore, in current usage, there is no possibility of infinite recursion.

For safety, I added a note that `raw_value_type` is not tracked, so that anyone using it in the future knows to be careful.

PR #20926, opened 2025-10-16
[semantic error tests]: refactor semantic error tests to separate files

11happy:
## Summary

This PR refactors semantic error tests in each seperate file

## CC
- @ntBre

### DIFF
--- a/crates/ruff_linter/src/linter.rs
+++ b/crates/ruff_linter/src/linter.rs
@@ -978,7 +978,7 @@ mod tests {
     #[test_case(Path::new("single_starred_assignment.py"), PythonVersion::PY310)]
     #[test_case(Path::new("write_to_debug.py"), PythonVersion::PY312)]
     #[test_case(Path::new("write_to_debug.py"), PythonVersion::PY310)]
-    #[test_case(Path::new("invalid_expression.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("invalid_expression.py"), PythonVersion::PY312)]
     #[test_case(Path::new("global_parameter.py"), PythonVersion::PY310)]
     fn test_semantic_errors(path: &Path, python_version: PythonVersion) -> Result<()> {
         let snapshot = format!(
--- a/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__semantic_syntax_error_invalid_expression.py_3.12.snap
+++ b/crates/ruff_linter/src/snapshots/ruff_linter__linter__tests__semantic_syntax_error_invalid_expression.py_3.12.snap
@@ -1,15 +1,6 @@
 ---
 source: crates/ruff_linter/src/linter.rs
 ---
-invalid-syntax: Cannot use `type` alias statement on Python 3.10 (syntax was added in Python 3.12)
- --> resources/test/fixtures/semantic_errors/invalid_expression.py:1:1
-  |
-1 | type X[T: (yield 1)] = int
-  | ^^^^
-2 |
-3 | type Y = (yield 1)
-  |
-
 invalid-syntax: yield expression cannot be used within a TypeVar bound
  --> resources/test/fixtures/semantic_errors/invalid_expression.py:1:12
   |
@@ -17,17 +8,6 @@ invalid-syntax: yield expression cannot be used within a TypeVar bound
   |            ^^^^^^^
 2 |
 3 | type Y = (yield 1)
-  |
-
-invalid-syntax: Cannot use `type` alias statement on Python 3.10 (syntax was added in Python 3.12)
- --> resources/test/fixtures/semantic_errors/invalid_expression.py:3:1
-  |
-1 | type X[T: (yield 1)] = int
-2 |
-3 | type Y = (yield 1)
-  | ^^^^
-4 |
-5 | def f[T](x: int) -> (y := 3): return x
   |
 
 invalid-syntax: yield expression cannot be used within a type alias
@@ -41,17 +21,6 @@ invalid-syntax: yield expression cannot be used within a type alias
 5 | def f[T](x: int) -> (y := 3): return x
   |
 
-invalid-syntax: Cannot use type parameter lists on Python 3.10 (syntax was added in Python 3.12)
- --> resources/test/fixtures/semantic_errors/invalid_expression.py:5:6
-  |
-3 | type Y = (yield 1)
-4 |
-5 | def f[T](x: int) -> (y := 3): return x
-  |      ^^^
-6 |
-7 | class C[T]((yield from [object])):
-  |
-
 invalid-syntax: named expression cannot be used within a generic definition
  --> resources/test/fixtures/semantic_errors/invalid_expression.py:5:22
   |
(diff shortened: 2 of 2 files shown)


ntBre:
Nice, thanks for working on this! Let me know when it's ready for review, or if you need any help. It looks like you're on the right track :)

11happy:
CI is green : ) , its ready for review.

Thank you

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,92 @@ mod tests {
...
+                &source_kind,
+                &path,
+                &LinterSettings {
+                    rules: settings::rule_table::RuleTable::empty(),
+                    unresolved_target_version: python_version.into(),
+                    preview: settings::types::PreviewMode::Enabled,
+                    ..Default::default()
+                },
+            );
+            insta::with_settings!({filters => vec![(r"\\", "/")]}, {
```
ntBre: Do we need this filter? It looks like we use `test_contents_syntax_errors` in other tests in this file without the filter, so it would probably be okay to omit it.

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,92 @@ mod tests {
...
+                assert_diagnostics!(format!("{snapshot}_{python_version}"), diagnostics);
```
ntBre: It looks like we're adding the Python version twice, once here and once above on line 1033.

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,92 @@ mod tests {
...
-                ..Default::default()
-            },
-        );
-        assert_diagnostics!(snapshot, diagnostics);
+    ) -> Result<()> {
+        let path = Path::new("resources/test/fixtures/semantic_errors").join(path);
+        let contents = std::fs::read_to_string(&path)?;
+        let source_kind = SourceKind::Python(contents);
+        for &python_version in python_versions {
+            let snapshot = format!("semantic_syntax_error_{error_type}_{python_version}");
```
ntBre: I think we could probably omit the `error_type` and just use the path instead. That seems closer to what we do for lint rules, and it will make the `test_case` entries a bit shorter.

Looking back at the blame, I'm actually not sure why we added `error_type` in the first place.

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,92 @@ mod tests {
...
-    #[test_case(
-        "invalid_expression_yield_from_in_base_class",
-        "
-        class C[T]((yield from [object])):
-            pass
-        ",
-        PythonVersion::PY312,
-        "InvalidExpression"
+        Path::new("global_parameter.py"),
+        &[PythonVersion::PY312],
```
ntBre: It looks like most of these only have one Python version, so I'd probably just pass a single Python version, like we had before. And then we can duplicate the `test_case` like:

```rust
#[test_case(Path::new("some_file.py"), PythonVersion::PY311)]
#[test_case(Path::new("some_file.py"), PythonVersion::PY312)]
```

I think that looks a bit neater.

crates/ruff_linter/src/linter.rs
11happy: done

crates/ruff_linter/src/linter.rs
11happy: done

crates/ruff_linter/src/linter.rs
11happy: done

crates/ruff_linter/src/linter.rs
11happy: done

crates/ruff_linter/src/linter.rs
11happy: path is causing windows test to fail .

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,50 @@ mod tests {
...
-    #[test_case(
-        "invalid_star_expression_for",
-        "
-        for *x in range(10):
-            pass
-        ",
-        PythonVersion::PY310,
-        "InvalidStarExpression"
+        Path::new("async_comprehension_outside_async_function.py"),
+        PythonVersion::PY312
```
ntBre: Should this be 3.11? It looks like the old tests used 3.10 and 3.11, not 3.12.

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,50 @@ mod tests {
...
+    #[test_case(Path::new("multiple_case_assignment.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("duplicate_match_key.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("duplicate_match_class_attribute.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("invalid_star_expression.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("irrefutable_case_pattern.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("single_starred_expression.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("write_to_debug.py"), PythonVersion::PY312)]
+    #[test_case(Path::new("write_to_debug.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("invalid_expression.py"), PythonVersion::PY312)]
+    #[test_case(Path::new("global_parameter.py"), PythonVersion::PY312)]
```
ntBre: This one was also 3.10 previously. I don't think it matters in this case, but I'd like to change as little as possible in the refactor.

crates/ruff_linter/src/linter.rs
```diff
@@ -972,253 +961,50 @@ mod tests {
...
-        let diagnostics = test_snippet_syntax_errors(
-            contents,
+    #[test_case(Path::new("rebound_comprehension.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("duplicate_type_parameter.py"), PythonVersion::PY312)]
+    #[test_case(Path::new("multiple_case_assignment.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("duplicate_match_key.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("duplicate_match_class_attribute.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("invalid_star_expression.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("irrefutable_case_pattern.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("single_starred_expression.py"), PythonVersion::PY310)]
```
ntBre: Small nit, but should we call this file `single_starred_assignment` like the error variant?

ntBre (approved):
Thank you! I had a handful of additional naming/version nits, but I think this is good to go otherwise.

crates/ruff_linter/src/linter.rs
11happy: sure done

crates/ruff_linter/src/linter.rs
11happy: done

crates/ruff_linter/src/linter.rs
11happy: done

ntBre (approved):

PR #20927, opened 2025-10-16
[ty] Prefer declared type for invariant collection literals

ibraheemdev:
## Summary

Prefer the declared type for collection literals, e.g.,
```py
x: list[Any] = [1, "2", (3,)]
reveal_type(x)  # list[Any]
```

This solves a large part of https://github.com/astral-sh/ty/issues/136 for invariant generics, where respecting the declared type is a lot more important. It also means that annotated dict literals with `dict[_, Any]` is a way out of https://github.com/astral-sh/ty/issues/1248.


AlexWaygood:
(Retitled the PR because tuples are also collection literals, but those don't look like they're touched by this PR -- a decision which I support!)

ibraheemdev:
Looks like there are no regressions in the ecosystem report and a lot of clear improvements, so hopefully this is fairly uncontroversial (I also discussed this with @carljm and he agreed).

PR #20928, opened 2025-10-16
[ty] Truncate Literal type display in some situations

InvalidPathException:
## Summary

Fixes [astral-sh/ty#1369](https://github.com/astral-sh/ty/issues/1369)

## Test Plan

A new mdtest case is created
`Union[Literal[1, 2, 3, 4, 5, 6, 7, 8], A, B, C, D, E, F]` -> `Literal[1, 2, 3, 4, 5, ... omitted 3 literals] | A | B | ... omitted 4 union elements` because there was no appropriate existing case.

The non-truncation case for reveal_type has many existing tests to verify, for example, `crates/ty_python_semantic/resources/mdtest/subscript/tuple.md`.

Tentatively setting truncation threshold to 7. When truncated, we keep the first 5 elements.

### DIFF
--- a/crates/ty_python_semantic/src/types/display.rs
+++ b/crates/ty_python_semantic/src/types/display.rs
@@ -1322,7 +1322,7 @@ impl Display for DisplayParameter<'_> {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 struct TruncationPolicy {
     max: usize,
     max_when_elided: usize,
@@ -1342,6 +1342,7 @@ impl TruncationPolicy {
     }
 }
 
+#[derive(Debug)]
 struct DisplayOmitted {
     count: usize,
     singular: &'static str,


crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1322,6 +1322,43 @@ impl Display for DisplayParameter<'_> {
     }
 }
 
+#[derive(Copy, Clone)]
```
```suggestion
#[derive(Debug, Copy, Clone)]
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1322,6 +1322,43 @@ impl Display for DisplayParameter<'_> {
...
+        let limit = if total > self.max {
+            self.max_when_elided
+        } else {
+            self.max
+        };
+        limit.min(total)
+    }
+}
+
+struct DisplayOmitted {
```
AlexWaygood, line 1346:
```suggestion
#[derive(Debug)]
struct DisplayOmitted {
```

PR #20933, opened 2025-10-17
[ty] Use type context for inference of generic constructors

ibraheemdev:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1228.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -2687,6 +2687,7 @@ struct ArgumentTypeChecker<'a, 'db> {
     arguments: &'a CallArguments<'a, 'db>,
     argument_matches: &'a [MatchedArgument<'db>],
     parameter_tys: &'a mut [Option<Type<'db>>],
+    callable_type: Type<'db>,
     call_expression_tcx: TypeContext<'db>,
     return_ty: Type<'db>,
     errors: &'a mut Vec<BindingError<'db>>,
@@ -2703,6 +2704,7 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
         arguments: &'a CallArguments<'a, 'db>,
         argument_matches: &'a [MatchedArgument<'db>],
         parameter_tys: &'a mut [Option<Type<'db>>],
+        callable_type: Type<'db>,
         call_expression_tcx: TypeContext<'db>,
         return_ty: Type<'db>,
         errors: &'a mut Vec<BindingError<'db>>,
@@ -2713,6 +2715,7 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
             arguments,
             argument_matches,
             parameter_tys,
+            callable_type,
             call_expression_tcx,
             return_ty,
             errors,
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2623,7 +2626,22 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
             .apply_specialization(self.db, isolated_specialization);
 
         let mut try_infer_tcx = || {
-            let return_ty = self.signature.return_ty?;
+            // For generic constructors, we use the type context to infer the specialization of the class
+            // instance instead of the method's return type.
+            let (inference_context, generic_constructor) = if let Type::BoundMethod(method) =
+                self.callable_type
```
ibraheemdev: Pretty annoyed with the formatting here.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2623,7 +2626,22 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
             .apply_specialization(self.db, isolated_specialization);
 
         let mut try_infer_tcx = || {
-            let return_ty = self.signature.return_ty?;
+            // For generic constructors, we use the type context to infer the specialization of the class
```
sharkdp: I was a bit confused here at first, because "generic constructor" sounded to me like the function itself was generic (which is probably illegal? or doesn't make sense because `__init__` always returns `None`, and so every generic parameter could be annotated with its upper bound). Maybe:
```suggestion
            // For constructors of generic classes, we use the type context to infer the specialization of the class
```

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2623,7 +2626,22 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+            // instance instead of the method's return type.
+            let (inference_context, generic_constructor) = if let Type::BoundMethod(method) =
```
sharkdp: Matching on explicit type variants is tempting, I know, but will break down when  the callable is not precisely a bound method. For example, in a case like the following, the callable would be a union type:
```py
class X[T]:
    def __init__(self, value: T):
        self.value = value

def _(flag: bool):
    b: X[int | None] = X(1) if flag else X(2)
    reveal_type(b)
```

This might be difficult to fix here, and might be an acceptable limitation for now, but we should maybe add a test case and a TODO?

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2623,7 +2626,22 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+                self.callable_type
+                && let Type::NominalInstance(instance) = method.self_instance(self.db)
+                && method.function(self.db).name(self.db) == "__init__"
```
sharkdp: Is it enough to match on `__init__` or do we need a similar treatment of `__new__` calls (which we generally don't support everywhere and fully yet)?

ibraheemdev:
The ecosystem report looks a lot better after rebasing this on https://github.com/astral-sh/ruff/pull/21210. Most of the added diagnostics seem to be existing issues that were hidden by an inference of `Unknown`.

sharkdp (approved):

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: There are a couple cases that this does not handle correctly, I left those as TODOs for now.

PR #20935, opened 2025-10-17
docs: revise Ruff setup instructions for Zed editor

LoicRiegel:
## Summary

Updated Ruff installation instructions for Zed editor.

## Details

Ruff support is now built in and does not require installing an extension anymore. https://github.com/zed-industries/zed/pull/37804

Also FYI this open issue is related: https://github.com/zed-industries/zed/issues/40266.

But I've just tried the last code snippet from the docs on a fresh Zed install on Windows and it worked well without installing any extension.

### DIFF
--- a/docs/editors/setup.md
+++ b/docs/editors/setup.md
@@ -424,9 +424,9 @@ bundle for TextMate.
 
 Ruff support is now built into Zed (no separate extension required).
 
-By default, Zed uses basedpyright as the primary LSP for language features, and Ruff is used by default for formatting and linting.
+By default, Zed uses Ruff for formatting and linting.
 
-To set up editor-wide Ruff option, provide the [server settings](settings.md)
+To set up editor-wide Ruff options, provide the [server settings](settings.md)
 under the [`lsp.ruff.initialization_options.settings`](https://zed.dev/docs/configuring-zed#lsp) key of your `settings.json` file:
 
 ```json
@@ -435,7 +435,7 @@ under the [`lsp.ruff.initialization_options.settings`](https://zed.dev/docs/conf
     "ruff": {
       "initialization_options": {
         "settings": {
-          // Ruff server settings goes here
+          // Ruff server settings go here
           "lineLength": 80,
           "lint": {
             "extendSelect": ["I"],
@@ -447,10 +447,8 @@ under the [`lsp.ruff.initialization_options.settings`](https://zed.dev/docs/conf
 }
 ```
 
-Alternatively, Zed reads project configuration files (`pyproject.toml`, `ruff.toml`, or `.ruff.toml`) to apply project-specific configurations.
-
 [`format_on_save`](https://zed.dev/docs/configuring-zed#format-on-save) is enabled by default.
-You can disable it for Python by adding this to your `settings.json` file:
+You can disable it for Python by changing `format_on_save` in your `settings.json` file:
 
 ```json
 {


MichaReiser:
Thank you. There's probably more that we can remove, now that ruff is used as the default formatter and linter (it's enabled by default) as mentioned here https://zed.dev/docs/languages/python#code-formatting--linting

LoicRiegel:
I can have a look

injust:
Zed comes with formatting and linting out-of-the-box now.

If you want Ruff to organize imports:

```json
{
    "languages": {
        "Python": {
            "code_actions_on_format": { "source.organizeImports.ruff": true }
        }
    }
}
```

If you want Ruff to apply fixes (which includes organizing imports):

```json
{
    "languages": {
        "Python": {
            "code_actions_on_format": { "source.fixAll.ruff": true }
        }
    }
}
```

LoicRiegel:
I'll do a commit tonight.

Strange that Zed's documentation does not mention the `code_actions_on_format` anywhere in their python docs. But they have a nice tool to automatically update `settings.json` settings.
They also mention that the ruff configuration can be read from `ruff.toml`, but `.ruff.toml` and `pyproject.toml` also work (just tested, I wanted to be sure).
Maybe I'll make a small PR on their side too.

BTW do we want do add this in ty's documentation, or is this section https://docs.astral.sh/ty/editors/#other-editors enough?:
```jsonc
{
  "languages": {
    "Python": {
      "language_servers": [
        // Disable basedpyright and enable Ty, and otherwise
        // use the default configuration.
        "ty",
        "!basedpyright",
        "..."
      ]
    }
  }
}
```

injust:
> Strange that Zed's documentation does not mention the `code_actions_on_format` anywhere in their python docs. But they have a nice tool to automatically update `settings.json` settings.

`code_actions_on_format` was slated to be removed from Zed (in favour of `code_action`), but it broke some uses cases, so it was added back. That might be why.

MichaReiser:
> BTW do we want do add this in ty's documentation, or is this section [docs.astral.sh/ty/editors#other-editors](https://docs.astral.sh/ty/editors/#other-editors) enough?:

Adding Zed to other editors makes sense to me, considering that we have a dedicated tab for Zed on the editor settings page.

> I'll do a commit tonight.

Thank you!

LoicRiegel:
Okay tell me what you think. I put back "code_actions_on_format" since this is what has to be used at the moment apparently

MichaReiser:
Sorry for the late review

> Okay tell me what you think. I put back "code_actions_on_format" since this is what has to be used at the moment apparently

Do you have a reference for this. Our recommendation is to run code actions (fixes and organize imports) **before** format but `code_actions_on_format` (at least according to the inline documentation), runs **after** formatting. 

That's why I'd prefer to keep our existing recommended configuration, unless there are issues with it that we aren't aware of.

injust:
> Do you have a reference for this. Our recommendation is to run code actions (fixes and organize imports) **before** format but `code_actions_on_format` (at least according to the inline documentation), runs **after** formatting.

See https://github.com/zed-industries/zed/pull/40409. Zed removed `code_actions_on_format`, then added it back and included a migrator that will convert this:

```json
{
    "languages": {
        "Python": {
            "formatter": [{ "code_action": "source.fixAll.ruff" }]
        }
    }
}
```

to this:

```json
{
    "languages": {
        "Python": {
            "code_actions_on_format": { "source.fixAll.ruff": true }
        }
    }
}
```

injust:
cc @probably-neb if you have more insight on the `code_action` -> `code_actions_on_format` migrator causing side effects for code action ordering before/after formatting.

probably-neb:
Thanks for the ping @injust 

> Do you have a reference for this. Our recommendation is to run code actions (fixes and organize imports) before format but code_actions_on_format (at least according to the inline documentation), runs after formatting.

Where are you seeing this documentation @MichaReiser? I'd love to fix it. I wrote the code that runs `code_actions_on_format` before the format steps specified in the `"formatter"` setting.

MichaReiser:
> Where are you seeing this documentation @MichaReiser? I'd love to fix it. I wrote the code that runs code_actions_on_format before the format steps specified in the "formatter" setting.

Oh great! That's exactly what we need. The description is from Zed's on hover tooltip

Screenshot 2025-11-04 at 16 57 32

probably-neb:
Docs are fixed in https://github.com/zed-industries/zed/pull/42128. Thanks all!

PR #20937, opened 2025-10-17
[ty] Re-enable fuzzer seeds that are no longer slow

AlexWaygood:
## Summary

And set a much smaller timeout so that it becomes obvious if any seeds become pathologically slow again

## Test Plan

CI on this PR

### DIFF
--- a/python/py-fuzzer/fuzz.py
+++ b/python/py-fuzzer/fuzz.py
@@ -152,16 +152,13 @@ class FuzzResult:
 
 def fuzz_code(seed: Seed, args: ResolvedCliArgs) -> FuzzResult:
     """Return a `FuzzResult` instance describing the fuzzing result from this seed."""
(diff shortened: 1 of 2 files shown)


python/py-fuzzer/fuzz.py
```diff
@@ -152,16 +152,13 @@ def print_description(self, index: int, num_seeds: int) -> None:
 
 def fuzz_code(seed: Seed, args: ResolvedCliArgs) -> FuzzResult:
     """Return a `FuzzResult` instance describing the fuzzing result from this seed."""
-    # TODO debug slowness of these seeds
-    skip_check = seed in {32, 56, 208}
-
```
MichaReiser, line 157: Nice, so we're back to that one failing seed (that panics)

AlexWaygood: Hopefully...!

PR #20942, opened 2025-10-17
[`ruff`] Update schemars to v1

### LINKED ISSUE #19173: Update schemars to v1
Upgrade the schemars crate to v1, see https://github.com/astral-sh/ruff/pull/19171

There are some breaking changes (mainly to the schema type) that our code needs to be adopted to.

TaKO8Ki:
## Summary

FIxes #19173

Update schemars to v1

**I’ve verified that the fundamentals of ruff.schema.json and ty.schema.json remain unchanged following the Schemars upgrade.**

The migration is based on https://graham.cool/schemars/migrating.

### Shared Changes

* ~~Upgraded both schemas to **JSON Schema draft 2020-12** andmigrated shared types into **`$defs`**, with all `$ref` paths now using `#/$defs/...`.~~
* Added reusable helpers such as `$defs/string` (and array helpers like `Array_of_string`) to avoid duplicating scalar/collection shapes.
* Reformatted descriptions with consistent newline breaks. **[Schemars 1.0]** — since **v1.0.0-alpha.3**, Schemars stops collapsing doc-comment whitespace; multi-line Rust docs flow through to `description` verbatim. ref: https://github.com/GREsau/schemars/issues/120
* Kept existing **deprecated** flags but moved them earlier within each property object to match the new formatting. 

### ty.schema.json

* Introduced `$defs` entries for path-related types (`RelativePathBuf`, `SystemPathBuf`) and updated path properties to reference them.
* Redirected overrides to a new `OverridesOptions` array definition with expanded multi-line documentation and examples, while maintaining the original rule descriptions under the updated formatting.

## Test Plan

Check if existing tests pass.

### DIFF
--- a/crates/ruff_workspace/src/options.rs
+++ b/crates/ruff_workspace/src/options.rs
@@ -601,7 +601,6 @@ impl schemars::JsonSchema for DeprecatedTopLevelLintOptions {
 // Don't add any new options to this struct. Add them to [`LintOptions`] directly to avoid exposing them in the
 // global settings.
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
-#[cfg_attr(feature = "schemars", schemars(inline))]
 #[derive(
     Clone, Debug, PartialEq, Eq, Default, OptionsMetadata, CombineOptions, Serialize, Deserialize,
 )]
@@ -3890,7 +3889,6 @@ pub struct AnalyzeOptions {
 
 /// Like [`LintCommonOptions`], but with any `#[serde(flatten)]` fields inlined. This leads to far,
 /// far better error messages when deserializing.
-#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct LintOptionsWire {
--- a/crates/ty_python_semantic/src/python_platform.rs
+++ b/crates/ty_python_semantic/src/python_platform.rs
@@ -76,6 +76,7 @@ mod schema {
                 schema.into()
             }
 
+            // Hard code some well known values, but allow any other string as well.
             let mut any_of = vec![schemars::json_schema!({ "type": "string" }).into()];
             // Promote well-known values for better auto-completion.
             // Using `const` over `enumValues` as recommended [here](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#documenting-enums).


ruff.schema.json
```diff
@@ -1415,7 +1405,7 @@
             "null"
           ],
           "items": {
-            "type": "string"
+            "$ref": "#/$defs/string"
```
TaKO8Ki: These changes looks weird, but they refer to `Name` struct, which means String in schemars. That is why they use `defs/string` instead of `string`.

https://github.com/astral-sh/ruff/blob/c4240076459bc9128fee6f83cbf0f51d134b663b/crates/ruff_workspace/src/options.rs#L1970-L1977

https://github.com/astral-sh/ruff/blob/c4240076459bc9128fee6f83cbf0f51d134b663b/crates/ruff_workspace/src/options.rs#L1960-L1968

https://github.com/astral-sh/ruff/blob/c4240076459bc9128fee6f83cbf0f51d134b663b/crates/ruff_python_ast/src/name.rs#L204-L227

ruff.schema.json
```diff
@@ -1185,7 +1175,7 @@
...
```
TaKO8Ki: Same as below.

https://github.com/astral-sh/ruff/pull/20942#discussion_r2440788641

ty.schema.json
```diff
@@ -1104,6 +1128,9 @@
         }
       },
       "additionalProperties": false
+    },
+    "string": {
+      "type": "string"
```
TaKO8Ki, line 1137: This is related to https://github.com/astral-sh/ruff/pull/20942#discussion_r2440788641

ruff.schema.json
```diff
@@ -4368,18 +4340,17 @@
...
           "type": "string",
-          "enum": [
-            "all"
-          ]
+          "const": "all"
         }
       ]
+    },
+    "string": {
+      "type": "string"
```
TaKO8Ki: This is related to https://github.com/astral-sh/ruff/pull/20942#discussion_r2440788641.

ty.schema.json
```diff
@@ -1031,56 +1045,66 @@
...
-          "deprecated": true,
-          "type": [
-            "string",
-            "null"
-          ]
+          "description": "The root of the project, used for finding first-party modules.\n\nIf left unspecified, ty will try to detect common project layouts and initialize `src.root` accordingly:\n\n* if a `./src` directory exists, include `.` and `./src` in the first party search path (src layout or flat)\n* if a `./<project-name>/<project-name>` directory exists, include `.` and `./<project-name>` in the first party search path\n* otherwise, default to `.` (flat layout)\n\nBesides, if a `./tests` directory exists and is not a package (i.e. it does not contain an `__init__.py` file),\nit will also be included in the first party search path.",
+          "anyOf": [
+            {
+              "$ref": "#/$defs/RelativePathBuf"
```
TaKO8Ki: With **Schemars 1.0** (defaulting to JSON Schema **2020-12**), named newtypes are **emitted as reusable `$ref`s in `#/$defs` by default**, even when `#[serde(transparent)]` makes their schema equal to the inner field. This preserves **type identity** and **deduplicates** shared shapes.

https://github.com/astral-sh/ruff/blob/c4240076459bc9128fee6f83cbf0f51d134b663b/crates/ty_project/src/metadata/value.rs#L330-L332

ty.schema.json
```diff
@@ -1031,56 +1045,66 @@
...
+          ],
+          "deprecated": true
         }
       },
       "additionalProperties": false
     },
+    "SystemPathBuf": {
+      "description": "An owned, mutable path on [`System`](`super::System`) (akin to [`String`]).\n\nThe path is guaranteed to be valid UTF-8.",
+      "type": "string"
+    },
```
TaKO8Ki, line 1111: This is related to https://github.com/astral-sh/ruff/pull/20942#discussion_r2440836378.

ty.schema.json
```diff
@@ -246,7 +256,7 @@
       "anyOf": [
         {
           "type": "string",
-          "pattern": "^\\d+\\.\\d+$"
+          "pattern": "^\\\\d+\\\\.\\\\d+$"
```
MichaReiser: The escaping here looks incorrect to me

crates/ruff_python_ast/src/name.rs
```diff
@@ -203,30 +203,26 @@ impl PartialEq<Name> for &String {
 
 #[cfg(feature = "schemars")]
 impl schemars::JsonSchema for Name {
-    fn is_referenceable() -> bool {
-        String::is_referenceable()
-    }
-
-    fn schema_name() -> String {
-        String::schema_name()
+    fn schema_name() -> std::borrow::Cow<'static, str> {
```
MichaReiser: I think we could replace this with

```rust
#[cfg_attr(
    feature = "serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(transparent)
)]
#[cfg_attr(feature = "cache", derive(ruff_macros::CacheKey))]
#[cfg_attr(feature = "salsa", derive(salsa::Update))]
#[cfg_attr(feature = "get-size", derive(get_size2::GetSize))]
#[cfg_attr(
    feature = "schemars",
    derive(schemars::JsonSchema),
    schemars(with = "String")
)]
```

(serde(transparent) and `schemars(with="String")`)

crates/ruff_python_ast/src/python_version.rs
```diff
@@ -188,42 +188,39 @@ mod serde {
...
-                ..SchemaObject::default()
-            })
+        fn json_schema(_gen: &mut SchemaGenerator) -> Schema {
+            let string_with_pattern = schemars::json_schema!({
+                "type": "string",
+                "pattern": r"^\\d+\\.\\d+$",
+            });
+
+            let mut any_of: Vec<Value> = Vec::new();
+            any_of.push(string_with_pattern.into());
```
MichaReiser: Nit
```suggestion
            let mut any_of: Vec<Value> = vec![
            	schemars::json_schema!({
	                "type": "string",
	                "pattern": r"^\\d+\\.\\d+$",
	            })
            ];
```

crates/ruff_workspace/src/options.rs
```diff
@@ -3938,8 +3938,8 @@ pub struct LintOptionsWire {
     pyflakes: Option<PyflakesOptions>,
     pylint: Option<PylintOptions>,
     pyupgrade: Option<PyUpgradeOptions>,
-    per_file_ignores: Option<FxHashMap<String, Vec<RuleSelector>>>,
-    extend_per_file_ignores: Option<FxHashMap<String, Vec<RuleSelector>>>,
+    per_file_ignores: Option<BTreeMap<String, Vec<RuleSelector>>>,
+    extend_per_file_ignores: Option<BTreeMap<String, Vec<RuleSelector>>>,
```
MichaReiser: What's the reason for changing those fields to a `BTreeMap`?

crates/ty_project/src/metadata/options.rs
```diff
@@ -801,6 +799,59 @@ impl FromIterator<(RangedValue<String>, RangedValue<Level>)> for Rules {
     }
 }
 
+#[cfg(feature = "schemars")]
```
MichaReiser: Nit: I'd prefer to keep this in its own `mod` (and in the same position, it makes attributing changes with git easier)

crates/ty_python_semantic/src/python_platform.rs
```diff
@@ -58,77 +58,42 @@ impl Default for PythonPlatform {
...
-            Schema::Object(SchemaObject {
-                // Hard code some well known values, but allow any other string as well.
-                subschemas: Some(Box::new(SubschemaValidation {
-                    any_of: Some(vec![
-                        Schema::Object(SchemaObject {
-                            instance_type: Some(schemars::schema::InstanceType::String.into()),
-                            ..SchemaObject::default()
-                        }),
-                        // Promote well-known values for better auto-completion.
-                        // Using `const` over `enumValues` as recommended [here](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#documenting-enums).
```
MichaReiser, line 81: Can we preserve this comment

ruff.schema.json
```diff
@@ -1,15 +1,15 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
MichaReiser: SchemaStore's recommendation is to still use draft-07 because many editors lack support for newer drafts, see 

> We recommend using the draft-07 JSON schema version. Later versions of JSON Schema are not yet recommended for use in SchemaStore until IDE and language support improves for those versions.

https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#best-practices

That's why I think we should not change the schema version as part of this PR

TaKO8Ki: Ok. It's also possible to continue to use draft-07. I will replace the schema version.

crates/ty_project/src/metadata/options.rs
TaKO8Ki: What I want to do is implementing JsonSchema for Rules instead of `schemars(with = "schema::Rules")`, so I will keep the mod and implement it inside the module.

TaKO8Ki:
@MichaReiser Thank you for the review. I have addressed your comments!

MichaReiser:
This is great. Thank you so much!

PR #20946, opened 2025-10-17
Dogfood ty on py-fuzzer in CI

AlexWaygood:
## Summary

Ensure we feel the pain when we introduce new false positives or false negatives

## Test Plan

CI on this PR

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -277,7 +277,8 @@ jobs:
         run: cargo test -p ty_python_semantic --test mdtest || true
       - name: "Run tests"
         run: cargo insta test --all-features --unreferenced reject --test-runner nextest
-
+      # Dogfood ty on py-fuzzer
+      - run: uv run --project=./python/py-fuzzer cargo run -p ty check --project=./python/py-fuzzer
       # Check for broken links in the documentation.
       - run: cargo doc --all --no-deps
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -533,6 +535,11 @@ jobs:
           ./scripts/add_plugin.py test --url https://pypi.org/project/-test/0.1.0/ --prefix TST
           ./scripts/add_rule.py --name FirstRule --prefix TST --code 001 --linter test
       - run: cargo check
+      # Lint/format/type-check py-fuzzer
+      # (dogfooding with ty is done in a separate job)
+      - run: uv run --directory=./python/py-fuzzer mypy
```
MichaReiser, line 540: Is it worth adding mypy. Shouldn't we also feel the pain for false-negatives ;)

.github/workflows/ci.yaml
```diff
@@ -277,7 +277,8 @@ jobs:
         run: cargo test -p ty_python_semantic --test mdtest || true
       - name: "Run tests"
         run: cargo insta test --all-features --unreferenced reject --test-runner nextest
-
+      # Dogfood ty on py-fuzzer
+      - run: uv run --project=./python/py-fuzzer cargo run -p ty check --project=./python/py-fuzzer
```
sharkdp, line 281: Let's see how annoying it will be to add `# ty: ignore` comments...

.github/workflows/ci.yaml
AlexWaygood: the mypy config is already there and I already run mypy locally when I make changes to the script 😄

PR #20948, opened 2025-10-17
[ty] Add suggestion to unknown rule diagnostics, rename `unknown-rule` lint to `ignore-comment-unknown-rule`

MichaReiser:
## Summary

Show suggestions for misspelled rule names. 

Rename the `unknown-rule` lint because it overlaps with the non-suppressable `unknown-rule` error code used in configurations.

## Test Plan

Updated mdtests/CLI tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
+++ b/crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
@@ -127,7 +127,7 @@ a = 10 / 0  # ty: ignore[*-*]
 <!-- blacken-docs:off -->
 
 ```py
-a = 10 / 0  # ty: ignore[division-by-zero]
+a = 10 / 0  # ty: ignore[division-by-zero]       
             #                               ^^^^^^ trailing whitespace
 ```
 


crates/ty_python_semantic/src/diagnostic.rs
AlexWaygood, line 1: Should this module be merged with https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/util/diagnostics.rs ? I don't have a strong opinion about whether a top-level `diagnostic` module or a `util::diagnostic` module is better 😄

MichaReiser: I searched like 2 minutes for where to put it and didn't notice the module in util (probably because I have an aversion to util modules)

crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
```diff
@@ -127,7 +127,7 @@ a = 10 / 0  # ty: ignore[*-*]
 <!-- blacken-docs:off -->
 
 ```py
-a = 10 / 0  # ty: ignore[division-by-zero]      
+a = 10 / 0  # ty: ignore[division-by-zero]
             #                               ^^^^^^ trailing whitespace
```
AlexWaygood: judging by the comment, I think the trailing whitespace here was deliberate :-)

crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
```diff
@@ -88,7 +88,7 @@ def test($):  # ty: ignore
 ```py
 a = 10
 # revealed: Literal[10]
-# error: [unknown-rule] "Unknown rule `revealed-type`"
+# error: [ignore-comment-unknown-rule] "Unknown rule `revealed-type`"
```
AlexWaygood, line 91: Is it worth this even being a separate error code to `invalid-ignore-comment`? Is there any situation where you'd want to globally ignore one but not the other? (Is there any situation where you'd want to globally ignore either?!)

MichaReiser: I guess that depends on if we ever want to support `type: ignore[code]`. I could then see cases where you would want to disable `ignore-comment-unknown-rule` but not `invalid-ignore-comment` when using ty alongside mypy and pyright.

But you know that I'm almost always on-board with removing rules ;)

AlexWaygood: hmm, that's a reasonable point. I guess the current division keeps our options open!

PR #20950, opened 2025-10-17
[ty] Avoid promoting invariant generics during literal promotion

ibraheemdev:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1284.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6648,7 +6648,7 @@ impl<'db> Type<'db> {
         tcx: TypeContext<'db>,
         visitor: &ApplyTypeMappingVisitor<'db>,
     ) -> Type<'db> {
-        match self {
+        let ty = match self {
             Type::TypeVar(bound_typevar) => match type_mapping {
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -6837,6 +6837,15 @@ impl<'db> Type<'db> {
             | Type::BoundSuper(_)
             | Type::SpecialForm(_)
             | Type::KnownInstance(_) => self,
+        };
+
+        match type_mapping {
+            TypeMapping::PromoteLiterals => {
+                // It is only sound to promote to a supertype (i.e. it is unsound to promote
+                // invariant generics).
+                if self.is_subtype_of(db, ty) { ty } else { self }
```
ibraheemdev, line 6846: We could technically do better and partially promote types with mixed variance generics (e.g. `Callable`), but that doesn't seem necessary (and arguably, the only type we really care recursively promoting are tuples).

AlexWaygood:
While there are some new diagnostics on other projects too, the vast majority of the new diagnostics appear to be new `unknown-argument` diagnostics on ddtrace, which is pretty interesting

PR #20951, opened 2025-10-17
[ty] Provide completions at the end of the file

### LINKED ISSUE #1392: Missing completions in playground and VS Code
I recently added these tests: https://github.com/astral-sh/ruff/blob/8ca2b5555d47e211c04136637477c5227172d885/crates/ty_ide/src/completion.rs#L3899-L3946

They show that
```py
def f[T: str](msg: T):
    msg.<CURSOR>
```

should show completions like `upper` and `capitalize`. This does work in the LSP with neovim. But it does not work with VS Code or in the playground:

neovim:
Image

VS Code:
Image

Playground:
Image

What's very weird is that I *can* see those completions if I make other random changes to the file. For example, if I add a comment in that function:

Image

So I'm afraid this might be some caching problem?

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1392

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -409,7 +409,7 @@ impl<'t> CompletionTargetTokens<'t> {
             TokenAt::Single(tok) => tok.end(),
             TokenAt::Between(_, tok) => tok.start(),
         };
-        let before = tokens_start_before(parsed.tokens(), offset);
+        let before = strip_eof_newline(tokens_start_before(parsed.tokens(), offset));
         Some(
             // Our strategy when it comes to `object.attribute` here is
             // to look for the `.` and then take the token immediately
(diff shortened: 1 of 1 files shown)


crates/ty_ide/src/completion.rs
```diff
@@ -627,6 +627,19 @@ fn tokens_start_before(tokens: &Tokens, offset: TextSize) -> &[Token] {
...
+/// end of the file.
+fn strip_eof_newline(tokens: &[Token]) -> &[Token] {
+    if tokens
+        .last()
+        .is_some_and(|t| t.kind() == TokenKind::Newline)
+    {
+        &tokens[..tokens.len() - 1]
+    } else {
+        tokens
+    }
```
MichaReiser, line 640: Nit: Use https://doc.rust-lang.org/std/primitive.slice.html#method.split_last

sharkdp: Thanks. This is still in draft because it's authored 100% by Claude and I still wanted to review (and understand) it myself.

PR #20953, opened 2025-10-18
Refactor format tests to use CliTest helper

Renkai:
Part of https://github.com/astral-sh/ruff/issues/19016

### DIFF
--- a/crates/ruff/tests/cli/format.rs
+++ b/crates/ruff/tests/cli/format.rs
@@ -6,13 +6,13 @@ use std::path::Path;
 use anyhow::Result;
 use insta_cmd::assert_cmd_snapshot;
 
-use super::CliTest;
+use super::{CliTest, tempdir_filter};
 
 #[test]
 fn default_options() -> Result<()> {
     let test = CliTest::new()?;
     assert_cmd_snapshot!(test.format_command()
-        .args(["--stdin-filename", "test.py"])
+        .args(["--isolated", "--stdin-filename", "test.py"])
         .arg("-")
         .pass_stdin(r#"
 def foo(arg1, arg2,):
@@ -44,12 +44,14 @@ if condition:
 
 #[test]
 fn default_files() -> Result<()> {
-    let test = CliTest::new()?;
-    test.write_file("foo.py", r#"foo =     "needs formatting""#)?;
-    test.write_file("bar.py", r#"bar =     "needs formatting""#)?;
+    let test = CliTest::with_files([
+        ("foo.py", r#"foo =     "needs formatting""#),
+        ("bar.py", r#"bar =     "needs formatting""#),
+    ])?;
 
-    assert_cmd_snapshot!(test.command()
-        .args(["format", "--isolated", "--no-cache", "--check"]), @r"
+    assert_cmd_snapshot!(test.format_command()
+        .arg("--isolated")
+        .arg("--check"), @r"
     success: false
     exit_code: 1
     ----- stdout -----
@@ -67,7 +69,7 @@ fn default_files() -> Result<()> {
 fn format_warn_stdin_filename_with_files() -> Result<()> {
     let test = CliTest::new()?;
     assert_cmd_snapshot!(test.format_command()
-        .args(["--stdin-filename", "foo.py"])
+        .args(["--isolated", "--stdin-filename", "foo.py"])
         .arg("foo.py")
         .pass_stdin("foo =     1"), @r"
     success: true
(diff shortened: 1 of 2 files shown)


crates/ruff/tests/cli/format.rs
```diff
@@ -782,7 +734,14 @@ if __name__ == "__main__":
     ");
 
     // Second format should exit with code 0 since no files needed formatting
-    assert_cmd_snapshot!(cmd, @r"
+    assert_cmd_snapshot!(test.command()
```
MichaReiser: Nit: You could avoid some repetition by adding a `format_command` method to `test` that always sets `format --no-cache` and maybe `--isolated`

crates/ruff/tests/cli/main.rs
```diff
@@ -15,13 +15,34 @@ use std::{
...
+        // - Backslash: \
+        // This handles paths in JSON output, file URIs, and regular Windows paths
+        let pattern = escaped.replace("/", r"(?:/|\\/|\\)");
+
+        // Include optional leading slash to match the third slash in file:///
+        // and optional trailing slash, so file:/// becomes file:// after replacement
+        format!(r"(?:/|\\)?{}(?:/|\\/|\\)?", pattern)
+    } else {
+        format!(r"{}[\\/]?", regex::escape(path_str))
+    }
```
MichaReiser: Can you tell me why this is now necessary? Is it because of the output format tests? If so, could we use the same approach as in linter tests https://github.com/astral-sh/ruff/blob/15af4c0a34dd6d86d9aa98115ae1e1e9392c7eef/crates/ruff/tests/cli/lint.rs#L3293-L3298

crates/ruff/tests/cli/snapshots/cli__format__output_format_grouped.snap
```diff
@@ -1,12 +1,13 @@
...
-source: crates/ruff/tests/format.rs
+source: crates/ruff/tests/cli/format.rs
 info:
   program: ruff
   args:
     - format
     - "--no-cache"
     - "--output-format"
     - grouped
+    - "--preview"
```
MichaReiser, line 10: Is using `--preview` here intentional?

crates/ruff/tests/cli/format.rs
```diff
@@ -50,22 +46,12 @@ if condition:
 
 #[test]
 fn default_files() -> Result<()> {
```
MichaReiser, line 46: I suggest that we use `CliTest` for all tests, including tests that don't use `with_files` or `write_file`. Using `CliTest` has the benefit that it e.g. clears all environment variables, ensuring better test isolation

crates/ruff/tests/cli/main.rs
Renkai: Yes it's for  output format tests. I'm not successful yet, windows is too weird 😢. Will take a look at your suggested approach soon.

Renkai: Why should we add a different filter in a test, but not the common class?

MichaReiser: I decided to use a specific filter for that test because the criteria aren't very specific, risking that we accidentally filter out unrelated content.

crates/ruff/tests/cli/snapshots/cli__format__output_format_grouped.snap
Renkai: The diff is from https://github.com/astral-sh/ruff/pull/20443 I think, not sure why the snapshot is not change on his PR but mine

crates/ruff/tests/cli/format.rs
```diff
@@ -239,15 +225,16 @@ def foo():
 
 #[test]
 fn config_doubly_overridden_via_cli() -> Result<()> {
-    let tempdir = TempDir::new()?;
-    let ruff_toml = tempdir.path().join("ruff.toml");
-    fs::write(&ruff_toml, "line-length = 70")?;
+    let test = CliTest::new()?;
+    test.write_file("ruff.toml", "line-length = 70")?;
+
+    let ruff_toml = test.root().join("ruff.toml");
```
MichaReiser: We can just pass "ruff.toml" now, because `CliTest::command` ensures that the current working directory is the test root.

crates/ruff/tests/cli/format.rs
```diff
@@ -629,52 +596,52 @@ if __name__ == "__main__":
...
+    let test = CliTest::new()?;
+    let project_root = Path::new(env!("CARGO_MANIFEST_DIR"));
+    let fixtures = project_root.join("resources").join("test").join("fixtures");
     let path = fixtures.join("unformatted.ipynb");
+    // Normalize path separators to forward slashes for cross-platform compatibility
+    let project_root_normalized = project_root
+        .join("resources")
+        .display()
+        .to_string()
+        .replace('\\', "/");
```
MichaReiser: What's the reason what we need those new escapes that weren't needed before? Are they also needed when using `CliTest::with_settings`?

MichaReiser (approved):
Thank you

PR #20955, opened 2025-10-18
[ty] Fix panic on recursive class definitions in a stub that use constrained type variables

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1390

## Test Plan

Added mdtests that cause us to panic on `main`

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8303,7 +8303,11 @@ impl<'db> TypeVarInstance<'db> {
         ))
     }
 
-    #[salsa::tracked(cycle_fn=lazy_bound_cycle_recover, cycle_initial=lazy_bound_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -8324,7 +8324,7 @@ impl<'db> TypeVarInstance<'db> {
         Some(TypeVarBoundOrConstraints::UpperBound(ty))
     }
 
-    #[salsa::tracked(heap_size=ruff_memory_usage::heap_size)]
+    #[salsa::tracked(cycle_fn=lazy_bound_cycle_recover, cycle_initial=lazy_bound_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
```
MichaReiser: Nit: Rename the cycle function to `lazy_bound_or_constraints_recover`?

AlexWaygood: aw but then it's not a 1-line fix anymore 🤣

PR #20958, opened 2025-10-18
Reduce duplicated logic between the macOS and Windows CI jobs

AlexWaygood:
## Summary

These CI jobs have almost identical logic in them. We can just use a matrix for the runner OS, to avoid writing it all out twice (which is bug-prone). Neither is marked as being required for (auto-)merge to succeed.

## Test Plan

CI on this PR

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -334,9 +334,14 @@ jobs:
       - name: "Run tests"
         run: cargo insta test --release --all-features --unreferenced reject --test-runner nextest
 
-  cargo-test-windows:
-    name: "cargo test (windows)"
-    runs-on: ${{ github.repository == 'astral-sh/ruff' && 'depot-windows-2022-16' || 'windows-latest' }}
+  cargo-test-other:
+    strategy:
+      matrix:
+        platform:
+          - ${{ github.repository == 'astral-sh/ruff' && 'depot-windows-2022-16' || 'windows-latest' }}
+          - macos-latest
+    name: "cargo test (${{ matrix.platform }})"
+    runs-on: ${{ matrix.platform }}
     needs: determine_changes
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
     timeout-minutes: 20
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -334,38 +334,14 @@ jobs:
...
-        with:
-          tool: cargo-nextest
-      - name: "Install uv"
-        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0
-        with:
-          enable-cache: "true"
-      - name: "Run tests"
-        env:
-          # Workaround for <https://github.com/nextest-rs/nextest/issues/1493>.
-          RUSTUP_WINDOWS_PATH_ADD_BIN: 1
```
AlexWaygood, line 361: This issue has long-since been closed, so there's no need to set this environment variable anymore

.github/workflows/ci.yaml
```diff
@@ -376,8 +352,6 @@ jobs:
       - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
       - name: "Install Rust toolchain"
         run: rustup show
-      - name: "Install mold"
-        uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
```
AlexWaygood, line 380: This action is a no-op on both Windows and macOS, so this step was redundant: https://github.com/rui314/setup-mold/blob/725a8794d15fc7563f59595bd9556495c0564878/action.yml#L20

.github/workflows/ci.yaml
```diff
@@ -675,7 +649,7 @@ jobs:
       - determine_changes
     # Only runs on pull requests, since that is the only we way we can find the base version for comparison.
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && github.event_name == 'pull_request' && (needs.determine_changes.outputs.ty == 'true' || needs.determine_changes.outputs.py-fuzzer == 'true') }}
-    timeout-minutes: 5
+    timeout-minutes: ${{ github.repository == 'astral-sh/ruff' && 5 || 20 }}
```
AlexWaygood, line 652: the 5-minute timeout isn't long enough on forks, where a free GitHub runner will be used instead of a depot runner

PR #20961, opened 2025-10-18
[ty] `dataclass_transform`: Support for fields with an `alias`

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1385

## Conformance tests

Two false positives removed, as expected.

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
@@ -451,7 +451,7 @@ checkers do not seem to support this either.
 ```py
 from typing_extensions import dataclass_transform, Any
 
-def fancy_field(*, init: bool = True, kw_only: bool = False) -> Any: ...
+def fancy_field(*, init: bool = True, kw_only: bool = False, alias: str | None = None) -> Any: ...
 @dataclass_transform(field_specifiers=(fancy_field,))
 def fancy_model[T](cls: type[T]) -> type[T]:
     ...
@@ -460,7 +460,7 @@ def fancy_model[T](cls: type[T]) -> type[T]:
 @fancy_model
 class Person:
     id: int = fancy_field(init=False)
-    name: str = fancy_field()
+    internal_name: str = fancy_field(alias="name")
     age: int | None = fancy_field(kw_only=True)
 
 reveal_type(Person.__init__)  # revealed: (self: Person, name: str, *, age: int | None) -> None
@@ -468,7 +468,7 @@ reveal_type(Person.__init__)  # revealed: (self: Person, name: str, *, age: int 
 alice = Person("Alice", age=30)
 
 reveal_type(alice.id)  # revealed: int
-reveal_type(alice.name)  # revealed: str
+reveal_type(alice.internal_name)  # revealed: str
 reveal_type(alice.age)  # revealed: int | None
 ```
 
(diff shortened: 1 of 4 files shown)


sharkdp:
I noticed that https://github.com/astral-sh/ty/issues/1159 is still not fixed after this PR (only one of the two errors goes away). But I don't think anything is wrong with `dataclass_transform` or `field_specifiers`. There's something wrong with overload resolution or generics, which causes us to infer unknown for calls to Pydantics `Field` if `default` and `alias` are both passed as keyword arguments. Everything works fine if `default` is passed positional-only or not at all. Will address/report this separately.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8000,6 +8000,9 @@ pub struct FieldInstance<'db> {
 
     /// Whether or not this field can only be passed as a keyword argument to `__init__`.
     pub kw_only: Option<bool>,
+
+    /// This name is used to provide an alternative parameter name in the synthesized `__init__` method.
+    pub alias: Option<Box<str>>,
```
AlexWaygood, line 8005: It's possible using `Option<Name>` might lead to fewer allocations here

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -650,6 +653,14 @@ impl<'db> Bindings<'db> {
                             None
                         };
 
+                        let alias = alias.and_then(|alias_ty| {
+                            if let Type::StringLiteral(s) = alias_ty {
+                                Some(s.value(db).to_string().into_boxed_str())
+                            } else {
+                                None
+                            }
+                        });
```
```suggestion
                        let alias = alias
                            .and_then(Type::as_string_literal)
                            .map(|literal| Box::from(literal.value(db)));
```

(though this would need to be revised slightly if you do refactor to use `Option<Name>` instead of `Option<Box<str>>` 😄

crates/ty_python_semantic/src/types.rs
sharkdp: Yes, but then `FieldInstance` itself will be larger for *every* field, not just those that use `alias`. I haven't measured it but I'd rather go with the more memory-conservative option here.

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -549,6 +549,58 @@ reveal_type(Person.__init__)  # revealed: (self: Person, name: str) -> None
...
+
+def field_with_alias(*, alias: str | None = None, kw_only: bool = False) -> Any: ...
+@dataclass_transform(field_specifiers=(field_with_alias,))
+def model[T](cls: type[T]) -> type[T]:
+    return cls
+
+@model
+class Person:
+    internal_name: str = field_with_alias(alias="name")
+    internal_age: int = field_with_alias(alias="age", kw_only=True)
```
AlexWaygood, line 568: It looks like we just ignore the `alias` argument if you try to pass a value for `alias` that's an instance of `str`, but isn't a string literal. We should probably emit a diagnostic for that, to warn the user that we can't figure out what they're trying to do? Though it's admittedly a bit of an edge case. E.g.

```py
from typing_extensions import dataclass_transform, Any

def field_with_alias(*, alias: str | None = None, kw_only: bool = False) -> Any: ...
@dataclass_transform(field_specifiers=(field_with_alias,))
def model[T](cls: type[T]) -> type[T]:
    return cls

def _(alias: str):
    @model
    class Person:
        internal_name: str = field_with_alias(alias=alias)
        internal_age: int = field_with_alias(alias=alias, kw_only=True)
    
    reveal_type(Person.__init__)
```

sharkdp: Right. Will write this down.

PR #20962, opened 2025-10-18
[ty] Type inference for comprehensions

sharkdp:
## Summary

Adds type inference for list/dict/set comprehensions, including bidirectional inference:

```py
reveal_type({k: v for k, v in [("a", 1), ("b", 2)]})  # dict[Unknown | str, Unknown | int]

squares: list[int | None] = [x for x in range(10)]
reveal_type(squares)  # list[int | None]
```

## Ecosystem impact

I did spot check the changes and most of them seem like known limitations or true positives. Without proper bidirectional inference, we saw a lot of false positives.

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/regression/pr_20962_comprehension_panics.md
+++ b/crates/ty_python_semantic/resources/mdtest/regression/pr_20962_comprehension_panics.md
@@ -1,0 +1,50 @@
+# Documentation of two fuzzer panics involving comprehensions
+
+Type inference for comprehensions was added in <https://github.com/astral-sh/ruff/pull/20962>. It
+added two new fuzzer panics that are documented here for regression testing.
+
+## Too many cycle iterations in `place_by_id`
+
+<!-- expect-panic: too many cycle iterations -->
+
+```py
+name_5(name_3)
+[0 for unique_name_0 in unique_name_1 for unique_name_2 in name_3]
+
+@{name_3 for unique_name_3 in unique_name_4}
+class name_4[**name_3](0, name_2=name_5):
+    pass
+
+try:
+    name_0 = name_4
+except* 0:
+    pass
+else:
+    match unique_name_12:
+        case 0:
+            from name_2 import name_3
+        case name_0():
+
+            @name_4
+            def name_3():
+                pass
+
+(name_3 := 0)
+
+@name_3
+async def name_5():
+    pass
+```
+
+## Too many cycle iterations in `infer_definition_types`
+
+<!-- expect-panic: too many cycle iterations -->
+
+```py
+for name_1 in {
+    {{0: name_4 for unique_name_0 in unique_name_1}: 0 for unique_name_2 in unique_name_3 if name_4}: 0
+    for unique_name_4 in name_1
+    for name_4 in name_1
+}:
+    pass
+```


crates/ty_python_semantic/resources/mdtest/comprehensions/basic.md
```diff
@@ -103,3 +103,94 @@ async def _():
...
+
+This also works for nested comprehensions:
+
+```py
+table = [[(x, y) for x in range(3)] for y in range(3)]
+reveal_type(table)  # revealed: list[list[tuple[int, int]]]
+
+# TODO: no error here
+# error: [invalid-assignment]
+table_with_content: list[list[tuple[int, int, str | None]]] = [[(x, y, None) for x in range(3)] for y in range(3)]
```
sharkdp, line 155: As discussed with @ibraheemdev, solving this will require pushing down the type context into `infer_scope_types` in order to infer the `None` in `(x, y, None)` as `str | None` inside the comprehension. Similar for the `TypedDict` test below. This change is a bit more invasive, as it requires us to skip comprehension scopes in `check_types`, because we would not have the required type context available when simply looping over all scopes. This change would be fine because with this PR, we're now calling `infer_scope_types` for comprehension scopes when we check the *outer* scope. A similar change would probably be required for lambdas as well.

AlexWaygood (approved):
LGTM!

AlexWaygood:
Looks like the remaining fuzzer panics are seed 20, which minimize to this:

```py
name_5(name_3)
[0 for unique_name_0 in unique_name_1 for unique_name_2 in name_3]

@{name_3 for unique_name_3 in unique_name_4}
class name_4[**name_3](0, name_2=name_5):
    pass
try:
    name_0 = name_4
except* 0:
    pass
else:
    match unique_name_12:
        case 0:
            from name_2 import name_3
        case name_0():

            @name_4
            def name_3():
                pass
(name_3 := 0)

@name_3
async def name_5():
    pass
```

with this stacktrace:

```
error[panic]: Panicked at /Users/alexw/.cargo/git/checkouts/salsa-e6f3bb7c2a062968/cdd0b85/src/function/execute.rs:419:17 when checking `/Users/alexw/dev/ruff/foo.py`: `place_by_id(Id(3002)): execute: too many cycle iterations`
info: This indicates a bug in ty.
info: If you could open an issue at https://github.com/astral-sh/ty/issues/new?title=%5Bpanic%5D, we'd be very appreciative!
info: Platform: macos aarch64
info: Version: ruff/0.14.3+22 (8f293eaf3 2025-10-31)
info: Args: ["target/debug/ty", "check", "foo.py", "--python-version=3.13"]
info: run with `RUST_BACKTRACE=1` environment variable to show the full backtrace information
info: query stacktrace:
   0: infer_expression_types_impl(Id(1401))
             at crates/ty_python_semantic/src/types/infer.rs:184
   1: infer_definition_types(Id(1801))
             at crates/ty_python_semantic/src/types/infer.rs:94
   2: infer_scope_types(Id(1001))
             at crates/ty_python_semantic/src/types/infer.rs:70
   3: infer_scope_types(Id(1000))
             at crates/ty_python_semantic/src/types/infer.rs:70
   4: check_file_impl(Id(c00))
             at crates/ty_project/src/lib.rs:535
```

and seed 486, which minimizes to this:

```py
for name_1 in {{{0: name_4 for unique_name_0 in unique_name_1}: 0 for unique_name_2 in unique_name_3 if name_4}: 0 for unique_name_4 in name_1 for name_4 in name_1}:
    pass
```

with this stacktrace:

```
error[panic]: Panicked at /Users/alexw/.cargo/git/checkouts/salsa-e6f3bb7c2a062968/cdd0b85/src/function/execute.rs:419:17 when checking `/Users/alexw/dev/ruff/bar.py`: `infer_definition_types(Id(1804)): execute: too many cycle iterations`
info: This indicates a bug in ty.
info: If you could open an issue at https://github.com/astral-sh/ty/issues/new?title=%5Bpanic%5D, we'd be very appreciative!
info: Platform: macos aarch64
info: Version: ruff/0.14.3+22 (8f293eaf3 2025-10-31)
info: Args: ["target/debug/ty", "check", "bar.py", "--python-version=3.13"]
info: run with `RUST_BACKTRACE=1` environment variable to show the full backtrace information
info: query stacktrace:
   0: infer_scope_types(Id(1000))
             at crates/ty_python_semantic/src/types/infer.rs:70
   1: check_file_impl(Id(c00))
             at crates/ty_project/src/lib.rs:535
```

ISTM that neither should block this being merged, but we could possibly add expect-panicking mdtests so we get notified when they get fixed.

sharkdp:
Yes, thank you. I was also looking into these earlier today and came to the same conclusion. Will panics in the baseline 
 cause longer CI times? Or does the costly minimization only kick in for the new panics?

AlexWaygood:
> Will panics in the baseline cause longer CI times? Or does the costly minimization only kick in for the new panics?

The costly minimization only kicks in for the new panics that don't exist in the baseline executable -- this should be fine!

AlexWaygood:
Okay, in CI it does seem like this makes us a fair bit slower to execute on seed 742 (a seed which we're not panicking on). But I assume that this is another instance where our performance is much slower on a debug build than on a release build; the fuzzer script still completes in a very reasonable time locally when I use a release build.

If it feels like the fuzzer CI job is now taking an annoying amount of time, we could experiment with using `--profile=profiling` like we do for mypy_primer jobs in CI.

PR #20967, opened 2025-10-19
[ty] Add cycle handling to `lazy_default`

MichaReiser:
## Summary

Add cycle handling to `lazy_default`.

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -316,10 +316,7 @@ static_assert(is_subtype_of(Bottom[JsonDict], Top[JsonDict]))
 ### Cyclic defaults
 
 ```py
-from __future__ import annotations
-
-from typing import Protocol
-from typing_extensions import TypeVar
+from typing_extensions import Protocol, TypeVar
 
 T = TypeVar("T", default="C", covariant=True)
 
@@ -328,6 +325,9 @@ class P(Protocol[T]):
 
 class C(P[T]):
     pass
+
+reveal_type(C[int]())  # revealed: C[int]
+reveal_type(C())  # revealed: C[Divergent]
 ```
 
 ### Union inside generic
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8344,7 +8344,7 @@ impl<'db> TypeVarInstance<'db> {
         Some(TypeVarBoundOrConstraints::Constraints(ty))
     }
 
-    #[salsa::tracked(cycle_fn=lazy_default_cycle_recover, cycle_initial=lazy_default_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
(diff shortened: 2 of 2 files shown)


crates/ty_python_semantic/resources/primer/bad.txt
```diff
@@ -1,2 +1,2 @@
 spark  # too many iterations (in `exported_names` query), `should not be able to access instance member `spark` of type variable IndexOpsLike@astype in inferable position`
```
AlexWaygood, line 1: The second of the two panics mentioned here is presumably gone after Doug's recent work in https://github.com/astral-sh/ruff/pull/20677?

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -313,6 +313,26 @@ static_assert(is_subtype_of(Bottom[JsonDict], Bottom[JsonDict]))
 static_assert(is_subtype_of(Bottom[JsonDict], Top[JsonDict]))
 ```
 
+
+### Cyclic defaults
```
MichaReiser, line 316: This is a bit too advanced Python typing for me 😅 

I think it would be good to have some `reveal_type` or other assertion in this test but I need some help with what that assertion could be.

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -313,6 +313,23 @@ static_assert(is_subtype_of(Bottom[JsonDict], Bottom[JsonDict]))
 static_assert(is_subtype_of(Bottom[JsonDict], Top[JsonDict]))
 ```
 
+### Cyclic defaults
+
+```py
+
+from typing import Protocol
+from typing_extensions import TypeVar
```
sharkdp: Minor:
```suggestion
from typing_extensions import Protocol, TypeVar
```

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -313,6 +313,23 @@ static_assert(is_subtype_of(Bottom[JsonDict], Bottom[JsonDict]))
 static_assert(is_subtype_of(Bottom[JsonDict], Top[JsonDict]))
 ```
 
+### Cyclic defaults
+
+```py
+
```
sharkdp: This shouldn't be necessary if `"C"` is a stringified annotation.
```suggestion
```

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -313,6 +313,23 @@ static_assert(is_subtype_of(Bottom[JsonDict], Bottom[JsonDict]))
...
+from typing import Protocol
+from typing_extensions import TypeVar
+
+T = TypeVar("T", default="C", covariant=True)
+
+class P(Protocol[T]):
+    pass
+
+class C(P[T]):
+    pass
```
sharkdp, line 327: I think two assertions like these are probably sufficient here:
```suggestion
    pass

reveal_type(C[int]())  # revealed: C[int]
reveal_type(C())  # revealed: C[Divergent]
```

PR #20969, opened 2025-10-19
[ty] fix infinite recursion with generic type aliases

mtshiba:
## Summary

While working on #20900, I noticed that the calculation of variance for recursive type aliases was leading to a stack overflow.

This PR fixes the `apply_type_mapping_impl` handling of type aliases, avoiding infinite recursion.

## Test Plan

New tests in `generics/pep695/aliases.md` and `types::tests::type_alias_variance`.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6765,16 +6765,7 @@ impl<'db> Type<'db> {
                 // In the case of recursive type aliases, this leads to infinite recursion.
                 // Instead, call `raw_value_type` and perform the specialization after the `visitor` cache has been created.
                 let value_type = visitor.visit(self, || alias.raw_value_type(db).apply_type_mapping_impl(db, type_mapping, tcx, visitor));
-                if let Some(generic_context) = alias.generic_context(db) {
-                    let specialization = alias
-                        .specialization(db)
-                        .unwrap_or_else(|| generic_context.default_specialization(db, None));
-
-                    value_type.apply_specialization(db, specialization)
-                        .apply_type_mapping_impl(db, type_mapping, tcx, visitor)
-                } else {
-                    value_type
-                }
+                alias.apply_function_specialization(db, value_type).apply_type_mapping_impl(db, type_mapping, tcx, visitor)
             }
 
             Type::ModuleLiteral(_)
@@ -10719,17 +10710,7 @@ impl<'db> PEP695TypeAliasType<'db> {
 
     /// The RHS type of a PEP-695 style type alias with specialization applied.
     pub(crate) fn value_type(self, db: &'db dyn Db) -> Type<'db> {
-        let value_type = self.raw_value_type(db);
-
-        if let Some(generic_context) = self.generic_context(db) {
-            let specialization = self
-                .specialization(db)
-                .unwrap_or_else(|| generic_context.default_specialization(db, None));
-
-            value_type.apply_specialization(db, specialization)
-        } else {
-            value_type
-        }
+        self.apply_function_specialization(db, self.raw_value_type(db))
     }
 
     /// The RHS type of a PEP-695 style type alias with *no* specialization applied.
@@ -10741,6 +10722,17 @@ impl<'db> PEP695TypeAliasType<'db> {
         let definition = self.definition(db);
 
         definition_expression_type(db, definition, &type_alias_stmt_node.node(&module).value)
+    }
+
+    fn apply_function_specialization(self, db: &'db dyn Db, ty: Type<'db>) -> Type<'db> {
+        if let Some(generic_context) = self.generic_context(db) {
+            let specialization = self
+                .specialization(db)
+                .unwrap_or_else(|| generic_context.default_specialization(db, None));
+            ty.apply_specialization(db, specialization)
+        } else {
+            ty
+        }
     }
 
     pub(crate) fn apply_specialization(
@@ -10929,6 +10921,13 @@ impl<'db> TypeAliasType<'db> {
         match self {
             TypeAliasType::PEP695(type_alias) => type_alias.specialization(db),
             TypeAliasType::ManualPEP695(_) => None,
+        }
+    }
+
+    fn apply_function_specialization(self, db: &'db dyn Db, ty: Type<'db>) -> Type<'db> {
+        match self {
+            TypeAliasType::PEP695(type_alias) => type_alias.apply_function_specialization(db, ty),
+            TypeAliasType::ManualPEP695(_) => ty,
         }
     }
 


crates/ty_python_semantic/src/types.rs
```diff
@@ -6791,7 +6791,20 @@ impl<'db> Type<'db> {
             Type::TypeIs(type_is) => type_is.with_type(db, type_is.return_type(db).apply_type_mapping(db, type_mapping, tcx)),
 
             Type::TypeAlias(alias) => {
-                visitor.visit(self, || alias.value_type(db).apply_type_mapping_impl(db, type_mapping, tcx, visitor))
+                // Do not call `value_type` here. `value_type` does the specialization internally, so `apply_type_mapping` is performed without `visitor` inheritance.
+                // In the case of recursive type aliases, this leads to infinite recursion.
+                // Instead, call `raw_value_type` and perform the specialization after the `visitor` cache has been created.
+                let value_type = visitor.visit(self, || alias.raw_value_type(db).apply_type_mapping_impl(db, type_mapping, tcx, visitor));
```
AlexWaygood, line 6767: the warning note above `raw_value_type` (which I know you added just the other day!) says that we should be careful not to call it from functions that might be called while analysing other files:

https://github.com/astral-sh/ruff/blob/590ce9d9ed70db0578b14a4c93aaaba31485724a/crates/ty_python_semantic/src/types.rs#L10738-L10746

But `apply_type_mapping` _can_ be called while analysing other files, so I'm not sure this call is safe here. But maybe we could make it safe by just moving the Salsa caching from `PEP695TypeAliasType::value_type` to `PEP695TypeAliasType::raw_value_type`? No tests fail on your branch if I make this change:

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 7a16a1e4b7..dcef24a84d 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -10731,7 +10731,6 @@ impl<'db> PEP695TypeAliasType<'db> {
     }
 
     /// The RHS type of a PEP-695 style type alias with specialization applied.
-    #[salsa::tracked(cycle_fn=value_type_cycle_recover, cycle_initial=value_type_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
     pub(crate) fn value_type(self, db: &'db dyn Db) -> Type<'db> {
         let value_type = self.raw_value_type(db);
 
@@ -10756,6 +10755,7 @@ impl<'db> PEP695TypeAliasType<'db> {
     /// over-invalidation.
     /// This method also calls the type inference functions, and since type aliases can have recursive structures,
     /// we should be careful not to create infinite recursions in this method (or make it tracked if necessary).
+    #[salsa::tracked(cycle_fn=value_type_cycle_recover, cycle_initial=value_type_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
     pub(crate) fn raw_value_type(self, db: &'db dyn Db) -> Type<'db> {
         let scope = self.rhs_scope(db);
         let module = parsed_module(db, scope.file(db)).load(db);
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/aliases.md
```diff
@@ -170,3 +170,16 @@ type X[T: X] = T
...
+
+```py
+type RecursiveList[T] = T | list[RecursiveList[T]]
+
+r1: RecursiveList[int] = 1
+r2: RecursiveList[int] = [1, [1, 2, 3]]
+# error: [invalid-assignment] "Object of type `Literal["a"]` is not assignable to `RecursiveList[int]`"
+r3: RecursiveList[int] = "a"
+# TODO: this should be an error
+r4: RecursiveList[int] = [1, ["a"]]
```
AlexWaygood: A couple more suggested tests to make our behaviour here clear:

```suggestion
# error: [invalid-assignment] "Object of type `Literal["a"]` is not assignable to `RecursiveList[int]`"
r3: RecursiveList[int] = "a"
# error: [invalid-assignment]
r4: RecursiveList[int] = ["a"]
# TODO: this should be an error
r5: RecursiveList[int] = [1, ["a"]]

def _(x: RecursiveList[int]):
    if isinstance(x, list):
        # TODO: should be `list[RecursiveList[int]]
        reveal_type(x[0])  # revealed: list[int | list[Any]]
    if isinstance(x, list) and isinstance(x[0], list):
        # TODO: should be `list[RecursiveList[int]]`
        reveal_type(x[0])  # revealed: list[Any]
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -6791,7 +6791,20 @@ impl<'db> Type<'db> {
...
+                    let specialization = alias
+                        .specialization(db)
+                        .unwrap_or_else(|| generic_context.default_specialization(db, None));
+
+                    value_type.apply_specialization(db, specialization)
+                        .apply_type_mapping_impl(db, type_mapping, tcx, visitor)
+                } else {
+                    value_type
+                }
             }
```
MichaReiser, line 6769: Instead of duplicating the logic here, I'm inclined to change the signature of `value_type` and add the visitor as an argument. Not only does it reduce code duplication, it also makes callers aware that they need to think about how to handle the recursive case. We can do this now because `value_type` is no longer a salsa query.

ibraheemdev (commented):
It's interesting that `apply_type_mapping_impl` is the only method that causes a panic. I suspect we will need to generalize this to storing a lazily-applied specialization on `Type::TypeAlias`, but given that this seems to resolve the panics, it seems fine to merge the easier fix for now?

crates/ty_python_semantic/src/types.rs
dcreager: I'm not sure that passing in a visitor to `value_type` would help here. It wouldn't be correct to use that visitor when applying the specialization, because then you'd be using the same visitor for two separate type mappings. I think the visitor's internal state would get confused about which types had been visited. We would need to construct a _separate_ visitor to pass in to that call, and the only new visitor we could create is an empty one, and so that wouldn't really be different than what @mtshiba is currently doing.

I think the issue is more that we now perform the two type mappings in a different order: before, we applied the specialization first (via the call to `value_type`), then we applied the `type_mapping` passed into this call. Now we do that in the reverse order, so that we can skip the specialization step if we've already visited this type for `type_mapping`.

So a different suggestion would be to factor out a new internal helper method:

```rust
fn apply_function_specialization(self, db: &'db dyn Db, ty: Type<'db>) -> Type<'db> {
    if let Some(generic_context) = self.generic_context(db) {
        let specialization = self
            .specialization(db)
            .unwrap_or_else(|| generic_context.default_specialization(db, None));
        ty.apply_specialization(db, specialization)
    } else {
        ty
    }
}
```

Then `value_type` would become much simpler:

```rust
pub(crate) fn value_type(self, db: &'db dyn Db) -> Type<'db> {
    self.apply_function_specialization(db, self.raw_value_type(db))
}
```

and then you could call that here to avoid the duplication

PR #20971, opened 2025-10-19
Add source to testimonial

ShalokShalom:
The testimonial quotes two different sources. 
This PR adds the second source, which was missing.  (One under Tim, the second via Abbott)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ creator of [isort](https://github.com/PyCQA/isort):
 > Just switched my first project to Ruff. Only one downside so far: it's so fast I couldn't believe
 > it was working till I intentionally introduced some errors.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,8 +89,7 @@ creator of [isort](https://github.com/PyCQA/isort):
 > Just switched my first project to Ruff. Only one downside so far: it's so fast I couldn't believe
 > it was working till I intentionally introduced some errors.
 
-[**Tim Abbott**](https://github.com/astral-sh/ruff/issues/465#issuecomment-1317400028), lead
-developer of [Zulip](https://github.com/zulip/zulip):
+[**Tim_**](https://github.com/zulip/zulip/pull/23431#issuecomment-1302557034)[**Abbott**](https://github.com/astral-sh/ruff/issues/465#issuecomment-1317400028), lead developer of [Zulip](https://github.com/zulip/zulip):
```
MichaReiser: Thanks. I'm a bit confused. You say that it quotes two testimonial, but I'm only aware of one, the *This is just ridiculously fast...*

I think the proper fix here is to just change the source to https://github.com/zulip/zulip/pull/23431#issuecomment-1302557034

MichaReiser: Oh I see now. The ruff is amazing. Hmm

PR #20972, opened 2025-10-19
[`pydoclint`] Support NumPy-style comma-separated parameters (`DOC102`)

### LINKED ISSUE #20959: DOC102 false negative for NumPy-style comma-separated parameters
### Summary

[The NumPy docstring style guide says](https://numpydoc.readthedocs.io/en/stable/format.html#parameters) “When two or more input parameters have exactly the same type, shape and description, they can be combined”, but [`docstring-extraneous-parameter` (DOC102)](https://docs.astral.sh/ruff/rules/docstring-extraneous-parameter/) doesn’t support that, which causes false negatives. [Example](https://play.ruff.rs/118cf9ce-c72c-4cd8-b967-42d65e646f68):
```console
$ cat >doc102.py <<'# EOF'
def leq(x: object, y: object) -> bool:
    """Compare two objects for loose equality.

    Parameters
    ----------
    x1, x2 : object
        Objects.

    Returns
    -------
    bool
        Whether the objects are identical or equal.
    """
    return x is y or x == y
# EOF

$ ruff --isolated check doc102.py --select DOC102 --preview
All checks passed!
```

### Version

danparizher:
## Summary

Fixes #20959. Enhanced the DOC102 rule to recognize and validate multiple parameters listed together in NumPy-style docstrings using comma separation.

## Problem Analysis

The existing DOC102 rule failed to detect extraneous parameters when NumPy-style docstrings combined multiple parameters with commas (e.g., `x1, x2 : object`), causing false negatives.

## Approach

Updated the `parse_parameters_numpy` function to split comma-separated parameter names and validate each individually, added comprehensive test cases covering the original issue and edge cases, and ensured code compliance with project standards.

### DIFF
--- a/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
+++ b/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
@@ -672,9 +672,7 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
                             // Calculate the position of this specific parameter part within the line
                             // Account for leading whitespace that gets trimmed
                             let param_start_in_line = current_offset
-                                + TextSize::from(u32::from(
-                                    param_part.text_len() - param_part_trimmed.text_len(),
-                                ));
+                                + (param_part.text_len() - param_part_trimmed.text_len());
                             let param_start =
                                 line_start + indentation.text_len() + param_start_in_line;
 


crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -653,19 +653,32 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+                            // Calculate the position of this specific parameter part within the line
+                            let param_start_in_line =
+                                param_line.find(param_part_trimmed).unwrap_or(0);
+                            let param_start = line_start
+                                + indentation.text_len()
+                                + TextSize::from(u32::try_from(param_start_in_line).unwrap_or(0));
+                            let param_end = param_start
+                                + TextSize::from(
+                                    u32::try_from(param_part_trimmed.len()).unwrap_or(0),
+                                );
```
ntBre: I think this is just
```suggestion
                                + param_part_trimmed.text_len();
```

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -653,19 +653,32 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+                    for param_part in param_line.split(',') {
+                        let param_part_trimmed = param_part.trim();
+                        let param_name = param_part_trimmed.trim_start_matches('*');
+                        if is_identifier(param_name) {
+                            // Calculate the position of this specific parameter part within the line
+                            let param_start_in_line =
+                                param_line.find(param_part_trimmed).unwrap_or(0);
+                            let param_start = line_start
+                                + indentation.text_len()
+                                + TextSize::from(u32::try_from(param_start_in_line).unwrap_or(0));
```
ntBre:
```suggestion
                                + TextSize::try_from(param_start_in_line).unwrap();
```

This is one of the rare cases where plain `unwrap` is fine. Any usize within a document that we're linting (and in this case an offset within a single line) should definitely be convertible into a u32 or `TextSize`.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -653,19 +653,32 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+                    let param_line = before_colon.trim_end();
+
+                    // Split on commas to handle comma-separated parameters
+                    for param_part in param_line.split(',') {
+                        let param_part_trimmed = param_part.trim();
+                        let param_name = param_part_trimmed.trim_start_matches('*');
+                        if is_identifier(param_name) {
+                            // Calculate the position of this specific parameter part within the line
+                            let param_start_in_line =
+                                param_line.find(param_part_trimmed).unwrap_or(0);
```
ntBre: Let's just make this an `unwrap` too. We're searching for a substring in a line we just split, so it should be okay. And if we end up unwrapping to zero  somehow, the rendered range could look really bad. It might be worth a short safety comment to that effect, e.g.:

> Safety: searching for a substring we just split above

Ideally I feel like we'd just update the `current_pos` or a similar offset value as we go instead of `find`ing after the fact, but I think that's hard to do with `split`, so this seems okay.

augustelalande:
Can you add a test case for a type less docstring with comma, e.g.:
```python
def add_numbers(a, b):
    """
    Adds two numbers and returns the result.

    Parameters
    ----------
    a, b
        The numbers to add.

    Returns
    -------
    int
        The sum of the two numbers.
    """
    return a + b
```

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -661,19 +661,29 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
-                        let param_end = param_start + param.text_len();
-
-                        entries.push(ParameterEntry {
-                            name: param_name,
-                            range: TextRange::new(
-                                content_start + param_start,
-                                content_start + param_end,
-                            ),
-                        });
+                    let param_line = before_colon.trim_end();
```
MichaReiser, line 664: I suspect that the `trim_end` call here now is unnecessary, given that you also trim each parameter part.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -661,19 +661,29 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+
+                    // Split on commas to handle comma-separated parameters
+                    for param_part in param_line.split(',') {
+                        let param_part_trimmed = param_part.trim();
+                        let param_name = param_part_trimmed.trim_start_matches('*');
+                        if is_identifier(param_name) {
+                            // Calculate the position of this specific parameter part within the line
+                            // Safety: searching for a substring we just split above
+                            let param_start_in_line = param_line.find(param_part_trimmed).unwrap();
```
MichaReiser: I'd be inclined to keep track of the offset instead of finding the parameter aggain. 

```
let mut current_offset = TextSize::MIN;

for param_part in ... {
	...

	let param_start_in_line = current_offset + (param_part.text_len() - param_part_trimmed.text_len());

	current_offset = current_offset + param_part.text_len() + ','.text_len()
}
```

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -661,19 +661,29 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+                                + indentation.text_len()
+                                + TextSize::try_from(param_start_in_line).unwrap();
+                            let param_end = param_start + param_part_trimmed.text_len();
+
+                            entries.push(ParameterEntry {
+                                name: param_name,
+                                range: TextRange::new(
+                                    content_start + param_start,
+                                    content_start + param_end,
+                                ),
```
MichaReiser:
```suggestion
                                range: TextRange::at(
                                    content_start + param_start,
                                    param_part_trimmed.text_len(),
                                ),
```

MichaReiser (approved):

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
danparizher: Removing it causes incorrect column positions in the diagnostic ranges, from my testing.

Even though we `trim()` each parameter part, the offset calculation happens before trimming, so the trailing whitespace in `before_colon` affects the positions.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -661,19 +661,33 @@ fn parse_parameters_numpy(content: &str, content_start: TextSize) -> Vec<Paramet
...
+                    for param_part in param_line.split(',') {
+                        let param_part_trimmed = param_part.trim();
+                        let param_name = param_part_trimmed.trim_start_matches('*');
+                        if is_identifier(param_name) {
+                            // Calculate the position of this specific parameter part within the line
+                            // Account for leading whitespace that gets trimmed
+                            let param_start_in_line = current_offset
+                                + TextSize::from(u32::from(
+                                    param_part.text_len() - param_part_trimmed.text_len(),
+                                ));
```
MichaReiser: This seems unnecessarily complicated to go from `TextSize -> u32 -> TextSize`. Can you say more why you added that?

danparizher: Oversight on my part - Since both `param_part.text_len()` and `param_part_trimmed.text_len()` return `TextSize`, we can subtract them directly without the conversion.

PR #20974, opened 2025-10-19
[ty] implement `typing.TypeGuard`

ericmarkmartin:
## Summary

Resolve(s) astral-sh/ty#117, astral-sh/ty#1569

Implement `typing.TypeGuard`. Due to the fact that it [overrides anything previously known about the checked value](https://typing.python.org/en/latest/spec/narrowing.html#typeguard)---

> When a conditional statement includes a call to a user-defined type guard function, and that function returns true, the expression passed as the first positional argument to the type guard function should be assumed by a static type checker to take on the type specified in the TypeGuard return type, unless and until it is further narrowed within the conditional code block.

---we have to substantially rework the constraints system. In particular, we make constraints represented as a disjunctive normal form (DNF) where each term includes a regular constraint, and one or more disjuncts with a typeguard constraint. Some test cases (including some with more complex boolean logic) are added to `type_guards.md`.

## Test Plan

- update existing tests
- add new tests for more complex boolean logic with `TypeGuard`
- add new tests for `TypeGuard` variance

### DIFF
--- a/crates/ty_python_semantic/src/types/narrow.rs
+++ b/crates/ty_python_semantic/src/types/narrow.rs
@@ -285,15 +285,15 @@ impl ClassInfoConstraintFunction {
 /// For example:
 /// - `f(x) and g(x)` where f returns `TypeIs[A]` and g returns `TypeGuard[B]`
 ///   => and
-///     => `NarrowingConstraint { regular_disjunct: Some(A), typeguard_disjuncts: [] }`
-///     => `NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }`
-///   => NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }
+///   ===> `NarrowingConstraint { regular_disjunct: Some(A), typeguard_disjuncts: [] }`
+///   ===> `NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }`
+///   => `NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }`
 ///   => evaluates to `B` (`TypeGuard` clobbers any previous type information)
 ///
 /// - `f(x) or g(x)` where f returns `TypeIs[A]` and g returns `TypeGuard[B]`
 ///   => or
-///     => `NarrowingConstraint { regular_disjunct: Some(A), typeguard_disjuncts: [] }`
-///     => `NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }`
+///   ===> `NarrowingConstraint { regular_disjunct: Some(A), typeguard_disjuncts: [] }`
+///   ===> `NarrowingConstraint { regular_disjunct: None, typeguard_disjuncts: [B] }`
 ///   => `NarrowingConstraint { regular_disjunct: Some(A), typeguard_disjuncts: [B] }`
 ///   => evaluates to `(P & A) | B`, where `P` is our previously-known type
 #[derive(Hash, PartialEq, Debug, Eq, Clone, salsa::Update, get_size2::GetSize)]
@@ -302,8 +302,8 @@ pub(crate) struct NarrowingConstraint<'db> {
     /// because we can eagerly union disjunctions and eagerly intersect conjunctions.
     regular_disjunct: Option<Type<'db>>,
 
-    /// TypeGuard constraints. We can't eagerly union disjunctions because `TypeGuard` clobbers the
-    /// previously-known type; within each TypeGuard disjunct, we may eagerly intersect
+    /// `TypeGuard` constraints. We can't eagerly union disjunctions because `TypeGuard` clobbers
+    /// the previously-known type; within each `TypeGuard` disjunct, we may eagerly intersect
     /// conjunctions with a later regular narrowing.
     typeguard_disjuncts: SmallVec<[Type<'db>; 1]>,
 }
@@ -331,13 +331,13 @@ impl<'db> NarrowingConstraint<'db> {
         // Distribute AND over OR: (A1 | A2 | ...) AND (B1 | B2 | ...)
         // becomes (A1 & B1) | (A1 & B2) | ... | (A2 & B1) | ...
         //
-        // In our representation, the RHS `typeguard_disjuncts` will all clobber
-        // the LHS disjuncts when they are anded, so they'll just stay as is.
+        // In our representation, the RHS `typeguard_disjuncts` will all clobber the LHS disjuncts
+        // when they are anded, so they'll just stay as is.
         //
-        // The thing we actually need to deal with is the RHS `regular_disjunct`.
-        // It gets intersected with the LHS `regular_disjunct` to form the new
-        // `regular_disjunct`, and intersected with each LHS `typeguard_disjunct`
-        // to form new additional `typeguard_disjuncts`.
+        // The thing we actually need to deal with is the RHS `regular_disjunct`. It gets
+        // intersected with the LHS `regular_disjunct` to form the new `regular_disjunct`, and
+        // intersected with each LHS `typeguard_disjunct` to form new additional
+        // `typeguard_disjuncts`.
         let Some(other_regular_disjunct) = other.regular_disjunct else {
             return other;
         };


ericmarkmartin:
Actually I hadn't seen astral-sh/ty#690 before, but maybe this change isn't as terrible as I initially thought if it helps address that?

carljm:
Hi! Sorry this fell off my radar; it wasn't in my GH notifications because it was still in draft.

This looks pretty good to me, and TypeGuard support would be great to have; if you can get it rebased and the conformance suite changes look right, I'd happily merge it. If the constraint DNF stuff can help us resolve https://github.com/astral-sh/ty/issues/690 that's a bonus, although I kind of think we may need to eventually switch to an "every new narrowing is tracked as a new definition" approach instead, in order to also fix https://github.com/astral-sh/ty/issues/685

ericmarkmartin:
> This looks pretty good to me, and TypeGuard support would be great to have; if you can get it rebased and the conformance suite changes look right, I'd happily merge it

Rebased---fortunately not to bad conflict-wise and tests seem to pass. Not sure how to evaluate the conformance suite side but 🤞. 

> If the constraint DNF stuff can help us resolve [astral-sh/ty#690](https://github.com/astral-sh/ty/issues/690) that's a bonus

I'm not super confident about this, I'm mostly just going off of
> we just track an implicit AND of multiple narrowing constraints per definition. Since we have no way to represent an OR of narrowing constraints

from that issue

carljm:
So regarding the conformance suite, the way to evaluate that is to open up the corresponding conformance suite test file (e.g. https://github.com/python/typing/blob/main/conformance/tests/narrowing_typeguard.py is the main relevant one here). Lines with an `# E` comment are supposed to emit a diagnostic, lines without are not supposed to.

Overall this PR is clearly an improvement (every change is in the right direction), but the conformance suite does reveal a gap in the implementation that we should probably fix, when a `TypeGuard` is an instance or class method. I would guess maybe we fail to skip the `self` / `cls` argument when deciding what is the "first argument" to which narrowing applies? See lines 69, 73, 77, 89, 93, where we still wrongly emit a diagnostic with this PR.

(When fixing this, we should also add corresponding mdtests to our own test suite.)

carljm:
It looks like the same is true for our existing implementation of `TypeIs`, I just filed https://github.com/astral-sh/ty/issues/1569

carljm:
The ecosystem report looks generally fantastic! One thing I notice is that in cases like https://github.com/pydata/xarray/blob/main/xarray/namedarray/utils.py#L107 where the returned `TypeGuard` is a gradual type, this PR takes the top materialization of it, like we do for `TypeIs`. This is important for `TypeIs`, in order to get intersections to simplify and give the expected narrowing. I'm not sure we should do the same for `TypeGuard`. Because of the "replaces existing type" semantics, intersection simplification is less important for `TypeGuard`. And taking the top materialization causes us to emit diagnostics (which the author would likely consider false positives) in the xarray case, since we treat it as narrowing to `Mapping[object, object]` rather than `Mapping[Any, Any]`.

So I think we should remove the top materialization behavior from this PR and see how that impacts the ecosystem report.

carljm:
(I'm not sure why the ecosystem analyzer failed to post its comment here, but the report can be downloaded at https://github.com/astral-sh/ruff/actions/runs/19380669883/artifacts/4574376479 )

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+/// Represents a single conjunction (AND) of constraints in Disjunctive Normal
+/// Form (DNF).
+///
+/// A conjunction may contain: - A regular constraint (intersection of types) -
+/// An optional `TypeGuard` constraint that "replaces" the type rather than
+/// intersecting
+///
+/// For example, `(Conjunction { constraint: A, typeguard: Some(B) } &
+/// Conjunction { constraint: C, typeguard: Some(D)})` evlaluates to
+/// `Conjunction { constraint: C, typeguard: Some(D) }` because the type guard
```
carljm: Shouldn't this be
```suggestion
/// `Conjunction { constraint: A & C, typeguard: Some(D) }` because the type guard
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+/// in the second clobbers the first.
+#[derive(Hash, PartialEq, Debug, Eq, Clone, Copy)]
+struct Conjunction<'db> {
+    /// The intersected constraints (represented as an intersection type)
```
carljm: The type here is not necessarily an "intersection type", but it is a "type we will intersect with our previous knowledge of the type of the variable"

```suggestion
    /// The intersected constraints (represented as a type to intersect with)
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+    constraint: Type<'db>,
+    /// If any constraint in this conjunction is a `TypeGuard`, this is Some and
+    /// contains the union of all `TypeGuard` types in this conjunction
```
carljm: Why would a conjunction be a union, not an intersection? And why would we need either one, when a typeguard clobbers previous information (as mentioned in the comment above)?

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+    /// Evaluate this conjunction to a single type.
+    /// If there's a `TypeGuard` constraint, it replaces the regular constraint.
+    /// Otherwise, returns the regular constraint.
+    fn evaluate_type_constraint(self, db: &'db dyn Db) -> Type<'db> {
+        self.typeguard.map_or_else(
+            || self.constraint,
+            |typeguard_constraint| {
+                IntersectionBuilder::new(db)
+                    .add_positive(typeguard_constraint)
+                    .add_positive(self.constraint)
```
carljm: Building intersections isn't free. I think we should probably make `constraint` an `Option` (rather than defaulting it to `object`) as well, so we can build this intersection only if we actually need to.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+/// through boolean operations.
+///
+/// For example:
+/// - `f(x) and g(x)` where f returns `TypeIs[A]` and g returns `TypeGuard[B]`
+///   => `[Conjunction { constraint: A, typeguard: Some(B) }]`
+///   => evaluates to `B` (`TypeGuard` clobbers)
+///
+/// - `f(x) or g(x)` where f returns `TypeIs[A]` and g returns `TypeGuard[B]`
+///   => `[Conjunction { constraint: A, typeguard: None }, Conjunction { constraint: object, typeguard: Some(B) }]`
+///   => evaluates to `A | B`
```
carljm: I don't think this is quite right, as discussed below; we need to say something more like
```suggestion
///   => evaluates to `(P & A) | B` where `P` is our previously-known type
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+#[derive(Hash, PartialEq, Debug, Eq, Clone)]
+struct NarrowingConstraint<'db> {
+    /// Disjunctions of conjunctions (DNF)
```
carljm:
```suggestion
    /// Disjunction of conjunctions (DNF)
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+    disjuncts: SmallVec<[Conjunction<'db>; 4]>,
```
carljm: I sort of suspect that 4 is too large here, and either 1 or 2 will be better? We have to clone these quite a bit.

Did you do any experimentation to settle on 4? Can you try 1 and 2 and see if it makes any noticeable difference on benchmarks?

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+    ///
+    /// Forgets whether each constraint originated from a `TypeGuard` or not
+    fn evaluate_type_constraint(self, db: &'db dyn Db) -> Type<'db> {
+        UnionType::from_elements(
+            db,
+            self.disjuncts
+                .into_iter()
+                .map(|disjunct| Conjunction::evaluate_type_constraint(disjunct, db)),
+        )
+    }
```
carljm, line 380: I don't think this is right? As mentioned above, given a regular constraint of `A` OR a typeguard constraint of `B`, and a previously-known type of `P`, the answer we should narrow to is `(P & A) | B`, but the way this is currently implemented will give `P & (A | B)` instead. In the typeguard branch of the union we need to ignore the prior type.

Or, translating the same thing to testable Python code:

```py
from typing import TypeGuard, reveal_type

class P:
    pass

class A:
    pass

class B:
    pass

def is_b(val: object) -> TypeGuard[B]:
    return isinstance(val, B)

def _(x: P):
    if isinstance(x, A) or is_b(x):
        reveal_type(x)  # currently reveals `(P & A) | (P & B)`, should reveal `(P & A) | B`
```

I think this will require passing the `NarrowingConstraint` all the way out to the type-inference code (i.e. `InternalConstraints` should replace `NarrowingConstraints`), because there is no way to correctly implement this method without also taking in the prior type.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+            .insert(place, NarrowingConstraint::regular(ty));
+    }
+
+    /// Convert internal constraints to public constraints by evaluating each DNF constraint to a Type
+    fn evaluate_type_constraints(self, db: &'db dyn Db) -> NarrowingConstraints<'db> {
+        self.constraints
+            .into_iter()
+            .map(|(place, constraint)| (place, constraint.evaluate_type_constraint(db)))
+            .collect()
+    }
```
carljm: I don't think this method can exist in a correct implementation, there is no correct way to do this translation without access to the prior type for each place.

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_forms.md
```diff
@@ -19,7 +19,9 @@ def f(*args: Unpack[Ts]) -> tuple[Unpack[Ts]]:
     reveal_type(Alias)  # revealed: @Todo(Support for `typing.TypeAlias`)
     return args
 
-def g() -> TypeGuard[int]: ...
+def g() -> TypeGuard[int]:
+    return True
+
```
carljm: I think we can just remove this, it's no longer an "unsupported special form"

crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
```diff
@@ -17,16 +17,17 @@ def _(
...
     f: TypeIs,  # error: [invalid-type-form]
 ):
-    # TODO: Should be `TypeGuard[str]`
-    reveal_type(a)  # revealed: @Todo(`TypeGuard[]` special form)
+    reveal_type(a)  # revealed: TypeGuard[str]
     reveal_type(b)  # revealed: TypeIs[str | int]
-    # TODO: Should be `TypeGuard[complex & ~int & ~float]`
-    reveal_type(c)  # revealed: @Todo(`TypeGuard[]` special form)
+    # not `TypeGuard[complex & ~int & ~float]`: `complex` in argument position
+    # means `complex & int & float` semantically so `Intersection[complex,
```
carljm: I think you mean 
```suggestion
    # means `complex | int | float` semantically so `Intersection[complex,
```

But honestly I would just use a different type above and skip all this complex/int/float nonsense, it's not germane

crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
```diff
@@ -215,23 +214,19 @@ class C(Generic[T]):
     v: T
 
 def _(a: tuple[Foo, Bar] | tuple[Bar, Foo], c: C[Any]):
-    # TODO: Should be `TypeGuard[Foo @ a[1]]`
-    if reveal_type(guard_foo(a[1])):  # revealed: @Todo(`TypeGuard[]` special form)
+    if reveal_type(guard_foo(a[1])):  # revealed: TypeGuard[Foo @ a[1]]
         # TODO: Should be `tuple[Bar, Foo]`
```
carljm: Orthogonal to this PR, but I'm not sure this should even be a TODO. No other type checker draws this conclusion, and I think the odd semantics of `TypeGuard` make it unclear whether it's even valid to do so.
```suggestion
```

crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
```diff
@@ -215,23 +214,19 @@ class C(Generic[T]):
...
         reveal_type(a)  # revealed: tuple[Foo, Bar] | tuple[Bar, Foo]
-        # TODO: Should be `Foo`
-        reveal_type(a[1])  # revealed: Bar | Foo
+        reveal_type(a[1])  # revealed: Foo
 
     if reveal_type(is_bar(a[0])):  # revealed: TypeIs[Bar @ a[0]]
         # TODO: Should be `tuple[Bar, Bar & Foo]`
```
carljm: Also orthogonal, but this is wrong IMO, if anything we could reveal `tuple[Foo & Bar, Bar] | tuple[Bar, Foo]`, but no other type checker does this either. It seems of questionable utility.
```suggestion
```

crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
```diff
@@ -215,23 +214,19 @@ class C(Generic[T]):
...
         reveal_type(a)  # revealed: tuple[Foo, Bar] | tuple[Bar, Foo]
         reveal_type(a[0])  # revealed: Bar
 
-    # TODO: Should be `TypeGuard[Foo @ c.v]`
-    if reveal_type(guard_foo(c.v)):  # revealed: @Todo(`TypeGuard[]` special form)
+    if reveal_type(guard_foo(c.v)):  # revealed: TypeGuard[Foo @ c.v]
         reveal_type(c)  # revealed: C[Any]
-        # TODO: Should be `Foo`
-        reveal_type(c.v)  # revealed: Any
+        reveal_type(c.v)  # revealed: Any & Foo
```
carljm: Shouldn't this be just `Foo` per the semantics of `TypeGuard`?

crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
```diff
@@ -350,20 +344,72 @@ def is_bar(a: object) -> TypeIs[Bar]:
...
         reveal_type(a)  # revealed: Foo | Bar
     else:
-        reveal_type(a)  # revealed: Foo | Bar
+        reveal_type(a)  # revealed: Foo
 
 def narrowed_type_must_be_exact(a: object, b: Baz):
     if guard_foo(b):
-        # TODO: Should be `Foo`
-        reveal_type(b)  # revealed: Baz
+        reveal_type(b)  # revealed: Baz & Foo
```
carljm: Shouldn't this be just `Foo`?

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -1280,10 +1280,27 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
...
+                        diagnostic::add_type_expression_reference_link(diag);
+                    }
+
+                    Type::unknown()
+                }
+                _ => TypeGuardType::unbound(
+                    self.db(),
+                    // Similar to TypeIs, use top materialization
+                    self.infer_type_expression(arguments_slice)
+                        .top_materialization(self.db()),
```
carljm: As mentioned in another comment, I think for `TypeGuard`, given its "clobber" semantics, we maybe shouldn't do this? Or at least I'm curious what the impact on the ecosystem report would be. We do see some false positives in the ecosystem from doing this.

crates/ty_python_semantic/src/types/infer/builder.rs
carljm, line 9019: The bug with method guards would be fixed in here, I think.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+                entry
+                    .get_mut()
+                    .disjuncts
+                    .extend(from_constraint.disjuncts.clone());
             }
             Entry::Vacant(entry) => {
-                entry.insert(Type::object());
+                // Place only appears in `from`, not in `into`.
+                // Widen to object since the other branch doesn't constrain it.
+                entry.insert(NarrowingConstraint::regular(Type::object()));
```
carljm: I wonder why we don't just do nothing here? We don't want to constrain, and there's already no constraint in `into`, so let's just... leave it that way. Seems to pass tests.
```suggestion
                // Place only appears in `from`, not in `into`. No constraint needed.
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -269,48 +271,244 @@ impl ClassInfoConstraintFunction {
...
+    _db: &'db dyn Db,
 ) {
-    for (key, value) in from {
-        match into.entry(*key) {
+    // For places that appear in `into` but not in `from`, widen to object
+    for (key, value) in &mut into.constraints {
+        if !from.constraints.contains_key(key) {
+            *value = NarrowingConstraint::regular(Type::object());
+        }
+    }
```
carljm: This seems better, and passes tests, not sure why we didn't do it previously:

```suggestion
    // For places that appear in `into` but not in `from`, do not constrain
    into.constraints
        .retain(|key, _| from.constraints.contains_key(key));

```

crates/ty_python_semantic/src/types.rs
```diff
@@ -12328,6 +12381,73 @@ impl<'db> VarianceInferable<'db> for TypeIsType<'db> {
...
+        Type::TypeGuard(Self::new(db, ty, self.place_info(db)))
+    }
+
+    pub(crate) fn is_bound(self, db: &'db dyn Db) -> bool {
+        self.place_info(db).is_some()
+    }
+}
+
+impl<'db> VarianceInferable<'db> for TypeGuardType<'db> {
+    // TODO: comment
```
carljm: Let's remove this or flesh it out

crates/ty_python_semantic/src/types/narrow.rs
ericmarkmartin: I didn't think so...

```py
class A: ...
class B: ...
class C: ...
class D: ...

def is_a(x: object) -> TypeIs[A]: ...
def guard_b(x: object) -> TypeGuard[B]: ...
def is_c(x: object) -> TypeIs[C]: ...
def guard_d(x: object) -> TypeGuard[D]: ...
```

Now I think `Conjunction { constraint: A, typeguard: Some(B) }` is produced by `guard_b(x) and is_a(x)` (if it were the other way around, it'd just be `Conjunction { constraint: None, tyepguard: Some(B) }` (assuming for a second we already have optionality of `constraint`)). 

Likewise `Conjunction { constraint: C, typeguard: Some(D)})` is produced by `guard_d(x) and is_c(x)`. If you put those together, i.e. `guard_b(x) and is_a(x) and guard_d(x) and is_c(x)`, I think the `guard_d` should wipe out the `guard_b` and `is_a` and leave you with just the `guard_d` and the `is_c`.

crates/ty_python_semantic/src/types/narrow.rs
ericmarkmartin: Yeah I think this was just really poor wording on my end

crates/ty_python_semantic/src/types/narrow.rs
carljm: Oh yeah, I think you're right! Sorry, I think I was mixing up AND and OR somewhere in there.

carljm:
Hi @ericmarkmartin -- thanks again for this contribution! Do you have any timeframe on getting back to this PR? Would you like someone else to pick it up and push it forward? Would love to get this feature in!

ericmarkmartin:
Hey @carljm,

Sorry this fell off my radar a bit. I'd be happy to get back into this if it's still wanted!

carljm:
Definitely still wanted, if you're up for picking it back up I'll try to be speedy on reviews!

ericmarkmartin:
This PR should also resolve https://github.com/astral-sh/ty/issues/1569 now, updated description as such

MichaReiser:
Thank you for your continuous work on this. Almost the entire team is out this week. It may take a few days before someone finds time to review your PR. Happy holidays.

carljm:
Conformance tests look great! I filed https://github.com/astral-sh/ty/issues/2267, which I think captures what remains (after this PR) for us to fully pass the `TypeIs` and `TypeGuard` conformance tests.

Ecosystem also looks great! Mostly it's removing false positives due to our new understanding of `TypeGuard`.

There are a lot of new diagnostics on scipy -- I spot-checked these and they look like true positives (at least given the specified behavior of `TypeGuard`.) Scipy has a lot of functions that take in an object of unknown type, and then do something like `if not isscalar(obj) or (obj < 1200)`, where `isscalar` is a `TypeGuard`, and the union it narrows to includes types that can't be compared `< 1200`. Erroring on this looks correct, and other type checkers also error in similar usages.

The new diagnostics on apprise are a bit weird, and I haven't fully tracked down where a `TypeGuard` is even involved. But the underlying issue there looks like https://github.com/astral-sh/ty/issues/1714, so I'm not concerned that it's surfacing a problem with this PR.

crates/ty_python_semantic/src/types.rs
```diff
@@ -2809,15 +2829,29 @@ impl<'db> Type<'db> {
...
+                    db,
+                    right.return_type(db),
+                    inferable,
+                    relation,
+                    relation_visitor,
+                    disjointness_visitor,
+                )
+            }
+
+            // `TypeIs[T]` and `TypeGuard[T]` is a subtype of `bool`.
```
```suggestion
            // `TypeIs[T]` and `TypeGuard[T]` are subtypes of `bool`.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -4324,6 +4367,7 @@ impl<'db> Type<'db> {
             }
 
             Type::TypeIs(type_is) => type_is.is_bound(db),
+            Type::TypeGuard(type_is) => type_is.is_bound(db),
```
```suggestion
            Type::TypeGuard(type_guard) => type_guard.is_bound(db),
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -8038,6 +8091,9 @@ impl<'db> Type<'db> {
             // TODO(jelle): Materialize should be handled differently, since TypeIs is invariant
             Type::TypeIs(type_is) => type_is.with_type(db, type_is.return_type(db).apply_type_mapping(db, type_mapping, tcx)),
 
+            // TODO: check variance
```
carljm: I don't think we need this TODO here, unless you're aware of some specific reason it's needed?

Unlike `TypeIs`, `TypeGuard` is covariant, so it should be correct (even for a "materialize" mapping) to just propagate the mapping to the return type. The TODO above is because `TypeIs` is invariant, so it's not correct to just propagate a materialization through it.

```suggestion
```

carljm (commented):
Still looking at the core implementation in `narrow.rs`, but submitting a few other comments... looking really good so far!

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -277,48 +276,221 @@ impl ClassInfoConstraintFunction {
...
+    fn evaluate_constraint_type(self, db: &'db dyn Db) -> Type<'db> {
+        match self.refinement {
+            Some(refinement) => IntersectionBuilder::new(db)
+                .add_positive(self.typeguard)
+                .add_positive(refinement)
+                .build(),
+            None => self.typeguard,
+        }
+    }
+}
```
carljm: Did you consider eliminating this struct, making `typeguard_disjuncts` below a `SmallVec` of `Type<'db>'` instead, and eagerly intersecting refinements into the type instead of doing it lazily? Locally that seems to simplify things a bit, and passes all tests. I'm pretty sure it's semantically equivalent, because there wasn't any place where we would "break apart" a `TypeGuardConstraint` and consider its refinement separately from its base typeguard type; we would always keep them together (possibly intersecting more things into `refinement`) and just end up intersecting them anyway.

I'll try pushing a commit that does this and see how it impacts performance and ecosystem.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -277,48 +276,186 @@ impl ClassInfoConstraintFunction {
...
+        // Distribute AND over OR: (A1 | A2 | ...) AND (B1 | B2 | ...)
+        // becomes (A1 & B1) | (A1 & B2) | ... | (A2 & B1) | ...
+        //
+        // In our representation, the RHS `typeguard_disjuncts` will all clobber
+        // the LHS disjuncts when they are anded, so they'll just stay as is.
+        //
+        // The thing we actually need to deal with is the RHS `regular_disjunct`.
+        // It gets anded onto the LHS `regular_disjunct` to form the new
+        // `regular_disjunct`, and anded onto each LHS `typeguard_disjunct` (via
+        // the refinement) to form new additional `typeguard_disjuncts`.
```
carljm:
```suggestion
        // The thing we actually need to deal with is the RHS `regular_disjunct`.
        // It gets intersected with the LHS `regular_disjunct` to form the new
        // `regular_disjunct`, and intersected with each LHS `typeguard_disjunct`
        // to form new additional `typeguard_disjuncts`.
```

crates/ty_python_semantic/src/types/narrow.rs
carljm: Went ahead and merged with this simplification. If you think it's wrong, let me know (ideally with a code example we now fail) and I'll be happy to make a follow-up PR to restore this.

carljm:
Merged -- thank you for this!!

PR #20985, opened 2025-10-20
[ty] Refactor: Use `let`-chains in a few places

sharkdp:
## Summary

Just ran a Claude prompt I used on one of my own projects (limited to `ty_python_semantic`). I didn't add all of the proposed changes and only selected those that were unlikely to conflict with any open PRs, but those look like reasonable improvements to me.

## Test Plan

Pure refactor

### DIFF
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -616,10 +616,10 @@ impl<'db> FunctionLiteral<'db> {
         // We only include an implementation (i.e. a definition not decorated with `@overload`) if
         // it's the only definition.
         let (overloads, implementation) = self.overloads_and_implementation(db);
-        if let Some(implementation) = implementation {
-            if overloads.is_empty() {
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/types/function.rs
```diff
@@ -616,10 +616,10 @@ impl<'db> FunctionLiteral<'db> {
         // We only include an implementation (i.e. a definition not decorated with `@overload`) if
         // it's the only definition.
         let (overloads, implementation) = self.overloads_and_implementation(db);
-        if let Some(implementation) = implementation {
-            if overloads.is_empty() {
-                return CallableSignature::single(implementation.signature(db));
-            }
+        if let Some(implementation) = implementation
+            && overloads.is_empty()
```
MichaReiser, line 620: It will take me a while to get used to this syntax. My brain immediately raises a syntax error after `implementation` that there's a missing `{` 😅

MichaReiser (approved):

AlexWaygood (approved):

PR #19201, opened 2025-07-08
[ty] Use diagnostic rendering for semantic token tests

MichaReiser:
## Summary

This PR changes the test infrastructure for semantic tokens to use the diagnostic rendering over printing the tokens with their text range.

The advantage of this approach is that I don't need a hex viewer to verify that the byte offsets are correct. 

However, it does come at the cost of much larger snapshots. I still think it's worth it considering that it improves readability of tests but it's certainly a trade off.

## Test Plan

Updated tests

PR #20986, opened 2025-10-20
[ty] Fix inconsistent highlighting of self

MichaReiser:
## Summary

Fix the inconsistent highlighting of `self`.

The root cause of the inconsistent highlighting was that the semantic token visitor failed to classify `self.x` (or any other non-name expression) in an annotated assignment. VS Code then applies its default highlighting rules (for any unclassified token). 

The fix is simple, remove the custom handling for `StmtAnnAssign` and override `visit_annotation`, the `SourceOrderVisitor` then does the traversing as expected.

We'd ideally classifiy `self` as `SelfParameter` but that's separate from this bug fix. 

Fixes https://github.com/astral-sh/ty/issues/1126

## Test Plan

Added test, updated existing tests, and...

Screenshot 2025-10-20 at 11 24 39

### DIFF
--- a/crates/ty_ide/src/semantic_tokens.rs
+++ b/crates/ty_ide/src/semantic_tokens.rs
@@ -467,13 +467,6 @@ impl<'db> SemanticTokenVisitor<'db> {
         self.classify_from_type_and_name_str(ty, local_name.id.as_str())
     }
 
-    fn visit_type_annotation(&mut self, annotation: &ast::Expr) {
-        let prev_in_type_annotation = self.in_type_annotation;
-        self.in_type_annotation = true;
-        self.visit_expr(annotation);
-        self.in_type_annotation = prev_in_type_annotation;
-    }
-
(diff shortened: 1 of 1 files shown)


crates/ty_ide/src/semantic_tokens.rs
```diff
@@ -2222,4 +2213,46 @@ finally:
...
+        "1" @ 96..97: Number
+        "self" @ 106..110: Parameter
+        "non_annotated" @ 111..124: Variable
+        "1" @ 127..128: Number
+        "self" @ 137..141: Parameter
+        "x" @ 142..143: Variable
+        "test" @ 144..148: Variable
+        "self" @ 159..163: Parameter
+        "x" @ 164..165: Variable
+        "#);
```
AlexWaygood, line 2256: I still sort-of wish we'd landed https://github.com/astral-sh/ruff/pull/19201 :(

PR #20987, opened 2025-10-20
[ty] Fix auto import for files with `from __future__` import

MichaReiser:
## Summary

Ensure that new imports are added after any pre-existing `from __future__` import.

Fixes https://github.com/astral-sh/ty/issues/1382

## Test Plan

Added test


sharkdp (approved):
Thank you!

PR #20566, opened 2025-09-25
[ty] WIP: handle recursive type inference properly

mtshiba:
## Summary

Derived from #17371

Fixes astral-sh/ty#256

Properly handles any kind of recursive inference and prevents panics.
The discussion in https://github.com/astral-sh/ruff/pull/17371#discussion_r2356324183 revealed that additional changes to the salsa API are required to complete this PR, and I will update this PR as soon as those changes are made.

The technique we used here to converge fixed-point iterations was roughly described in #17371, but will be properly documented in this PR.

PR #20988, opened 2025-10-20
[ty] Fall back to `Divergent` for deeply nested specializations

sharkdp:
## Summary

Fall back to `C[Divergent]` if we are trying to specialize `C[T]` with a type that itself already contains deeply nested specialized generic classes. This is a way to prevent infinite recursion for cases like `self.x = [self.x]` where type inference for the implicit instance attribute would not converge.

closes https://github.com/astral-sh/ty/issues/1383
closes https://github.com/astral-sh/ty/issues/837

## Test Plan

Regression tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/visitor.rs
+++ b/crates/ty_python_semantic/src/types/visitor.rs
@@ -372,7 +372,7 @@ pub(super) fn exceeds_max_specialization_depth(db: &dyn Db, ty: Type<'_>) -> boo
     //
     // reveal_type(x)  # Unknown | Literal[1] | list[Divergent]
     // ```
-    const MAX_SPECIALIZATION_DEPTH: usize = 14;
+    const MAX_SPECIALIZATION_DEPTH: usize = 10;
 
     specialization_depth(db, ty) > MAX_SPECIALIZATION_DEPTH
 }
--- a/crates/ty_python_semantic/resources/mdtest/attributes.md
+++ b/crates/ty_python_semantic/resources/mdtest/attributes.md
@@ -2624,10 +2624,10 @@ def duplicate(x: T) -> tuple[T, T]:
     return (x, x)
 
 class E:
-    def f(self: Self):
+    def f(self: "E"):
         self.x = duplicate(self.x)
 
-reveal_type(C().x)  # revealed: Unknown | tuple[Divergent, Divergent]
+reveal_type(E().x)  # revealed: Unknown | tuple[Divergent, Divergent]
 ```
 
 And it also works for homogeneous tuples:


crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1607,10 +1607,34 @@ impl<'db> ClassLiteral<'db> {
...
+        // example:
+        //
+        // ```py
+        // x = 1
+        // while random_bool():
+        //     x = [x]
+        //
+        // reveal_type(x)  # Unknown | Literal[1] | list[Divergent]
+        // ```
+        const MAX_SPECIALIZATION_DEPTH: usize = 10;
```
sharkdp: This still needs to be fine-tuned (using ecosystem impact and benchmarks).

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1607,10 +1607,34 @@ impl<'db> ClassLiteral<'db> {
...
             Some(generic_context) => {
-                let specialization = f(generic_context);
+                let mut specialization = f(generic_context);
+
+                for (idx, ty) in specialization.types(db).iter().enumerate() {
+                    if specialization_depth(db, *ty) > MAX_SPECIALIZATION_DEPTH {
+                        specialization = specialization.with_replaced_type(
+                            db,
+                            idx,
+                            Type::divergent(self.body_scope(db)),
```
sharkdp: I haven't understood the `scope` parameter of `Type::Divergent`. When we see an infinitely wrapped `list[list[list[…]]]` in a method of class `C`, is it still correct to set the scope parameter to the body scope of `list` (not of `C` or the method) here?

mtshiba: Your use of `Divergent` in this PR seems to differ slightly from the original intended use. The intended use is to make functions with the potential for divergent type inference tracked, and to set the initial cycle value to `Divergent` instead of `Never`. If the resulting type contains `Divergent`, replace that type with `Divergent` itself to converge the inference.

https://github.com/astral-sh/ruff/pull/17371/files#diff-20b910c6e20faa962bb1642e111db1cbad8e66ace089bdd966ac9d7f9fa99ff2R649-R665

When I first introduced `Divergent` in #20312, I included `scope` to prevent `Divergent` from propagating to other types of inference. This was based on the use case in #17371.

https://github.com/astral-sh/ruff/pull/17371/commits/e3b896df046afbea1674c302617f44ac2ea9cda8#diff-fdebb876f3c3f16730728018e4b1223f844d9493cd7caa7ee94e4a7a7b02184f

`call_divergent` calls `divergent`, but it does not diverge itself. Knowing where the divergence occurs can help prevent excessive type replacement.

#20333 and #20359 may also be helpful in understanding `Divergent`.

sharkdp: I currently don't see how that approach scales. We would need to add dedicated "turn this type into `Divergent` if it contains `Divergent`" logic everywhere. We can do this for tuples, but we also need to do it for sets, lists, and dicts. And it doesn't really stop with those syntactical constructs. What about something like
```py
from typing import TypeVar, Literal

T = TypeVar("T")

def make_tuple(a: T) -> tuple[T, Literal[1]]:
    return (a, 1)

class D:
    def f(self, other: "D"):
        self.x = make_tuple(other.x)

reveal_type(D().x)
```

Do we also specialize all call expressions to detect nested `Divergent` types? At that point, we could probably add it to `infer_expression_type` (and in fact, that might work, but done naively, we would get `Divergent` instead of `tuple[Divergent, …]` etc).

What I'm proposing here is certainly not very elegant, I agree. But it does solve all of these cases (and more) without any special casing to certain syntactical forms.

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
...
+    // ```py
+    // x = 1
+    // while random_bool():
+    //     x = [x]
+    //
+    // reveal_type(x)  # Unknown | Literal[1] | list[Divergent]
+    // ```
+    const MAX_SPECIALIZATION_DEPTH: usize = 10;
+
+    specialization_depth(db, ty) > MAX_SPECIALIZATION_DEPTH
```
MichaReiser, line 377: Would that mean that we also fall back to `DIVERGENT` for cases where a user combines complex types, but without any fixpoint involved?

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
     visitor.visit_type(db, ty);
     visitor.found_matching_type.get()
 }
+
+/// Returns the maximum number of layers of generic specializations for a given type.
+///
+/// For example, `int` has a depth of `0`, `list[int]` has a depth of `1`, and `list[set[int]]`
+/// has a depth of `2`. A set-theoretic type like `list[int] | list[list[int]]` has a maximum
+/// depth of `2`.
```
MichaReiser, line 308: This is a very helpful comment.

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
...
+                        .insert(non_atomic_type, Some(self.max_depth.get()));
+                }
+            }
+        }
+    }
+
+    let visitor = SpecializationDepthVisitor {
+        seen_types: RefCell::new(FxHashMap::default()),
+        max_depth: Cell::new(0),
+    };
```
AlexWaygood: nit: derive `Default`? then this could be

```suggestion
    let visitor = SpecializationDepthVisitor::default();
```

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
...
+        max_depth: Cell<usize>,
+    }
+
+    impl<'db> TypeVisitor<'db> for SpecializationDepthVisitor<'db> {
+        fn should_visit_lazy_type_attributes(&self) -> bool {
+            false
+        }
+
+        fn visit_type(&self, db: &'db dyn Db, ty: Type<'db>) {
+            match TypeKind::from(ty) {
```
AlexWaygood, line 322: can you optimize and return early here, avoiding the `match` entirely, if `self.max_depth == usize::MAX`?

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
...
+    specialization_depth(db, ty) > MAX_SPECIALIZATION_DEPTH
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{db::tests::setup_db, types::KnownClass};
+
+    #[test]
+    fn test_generics_layering_depth() {
```
AlexWaygood, line 386: I'd be tempted to define some helpers to make this test more readable, e.g.

```diff
diff --git a/crates/ty_python_semantic/src/types/visitor.rs b/crates/ty_python_semantic/src/types/visitor.rs
index d3f848dbbd..9644c932c1 100644
--- a/crates/ty_python_semantic/src/types/visitor.rs
+++ b/crates/ty_python_semantic/src/types/visitor.rs
@@ -388,62 +388,45 @@ mod tests {
     fn test_generics_layering_depth() {
         let db = setup_db();
 
-        let list_of_int =
-            KnownClass::List.to_specialized_instance(&db, [KnownClass::Int.to_instance(&db)]);
+        let int = || KnownClass::Int.to_instance(&db);
+        let list = |element| KnownClass::List.to_specialized_instance(&db, [element]);
+        let dict = |key, value| KnownClass::Dict.to_specialized_instance(&db, [key, value]);
+        let set = |element| KnownClass::Set.to_specialized_instance(&db, [element]);
+        let str = || KnownClass::Str.to_instance(&db);
+        let bytes = || KnownClass::Bytes.to_instance(&db);
+        let union = |elements| UnionType::from_elements(&db, elements);
+
+        let list_of_int = list(int());
         assert_eq!(specialization_depth(&db, list_of_int), 1);
 
-        let list_of_list_of_int = KnownClass::List.to_specialized_instance(&db, [list_of_int]);
+        let list_of_list_of_int = list(list_of_int);
         assert_eq!(specialization_depth(&db, list_of_list_of_int), 2);
 
-        let list_of_list_of_list_of_int =
-            KnownClass::List.to_specialized_instance(&db, [list_of_list_of_int]);
+        let list_of_list_of_list_of_int = list(list_of_list_of_int);
         assert_eq!(specialization_depth(&db, list_of_list_of_list_of_int), 3);
 
-        let set_of_dict_of_str_and_list_of_int = KnownClass::Set.to_specialized_instance(
-            &db,
-            [KnownClass::Dict
-                .to_specialized_instance(&db, [KnownClass::Str.to_instance(&db), list_of_int])],
-        );
+        let set_of_dict_of_str_and_list_of_int = set(dict(str, list_of_int));
         assert_eq!(
             specialization_depth(&db, set_of_dict_of_str_and_list_of_int),
             3
         );
 
-        let union_type_1 =
-            UnionType::from_elements(&db, [list_of_list_of_list_of_int, list_of_list_of_int]);
+        let union_type_1 = union([list_of_list_of_list_of_int, list_of_list_of_int]);
         assert_eq!(specialization_depth(&db, union_type_1), 3);
 
-        let union_type_2 =
-            UnionType::from_elements(&db, [list_of_list_of_int, list_of_list_of_list_of_int]);
+        let union_type_2 = union([list_of_list_of_int, list_of_list_of_list_of_int]);
         assert_eq!(specialization_depth(&db, union_type_2), 3);
 
-        let tuple_of_tuple_of_int = Type::heterogeneous_tuple(
-            &db,
-            [Type::heterogeneous_tuple(
-                &db,
-                [KnownClass::Int.to_instance(&db)],
-            )],
-        );
+        let tuple_of_tuple_of_int =
+            Type::heterogeneous_tuple(&db, [Type::heterogeneous_tuple(&db, [int()])]);
         assert_eq!(specialization_depth(&db, tuple_of_tuple_of_int), 2);
 
-        let tuple_of_list_of_int_and_str = KnownClass::Tuple
-            .to_specialized_instance(&db, [list_of_int, KnownClass::Str.to_instance(&db)]);
+        let tuple_of_list_of_int_and_str =
+            KnownClass::Tuple.to_specialized_instance(&db, [list_of_int, str()]);
         assert_eq!(specialization_depth(&db, tuple_of_list_of_int_and_str), 1);
 
-        let list_of_union_of_lists = KnownClass::List.to_specialized_instance(
-            &db,
-            [UnionType::from_elements(
-                &db,
-                [
-                    KnownClass::List
-                        .to_specialized_instance(&db, [KnownClass::Int.to_instance(&db)]),
-                    KnownClass::List
-                        .to_specialized_instance(&db, [KnownClass::Str.to_instance(&db)]),
-                    KnownClass::List
-                        .to_specialized_instance(&db, [KnownClass::Bytes.to_instance(&db)]),
-                ],
-            )],
-        );
+        let list_of_union_of_lists = list(union([list(int()), list(str()), list(bytes())]));
+
         assert_eq!(specialization_depth(&db, list_of_union_of_lists), 2);
```

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,150 @@ pub(super) fn any_over_type<'db>(
...
+            &db,
+            [Type::heterogeneous_tuple(
+                &db,
+                [KnownClass::Int.to_instance(&db)],
+            )],
+        );
+        assert_eq!(specialization_depth(&db, tuple_of_tuple_of_int), 2);
+
+        let tuple_of_list_of_int_and_str = KnownClass::Tuple
+            .to_specialized_instance(&db, [list_of_int, KnownClass::Str.to_instance(&db)]);
```
AlexWaygood: Should this be `Type::homogeneous_tuple()` rather than `KnownClass::Tuple.to_instance()`?

sharkdp: Yes. I actually fell into this trap where I thought I could just run `KnownClass::Tuple.to_specialized_instance(…)` to get a `tuple` type. I don't know *what* it generates, but it's not a tuple, apparently. Maybe we should prevent other developers from doing that by adding some kind of check?

AlexWaygood: Yes, some kind of debug assertion in `.to_instance()` that checks you're not trying to create a tuple type using that method would probably be great

crates/ty_python_semantic/src/types/visitor.rs
sharkdp: That is the "not so elegant" part of this PR, exactly. But 10 already seems to be deep enough not to cause any ecosystem changes here. Well, that's a bit hard to tell, of course, because `Divergent` acts like a dynamic type. But I *did* see changes when the limit was too low (or rather: when I was accidentally computing depths that were too large). Because introducing dynamic types all over the place typically leads to a reduction of diagnostics.

crates/ty_python_semantic/src/types/class.rs
mtshiba: I don't actually think it's necessary to apply this logic everywhere. That's work in progress in #20566.

The only place where this logic needs to be applied is within tracked functions that may diverge (e.g. `implicit_attribute_inner`, `infer_expression_type_impl`). That means that a `Divergent`-type suppression needs to be performed by the end of a cycle, and it would be most natural to do this at the end of the tracked function. Other places, such as `infer_tuple_expression`, don't actually need this logic. In fact, it's been removed in https://github.com/astral-sh/ruff/pull/20566/files#diff-5ade3020ce233d81ee50c1ec2debd673afc8ba08ede8ec664ca62e62aefbb2f7L5273-L5279, https://github.com/astral-sh/ruff/pull/20566/files#diff-411da1a5c9e0560d6b43091ca27247357a1e94a9077f9adb2cd6a23f495578b6L24-L28.

cc (as author of #20333, etc.): @carljm

crates/ty_python_semantic/src/types/visitor.rs
sharkdp: We probably could, but the `self.max_depth == usize::MAX` cases are hopefully *not* what we need to optimize for. It's not important that this is extremely fast for (too) deeply nested generic types. But it is important that this is fast for literally everything else.

AlexWaygood: Thanks, that's helpful. It might be good to put that in a comment (that this "optimization" could actually be a pessimisation) -- I'm used to seeing that kind of "return early" clause in ad-hoc AST visitors like this that we write for Ruff rules, and was surprised not to see it here

crates/ty_python_semantic/src/types/visitor.rs
AlexWaygood: For posterity: we did this in https://github.com/astral-sh/ruff/pull/21027

crates/ty_python_semantic/src/types/class.rs
sharkdp: I had completely missed https://github.com/astral-sh/ruff/pull/20566 — thank you for the reference. It looks like that still requires adding divergent-handling in lots of places, but I'm certainly not opposed to that. I would very much like there to be a more principled solution.

My problem is that I want something that fixes these panics *now* (to finally merge the type-of-`self` PR). I rolled back most of the changes here to leave the `CycleRecovery` setup intact and to be less invasive. I am proposing that we merge this PR to solve the immediate problem, and I'll be happy to remove all of this once we have a better solution (that still solves all of the test cases that are introduced in this PR).

crates/ty_python_semantic/src/types/visitor.rs
sharkdp: It looks like we'll soon remove all of this code anyway, so I'll leave it as is for now.

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -295,3 +300,148 @@ pub(super) fn any_over_type<'db>(
...
+    // example:
+    //
+    // ```py
+    // x = 1
+    // while random_bool():
+    //     x = [x]
+    //
+    // reveal_type(x)  # Unknown | Literal[1] | list[Divergent]
+    // ```
+    const MAX_SPECIALIZATION_DEPTH: usize = 10;
```
AlexWaygood, line 375: 10 does feel quite conservative here. My instinct would be to go with something a _bit_ larger (32?) if it's possible. But it also probably doesn't matter that much if we're planning to get rid of this soon anyway. This is probably fine for a stopgap measure.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1264,6 +1264,27 @@ impl<'db> Specialization<'db> {
...
     }
+
+    /// Returns a copy of this specialization with the type at a given index replaced.
+    pub(crate) fn with_replaced_type(
+        self,
+        db: &'db dyn Db,
+        index: usize,
+        new_type: Type<'db>,
+    ) -> Self {
+        debug_assert!(index < self.types(db).len());
```
MichaReiser: Nit: I don't think this assertion is necessary. Rust always panics on out of bound index (even in release builds as not doing so would be UB)

crates/ty_python_semantic/src/types/visitor.rs
sharkdp: I made it 32 for now. I think we can only really fine-tune it once we combine this with type-of-`self`, which introduces those cases. I was worried that there are examples like
```py
from typing import Self

def duplicate[T](x: T) -> tuple[T, T]:
    return (x, x)

class C:
    def f(self: Self):
        self.x = duplicate(self.x)

reveal_type(C().x)
```
where the type expands *in breadth* exponentially at each layer.

sharkdp: Actually, I'm setting it back to 14, because that seems to be at the edge of what we can still handle in that particular scenario. I added this as a test case.

AlexWaygood: Thanks for trying it out

crates/ty_python_semantic/src/types/class.rs
mtshiba: Yes, as I said [here](https://github.com/astral-sh/ruff/pull/20566#issuecomment-3431449139), I'm in favor of provisionally merging the PR  anyway; the lack of self types means that a lot of recursive type inference is hidden by dynamic types.

crates/ty_python_semantic/src/types/visitor.rs
sharkdp: And here's an example of a case that we "can't handle" anymore with this PR (in the sense that we would lose precise type inference for that dict): https://play.ty.dev/e9b94a3d-704c-4078-90fc-ed9915cfb5e7

crates/ty_python_semantic/src/types/class.rs
sharkdp: Thank you for chiming in here, very much appreciated!

PR #20993, opened 2025-10-20
[ty] Fix completions at end of file

MichaReiser:
## Summary
Fixes https://github.com/astral-sh/ty/issues/1392

The methods on `Tokens` incorrectly assumed that there can be at most one token starting or ending at `offset`. 
Unfortunately, this isn't the case because Ruff's lexer can produce zero-length tokens. The most common
zero length tokens are the `Newline` at the end of the file (The lexer must emit this token because every logical line must end with a newline token), 
or `Dedent` tokens (which are emitted when there are "fewer" indents, they're always zero length).

The fix in this PR is to make the methods on `Tokens` aware of this and return the first token starting at 
or last token ending at a given offset (instead of any token starting or ending at).

## Test Plan

Added test

### DIFF
--- a/crates/ruff_python_parser/src/lib.rs
+++ b/crates/ruff_python_parser/src/lib.rs
@@ -646,9 +646,9 @@ impl Tokens {
         Self::before_impl(&self.raw, offset)
     }
 
-    fn before_impl(me: &[Token], offset: TextSize) -> &[Token] {
-        let partition_point = me.partition_point(|token| token.start() < offset);
-        let before = &me[..partition_point];
+    fn before_impl(tokens: &[Token], offset: TextSize) -> &[Token] {
+        let partition_point = tokens.partition_point(|token| token.start() < offset);
+        let before = &tokens[..partition_point];
 
         if let Some(last) = before.last() {
             // If it's equal to the end offset, then it's at a token boundary which is


crates/ty_ide/src/completion.rs
```diff
@@ -1453,6 +1452,26 @@ def frob(): ...
...
+
+def f(msg: Test):
+    msg.<CURSOR>",
+        );
+
+        assert_snapshot!(test.completions_if(|completion| matches!(&*completion.name, "x" | "y")), @r"
+        x
+        y
+        ");
+    }
```
sharkdp, line 1467: I also had a regression test here: https://github.com/astral-sh/ruff/pull/20951/files. That `Test` protocol seems overly complicated here?

crates/ruff_python_parser/src/lib.rs
```diff
@@ -489,6 +489,59 @@ impl Tokens {
...
+                Ok(idx)
+            }
+            Err(idx) => Err(idx),
+        }
+    }
+
+    /// Performs a binary search to find the index of the **last** token that ends at the given `offset`.
+    ///
+    /// Unlike `binary_search_by_key`, this method ensures that if multiple tokens end at the same offset,
+    /// it returns the index of the last one. Multiple tokens can emd at the same offset in cases where
```
sharkdp:
```suggestion
    /// it returns the index of the last one. Multiple tokens can end at the same offset in cases where
```

crates/ruff_python_parser/src/lib.rs
```diff
@@ -489,6 +489,59 @@ impl Tokens {
...
+                // For those, we want to get the first token starting at the given offset.
+                while let Some(previous) = idx.checked_sub(1).map(|idx| self[idx]) {
+                    if previous.start() == offset {
+                        idx -= 1;
+                    } else {
+                        break;
+                    }
+                }
+
+                Ok(idx)
```
sharkdp: An alternative functional form would be the following. I'll leave it up to you which one you prefer.
```suggestion
            Ok(idx) => {
                // Some tokens have an empty range (e.g. `Newline` at the end of the file or `Dedent`).
                // For those, we want to get the first token starting at the given offset.
                Ok((0..=idx)
                    .rev()
                    .take_while(|&i| self[i].start() == offset)
                    .last()
                    .unwrap_or(idx))
```

crates/ruff_python_parser/src/lib.rs
```diff
@@ -489,6 +489,59 @@ impl Tokens {
...
+                // we need to walk backwards.
+                while let Some(next) = me.get(idx + 1) {
+                    if next.end() == offset {
+                        idx += 1;
+                    } else {
+                        break;
+                    }
+                }
+
+                Ok(idx)
```
sharkdp: Similar here
```suggestion
            Ok(idx) => {
                // Some tokens have an empty range (e.g. `Newline` at the end of the file or `Dedent`).
                // For those, we want to return the last token that ends at the given offset. That's why
                // we need to walk backwards.
                Ok((idx..me.len())
                    .take_while(|&i| me[i].end() == offset)
                    .last()
                    .unwrap_or(idx))
```

crates/ruff_python_parser/src/lib.rs
MichaReiser: I think I could maybe use `partition` instead

sharkdp: `slice::partition_point`? Yes, that could work, but you would need to do an additional comparison to decide whether it's an `Ok` or `Err` variant.

crates/ruff_python_parser/src/lib.rs
```diff
@@ -649,28 +643,25 @@ impl Tokens {
...
-                            prev.range()
-                        );
-                    }
-                }
+        Self::before_impl(&self.raw, offset)
+    }
 
-                &self[..idx]
-            }
+    fn before_impl(me: &[Token], offset: TextSize) -> &[Token] {
```
dhruvmanila: nit: `tokens` instead of `me`?

PR #20994, opened 2025-10-20
[ty] Fix `None`, `True` and `False` highlighting in playground

MichaReiser:
## Summary

Add a highligthing rule for `builtinConstant` for proper highlighting of `True`, `None`, and `False` in the playground.

## Test Plan

Screenshot 2025-10-20 at 14 43 50


playground/issue.py
AlexWaygood, line 1: did you mean to add this empty file?

PR #20995, opened 2025-10-20
[ty] Fix bug where ty would think all types had an `__mro__` attribute

AlexWaygood:
## Summary

**This PR is best reviewed with the "hide whitespace changes" option enabled on GitHub.**

This PR fixes a bug where ty would consider all types as having an `__mro__` attribute, when in reality this is only true for instances of `type`. The bug is fixed by lifting our special-casing for the `__mro__` attribute out of `Type::member()` and into `TypeInferenceBuilder::infer_attribute_load()`. This means that we now only infer precise types for the `__mro__` attribute for literal attribute accesses, whereas we previously also inferred this precise type for implicit attribute accesses. I think that's okay, however: we only really need the precise inference of `__mro__` for our internal tests. Calling `iter_mro()` at a higher level (rather than in the guts of `Type::member()`) also appears to have the effect that we properly propagate the specialisation of a generic class down through the entire MRO when inferring the type of `someclass.__mro__`, resulting in more intuitive types being revealed in our `mro.md` test.

The type displayed in the tooltip for some autocomplete suggestions also becomes less precise, but that also seems fine. I don't think users of autocompletion need a type that's so precise there; the less precise type seems like it's probably less noisy.

Fixes https://github.com/astral-sh/ty/issues/986

## Test Plan

Existing mdtests updated and TODOs removed

### DIFF
--- a/crates/ty_python_semantic/src/types/function.rs
+++ b/crates/ty_python_semantic/src/types/function.rs
@@ -1677,6 +1677,7 @@ impl KnownFunction {
                         message.push('(');
                         for class in class.iter_mro(db) {
                             message.push_str(&class.display(db).to_string());
+                            // Omit the comma for the last element (which is always `object`)
                             if class
                                 .into_class()
                                 .is_none_or(|base| !base.is_object(context.db()))
@@ -1684,7 +1685,11 @@ impl KnownFunction {
                                 message.push_str(", ");
                             }
                         }
-                        if class.is_known(db, KnownClass::Object) {
+                        // If the last element was also the first element
+                        // (i.e., it's a length-1 tuple -- which can only happen if we're revealing
+                        // the MRO for `object` itself), add a trailing comma so that it's still a
+                        // valid tuple display.
+                        if class.is_object(db) {
                             message.push(',');
                         }
                         message.push(')');


sharkdp:
> This PR fixes a bug where ty would consider all types as having an `__mro__` attribute, when in reality this is only true for instances of `type`. The bug is fixed by lifting our special-casing for the `__mro__` attribute out of `Type::member()` and into `TypeInferenceBuilder::infer_attribute_load()`. This means that we now only infer precise types for the `__mro__` attribute for literal attribute accesses, whereas we previously also inferred this precise type for implicit attribute accesses.

It seems like this could have been fixed by moving that special handling to `class_member` instead?

Special-casing things in type inference (instead of core `Type` operations) is something I would wish we would do less of, not more. It's probably not very relevant for `.__mro__`, but in general, there are a lot of disadvantages to specializing things in type inference. The whole attribute access "machinery" with the descriptor protocol etc. will be overwritten, unions/intersections won't be handled, etc.

> The type displayed in the tooltip for some autocomplete suggestions also becomes less precise, but that also seems fine. I don't think users of autocompletion need a type that's so precise there

They probably don't *need* it, but I found it kind of nice?

dcreager:
> Special-casing things in type inference (instead of core `Type` operations) is something I would wish we would do less of, not more.

+1

AlexWaygood:
> Special-casing things in type inference (instead of core `Type` operations) is something I would wish we would do less of, not more.

In general, I absolutely agree. In this specific case, though, I worry that ty's inferred MRO for a class is often pretty different to a class's actual MRO at runtime, because of various fictions typeshed would (for very good reasons) have us believe. For example, we infer the MRO of `tuple` on `main` as `tuple[<class 'tuple[Unknown, ...]'>, <class 'Sequence[_T_co@tuple]'>, <class 'Reversible[_T_co@tuple]'>, <class 'Collection[_T_co@tuple]'>, <class 'Iterable[_T_co@tuple]'>, <class 'Container[_T_co@tuple]'>, typing.Protocol, typing.Generic, <class 'object'>]`. But at runtime:

```pycon
>>> tuple.__mro__
(<class 'tuple'>, <class 'object'>)
```

Not only is typeshed's MRO for `tuple` much longer than the actual one at runtime, the MRO that we display also contains generic aliases in it, which can never appear in an MRO at runtime -- only actual class objects can appear in MROs at runtime. The lengthy MRO typeshed gives us is a white lie from typeshed, because otherwise there'd be no way to get us or other type checkers to understand that `tuple` should be considered a subtype of `Sequence` (`Sequence` is not a `Protocol`, and that's an intentional design decision in the type system). But I think displaying these frankly inaccurate MROs in things like autocomplete tooltips could honestly be pretty confusing for users, and having them be propagated through all type inference by baking the special case into `Type::member()` could have unexpected results in far-off places; to me it seems much better to keep the special case localized to actual attribute expressions in the source code. As the codspeed report shows, keeping the special case localized also seems to result in performance improvements on some microbenchmarks!

Honestly, after writing this all out, I'm not even sure whether we should special-case inference of the `__mro__` attribute at all. Maybe our MRO tests should instead use a custom `ty_extensions.reveal_mro()` helper instead? Precise inference for the `__mro__` attribute here wasn't ever implemented because it was something we felt users needed; it was implemented so that we'd have an easy way to write isolated unit tests for our MRO inference logic.

sharkdp:
> Honestly, after writing this all out, I'm not even sure whether we should special-case inference of the `__mro__` attribute at all. Maybe our MRO tests should instead use a custom `ty_extensions.reveal_mro()` helper instead?

I got the same feeling when reading your comment and was glad to find that paragraph at the end :+1: :+1:

crates/ty_python_semantic/resources/mdtest/mro.md
```diff
@@ -160,23 +188,22 @@ An intersection that includes `Unknown` or `Any` is permitted as long as the int
...
+from ty_extensions import reveal_mro
 from does_not_exist import DoesNotExist  # error: [unresolved-import]
 
 reveal_type(DoesNotExist)  # revealed: Unknown
 
 if hasattr(DoesNotExist, "__mro__"):
-    # TODO: this should be `Unknown & <Protocol with members '__mro__'>` or similar
-    # (The second part of the intersection is incorrectly simplified to `object` due to https://github.com/astral-sh/ty/issues/986)
-    reveal_type(DoesNotExist)  # revealed: Unknown
+    reveal_type(DoesNotExist)  # revealed: Unknown & <Protocol with members '__mro__'>
```
AlexWaygood, line 197: this demonstrates the improved semantics from this PR

crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -329,9 +329,8 @@ reveal_type(typing.NamedTuple.__name__)  # revealed: str
 reveal_type(typing.NamedTuple.__qualname__)  # revealed: str
 reveal_type(typing.NamedTuple.__kwdefaults__)  # revealed: dict[str, Any] | None
 
-# TODO: this should cause us to emit a diagnostic and reveal `Unknown` (function objects don't have an `__mro__` attribute),
-# but the fact that we don't isn't actually a `NamedTuple` bug (https://github.com/astral-sh/ty/issues/986)
-reveal_type(typing.NamedTuple.__mro__)  # revealed: tuple[<class 'FunctionType'>, <class 'object'>]
+# error: [unresolved-attribute]
+reveal_type(typing.NamedTuple.__mro__)  # revealed: Unknown
```
AlexWaygood, line 333: this demonstrates the improved semantics from this PR

crates/ty_completion_eval/completion-evaluation-tasks.csv
```diff
@@ -20,6 +20,6 @@ scope-existing-over-new-import,main.py,0,474
 scope-prioritize-closer,main.py,0,2
 scope-simple-long-identifier,main.py,0,1
 tstring-completions,main.py,0,1
-ty-extensions-lower-stdlib,main.py,0,7
+ty-extensions-lower-stdlib,main.py,0,8
```
AlexWaygood, line 23: this change is because there is now another `ty_extensions` function (`reveal_mro`) that currently has a higher rank than `typing.reveal_type` in autocomplete suggestions, but would have a lower rank in an ideal world: https://github.com/astral-sh/ruff/blob/64ab79e5721ec6fdd2182fbf9d39a26534ccca43/crates/ty_completion_eval/truth/ty-extensions-lower-stdlib/main.py#L2

AlexWaygood:
The beartype primer hit is clearly a false negative going away. The sympy primer hits are a bit more complicated, but also look like true positives to me. The issue is that at line 177, we infer the type of `nargs` as `object | Any`: `Any` is the type of `nargs` before the `for` loop, and `object` is the type assigned to it from inside the `for` loop. `object | Any` just simplifies to `object`.

The conformance suite diff is also good: two false positives going away.

This PR should be ready for another round of review now.

sharkdp (approved):
Thank you! This looks great now.

PR #20996, opened 2025-10-20
[ty] Add basic `__slots__` type inference support

BlairAllan:
## Summary

Implements initial `__slots__` type checking as requested in astral-sh/ty#1268. This is the first PR which has come from PR #20636 which ballooned in size trying to solve all of the issues requests.

Now, when a class defines `__slots__`, we restricts instance attribute access to only the names listed in the slots:

- Accessing an attribute not in __slots__ produces `unresolved-attribute`.
- Accessing a valid slot name that was never assigned does not produce a diagnostic (but returns `Unknown` type).
- `__slots__` defined as a string literal (e.g., `__slots__ = "xyz"`) is treated as a single slot name, not individual
characters.

For example, like Pyright, the following no longer produces `unresolved-attribute`:

```python
class Foo:
    __slots__ = ("bar",)

foo = Foo()
reveal_type(foo.bar)
```

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/slots.md
+++ b/crates/ty_python_semantic/resources/mdtest/slots.md
@@ -1,0 +1,62 @@
+# `__slots__`
+
+## Basic slot access
+
+```py
+class A:
+    __slots__ = ("foo", "bar")
+
+    def __init__(self, foo: int, bar: str):
+        self.foo = foo
+        self.bar = bar
+
+a = A(1, "zip")
+a.foo = 2
+a.bar = "woo"
+a.baz = 3  # error: [unresolved-attribute]
+```
+
+## Accessing undefined attributes
+
+```py
+class A:
+    __slots__ = ("x",)
+
+a = A()
+a.y = 1  # error: [unresolved-attribute]
+```
+
+## Empty slots
+
+```py
+class A:
+    __slots__ = ()
+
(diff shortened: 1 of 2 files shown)


sharkdp:
This should have a bigger ecosystem impact now that type-of-self has landed

AlexWaygood:
I'm not sure I agree that we should emit `possibly-unresolved-attribute` if an item exists in `__slots__` without any explicit initialisation on the class or any instance methods, and is then accessed on an instance of the class. For a class like this, we assume instances of the class have the attribute `x` available at runtime, even though the attribute might never have been initialized:

```py
class F:
    x: int
```

`G` here doesn't seem all that different to me; it still has an explicit declaration of the attribute (via `__slots__`), it just doesn't have a type attached to that declaration (so we should infer `Unknown` — but that doesn't impact the boundness we should infer, I don't think)

```py
class G:
    __slots__ = "x",
```

charliermarsh:
I'm fine with that. I assumed it was intended / discussed elsewhere that we'd make it possibly-uninitialized.

AlexWaygood:
Not to my knowledge, i don't think we've really had any detailed discussions publicly or privately about exactly what the design of our `__slots__` support should look like

charliermarsh:
(Fixed.)

PR #20999, opened 2025-10-20
ci: adjust zizmor config, bump dist

woodruffw:
## Summary

Also bumps `cargo dist` to 0.30, and moves us
back to the upstream copy of `dist` now that
the latest version has integrated our fork's
patches.

## Test Plan

See what happens in CI 🙂

### DIFF
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -17,7 +17,26 @@ windows-archive = ".zip"
 # The archive format to use for non-windows builds (defaults .tar.xz)
 unix-archive = ".tar.gz"
 # Target platforms to build apps for (Rust target-triple syntax)
-targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "aarch64-unknown-linux-musl", "aarch64-pc-windows-msvc", "arm-unknown-linux-musleabihf", "armv7-unknown-linux-gnueabihf", "armv7-unknown-linux-musleabihf", "x86_64-apple-darwin", "powerpc64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu", "s390x-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc", "i686-unknown-linux-gnu", "i686-unknown-linux-musl", "i686-pc-windows-msvc"]
+targets = [
+    "aarch64-apple-darwin",
+    "aarch64-unknown-linux-gnu",
+    "aarch64-unknown-linux-musl",
+    "aarch64-pc-windows-msvc",
+    "arm-unknown-linux-musleabihf",
+    "armv7-unknown-linux-gnueabihf",
+    "armv7-unknown-linux-musleabihf",
+    "x86_64-apple-darwin",
+    "powerpc64-unknown-linux-gnu",
+    "powerpc64le-unknown-linux-gnu",
+    "riscv64gc-unknown-linux-gnu",
+    "s390x-unknown-linux-gnu",
+    "x86_64-unknown-linux-gnu",
+    "x86_64-unknown-linux-musl",
+    "x86_64-pc-windows-msvc",
+    "i686-unknown-linux-gnu",
+    "i686-unknown-linux-musl",
+    "i686-pc-windows-msvc"
+]
 # Whether to auto-include files like READMEs, LICENSEs, and CHANGELOGs (default true)
 auto-includes = false
 # Whether dist should create a Github Release or use an existing draft


.github/workflows/publish-ty-playground.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           node-version: 22
+          package-manager-cache: false
```
woodruffw: Noting: the reason we disable this cache is because there's a dataflow from the cache here to a "release" artifact, i.e. the CloudFlare deployment artifact. 

(Whether or not this constitutes a significant cache poisoning risk is a matter of taste/discretion 😅)

dist-workspace.toml
```diff
@@ -17,26 +17,7 @@ windows-archive = ".zip"
...
-    "riscv64gc-unknown-linux-gnu",
-    "s390x-unknown-linux-gnu",
-    "x86_64-unknown-linux-gnu",
-    "x86_64-unknown-linux-musl",
-    "x86_64-pc-windows-msvc",
-    "i686-unknown-linux-gnu",
-    "i686-unknown-linux-musl",
-    "i686-pc-windows-msvc"
-]
+targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "aarch64-unknown-linux-musl", "aarch64-pc-windows-msvc", "arm-unknown-linux-musleabihf", "armv7-unknown-linux-gnueabihf", "armv7-unknown-linux-musleabihf", "x86_64-apple-darwin", "powerpc64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu", "s390x-unknown-linux-gnu", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc", "i686-unknown-linux-gnu", "i686-unknown-linux-musl", "i686-pc-windows-msvc"]
```
AlexWaygood: ew, I preferred the previous formatting 😆 maybe we should add a TOML formatter to pre-commit...

AlexWaygood:
Does this PR mean we can get rid of this ignore from our zizmor config?

https://github.com/astral-sh/ruff/blob/44e678a222b37164c4ded4b7b6a19c9feffb4707/.github/zizmor.yml#L12

dist-workspace.toml
woodruffw: Yeah, it's unfortunate these churned 😅

woodruffw:
> Does this PR mean we can get rid of this ignore from our zizmor config?

Oh hmm, that's for `publish-playground` not `publish-ty-playground`. I'll take a look there.

.github/zizmor.yml
```diff
@@ -9,13 +9,18 @@ rules:
   cache-poisoning:
     ignore:
       - build-docker.yml
-      - publish-playground.yml
-      - ty-ecosystem-analyzer.yaml
-      - ty-ecosystem-report.yaml
```
woodruffw, line 14: I've removed these, since we now use inline ignore comments to ignore the exact acceptable cache risks rather than blanket ignoring all cache poisoning vectors 🙂 

(I did a review pass on each of these, and the primary risk in all cases is that a poisoned cache could influence one of the various HTML artifacts pushed to CloudFlare Pages. I believe that's an acceptable risk in practice!)

.github/zizmor.yml
```diff
@@ -9,13 +9,18 @@ rules:
...
     # it's hard to test what the impact of removing these ignores would be
     # without actually running the release workflow...
     ignore:
       - build-docker.yml
-      - publish-playground.yml
       - publish-docs.yml
+  secrets-inherit:
+    # `cargo dist` makes extensive use of `secrets: inherit`,
+    # and we can't easily fix that until an upstream release changes that.
+    disable: true
```
woodruffw, line 21: I went ahead and blanket-disabled these findings, since we don't have a remediation strategy until `cargo dist` uses fine-grained secret inheritance.

.github/zizmor.yml
```diff
@@ -9,13 +9,18 @@ rules:
...
+  template-injection:
+    ignore:
+      # like with `secrets-inherit`, `cargo dist` introduces some
+      # template injections. We've manually audited these usages for safety.
+      - release.yml
```
woodruffw, line 26: Same as above -- I manually reviewed these for risk and they're not a danger in their current trigger settings. We'll need cargo dist to remove them, at which point we can regenerate here.

.github/workflows/release.yml
```diff
@@ -1,7 +1,6 @@
-# This file was autogenerated by dist: https://github.com/astral-sh/cargo-dist
+# This file was autogenerated by dist: https://axodotdev.github.io/cargo-dist
```
AlexWaygood, line 1: Would like @Gankra's confirmation that migrating off the fork is okay and desirable 😄

dist-workspace.toml
AlexWaygood: I would actually prefer it if we reverted this change, if that's okay (sorry to be a pain) -- this is now a really long line, and it's also going to make `git blame` muddier for this section of the config

woodruffw: No worries, completely reasonable!

woodruffw: abc011d9c..f5aaf9d50

woodruffw:
CI is green, but waiting for @Gankra's blessing 🙂

.github/workflows/release.yml
Gankra: Yes

PR #21001, opened 2025-10-20
[ty] Implement go-to for binary and unary operators

MichaReiser:
## Summary

This PR implements goto and hover for binary operations and unary expressions. 

As for other hover targets, the implementation doesn't try to match overloads yet. 

This PR doesn't implement go-to support for augmented assign, which seems to match Pylance. I don't see a reason why we couldn't but I suggest leaving that for a separate PR

Fixes https://github.com/astral-sh/ty/issues/1151

## Test plan

Added tests

https://github.com/user-attachments/assets/5b04daea-c04f-4937-a899-2c9879e2d99a

### DIFF
--- a/crates/ty_python_semantic/src/types/ide_support.rs
+++ b/crates/ty_python_semantic/src/types/ide_support.rs
@@ -999,7 +999,7 @@ pub fn definitions_for_unary_op<'db>(
 
 /// Promotes literal types in `self` positions to their fallback instance types.
 ///
-/// This is so that we show `int.__add__` instead of `IntLiteral.__add__`.
+/// This is so that we show e.g. `int.__add__` instead of `Literal[4].__add__`.
 fn promote_literals_for_self<'db>(db: &'db dyn Db, ty: Type<'db>) -> Type<'db> {
     match ty {
         Type::BoundMethod(method) => Type::BoundMethod(method.map_self_type(db, |self_ty| {


crates/ty_ide/src/goto_definition.rs
```diff
@@ -798,26 +798,6 @@ my_func(my_other_func(a<CURSOR>b=5, y=2), 0)
         ");
     }
 
-    impl CursorTest {
```
MichaReiser, line 801: I moved this to the end. It felt weird having it in the middle of the tests

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -929,6 +930,72 @@ pub fn call_signature_details<'db>(
...
+    unary_op: &ast::ExprUnaryOp,
+) -> Option<(Vec<ResolvedDefinition<'db>>, Type<'db>)> {
+    let operand_ty = unary_op.operand.inferred_type(model);
+
+    let unary_dunder_method = match unary_op.op {
+        ast::UnaryOp::Invert => "__invert__",
+        ast::UnaryOp::UAdd => "__pos__",
+        ast::UnaryOp::USub => "__neg__",
+        ast::UnaryOp::Not => "__bool__",
+    };
```
MichaReiser, line 974: It didn't feel worth pulling this out into its own `Type` method, especially because `__bool__` is somewhat complicated when all we need here is to get the`__bool__` dunder

crates/ty_python_semantic/src/types/call.rs
```diff
@@ -1,14 +1,87 @@
...
+        // We either want to call lhs.__op__ or rhs.__rop__. The full decision tree from
+        // the Python spec [1] is:
+        //
+        //   - If rhs is a (proper) subclass of lhs, and it provides a different
+        //     implementation of __rop__, use that.
+        //   - Otherwise, if lhs implements __op__, use that.
+        //   - Otherwise, if lhs and rhs are different types, and rhs implements __rop__,
+        //     use that.
+        //
+        // [1] https://docs.python.org/3/reference/datamodel.html#object.__radd__
```
MichaReiser, line 29: I extracted this from `builder.rs`. The lines are unchanged except that the `map(|outcome| outcome.return_type())` remains in `builder.rs`

crates/ty_ide/src/goto.rs
```diff
@@ -835,9 +894,19 @@ pub(crate) fn find_goto_target(
...
+
+            // if we have a<CURSOR>+b`, prefer the `+` token (by respecting the token ordering)
+            // This matches VS Code's behavior where it sends the start of the clicked token as offset.
+            kind if kind.as_binary_operator().is_some() || kind.as_unary_operator().is_some() => 1,
             _ => 0,
         })?;
 
+    if token.kind().is_comment() {
+        return None;
+    }
```
dhruvmanila, line 939: Just out of curiosity, is this to avoid returning `Some` when the cursor is inside a comment?

MichaReiser: Short answer: yes :)

The problem with binary expressions and unary expressions is that we don't know the operator's range. That means, we would return `Some` for

```py
(
	a + # comment<CURSOR>
	b
)
```

Because the range is between `a` and `b`. But that's obviously not what we want.

AlexWaygood:
One _tiny_ UI nit (possibly not solvable, possibly playground-specific, just posting it because I noticed it):

If I have the `cmd` key held down and hover over a symbol, the symbol changes colour and becomes underlined, and my mouse symbol changes to a little ✋ to tell me that I can click on it to go to its definition. But the same thing doesn't happen if I have the `cmd` key held down and hover over an operator, so it's hard for me to tell that cmd-clicking on it is going to take me to the method being displayed by the on-hover tooltip. Here's a video in which I have the `cmd` key held down for the duration:

https://github.com/user-attachments/assets/503ce993-572c-4a66-a91e-87433be73df2

crates/ty_python_semantic/src/types/ide_support.rs
AlexWaygood: You could _possibly_ consider adding methods like these to the AST nodes directly, similar to what we already have for the bin-op nodes:

https://github.com/astral-sh/ruff/blob/e1cada1ec30d289acb00390058fab8bce765836e/crates/ruff_python_ast/src/nodes.rs#L2546-L2602

But yeah, `not` is complicated because it falls back to `__len__` as well as `__bool__`, even if you ignore all the fancy special casing we have baked in for the truthiness of various types.

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -929,6 +930,91 @@ pub fn call_signature_details<'db>(
...
+            ))
+        })
+        .collect();
+
+    Some((definitions, callable_type))
+}
+
+/// Promotes literal types in `self` positions to their fallback instance types.
+///
+/// This is so that we show `int.__add__` instead of `IntLiteral.__add__`.
```
AlexWaygood:
```suggestion
/// This is so that we show e.g. `int.__add__` instead of `Literal[4].__add__`.
```

MichaReiser:
> If I have the cmd key held down and hover over a symbol, the symbol changes colour and becomes underlined, and my mouse symbol changes to a little ✋ to tell me that I can click on it to go to its definition. But the same thing doesn't happen if I have the cmd key held down and hover over an operator, so it's hard for me to tell that cmd-clicking on it is going to take me to the method being displayed by the on-hover tooltip. Here's a video in which I have the cmd key held down for the duration:

I noticed this too but it's not clear to me why it is happening, because we report the full range of the binary expression as its range. I can try to use a narrower range to see if that helps

MichaReiser:
Nope, that doesn't help (but I think is strictly better). I did give r-a a try and it also doesn't underline `+` and other operands. We also correctly set the [`originSelectionRange`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink) but it seems VS code doesn't care?

It does work fine with `not`.

Confirmed, Zed renders this correctly

PR #21002, opened 2025-10-20
remove `playground/issue.py`

AlexWaygood:
Pretty sure we don't need this file :-)

(Added in #20994)


MichaReiser:
Hmm. I did klick discard in the github ui... 

Thank you

PR #21003, opened 2025-10-20
[`pyupgrade`] Always parenthesize assignment expressions in fix for `f-string` (`UP032`)

### LINKED ISSUE #21000: ruff suggests invalid code when fixing UP032
### Summary

Consider the following code:
```python
if __name__ == "__main__":
    number = 0
    string = "{}".format(number := number + 1)
    print(string)
```

It executes successfully and prints `1`. 

When I run ruff, it returns
```
bad_ruff.py:3:14: UP032 [*] Use f-string instead of `format` call
  |
1 | if __name__ == "__main__":
2 |     number = 0
3 |     string = "{}".format(number := number + 1)
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP032
4 |     print(string)
  |
  = help: Convert to f-string
```

When I run `ruff check --fix bad_ruff.py`, the resulting code is invalid and results in `ValueError`:

```python
if __name__ == "__main__":
    number = 0
    string = f"{number := number + 1}"
    print(string)

```

### Version

ruff 0.11.13 (5faf72a4d 2025-06-05)

dylwil3:
Closes #21000

### DIFF
--- a/crates/ruff_linter/src/rules/pyupgrade/snapshots/ruff_linter__rules__pyupgrade__tests__UP032_0.py.snap
+++ b/crates/ruff_linter/src/rules/pyupgrade/snapshots/ruff_linter__rules__pyupgrade__tests__UP032_0.py.snap
@@ -1320,6 +1320,8 @@ UP032 [*] Use f-string instead of `format` call
 270 | # Refer: https://github.com/astral-sh/ruff/issues/11736
 271 | x: "'{} + {}'.format(x, y)"
     |     ^^^^^^^^^^^^^^^^^^^^^^
+272 |
+273 | # Regression https://github.com/astral-sh/ruff/issues/21000
     |
 help: Convert to f-string
 268 | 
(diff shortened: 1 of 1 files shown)


crates/ruff_linter/src/rules/pyupgrade/rules/f_strings.rs
```diff
@@ -167,6 +166,7 @@ fn parenthesize(expr: &Expr, text: &str, context: FormatContext) -> bool {
             _,
             Expr::Generator(_)
             | Expr::Dict(_)
+            | Expr::Named(_)
```
MichaReiser, line 169: It would be nice if we could use OperatorPrecedence here (I believe that's the name) instead of listing all expressions

MichaReiser (approved):

crates/ruff_linter/src/rules/pyupgrade/rules/f_strings.rs
dylwil3: Hm... it's not immediately clear how to do this here - especially because some of this logic is to avoid double escaped braces for sets and dicts, like `f"{{x}}"`

Issue #20482, opened 2025-09-19
Ruff 2026 Style Guide


dylwil3:
From [Black's changelog](https://github.com/psf/black/blob/main/CHANGES.md) and [`Preview` enum](https://github.com/psf/black/blob/43135e9fafccbca723910a45aa62f0f182e85e5f/src/black/mode.py#L223). Last checked 19.09.2025, Last release: v25.9.0.

Goals:
* Close parity with Black when migrating from Black to Ruff
* Fix Ruff specific bugs
* Innovate on local formatting (to keep close parity with black)

### Stabilized in Black and not available in Ruff (even in preview)
* [ ] [`remove_lone_list_item_parens`](https://github.com/psf/black/pull/4312)
* [x] https://github.com/astral-sh/ruff/issues/9745  -- progress towards this here https://github.com/astral-sh/ruff/pull/21110

### Black preview styles

* [x] [`wrap_long_dict_values_in_parens`](https://github.com/psf/black/pull/3440) Non-goal, see https://github.com/psf/black/issues/4123 and https://github.com/astral-sh/ruff/issues/12856
* [ ] [`always_one_newline_after_import`](https://github.com/psf/black/pull/4489) This is enforced by [unsorted-imports (I001)](https://docs.astral.sh/ruff/rules/unsorted-imports/#unsorted-imports-i001) and configurable via [`lines-after-imports`](https://docs.astral.sh/ruff/settings/#lint_isort_lines-after-imports). There is already some small conflict with the formatter depending on how the user configures this, and this may cause a little more. Unsure about this one...
* [x] [`fix_fmt_skip_in_one_liners`](https://github.com/psf/black/pull/4552) https://github.com/astral-sh/ruff/pull/20633 https://github.com/astral-sh/ruff/pull/22119
* [x] [`wrap_comprehension_in`](https://github.com/psf/black/pull/4699)
  * https://github.com/astral-sh/ruff/pull/21005
* [x] [`remove_parens_around_except_types`](https://github.com/psf/black/pull/4720)
  * https://github.com/astral-sh/ruff/pull/20768
* [x] [`normalize_cr_newlines`](https://github.com/psf/black/pull/4710) Already handled by Ruff https://github.com/astral-sh/ruff/issues/20482#issuecomment-3314361284

### Black unstable styles

[...]

Issue #12870, opened 2024-08-13
Comprehensions do not split as far as they can go, which preserves long line lengths


rogersheu:
**Keywords**
Line length, comprehension

**Example**
*Before `ruff --fix`
```
really_really_really_long_dict_name = {"foo": "bar"}

dict_with_really_long_names = [(really_really_really_long_key_name, really_really_really_long_key_value) for really_really_really_long_key_name, really_really_really_long_key_value in really_really_really_long_dict_name.items()]
```

After `ruff --fix`

```
dict_with_really_long_names = [
    (really_really_really_long_key_name, really_really_really_long_key_value)
    for really_really_really_long_key_name, really_really_really_long_key_value in really_really_really_long_dict_name.items()
]
```

That `for/in` line is 127 characters long, but it could come under the line length (default of 88, not set here) by splitting the `for/in` statements, which does not affect behavior in this case.

**Version/Command**
Using ruff v0.5.1 with pre-commit calling `ruff --fix`.

MichaReiser:
To clarify. Are you running `ruff format` or `ruff check --fix`? I  suspect it's the former because `E501` doesn't come with an autofix.

MichaReiser:
Thanks.  Yeah, it is then the formatter that doesn't split the comprehension further. 

The main question here is what should happen when the comprehension does break over multiple lines:

```diff
 lcomp3 = [
         # This one is actually too long to fit in a single line.
         element.split("\n", 1)[0]
         yup
-        for element in collection.select_elements()
+        for element
+        in collection.select_elements()
         # right
         if element is not None
 ]
```

Because this does look worse. The desired result is probably that the `for ... in` should be kept together if possible

MichaReiser:
I think my biggest concern is that it hurts readability if there are multiple generators

```python
[
    [
        x 
        for x
        in bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
        for y 
        in xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxdxxxxxxxxxxxxxxxxxxxxxxxx
    ]
]
```

MichaReiser:
Thanks for linking to the relevant black issue. This is very helpful for both projects :)

MichaReiser:
[...]

Issue #12856, opened 2024-08-13
Nested expression formatting


MichaReiser:
I was surprised that there's no existing issue for this. 

We decided not to implement the following Black 24 style changes:

* `parenthesize_long_type_hints,` 
* `wrap_long_dict_values_in_parens` 
* `parenthesize_conditional_expressions`

Because a) implementing this in Ruff is kind of hard (or slow) and b) it goes against the overall goal to avoid unnecessary parentheses. 

[This Black issue](https://github.com/psf/black/issues/4123) goes into more detail about our concerns. We should explore different designs that yields better sub-expression formatting without requiring optional parentheses to improve readability.

PR #21005, opened 2025-10-20
[WIP] Implement the `wrap_comprehension_in` preview style from Black

ntBre:
This is a rough draft with a naive fix. We still disagree with Black on some formatting, for example this snippet from our docstring formatting tests:

```py
def by_first_letter_of_column_values(self, col: str) -> list[pl.DataFrame]:
    return [
        self._df.filter(pl.col(col).str.starts_with(c))
        for c in sorted(set(df.select(pl.col(col).str.slice(0, 1)).to_series()))
    ]
```

Black reuses the parentheses from the `sorted` call instead of adding new parentheses around the whole thing, which seems preferable.

**Black**:

```py
def by_first_letter_of_column_values(self, col: str) -> list[pl.DataFrame]:
    return [
        self._df.filter(pl.col(col).str.starts_with(c))
        for c in sorted(
            set(df.select(pl.col(col).str.slice(0, 1)).to_series())
        )
    ]
```

**This PR**:
```py
def by_first_letter_of_column_values(self, col: str) -> list[pl.DataFrame]:
    return [
        self._df.filter(pl.col(col).str.starts_with(c))
        for c in (
            sorted(set(df.select(pl.col(col).str.slice(0, 1)).to_series()))
        )
    ]
```

I can't quite tell if I'm having trouble here because this is tricky to implement in Ruff as Dylan mentioned [here](https://github.com/astral-sh/ruff/issues/20482#issuecomment-3340449328), or if I'm still just unfamiliar with the formatter.

Summary
--

This PR implements the `wrap_comprehension_in` style added in
https://github.com/psf/black/pull/4699. This wraps `in` clauses in
comprehensions if they get too long. Using some examples from the upstream
issue, this code:

```py
[a for graph_path_expression in refined_constraint.condition_as_predicate.variables]

[
    a
    for graph_path_expression
    in refined_constraint.condition_as_predicate.variables
]
```

is currently formatted to:

```py
[
    a
    for graph_path_expression in refined_constraint.condition_as_predicate.variables
]

[
    a
    for graph_path_expression in refined_constraint.condition_as_predicate.variables
]
```

even if the second line of the comprehension exceeds the configured line length.

In preview, black will now break these lines by parenthesizing the expression
following `in`:

```py
[
    a
    for graph_path_expression in (
        refined_constraint.condition_as_predicate.variables
    )
]

[
    a
    for graph_path_expression in (
        refined_constraint.condition_as_predicate.variables
    )
]
```

I actually kind of like the alternative formatting mentioned on the original
Black issue and in our #12870 which would be more like:

```py
[
    a
    for graph_path_expression
	in refined_constraint.condition_as_predicate.variables
]
```

but I think I'm in the minority there.

Test Plan
--

Existing Black compatibility tests showing fewer differences

### DIFF
--- a/crates/ruff_python_formatter/src/expression/parentheses.rs
+++ b/crates/ruff_python_formatter/src/expression/parentheses.rs
@@ -70,6 +70,10 @@ pub(crate) enum Parenthesize {
     /// when outer parentheses are present.
     IfBreaksParenthesizedNested,
 
+    /// Parenthesize the expression if it doesn't fit on a line or if the expression is
+    /// parenthesized in the source code. This is similar to `Self::Optional` except that this
+    /// variant will also include parentheses in nested calls, like
+    /// `Self::IfBreaksParenthesizedNested`.
     IfBreaksComprehension,
 }
 


crates/ruff_python_formatter/src/other/comprehension.rs
```diff
@@ -166,3 +192,68 @@ impl Format<PyFormatContext<'_>> for ExprTupleWithoutParentheses<'_> {
...
+/// ```python
+/// [
+///     x
+///     for x in {  # _not_ `({`
+///         x
+///         for x in "long line long line long line long line long line long line long line"
+///     }
+/// ]
+/// ```
+fn needs_nested_parentheses(expr: &Expr) -> bool {
```
MichaReiser: You probably want has_own_parentheses

MichaReiser:
I think the comment here is relevant for your work 

https://github.com/astral-sh/ruff/blob/c9dff5c7d5bcab2a83f1c3b4a1a80af230ece541/crates/ruff_python_formatter/src/expression/mod.rs#L386-L397

I'm also not sure if we should implement this preview style, as @dylwil3 pointed out in https://github.com/astral-sh/ruff/issues/20482#issuecomment-3340449328 because it might fall into the same category as https://github.com/astral-sh/ruff/issues/12856 where Black now starts introducing otherwise unnecessary parentheses. Are there alternative formattings that we could use that avoid the need for inserting parentheses?

MichaReiser:
Reading through the issue, the main concern seems to be that very long attribute chains aren't split. That makes me wonder if the proper fix instead is to split attribute expressions in parenthesized expressions. I suspect this would be a bigger change and we probably want to give attribute chains a very low split priority. Or we could decide to only indent the content rather than adding an extra pair of parentheses?

ntBre:
Thanks for the pointers here and in our 1:1! The draft is in a slightly better state now, at least in terms of the tests and the ecosystem results. However, I'm now splitting expressions like this too eagerly on the `in`:

```py
unformatted: File would be reformatted
 --> /tmp/tmp.h5wCjfHytw/try.py:1:1
2 | async def api_get_user_extensions(
3 |     user: User = Depends(check_user_exists),
4 | ) -> list[Extension]:
  - 
  -     user_extensions_ids = [
  -         ue.extension for ue in await get_user_extensions(user.id)
  -     ]
  -     return [
  -         ext
  -         for ext in await get_valid_extensions(False)
  -         if ext.code in user_extensions_ids
  -     ]
5 +     user_extensions_ids = [ue.extension for ue in await get_user_extensions(user.id)]
6 +     return [ext for ext in await get_valid_extensions(
7 +             False
8 +         ) if ext.code in user_extensions_ids]
```

Similar cases make up all of the ecosystem results I've looked at.

Is there an easy way to avoid that? It does seem to be something with `can_omit_optional_parentheses` right where you linked me on Discord. The split priority you mentioned above sounded perfect, but I didn't turn up anything with grep.

I will probably move on to another preview style as you and Dylan said, unless this is looking promising. It seems like there are other designs worth exploring in the future anyway.

MichaReiser:
> Thanks for the pointers here and in our 1:1! The draft is in a slightly better state now, at least in terms of the tests and the ecosystem results. However, I'm now splitting expressions like this too eagerly on the in:

This makes me wonder if you even want `maybe_parenthesize_expression`. Can you formalize the desired split and parenthesizing behavior?

* Should lines split from left to right? (We parenthesize the `in` before parenthesizing the if)
* Should lines split from right to left (We parenthesize the `if` part before parenthesizing the `in` part?
* Does parenthesizing the `in` part always expand the comprehension (`[` and `]` are on their own line)?
* Where should newlines be added when the `in` part gets parenthesized or splits over multiple lines (as you can see with `await`)
* ...

ntBre:
This doesn't feel very formal, so I don't think it fully answers your questions, but my understanding based on the black tests are that this preview style should only come into effect after all of the usual splits have been done and the `in` clause is still too long. For the example above, black leaves the input alone:

```py
@extension_router.get("")
async def api_get_user_extensions(
    user: User = Depends(check_user_exists),
) -> list[Extension]:

    user_extensions_ids = [
        ue.extension for ue in await get_user_extensions(user.id)
    ]
    return [
        ext
        for ext in await get_valid_extensions(False)
        if ext.code in user_extensions_ids
    ]
```

If I extend the `await` expression enough, it eventually breaks like this on both stable and preview:

```py
@extension_router.get("")
async def api_get_user_extensions(
    user: User = Depends(check_user_exists),
) -> list[Extension]:

    user_extensions_ids = [
        ue.extension for ue in await get_user_extensions(user.id)
    ]
    return [
        ext
        for ext in await get_valid_extensions(
            Falseeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
        )
        if ext.code in user_extensions_ids
    ]
```

which also makes sense to me.

This part is a bit confusing to me, but if I remove the `await` and the `call`, no split happens with the new preview style:

```py
@extension_router.get("")
async def api_get_user_extensions(
    user: User = Depends(check_user_exists),
) -> list[Extension]:

    user_extensions_ids = [
        ue.extension for ue in await get_user_extensions(user.id)
    ]
    return [
        ext
        for ext in aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    ]
```

If I make that an attribute expression instead (replacing an `a` with a `.` to avoid changing the length), it does split:

```py
@extension_router.get("")
async def api_get_user_extensions(
    user: User = Depends(check_user_exists),
) -> list[Extension]:

    user_extensions_ids = [
        ue.extension for ue in await get_user_extensions(user.id)
    ]
    return [
        ext
        for ext in (
            aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
        )
    ]
```

I get that same splitting even with a following `if` clause, as I would expect.

Black also explicitly tests that they _don't_ parenthesize the `in` clause if it's still too long [here](https://github.com/psf/black/blob/6122359d2029f24bcac9fd34c930ff78672ae840/tests/data/cases/preview_wrap_comprehension_in.py#L117-L127).

So at least from these tests, it seems that the comprehension always gets fully expanded like the stable behavior, and then if the for-in part is still too long, and it's a specific type of expression like an attribute chain, it can be parenthesized and split across multiple lines.

Does that help at all? Maybe I need to take a look at `normalize_invisible_parens` and the rest of the Black implementation too.

MichaReiser:
Thanks for explaining. What I understand is:

* We want a right-to-left layout where the `in` part breaks before the `for element in`
* We only want to break the `for element in` over multiple lines if breaking the `in` on its own isn't enough. But, if we do so, we don't want to break the `in` unless it's necessary

> If I make that an attribute expression instead (replacing an a with a . to avoid changing the length), it does split:

But you changed more than just from identifier to attribute. The attribute expression is shorter. I think this is just

> Black also explicitly tests that they don't parenthesize the in clause if it's still too long [here](https://github.com/psf/black/blob/6122359d2029f24bcac9fd34c930ff78672ae840/tests/data/cases/preview_wrap_comprehension_in.py#L117-L127).

We did implement this behavior in a few places but it's fairly tricky and I'm honestly not sure if it's worth it. Some users find it confusing (it goes against: predictable formatting), but it does avoid parentheses in some places, like assignment, where I think it makes sense.

As discussed with Dylan. I'm not convinced that this always improves formatting. Have you considered alternative formattings that don't require inserting parentheses but result in the same improvement? 


Either way. I haven't seen this come up. I suggest deprioritizing it. We have other formatter issues that have been requested more frequently (I don't remember that this ever came up)

ntBre:
> Either way. I haven't seen this come up. I suggest deprioritizing it. We have other formatter issues that have been requested more frequently (I don't remember that this ever came up)

Sounds good, thanks for talking through it. I still learned a lot!

> Have you considered alternative formattings that don't require inserting parentheses but result in the same improvement?

I kind of liked this formatting mentioned on the upstream issue, but that's about as much as I've considered it.

```
[
    a
    for graph_path_expression
	in refined_constraint.condition_as_predicate.variables
]
```

PR #21010, opened 2025-10-21
[ty] Add new "constraint implication" typing relation

dcreager:
This PR adds the new **_constraint implication_** relationship between types, aka `is_subtype_of_given`, which tests whether one type is a subtype of another _assuming that the constraints in a particular constraint set hold_.

For concrete types, constraint implication is exactly the same as subtyping. (A concrete type is any fully static type that is not a typevar. It can _contain_ a typevar, though — `list[T]` is considered concrete.)

The interesting case is typevars. The other typing relationships (TODO: will) all "punt" on the question when considering a typevar, by translating the desired relationship into a constraint set. At some point, though, we need to resolve a constraint set; at that point, we can no longer punt on the question. Unlike with concrete types, the answer will depend on the constraint set that we are considering.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -213,6 +213,37 @@ impl<'db> ConstraintSet<'db> {
 
     /// Returns the constraints under which `lhs` is a subtype of `rhs`, assuming that the
     /// constraints in this constraint set hold.
+    ///
+    /// For concrete types (types that are not typevars), this returns the same result as
+    /// [`when_subtype_of`][Type::when_subtype_of]. (Constraint sets place restrictions on
+    /// typevars, so if you are not comparing typevars, the constraint set can have no effect on
+    /// whether subtyping holds.)
+    ///
+    /// If you're comparing a typevar, we have to consider what restrictions the constraint set
+    /// places on that typevar to determine if subtyping holds. For instance, if you want to check
+    /// whether `T ≤ int`, then answer will depend on what constraint set you are considering:
+    ///
+    /// ```text
+    /// when_subtype_of_given(T ≤ bool, T, int) ⇒ true
+    /// when_subtype_of_given(T ≤ int, T, int)  ⇒ true
+    /// when_subtype_of_given(T ≤ str, T, int)  ⇒ false
+    /// ```
+    ///
+    /// In the first two cases, the constraint set ensures that `T` will always specialize to a
+    /// type that is a subtype of `int`. In the final case, the constraint set requires `T` to
+    /// specialize to a subtype of `str`, and there is no such type that is also a subtype of
+    /// `int`.
+    ///
+    /// There are two constraint sets that deserve special consideration.
+    ///
+    /// - The "always true" constraint set does not place any restrictions on any typevar. In this
+    ///   case, `when_subtype_of_given` will return the same result as `when_subtype_of`, even if
+    ///   you're comparing against a typevar.
+    ///
+    /// - The "always false" constraint set represents an impossible situation. In this case, every
+    ///   subtype check will be vacuously true, even if you're comparing two concrete types that
+    ///   are not actually subtypes of each other. (That is,
+    ///   `when_subtype_of_given(false, int, str)` will return true!)
     pub(crate) fn when_subtype_of_given(
         self,
         db: &'db dyn Db,


crates/ty_python_semantic/src/types.rs
dcreager, line 1: I had originally written this as a new `TypeRelation` variant, but that was too heavy-handed, given that constraint implication is exactly the same as subtyping for everything except for typevars.

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of_given.md
```diff
@@ -0,0 +1,184 @@
...
+
+def even_given_unsatisfiable_constraints():
+    static_assert(is_subtype_of_given(False, bool, int))
+    static_assert(not is_subtype_of_given(False, bool, str))
+```
+
+## Type variables
+
+The interesting case is typevars. The other typing relationships (TODO: will) all "punt" on the
+question when considering a typevar, by translating the desired relationship into a constraint set.
```
dcreager, line 50: This TODO is #20093

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of_given.md
```diff
@@ -0,0 +1,184 @@
...
+    static_assert(is_subtype_of_given(given_bool, T, bool))
+    static_assert(not is_subtype_of_given(given_bool, T, str))
+
+    given_str = range_constraint(Never, T, str)
+    static_assert(not is_subtype_of_given(given_str, T, int))
+    static_assert(not is_subtype_of_given(given_str, T, bool))
+    static_assert(is_subtype_of_given(given_str, T, str))
+```
+
+This might require propagating constraints from other typevars.
```
dcreager, line 177: This will be involved enough that I'm tackling it separately, in #21068

crates/ty_python_semantic/src/types.rs
```diff
@@ -8500,7 +8524,6 @@ pub struct BoundTypeVarIdentity<'db> {
 /// A type variable that has been bound to a generic context, and which can be specialized to a
 /// concrete type.
 #[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
-#[derive(PartialOrd, Ord)]
```
dcreager, line 8537: We should always be comparing a typevar's identity, and this makes it harder to accidentally compare specific instances of a typevar.

crates/ty_python_semantic/src/types/type_ordering.rs
```diff
@@ -137,7 +139,7 @@ pub(super) fn union_or_intersection_elements_ordering<'db>(
         (Type::ProtocolInstance(_), _) => Ordering::Less,
         (_, Type::ProtocolInstance(_)) => Ordering::Greater,
 
-        (Type::TypeVar(left), Type::TypeVar(right)) => left.cmp(right),
+        (Type::TypeVar(left), Type::TypeVar(right)) => left.as_id().cmp(&right.as_id()),
```
dcreager, line 144: but this is the one place where we do still want to compare typevar instances, not identities

crates/ty_vendored/ty_extensions/ty_extensions.pyi
```diff
@@ -62,26 +62,35 @@ def negated_range_constraint(
...
+    """Returns a constraint set that is satisfied when `ty` is a `subtype`_ of `of`.
+
+    .. _subtype: https://typing.python.org/en/latest/spec/concepts.html#subtype-supertype-and-type-equivalence
+    """
+
+def is_subtype_of_given(
+    constraints: bool | ConstraintSet, ty: Any, of: Any
+) -> ConstraintSet:
+    """Returns a constraint set that is satisfied when `ty` is a `subtype`_ of `of`,
+    assuming that all of the constraints in `constraints` hold.
```
MichaReiser, line 82: I just wanted to suggest making this a method on `ConstraintSet` but that doesn't work because you also want to allow `bool` as `constraints`. 

I'm not sure I fully understand the semantics here. Is this roughly if and only if `constraints` semantics? If that's the case, I think I'd find `is_subtype_of_iff(ty, of, constraints)` easier to understand.

But I don't think that's what this API does after reading the mdtests. Hmm :)


I'm asking because I first assumed the relation between `is_subtype_of` and `is_subtype_of_given` is similar to `std::mem::size_of` and `std::mem::size_of_val` where the latter takes a value. I don't think this is the case here because `constraints` would than have to be a subtype of `ty`, which the API doesn't require.

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of_given.md
```diff
@@ -0,0 +1,184 @@
...
+
+```py
+from ty_extensions import is_subtype_of, is_subtype_of_given, static_assert
+
+def equivalent_to_other_relationships[T]():
+    static_assert(is_subtype_of(bool, int))
+    static_assert(is_subtype_of_given(True, bool, int))
+
+    static_assert(not is_subtype_of(bool, str))
+    static_assert(not is_subtype_of_given(True, bool, str))
```
MichaReiser, line 26: Is `is_subtype_of(T, V)` guaranteed to always be the same as `is_subtype_of(True, T, V)`. If so, maybe that's something you could add to the documentation of `is_subtype_of_given`

crates/ty_python_semantic/src/types.rs
```diff
@@ -4837,6 +4837,30 @@ impl<'db> Type<'db> {
...
+                        Parameters::new([
+                            Parameter::positional_only(Some(Name::new_static("a")))
+                                .with_annotated_type(UnionType::from_elements(
+                                    db,
+                                    [
+                                        KnownClass::Bool.to_instance(db),
+                                        KnownClass::ConstraintSet.to_instance(db),
+                                    ],
+                                )),
+                            Parameter::positional_only(Some(Name::new_static("a")))
```
MichaReiser: Is it intentional that the first two parameters have the same name? Should we use the same names as in the `pyi` files (it would allow me to easier match the parameters)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -175,6 +176,31 @@ impl<'db> ConstraintSet<'db> {
...
+    /// Returns a constraint set that constraints a typevar to a particular range of types.
+    pub(crate) fn constrain_typevar(
+        db: &'db dyn Db,
+        typevar: BoundTypeVarInstance<'db>,
+        lower: Type<'db>,
+        upper: Type<'db>,
+        relation: TypeRelation,
+    ) -> Self {
+        let (lower, upper) = match relation {
+            // TODO: Is this the correct constraint for redundancy?
```
MichaReiser, line 188: Is it? ;)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -297,7 +348,69 @@ impl<'db> ConstrainedTypeVar<'db> {
...
+
+        // We have an (arbitrary) ordering for typevars. If the upper and/or lower bounds are
+        // typevars, we have to ensure that the bounds are "later" according to that order than the
+        // typevar being constrained.
+        //
+        // In the comments below, we use brackets to indicate which typevar is "earlier", and
+        // therefore the typevar that the constraint applies to.
+        match (lower, upper) {
+            // L ≤ T ≤ L == (T ≤ [L] ≤ T)
+            (Type::TypeVar(lower), Type::TypeVar(upper)) if lower == upper => {
```
MichaReiser: Should this use `is_same_typevar`?

crates/ty_python_semantic/src/types/type_ordering.rs
MichaReiser: Should we add an inline comment explaining why that's the case?

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of_given.md
dcreager: This is a great catch! I was in the middle of writing that this would not be true when comparing typevars — but you're right, and it _is_ true even then:

```py
is_subtype_of(bool, int)               # yes
is_subtype_of_given(True, bool, int)   # yes
is_subtype_of_given(False, bool, int)  # yes

def f[T]():
    is_subtype_of(T, int)              # no (T could be anything!)
    is_subtype_of_given(True, T, int)  # also no (T could be anything!)
```

This also made me realize that I don't need to (and shouldn't) project away the other typevars like I was doing.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: @AlexWaygood do you have thoughts here? I was thinking that we should treat redundancy the same as subtyping here. Adding more context, the question is, if you call `is_redundant_with(T, list[Any])`, what constraint should we turn that into? For `is_subtype_of(T, list[Any])`, we turn it into `T ≤ Bottom[list[Any]]`, since for subtyping, the relationship has to hold for _all_ materializations. For `is_assignable_to(T, list[Any])`, we turn it into `T ≤ Top[list[Any]]`, since the relationship only has to hold for _some_ materialization.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Yes!

crates/ty_python_semantic/src/types/type_ordering.rs
dcreager: Done

crates/ty_python_semantic/src/types.rs
dcreager: Nope, that was bad copy/pasta. Done

(And once we have support for `TypeForm`, all of these special cases can go away)

crates/ty_vendored/ty_extensions/ty_extensions.pyi
dcreager: > I just wanted to suggest making this a method on `ConstraintSet` but that doesn't work because you also want to allow `bool` as `constraints`.

Yeah, I thought the same thing, but our `KnownFunction` machinery is focused on top-level functions in a module. I had considered adding support for "known methods of a class", but that seemed like it would be out of scope for this PR. I can open an issue to track that.

> I'm not sure I fully understand the semantics here. Is this roughly if and only if `constraints` semantics? If that's the case, I think I'd find `is_subtype_of_iff(ty, of, constraints)` easier to understand.

I can copy over more of the commentary from the mdtest to explain the method better.

crates/ty_python_semantic/src/types/constraints.rs
AlexWaygood: Happy to take a look at this, but might not get to it until Wednesday as I'm busy travelling tomorrow. Don't feel you have to wait for me before merging!

In general, if we're in doubt about how redundancy should behave, the safe thing to do in that situation is always to treat it like subtyping

crates/ty_vendored/ty_extensions/ty_extensions.pyi
dcreager: Done. I put the expanded documentation on the Rust side, to be consistent with where we are documenting the other typing internals.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: Thanks! This isn't hooked up to anything yet except isolated mdtests, so no urgency.

dcreager: (But I will still probably merge this and address anything you might find in a follow-on PR)

crates/ty_vendored/ty_extensions/ty_extensions.pyi
dcreager: > Yeah, I thought the same thing, but our `KnownFunction` machinery is focused on top-level functions in a module. I had considered adding support for "known methods of a class", but that seemed like it would be out of scope for this PR. I can open an issue to track that.

https://github.com/astral-sh/ty/issues/1447

PR #21011, opened 2025-10-21
[`pydoclint`] Fix false positive on explicit exception re-raising (`DOC501`, `DOC502`)

### LINKED ISSUE #20973: DOC502 false positive on `raise e`
### Summary

[`docstring-extraneous-exception` (DOC502)](https://docs.astral.sh/ruff/rules/docstring-extraneous-exception/) reports a false positive when an exception mentioned in the docstring is caught and explicitly reraised (i.e. `raise e` or `raise e from None`, assuming `e` is bound in the `except` clause and not rebound before it is raised). It correctly ignores implicit reraising (i.e. just `raise`) of a documented exception. The upstream implementation in pydoclint gets it right both ways. [Example](https://play.ruff.rs/ffc4e2ea-af0f-4ca4-a56e-ea2c40c373f8):
```console
$ cat >doc502.py <<'# EOF'
def f():
    """Do nothing.

    Raises
    ------
    OSError
        If the OS errors.
    """
    try:
        pass
    except OSError as e:
        raise e
# EOF

$ ruff --isolated check doc502.py --select DOC502 --preview --output-format concise -q
doc502.py:2:5: DOC502 Raised exception is not explicitly raised: `OSError`

$ pydoclint doc502.py
Skipping files that match this pattern: \.git|\.tox
doc502.py
🎉 No violations 🎉
```

### Version

danparizher:
## Summary
Fixes #20973 (`docstring-extraneous-exception`) false positive when exceptions mentioned in docstrings are caught and explicitly re-raised using `raise e` or `raise e from None`.

## Problem Analysis
The DOC502 rule was incorrectly flagging exceptions mentioned in docstrings as "not explicitly raised" when they were actually being explicitly re-raised through exception variables bound in `except` clauses. 

**Root Cause**: The `BodyVisitor` in `check_docstring.rs` only checked for direct exception references (like `raise OSError()`) but didn't recognize when a variable bound to an exception in an `except` clause was being re-raised.

**Example of the bug**:
```python
def f():
    """Do nothing.

    Raises
    ------
    OSError
        If the OS errors.
    """
    try:
        pass
    except OSError as e:
        raise e  # This was incorrectly flagged as not explicitly raising OSError
```

The issue occurred because `resolve_qualified_name(e)` couldn't resolve the variable `e` to a qualified exception name, since `e` is just a variable binding, not a direct reference to an exception class.

## Approach
Modified the `BodyVisitor` in `crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs` to:

1. **Track exception variable bindings**: Added `exception_variables` field to map exception variable names to their exception types within `except` clauses
2. **Enhanced raise statement detection**: Updated `visit_stmt` to check if a `raise` statement uses a variable name that's bound to an exception in the current `except` clause
3. **Proper scope management**: Clear exception variable mappings when leaving `except` handlers to prevent cross-contamination

**Key changes**:
- Added `exception_variables: FxHashMap<&'a str, QualifiedName<'a>>` to track variable-to-exception mappings
- Enhanced `visit_except_handler` to store exception variable bindings when entering `except` clauses
- Modified `visit_stmt` to check for variable-based re-raising: `raise e` → lookup `e` in `exception_variables`
- Clear mappings when exiting `except` handlers to maintain proper scope

### DIFF
--- a/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
+++ b/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
@@ -861,6 +861,20 @@ impl<'a> BodyVisitor<'a> {
             raised_exceptions,
         }
     }
+
+    /// Store `exception` if its qualified name does not correspond to one of the exempt types.
+    fn maybe_store_exception(&mut self, exception: &'a Expr, range: TextRange) {
+        let Some(qualified_name) = self.semantic.resolve_qualified_name(exception) else {
+            return;
+        };
+        if is_exception_or_base_exception(&qualified_name) {
+            return;
+        }
+        self.raised_exceptions.push(ExceptionEntry {
+            qualified_name,
+            range,
+        });
+    }
 }
 
 impl<'a> Visitor<'a> for BodyVisitor<'a> {
@@ -896,55 +910,26 @@ impl<'a> Visitor<'a> for BodyVisitor<'a> {
                             range: exc.range(),
                         });
                     } else if let ast::Expr::Name(name) = exc.as_ref() {
-                        // If it's a variable name, check if it's bound to an exception in the current except clause
+                        // If it's a variable name, check if it's bound to an exception in the
+                        // current except clause
                         if let Some(exception_expr) = self.exception_variables.get(name.id.as_str())
                         {
-                            // Use the same logic as currently_suspended_exceptions to handle tuples
-                            let mut maybe_store_exception = |exception| {
-                                let Some(qualified_name) =
-                                    self.semantic.resolve_qualified_name(exception)
-                                else {
-                                    return;
-                                };
-                                if is_exception_or_base_exception(&qualified_name) {
-                                    return;
-                                }
-                                self.raised_exceptions.push(ExceptionEntry {
-                                    qualified_name,
-                                    range: exc.range(),
-                                });
-                            };
-
                             if let ast::Expr::Tuple(tuple) = exception_expr {
                                 for exception in tuple {
-                                    maybe_store_exception(exception);
+                                    self.maybe_store_exception(exception, stmt.range());
                                 }
                             } else {
-                                maybe_store_exception(exception_expr);
+                                self.maybe_store_exception(exception_expr, stmt.range());
                             }
                         }
                     }
                 } else if let Some(exceptions) = self.currently_suspended_exceptions {
-                    let mut maybe_store_exception = |exception| {
-                        let Some(qualified_name) = self.semantic.resolve_qualified_name(exception)
-                        else {
-                            return;
-                        };
-                        if is_exception_or_base_exception(&qualified_name) {
-                            return;
-                        }
-                        self.raised_exceptions.push(ExceptionEntry {
-                            qualified_name,
-                            range: stmt.range(),
-                        });
-                    };
-
                     if let ast::Expr::Tuple(tuple) = exceptions {
                         for exception in tuple {
-                            maybe_store_exception(exception);
+                            self.maybe_store_exception(exception, stmt.range());
                         }
                     } else {
-                        maybe_store_exception(exceptions);
+                        self.maybe_store_exception(exceptions, stmt.range());
                     }
                 }
             }


danparizher:
The +109 DOC501 violations represent improved detection accuracy rather than regressions.

From what I'm seeing, the fix correctly identifies cases where exceptions are explicitly re-raised through variables like `raise e` but weren't being detected before.

Previously, DOC501 missed these patterns because it couldn't resolve the variable `e` to an exception type, but now it properly recognizes that `raise e` is indeed raising the exception that `e` is bound to.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -863,21 +867,55 @@ impl<'a> Visitor<'a> for BodyVisitor<'a> {
...
+                };
+
+                if let ast::Expr::Tuple(tuple) = exceptions {
+                    // For tuple exceptions, we'll store the first one
+                    if let Some(first_exception) = tuple.elts.first() {
+                        store_exception_variable(first_exception);
+                    }
+                } else {
+                    store_exception_variable(exceptions);
+                }
```
ntBre: Why would we only store the first element in a tuple of exceptions? It seems like we should store all of them to me.

crates/ruff_linter/src/rules/pydoclint/snapshots/ruff_linter__rules__pydoclint__tests__docstring-missing-exception_DOC501_google.py.snap
```diff
@@ -61,6 +61,46 @@ DOC501 Raised exception `FasterThanLightError` missing from docstring
...
+74 | |         time: Time spent traveling.
+75 | |
+76 | |     Returns:
+77 | |         Speed as distance divided by time.
+78 | |     """
+   | |_______^
+79 |       try:
+80 |           return distance / time
+   |
+help: Add `ZeroDivisionError` to the docstring
```
ntBre, line 82: We shouldn't do this in this PR, but this looks like a nice place for  a secondary annotation showing where the `ZeroDivisionError` is raised.

crates/ruff_linter/resources/test/fixtures/pydoclint/DOC502_google.py
```diff
@@ -81,3 +81,33 @@ def calculate_speed(distance: float, time: float) -> float:
...
+
+    Raises
+    ------
+    OSError
+        If the OS errors.
+    """
+    try:
+        pass
+    except OSError as e:
+        raise e from None
```
ntBre, line 109: I think it would be good to include an example or two of a tuple of exceptions too.

crates/ruff_linter/resources/test/fixtures/pydoclint/DOC502_google.py
ntBre, line 1: These tests don't actually trigger DOC502 on main because they're numpy-style docstrings, not Google style. We need to either rewrite them to be Google style or move them to the DOC502_numpy.py file.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -826,6 +829,7 @@ impl<'a> BodyVisitor<'a> {
             currently_suspended_exceptions: None,
             raised_exceptions: Vec::new(),
             semantic,
+            exception_variables: FxHashMap::default(),
```
ntBre, line 832: I was wondering if we could avoid collecting these here and instead try to resolve the name down below,  but I tried it locally, and it seems that we're not actually in this `Scope` in the `SemanticModel` during this check, so this makes sense to me.

crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
```diff
@@ -863,21 +867,55 @@ impl<'a> Visitor<'a> for BodyVisitor<'a> {
...
         self.currently_suspended_exceptions = handler_inner.type_.as_deref();
+
+        // Track exception variable bindings
+        if let Some(name) = handler_inner.name.as_ref() {
+            if let Some(exceptions) = self.currently_suspended_exceptions {
+                let mut store_exception_variable = |exception| {
+                    if let Some(qualified_name) = self.semantic.resolve_qualified_name(exception) {
+                        self.exception_variables
+                            .insert(name.id.as_str(), qualified_name);
+                    }
```
ntBre: I think it might make sense to defer this as resolving qualified names can be somewhat expensive. Then we might be able to reuse the `maybe_store_exception` closure below. In other words, just store the `name.id.as_str()` and `exceptions` in the map here.

That will also help to resolve my other comment about tuples because we can just store the name -> tuple mapping and handle the tuple like the `currently_suspended_exceptions` case too.

ntBre (approved):
Thanks! This looks good. I pushed one commit factoring out the closure into a method instead of duplicating it. I'll just make sure I didn't break anything in the ecosystem report and then merge this.

PR #21012, opened 2025-10-21
[ty] Teach the resolver that vendored copies of typeshed in the ty cache are in fact typeshed

### LINKED ISSUE #1054: goto-definition on vendored typeshed definitions doesn't work
This is a followup to #1004. 

It's exactly the same issue but specifically for when goto-declaration opens up a copy of our vendored typeshed in your IDE.

```python
from warnings import deprecated
```

If you goto-declaration on "deprecated" it will open up our vendored typeshed in the IDE.
If you goto-definition on "deprecated" it will open up the actual stdlib in the IDE.
...but if you goto-declaration on *the vendored typeshed definition* it won't go anywhere.

From initial debugging it seems like file_to_module doesn't recognize the opened file, and so stub mapping fails. I assume there's some trickery going on with vendored typeshed files that makes this file a different one from the one used normally in our stdlib analysis.

Gankra:
Fixes https://github.com/astral-sh/ty/issues/1054

I've confirmed it works on my machine in vscode, but it's not clear to me if there's a good way to test this in our test suite.

### DIFF
--- a/crates/ty_python_semantic/src/module_resolver/path.rs
+++ b/crates/ty_python_semantic/src/module_resolver/path.rs
@@ -607,7 +607,11 @@ impl SearchPath {
     }
 
     #[must_use]
-    pub(crate) fn relativize_system_path(&self, path: &SystemPath) -> Option<ModulePath> {
+    pub(crate) fn relativize_system_path(
+        &self,
+        db: &dyn Db,
+        path: &SystemPath,
+    ) -> Option<ModulePath> {
         if path
             .extension()
             .is_some_and(|extension| !self.is_valid_extension(extension))
@@ -629,7 +633,15 @@ impl SearchPath {
                         relative_path: relative_path.as_utf8_path().to_path_buf(),
                     })
             }
-            SearchPathInner::StandardLibraryVendored(_) => None,
+            // Check if this system path is actually a cached vendored path
+            SearchPathInner::StandardLibraryVendored(search_path) => path
+                .as_utf8_path()
+                .strip_prefix(cached_vendored_path(db, search_path)?.as_utf8_path())
+                .ok()
+                .map(|relative_path| ModulePath {
+                    search_path: self.clone(),
+                    relative_path: relative_path.to_path_buf(),
+                }),
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/ty_python_semantic/src/module_resolver/path.rs
```diff
@@ -721,6 +733,38 @@ impl SearchPath {
...
+        path.as_str()
+    ))
+}
+
+/// Get the full `SystemPath` in the cache that a `VendoredPath` refers to
+pub(crate) fn cached_vendored_path(db: &dyn Db, path: &VendoredPath) -> Option<SystemPathBuf> {
+    let writable = db.system().as_writable()?;
+    let system_path = vendored_path_for_cache(path);
+    Some(writable.cache_dir()?.join(system_path))
+}
```
Gankra, line 751: We invoke these *a lot* now, so they might need to be cached...

crates/ty_python_semantic/src/module_resolver/path.rs
```diff
@@ -721,6 +733,38 @@ impl SearchPath {
     }
 }
 
+/// Get the cache-relative `SystemPath` that a `VendoredPath` refers to
+pub fn vendored_path_for_cache(path: &VendoredPath) -> SystemPathBuf {
```
MichaReiser, line 737: Hmm. I'd prefer if we can keep this logic in `ty_server`, if possible because not all clients have a cache location or they use a different cache location (e.g. playground)

MichaReiser (commented):
Could we achive the same if the LSP registered an additional search path for the cached vendored files (that comes after the standard library?) That would require less custom casing for which we have not tests for. 

Another alternative would be to remap cached vendored file path to  the corresponding vendored path within the LSP request handlers. 

Could any of those solution lead to problems if the content of the cached vendored file and the vendored file diverge (e.g. because the user starts to make edits?)

Gankra:
> Could we achive the same if the LSP registered an additional search path for the cached vendored files (that comes after the standard library?) That would require less custom casing for which we have not tests for.

The additional search-path still requires something like `path_is_equivalent` logic to not get rejected on the `file.path() == module.file().path()` logic. I'd also be worried that "the stdlib can't be shadowed" rules might get in the way? But probably it would do the right thing for this code specifically so maybe it's fine?

> Another alternative would be to remap cached vendored file path to the corresponding vendored path within the LSP request handlers.

Hmm that was concerning to me as an approach earlier but maybe it's not actually that bad? We already do that mapping in the other direction. I'll give it a try.

PR #21019, opened 2025-10-21
[ty] Timeout based workspace diagnostic progress reports

MichaReiser:
## Summary

This PR changes the workspace diagnostic's progress report (how many files were checked) from sending
a report for every 100th file to sending a report every 50ms. It makes ty feel snappier. 

The only motivation for having a timeout in the first place is to not overwhelm the client
(and not spend more time serializing messages than checking files ;) )

## Test Plan

https://github.com/user-attachments/assets/0c7513bb-8280-4196-9ca0-60fc6517a1b1

### DIFF
--- a/crates/ty_server/src/server/api/requests/workspace_diagnostic.rs
+++ b/crates/ty_server/src/server/api/requests/workspace_diagnostic.rs
@@ -23,6 +23,7 @@ use ruff_db::files::File;
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
+use std::sync::Mutex;
 use std::time::{Duration, Instant};
 use ty_project::{Db, ProgressReporter};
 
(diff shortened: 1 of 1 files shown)


crates/ty_server/src/server/api/requests/workspace_diagnostic.rs
```diff
@@ -199,77 +198,64 @@ impl RetriableRequestHandler for WorkspaceDiagnosticRequestHandler {
 ///
 /// Diagnostics are only streamed if the client sends a partial result token.
 struct WorkspaceDiagnosticsProgressReporter<'a> {
-    total_files: usize,
-    checked_files: AtomicUsize,
     work_done: LazyWorkDoneProgress,
-    response: std::sync::Mutex<ResponseWriter<'a>>,
+    state: std::sync::Mutex<ProgressReporterState<'a>>,
```
dcreager: nit: Can we import `std::sync::Mutex` instead of fully qualifying here and below?

crates/ty_server/src/server/api/requests/workspace_diagnostic.rs
```diff
@@ -199,77 +198,64 @@ impl RetriableRequestHandler for WorkspaceDiagnosticRequestHandler {
...
+        let Ok(mut state) = self.state.lock() else {
             return;
         };
 
+        state.checked_files += 1;
+
+        if state.checked_files == state.total_files {
+            state.report_progress(&self.work_done);
+        } else if state.last_response_sent.elapsed() >= Duration::from_millis(50) {
+            state.last_response_sent = Instant::now();
```
dcreager, line 245: very optional nit: For this use case, it doesn't matter all that much, but this will cause the report times to creep by a bit, instead of being precisely every 50ms.


```suggestion
            state.last_response_sent += Duration::from_millis(50);
```

(That does make the name of the field slightly incorrect, which you could fix by calling it `next_report_time` instead and comparing it against `now`)

MichaReiser: This can also lead to drifts if `reprot_file` isn't called for a long time because `Duration::from_millis` might be shorter than the time between two report_file calls. If that happens, then we'll constantly be "behind" and send a report on every future `report_file` call until it eventually catches up again with `now`.

PR #21020, opened 2025-10-21
[ty] Support goto-definition on vendored typeshed stubs

Gankra:
This is an alternative to #21012 that more narrowly handles this logic in the stub-mapping machinery rather than pervasively allowing us to identify cached files as typeshed stubs. Much of the logic is the same (pulling the logic out of ty_server so it can be reused).

I don't have a good sense for if one approach is "better" or "worse" in terms of like, semantics and Weird Bugs that this can cause. This one is just "less spooky in its broad consequences" and "less muddying of separation of concerns" and puts the extra logic on a much colder path. I won't be surprised if one day the previous implementation needs to be revisited for its more sweeping effects but for now this is good.

Fixes https://github.com/astral-sh/ty/issues/1054

### DIFF
--- a/crates/ty_ide/src/lib.rs
+++ b/crates/ty_ide/src/lib.rs
@@ -45,7 +45,11 @@ pub use signature_help::{ParameterDetails, SignatureDetails, SignatureHelpInfo, 
 pub use symbols::{FlatSymbols, HierarchicalSymbols, SymbolId, SymbolInfo, SymbolKind};
 pub use workspace_symbols::{WorkspaceSymbolInfo, workspace_symbols};
 
-use ruff_db::files::{File, FileRange};
+use ruff_db::{
+    files::{File, FileRange},
+    system::SystemPathBuf,
+    vendored::VendoredPath,
+};
 use ruff_text_size::{Ranged, TextRange};
 use rustc_hash::FxHashSet;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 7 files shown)


crates/ty_ide/src/lib.rs
```diff
@@ -287,6 +291,38 @@ impl HasNavigationTargets for TypeDefinition<'_> {
     }
 }
 
+/// Get the cache-relative path where vendored paths should be written to.
+pub fn relative_cached_vendored_root() -> SystemPathBuf {
+    // The vendored files are uniquely identified by the source commit.
+    SystemPathBuf::from(format!("vendored/typeshed/{}", ty_vendored::SOURCE_COMMIT))
+}
```
Gankra, line 298: This is morally a const fn but I didn't want to deal with making that work and it's not a huge deal.

MichaReiser: haha, I wanted something similar recently too and did draw the very same conclusion: I don't want to deal with this right now :)

crates/ty_ide/src/lib.rs
```diff
@@ -287,6 +291,38 @@ impl HasNavigationTargets for TypeDefinition<'_> {
     }
 }
 
+/// Get the cache-relative path where vendored paths should be written to.
+pub fn relative_cached_vendored_root() -> SystemPathBuf {
+    // The vendored files are uniquely identified by the source commit.
+    SystemPathBuf::from(format!("vendored/typeshed/{}", ty_vendored::SOURCE_COMMIT))
```
MichaReiser, line 297:
```suggestion
    SystemPath::from(concat!("vendored/typeshed", ty_vendored::SOURCE_COMMIT))
```

Gankra: Unfortunately `concat` concatenates only literals so this is a compiler error :(

Issue #15655, opened 2025-01-21
Auto-generate more of the AST representation


dcreager:
In #15544 we added a [script](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_ast/generate.py) to [auto-generate](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_ast/src/generated.rs) large parts of the Rust data model that we use to store the AST of parsed Python code.  That script consumes a [TOML file](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_ast/ast.toml), which describes all of the possible syntax nodes (e.g., `StmtIf`, `ExprBinOp`) and any groups those nodes belong to (e.g. `Stmt`, `Expr`).  The details of each syntax node are still [defined manually](https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_ast/src/nodes.rs) in Rust.

We could go further with auto-generation, with existing art that we could build on.  rust-analyzer uses [ungrammar](https://rust-analyzer.github.io//blog/2020/10/24/introducing-ungrammar.html), while Python itself uses [ASDL](https://www.oilshell.org/blog/2016/12/11.html) ([asdl](https://github.com/python/cpython/blob/main/Parser/Python.asdl), [parser](https://github.com/python/cpython/blob/main/Parser/asdl.py), [codegen](https://github.com/python/cpython/blob/main/Parser/asdl_c.py)).  This would eliminate even more tedious hand-written Rust code — not just the struct/enum definitions themselves, but even things like the [`visit_source_order`](https://github.com/astral-sh/ruff/blob/ef85c682bdd76ed63457bdfbec72e09424ac20ab/crates/ruff_python_ast/src/node.rs#L18-L24) methods for each syntax node.  It would also allow us to experiment more easily with other internal representations for the parsed AST — such as using `IndexVec` to store the syntax node content (as alluded to in https://github.com/astral-sh/ruff/pull/12419#issuecomment-2262727929).

Glyphack:
Hey, I'd like to help with this and the linked issue.
First I went through the original PR and tried resolving one of the comments to see how the code generation works(https://github.com/astral-sh/ruff/pull/16144).

I'll look into other files and see what other stuff can we auto generate with the current information.

I want to continue with the implementation, I appreciate your help with a few questions:

[...]

PR #21024, opened 2025-10-21
[`ruff`] Auto generate ast Pattern nodes

TaKO8Ki:
## Summary

Fixes a part of #15655

Auto generate ast Pattern nodes

## Test Plan

I have fixed some syntax tests. All other tests should pass without any changes. I have confirmed the generated code to make sure it's the same as old code.

### DIFF
--- a/crates/ruff_python_ast/src/nodes.rs
+++ b/crates/ruff_python_ast/src/nodes.rs
@@ -2848,10 +2848,10 @@ pub enum IrrefutablePatternKind {
     Wildcard,
 }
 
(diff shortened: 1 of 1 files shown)


TaKO8Ki:
I have fixed some snapshots in `ruff_python_parser/tests`. If we change the order of node_index and range in `generate.py`, It's unnecessary. However, other definitions such as `StmtExpr` is already based on the fields order, so I will keep the following order.

```rust
pub node_index: crate::AtomicNodeIndex,
pub range: ruff_text_size::TextRange,
```

PR #21032, opened 2025-10-22
[syntax-error]: no binding for nonlocal  PLE0117 as a semantic syntax error

11happy:
## Summary

This PR ports PLE0117 as a semantic syntax error.

## Test Plan

Tests previously written

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -2698,7 +2698,7 @@ impl SemanticSyntaxContext for SemanticIndexBuilder<'_, '_> {
     }
 
     // We handle the one syntax error that relies on this method (`NonlocalWithoutBinding`) directly
-    // in `infer_nonlocal_statement`, so this just returns `true`.
+    // in `TypeInferenceBuilder::infer_nonlocal_statement`, so this just returns `true`.
     fn has_nonlocal_binding(&self, _name: &str) -> bool {
         true
     }


11happy:
@ntBre its ready for review : )

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1154,6 +1166,9 @@ impl Display for SemanticSyntaxError {
             SemanticSyntaxErrorKind::DifferentMatchPatternBindings => {
                 write!(f, "alternative patterns bind different names")
             }
+            SemanticSyntaxErrorKind::NonlocalWithoutBinding(name) => {
+                write!(f, "Nonlocal name `{name}` found without binding")
```
ntBre: It looks like most of our other error messages start with a lowercase letter. I might just stick with the CPython error message here:

```pycon
>>> def f():
...     nonlocal x
...
  File "<python-input-0>", line 2
    nonlocal x
    ^^^^^^^^^^
SyntaxError: no binding for nonlocal 'x' found
```

```suggestion
                write!(f, "no binding for nonlocal `{name}` found")
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -219,21 +219,33 @@ impl SemanticSyntaxChecker {
...
-                //     nonlocal x
-
+            Stmt::Nonlocal(ast::StmtNonlocal { names, range, .. }) => {
                 // test_err nonlocal_declaration_at_module_level
                 // nonlocal x
                 // nonlocal x, y
+
+                // test_ok nonlocal_declaration_at_module_level
+                // def _():
+                //     nonlocal x
```
ntBre: I'd probably revert this change since it just moves the test down, right?

crates/ruff_python_parser/src/parser/statement.rs
```diff
@@ -934,7 +934,6 @@ impl<'src> Parser<'src> {
                 self.current_token_range(),
             );
         }
-
```
ntBre, line 937: I think we could also revert this whitespace change.

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -641,6 +642,12 @@ impl SemanticSyntaxContext for Checker<'_> {
         self.semantic.global(name)
     }
 
+    fn nonlocal(&self, name: &str) -> Option<TextRange> {
+        self.semantic
+            .nonlocal(name)
+            .map(|(_scope_id, binding_id)| self.semantic.binding(binding_id).range)
+    }
```
ntBre: What do you think about just making this function something like `has_nonlocal_binding` that returns a bool? It doesn't look like we're using the returned range anywhere, and it might be easier for ty to implement `has_nonlocal_binding` (or at least the stub version is simpler).

crates/ruff_python_parser/src/semantic_errors.rs

crates/ruff_python_parser/src/parser/statement.rs
11happy: done

crates/ruff_linter/src/checkers/ast/mod.rs
11happy: Yes, we can change it bool as the range is not used anywhere rightnow.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -2697,6 +2697,12 @@ impl SemanticSyntaxContext for SemanticIndexBuilder<'_, '_> {
         None
     }
 
+    // We handle the one syntax error that relies on this method (`NonlocalWithoutBinding`) directly
+    // in `infer_nonlocal_statement`, so this just returns `true`.
```
AlexWaygood:
```suggestion
    // in `TypeInferenceBuilder::infer_nonlocal_statement`, so this just returns `true`.
```

PR #21034, opened 2025-10-22
[`ISC001`] fix panic when string literals are unclosed

### LINKED ISSUE #21023: ISC001 panic: begin <= end (1 <= 0) when slicing `'`
### Summary

`ruff check --select=ISC001 s.py` panics with the following code:

**s.py**
```python
'' '
```

Output:
```
panic: Panicked at crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs:200:25 when checking `/tmp/ruff/s.py`: `begin <= end (1 <= 0) when slicing `'``
--> s.py:1:1
```

Version info:
- Python 3.13.7 on Arch Linux

### Version

ruff 0.14.1

TaKO8Ki:
## Summary

Fixes #21023

Return None from trailing_quote for single-quote tokens

## Test Plan

I created `crates/ruff_linter/resources/test/fixtures/flake8_implicit_str_concat/ISC_syntax_error_2.py`.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs
+++ b/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs
@@ -1,14 +1,12 @@
 use std::borrow::Cow;
 
 use itertools::Itertools;
-
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::StringFlags;
-use ruff_python_ast::str::{leading_quote, trailing_quote};
 use ruff_python_index::Indexer;
 use ruff_python_parser::{Token, TokenKind, Tokens};
 use ruff_source_file::LineRanges;
-use ruff_text_size::{Ranged, TextRange};
+use ruff_text_size::{Ranged, TextLen, TextRange};
 
 use crate::Locator;
 use crate::checkers::ast::LintContext;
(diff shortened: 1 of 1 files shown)


crates/ruff_python_ast/src/str.rs
```diff
@@ -307,6 +307,12 @@ pub fn leading_quote(content: &str) -> Option<&str> {
 
 /// Return the trailing quote string for a string, template, or bytes literal (e.g., `"""`).
 pub fn trailing_quote(content: &str) -> Option<&str> {
+    // Tokens like "'" or "\"" represent unterminated literals that end with a single quote
+    // character, so they can't provide both the opening and closing delimiter.
```
MichaReiser: I'll have a closer look tomorrow but I wonder if we could change the rule instead to get the quote style from the token flags (and/or not call this function if the string token is unterminated)

TaKO8Ki: I guess it's possible. The problem is `concatenate_strings` that calls `trailing_quote`, so If we check the string flags `a_token` and `b_token` and avoid calling `concatenate_strings` when there’s an issue, it will resolve the problem.

https://github.com/astral-sh/ruff/blob/735c3e087e6a31dd45e8507007a6743348a94113/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/implicit.rs#L170

MichaReiser: I think I'd change `concatenate_strings` to take a `Token` instead and make it return `None` if `a_range.string_flags()?` is `None`. I'm also leaning towards not providing a fix when either of the strings are uncloded, but I could be convinced that we should still provide a fix.

Using `StringFlags` has the benefit that it also works for triple quoted strings (where the `content.len() <= 1 check isn't sufficient)

TaKO8Ki: I updated `concatenate_strings` to validate `token.string_flags`.

PR #21035, opened 2025-10-22
Document when a rule was added

ntBre:
Summary
--

Inspired by #20859, this PR adds the version a rule was added, and the file and line where it was defined, to `ViolationMetadata`. The file and line just use the standard `file!` and `line!` macros, while the more interesting version field uses a new `violation_metadata` attribute parsed by our `ViolationMetadata` derive macro.

I moved the commit modifying all of the rule files to the end, so it should be a lot easier to review by omitting that one.

As a curiosity and a bit of a sanity check, I also plotted the rule numbers over time:

image

I think this looks pretty reasonable and avoids some of the artifacts the earlier versions of the script ran into, such as the `rule` sub-command not being available or `--explain` requiring a file argument.

<details><summary>Script and summary data</summary>

```shell
gawk --csv '
NR > 1 {
    split($2, a, ".")
    major = a[1]; minor = a[2]; micro = a[3]
    # sum the number of rules added per minor version
    versions[minor] += 1
}
END {
    tot = 0
    for (i = 0; i <= 14; i++) {
        tot += versions[i]
        print i, tot
    }
}
' ruff_rules_metadata.csv > summary.dat
```

```
0 696
1 768
2 778
3 803
4 822
5 848
6 855
7 865
8 893
9 915
10 916
11 924
12 929
13 932
14 933
```

</details>

Test Plan
--

I built and viewed the documentation locally, and it looks pretty good!

image

The spacing seems a bit awkward following the `h1` at the top, so I'm wondering if this might look nicer as a footer in Ruff. The links work well too:
- [v0.0.271](https://github.com/astral-sh/ruff/releases/tag/v0.0.271)
- [Related issues](https://github.com/astral-sh/ruff/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20airflow-variable-name-task-id-mismatch)
- [View source](https://github.com/astral-sh/ruff/blob/main/crates%2Fruff_linter%2Fsrc%2Frules%2Fairflow%2Frules%2Ftask_variable_name.rs#L34)

The last one even works on `main` now since it points to the `derive(ViolationMetadata)` line.

In terms of binary size, this branch is a bit bigger than main with 38,654,520 bytes compared to 38,635,728 (+20 KB). I guess that's not _too_ much of an increase, but I wanted to check since we're generating a lot more code with macros.

### DIFF
--- a/crates/ruff_linter/src/rules/airflow/rules/dag_schedule_argument.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/dag_schedule_argument.rs
@@ -41,7 +41,7 @@ use crate::checkers::ast::Checker;
 /// dag = DAG(dag_id="my_dag", schedule=timedelta(days=1))
 /// ```
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.11.3")]
+#[violation_metadata(stable_since = "0.13.0")]
 pub(crate) struct AirflowDagNoScheduleArgument;
 
 impl Violation for AirflowDagNoScheduleArgument {
--- a/crates/ruff_linter/src/rules/airflow/rules/moved_to_provider_in_3.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/moved_to_provider_in_3.rs
@@ -35,7 +35,7 @@ use crate::{FixAvailability, Violation};
 /// fab_auth_manager_app = FabAuthManager().get_fastapi_app()
 /// ```
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.8.2")]
+#[violation_metadata(stable_since = "0.13.0")]
 pub(crate) struct Airflow3MovedToProvider<'a> {
     deprecated: QualifiedName<'a>,
     replacement: ProviderReplacement,
--- a/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
@@ -41,7 +41,7 @@ use ruff_text_size::TextRange;
 /// yesterday = today - timedelta(days=1)
 /// ```
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.8.1")]
+#[violation_metadata(stable_since = "0.13.0")]
 pub(crate) struct Airflow3Removal {
     deprecated: String,
     replacement: Replacement,
--- a/crates/ruff_linter/src/rules/airflow/rules/suggested_to_move_to_provider_in_3.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/suggested_to_move_to_provider_in_3.rs
@@ -51,7 +51,7 @@ use ruff_text_size::TextRange;
 /// )
 /// ```
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.11.5")]
+#[violation_metadata(stable_since = "0.13.0")]
 pub(crate) struct Airflow3SuggestedToMoveToProvider<'a> {
     deprecated: QualifiedName<'a>,
     replacement: ProviderReplacement,
--- a/crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/suggested_to_update_3_0.rs
@@ -37,7 +37,7 @@ use ruff_text_size::TextRange;
 /// Asset(uri="test://test/")
 /// ```
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.11.6")]
+#[violation_metadata(stable_since = "0.13.0")]
 pub(crate) struct Airflow3SuggestedUpdate {
     deprecated: String,
     replacement: Replacement,
--- a/crates/ruff_linter/src/rules/fastapi/rules/fastapi_non_annotated_dependency.rs
+++ b/crates/ruff_linter/src/rules/fastapi/rules/fastapi_non_annotated_dependency.rs
@@ -79,7 +79,7 @@ use ruff_python_ast::PythonVersion;
 /// [typing-annotated]: https://docs.python.org/3/library/typing.html#typing.Annotated
 /// [typing-extensions]: https://typing-extensions.readthedocs.io/en/stable/
 #[derive(ViolationMetadata)]
-#[violation_metadata(stable_since = "0.5.5")]
+#[violation_metadata(stable_since = "0.8.0")]
 pub(crate) struct FastApiNonAnnotatedDependency {
     py_version: PythonVersion,
 }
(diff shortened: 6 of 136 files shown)


MeGaGiGaGon:
That's fun, I wasn't aware `ruff rule` existed, that would have made my other two previous adventures with iterating over all the rules much easier. I'll have to go back at some point and update my scripts for those, since I may have missed a couple rules while regexing my way through the problems.

Related to that, this may be a stupid question, but will the changes in this PR be reflected in `ruff rule`? (As in, after this will `ruff rule --all --output-format json` also include items for version, file, and line?) That would be even more convenient for my purposes, since I wouldn't need to do any of the regex spaghetti to get rule definition location info in the first place.

crates/ruff_dev/src/generate_docs.rs
```diff
@@ -31,6 +31,25 @@ pub(crate) fn main(args: &Args) -> Result<()> {
...
+                version = rule.version().unwrap_or(env!("CARGO_PKG_VERSION")),
+                encoded_name =
+                    url::form_urlencoded::byte_serialize(rule.name().as_str().as_bytes())
+                        .collect::<String>(),
+                file =
+                    url::form_urlencoded::byte_serialize(rule.file().replace('\\', "/").as_bytes())
+                        .collect::<String>(),
+                line = rule.line(),
+            );
+
```
MeGaGiGaGon: Apply with caution, completely untested.

Two suggestions on the issues link:
- Surround the rule name with quotes (`%27` is `'`), since otherwise GitHub doesn't treat it as a single word, and a lot of extra results appear
- Use the new-ish `OR` operator + parenthesis to search for both the rule name and code at the same time, since some issues include only one of the two.

As an example with a rule that has less, more common words: The [generated link](https://github.com/astral-sh/ruff/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20non-self-return-type) for `non-self-return-type` gives 16 open and 22 closed, many of which are not related, and do not mention the rule name at all. This [new link](https://github.com/astral-sh/ruff/issues?q=is%3Aissue%20state%3Aopen%20(%27non-self-return-type%27%20OR%20PYI034)) only gives 3 open and 8 closed, all of which at least mention the rule.

```suggestion
            let _ = writeln!(
                &mut output,
                r#"<small>
Added in <a href="https://github.com/astral-sh/ruff/releases/tag/{version}">{version}</a> ·
<a href="https://github.com/astral-sh/ruff/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20(%27{encoded_name}%27%20OR%20{rule_code})" target="_blank">Related issues</a> ·
<a href="https://github.com/astral-sh/ruff/blob/main/{file}#L{line}" target="_blank">View source</a>
</small>

"#,
                version = rule.version().unwrap_or(env!("CARGO_PKG_VERSION")),
                encoded_name =
                    url::form_urlencoded::byte_serialize(rule.name().as_str().as_bytes())
                        .collect::<String>(),
                rule_code = rule.noqa_code(),
                file =
                    url::form_urlencoded::byte_serialize(rule.file().replace('\\', "/").as_bytes())
                        .collect::<String>(),
                line = rule.line(),
            );
```

MichaReiser (commented):
Nice! That's what you were up to yesterday :)

`ViolationMetadata` seems the right spot for `file` and `location` but I'm less sure about the version. Mainly because the rule status is now spread over two places:

* `code_to_rule`: Defines the status of the rule (preview, removed, stable)
* `violation_metadata`: Defines the version. 

This has the downside that It's unclear what the version represents. Is it when the rule was added, deprecated or removed?

That's why I'm inclined to combine the version and the `RuleStatus`, similar to `LintStatus` in ty. Whether we simply extend the `RuleStatus` variants to have extra fields or move `RuleStatus` to `ViolationMetadata` (that's probably what I'd do if I would do it anew) isn't important and I'd probably go with what's easier.

crates/ruff_dev/src/generate_docs.rs
ntBre: That seems like a good idea, thank you!

ntBre:
> Related to that, this may be a stupid question, but will the changes in this PR be reflected in `ruff rule`? (As in, after this will `ruff rule --all --output-format json` also include items for version, file, and line?) That would be even more convenient for my purposes, since I wouldn't need to do any of the regex spaghetti to get rule definition location info in the first place.

@MeGaGiGaGon I think that's a great question and a good idea to add. It crossed my mind early on but then I kind of forgot about it. I think it should be straightforward to add here.

@MichaReiser that makes a lot of sense. I did like how `LintStatus` looked in ty. I'll try to see how it looks to move `RuleGroup` into the new macro attribute this morning, but I think adding the version as a field is probably easier.

ntBre:
As Micha and I discussed on Discord, the `RuleGroup` and version information have now been combined, more like `LintStatus` in ty. This is an example usage of the new attribute:

```rust
#[derive(ViolationMetadata)]
#[violation_metadata(stable_since = "v0.0.155")]
pub(crate) struct UnusedNOQA {
    pub codes: Option<UnusedCodes>,
}
```

One interesting and somewhat intentional aspect of the attribute is that the macro will preserve the last entry, so we could theoretically stack versions if the status of a rule changes over time. For example, this would be accepted by the current macro implementation but only use the `removed_since` field:

```rust
#[derive(ViolationMetadata)]
#[violation_metadata(
    preview_since = "v0.0.123", 
    stable_since = "v0.0.155", 
    deprecated_since = "v1.2.3", 
    removed_since = "v4.5.6",
)]
pub(crate) struct UnusedNOQA {
    pub codes: Option<UnusedCodes>,
}
```

Another alternative we considered was just using the `RuleGroup` variants directly, like:

```rust
#[derive(ViolationMetadata)]
#[violation_metadata(status = RuleGroup::Stable { since: "v0.0.155" })]
pub(crate) struct UnusedNOQA {
    pub codes: Option<UnusedCodes>,
}
```

which is also pretty nice. If we ever decide to  add a `reason` field like ty has for the deprecated and removed variants this might be even more compelling. Or we could just add a `reason` field to the macro too.

<hr>

I also updated the JSON output format for `ruff rule` to include the new file and line information (the version was added automatically by including it in `RuleGroup`).

And I rebased a bit more to keep the big commits at the end. The last commit is still `add metadata to rules` and the second-to-last is the big edit to remove the `RuleGroup` entries from the `map_codes` table.

ntBre:
Sounds good! I guess I'll go ahead and land this once I fix the doctests and remove the scripts and CSV since it should be easy to iterate on the syntax now that we have the infrastructure in place.

Maybe this is just my editor, but I wasn't getting completions when I tried typing `RuleGroup::` in the attribute. But otherwise I like that fine too, no strong feelings on the exact syntax.

ntBre:
One last check to make sure my scripts didn't change any of the rule statuses:

```shell
diff \
    <(./target/debug/ruff rule --all --output-format json | jq '[ .[] | { code, status: ( .status | keys[0] ) } ]') \
    <(../worktrees/ruff1/target/debug/ruff rule --all --output-format json | jq '[ .[] | { code, status} ]')
```

passes cleanly with main checked out in my other worktree.

ntBre:
Oh, I just realized that many of the versions are actually wrong after the `RuleGroup` change... I collected all of the `Added in` versions but now they need to be when they were stabilized or removed for many rules instead.

ntBre:
All of the `removed_since` and `stable_since` versions are updated. Preview should have already been correct, and we don't currently have any deprecated rules. I'll merge once CI finishes and then start the release for the day. Thanks again!

PR #21038, opened 2025-10-23
[ty] Fix rare multithreaded related hang

MichaReiser:
## Summary

This PR updates salsa to pull in https://github.com/salsa-rs/salsa/pull/1010 which fixes two cases where Salsa hangs because it enters an infinite loop when multiple threads are involved in a single fixpoint iteration. 

The update also pulls in https://github.com/salsa-rs/salsa/pull/1012 which a) makes the `cycle_fn` optional if it only returns `Iterate`, allowing us to remove a ton of code and b) exposes the last provisional value and the query id to the `cycle_fn`, which should help with https://github.com/astral-sh/ruff/pull/20566

## Test Plan

Existing tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3563,7 +3563,7 @@ checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
 [[package]]
 name = "salsa"
 version = "0.24.0"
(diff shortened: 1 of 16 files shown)


sharkdp:
> The update also pulls in [salsa-rs/salsa#1012](https://github.com/salsa-rs/salsa/pull/1012) which a) makes the `cycle_fn` optional if it only returns `Iterate`, allowing us to remove a ton of code

ohhh :heart_eyes:

PR #21041, opened 2025-10-23
Configurable "unparse mode" for `ruff_python_codegen::Generator`

ShaharNaveh:
## Summary

At [RustPython](https://github.com/RustPython/RustPython) we want to fully replace https://github.com/RustPython/RustPython/blob/153d0eef51ea109d8a322fd351678847b2fb8fe2/compiler/codegen/src/unparse.rs.

Currently our only regression when using Ruff's unparsing, is forcing tuple parentheses. 

## Test Plan

Added tests

### DIFF
--- a/crates/ruff_python_codegen/src/generator.rs
+++ b/crates/ruff_python_codegen/src/generator.rs
@@ -93,7 +93,7 @@ pub struct Generator<'a> {
     /// The line ending to use.
     line_ending: LineEnding,
     /// Unparsed code style. See [`Mode`] for more info.
-    unparse_mode: Mode,
+    mode: Mode,
     buffer: String,
     indent_depth: usize,
     num_newlines: usize,
@@ -105,7 +105,7 @@ impl<'a> From<&'a Stylist<'a>> for Generator<'a> {
         Self {
             indent: stylist.indentation(),
             line_ending: stylist.line_ending(),
-            unparse_mode: Mode::default(),
+            mode: Mode::default(),
             buffer: String::new(),
             indent_depth: 0,
             num_newlines: 0,
@@ -120,7 +120,7 @@ impl<'a> Generator<'a> {
             // Style preferences.
             indent,
             line_ending,
-            unparse_mode: Mode::Default,
+            mode: Mode::Default,
             // Internal state.
             buffer: String::new(),
             indent_depth: 0,
@@ -131,8 +131,8 @@ impl<'a> Generator<'a> {
 
     /// Sets the mode for code unparsing.
     #[must_use]
-    pub fn with_unparse_mode(mut self, mode: Mode) -> Self {
-        self.unparse_mode = mode;
+    pub fn with_mode(mut self, mode: Mode) -> Self {
+        self.mode = mode;
         self
     }
 
@@ -194,7 +194,7 @@ impl<'a> Generator<'a> {
                 return;
             }
         }
-        let quote_style = self.unparse_mode.quote_style(flags);
+        let quote_style = self.mode.quote_style(flags);
         let escape = AsciiEscape::with_preferred_quote(s, quote_style);
         if let Some(len) = escape.layout().len {
             self.buffer.reserve(len);
@@ -214,7 +214,7 @@ impl<'a> Generator<'a> {
         }
         self.p(flags.prefix().as_str());
 
-        let quote_style = self.unparse_mode.quote_style(flags);
+        let quote_style = self.mode.quote_style(flags);
         let escape = UnicodeEscape::with_preferred_quote(s, quote_style);
         if let Some(len) = escape.layout().len {
             self.buffer.reserve(len);
@@ -1324,7 +1324,7 @@ impl<'a> Generator<'a> {
                 if tuple.is_empty() {
                     self.p("()");
                 } else {
-                    let lvl = match self.unparse_mode {
+                    let lvl = match self.mode {
                         Mode::Default => precedence::TUPLE,
                         Mode::AstUnparse => precedence::MIN,
                     };
@@ -1550,7 +1550,7 @@ impl<'a> Generator<'a> {
             return;
         }
 
-        let quote_style = self.unparse_mode.quote_style(flags);
+        let quote_style = self.mode.quote_style(flags);
         let escape = UnicodeEscape::with_preferred_quote(&s, quote_style);
         if let Some(len) = escape.layout().len {
             self.buffer.reserve(len);
@@ -1577,7 +1577,7 @@ impl<'a> Generator<'a> {
     ) {
         self.p(flags.prefix().as_str());
 
-        let quote_style = self.unparse_mode.quote_style(flags);
+        let quote_style = self.mode.quote_style(flags);
         let flags = flags.with_quote_style(quote_style);
         self.p(flags.quote_str());
         self.unparse_interpolated_string_body(values, flags);
@@ -1637,8 +1637,7 @@ mod tests {
         contents: &str,
     ) -> String {
         let module = parse_module(contents).unwrap();
-        let mut generator =
-            Generator::new(indentation, line_ending).with_unparse_mode(unparse_mode);
+        let mut generator = Generator::new(indentation, line_ending).with_mode(unparse_mode);
         generator.unparse_suite(module.suite());
         generator.generate()
     }


crates/ruff_python_codegen/src/generator.rs
```diff
@@ -1605,16 +1628,18 @@ mod tests {
     }
 
     /// Like [`round_trip`] but configure the [`Generator`] with the requested
-    /// `indentation`, `line_ending` and `preferred_quote` settings.
+    /// `indentation`, `line_ending`, `preferred_quote` and `forced_tuple_parentheses` settings.
     fn round_trip_with(
```
MichaReiser, line 1633: The signature here becomes a bit awkward. Should we introduce a `GeneratorOptions` struct with: 

* `indentation`
* `line_ending`
* `preferred_quote`: Option
* `forced_tuple_parentheses`

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -115,6 +120,18 @@ impl<'a> Generator<'a> {
...
     }
 
+    /// Sets whether tuple expressions should be enclosed in parentheses during unparsing.
+    ///
+    /// When set to true this forces tuple nodes to always include parentheses,
+    /// even in contexts where they might otherwise be optional (e.g., a single-element tuple).
+    ///
+    /// When set to false (the default), parentheses are emitted only when syntactically required.
+    #[must_use]
+    pub fn with_forced_tuple_parentheses(mut self, forced: bool) -> Self {
```
MichaReiser: This option seems fairly low-level and I worry we'll have more conflicts in the future because we're likely to go to preserve the source formatting in more places where what you want is some fixed formatting (not sure I understand why it matters to you how tuples are formatted). 

IMO, this does raise the question if our goals are well aligned or if we'll keep patching up differences forever. Have you considerd just forking the generator?

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: sounds good

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: > This option seems fairly low-level and I worry we'll have more conflicts in the future because we're likely to go to preserve the source formatting in more places where what you want is some fixed formatting (not sure I understand why it matters to you how tuples are formatted).

Because at RustPython we are using CPython test suite. when you do:
```python
import ast
parsed = ast.parse('x = 1,')
ast.unparse(parsed)
```
The returned value is `x = (1,)`, and our tests fails because it's getting `x = 1,` instead

> IMO, this does raise the question if our goals are well aligned or if we'll keep patching up differences forever. Have you considerd just forking the generator?

We have considered that, and agreed on try to get it upstream and in the worst case we will fork, but that's less ideal for us. obviously I get the concern with merging code that solves a very niche issue

MichaReiser: > We have considered that, and agreed on try to get it upstream and in the worst case we will fork, but that's less ideal for us. obviously I get the concern with merging code that solves a very niche issue

I'm less concerned about the niche use case. My concern is that our goals aren't aligned. Our long-term goal is to preserve as much as possible about the source formatting as possible. This is in direct contrast with what you want. You want a generator that's `ast.unparse` compatible.

youknowone: Hi! Long time no see.

I fully understand that it’s not natural for unnecessary features to be included in the Ruff repository.

However, from interpreting your explanation, it seems that @ShaharNaveh’s requirements may not actually be in direct contrast with your long-term goals.

To preserve the source formatting, the system would need to store whether the original text included parentheses. Only then it could decide whether to restore them. Currently, since the original parenthesis information isn’t preserved, Ruff can only offer two fixed options: always include parentheses or never include them.

In the long term, If ruff_python_codegen::Generator will need to include information from the original source anyway, then perhaps @ShaharNaveh could help implement this based on your suggestion how to approach consistent with the project’s goals. 
Meanwhile, RustPython could achieve the same purpose by simply using a fixed value instead of the original formatting information.

MichaReiser: The issue is that, once added, other users than RustPython will start using it and it will be harder for us to remove or even remember what the use case was. 

I also don't fully understand the hesitation of just forking that one module. It's not that much code.

youknowone: My point is that if, as you say, your goal is to preserve the exact syntax of the original input, then users also can use the same feature aligned to your goal.

It’s unfortunate that this discussion was seen merely as hesitation about forking. Everyone knows the code can be forked at any time, but I don’t think that was ever the focus of this conversation.

Anyway, I understand that you’re currently not interested in this issue, and that the Ruff project is more focused on the concern that “it will be harder for us to remove or even remember what the use case was” rather than on a technical discussion about whether a similar feature might be needed in the future to restore the original syntax.

Thanks

MichaReiser: > My point is that if, as you say, your goal is to preserve the exact syntax of the original input, then users also can use the same feature aligned to your goal.

I don't understand that part. The feature you need is to always set parentheses because that's the only formatting that's compatible with `unparse`.

> rather than on a technical discussion about whether a similar feature might be needed in the future to restore the original syntax.

I don't think that's what you need? 

The example from @ShaharNaveh shows that you don't want to preserve formatting:

```py
import ast
parsed = ast.parse('x = 1,')
ast.unparse(parsed) # should become x = (1,)
```

youknowone: To preserve the original syntax, there must be a field to store information that it had parenthesis or not before.
Once if there is a field. @ShaharNaveh can fill the field as he needs.

That's based on your description:

> ... because we're likely to go to preserve the source formatting in more places ...

> Our long-term goal is to preserve as much as possible about the source formatting as possible. 

Is the preservation already implemented? Then he might be able to look for it if it is reusable or not.
Otherwise it will be required later to preserve the original formatting.

MichaReiser: No such field exists today because Ruff always uses the precedence. But we ran into the same issue with quotes, where Ruff used to preserve parentheses, and we had to add an extra option to allow customizing the quotes. That's where my concern comes from, the "better" we make the generator for Ruff's purpose, the more flags we'll need. 

I'd be open to removing the quote style option again and instead introducing a Mode option with two variants: `Default` and `AstUnparse`. This at least cleary documents the intent of all those options

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -63,13 +64,20 @@ mod precedence {
     pub(crate) const MAX: u8 = 63;
 }
 
+#[derive(Default)]
+pub enum UnparseMode {
+    #[default]
+    Default,
+    Ast,
+}
```
MichaReiser: Nit: I think just `Mode` is enough. I'd rename `Ast` to `AstUnparse` and document that the intent is to emit the same output as Python's `ast.unparse` method

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -173,7 +182,10 @@ impl<'a> Generator<'a> {
                 return;
             }
         }
-        let quote_style = self.preferred_quote.unwrap_or_else(|| flags.quote_style());
+        let quote_style = match self.unparse_mode {
+            UnparseMode::Default => flags.quote_style(),
+            UnparseMode::Ast => Quote::Single,
+        };
```
MichaReiser: We could make this a method on `Mode` insteada of repeating it at every call site

```
impl Mode {
	const fn quote_style(self, flags: AnyStringFlags) -> QuoteStyle {
		...
	}
}
```

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: np, would you prefer this documented at the enum variant, or the `Generator::with_unparse_mode`?

MichaReiser: I suggest documenting it at the variant level

crates/ruff_python_codegen/src/generator.rs
ShaharNaveh: I guess that it would make sense to move the tuple precedence value to it's own method as well, right? although it's only used once it kind of makes more sense to me for some reason. I'll do that if you think it's a good idea

MichaReiser: I'm less concerned about the tuple case because it's used exactly once.

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -63,13 +64,36 @@ mod precedence {
...
+    AstUnparse,
+}
+
+impl Mode {
+    /// Quote style to use.
+    ///
+    /// - [`Default`](`Mode::Default`): Output of `[AnyStringFlags.quote_style`].
+    /// - [`AstUnparse`](`Mode::AstUnparse`): Always return [`Quote::Single`].
+    #[must_use]
+    fn quote_style(&self, flags: impl StringFlags) -> Quote {
```
ShaharNaveh, line 82: had to use generics because this gets called with either `AnyStringFlags` or `BytesLiteralFlags`

crates/ruff_python_codegen/src/generator.rs
```diff
@@ -1586,14 +1611,13 @@ impl<'a> Generator<'a> {
...
 mod tests {
-    use ruff_python_ast::str::Quote;
     use ruff_python_ast::{Mod, ModModule};
     use ruff_python_parser::{self, Mode, ParseOptions, parse_module};
     use ruff_source_file::LineEnding;
 
     use crate::stylist::Indentation;
 
-    use super::Generator;
+    use super::{Generator, Mode as UnparseMode};
```
ShaharNaveh, line 1620: Renamed to `UnparsedMode` here because the name collides with 
```rust
use ruff_python_parser::{..., Mode, ...};
```

crates/ruff_python_codegen/src/generator.rs
youknowone: Thank you all for trying to find common ground.

ShaharNaveh:
tysm @MichaReiser <3 !

PR #21042, opened 2025-10-23
[ty] Simplify `Self` for final types

### LINKED ISSUE #1404: Failed exhaustiveness matching on type variable
Consider this example where we currently emit a false positive.

```py
from typing import Self
from enum import Enum

class Answer(Enum):
    YES = "yes"
    NO = "no"

    def is_yes(self: Self) -> bool:    # ty: Function can implicitly return `None`, which is not assignable to return type `bool`
        match self:
            case Answer.YES:
                return True
            case Answer.NO:
                return False
```

When the annotation is changed to `self: Answer`, everything works as expected. But we do not recognize that `Self@is_yes` with an upper bound of `Answer` can be narrowed down.

Similar problems certainly exist for other final types and other narrowing patterns.

It seems like we can generally simplify a non-inferable typevar `T` with a `final` upper bound `F` to that upper bound `F`? Or let `T` share more properties of `F` (like whether or not we consider it to be `final`).

sharkdp:
## Summary

For final classes, we simplify `Self` to an instance of the class.

Ideally, we would also do that for other type variables with a final bound, but those seem like a rare edge-case, because there's not really any benefit to using a type variable in the first place, if it's bound to a final type (thank you, @AlexWaygood).

closes https://github.com/astral-sh/ty/issues/1404

## Test Plan

Added regression test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/self.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/self.md
@@ -508,4 +508,22 @@ def _(c: CallableTypeOf[C().method]):
     reveal_type(c)  # revealed: (...) -> None
 ```
 
+## Final classes
+
+For final classes, we simplify `Self` to an instance of the class:
+
+```py
+from typing import final, Self
+
+@final
+class FinalClass:
+    def implicit_self(self) -> Self:
+        reveal_type(self)  # revealed: FinalClass
+        return self
+
+    def explicit_self(self: Self) -> Self:
+        reveal_type(self)  # revealed: FinalClass
+        return self
+```
+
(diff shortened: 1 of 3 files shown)


AlexWaygood (commented):
Thinking about this some more, I'm not totally sure this is sound for generic methods. Just because an instance type comes from a `@final` class doesn't mean that there aren't subtypes of that instance type, and the `Self` type variable could be solved to be one of those subtypes. For example:

```py
import enum
from typing import Self

class F(enum.Enum):
    X = 0
    Y = 1

    def method(self) -> Self:
        return self

reveal_type(F.X.method())  # `Literal[F.X]` on `main`, `F` on this branch
```

PR #21043, opened 2025-10-23
[`airflow`] Extend `airflow.models..Param` check (`AIR311`)

Lee-W:
## Summary

* Extend `airflow.models.Param` to include `airflow.models.param.Param` case and include both `airflow.models.param.ParamDict` and `airflow.models.param.DagParam` and their `airflow.models.` counter part

## Test Plan

update the text fixture accordingly and reorganize them in the third commit


ntBre (approved):
Thank you!

PR #21044, opened 2025-10-23
[ty] Improve `invalid-argument-type` diagnostics where a union type was provided

AlexWaygood:
Fixes https://github.com/astral-sh/ty/issues/1411

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -3527,28 +3527,23 @@ impl<'db> BindingError<'db> {
                         .filter(|element| !element.is_assignable_to(context.db(), *expected_ty))
                         .copied()
                         .collect();
+                    let first_invalid_element = invalid_elements[0].display(context.db());
                     if invalid_elements.len() < union_elements.len() {
-                        let (first, rest) = invalid_elements
-                            .split_first()
-                            .expect("There must be at least one invalid union element");
-
-                        match rest {
+                        match &invalid_elements[1..] {
                             [] => diag.info(format_args!(
-                                "Element `{}` of this union is not assignable to `{}`",
-                                first.display(context.db()),
-                                expected_ty_display,
+                                "Element `{first_invalid_element}` of this union \
+                                is not assignable to `{expected_ty_display}`",
                             )),
                             [single] => diag.info(format_args!(
-                                "Union elements `{}` and `{}` are not assignable to `{}`",
-                                first.display(context.db()),
+                                "Union elements `{first_invalid_element}` and `{}` \
+                                are not assignable to `{expected_ty_display}`",
                                 single.display(context.db()),
-                                expected_ty_display,
                             )),
-                            _ => diag.info(format_args!(
-                                "Union element `{}`, and {} more union elements, are not assignable to `{}`",
-                                first.display(context.db()),
+                            rest => diag.info(format_args!(
+                                "Union element `{first_invalid_element}`, \
+                                and {} more union elements, \
+                                are not assignable to `{expected_ty_display}`",
                                 rest.len(),
-                                expected_ty_display,
                             )),
                         }
                     }


crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap
```diff
@@ -0,0 +1,135 @@
...
+  --> src/mdtest_snippet.py:14:7
+   |
+12 |     d: list[str] | str | dict[str, str] | tuple[str, ...] | bytes | frozenset[str] | set[str] | Foo | Bar | Baz,
+13 | ):
+14 |     f(a)  # error: [invalid-argument-type]
+   |       ^ Expected `Sized`, found `str | Foo`
+15 |     f(b)  # error: [invalid-argument-type]
+16 |     f(c)  # error: [invalid-argument-type]
+   |
+info: Element `Foo` of this union is not assignable to `Sized`
```
MichaReiser: The secondary info message won't be shown with `--output-format=concise` or in the main diagnostic shown in editors (unless they supported related information which many dont). 

Given that this is the most important bit, I'm inclined to make it part of the primary annotation message, although that's somewhat tricky

AlexWaygood: I think we may have had this conversation before... I think we may need to rethink our diagnostic model here. It's often not feasible to encapsulate all relevant information in a single line, and mypy, for example, will often display secondary diagnostics even if you do not specify `--pretty`. (Mypy uses a concise output format by default; `--pretty` gives you annotations as part of diagnostics, similar to our output format.) E.g. https://mypy-play.net/?mypy=latest&python=3.12&gist=8322703266ffc5e8e2ccb9538e2487c3

The approach I'm taking here is also exactly what we already do in similar diagnostics, so I'm inclined to say that if we want to change this approach, we should do so for all our diagnostics together: https://github.com/astral-sh/ruff/blob/01695513ce33f1f1615309323ba145c42f4720c1/crates/ty_python_semantic/resources/mdtest/snapshots/union_call.md_-_Calling_a_union_of_f%E2%80%A6_-_A_smaller_scale_exam%E2%80%A6_(c24ecd8582e5eb2f).snap#L34-L53

MichaReiser: It's probably worth changing it everywhere because using sub diagnostics with info won't show with `--output-format=concise` or in the editor. The reason I opened this issue was because I didn't had enough information in the editor to fix the diagnostic. 

I'd be inclined to change the message to:

```
Expected `Sized`, element `Foo` of `str | Foo` is not assignable to `Sized`

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > using sub diagnostics with info won't show with `--output-format=concise`

yes, that's the thing I'm suggesting we might want to rethink when I said "we may need to rethink our diagnostic model here". I don't think that's a great outcome. I think it's going to end up with worse diagnostics than mypy for CLI users. It's often silly to try to cram all relevant information onto one line, and for things like overloads it's arguably impossible.

> or in the editor

Whenever I need more information on a diagnostic rust-analyzer is showing me in the "Problems" tab in VSCode, I always click on the "see full diagnostic" option, and then a new tab opens up with all subdiagnostics displayed as well as the main diagnostic. Is it not possible for us to implement something similar for users of the ty server?

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: And it seems to me that I'm already being shown several subdiagnostics on-hover here in VSCode, for example, even without clicking on the "Click for full compiler diagnostic" option?

image

Comment (MichaReiser) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > Whenever I need more information on a diagnostic rust-analyzer is showing me in the "Problems" tab in VSCode, I always click on the "see full diagnostic" option, and then a new tab opens up with all subdiagnostics displayed as well as the main diagnostic. Is it not possible for us to implement something similar for users of the ty server?

We do implement this but it's not supported by all editors

> I think it's going to end up with worse diagnostics than mypy for CLI users. 

Only when using `--output-format=full`. 

I do share the sentiment that we shouldn't try to put everything on a single line. But I think it's crucial that the most important bits are visible on that single line. In this case, it's not the union, but which element of the union doesn't implement `Sized`. We can move the union out of the main message into a secondary info message. But I think we need to have the problematic element in the primary message.

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > `` Expected `Sized`, element `Foo` of `str | Foo` is not assignable to `Sized` ``

I think if I'm a user, this message could easily raise more questions than it answers. The user might not have any idea why we're talking about either `Foo` or `Foo | str` -- we haven't yet told them that we inferred a type of `str | Foo` for the argument they passed into the function call. Users think in terms of values, not types; all they'll see in their source code is a variable `bar`, and they might not even be aware of what type `bar` has at runtime, let alone the union type we've inferred for it in ty. I don't think we can start talking about which specific element in the union is causing an issue before we've told them that we've inferred a union type for the argument being passed in

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > We do implement this but it's not supported by all editors

Feels like a feature request that should be brought up with those editors, since it'll also lead to a bad experience when using other LSP servers like rust-analyzer ;)

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: This discussion does make me wonder whether it might be worth revisiting my original idea in https://github.com/astral-sh/ruff/pull/20730#discussion_r2410446248 of having a "smart filter" option for union displays (so that we can make sure at least one "problematic union element" is always included in the display)

Comment (MichaReiser) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > Feels like a feature request that should be brought up with those editors, since it'll also lead to a bad experience when using other LSP servers like rust-analyzer ;)

Maybe, but that's out of our control and it's still an issue when using `--output-format=concise`. 

> This discussion does make me wonder whether it might be worth revisiting my original idea in https://github.com/astral-sh/ruff/pull/20730#discussion_r2410446248 of having a "smart filter" option for union displays (so that we can make sure at least one "problematic union element" is always included in the display)

I'd sort of like this more. It seems very unfortunate to just truncate anywhere. I think it also degraded the LSP experience because it's now impossible to see the full type when hovering (and I assume we also truncate when using `reveal_type`?)

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: > and I assume we also truncate when using `reveal_type`?

no, we do not. There's an option to switch off truncation when displaying types and `reveal_type` already uses it. The type displayed for on-hover should probably also use it; this was probably an oversight in #20730!

Comment (MichaReiser) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: It's probably worth exploring how other type checkers (across ecosystem) handle long types (and union types) in diagnostics and editors

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: Though even for on-hover, we may want to avoid displaying all union elements in situations where the union is [60 elements long](https://github.com/astral-sh/ruff/pull/20730#issuecomment-3381454378)...?

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: In general I think we infer unions more than any other Python type checker. This is not only because we treat intersections as first-class types, but also because we have concepts such as "class-literal types" and "function-literal types". To my knowledge, no other Python type checker treats these concepts as first-class types, and treating them as such seems to mean that huge unions are more common for us than for other type checkers

Comment (MichaReiser) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: Screenshot 2025-10-23 at 14 32 24

Typescript does truncate very long unions

Screenshot 2025-10-23 at 14 34 04

https://www.typescriptlang.org/play/?#code/PTAEDEFcDsGMBcCWB7ap4AsCG9SoDYCeoWssApgA7wDOoN8ATotAOY0BQAZjAimpWbR4AZSYtWACgBuWfJHIAueuLYBKZdOSIAJqADeHUKFioayfOQB0+ZFNnzyagNwcAvhw4hQAFQyI6AHdEfHxQckZGZEZQACNyWCxIGnJ0DFSsaFRCAFtkZNAYfhNUeCwWOgAWACZuXiRUdHIGSSNQRHhyHOU24wAfUAAiLEpKS0He0AHB2My5ieN+odh0yMIFxemdHHINpcHyfB0I+LW9qaGuRFZz6dZGEd3J6YxUckJjwNuhgGtEYO+g0seWggJymVYyEB0ASZSET0WF0G0QhCM2Q0oIywhCwgIAjpAWBRAQ8aJRTox1s8hgwHoEKVTEdMymwIiw0ftAjtGDlDqhASNmKZ4IDYvhSD8GaLHFLqYNEsI5PlKBykaZTNBICK5bAHtBZUyhjoHpDoFxGITtYbBpDkCkDeibQ8VebLYDWJAsLJAQArCWujqA-CIXlBwgrciq6bg6DbIRgyD4B37CyIaRRjHkUgYQEqrCMXMRGghkG59kjMYZwZjSA5XPIXn3OadXMWmEk8rF0Fyhj5gNWx3wBs4KHUmrGbwAOVrp3aaEwqSKqAAhBwNKAtLoDG1BCxRKopB0ui5QN4AKKRaLKTU5WcBUBZXBYGjF1jQLBi1JDlRCVirjxeGAZ7pmgeSMKkpg5JWAAeyiZNkeQFEuaDBJgeCxD6sLoIQKqcDwcANAIUQUC+ADCDYwQAkp0OStMYR7dNS+jtDo14zhEzgPlgvLKLSEicaQSDpsosTIBYWZoB4hrMbobG3hx6AdJYvEHpxlCQGKATpKxcRiZYmSgFJ6IyTpN6nJx4rxPgKm-px0gBIgn4iXpEmGUxLFyeZoDHDQuqINQ-A2fx4Tvp+OmieJBlGUsJmeQpZSsEFbCcVwWbwJA4HhS5UXubJD7sYwnGJJ0kKUklrACYwKxpuQWWRZJuWmQVdlyAo5WcSklgILVzn1W50kefl8mFSUwjkMI7UkKMUTpnV+kNQNeVmfF5DQfAk3pswVw9bpfXRRcsVDV5iQBagk3+KwGDBpdnRza5+0DIdy0jekWA6BIk27uCZW7fN-XGYNz0dRpr3vWwk0pBqcaEL1f0PQYgPNfQtbfTDP7BatmKxjtEVw41cUjY+Sjo8loCUCwMJ3Tli1NcNRUNryE0kxVoBcOKrCsDj2ULQDS1I7yL5YJzk3gW9sP3fjR0KTQGmYQgENwpl4vU7ztNeUO5OwJNTDjWDiW-RLNME5xPzvIE0Q6XxpNMBKXN7ZLQNkxgpLE1bLMyxzzS3crPMxYjdOgASyCdJNXBetER5U77B3+15NDQP5KrrcznERT832m1H-1+3zAerRQjDUBD2BKwbKs52rCmYiazoYJNxpYFw3tl9Hj2x9L42dHAruqSUUGWM3uOG6rxugObjCW73gjICRKRZ-DT1I50PKTSwxzQXbeNG1LI3HFcCeESL5D2eQgSb8PFej5j4rvofKcbnIug7PPDtI7A4pbYgxWBffGpXDy59y4x1zl5CoTBIB8DOvfVaCQtSANbgjEBCl3rgT4MJe+IZKyM0HtzbOwDK4jQ9LoQ47JQ4WFsGfF+29HYWmUtAs00QKBUJHjvNSqZYBozdgJUYwZ4F4Lbkgka08mEZR7rZbyM9aydz4QvduQiohDlMNZaBvZNI0G0j7fhiCCEdRZHGSe4i3rIGoDI1+AcyQJEQFcb+UCuHeXIPvUx1CkbgQJIgcC2CIY4ACNtZhl9WEmGYMvH+djeTJyHkAgROjRrvTvnYng+ArihCcSwx2GpaTlCZnY8CFjup+PwaPYMOQOjeNsb3CUWRAiWB0JzfJUTR6mGOKPBwT8cF7QnGAAA5I08gnT2h0Gequdcm49CGA6b4Sk6BkAkDIM0OgWAyZRBVIweAxBMA4HCNBAItA4halANgOgRIKLeM-NhXCbR0m4B6QANVamIiQoAAC87QaJWB6ZxU8QFLyMH-EAA

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: No other Python type checker attempts to highlight the problematic union element as part of the _primary_ diagnostic for something like this, from what I can tell:
- Mypy: https://mypy-play.net/?mypy=latest&python=3.12&gist=e52205a0769db5dd033a163d85f561f8
- Pyright: https://pyright-play.net/?pyrightVersion=1.1.405&strict=true&code=CYUwZgBGAUAeBcECWA7ALhAPhAzmgTgJTwCwAUBJRADYgpyHlA
- Pyrefly: https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSImMYABGABT6I0ToAuNAPjXGwE4BKRAB10NcTVjpGA0SAA0IAK5tocEuUQgAxDQCqqqBDak6S9AGNVudHFGiqtMLj4BbVGwD66Ja%2Bww%2BRmZWNgEaAFoAPh5%2BETEJPhg2JT4xMGEQADlffz5mYHwAXwy5RTJEsChSQjZcVygKXQAFUgqqngwcAhoLG0gAcxSPCBtCUV0AZRgYGgALNjZiOEQAehXy6irCF36VmHQVzFwLOBXe9AGh6wO6FxpUADdUaFRsWB6%2BiEG%2BYZsaXGI13UojIbFmNnCDwCcBGYgAvDQMgBmQgARgATCV0CBCopUFYIFCAGLQGAUNBYPBEMg4oA

Pyright highlights it in a subdiagnostic, the same as my PR currently proposes

Comment (MichaReiser) [reply crates/ty_python_semantic/resources/mdtest/snapshots/function.md_-_Call_expression_-_Wrong_argument_type_-_Diagnostics_for_unio…_(5396a8f9e7f88f71).snap]: That's true. But no other type checker seems to truncate the union type :)

I'm fine merging this as is but I really think our approach to truncating unions is problematic

File: crates/ty_python_semantic/src/types/call/bind.rs [hunk 2]
```diff
@@ -3519,6 +3519,36 @@ impl<'db> BindingError<'db> {
...
+                            )),
+                            rest => diag.info(format_args!(
+                                "Union element `{first_invalid_element}`, \
+                                and {} more union elements, \
+                                are not assignable to `{expected_ty_display}`",
+                                rest.len(),
+                            )),
+                        }
+                    }
+                }
```
sharkdp, line 3550: I assumed we would fix this somewhere else, e.g. when actually checking assignability in the first place. This code here only affects invalid-argument-type, but we should do the same thing in other situations, like for actual assignments to variables. Consider:
```py
from typing import Sized

class Foo: ...

def g(
    b: list[str]
    | str
    | dict[str, str]
    | tuple[str, ...]
    | bytes
    | frozenset[str]
    | set[str]
    | Foo,
):
    y: Sized = b  # still only shows the truncated union
```

AlexWaygood: A more general solution would probably be https://github.com/astral-sh/ruff/pull/19580 (so that we always propagate up the reason _why_ an assignability/subtyping relation did not hold). I do think we need something like that eventually, for better diagnostics in cases where Protocol subtyping doesn't hold (our diagnostics are currently terrible for protocols with lots of members — you can't tell which specific member of the protocol wasn't satisfied). But that would be a big change, and I'm not yet sure how it would fit in with Doug's work on the new constraint solver.

We already do ad-hoc special-casing of union types like this for several other diagnostics. I agree it's not great but I think it's the best we can do for now, and it's better than nothing in the short term.

PR #21047, opened 2025-10-23
[ty] Make the default settings try harder to avoid crashes

Gankra:
This introduces a safe_mode to our settings type that tells our initialization code to more aggressively believe Settings Failure Is Not An Error. This allows the full settings initialization logic to be used as much as possible for the default database while trying to keep any discarded results as narrow as possible for graceful degradation.

Finally, Python has SFINAE.

Fixes https://github.com/astral-sh/ty/issues/859

### DIFF
--- a/crates/ty_project/src/metadata/options.rs
+++ b/crates/ty_project/src/metadata/options.rs
@@ -94,9 +94,23 @@ pub struct Options {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[option_group]
     pub overrides: Option<OverridesOptions>,
+
+    /// Settings Failure Is Not An Error.
+    ///
+    /// This is used by the default database, which we are incentivized to make infallible,
+    /// while still trying to "do our best" to set things up properly where we can.
+    #[serde(default, skip)]
+    pub safe_mode: bool,
 }
 
 impl Options {
+    pub fn safe() -> Self {
+        Self {
+            safe_mode: true,
+            ..Self::default()
+        }
+    }
+
     pub fn from_toml_str(content: &str, source: ValueSource) -> Result<Self, TyTomlError> {
         let _guard = ValueSourceGuard::new(source, true);
         let options = toml::from_str(content)?;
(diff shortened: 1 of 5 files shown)


Gankra:
Ah, rude. The exact case of `VIRTUAL_ENV=gibberish` actually blows up in asserts in `dynamic_resolution_paths`. This is a big ol' game of whackamole.

Gankra:
The failure with a gibberish `VIRTUAL_ENV` is a fascinating maybe-independent bug in our handling of symlinked python environments (homebrew).

We actually handle the gibberish VIRTUAL_ENV value gracefully but then we fallback to grabbing a reference to some homebrew site-packages and then blowup because symlinks are happening and the same dir looks like two different dirs.

(just on main, not with my patches, although my patches don't affect this):

```
No root found for path '/opt/homebrew/Cellar/python@3.13/3.13.5/lib/python3.13/site-packages'. Known roots: FileRoots(
...
path: "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages",
```

Here we have a symlink:

```
/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages 
-> 
../../../../../../lib/python3.13/site-packages
```

MichaReiser:
@sharkdp did you intentionally mark this pr as ready for review?

sharkdp:
No, sorry. I was reading through it on mobile and must have hit something accidentally. And I can't even find the button to change it back 🙃

By the way, love the SFINAE branch name.

MichaReiser:
You mentioned over chat if we should return a Result with the errors instead of a safe flag. I'm starting to see the appeal of this, if we try to get as much right as possible even in safe mode

The resolution methods could return a result where the error type wraps the fallback settings and the first (or all errors). A caller could then decide to build the program with the fallback settings (or error). 

I'm not sure how much harder that is

Gankra:
We no longer have an input that crashes the default server so I'm going to close this as WONTFIX until someone reports a new one.

PR #21049, opened 2025-10-23
[ty] Consider `__len__` when determining the truthiness of an instance of a tuple class or a `@final` class

AlexWaygood:
## Summary

At runtime, if a class does not define `__bool__` but does define `__len__`, the result of the `__len__` call will be used to determine the truthiness of that object. The most common examples of this are various builtin containers such as tuple, list and others:

```pycon
>>> bool(())
False
>>> bool((1,))
True
>>> list.__bool__
  File "<python-input-6>", line 1, in <module>
    list.__bool__
AttributeError: type object 'list' has no attribute '__bool__'. Did you mean: '__doc__'?
>>> bool([])
False
>>> bool([1])
True
```

Up till now, we haven't attempted to represent this in ty: if a class `N` defines a custom `__len__` method, we've ignored it for the purposes of determining the truthiness of that class's instances. This is because `__bool__` always takes priority over `__len__`, and it's always possible that a subclass of `N` could override `__bool__` (thus invalidating, for some subtypes of `N`, the assumptions regarding truthiness that we made from looking at the return type of `N.__len__`).

However, this approach has several problems:
- It ignores the fact that if a `@final` class defines `__len__` but does not define `__bool__`, it's perfectly safe to look at the return type of `__len__`
- It ignores the fact that if a `@final` class defines neither `__len__` nor `__bool__`, it's perfectly safe to consider the class as being always truthy
- It's led us to synthesize `__bool__` methods for tuple classes that actually don't exist at runtime

This PR fixes these problems. It still leaves several issues unresolved: arguably, if we see that the runtime implicitly calls a badly defined `__len__` method, we should emit diagnostics warning the user about this. But these false negatives already exist on `main`, and it would take quite some time to write out all the code and tests for this kind of diagnostic. It doesn't seem to me to be a priority right now, so I chose to cut scope here.

Helps with https://github.com/astral-sh/ty/issues/1420. It doesn't close that issue (our behaviour will remain the same for the snippet the author posted in that issue!), but it'll make it easier for me to justify our current behaviour to the author of that issue. I have trouble justifying our current behaviour, since it just doesn't emulate the semantics at runtime fully enough IMO 😄

## Test plan

Added mdtests

### DIFF
--- a/crates/ty_ide/src/goto_definition.rs
+++ b/crates/ty_ide/src/goto_definition.rs
@@ -1324,6 +1324,44 @@ a = Test()
         ");
     }
 
+    /// We jump to the `__invert__` definition here even though its signature is incorrect.
+    #[test]
+    fn goto_definition_unary_operator_with_bad_dunder_definition() {
+        let test = CursorTest::builder()
+            .source(
+                "main.py",
+                "
+class Test:
+    def __invert__(self, extra_arg) -> 'Test': ...
+
+a = Test()
+
+<CURSOR>~a
+",
+            )
+            .build();
+
+        assert_snapshot!(test.goto_definition(), @r"
+        info[goto-definition]: Definition
+         --> main.py:3:9
+          |
+        2 | class Test:
+        3 |     def __invert__(self, extra_arg) -> 'Test': ...
+          |         ^^^^^^^^^^
+        4 |
+        5 | a = Test()
+          |
+        info: Source
+         --> main.py:7:1
+          |
+        5 | a = Test()
+        6 |
+        7 | ~a
+          | ^
+          |
+        ");
+    }
+
     #[test]
     fn goto_definition_unary_after_operator() {
         let test = CursorTest::builder()
@@ -1474,6 +1512,7 @@ a = Test()
     }
 
     /// If `__bool__` is defined incorrectly, `not` does not fallback to `__len__`.
+    /// Instead, we jump to the `__bool__` definition as usual.
     /// The fallback only occurs if `__bool__` is not defined at all.
     #[test]
     fn goto_definition_unary_not_with_bad_dunder_bool_and_dunder_len() {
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/type_compendium/always_truthy_falsy.md
```diff
@@ -17,7 +17,7 @@ Here, we give a few examples of values that belong to these types:
 
 ```py
 from ty_extensions import AlwaysTruthy, AlwaysFalsy
-from typing_extensions import Literal
+from typing_extensions import Literal, final
```
dcreager: Is this spurious?

crates/ty_python_semantic/src/types.rs
```diff
@@ -4509,18 +4506,65 @@ impl<'db> Type<'db> {
...
                 }
 
-                Err(CallDunderError::MethodNotAvailable) => Ok(Truthiness::Ambiguous),
+                Err(CallDunderError::MethodNotAvailable) => {
+                    // We only consider `__len__` for tuples and `@final` types,
+                    // since `__bool__` takes precedence
+                    // and a subclass could add a `__bool__` method.
+                    // With regards to tuple types, we intend to emit a diagnostic
+                    // if a tuple subclass defines a `__bool__` method with a return type
+                    // that is inconsistent with the tuple's length.
```
dcreager: nit: Since this calls out an intention, can you mark it with `TODO`?

dcreager (approved):
bueno

MichaReiser (commented):
Hmm, I worry this will be painful but we'll need to also implement the same logic for the not goto definition

crates/ty_python_semantic/resources/mdtest/type_compendium/always_truthy_falsy.md
AlexWaygood: oops, yes

AlexWaygood:
> Hmm, I worry this will be painful but we'll need to also implement the same logic for the not goto definition

Great point. I think I fixed this -- want to take another look?

For goto definition, I don't think it's so important that we do the "only respect the dunder if it's `@final` dance. We don't really need to worry so much about soundness when the user is just asking us to jump to the definition that will be used at runtime 😄

crates/ty_ide/src/goto_definition.rs
```diff
@@ -1294,11 +1294,11 @@ class Test:
                 "main.py",
                 "
 class Test:
-    def __bool__(self) -> bool: ...
+    def __invert__(self) -> 'Test': ...
```
MichaReiser, line 1297: What's the reason for changing this test?

crates/ty_ide/src/goto_definition.rs
```diff
@@ -1399,6 +1399,102 @@ a = Test()
...
+        5 | a = Test()
+        6 |
+        7 | not a
+          | ^^^
+          |
+        ");
+    }
+
+    /// If `__bool__` is defined incorrectly, `not` does not fallback to `__len__`.
+    /// The fallback only occurs if `__bool__` is not defined at all.
```
MichaReiser, line 1516: This seems somewhat unfortunate. I think I'd expect the IDE to jump to either `__bool__` or `__len__` (it should be forgiving). E.g, huh... why does ty complain about `not a` not being supported. Click... ahhh, `__bool__` has the wrong signature. 

We don't need to change this here but I think we should change the documentation to make it sound less so that this is our desired behavior.

AlexWaygood: Oh, I was just trying to preserve our existing behaviour here. goto already doesn't work for the other operators if the dunders are defined slightly incorrectly. I'll fix this up, thanks

crates/ty_ide/src/goto_definition.rs
AlexWaygood: `not` now works slightly differently to all other unary operators when it comes to goto-definition, so I felt like we should probably test with a different unary operator for "generic unary operator functionality" tests like this, and then have some different tests on top of that for the specific ways in which `not` differs from the rest

crates/ty_python_semantic/src/types.rs
```diff
@@ -4509,18 +4506,67 @@ impl<'db> Type<'db> {
...
+                    // since `__bool__` takes precedence
+                    // and a subclass could add a `__bool__` method.
+                    //
+                    // TODO: with regards to tuple types, we intend to emit a diagnostic
+                    // if a tuple subclass defines a `__bool__` method with a return type
+                    // that is inconsistent with the tuple's length. Otherwise, the special
+                    // handling for tuples here isn't sound.
+                    if let Some(instance) = self.into_nominal_instance() {
+                        if let Some(tuple_spec) = instance.tuple_spec(db) {
+                            Ok(tuple_spec.truthiness())
```
carljm, line 4520: This is implicitly depending for its correctness on our plan (not yet implemented AFAIK?) to ban overriding `__len__` or `__bool__` on a tuple subclass?

AlexWaygood: yes, as stated in the comment immediately above this line of code

PR #21050, opened 2025-10-23
[ty] Consider domain of BDD when checking whether always satisfiable

dcreager:
That PR title might be a bit inscrutable.

Consider the two constraints `T ≤ bool` and `T ≤ int`. Since `bool ≤ int`, by transitivity `T ≤ bool` implies `T ≤ int`. (Every type that is a subtype of `bool` is necessarily also a subtype of `int`.) That means that `T ≤ bool ∧ T ≰ int` is an impossible combination of constraints, and is therefore not a valid input to any BDD. We say that that assignment is not in the _domain_ of the BDD.

The implication `T ≤ bool → T ≤ int` can be rewritten as `T ≰ bool ∨ T ≤ int`. (That's the definition of implication.) If we construct that constraint set in an mdtest, we should get a constraint set that is always satisfiable. Previously, that constraint set would correctly _display_ as `always`, but a `static_assert` on it would fail.

The underlying cause is that our `is_always_satisfied` method would only test if the BDD was the `AlwaysTrue` terminal node. `T ≰ bool ∨ T ≤ int` does not simplify that far, because we purposefully keep around those constraints in the BDD structure so that it's easier to compare against other BDDs that reference those constraints.

To fix this, we need a more nuanced definition of "always satisfied". Instead of evaluating to `true` for _every_ input, we only need it to evaluate to `true` for every _valid_ input — that is, every input in its domain.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
@@ -633,7 +633,7 @@ it!
 
 ```py
 from typing import Never
-from ty_extensions import constraint_set_domain, static_assert
+from ty_extensions import static_assert
 
 def f[T]():
     t_int = range_constraint(Never, T, int)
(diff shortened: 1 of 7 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/constraints.md
```diff
@@ -626,11 +628,50 @@ def f[T, U]():
...
+    static_assert(implication)
+
+    # However, because of that implication, some inputs aren't valid: it's not possible for
+    # `T ≤ bool` to be true and `T ≤ int` to be false. This is reflected in the constraint set's
+    # "domain", which maps valid inputs to `true` and invalid inputs to `false`.
+    # TODO: Under the covers, the domain BDD actually does include the `¬(T@f ≤ bool)` constraint,
+    # but it's currently being hidden by our display simplification logic.
+    # TODO: revealed: ty_extensions.ConstraintSet[((T@f ≤ int) ∧ ¬(T@f ≤ bool))]
+    # revealed: ty_extensions.ConstraintSet[(T@f ≤ int)]
+    reveal_type(constraint_set_domain(implication))
```
dcreager: Ugh... I wanted to display this without simplifying the DNF so that we could see (and test!) that its internal structure is not just the `AlwaysTrue` terminal. But the display simplification rules are important to get consistent output even when BDD variable ordering changes. So this is a flaky assertion, as can be seen in the CI failures.

For now, I think I'm going to back out the parts that let us display BDD domains in the mdtests. The equality test below still lets us test that the domain is not just `AlwaysTrue`, even if we can't display it nicely.

ibraheemdev (approved):
Nice! We do something similar in uv when simplifying marker trees based on the `requires-python` range of a project.

PR #21052, opened 2025-10-23
[ty] Do not crash if a FileRoot is a symlink

Gankra:
This is a fix for the issue [described here](https://github.com/astral-sh/ruff/pull/21047#issuecomment-3437837018), where ty found my homebrew python site-packages, and then panicked because it was handling both symlink and non-symlink versions of the same directory (which, syntactically, are not subdirectories of eachother).

I am always suspicious of using canonicalize but it's not clear to me there's a way to avoid it here.

### DIFF
--- a/crates/ruff_db/src/files.rs
+++ b/crates/ruff_db/src/files.rs
@@ -190,10 +190,8 @@ impl Files {
         let roots = self.inner.roots.read().unwrap();
 
         let absolute = SystemPath::absolute(path, db.system().current_directory());
-        // We need to resolve away symlinks here to avoid getting confused about subdirectories.
-        let canonicalized = db.system().canonicalize_path(&absolute).unwrap_or(absolute);
 
-        roots.at(&canonicalized)
+        roots.at(&absolute)
     }
 
     /// The same as [`Self::root`] but panics if no root is found.


crates/ruff_db/src/files.rs
```diff
@@ -190,7 +190,10 @@ impl Files {
         let roots = self.inner.roots.read().unwrap();
 
         let absolute = SystemPath::absolute(path, db.system().current_directory());
-        roots.at(&absolute)
+        // We need to resolve away symlinks here to avoid getting confused about subdirectories.
+        let canonicalized = db.system().canonicalize_path(&absolute).unwrap_or(absolute);
```
Gankra: On paper the first `absolute` is redundant but I'm not sure if `db.system().current_directory()` is able to be not-the-process-level-current-directory that  canonicalize_path uses.

crates/ty_test/src/db.rs
```diff
@@ -180,11 +180,18 @@ impl System for MdtestSystem {
...
+            if let Some(Utf8Component::Prefix(prefix)) = canonicalized.components().next()
+                && let Utf8Prefix::VerbatimDisk(_) = prefix.kind()
+            {
+                Ok(canonicalized)
+            } else {
+                Ok(canonicalized
+                    .strip_prefix(os_system.current_directory())
+                    .unwrap()
+                    .to_owned())
+            }
```
Gankra, line 194: This is a "Hillarious" interaction between `strip_prefix` and `dunce`. 

`dunce` removes "unnecessary" UNC prefixes, but the import_basic mdtest actually passes in a really really really long path, which `dunce` then goes "ah UNC is needed". 

This results in `canonicalized.strip_prefix(os_system.current_directory())` failing.

If you then go "ah I will simply canonicalize `os_system.current_directory()`" **you still lose** because `dunce` is very helpful and removes the unnecessary UNC prefix from the much shorter cwd!

MichaReiser:
I'd very much prefer not to call `canonicalize` in `Files` and, instead, call `canonicalize` where we add the `FileRoot` (and where we look it up if that turns out to be necessary). It limits the scope of where we have an undesired `canonicalize` call and also allows us to better document why it is necessary in this specific instance.

Gankra:
> I'd very much prefer not to call canonicalize in Files and, instead, call canonicalize where we add the FileRoot (and where we look it up if that turns out to be necessary).

Do you mean you want this logic moved into the routine it's calling (FileRoot::try_add, FileRoot::at)? 
Or do you mean out to each caller of  `try_add_root` / `expect_root` / `root`?

MichaReiser:
Ideally, we'd canonicalize the path where we create it before adding it as a search path (or site package path or whatever it is). This should then also ensure that we only create paths that use the canonicalized representation. But I don't understand the specific problem enough to say whether that's possible

Gankra:
Bafflingly I can't reproduce this anymore so I'm just gonna close this

Gankra:
Ok micha got a reproducer. If you go in vscode and via the pallete select "Python Interpretter: /opt/homebrew/bin/python3" the crash occurs:

```text
No root found for path '/opt/homebrew/Cellar/python@3.13/3.13.5/lib/python3.13/site-packages'. Known roots: FileRoots(
...
/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages
```

On my system the path we select is a fractal symlink:

1. `/opt/homebrew/bin/python3`
2. `/opt/homebrew/Cellar/python@3.13/3.13.5/bin/python3`
3. `/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/bin/python3`
4. `/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/bin/python3.13`

Path 4 is a real file. If you pass 1 or 2 to `ty check --python ...` everything works fine. If you pass 3 or 4 we have this panic.

The actual crash involves two other dirs in a symlink relationship:

5. `/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages`
6. `/opt/homebrew/Cellar/python@3.13/3.13.5/lib/python3.13/site-packages`

Path 6 is an actual directory. We also see that path 6 is, weirdly, what you would discover if you searched for site-packages from *path 2*, which is wild because path 2 was a bloody symlink!

So when handed path 3 or 4 as `--python` we find that we're in a *real* directory (so even if we canonicalized at that point it would do nothing). At this point we search for and discover site-packages at path 5, which *is* a symlink but we don't normalize that away.

Then at some later point we end up with a copy of the same path but with the symlink resolved (path 6), and we freak out because the two paths aren't lexically nested.

Gankra:
Ok so the "root cause" is this canonicalize (and the absence of a paired canonicalize for *adding* a root):

https://github.com/astral-sh/ruff/blob/b93d8f2b9fa834ab7c672d801c725c0031d9408e/crates/ty_python_semantic/src/module_resolver/resolver.rs#L451-L464

Gankra:
An unfortunate game of whackamole if you don't want to unconditionally canonicalize all paths.

Gankra:
Obsolete

PR #21054, opened 2025-10-24
[ty] Use constructor parameter types as type context

ibraheemdev:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1408.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/bidirectional.md
+++ b/crates/ty_python_semantic/resources/mdtest/bidirectional.md
@@ -153,72 +153,76 @@ Type context is sourced from various places, including annotated assignments:
 ```py
 from typing import Literal
 
-a: Literal["x"] = "x"
+a: list[Literal[1]] = [1]
 ```
 
 Function parameter annotations:
 
 ```py
-def b(x: Literal["x"]): ...
+def b(x: list[Literal[1]]): ...
 
-b("x")
+b([1])
 ```
 
 Bound method parameter annotations:
 
 ```py
 class C:
-    def __init__(self, x: Literal["x"]): ...
-    def foo(self, x: Literal["x"]): ...
+    def __init__(self, x: list[Literal[1]]): ...
+    def foo(self, x: list[Literal[1]]): ...
 
-C("x").foo("x")
+C([1]).foo([1])
 ```
 
 Declared variable types:
 
 ```py
-d: Literal["x"]
-d = "x"
+d: list[Literal[1]]
+d = [1]
 ```
 
 Declared attribute types:
 
 ```py
 class E:
-    e: Literal["x"]
+    e: list[Literal[1]]
 
 def _(e: E):
-    e.e = "x"
+    # TODO: Implement attribute type context.
+    # error: [invalid-assignment] "Object of type `list[Unknown | int]` is not assignable to attribute `e` of type `list[Literal[1]]`"
+    e.e = [1]
 ```
 
 Function return types:
 
 ```py
-def f() -> Literal["x"]:
-    return "x"
+def f() -> list[Literal[1]]:
+    return [1]
 ```
 
 ## Class constructor parameters
+
+```toml
+[environment]
+python-version = "3.12"
+```
 
 The parameters of both `__init__` and `__new__` are used as type context sources for constructor
 calls:
 
 ```py
-from typing import TypedDict
-
-class T1(TypedDict):
-    x: int
-
-class T2(TypedDict):
-    x: int
+def f[T](x: T) -> list[T]:
+    return [x]
 
 class A:
-    def __new__(cls, value: T1):
+    def __new__(cls, value: list[int | str]):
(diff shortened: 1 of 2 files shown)


ibraheemdev:
The mypy-primer diff changes are due to a known issue with duplicated diagnostics (see https://github.com/astral-sh/ty/issues/1428), there are no actual new diagnostics (somewhat surprisingly).

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -145,3 +145,80 @@ def h[T](x: T, cond: bool) -> T | list[T]:
...
+calls:
+
+```py
+from typing import TypedDict
+
+class T1(TypedDict):
+    x: int
+
+class T2(TypedDict):
+    x: int
```
dcreager: Is it important that these are separate `TypedDict` types? If so, can you call out why in a comment or in the markdown documentation?

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -145,3 +145,80 @@ def h[T](x: T, cond: bool) -> T | list[T]:
...
 ```
+
+## Type context sources
+
+Type context is sourced from various places, including annotated assignments:
+
+```py
+from typing import Literal
+
+a: Literal["x"] = "x"
```
dcreager: The examples in this section look like they're not exercising bidi typing to their fullest, since the expressions will already be inferred as `Literal["x"]` even without the type context. Would it be possible to update these so that the type context is different from what we'd normally infer for the expression? (Maybe a list literal and a more specific `list[something]` type context?) That might overlap with some of the tests elsewhere in the file, but I think it would make this section clearer.

(And ideally, it would be nice to be able to then show that the bidi-inferred type is different, by placing this alongside the same expression without a type context. But I think that would require `hover` annotations #20786)

crates/ty_python_semantic/src/types.rs
```diff
@@ -6030,11 +6030,64 @@ impl<'db> Type<'db> {
...
+        // Infer the call argument types, using both `__new__` and `__init__` for type-context.
+        let bindings = match (
+            new_method.as_ref().map(|method| &method.place),
+            &init_method.place,
+        ) {
+            (Some(Place::Defined(new_method, ..)), Place::Undefined) => Some(
+                new_method
+                    .bindings(db)
+                    // Inject a synthetic `cls` argument.
+                    .map(|binding| binding.with_bound_type(Type::unknown())),
```
dcreager: We use `self_type` [down below](https://github.com/astral-sh/ruff/pull/21054/files?w=1#diff-8049ab5af787dba29daa389bbe2b691560c15461ef536f122b1beab112a4b48aR6089) when we bind the `cls` parameter when calling `__new__`. Can we use that here too?

crates/ty_python_semantic/src/types.rs
```diff
@@ -6030,11 +6030,64 @@ impl<'db> Type<'db> {
...
+            (Some(Place::Defined(new_method, ..)), Place::Defined(init_method, ..)) => {
+                let callable = UnionBuilder::new(db)
+                    .add(*new_method)
+                    .add(*init_method)
+                    .build();
+
+                let new_method_bindings = new_method
+                    .bindings(db)
+                    // Inject a synthetic `cls` argument.
+                    .map(|binding| binding.with_bound_type(Type::unknown()));
```
dcreager: ditto here

crates/ty_python_semantic/src/types.rs
```diff
@@ -5971,10 +5971,10 @@ impl<'db> Type<'db> {
     ///
     /// Foo()
     /// ```
-    fn try_call_constructor(
+    fn try_call_constructor<'ast>(
         self,
         db: &'db dyn Db,
-        argument_types: CallArguments<'_, 'db>,
+        infer_argument_types: impl FnOnce(Option<Bindings<'db>>) -> CallArguments<'ast, 'db>,
```
dcreager, line 5980: Can you update the docs for this method to explain why this is a closure? (So that we can wait to infer types for arguments until we know the signatures of `__new__` and `__init__`, so that we can use their parameter annotations as type context for bidi inference)

dcreager (approved):

crates/ty_python_semantic/resources/mdtest/bidirectional.md
ibraheemdev: Sorry, that was an oversight, fixed. `hover` tests would nice :smile:

crates/ty_python_semantic/resources/mdtest/bidirectional.md
ibraheemdev: I realized this was actually incorrect because of structural subtyping. The idea was that both type contexts would be required for a valid inference, I updated the test to use a generic call instead.

PR #21056, opened 2025-10-24
[ty] Avoid duplicate diagnostics during multi-inference of standalone expressions

ibraheemdev:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1428.

### DIFF
--- a/crates/ty_python_semantic/src/types/context.rs
+++ b/crates/ty_python_semantic/src/types/context.rs
@@ -97,7 +97,9 @@ impl<'db, 'ast> InferContext<'db, 'ast> {
     }
 
     pub(crate) fn extend(&mut self, other: &TypeCheckDiagnostics) {
-        self.diagnostics.get_mut().extend(other);
+        if !self.is_in_multi_inference() {
+            self.diagnostics.get_mut().extend(other);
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/context.rs
```diff
@@ -97,7 +97,9 @@ impl<'db, 'ast> InferContext<'db, 'ast> {
     }
 
     pub(crate) fn extend(&mut self, other: &TypeCheckDiagnostics) {
-        self.diagnostics.get_mut().extend(other);
+        if !self.is_in_multi_inference() {
```
MichaReiser, line 100: Should it be the other way where we don't extend self if other is a multi inference context?

MichaReiser: Okay, I think I figured it out. `other` is guaranteed to be empty if it is a `multi_inference` region. That's why we only have to check self

ibraheemdev: Right, the issue is that if `other` was created using a different context (e.g. because it was executed as a salsa query), we still want to respect to context of `self`.

PR #21059, opened 2025-10-24
[ty] Fix rare panic with highly cyclic `TypeVar` definitions

MichaReiser:
Updates salsa to include a fix for https://github.com/salsa-rs/salsa/pull/1014

## Testing

I tested that ty now "successfully" and consistently crashes with "too many cycle iterations" instead of an internal salsa panic (mismatching iteration counts). 

Since I didn't manage to write a regression test in Salsa, I wrote one in ty instead. I extended mdtest to support tests that expect panics (needed because the example now panics with *too many iterations*) and added a minified version of steam.py. 

This mdtest might also be helpful when debugging why steam.py still panics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3563,7 +3563,7 @@ checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
 [[package]]
 name = "salsa"
 version = "0.24.0"
-source = "git+https://github.com/salsa-rs/salsa.git?rev=ba811277fcb7f1abc9dca8b480f74960184d558a#ba811277fcb7f1abc9dca8b480f74960184d558a"
+source = "git+https://github.com/salsa-rs/salsa.git?rev=25b3ef146cfa2615f4ec82760bd0c22b454d0a12#25b3ef146cfa2615f4ec82760bd0c22b454d0a12"
 dependencies = [
  "boxcar",
  "compact_str",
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/regression/1377_iteration_count_mismatch.md
```diff
@@ -0,0 +1,149 @@
+# Iteration count mismatch for highly cyclic type vars
+
+Regression test for <https://github.com/astral-sh/ty/issues/1377>.
+
+The code is an excerpt from <https://github.com/Gobot1234/steam.py> that is minimal enough to
+trigger the iteration count mismatch bug in Salsa.
+
+<!-- expect-panic: execute: too many cycle iterations -->
```
AlexWaygood, line 8: haha, nice feature! This might be useful to track the remaining panics in https://github.com/astral-sh/ty/issues/256

crates/ty_test/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
+
```
AlexWaygood: is this necessary?

crates/ty_python_semantic/resources/mdtest/regression/1377_iteration_count_mismatch.md
AlexWaygood, line 1: Playing around with this locally, I can make this test quite a lot more minimal while still triggerin the `too many cycle iterations` panic. But I'm guessing you can no longer repro the Salsa bug on the more minimal versions of this test?

MichaReiser: Yes, removing any line immediately goes from the panic that I'm fixing in salsa to `too many cycle iterations`. 

I suspect that it might be possible to inline some `TypeVar`s (reduce some indirection) and still get the same outcome but I decided that it's already pretty good and not to spend more time (2h!) on minimizing

crates/ty_test/src/lib.rs
MichaReiser: na. Thanks

crates/ty_python_semantic/resources/mdtest/regression/1377_iteration_count_mismatch.md
AlexWaygood: oh yeah, for sure. This is fine as-is

PR #21063, opened 2025-10-24
[ty] Add `--no-progress` option

MichaReiser:
## Summary

Add a `--no-progress` option to suppress the progress bar (matching uv's interface)

## Test Plan

I don't think it's possible to test this in our snapshot tests because the progress bar gets hidden for non-interactive terminals.

https://github.com/user-attachments/assets/40468d13-f646-4dc4-85be-04dbdf098248


AlexWaygood (approved):
thanks!

zanieb (approved):

crates/ty/src/printer.rs
```diff
@@ -12,11 +12,10 @@ pub(crate) struct Printer {
 }
 
 impl Printer {
-    #[must_use]
-    pub(crate) fn with_verbosity(self, verbosity: VerbosityLevel) -> Self {
+    pub(crate) fn new(verbosity: VerbosityLevel, no_progress: bool) -> Self {
```
zanieb, line 15: crushed that you changed my API :)

MichaReiser: Sorry. I still very much like the `Printer` API overall. It's nice to have something that's responsible to make these kind of decisions. I hope that makes up for my API change

PR #21067, opened 2025-10-24
Fix finding keyword range for clause header after statement ending with semicolon

### LINKED ISSUE #21065: Formatter panic when range formatting certain clause headers with semicolons
### Summary

Given `ex.py` with content:

```python
x =7
try:
    a;
finally:
    b
```

We get:

```console
❯ ruff format --range=2 --diff ex.py
error: Failed to format tarmin.py: syntax error: Expected the keyword token but found another token instead.
```

This only happens with range formatting, since it comes from this call of `self.range`:

https://github.com/astral-sh/ruff/blob/eb8c0ad87c4e10e2d66ada5e4bb7f436214ca9a8/crates/ruff_python_formatter/src/statement/clause.rs#L368-L373

The `SimpleTokenizer` finding the keyword range is expecting to find `Finally` but instead it finds the (unnecessary) semi-colon.

### Version

_No response_

dylwil3:
When formatting clause headers for clauses that are not their own node, like an `else` clause or `finally` clause, we begin searching for the keyword at the end of the previous statement. However, if the previous statement ended in a semicolon this caused a panic because we only expected trivia between the end of the last statement and the keyword.

This PR adjusts the starting point of our search for the keyword to begin after the optional semicolon in these cases.

Closes #21065

### DIFF
--- a/crates/ruff_python_formatter/src/statement/clause.rs
+++ b/crates/ruff_python_formatter/src/statement/clause.rs
@@ -216,7 +216,11 @@ impl ClauseHeader<'_> {
                     .decorator_list
                     .last()
                     .map_or_else(|| header.start(), Ranged::end);
-                find_keyword(start_position, SimpleTokenKind::Class, source)
+                find_keyword(
+                    StartPosition::ClauseStart(start_position),
+                    SimpleTokenKind::Class,
+                    source,
+                )
             }
             ClauseHeader::Function(header) => {
                 let start_position = header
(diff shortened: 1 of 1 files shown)


crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -434,7 +444,7 @@ impl Format<PyFormatContext<'_>> for FormatClauseBody<'_> {
     }
 }
 
-/// Finds the range of `keyword` starting the search at `start_position`. Expects only comments and `(` between
+/// Finds the range of `keyword` starting the search at `start_position`. Expects only trivia between
```
dylwil3: This isn't related to this PR but this doc-comment looked incorrect to me - there was nothing in the code that allowed an open parentheses, as far as I could tell.

crates/ruff_python_formatter/src/statement/clause.rs
```diff
@@ -466,6 +476,18 @@ fn find_keyword(
     }
 }
 
+fn after_optional_semicolon(end_of_statement: TextSize, source: &str) -> TextSize {
+    let mut tokenizer = SimpleTokenizer::starts_at(end_of_statement, source);
```
MichaReiser: I'm inclined to instead add an argument to `find_keyword` to force every caller to think about whether the body can contain a semicolon or not that may need to be skipped. It also avoids the need to lex the token at `end_of_statement` twice in the common case, just so that we can skip the semi.

The way I'd do it is to add a new `enum SkipSemicolon` which also gives us the opportunity to document when a semicolon needs to be skipped.

A maybe simpler API would be to change the `start_position` of `find_keywrod` to an enum `StartOffset` that has two variants

* ClauseStart(TextSize)
* LastStatement(TextSize)


The advantage of this API is that it doesn't introduce more arguments and it also allows `find_keyword` to just do the right thing. 

This might even allow us to remove some of the `unwrap` by adding a `StartOffset::end_of_last_statement(&suite)` and `StartOffset::clause_start(impl Ranged)`

dylwil3: Nice! I added `StartOffset::clause_start(impl Ranged)` but not `StartOffset::end_of_last_statement(&suite)` because it looks like each of the "last statement" examples required some custom logic depending on the statement type. I removed `after_optional_semicolon` and replaced it with inlined logic that does not re-lex the keyword in the common case.

PR #21068, opened 2025-10-24
[ty] Update "constraint implication" relation to work on constraints between two typevars

dcreager:
It's possible for a constraint to mention two typevars. For instance, in the body of

```py
def f[S: int, T: S](): ...
```

the baseline constraint set would be `(T ≤ S) ∧ (S ≤ int)`. That is, `S` must specialize to some subtype of `int`, and `T` must specialize to a subtype of the type that `S` specializes to.

This PR updates the new "constraint implication" relationship from #21010 to work on these kinds of constraint sets. For instance, in the example above, we should be able to see that `T ≤ int` must always hold:

```py
def f[S, T]():
    constraints = ConstraintSet.range(Never, S, int) & ConstraintSet.range(Never, T, S)
    static_assert(constraints.implies_subtype_of(T, int))  # now succeeds!
```

This did not require major changes to the implementation of `implies_subtype_of`. That method already relies on how our `simplify` and `domain` methods expand a constraint set to include the transitive closure of the constraints that it mentions, and to mark certain combinations of constraints as impossible. Previously, that transitive closure logic only looked at pairs of constraints that constrain the same typevar. (For instance, to notice that `(T ≤ bool) ∧ ¬(T ≤ int)` is impossible.)

Now we also look at pairs of constraints that constraint different typevars, if one of the constraints is bound by the other — that is, pairs of the form `T ≤ S` and `S ≤ something`, or `S ≤ T` and `something ≤ S`. In those cases, transitivity lets us add a new derived constraint that `T ≤ something` or `something ≤ T`, respectively. Having done that, our existing `implies_subtype_of` logic finds and takes into account that derived constraint.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -753,34 +753,24 @@ impl<'db> Node<'db> {
         rhs: Type<'db>,
         inferable: InferableTypeVars<'_, 'db>,
     ) -> Self {
-        match (lhs, rhs) {
-            // When checking subtyping involving a typevar, we project the BDD so that it only
-            // contains that typevar, and any other typevars that could be its upper/lower bound.
-            // (That is, other typevars that are "later" in our arbitrary ordering of typevars.)
-            //
-            // Having done that, we can turn the subtyping check into a constraint (i.e, "is `T` a
-            // subtype of `int` becomes the constraint `T ≤ int`), and then check when the BDD
-            // implies that constraint.
+        // When checking subtyping involving a typevar, we can turn the subtyping check into a
+        // constraint (i.e, "is `T` a subtype of `int` becomes the constraint `T ≤ int`), and then
+        // check when the BDD implies that constraint.
+        let constraint = match (lhs, rhs) {
             (Type::TypeVar(bound_typevar), _) => {
-                let constraint = ConstrainedTypeVar::new_node(db, bound_typevar, Type::Never, rhs);
-                let simplified_self = self.simplify(db);
-                let implication = simplified_self.implies(db, constraint);
-                let (simplified, domain) = implication.simplify_and_domain(db);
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
```diff
@@ -180,16 +180,12 @@ This might require propagating constraints from other typevars.
...
     # If [T = U ∧ U ≤ int], then [T ≤ int] must be true as well.
     given_int = ConstraintSet.range(U, T, U) & ConstraintSet.range(Never, U, int)
-    # TODO: no static-assert-error
-    # error: [static-assert-error]
     static_assert(given_int.implies_subtype_of(T, int))
     static_assert(not given_int.implies_subtype_of(T, bool))
     static_assert(not given_int.implies_subtype_of(T, str))
 
     # If [T ≤ U ∧ U ≤ int], then [T ≤ int] must be true as well.
     given_int = ConstraintSet.range(Never, T, U) & ConstraintSet.range(Never, U, int)
```
MichaReiser, line 188: More an understand question for me. What's the difference between `Never` and passing `U`?

MichaReiser (approved):
Makes sense to me

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
dcreager: `range(Never, T, U)` is `Never ≤ T ≤ U`, which simplifies to `T ≤ U` since every type is a supertype of `Never`.

`range(U, T, U)` is `U ≤ T ≤ U`, which simplifies to `T = U`.

So the first one is "`T` must be a subtype of `U`", and the second one is "`T` must be exactly the same type as `U`"

carljm (approved):

PR #21069, opened 2025-10-24
[`ruff`] Suppress diagnostic for strings with backslashes in interpolations before Python 3.12 (`RUF027`)

### LINKED ISSUE #21033: RUF027 creates invalid f-string interpolations with backslashes before Python 3.12
### Summary

[`missing-f-string-syntax` (RUF027)](https://docs.astral.sh/ruff/rules/missing-f-string-syntax/) applies to putative interpolations containing backslashes, which are not valid until Python 3.12. The fix can therefore introduce syntax errors. RUF027 should probably be suppressed for such strings before Python 3.12. [Example](https://play.ruff.rs/9e3243a1-379b-4ae2-ab31-1355d71b85c8):
```console
$ cat >ruf027.py <<'# EOF'
x = "!"
print("{'\''}{x\
}")
# EOF

$ python3.11 ruf027.py
{'''}{x}

$ ruff --isolated check ruf027.py --select RUF027 --preview --target-version py311 --unsafe-fixes --fix --output-format concise
ruf027.py:2:11: invalid-syntax: Cannot use an escape sequence (backslash) in f-strings on Python 3.11 (syntax was added in Python 3.12)
ruf027.py:2:17: invalid-syntax: Cannot use an escape sequence (backslash) in f-strings on Python 3.11 (syntax was added in Python 3.12)
Found 3 errors (1 fixed, 2 remaining).

$ python3.11 ruf027.py 2>&1 | tail -n 1
SyntaxError: f-string expression part cannot include a backslash
```

### Version

danparizher:
## Summary
Fixes #21033 (RUF027) to suppress f-string suggestions for strings containing backslashes in interpolation expressions when targeting Python versions below 3.12, preventing syntax errors.

## Problem Analysis
RUF027 was suggesting f-string conversions for strings with backslashes in interpolation expressions (e.g., `"{'\\n'}"`), but f-strings with backslashes in interpolations are only valid syntax in Python 3.12+. This caused syntax errors when the fix was applied to code targeting older Python versions.

The issue occurred because the `should_be_fstring` function was checking for backslashes anywhere in the string literal, rather than specifically within the interpolation expressions where the syntax restriction applies.

## Approach
Added a version check in the `should_be_fstring` function to suppress RUF027 for strings containing backslashes within interpolation expressions when the target Python version is below 3.12. This ensures:

- For Python < 3.12: RUF027 is suppressed for strings with backslashes in interpolations (e.g., `"{'\\n'}{x}"`)
- For Python ≥ 3.12: RUF027 works normally for strings with backslashes in interpolations

### DIFF
--- a/crates/ruff_linter/src/rules/ruff/rules/missing_fstring_syntax.rs
+++ b/crates/ruff_linter/src/rules/ruff/rules/missing_fstring_syntax.rs
@@ -229,7 +229,7 @@ fn should_be_fstring(
             // Check if the interpolation expression contains backslashes
             // F-strings with backslashes in interpolations are only valid in Python 3.12+
             let interpolation_text = &fstring_expr[element.range()];
-            if interpolation_text.contains('\\') && target_version < PythonVersion::PY312 {
+            if target_version < PythonVersion::PY312 && interpolation_text.contains('\\') {
                 return false;
             }
 


danparizher:
The ecosystem check looks to be accurate; all 3 repositories are all <3.12.

dscorbett:
Backslashes in f-strings aren’t the problem.
```pycon
>>> f"\n"
'\n'
```
The problem is backslashes in f-string interpolations.
```pycon
>>> f"{'\n'}"
  File "<stdin>", line 1
    f"{'\n'}"
             ^
SyntaxError: f-string expression part cannot include a backslash
```

ntBre:
I just worked on two issues for our syntax error related to this: https://github.com/astral-sh/ruff/pull/20867 and https://github.com/astral-sh/ruff/pull/20949. I think the fix here is similar to the incorrect code I had originally in the parser, which turned out to be too naive in the presence of nested interpolations. To summarize the two issues, code like this is valid on 3.11:

```pycon
>>> f"{1:\"d\"}"
ValueError: Invalid format specifier '"d"' for object of type 'int'
```

Note the `ValueError`, not an f-string syntax error. It looks like the code here would flag this because it's within the interpolation but not within the interpolation _expression_.

In contrast, code like this is still invalid on 3.11 because it's in the expression part of a nested interpolation:

```pycon
>>> f'{1: abcd "{'aa'}" }'
  File "<stdin>", line 1
    f'{1: abcd "{'aa'}" }'
                  ^^
SyntaxError: f-string: expecting '}'
```

But the `interpolations` iterator doesn't recurse into nested interpolations, so the fix is also not as easy as only checking the expression part (that's what I got wrong in #20867).

It might be fair to be more conservative here since it's almost definitely easier to implement, but I wanted to mention this since it was relevant.

crates/ruff_linter/src/rules/ruff/rules/missing_fstring_syntax.rs
```diff
@@ -180,6 +180,7 @@ fn should_be_fstring(
     literal: &ast::StringLiteral,
     locator: &Locator,
     semantic: &SemanticModel,
+    checker: &Checker,
```
ntBre: I think we could drop the `locator` and `semantic` arguments if we're passing the `Checker` they both come from. Alternatively, we could pass only the `target_version`, which is the only additional field we're accessing now.

crates/ruff_linter/src/rules/ruff/snapshots/ruff_linter__rules__ruff__tests__RUF027_RUF027_0.py.snap
ntBre, line 1: We should add a variant of this test where we use a Python version before 3.12. This is still showing a diagnostic being emitted.

PR #21070, opened 2025-10-25
[ty] Prefer declared type of generic classes

ibraheemdev:
## Summary

Extends https://github.com/astral-sh/ruff/pull/20927 to generic call inference, preferring the declared type of generic classes:

```py
def f[T](x: T) -> list[T]:
    return [x]

x: list[Any] = f(1)
reveal_type(x)  # before: list[Literal[1]], after: list[Any]
```

Part of https://github.com/astral-sh/ty/issues/136.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -427,14 +427,13 @@ a = f("a")
 reveal_type(a)  # revealed: list[Literal["a"]]
 
 b: list[int | Literal["a"]] = f("a")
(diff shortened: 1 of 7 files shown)


ibraheemdev:
I'm not sure if we should restrict this change to invariant generics. It seems reasonable to use `frozendict[str, Any]` as an implicit typed dict, so preferring the `Any` there might still be meaningful (even thought it feels strictly worse).

carljm:
@ibraheemdev Is this blocked? What's keeping it in draft?

ibraheemdev:
The ecosystem changes all seem correct.

ibraheemdev:
Superceded by https://github.com/astral-sh/ruff/pull/21210.

PR #21071, opened 2025-10-25
[ty] Infer arguments of generic calls with declared type context

ibraheemdev:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1356.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -476,10 +476,8 @@ def _(i: int):
     b: list[int | None] | None = id([i])
     c: list[int | None] | int | None = id([i])
     reveal_type(a)  # revealed: list[int | None]
-    # TODO: these should reveal `list[int | None]`
-    # we currently do not use the call expression annotation as type context for argument inference
-    reveal_type(b)  # revealed: list[Unknown | int]
-    reveal_type(c)  # revealed: list[Unknown | int]
+    reveal_type(b)  # revealed: list[int | None]
+    reveal_type(c)  # revealed: list[int | None]
 
     a: list[int | None] | None = [i]
     b: list[int | None] | None = lst(i)
@@ -494,4 +492,27 @@ def _(i: int):
     reveal_type(a)  # revealed: list[Unknown]
     reveal_type(b)  # revealed: list[Unknown]
     reveal_type(c)  # revealed: list[Unknown]
+```
+
+The function arguments are inferred using the type context:
+
+```py
+from typing import TypedDict
+
+class TD(TypedDict):
+    x: int
+
+def f[T](x: list[T]) -> T:
+    return x[0]
+
+a: TD = f([{"x": 0}, {"x": 1}])
+reveal_type(a)  # revealed: TD
+
+b: TD | None = f([{"x": 0}, {"x": 1}])
+# TODO: Narrow away the `None` here.
+reveal_type(b)  # revealed: TD | None
+
+# error: [missing-typed-dict-key] "Missing required key 'x' in TypedDict `TD` constructor"
+# error: [invalid-key] "Invalid key access on TypedDict `TD`: Unknown key "y""
+c: TD | None = f([{"y": 0}, {"x": 1}])
 ```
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -495,3 +493,26 @@ def _(i: int):
...
+
+def f[T](x: list[T]) -> T:
+    return x[0]
+
+a: TD = f([{ "x": 0 }, { "x": 1 }])
+reveal_type(a)  # revealed: TD
+
+b: TD | None = f([{ "x": 0 }, { "x": 1 }])
+# TODO: Narrow away the `None` here.
+reveal_type(b)  # revealed: TD | None
```
ibraheemdev, line 513: This is really tricky. We infer the list argument to be `list[TD | None]` due to the annotated type of `T` being `TD | None`. I'm not sure there's a clear heuristic to filter away the `None` here (and similarly for any other type than `None`).

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5603,10 +5604,28 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        SpecializationBuilder::new(db, generic_context.inferable_typevars(db));
+
+                    let _ = builder.infer(return_ty, declared_return_ty);
+                    let specialization = builder.build(generic_context, call_expression_tcx);
+
+                    // Note that we are not necessarily "preferring the declared type" here, as the
+                    // type context will only be preferred during the inference of this expression
+                    // by the same heuristics we use for the inference of the outer generic call.
+                    parameter_type = parameter_type.apply_specialization(db, specialization);
+                }
```
dcreager, line 5626: If I'm following the control flow correctly, this has to happen here because we haven't yet inferred a specialization in the `Bindings`, since that doesn't happen until you call `check_types`. And we need this specialization applied to the parameter type in order to have a type context to use when infering the argument types. Does that sound right?

My worry here is that this specialization that you're building isn't guaranteed to line up with the one that `check_types` will infer — or at least, it won't once we have the better constraint solver, where the inferred specialization can be influenced by constraints coming from e.g. other arguments.

Ideally we'd find a way to be able to reuse the specialization that `Bindings` is already building up for us. Should we consider moving the specialization inference step so that it happens at the end of `match_parameters`, instead of at the beginning of `check_types`? Or as an explicit third step in between the two?

That might be something we can queue up as follow-on work, in order to land this. But I'd like to at least think through what would be involved, since I think might add more blockers to getting the new constraint solver landed.

ibraheemdev: > If I'm following the control flow correctly, this has to happen here because we haven't yet inferred a specialization in the Bindings, since that doesn't happen until you call check_types. And we need this specialization applied to the parameter type in order to have a type context to use when infering the argument types. Does that sound right?

Yeah that's correct.

> Ideally we'd find a way to be able to reuse the specialization that Bindings is already building up for us. Should we consider moving the specialization inference step so that it happens at the end of match_parameters, instead of at the beginning of check_types? Or as an explicit third step in between the two?

I'm not exactly sure how that would work. Constructing the complete binding specialization is only possible once we have the types of all arguments, which we want to infer using the type context — so there's a bit of a circular dependency here.

Consider:
```py
def f[T](x: list[T], y: T) -> list[T]:
    return x

def lst[T](x: T) -> list[T]:
    return [x]

def _(x: int, y: int | str):
    z: list[int | str] = f(lst(x), y) # error: expected list[int | str], found list[int]
```

In order to infer a specialization for `f`, we must first infer each of its arguments. The call to `lst(x)` has no way of specializing to `int | str` except by using the type-context `list[int | str]` to infer a *more* assignable type, such that it passes `check_types` once the complete specialization is inferred. Note that without the type-context, pyright also fails to type-check, which suggests it's using the type context in a similar way to this PR.

The question then is, can using the extra type-context lead to a *less* assignable type (as I think you are suggesting)? I'm not sure, as the specialized return type must be assignable to annotated type of the call expression. The only issue I see is the question of how to narrow the type context such that we don't overly widen the inferred type of the call, e.g.
```py
def f[T](x: list[T]) -> T:
    return x[0]

def lst[T](x: T) -> list[T]:
    return [x]

def _(x: int, y: int | str):
    z: int | str = f(lst(x))
    reveal_type(z) # should be int, not int | str
```

ibraheemdev:
Superceded by https://github.com/astral-sh/ruff/pull/21210.

Issue #19028, opened 2025-06-29
INT rules have false negatives for other ways to use gettext functions


dscorbett:
### Summary

[`f-string-in-get-text-func-call` (INT001)](https://docs.astral.sh/ruff/rules/f-string-in-get-text-func-call/), [`format-in-get-text-func-call` (INT002)](https://docs.astral.sh/ruff/rules/format-in-get-text-func-call/), and [`printf-in-get-text-func-call` (INT003)](https://docs.astral.sh/ruff/rules/printf-in-get-text-func-call/) do not recognize internalization functions when written as attribute references, e.g. `gettext.ngettext` instead of `ngettext`. This is a false negative because tools like xgettext recognize attribute references too.

These rules do not recognize internalization functions when bound to different names, e.g. `gettext.ngettext` imported as `N_` instead of `ngettext`. This is a false negative because if someone writes such a binding, they probably also specify it as a gettext keyword, e.g. with `xgettext -kN_:1,2`, even if they don’t configure Ruff with [`function-names`](https://docs.astral.sh/ruff/settings/#lint_flake8-gettext_function-names) or [`extend-function-names`](https://docs.astral.sh/ruff/settings/#lint_flake8-gettext_extend-function-names) explicitly.

[False negatives](https://play.ruff.rs/d54ecf1a-f020-4294-823a-d33ebcccaffb):
```python
import builtins
import gettext
from gettext import gettext as _gettext
gettext.install("int_fn")
name = "world"

print(builtins._(f"Hello, {name}!"))
print(gettext.gettext(f"Hello, {name}!"))
print(_gettext(f"Hello, {name}!"))

print(builtins._("Hello, {}!".format(name)))
print(gettext.gettext("Hello, {}!".format(name)))
print(_gettext("Hello, {}!".format(name)))

print(builtins._("Hello, %s!" % name))
print(gettext.gettext("Hello, %s!" % name))
print(_gettext("Hello, %s!" % name))
```

### Version

ruff 0.12.1 (32c54189c 2025-06-26)

ntBre:
It looks like these rules just check for a `Name` expression and compare the name directly instead of matching the qualified name, for example:

https://github.com/astral-sh/ruff/blob/e7aadfc28b5a1447c45d39017a37f1acd1001f14/crates/ruff_python_semantic/src/analyze/imports.rs#L73-L77

[...]

PR #21078, opened 2025-10-26
[`flake8-gettext`] Fix false negatives for plural argument of ngettext (`INT001`, `INT002`, `INT003`)

### LINKED ISSUE #21062: INT false negatives for the plural argument of `ngettext`
### Summary

[`f-string-in-get-text-func-call` (INT001)](https://docs.astral.sh/ruff/rules/f-string-in-get-text-func-call/), [`format-in-get-text-func-call` (INT002)](https://docs.astral.sh/ruff/rules/format-in-get-text-func-call/), and [`printf-in-get-text-func-call` (INT003)](https://docs.astral.sh/ruff/rules/printf-in-get-text-func-call/) only check the first argument (singular) of `ngettext` but skip the second (plural). The reasons to avoid f-strings etc. in the first argument apply equally to the second. This is true whether or not `ngettext` is imported, and if so whether it is aliased. [Example](https://play.ruff.rs/53b13c7c-340d-4f64-9e24-59ed0831d2ef):
```console
$ cat >int00.py <<'# EOF'
from gettext import ngettext
def f(n):
    ngettext("one", f"{n}", n)
    ngettext("one", "{}".format(n), n)
    ngettext("one", "%d" % n, n)
# EOF

$ ruff --isolated check int00.py --select INT001,INT002,INT003
All checks passed!

$ sed '/import/d' int00.py | ruff --isolated check - --select INT001,INT002,INT003
All checks passed!

$ sed '/import/s/$/ as n_/; s/ngettext(/n_(/' int00.py | ruff --isolated check - --select INT001,INT002,INT003
All checks passed!
```

### Version

danparizher:
## Summary
Fixes #21062. Extends the `INT001`, `INT002`, and `INT003` lint rules to check both the singular and plural arguments of `ngettext` function calls, fixing false negatives where formatting issues in the plural argument were not being detected.

## Problem Analysis
The three rules (`f-string-in-get-text-func-call`, `format-in-get-text-func-call`, and `printf-in-get-text-func-call`) only checked the first argument (singular) of `ngettext` calls but ignored the second argument (plural). This caused false negatives when f-strings, `.format()` calls, or `%` formatting were used in the plural argument.

The root cause was:
1. Rule functions only examined `args.first()` 
2. No logic to determine if a function call was specifically `ngettext` vs `gettext`
3. Missing checks for `args.get(1)` (the second argument)

Example of the bug:
```python
ngettext(f"one item", f"{n} items", n)  # Only flagged first f-string, not second
```

## Approach
1. **Modified function signatures**: Updated all three rule functions to accept the `func` parameter
2. **Added ngettext detection**: Created `is_ngettext_call()` helper that checks:
   - Direct name references (e.g., `ngettext(...)`)
   - Qualified names ending with `ngettext` (e.g., `gettext_mod.ngettext(...)`)
3. **Extended argument checking**: For `ngettext` calls, now check both first and second arguments
4. **Code refactoring**: Created shared `helpers.rs` module to eliminate duplication
5. **Updated main checker**: Modified expression analyzer to pass `func` parameter

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_gettext/rules/f_string_in_gettext_func_call.rs
+++ b/crates/ruff_linter/src/rules/flake8_gettext/rules/f_string_in_gettext_func_call.rs
@@ -73,14 +73,10 @@ pub(crate) fn f_string_in_gettext_func_call(checker: &Checker, func: &Expr, args
     }
 
     // Check second argument (plural) for ngettext calls
-    if is_ngettext_call(checker, func) {
-        if let Some(second) = args.get(1) {
-            if second.is_f_string_expr() {
-                checker.report_diagnostic(
-                    FStringInGetTextFuncCall { is_plural: true },
-                    second.range(),
-                );
-            }
-        }
+    if is_ngettext_call(checker, func)
+        && let Some(second) = args.get(1)
+        && second.is_f_string_expr()
+    {
+        checker.report_diagnostic(FStringInGetTextFuncCall { is_plural: true }, second.range());
     }
 }
--- a/crates/ruff_linter/src/rules/flake8_gettext/rules/format_in_gettext_func_call.rs
+++ b/crates/ruff_linter/src/rules/flake8_gettext/rules/format_in_gettext_func_call.rs
@@ -1,4 +1,4 @@
-use ruff_python_ast::{self as ast, Expr};
+use ruff_python_ast::Expr;
 
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_text_size::Ranged;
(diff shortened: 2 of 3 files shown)


crates/ruff_linter/src/rules/flake8_gettext/rules/f_string_in_gettext_func_call.rs
```diff
@@ -52,10 +53,20 @@ impl Violation for FStringInGetTextFuncCall {
...
+pub(crate) fn f_string_in_gettext_func_call(checker: &Checker, func: &Expr, args: &[Expr]) {
+    // Check first argument (singular)
     if let Some(first) = args.first() {
         if first.is_f_string_expr() {
             checker.report_diagnostic(FStringInGetTextFuncCall {}, first.range());
         }
     }
+
+    // Check second argument (plural) for ngettext calls
+    if helpers::is_ngettext_call(checker, func) {
```
ntBre: See my other comment about the `ngettext` validation. If we avoid that check, I think we can just make each of these `if let Some(first)` checks into loops:

```rust
for arg in args[..2] {
    if arg.is_f_string_expr() {
        checker.report_diagnostic(...);
    }
}
```

since as far as I can tell the insides of the conditionals are exactly the same.

crates/ruff_linter/src/rules/flake8_gettext/snapshots/ruff_linter__rules__flake8_gettext__tests__preview__printf-in-get-text-func-call_INT003.py.snap
```diff
@@ -160,3 +190,12 @@ INT003 printf-style format is resolved before function call; consider `_("string
...
    |                   ^^^^^^^^^^^^^
    |
+
+INT003 printf-style format is resolved before function call; consider `_("string %s") % arg`
+  --> INT003.py:49:34
+   |
+47 | builtins._("%s" % "line")
+48 | builtins.gettext("%s" % "line")
+49 | builtins.ngettext("%s" % "line", "%s" % "lines", 2)
+   |                                  ^^^^^^^^^^^^^^
```
ntBre: I think we have to customize the error messages depending on whether it's the first argument or the second. The `consider ...` suggestion doesn't make much sense here.

crates/ruff_linter/src/rules/flake8_gettext/helpers.rs
```diff
@@ -0,0 +1,21 @@
+use crate::checkers::ast::Checker;
+use ruff_python_ast::Expr;
+
+/// Returns true if the function call is ngettext
+pub(crate) fn is_ngettext_call(checker: &Checker, func: &Expr) -> bool {
```
ntBre: We have an existing helper in `flake8_gettext/mod.rs`. I kind of think we should keep them together, either by moving this there or moving the other helper here. I don't have a strong preference but would slightly lean toward reusing the existing file.

crates/ruff_linter/src/rules/flake8_gettext/helpers.rs
```diff
@@ -0,0 +1,21 @@
...
+    if let Some(name) = func.as_name_expr() {
+        if name.id == "ngettext" {
+            return true;
+        }
+    }
+
+    // Check if it's a qualified name ending with ngettext
+    if let Some(qualified_name) = semantic.resolve_qualified_name(func) {
+        return matches!(qualified_name.segments(), [.., "ngettext"]);
+    }
```
ntBre: We should double check if these two are redundant. I would expect `resolve_qualified_name` on a name to return `["ngettext"]`, which should pass the second check here, but I could be wrong. Checking the name might be a small perf optimization since the qualified name can be expensive, but I'd still be pretty tempted to turn the whole function into

```rust
checker.semantic().resolve_qualified_name(func).is_some_and(|qualified_name|
    matches!(qualified_name, [.., "ngettext"])
)
```

if it works.

MichaReiser:
Is it worth if we try to push this PR over the line, given that the changes seem pretty minimal

danparizher:
I can take a look this weekend

crates/ruff_linter/src/rules/flake8_gettext/helpers.rs
ntBre: I double checked this and can confirm that this is needed to support the case where `ngettext` is dynamically added to `builtins`. See https://github.com/astral-sh/ruff/issues/19028, for example.

ntBre (approved):
Thanks! I pushed a couple of commits adding the preview gating and factoring out a couple of helper functions to reuse existing code.

Issue #21037, opened 2025-10-23
FAST002: catch Union with Annotated


wilt-vetrec:
### Summary

I ran into an issue in FastAPI recently, where I defined a route with an optional array query parameter, like so:

```python
@router.get("/optional_none_outside")
def optional_none_outside(params: Annotated[list[Literal[1,2]], Query()] | None = None):
    """Incorrect; fails when called like /my_route?params=1 or like /my_route?params=1&params=2"""
    ...

@router.get("/optional_none_inside")
def optional_none_inside(params: Annotated[list[Literal[1,2]] | None, Query()] = None):
    """Correct; succeeds with any number of params"""
    ...
```

Full repro: https://replit.com/@wilt00/fastapi-query-annotation-repro#main.py

I'm not sure why this happens, but my understanding is that this is expected and not a bug in FastAPI. Ruff's lint rule FAST002 did not flag this error, and I think it should. I'm happy to work on adding this check.

ntBre:
I think this would have to be a different rule since FAST002 is only inspecting the default values looking for `Query`, `Depends`, etc. This would require validating the annotation itself and might be pretty tricky to get right.

CoderJoshDK:
After thinking about this more, I am not sure that this needs to be solved as a `FAST` based rule. This should either be a new lint or something a type aware system can handle. 

<sub>Or nuclear option of "this is a bug with fastAPI, and they should fix it"</sub> 

The reason I say that is because if we inspect the following:

```py
from typing import Annotated, get_type_hints

def f(a: Annotated[str, "test data"]): ...
def z(a: Annotated[str, "test data"] | None): ...
def b(a: Annotated[str | None, "test data"]): ...


get_type_hints(f, include_extras=True)
# {'a': typing.Annotated[str, 'test data']}
get_type_hints(z, include_extras=True)
# {'a': typing.Optional[typing.Annotated[str, 'test data']]}
get_type_hints(b, include_extras=True)
# {'a': typing.Annotated[str | None, 'test data']}
```

We can see that `a` gets converted into an optional annotated object. This obfuscates the `Annotated` with a compound type. That type should ideally be contained inside the `typing.Annotated`. IE, this is a bad practice regardless of `fastapi` or not. Very few libraries handles this case properly. And instead of saying all libraries have some bug, it might be easier to say that it is bad practice to obfuscate the annotated type behind some other type (optional in this instance, but union should also be discouraged).

[...]

PR #9911, opened 2024-02-09
rule(`RUF028`) Implement missing await for coroutine

mikeleppane:
## Summary

Detect never-awaited [coroutines](https://docs.python.org/3/library/asyncio-dev.html#detect-never-awaited-coroutines) 

See: https://github.com/astral-sh/ruff/issues/9833

## Test Plan

```bash
cargo test
```

PR #21079, opened 2025-10-26
[ruff] Implement union-with-annotated-type (RUF066)

wilt00:
## Summary

This PR implements a new rule, nested-annotated-type, as discussed in #21037 . This rule fails for type definitions including `Annotated` nested within some other type.

## Test Plan

New snapshosts in `RUF066.py` with `cargo insta test`.


MichaReiser:
I think this rule makes sense to me (catching `Annotated[...] | ...`) and I'd be in favor adding it because it catches a very likely mistake. But I'd be curious to hear @amyreese's opinion, too, just to ensure I'm not overlooking any valid use cases for this.


I do think we should come up with a less cryptic name :). We certainly can't use the nested terminology because the annotated documentation uses that to refer to `Annotated[Annotated]` nesting

> Nested Annotated types are flattened. The order of the metadata elements starts with the innermost annotation:
> https://docs.python.org/3/library/typing.html#typing.Annotated

Maybe something like `type-operation-with-annotated` or `mixed-annotated-regular-type`?

amyreese:
I agree that it's a useful rule — I'm not aware of any case where you would want to use an `Annotated` type with anything, let alone nest `Annotated` inside of anything other than another `Annotated` type.

For names I prefer `mixed-annotated-regular-type` over the other suggestion, but I might instead name it `union-with-annotated-type`

crates/ruff_linter/resources/test/fixtures/ruff/RUF066.py
```diff
@@ -0,0 +1,50 @@
+from typing import Annotated, Optional, TypeAlias, Union
+
+# Errors
+a: Union[Annotated[int, "An integer"], None] = 5
+b: Annotated[Union[int, str], "An integer or string"] | None = "World"
+c: Optional[Annotated[float, "A float"]] = 2.71
```
MichaReiser, line 6: What's the reason why this is an error? 

I'm asking because I like the name `union-with-annotated-type` but the name requires that it only detects usages of `Annotated` with `Union`

amyreese: > What's the reason why this is an error?
> 
> I'm asking because I like the name `union-with-annotated-type` but the name requires that it only detects usages of `Annotated` with `Union`

Assuming you're asking about the `Optional` example, that's functionally equivalent to `Union[Annotated[...], None]`, and it's still treated as a union in Python types. It's just a convenience because `Optional[X]` is more readable than `Union[X, None]`, though once `|` was allowed, everything converged on `X | None` rather than a dedicated syntax for optionals.

wilt00: Exactly, yes. `union-with-annotated-type` is my favorite of the suggested names as well - do you think it's common enough knowledge that optional is a subtype of union for that name to work? I can also clarify that in the diagnostic text.

amyreese: I do think it's common enough knowledge — and the context of the lint rule IMO helps bridge the gap for anyone unfamiliar — and as more projects move from `Optional[X]` to `X | None`, it becomes less and less of a concern.

wilt00: Sounds good. Are there other generic types that don't derive from Union that it makes sense to flag in this rule? Looking down the list of members of typing, Final is the only one that I feel could possibly fit and that's a stretch.

MichaReiser: Ah right, thanks

amyreese: I think it's reasonable to keep the rule relatively small in scope. I also think it's unlikely that someone would add `Final` or other type composition "by accident" the way they might with unions/optionals, not to mention being harder to reason about whether that's intended behavior or a potential bug.

ntBre:
I'll leave the main review to @amyreese, but this looks reasonable overall to me. Would you mind re-numbering the rule to RUF068? We have existing PRs for [RUF066](https://github.com/astral-sh/ruff/pull/9911) and [RUF067](https://github.com/astral-sh/ruff/pull/20585).

crates/ruff_linter/resources/test/fixtures/ruff/RUF066.py
```diff
@@ -0,0 +1,50 @@
...
+def f1(x: Union[Annotated[int, "An integer"], None]) -> None:
+    pass
+
+def f2(y: Annotated[Union[int, str], "An integer or string"] | None) -> None:
+    pass
+
+def f3(z: Optional[Annotated[float, "A float"]]) -> None:
+    pass
+
+e: Annotated[Annotated[int, "An integer"], "Another annotation"]
```
amyreese, line 18: This being an error implies that `list[Annotated[...]]` or similar would also produce an error. That part seems fine to me, but I wonder if it would be worth having a separate error message for these cases since the mention of union/optional doesn't make sense here? Would also like to have test cases showing that behavior one way or the other.

crates/ruff_linter/src/rules/ruff/rules/union_with_annotated_type.rs
```diff
@@ -0,0 +1,94 @@
...
+/// ```
+/// This fails to parse `param` as a query parameter. Use instead:
+/// ```python
+/// @app.get('/route')
+/// def route(param: Annotated[str | None, Query()] = None):
+///   ...
+/// ```
+///
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "0.14.4")]
```
amyreese, line 49: Needs a bump as well 😅

amyreese (changes requested):

CoderJoshDK:
Hey, I see this PR has not had any movement in the last month. If you don't mind, I think I am going to pick it up. I will include the original commits (rebased with main) and update based on the included comments. I hope I am not stepping on your toes with this.

ntBre:
Thank you for working on this! I'll close the PR for now as it's been stale for a couple of months. There has also been a bit more discussion on the rule's scope on the linked issue.

Issue #18409, opened 2025-06-01
[FURB] Path write


gbaian10:
## Case 1
```py
with open("hello.txt", "w", encoding="utf-8") as f:
    f.write("Hello, world!")
```

run `ruff check`

```console
PTH123 `open()` should be replaced by `Path.open()`
FURB103 `open` and `write` should be replaced by `Path("hello.txt").write_text("Hello, world!")`
```

## Case 2

If I try to fix `FURB103` first

```py
from pathlib import Path

Path("hello.txt").write_text("Hello, world!", encoding="utf-8")
```

run `ruff check`

```console
All checks passed!
```

## Case 3

But if we fix PTH123 first

```py
from pathlib import Path

with Path("hello.txt").open("w", encoding="utf-8") as f:
    f.write("Hello, world!")
```

run `ruff check`

```console
All checks passed!
```

---

Are Case3 and Case2 completely equivalent programs?
If they are, I would like to have a rule that can transform Case3 into Case2.

Similar to [unspecified-encoding (PLW1514)](https://docs.astral.sh/ruff/rules/unspecified-encoding/), will raise warnings for both `with open` and `Path.open`.

MichaReiser:
I think it would make sense to extend `PTH123` to also support `Path::open`

11happy:
@MichaReiser I have linked the PR, &  I assume you meant FURB103  as it more justifed to extend that rather than PTH123 ?

> I think it would make sense to extend `PTH123` to also support `Path::open`

PR #21080, opened 2025-10-26
[`refurb`] Extend support for `Path.open` (`FURB101`, `FURB103`)

11happy:
## Summary

This PR fixes https://github.com/astral-sh/ruff/issues/18409

## Test Plan

I have added tests in FURB103.


11happy:
@MichaReiser would appreciate your feedback on this whenever you get some time.

Thank you : )

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__preview_FURB103_FURB103.py.snap
ntBre, line 1: Oh, I didn't notice this when reviewing #20520, but `FURB101` and `FURB103` are already preview rules, so we don't actually need the additional preview checks on their fixes (or these preview tests/snapshot files). Feel free to follow up on that, I think it's fine to leave it for a separate PR, though!

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__preview_FURB103_FURB103.py.snap
```diff
@@ -257,4 +257,169 @@ help: Replace with `Path("file.txt").write_text(foobar, newline="\r\n")`
...
+72  | 
+73  | 
+74  | # FURB103
+--------------------------------------------------------------------------------
+147 |     for line in text:
+148 |         f.write(line)
+149 | 
+    - with Path("file.txt").open("w") as f:
+    -     f.write("test")
+150 + Path("file.txt").write_text("test")
```
ntBre: I don't think we need to add the pathlib import for cases where the input was already a `Path`. We're just adding another method call onto the existing `Path`.

Ah, this might actually be highlighting a different issue. If `pathlib.Path` hasn't been imported, we shouldn't be applying the rule since this could be something else that just happens to be named `Path`.

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -238,6 +243,79 @@ fn find_file_open<'a>(
...
+    } = item.context_expr.as_call_expr()?;
+    let attr = func.as_attribute_expr()?;
+    if attr.attr.as_str() != "open" {
+        return None;
+    }
+    let path_call = attr.value.as_call_expr()?;
+    let path_func = path_call.func.as_name_expr()?;
+    if path_func.id.as_str() != "Path" {
+        return None;
+    }
```
ntBre: Ah yeah, I think we need to be resolving the qualified name here. We may also be able to use our existing pathlib typing check:

https://github.com/astral-sh/ruff/blob/16b9b92182b7457478281e0cdfe4d14c74f33d13/crates/ruff_python_semantic/src/analyze/typing.rs#L1120-L1124

https://github.com/astral-sh/ruff/blob/16b9b92182b7457478281e0cdfe4d14c74f33d13/crates/ruff_python_semantic/src/analyze/typing.rs#L925-L928

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -238,6 +243,79 @@ fn find_file_open<'a>(
         mode,
         keywords,
         reference,
+        path_obj: None,
+    })
+}
+
+fn find_path_open<'a>(
```
ntBre, line 293: It feels like we should probably be able to reuse some more code from `find_file_open` instead of duplicating the whole thing, but maybe they're different enough that that's not true.

crates/ruff_linter/src/rules/refurb/rules/write_whole_file.rs
```diff
@@ -57,7 +57,13 @@ impl Violation for WriteWholeFile {
...
         let filename = self.filename.truncated_display();
         let suggestion = self.suggestion.truncated_display();
-        format!("`open` and `write` should be replaced by `Path({filename}).{suggestion}`")
+        if self.is_path_open {
+            format!(
+                "`Path.open()` followed by `write()` can be replaced by `Path({filename}).{suggestion}`"
+            )
+        } else {
+            format!("`open` and `write` should be replaced by `Path({filename}).{suggestion}`")
+        }
```
ntBre: I don't feel too strongly about this, but I kind of think the old message applies well enough to both cases and we could just leave it alone and not need to track `is_path_open` in the `Violation` struct.

ntBre (changes requested):
Thanks for working on this!

I think we need to be a bit more careful with how we identify existing `Path` instances and hopefully we can deduplicate some code too.

amyreese:
It might be helpful to reuse `is_open_call_from_pathlib` from ASYNC230: https://github.com/astral-sh/ruff/blob/1c7ea690a820deaa0e17ecf72593ebc4781f3752/crates/ruff_linter/src/rules/flake8_async/rules/blocking_open_call.rs#L73

amyreese:
Alternately PathlibPathChecker which can also match annotations: https://github.com/astral-sh/ruff/blob/1c7ea690a820deaa0e17ecf72593ebc4781f3752/crates/ruff_linter/src/rules/flake8_async/rules/blocking_path_methods.rs#L161

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__preview_FURB103_FURB103.py.snap
11happy: with the new name resolution utilizing the `is_open_call_from_pathlib` its not firing if pathlib.path has not been imported.

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: sure, I have used `is_open_call_from_pathlib` as suggested by @amyreese howevr had to make the function public.

crates/ruff_linter/src/rules/refurb/rules/write_whole_file.rs
11happy: I personally feel its a little bit more clear, however would be happy to change if you insist.
Thank  you : )

11happy:
the common helper has too many arguments causing the clippy error, should I break it down ?

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: yes have extracted the common logic from both functions. currently named it `resolve_file_open` would be happy to change if you have any suggestions.
Thank you : )

ntBre:
> the common helper has too many arguments causing the clippy error, should I break it down ?

I think it's okay just to add `#[expect(clippy::too_many_arguments)]` on the function.

crates/ruff_linter/src/rules/refurb/rules/write_whole_file.rs
ntBre: Sounds good, we can leave it then!

crates/ruff_linter/resources/test/fixtures/refurb/FURB103.py
```diff
@@ -1,3 +1,4 @@
+from pathlib import Path
```
ntBre: I think we might want a separate file for these new tests and this import. We're losing a lot of snapshot changes where the fix needed to add this import.

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -1,14 +1,14 @@
 use std::borrow::Cow;
 
+use crate::checkers::ast::Checker;
+use crate::rules::flake8_async::rules::blocking_open_call::is_open_call_from_pathlib;
+use crate::{Applicability, Edit, Fix};
```
ntBre: Can you move these back down to a separate `crate` import block like we had before? Stable rustfmt still doesn't handle this, unfortunately.

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -238,9 +207,87 @@ fn find_file_open<'a>(
...
+        func,
+        arguments: ast::Arguments { args, keywords, .. },
+        ..
+    } = item.context_expr.as_call_expr()?;
+    if !is_open_call_from_pathlib(func, semantic) {
+        return None;
+    }
+    let attr = func.as_attribute_expr()?;
+    let path_call = attr.value.as_call_expr()?;
+    let filename = path_call.arguments.args.first()?;
```
ntBre: Are we missing some of the argument validation code from `find_file_open`, especially these lines:

```rust
    if args.iter().any(Expr::is_starred_expr)
        || keywords.iter().any(|keyword| keyword.arg.is_none())
    {
        return None;
    }
```

I also think we're doing too much inspection on the `attr.value`. For example, I think `is_open_call_from_pathlib` should be able to identify a case like this:

```py
p = Path("file.txt")
with p.open("r") as f:
    f.write("test")
```

but the current implementation expects a literal `Path()` call, not a name that resolves to a `Path`.

We might need to make the `filename` field optional on the `FileOpen` struct.

crates/ruff_linter/resources/test/fixtures/refurb/FURB103.py
11happy: done I have added a seperate file suffixed _EXT

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: done

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: done , also the message is formatted accordingly but i feel its getting nested & dosent feel good , not sure although

11happy: also I have not added any comments yet , should i add similar comments like `find_file_open`

MichaReiser:
Would you mind rebasing your PR? Doing so will allow us to merge your PR without delayed once reviewed.

11happy:
done rebased it & resolved merge conflicts : )
Thank you

crates/ruff_linter/src/rules/refurb/helpers.rs
```diff
@@ -187,58 +246,62 @@ fn find_file_open<'a>(
...
-
-    // And even with all these restrictions, if the file handle gets used not exactly once,
-    // it doesn't fit the bill.
-    let [reference] = references.as_slice() else {
+    if !is_open_call_from_pathlib(func, semantic) {
         return None;
+    }
+    let attr = func.as_attribute_expr()?;
+    let filename = if let Expr::Call(path_call) = attr.value.as_ref() {
+        path_call.arguments.args.first()
```
ntBre: I don't think this is quite right. [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) takes `*args`, so it can be called like this:

```pycon
>>> from pathlib import Path
>>> Path("foo", "bar", "baz")
PosixPath('foo/bar/baz')
```

Taking only the first element would give the wrong result.

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB103_FURB103_1.py.snap
```diff
@@ -0,0 +1,131 @@
...
+
+FURB103 `Path.open()` followed by `write()` can be replaced by `p.write_text("test")`
+  --> FURB103_EXT.py:22:6
+   |
+21 | p = Path("file.txt")
+22 | with p.open("w") as f:
+   |      ^^^^^^^^^^^^^^^^
+23 |     f.write("test")
+   |
+help: Replace with `p.write_text("test")`
```
ntBre, line 131: Should this case have a fix?

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: actually this is handeled correctly already as we are we are using `path_obj` in fix generation
also have changed to require filename only in none case
```
let target = if let Some(path_obj) = open.path_obj {
        locator.slice(path_obj.range()).to_string()
    } else {
        let filename = open.filename?;
        let filename_code = locator.slice(filename.range());
        format!("{binding}({filename_code})")
    };
```

11happy: i think filename is redundant in path open cases as we are send `path_obj`

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB103_FURB103_1.py.snap
11happy: I believe for this case we can have a fix, but would like to know your opinion on this .
Thank you

11happy:
> Thanks for rebasing and for your work on this! I pushed a commit fixing a few nits but then noticed a few more minor issues. The implementation feels a bit complicated to me now, I want to keep thinking about it.

I agree with you implementation is complex ,felt the same, would appreciate any pointers to decomplicate/refactor it ?
Thank you : )

crates/ruff_linter/src/rules/refurb/helpers.rs
11happy: have added the case

crates/ruff_linter/src/rules/refurb/snapshots/ruff_linter__rules__refurb__tests__FURB103_FURB103_1.py.snap
ntBre: Sorry, yeah, I thought it should have a fix. This looks good now!

crates/ruff_linter/src/rules/refurb/helpers.rs
ntBre: Thanks, the new test looks good!

ntBre:
I just pushed a commit trying out an `OpenArgument` enum since the `path_obj` field and `filename` field were basically mutually exclusive. I think this makes things a little simpler, but I'm curious to get your thoughts on it too. I think the only remaining thing I don't love is the `is_path_open` check on the `Violation` struct. Maybe we could use the whole suggestion we generate for the fix in the error message?

I also fixed some new merge conflicts while I was here.

I also noticed that we haven't updated the code for FURB102, even though it's quite similar to FURB103. Do you want to do that in a follow-up PR?

11happy:
> I just pushed a commit trying out an `OpenArgument` enum since the `path_obj` field and `filename` field were basically mutually exclusive. I think this makes things a little simpler, but I'm curious to get your thoughts on it too. I think the only remaining thing I don't love is the `is_path_open` check on the `Violation` struct. Maybe we could use the whole suggestion we generate for the fix in the error message?
> 
> I also fixed some new merge conflicts while I was here.
> 
> I also noticed that we haven't updated the code for FURB102, even though it's quite similar to FURB103. Do you want to do that in a follow-up PR?

Thank you for these updates, will go through the commit in detail & will share my thoughts, I am also interested to work on the follow up PR for FURB102 , I feel a little occupied this week but will update this PR & my other open PRs soon.
Thanks again : )

11happy:
> I just pushed a commit trying out an `OpenArgument` enum since the `path_obj` field and `filename` field were basically mutually exclusive. I think this makes things a little simpler, but I'm curious to get your thoughts on it too. I think the only remaining thing I don't love is the `is_path_open` check on the `Violation` struct. Maybe we could use the whole suggestion we generate for the fix in the error message?

Yes, with the new OpenArgument enum code is a lot cleaner, what do you think of using `OpenArgument` directly instead of `is_path_open`  ?

ntBre (commented):
Thank you! The code for FURB103  looks good to me now, your last commit was a nice improvement :)

Based on the ecosystem report, this is already affecting FURB101 too (I had the wrong rule code when I mentioned FURB102 earlier). Would you mind going ahead and updating the FURB101 code and adding a test for it? I don't think it makes as much sense to save that for a follow-up if we already changed the rule's behavior here.

I think it's mostly just the `message`, `fix_title`, and `filename_code` that need to be updated, most of the logic is already in the shared helpers.

11happy:
> Thank you! The code for FURB103 looks good to me now, your last commit was a nice improvement :)
> 
> Based on the ecosystem report, this is already affecting FURB101 too (I had the wrong rule code when I mentioned FURB102 earlier). Would you mind going ahead and updating the FURB101 code and adding a test for it? I don't think it makes as much sense to save that for a follow-up if we already changed the rule's behavior here.
> 
> I think it's mostly just the `message`, `fix_title`, and `filename_code` that need to be updated, most of the logic is already in the shared helpers.

done, to keep it consistent with FURB103 separated the tests for FURB101, also should I add more tests ?
Thank  you : )

PR #21087, opened 2025-10-27
[`flynt`] Fix static-join-to-f-string when mixing raw and non-raw strings (`FLY002`)

### LINKED ISSUE #21082: Raw strings break the FLY002 fix
### Summary

When the argument to `str.join` contains only string literals and the first is raw, the fix for [`static-join-to-f-string` (FLY002)](https://docs.astral.sh/ruff/rules/static-join-to-f-string/) can introduce a syntax error or change the program’s behavior.

If a subsequent string isn’t raw, its contents will be inserted into the fix’s string, regardless of syntax errors. [Example](https://play.ruff.rs/372f5506-c7bf-41ba-8ef7-fc3eafe725a8):
```console
$ cat >fly002_1.py <<'# EOF'
"".join((r"", '"'))
# EOF

$ ruff check --isolated fly002_1.py --select FLY002 --unsafe-fixes --diff 2>&1 | grep error:
```

[Another example](https://play.ruff.rs/503176c6-9a2e-428c-befc-1130af96d1e0):
```console
$ cat >fly002_2.py <<'# EOF'
"".join((r"", "\\"))
# EOF

$ ruff check --isolated fly002_2.py --select FLY002 --unsafe-fixes --diff 2>&1 | grep error:
```

[Another example](https://play.ruff.rs/83a55ba4-cf6f-4bd1-85ed-6824692f6dbd):
```console
$ cat >fly002_3.py <<'# EOF'
"".join((r"", "\0"))
# EOF

$ ruff check --isolated fly002_3.py --select FLY002 --unsafe-fixes --fix

$ xxd fly002_3.py
00000000: 7222 0022 0a                             r".".

$ python fly002_3.py 2>&1 | tail -n 1
SyntaxError: source code cannot contain null bytes
```

It can also change the program’s behavior. [Example](https://play.ruff.rs/35984440-7add-4737-a205-77af0417c5d2):
```console
$ cat >fly002_4.py <<'# EOF'
print("".join((r"", r'".__class__.__name__, 0xA, sep="~')))
# EOF

$ python fly002_4.py
".__class__.__name__, 0xA, sep="~

$ ruff check --isolated fly002_4.py --select FLY002 --unsafe-fixes --fix

$ cat fly002_4.py
print(r"".__class__.__name__, 0xA, sep="~")

$ python fly002_4.py
str~10
```

If any string contains a carriage return via an escape sequence, the fix emits a multiline string containing a literal carriage return. Literal carriage returns are interpreted as line feeds, so this changes the program’s behavior. [Example](https://play.ruff.rs/a0fee72c-bb57-4a46-8410-672fd9039c5d):
```console
$ cat >fly002_5.py <<'# EOF'
print(repr("".join((r"", "\r"))))
# EOF

$ python fly002_5.py
'\r'

$ ruff check --isolated fly002_5.py --select FLY002 --unsafe-fixes --fix

$ xxd fly002_5.py
00000000: 7072 696e 7428 7265 7072 2872 2222 220d  print(repr(r""".
00000010: 2222 2229 290a                           """)).

$ python fly002_5.py
'\n'
```

### Version

danparizher:
## Summary

Refine `FLY002` to correctly handle mixed raw and non-raw strings in `str.join` calls. Instead of bailing on these transformations, the rule now preserves the raw flag when safe or automatically switches to a non-raw representation when raw string syntax would be invalid (e.g., trailing odd backslashes or results containing the quote character).

Closes #21082

## Problem Analysis

When joining strings where at least one is a raw string (`r""`):
1. Applying the raw flag to the joined result can lead to **syntax errors** if the result ends in an odd number of backslashes (e.g., `"".join((r"", "\\"))` → `r"\"` is invalid).
2. It can lead to **behavior changes** or **null byte issues** if non-raw strings containing escape sequences are joined into a raw result where those sequences are no longer processed.
3. Raw strings cannot safely contain the quote delimiter used or end with an un-escaped (odd) number of backslashes before the closing quote.

## Approach

Updated `build_fstring` to detect if any input string is raw and then validate if the joined content can be safely represented as such:

1.  **Detection:** Merged the `any_raw` check into the existing string literal iteration to identify if the output should prefer a raw prefix.
2.  **Safety Validation:** If the result contains `\r`, `\0`, the quote delimiter, or ends in an **odd number of backslashes**, it is deemed unsafe for a raw representation.
3.  **Fallback:** If unsafe for raw, the f-string/literal prefix is set to `Empty` (non-raw). This allows the fix to proceed safely by using standard escape sequences in the result.
4.  **Triple Quoting:** Preserved logic to use triple quotes for raw strings containing newlines, as long as it's safe.

## Testing

Updated and expanded tests in `FLY002.py`:
- `nok14`–`nok17`: Mixed strings that trigger non-raw fallback due to quotes, null bytes, or carriage returns.
- `nok18`: Verified that literal `\r` (backslash + `r`) in mixed joins is handled correctly.
- `nok19`–`nok20`: Added edge cases for trailing backslashes. An even number of trailing backslashes remains raw (`r"\\"`), while an odd number switches to non-raw (`"\\\\"`) to avoid syntax errors.
- `ok7`–`ok10`: Confirmed that homogeneous joins (all raw or all non-raw) continue to work correctly.


dscorbett:
Joining raw and non-raw strings is not the problem. The fix for FLY002 works fine for some combinations:
```console
$ cat >fly002.py <<'# EOF'
print("".join(('"', r"@")))
print("".join((r"@", "!")))
# EOF

$ python fly002.py
"@
@!

$ ruff check --isolated fly002.py --select FLY002 --unsafe-fixes --fix

$ cat fly002.py
print('"')
print(r"!")

$ python fly002.py
"@
@!
```
Removing support for such combinations would need to be done in preview, but it would be more helpful to instead keep the support and fix the bugs. In particular, that means that if the fix’s output string cannot be represented as a raw string, it should be represented as a non-raw string, even if the first input string is raw.

crates/ruff_linter/resources/test/fixtures/flynt/FLY002.py
```diff
@@ -29,3 +29,19 @@ def create_file_public_url(url, filename):
...
 nok12 = ''.join([foo, "'", '"'])
 nok13 = "".join([foo, "'", '"'])
+
+# Regression test for: https://github.com/astral-sh/ruff/issues/21082
+# Mixing raw and non-raw strings can cause syntax errors or behavior changes
+nok14 = "".join((r"", '"'))  # First is raw, second is not - would break syntax
+nok15 = "".join((r"", "\\"))  # First is raw, second has backslash - would break syntax
+nok16 = "".join((r"", "\0"))  # First is raw, second has null byte - would introduce null bytes
+nok17 = "".join((r"", "\r"))  # First is raw, second has carriage return - would change behavior
+nok18 = "".join((r"", "\\r"))  # First is raw, second has carriage return escape - would change behavior
```
dscorbett: `"\\r"` does not have a carriage return escape: it has a backslash escape followed by a literal ⟨r⟩. FLY002 doesn’t change its behavior.

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -91,6 +105,21 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
 
         let mut flags = flags?;
 
+        // If any input was raw but the content cannot be safely represented as a raw string,
+        // use non-raw representation. This handles cases where raw strings would create invalid
+        // syntax or behavior changes.
+        if any_raw && !content.is_empty() {
+            let needs_non_raw = content.contains(['\n', '\r', '\0']) || {
```
dscorbett: A line feed (`'\n'`) _can_ safely be represented in a raw string. The section after “If the result is a raw string and contains a newline, use triple quotes” handles that case.

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -91,6 +105,21 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
...
+                // Check if content contains characters that would break raw string syntax
+                let quote_char = flags.quote_str();
+                content.contains(quote_char) || content.contains('\\')
```
dscorbett: A backslash can safely appear in a raw string as long as it is not the last character. It’s probably common for raw strings to contain backslashes (because their main point is that they handle backslashes differently) so it might be worth trying to support that.

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -74,14 +76,26 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
     // If all elements are string constants, join them into a single string.
     if joinees.iter().all(Expr::is_string_literal_expr) {
         let mut flags: Option<ast::StringLiteralFlags> = None;
+        let mut any_raw = false;
```
ntBre, line 79: Can we just add the `any_raw` check into the existing `filter_map` call?

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -91,6 +105,25 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
...
+        // use non-raw representation. This handles cases where raw strings would create invalid
+        // syntax or behavior changes.
+        if any_raw && !content.is_empty() {
+            let needs_non_raw = content.contains(['\r', '\0']) || {
+                // Check if content contains characters that would break raw string syntax
+                let quote_char = flags.quote_str();
+                // A raw string cannot end with a single backslash if it's immediately
+                // followed by the quote delimiter, as that would be invalid syntax.
+                let ends_with_backslash = content.ends_with('\\')
+                    && (content.len() == 1 || content.chars().nth_back(1) != Some('\\'));
```
ntBre: This is a bit of an edge case, but I think this check is incorrect for odd numbers of backslashes. It checks for `\` and `\\`, but `\\\` would be allowed.

crates/ruff_linter/src/rules/flynt/snapshots/ruff_linter__rules__flynt__tests__FLY002_FLY002.py.snap
```diff
@@ -205,4 +226,183 @@ help: Replace with `f"{foo}'"`
...
+   |         ^^^^^^^^^^^^^^^^^^^
+36 | nok15 = "".join((r"", "\\"))  # First is raw, second has backslash - would break syntax
+37 | nok16 = "".join((r"", "\0"))  # First is raw, second has null byte - would introduce null bytes
+   |
+help: Replace with `'"'`
+32 | 
+33 | # Regression test for: https://github.com/astral-sh/ruff/issues/21082
+34 | # Mixing raw and non-raw strings can cause syntax errors or behavior changes
+   - nok14 = "".join((r"", '"'))  # First is raw, second is not - would break syntax
+35 + nok14 = '"'  # First is raw, second is not - would break syntax
```
ntBre: It looks like this comment is outdated?

ntBre (commented):
Sorry for the delay in reviewing this, and thanks as always for your work.

I wasn't quite sure if this was ready for another review, but I took a look and added a couple of comments. I'm always wary of these touchy string manipulation changes.

Would you mind updating the PR summary and also the comments in the new test cases? I think they've drifted a bit from the current version of the PR.

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -78,24 +78,17 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
...
             .iter()
             .filter_map(|expr| {
                 if let Expr::StringLiteral(ast::ExprStringLiteral { value, .. }) = expr {
+                    let curr_flags = value.first_literal_flags();
+                    if flags.is_none() {
+                        flags = Some(curr_flags);
+                        any_raw = curr_flags.prefix().is_raw();
+                    } else if curr_flags.prefix().is_raw() {
+                        any_raw = true;
+                    }
```
ntBre: I think we could simplify this slightly to:

```rust
if flags.is_none() {
    flags = Some(curr_flags);
}
any_raw |= curr_flags.prefix().is_raw();
```

If we feel really clever, we could even do something similar for the `Option`:

```rust
flags.get_or_insert(curr_flags);
```

crates/ruff_linter/src/rules/flynt/rules/static_join_to_fstring.rs
```diff
@@ -91,6 +98,32 @@ fn build_fstring(joiner: &str, joinees: &[Expr], flags: FStringFlags) -> Option<
 
         let mut flags = flags?;
 
+        // If any input was raw but the content cannot be safely represented as a raw string,
+        // use non-raw representation. This handles cases where raw strings would create invalid
+        // syntax or behavior changes.
+        if any_raw && !content.is_empty() {
+            let needs_non_raw = content.contains(['\r', '\0']) || {
+                // Check if content contains characters that would break raw string syntax
+                let quote_char = flags.quote_str();
```
ntBre: This is a pretty small nit, but it's a bit misleading to call this a `quote_char` when it contains multiple characters for triple-quoted strings. `quote_str` would be a better name, like the method.

Beyond that, should we move this `contains` check outside of this block? That would allow us to bypass the `content.chars().rev()` loop if `content` contains the quote string. `quote_char` doesn't appear to be used anywhere else until line 119.

ntBre (changes requested):
Thanks and sorry again for the delay. I left a couple of small simplification suggestions, but I think I also found another case with a raw, triple-quoted string that introduces a syntax error:

```py
nok21 = "".join((r"""prefix""", '"'))
```

I went ahead and merged the main branch while I was playing with the code locally. There weren't any git conflicts, but the ~~number of context lines~~ trailing whitespace handling included in our snapshots had changed slightly.

ntBre (changes requested):
Thanks for applying those fixes. However, it looks like the `nok21` case from my previous review is still unresolved. I went ahead and pushed it as a new test case so that it will show up in CI.

PR #21092, opened 2025-10-27
[`ruff-ecosystem`] Fix CLI crash on Python 3.14

### LINKED ISSUE #20965: `ruff-ecosystem` CLI crashes on Python 3.14
This line 

https://github.com/astral-sh/ruff/blob/3c229ae58abbe87523c9a2d3223ebcc67962137b/python/ruff-ecosystem/ruff_ecosystem/cli.py#L90

causes a crash on Python 3.14 when there is not yet an event loop (see https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop).

danparizher:
## Summary

Fixed the Python 3.14 crash in `ruff-ecosystem` by updating event loop handling.

Fixes #20965

### Changes Made

**1. `python/ruff-ecosystem/ruff_ecosystem/cli.py` (line 90-91)**
- Replaced `asyncio.get_event_loop()` with `asyncio.new_event_loop()` and `asyncio.set_event_loop(loop)`

**2. `scripts/check_ecosystem.py` (line 531-532)**
- Applied the same update

### Why This Works

In Python 3.14, `asyncio.get_event_loop()` raises when no loop exists. Creating the loop explicitly ensures:
- Python 3.10+ support
- Signal handler behavior preserved
- Loop is closed in the finally block

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -531,8 +531,7 @@ jobs:
           persist-credentials: false
       - uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0
         with:
-          # TODO: figure out why `ruff-ecosystem` crashes on Python 3.14
-          python-version: "3.13"
+          python-version: ${{ env.PYTHON_VERSION }}
           activate-environment: true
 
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Thank you.

PR #21093, opened 2025-10-27
[`airflow`] warning `airflow....DAG.create_dagrun` has been removed (`AIR301`)

Lee-W:
## Summary

* Warn that `airflow....DAG.create_dagrun` has been removed

## Test Plan

update accordingly and reorganize test cases

### DIFF
--- a/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
+++ b/crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
@@ -492,10 +492,12 @@ fn check_method(checker: &Checker, call_expr: &ExprCall) {
             "collected_datasets" => Replacement::AttrName("collected_assets"),
             _ => return,
         },
-        ["airflow", .., "DAG"] => match attr.as_str() {
-            "create_dagrun" => Replacement::None,
-            _ => return,
-        },
+        ["airflow", "models", "dag", "DAG"] | ["airflow", "models", "DAG"] | ["airflow", "DAG"] => {
+            match attr.as_str() {
+                "create_dagrun" => Replacement::None,
+                _ => return,
+            }
+        }
         ["airflow", "providers_manager", "ProvidersManager"] => match attr.as_str() {
             "initialize_providers_dataset_uri_resources" => {
                 Replacement::AttrName("initialize_providers_asset_uri_resources")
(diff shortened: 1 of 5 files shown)


MichaReiser:
Now that the rule is stable, changes like this may need to be gated behind preview and go through our normal stabilization period. 

I think the change here is fine as it isn't a significant expansion of the rule's scope:

crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
```diff
@@ -492,6 +492,10 @@ fn check_method(checker: &Checker, call_expr: &ExprCall) {
             "collected_datasets" => Replacement::AttrName("collected_assets"),
             _ => return,
         },
+        ["airflow", .., "DAG"] => match attr.as_str() {
```
MichaReiser: Would it be better to be explicit from which module `DAG` was removed to avoid false positives instead of allowing any `DAG` symbol imported from anywhere within airflow (e.g. this also flags `airflow.providers.my_provider.DAG`)

Lee-W:
> Now that the rule is stable, changes like this may need to be gated behind preview and go through our normal stabilization period.
> 
> I think the change here is fine as it isn't a significant expansion of the rule's scope:

We had a discussion with @ntBre previously. Our conclusion back then was this kind of small ones are fine.

But if we want to make it gated behind preview, what would be the process? Do we need to extract these logic to somewhere so that they can be activated through preview flag? Thanks!

ntBre:
> But if we want to make it gated behind preview, what would be the process? Do we need to extract these logic to somewhere so that they can be activated through preview flag? Thanks!

It should be pretty easy to add guards to the `match` arms and a new preview function like in [`preview.rs`](https://github.com/astral-sh/ruff/blob/c0fb235a703684bca1cc1767015bb7e19ab1c22f/crates/ruff_linter/src/preview.rs#L16-L19).

But I agree that this change seems fine without preview.

Lee-W:
Cool! This is super helpful! We might have some minor ones like this. Please let me know if any of them need to be in preview mode. Thank you!

crates/ruff_linter/src/rules/airflow/rules/removal_in_3.rs
Lee-W: it's unlikely, but definiely better. Let me make the change.

Lee-W: just updated. Thanks!

MichaReiser: Can you double-check if you pushed your most recent changes?

Lee-W: Hi, sorry for the mess, it might be missed due to wrong rebasing. Just updated it and checked again. Thanks for reminder!

MichaReiser (approved):

Issue #19552, opened 2025-07-25
Check Watch does not show output format error message from clearing too late


MeGaGiGaGon:
### Summary

I found this while helping on #19550. When running `check` in `watch` mode with a non-full `output-format`, this error:
```
warning: `--output-format full` is always used in watch mode.
``` 
Does not get displayed due to the terminal being cleared directly after. Here's the contents of `out.txt` I get when running `uvx ruff check --watch . --output-format json 2>&1 > out.txt`:
```
warning: `--output-format full` is always used in watch mode.
[H[2J[3J[23:04:31 PM] Starting linter in watch mode...

[23:04:31 PM] Found 7 errors. Watching for file changes.

... normal error output here
```

I assume `�[H�[2J�[3J` is the magic characters that cause a terminal clear in PowerShell, which shows that it's erasing the warning. This should be fixable by clearing the terminal prior to warnings being output, though I haven't looked at the code to see how easy it would be.

### Version

ruff 0.12.5 (d13228ab8 2025-07-24)

ntBre:
Possibly even more surprising is the fact that `--preview` enables the `full` format:

https://github.com/astral-sh/ruff/blob/6663571c9000a081f42ccb9c9f20a6f6905d9942/crates/ruff/src/printer.rs#L460-L462

Compare the `with_show_source` call to the `write_once` implementation:

https://github.com/astral-sh/ruff/blob/6663571c9000a081f42ccb9c9f20a6f6905d9942/crates/ruff/src/printer.rs#L262-L265

But yeah, without `--preview`, the default is definitely `concise`.

mikeleppane:
Hi! I've addressed the main issue, but I have a few comments regarding the output formatting:

- What is the reason for using only "one" formatting output logic in `--watch` mode compared to non-watch mode?
- As mentioned earlier, is it an oversight that the `--preview` option enables full mode formatting?
- Is the warning: `warning: `--output-format full` is always used in watch mode.` needed in the first place? Or is it trying to say that in `watch` mode, we only apply `full` mode?

Note: I looked at the [#19550](https://github.com/astral-sh/ruff/issues/19550) for more details. I think these two can be combined and tackled together.

ntBre:
Those are good questions. I'm not sure I know the exact answers, but I can at least weigh in :)

- I'm sure there was a reason at some point only to use one output format with `--watch`, but I think it makes sense to support `--output-format` now (possibly `--watch` was added well before we had multiple output formats)
[...]

PR #21097, opened 2025-10-27
Respect `--output-format` with `--watch`

### LINKED ISSUE #19550: How do i make ruff show more details in watch mode?
### Question

The ruff linter for python in watch mode does not show code details and position of error.

If I run "ruff check .", if there is a code error, it shows me the code where the issue is.

If I run "ruff check --watch .", it does not show me code details.

### Version

_No response_

ntBre:
Summary
--

Fixes #19550

This PR copies our non-watch diagnostic rendering code into `Printer::write_continuously` in preview mode, allowing it to use whatever output format is passed in.

I initially marked this as also fixing #19552, but I guess that's not true currently but will be true once this is stabilized and we can remove the warning.

Test Plan
--

Existing tests, but I don't think we have any `watch` tests, so some manual testing as well. The default with just `ruff check --watch` is still `concise`, adding just `--preview` still gives the `full` output, and then specifying any other output format works, with JSON as one example:

Screenshot 2025-10-27 at 9 21 41 AM

### DIFF
--- a/crates/ruff/src/printer.rs
+++ b/crates/ruff/src/printer.rs
@@ -9,9 +9,7 @@ use itertools::{Itertools, iterate};
 use ruff_linter::linter::FixTable;
 use serde::Serialize;
 
-use ruff_db::diagnostic::{
-    Diagnostic, DiagnosticFormat, DisplayDiagnosticConfig, DisplayDiagnostics, SecondaryCode,
-};
+use ruff_db::diagnostic::{Diagnostic, DisplayDiagnosticConfig, SecondaryCode};
 use ruff_linter::fs::relativize_path;
 use ruff_linter::logging::LogLevel;
 use ruff_linter::message::{EmitterContext, render_diagnostics};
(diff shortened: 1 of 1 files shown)


crates/ruff/src/printer.rs
```diff
@@ -389,22 +389,30 @@ impl Printer {
...
+                    .color(!cfg!(test) && colored::control::SHOULD_COLORIZE.should_colorize())
+                    .with_show_fix_status(show_fix_status(self.fix_mode, fixables.as_ref()))
+                    .format(DiagnosticFormat::Concise)
+                    .with_fix_applicability(self.unsafe_fixes.required_applicability());
+
+                write!(
+                    writer,
+                    "{}",
+                    DisplayDiagnostics::new(&context, &config, &diagnostics.inner)
+                )?;
```
MichaReiser: How's `DisplayDiagnostics` different from `render_diagnostics`? Or why can't we do

```
let format = if preview {
	self.format
} else {
	DiagnosticFormat::Concise
};

let config = DisplayDiagnosticConfig::default()
    .preview(preview)
    .hide_severity(true)
    .color(!cfg!(test) && colored::control::SHOULD_COLORIZE.should_colorize())
    .with_show_fix_status(show_fix_status(self.fix_mode, fixables.as_ref()))
    .with_fix_applicability(self.unsafe_fixes.required_applicability())
    .show_fix_diff(preview);

render_diagnostics(writer, self.format, config, &context, &diagnostics.inner)?;
```

ntBre: Yeah, I think that should be equivalent. I was trying not to change too much since we don't have tests for `--watch`, but this seems pretty easy to verify in hindsight 🤦

dylwil3:
> I initially marked this as also fixing https://github.com/astral-sh/ruff/issues/19552, but I guess that's not true currently but will be true once this is stabilized and we can remove the warning.

Don't we generally mark the issue as closed even if the solution is in `preview`?

ntBre:
> > I initially marked this as also fixing #19552, but I guess that's not true currently but will be true once this is stabilized and we can remove the warning.
> 
> Don't we generally mark the issue as closed even if the solution is in `preview`?

In general yes, but this felt a  little different  since it's basically a preview warning.

ntBre:
Actually the warning itself probably should be updated to mention preview now, but the warning isn't being shown anyway because of #19552 :upside_down_face:

PR #21100, opened 2025-10-27
[ty] Make auto-import skip symbols in current module

BurntSushi:
This adds a small check to ensure symbols from auto-import that are
defined in the current module aren't returned. In particular, these
symbols wind up being redundant and are just overall quite confusing.

It's best to review this commit-by-commit. In particular, the bulk of
this PR is a refactoring of the completion tests to make them a little
more declarative in more cases.

Fixes [#1445](https://github.com/astral-sh/ty/issues/1445)


crates/ty_ide/src/completion.rs
```diff
@@ -4034,66 +4017,144 @@ def f(x: T):
...
+    impl CompletionTestBuilder {
+        /// Returns completions based on this configuration.
+        fn build(&self) -> CompletionTest<'_> {
+            let original = completion(
+                &self.cursor_test.db,
+                &self.settings,
+                self.cursor_test.cursor.file,
+                self.cursor_test.cursor.offset,
+            );
+            let filtered = original
```
MichaReiser, line 4099: Nit: You'll probably role your eyes given that this is a test. Please feel free to disregard it ;)

Should we use `original.drain` when building `filtered` and a) only store the filtered out suggestions in `CompletionTest` (instead of `original`) or b) only store a bool saying if any suggestions were discarded.

crates/ty_completion_eval/truth/auto-import-skips-current-module/main.py
```diff
@@ -0,0 +1,3 @@
+Kadabra = 1
+
+Kada<CURSOR: Kadabra>
```
MichaReiser, line 19: To make sure I understand the test case. This used to suggest `Kadabra` twice, once as local symbol and once as import?

crates/ty_ide/src/completion.rs
```diff
@@ -330,6 +330,10 @@ fn add_unimported_completions<'db>(
     let members = importer.members_in_scope_at(scoped.node, scoped.node.start());
 
     for symbol in all_symbols(db, typed) {
+        if Some(file) == symbol.module.file(db) {
+            continue;
+        }
```
MichaReiser: I misread this as a `if let Some()` and was very confused why skipping files without a module would fix this bug. Maybe rewrite to `if symbol.module.file(db) == Some(file)` to disambiguate?

crates/ty_ide/src/completion.rs
BurntSushi: Yeah I think since this is a test, I don't mind it. I was mostly just trying to write the simplest thing.

crates/ty_completion_eval/truth/auto-import-skips-current-module/main.py
BurntSushi: That's right. I added this comment:

```
Kadabra = 1

# This is meant to reflect that auto-import
# does *not* include completions for `Kadabra`.
# That is, before a bug was fixed, completions
# would offer two variants for `Kadabra`: one
# for the current module (correct) and another
# from auto-import that would insert
# `from main import Kadabra` into this module
# (incorrect).
#
# Since the incorrect one wasn't ranked above
# the correct one, this task unfortunately
# doesn't change the evaluation results. But
# I've added it anyway in case it does in the
# future (or if we change our evaluation metric
# to something that incorporates suggestions
# after the correct one).
Kada<CURSOR: Kadabra>
```

crates/ty_ide/src/completion.rs
BurntSushi: Hah yeah, that's fair.

PR #21102, opened 2025-10-27
Fix some false positive for `is_non_empty_f_string`

### LINKED ISSUE #21048: `is_non_empty_f_string` has false positives
### Summary

[`is_non_empty_f_string`](https://github.com/astral-sh/ruff/blob/0.14.1/crates/ruff_python_ast/src/helpers.rs#L1350) has false positives for f-strings with certain interpolations. The interpolation `{f""!s}` is empty. Interpolations with format specs or containing comparison expressions have unknown emptiness. [Here is an example](https://play.ruff.rs/5612a49d-2c07-4852-9725-340777c1982a) with [`call-with-shell-equals-true` (S604)](https://docs.astral.sh/ruff/rules/call-with-shell-equals-true/):
```console
$ cat >is_non_empty_f_string_fp.py <<'# EOF'
print(bool(dict(shell=f"{f""!s}")["shell"]))

print(bool(dict(shell=f"{"x":.0}")["shell"]))

class C:
    def __init__(self, chars):
        self.chars = chars
    def __gt__(self, other):
        return "".join(sorted(c for c in self.chars if all(c > o for o in other.chars)))
print(bool(dict(shell=f"{C("abc") > C("xyz")}")["shell"]))
# EOF

$ python is_non_empty_f_string_fp.py
False
False
False

$ ruff --isolated check is_non_empty_f_string_fp.py --select S604 --output-format concise -q
is_non_empty_f_string_fp.py:1:12: S604 Function call with truthy `shell` parameter identified, security issue
is_non_empty_f_string_fp.py:3:12: S604 Function call with truthy `shell` parameter identified, security issue
is_non_empty_f_string_fp.py:10:12: S604 Function call with truthy `shell` parameter identified, security issue
```

### Version

IDrokin117:
## Summary
Closes #21048

Fix some false positive edge cases when determining string truthiness. The following cases are now treated as unknown truthiness:

1. Comparison expressions can return any value, even empty strings.
2. F-strings with format specs can produce either empty or non-empty strings:
```python
bool(f"{'mystring':.0}") # Returns False
``` 

Additionally, fixed a bug that caused `f"{f""!s}"` to be treated as non-empty. Since `.iter().all()` returns true if the iterator is empty, f-strings with no elements (like nested `f"{f""!s}"`) were incorrectly treated as non-empty.

## Test Plan
- Added test cases to S604 since `Truthiness` has no tests of its own

### DIFF
--- a/crates/ruff_python_ast/src/helpers.rs
+++ b/crates/ruff_python_ast/src/helpers.rs
@@ -1357,7 +1357,9 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
             Expr::ListComp(_) => true,
             Expr::SetComp(_) => true,
             Expr::DictComp(_) => true,
-            Expr::Compare(_) => true,
+            // Compare can return any value, even empty string
+            // https://docs.python.org/3/reference/datamodel.html#object.__ge__
+            Expr::Compare(_) => false,
             Expr::NumberLiteral(_) => true,
             Expr::BooleanLiteral(_) => true,
             Expr::NoneLiteral(_) => true,
@@ -1399,12 +1401,16 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
     expr.value.iter().any(|part| match part {
         ast::FStringPart::Literal(string_literal) => !string_literal.is_empty(),
         ast::FStringPart::FString(f_string) => {
-            f_string.elements.iter().all(|element| match element {
-                InterpolatedStringElement::Literal(string_literal) => !string_literal.is_empty(),
-                InterpolatedStringElement::Interpolation(f_string) => {
-                    f_string.debug_text.is_some() || inner(&f_string.expression)
-                }
-            })
+            !f_string.elements.is_empty()
+                && f_string.elements.iter().all(|element| match element {
+                    InterpolatedStringElement::Literal(string_literal) => {
+                        !string_literal.is_empty()
+                    }
+                    InterpolatedStringElement::Interpolation(f_string) => {
+                        f_string.format_spec.is_none()
+                            && (f_string.debug_text.is_some() || inner(&f_string.expression))
+                    }
+                })
         }
     })
 }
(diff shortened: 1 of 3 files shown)


crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1399,12 +1401,16 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
...
         ast::FStringPart::Literal(string_literal) => !string_literal.is_empty(),
         ast::FStringPart::FString(f_string) => {
-            f_string.elements.iter().all(|element| match element {
-                InterpolatedStringElement::Literal(string_literal) => !string_literal.is_empty(),
-                InterpolatedStringElement::Interpolation(f_string) => {
-                    f_string.debug_text.is_some() || inner(&f_string.expression)
-                }
-            })
+            !f_string.elements.is_empty()
+                && f_string.elements.iter().all(|element| match element {
```
ntBre, line 1405: Should this just be `any` instead? I could certainly be wrong, but it seems like we should check if any part is non-empty rather than `all` parts are non-empty. That would also give the correct behavior for the empty iterator and match the outer `any` on line 1401.

I tried this locally, and I think it fixed an existing bug in a SIM222 test too.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1357,7 +1357,9 @@ fn is_non_empty_f_string(expr: &ast::ExprFString) -> bool {
             Expr::ListComp(_) => true,
             Expr::SetComp(_) => true,
             Expr::DictComp(_) => true,
-            Expr::Compare(_) => true,
+            // Compare can return any value, even empty string
+            // https://docs.python.org/3/reference/datamodel.html#object.__ge__
+            Expr::Compare(_) => false,
```
ntBre, line 1362: Should we move this down to the other complex expressions below? It seems slightly out of place in this block now that the other variants are all `true`.

dscorbett: To be more precise, this could still be `true` if all the operators in the `Compare` are the ones guaranteed to return `bool`s (`in`, `is`, `is not`, and `not in`).

PR #21104, opened 2025-10-27
Fix syntax error false positive on nested alternative patterns

### LINKED ISSUE #21101: Syntax error false positive on nested alternative patterns
This change seems to now cause a SyntaxError on the following valid code that doesn't use different names:

```
invalid-syntax: alternative patterns bind different names
   --> src/sp_repo_review/checks/ruff.py:127:17
    |
125 |           match ruff:
126 |               case (
127 | /                 {"lint": {"select": x} | {"extend-select": x}}
128 | |                 | {"select": x}
129 | |                 | {"extend-select": x}
    | |______________________________________^
130 |               ):
131 |                   return cls.code in x or "ALL" in x
    |
```

See https://github.com/scientific-python/cookie/pull/663

_Originally posted by @henryiii in https://github.com/astral-sh/ruff/issues/20682#issuecomment-3452968119_

and here's a minimized [example](https://play.ruff.rs/f0eab842-e43c-4142-a764-df78b8c078ed):

```py
match ruff:
    case {"lint": {"select": x} | {"extend-select": x}} | {"select": x}:
        return cls.code in x or "ALL" in x
```

ntBre:
## Summary

Fixes #21101 by storing the child visitor's names in the parent visitor. This makes sure that `visitor.names` on line 1818 isn't empty after we visit a nested OR pattern.

## Test Plan

New inline test cases derived from the issue, [playground](https://play.ruff.rs/7b6439ac-ee8f-4593-9a3e-c2aa34a595d0)

### DIFF
--- a/crates/ruff_python_parser/src/semantic_errors.rs
+++ b/crates/ruff_python_parser/src/semantic_errors.rs
@@ -1857,6 +1857,7 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
                         );
                         break;
                     }
+                    self.names = visitor.names;
                 }
             }
         }
--- a/crates/ruff_python_parser/tests/snapshots/valid_syntax@nested_alternative_patterns.py.snap
+++ b/crates/ruff_python_parser/tests/snapshots/valid_syntax@nested_alternative_patterns.py.snap
@@ -1,0 +1,495 @@
+---
+source: crates/ruff_python_parser/tests/fixtures.rs
+input_file: crates/ruff_python_parser/resources/inline/ok/nested_alternative_patterns.py
+---
+## AST
+
+```
+Module(
+    ModModule {
+        node_index: NodeIndex(None),
+        range: 0..181,
+        body: [
+            Match(
+                StmtMatch {
+                    node_index: NodeIndex(None),
+                    range: 0..96,
+                    subject: Name(
+                        ExprName {
+                            node_index: NodeIndex(None),
+                            range: 6..10,
+                            id: Name("ruff"),
+                            ctx: Load,
+                        },
+                    ),
+                    cases: [
+                        MatchCase {
+                            range: 16..96,
+                            node_index: NodeIndex(None),
+                            pattern: MatchOr(
+                                PatternMatchOr {
+                                    node_index: NodeIndex(None),
+                                    range: 21..83,
+                                    patterns: [
+                                        MatchMapping(
+                                            PatternMatchMapping {
+                                                node_index: NodeIndex(None),
+                                                range: 21..67,
+                                                keys: [
+                                                    StringLiteral(
+                                                        ExprStringLiteral {
+                                                            node_index: NodeIndex(None),
+                                                            range: 22..28,
+                                                            value: StringLiteralValue {
+                                                                inner: Single(
+                                                                    StringLiteral {
+                                                                        range: 22..28,
+                                                                        node_index: NodeIndex(None),
+                                                                        value: "lint",
+                                                                        flags: StringLiteralFlags {
+                                                                            quote_style: Double,
+                                                                            prefix: Empty,
+                                                                            triple_quoted: false,
+                                                                            unclosed: false,
+                                                                        },
+                                                                    },
+                                                                ),
+                                                            },
+                                                        },
+                                                    ),
+                                                ],
+                                                patterns: [
+                                                    MatchOr(
+                                                        PatternMatchOr {
+                                                            node_index: NodeIndex(None),
+                                                            range: 30..66,
+                                                            patterns: [
+                                                                MatchMapping(
+                                                                    PatternMatchMapping {
+                                                                        node_index: NodeIndex(None),
+                                                                        range: 30..43,
+                                                                        keys: [
+                                                                            StringLiteral(
+                                                                                ExprStringLiteral {
+                                                                                    node_index: NodeIndex(None),
+                                                                                    range: 31..39,
+                                                                                    value: StringLiteralValue {
+                                                                                        inner: Single(
+                                                                                            StringLiteral {
+                                                                                                range: 31..39,
+                                                                                                node_index: NodeIndex(None),
+                                                                                                value: "select",
+                                                                                                flags: StringLiteralFlags {
+                                                                                                    quote_style: Double,
+                                                                                                    prefix: Empty,
+                                                                                                    triple_quoted: false,
+                                                                                                    unclosed: false,
+                                                                                                },
+                                                                                            },
+                                                                                        ),
+                                                                                    },
+                                                                                },
+                                                                            ),
+                                                                        ],
+                                                                        patterns: [
+                                                                            MatchAs(
+                                                                                PatternMatchAs {
+                                                                                    node_index: NodeIndex(None),
+                                                                                    range: 41..42,
+                                                                                    pattern: None,
+                                                                                    name: Some(
+                                                                                        Identifier {
+                                                                                            id: Name("x"),
+                                                                                            range: 41..42,
+                                                                                            node_index: NodeIndex(None),
+                                                                                        },
+                                                                                    ),
+                                                                                },
+                                                                            ),
+                                                                        ],
+                                                                        rest: None,
+                                                                    },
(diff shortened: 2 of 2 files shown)


crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1841,13 +1841,23 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
...
+                        // match 42:
+                        //     case [[x | x] | [x]] | x: ...
                         SemanticSyntaxChecker::add_error(
                             self.ctx,
                             SemanticSyntaxErrorKind::DifferentMatchPatternBindings,
                             *range,
                         );
                         break;
                     }
+                    self.names = visitor.names;
```
MichaReiser, line 1860: I don't have a good understanding of this part. So please disregard if not relevant or I'm just wrong ;)

Would it be better to return the union of all `visitor.names` here instead of just the last one in case we encountered a  `DifferentMatchPatternBindings` error? Or is there a risk that this introduces other false positives? If so, should we return the intersection of all names instead? 

(I suspect that either approach can lead to false positives depending on how the pattern are nested, so it might not be a case where there's no "better way" of doing this, it's just trade offs)

ntBre: Ah I think that's a good question. (This part was not intuitive to me at all, I was working on a more complicated fix when I realized this solved the issue)

I guess this could come into play with cases like this:

```py
match 42:
    case [{1: x} | [x] | [y]] | [y]: ...

match 42:
    case [{1: x} | [x] | [y]] | [x]: ...
```

We currently emit two diagnostics for the first and only one for the second:

```
invalid-syntax: alternative patterns bind different names
  --> /tmp/s.py:11:10
   |
10 | match 42:
11 |     case [{1: x} | [x] | [y]] | [y]: ...
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^
12 |
13 | match 42:
   |

invalid-syntax: alternative patterns bind different names
  --> /tmp/s.py:11:11
   |
10 | match 42:
11 |     case [{1: x} | [x] | [y]] | [y]: ...
   |           ^^^^^^^^^^^^^^^^^^
12 |
13 | match 42:
   |

invalid-syntax: alternative patterns bind different names
  --> /tmp/s.py:14:11
   |
13 | match 42:
14 |     case [{1: x} | [x] | [y]] | [x]: ...
   |           ^^^^^^^^^^^^^^^^^^
   |
```

Hmm, we actually get the same answer here with the union, but we get two errors in both cases with the intersection.

Maybe that makes the intersection the best option for consistency's sake?

I think I'd consider these all true positives at least, but they may be a bit redundant. CPython only emits the innermost diagnostic, so we could also consider that.

This is also pretty hard to hit. Because of the `break` after emitting a `DifferentMatchPatternBindings` diagnostic, you need at least two inner patterns with the same name, followed by one with a different name.

ntBre: Oops, the intersection breaks the tests, I guess because `self.names` is initially empty. So I guess we should either stick with replacing or the union.

MichaReiser: Can't we just not intersect until we visited at least one pattern? Like use an `Option<FxHashSet>` and only intersect when it's some?

ntBre: I may be misunderstanding the suggestion, but I tried a patch like the one below, and it caused some other tests to fail (and without changing the new diagnostics from this PR). It's not just when `self.names` is initially empty as I said before, this also affects cases like:

```py
match x:
     case [] | [a]: ...
```

where one of the branches is genuinely empty.

I initially included the .snap.new files in this patch, but they were way too big. We lost the diagnostics in the cases above.

I'm leaning toward sticking with the simple existing code unless I'm just missing something.

<details><summary>Patch</summary>
<p>

```diff
diff --git a/crates/ruff_python_parser/src/semantic_errors.rs b/crates/ruff_python_parser/src/semantic_errors.rs
index f35029d4b9..3312c72544 100644
--- a/crates/ruff_python_parser/src/semantic_errors.rs
+++ b/crates/ruff_python_parser/src/semantic_errors.rs
@@ -11,7 +11,7 @@ use ruff_python_ast::{
     visitor::{Visitor, walk_expr},
 };
 use ruff_text_size::{Ranged, TextRange, TextSize};
-use rustc_hash::{FxBuildHasher, FxHashSet};
+use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 use std::fmt::Display;
 
 #[derive(Debug, Default)]
@@ -117,7 +117,7 @@ impl SemanticSyntaxChecker {
                 Self::irrefutable_match_case(match_stmt, ctx);
                 for case in &match_stmt.cases {
                     let mut visitor = MatchPatternVisitor {
-                        names: FxHashSet::default(),
+                        names: None,
                         ctx,
                     };
                     visitor.visit_pattern(&case.pattern);
@@ -1671,7 +1671,7 @@ impl Visitor<'_> for ReboundComprehensionVisitor<'_> {
 }
 
 struct MatchPatternVisitor<'a, Ctx> {
-    names: FxHashSet<&'a ast::name::Name>,
+    names: Option<FxHashSet<&'a ast::name::Name>>,
     ctx: &'a Ctx,
 }
 
@@ -1810,15 +1810,19 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
                 let mut previous_names: Option<FxHashSet<&ast::name::Name>> = None;
                 for pattern in patterns {
                     let mut visitor = Self {
-                        names: FxHashSet::default(),
+                        names: None,
                         ctx: self.ctx,
                     };
                     visitor.visit_pattern(pattern);
                     let Some(prev) = &previous_names else {
-                        previous_names = Some(visitor.names);
+                        previous_names = visitor.names;
                         continue;
                     };
-                    if prev.symmetric_difference(&visitor.names).next().is_some() {
+                    if visitor
+                        .names
+                        .as_ref()
+                        .is_some_and(|names| prev.symmetric_difference(&names).next().is_some())
+                    {
                         // test_err different_match_pattern_bindings
                         // match x:
                         //     case [a] | [b]: ...
@@ -1857,7 +1861,13 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
                         );
                         break;
                     }
-                    self.names = visitor.names;
+                    if let Some(names) = &self.names
+                        && let Some(other) = visitor.names
+                    {
+                        self.names = Some(names.intersection(&other).cloned().collect());
+                    } else {
+                        self.names = visitor.names;
+                    }
                 }
             }
         }
@@ -1866,7 +1876,12 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
     /// Add an identifier to the set of visited names in `self` and emit a [`SemanticSyntaxError`]
     /// if `ident` has already been seen.
     fn insert(&mut self, ident: &'a ast::Identifier) {
-        if !self.names.insert(&ident.id) {
+        if self.names.is_none() {
+            self.names = Some(FxHashSet::default());
+        }
+        let names = self.names.as_mut().unwrap();
+
+        if !names.insert(&ident.id) {
             SemanticSyntaxChecker::add_error(
                 self.ctx,
                 SemanticSyntaxErrorKind::MultipleCaseAssignment(ident.id.clone()),
```

</p>
</details>

MichaReiser: That's not what I meant, I think? 

I'd created a `Option<FxHashSet>` right outside the `for pattern in patterns {` rather than globally in the visitor


But i also think that what we have is fine.

ntBre: We have one of those too:

https://github.com/astral-sh/ruff/blob/d38a5292d2a9c8509da6868046da0caa49ed5e46/crates/ruff_python_parser/src/semantic_errors.rs#L1809-L1811

We need an outer one in the visitor (or at least somewhere else) for recursive cases.

ntBre: I'll go ahead and land this for now then since it should resolve the bug!

PR #21109, opened 2025-10-28
[ty] Render `import <...>` in completions when "label details" isn't supported

BurntSushi:
This fixes a bug where the `import module` part of a completion for
unimported candidates would be missing. This makes it especially
confusing because the user can't tell where the symbol is coming from,
and there is no hint that an `import` statement will be inserted.

Previously, we were using [`CompletionItemLabelDetails`] to render the
`import module` part of the suggestion. But this is only supported in
clients that support version 3.17 (or newer) of the LSP specification.
It turns out that this support isn't widespread yet. In particular,
Heliex doesn't seem to support "label details."

To fix this, we take a [cue from rust-analyzer][rust-analyzer-details].
We detect if the client supports "label details," and if so, use it.
Otherwise, we push the `import module` text into the completion label
itself.

Fixes https://github.com/astral-sh/ruff/pull/20439#issuecomment-3313689568

[`CompletionItemLabelDetails`]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails
[rust-analyzer-details]: https://github.com/rust-lang/rust-analyzer/blob/5d905576d49233ed843bb40df4ed57e5534558f5/crates/rust-analyzer/src/lsp/to_proto.rs#L391-L404

### DIFF
--- a/crates/ty_server/src/server/api/requests/completion.rs
+++ b/crates/ty_server/src/server/api/requests/completion.rs
@@ -81,15 +81,30 @@ impl BackgroundDocumentRequestHandler for CompletionRequestHandler {
                         new_text: edit.content().map(ToString::to_string).unwrap_or_default(),
                     }
                 });
+
+                let name = comp.name.to_string();
+                let import_suffix = comp.module_name.map(|name| format!(" (import {name})"));
+                let (label, label_details) = if snapshot
+                    .resolved_client_capabilities()
+                    .supports_completion_item_label_details()
+                {
+                    let label_details = CompletionItemLabelDetails {
+                        detail: import_suffix,
+                        description: type_display.clone(),
+                    };
+                    (name, Some(label_details))
+                } else {
+                    let label = import_suffix
+                        .map(|suffix| format!("{name}{suffix}"))
+                        .unwrap_or_else(|| name);
+                    (label, None)
+                };
                 CompletionItem {
-                    label: comp.name.into(),
+                    label,
                     kind,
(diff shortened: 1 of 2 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/39ccbdb4-edfd-416b-b3f9-14d95e5fbe82

In this demo, I show the VS Code rendering, which is unchanged. Then I show Helix. Previously, it didn't include the modules for unimported symbols. But now it does!

crates/ty_server/src/server/api/requests/completion.rs
```diff
@@ -81,15 +81,30 @@ impl BackgroundDocumentRequestHandler for CompletionRequestHandler {
...
+                    .supports_completion_item_label_details()
+                {
+                    let label_details = CompletionItemLabelDetails {
+                        detail: import_suffix,
+                        description: type_display.clone(),
+                    };
+                    (name, Some(label_details))
+                } else {
+                    let label = import_suffix
+                        .map(|suffix| format!("{name}{suffix}"))
```
MichaReiser, line 98: Is there a leading whitespace at the start of `import_suffix` (I'm asking because it's hard to tell from Github's review page and I can't load your test plan with the onboard wifi ;)).

If not, should we add one?

BurntSushi: Yes there is! From above:

```rust
let import_suffix = comp.module_name.map(|name| format!(" (import {name})"));
```

Issue #9745, opened 2024-01-31
Allow newlines after block open (or not)


vaughnkoch:
Hi, I'm opening this issue per a suggestion from @MichaReiser, based on this closed issue: https://github.com/astral-sh/ruff/issues/8893.

The formatting question is whether Ruff should automatically remove a blank line after a block definition:

```
def some_func():

    some_more_code()
```

to 

```
def some_func():
    some_more_code()
```

The alternative is "allow zero or one" newlines, i.e. not to remove the newline if it exists, but also not to automatically add one. This is the approach that Black chose in their 2024 preview style. Full Black discussion here: https://github.com/psf/black/issues/4043.

ejovo13:
I second this. After exploring integrating ruff into my team's workflow as a replacement for black, isort, pydocstyle, etc., this is something I'm not willing to compromise on. Additionally, it was the first incompatibility with black that I discovered during testing. 

Excellent project though, I look forward to using it in the future.

cbandera:
Yes I can second this request as well. While investigating the transition to ruff, this inconsistency to black is causing many changes across our code base.

MichaReiser:
> Yes I can second this request as well. While investigating the transition to ruff, this inconsistency to black is causing many changes across our code base.

Thanks for reporting. Do you mainly mind that there are many differences or did you prefer black's formatting? Would you be able to share some examples where you preferred black's formatting over ruff's?

cbandera:
> > Yes I can second this request as well. While investigating the transition to ruff, this inconsistency to black is causing many changes across our code base.
> 
> Thanks for reporting. Do you mainly mind that there are many differences or did you prefer black's formatting? Would you be able to share some examples where you preferred black's formatting over ruff's?

For me it's mainly about the diff that is created. Working in a big monorepo, it's always easier for us toolies to roll out changes if we don't need to touch the other devs code. Having read that `ruff` would be a "drop-in" replacement for `black`, my expectation was that I would see no diff at all.

Having said that and after looking at the [known deviations](https://docs.astral.sh/ruff/formatter/black/) I can't say that I don't agree with ruff's design decisions though...

chicheng:
Hi, yes please add an option to preserve up to 1 empty line.

As empty lines used for separating/grouping lines, the first line(s) stay closer to block head but not the remaining parts feel like it's part of the block head (since empty line inside block content is inevitable), e.g.

```
block head
   part of block head

   block content
   block content
   block content

   block content
   block content
```

[...]

Issue #8893, opened 2023-11-29
Formatter: `allow_empty_first_line_before_new_block_or_comment` preview style


MichaReiser:
Implement Black's [`allow_empty_first_line_before_new_block_or_comment`](https://github.com/psf/black/pull/3967) as a Ruff preview style. 

```python
def foo():
    """
    Docstring
    """

    # Here we go
    if x:

        # This is also now fine
        a = 123

    else:
        # But not necessary
        a = 123
```

Notice how black now allows empty lines before the comments. Ruff seems to support this sometimes.

Goal: Implement the new formatting behind the preview flag and import the Black tests.

charliermarsh:
Heads up: it looks like this was changed to _always_ allow a blank line at the top of the block, _unless_ it's immediately followed by a docstring. See: https://github.com/psf/black/pull/4060.

charliermarsh:
Except, apparently, if it's a class immediately followed by a function: https://black.vercel.app/?version=main&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ACbAGldAD2IimZxl1N_WlXnON2nzNKOlH52D45tYviogiBy8ZlN26ZjqUAmZNQ-muxSkGcV-iAN9HZdhppScEmlx2Al7gJjtYIOGHyFQmjMSn_t7leQm1ctcrtU7IhNuHk3t8IsBD624xhI5TmvygAAAAC8TNkRzPz_WQABhQGcAQAALUwcRrHEZ_sCAAAAAARZWg==

MichaReiser:
I see the motivation behind this change but it can lead to unintentional deviations that outweigh the benefit of the change. 

Let's say you start with

```python
def foo():
    """
    Docstring
    """

    # Here we go
    if x:
		if no_longer_needed:
			print("Some complex statements")

        # This is also now fine
        a = 123
```

And you refactor the code to remove the `no_longer_needed` branch

```python
def foo():
    """
    Docstring
    """

    # Here we go
    if x:

        # This is also now fine
        a = 123
```

and you now hit save. Without this change, the comment and `a = 123` gets moved directly under the `if x:` condition which is what I would expect and gives us consistent formatting. With this change, this no longer happens and I have to manually remove the new line (or won't do it, which leads to inconsistent code).

vaughnkoch:
Hi, thanks for building Ruff! 

[...]

PR #21110, opened 2025-10-28
Allow newlines after function headers without docstrings

ntBre:
Summary
--

This is a first step toward fixing #9745. After reviewing our open issues and several Black issues and PRs, I personally found the function case the most compelling, especially with very long argument lists:

```py
def func(
	self,
	arg1: int,
	arg2: bool,
	arg3: bool,
	arg4: float,
	arg5: bool,
) -> tuple[...]:
	if arg2 and arg3:
		raise ValueError
```

or many annotations:

```py
def function(
    self, data: torch.Tensor | tuple[torch.Tensor, ...], other_argument: int
) -> torch.Tensor | tuple[torch.Tensor, ...]:
    do_something(data)
    return something
```

I think docstrings help the situation substantially both because syntax highlighting will usually give a very clear separation between the annotations and the docstring and because we already allow a blank line _after_ the docstring:

```py
def function(
    self, data: torch.Tensor | tuple[torch.Tensor, ...], other_argument: int
) -> torch.Tensor | tuple[torch.Tensor, ...]:
    """
	A function doing something.

	And a longer description of the things it does.
	"""

    do_something(data)
    return something
```

There are still other comments on #9745, such as [this one] with 9 upvotes, where users specifically request blank lines in all block types, or at least including conditionals and loops. I'm sympathetic to that case as well, even if personally I don't find an [example] like this:

```py
if blah:

    # Do some stuff that is logically related
    data = get_data()

    # Do some different stuff that is logically related
    results = calculate_results()

    return results
```

to be much more readable than:

```py
if blah:
    # Do some stuff that is logically related
    data = get_data()

    # Do some different stuff that is logically related
    results = calculate_results()

    return results
```

I'm probably just used to the latter from the formatters I've used, but I do prefer it. I also think that functions are the least susceptible to the accidental introduction of a newline after refactoring described in Micha's [comment] on #8893.

I actually considered further restricting this change to functions with multiline headers. I don't think very short functions like:

```py
def foo():

    return 1
```

benefit nearly as much from the allowed newline, but I just went with any function without a docstring for now. I guess a marginal case like:

```py
def foo(a_long_parameter: ALongType, b_long_parameter: BLongType) -> CLongType:

    return 1
```

might be a good argument for not restricting it.

I caused a couple of syntax errors before adding special handling for the ellipsis-only case, so I suspect that there are some other interesting edge cases that may need to be handled better.

Test Plan
--

Existing tests, plus a few simple new ones. As noted above, I suspect that we may need a few more for edge cases I haven't considered.

[this one]: https://github.com/astral-sh/ruff/issues/9745#issuecomment-2876771400
[example]: https://github.com/psf/black/issues/902#issuecomment-1562154809
[comment]: https://github.com/astral-sh/ruff/issues/8893#issuecomment-1867259744

### DIFF
--- a/crates/ruff_python_formatter/src/statement/suite.rs
+++ b/crates/ruff_python_formatter/src/statement/suite.rs
@@ -236,7 +236,7 @@ impl FormatRule<Suite, PyFormatContext<'_>> for FormatSuite {
                     )?;
                 } else {
                     // Preserve empty lines after a stub implementation but don't insert a new one if there isn't any present in the source.
-                    // This is useful when having multiple function overloads that should be grouped to getter by omitting new lines between them.
+                    // This is useful when having multiple function overloads that should be grouped together by omitting new lines between them.
                     let is_preceding_stub_function_without_empty_line = following
                         .is_function_def_stmt()
                         && preceding


ntBre:
The ecosystem results look good to me. There were only a couple of somewhat surprising cases:

- https://github.com/RasaHQ/rasa/blob/b8de3b231126747ff74b2782cb25cb22d2d898d7/tests/nlu/conftest.py#L79
  ```py
	 @pytest.fixture()
	def process_message(default_model_storage: ModelStorage) -> Callable[..., Message]:
	    def inner(loaded_pipeline: List[GraphComponent], message: Message) -> Message:
	
	        for component in loaded_pipeline:
	            component.process([message])
	
	        return message
	
	    return inner
  ```
- https://github.com/snowflakedb/snowflake-cli/blob/b0306e131063d26be11f14cb1f9dbd9ae67b32c8/src/snowflake/cli/_plugins/snowpark/snowpark_entity.py#L44
  ```py
	class SnowparkEntity(EntityBase[Generic[T]]):
	    def __init__(self, *args, **kwargs):
	
	        if not FeatureFlag.ENABLE_NATIVE_APP_CHILDREN.is_enabled():
	            raise NotImplementedError("Snowpark entity is not implemented yet")
	        super().__init__(*args, **kwargs)
	```

These both feel a bit awkward to me because they preserve a blank line inside the inner functions but not outside. But I guess we're only preserving the input whitespace, so this might still be what the authors intended.

crates/ruff_python_formatter/tests/snapshots/format@newlines.py.snap
```diff
@@ -342,6 +341,43 @@ def overload4():
...
+def removed3():
+
+    ...
+
+
+# And we discard empty lines after the first:
+def partially_preserved1():
+
+
+    return 1
```
amyreese, line 416: Is this expected? (edit: I'm guessing yes, missed the snapshot change further down with the single empty line)

amyreese (approved):

crates/ruff_python_formatter/tests/snapshots/format@newlines.py.snap
ntBre: You scared me! 😄 

I'm still getting used to the input appearing in the formatter snapshots.

crates/ruff_python_formatter/resources/test/fixtures/ruff/newlines.py
```diff
@@ -335,3 +335,40 @@ def overload4():
...
+def removed3():
+
+    ...
+
+
+# And we discard empty lines after the first:
+def partially_preserved1():
+
+
+    return 1
```
MichaReiser, line 410: Can we add more tests to this, specifically tests including comments or cases where the first element is a function on its own?

crates/ruff_python_formatter/src/statement/suite.rs
```diff
@@ -169,6 +171,21 @@ impl FormatRule<Suite, PyFormatContext<'_>> for FormatSuite {
...
             )
         } else {
+            // Allow an empty line after a function header in preview, if the function has no
+            // docstring and no initial comment and doesn't consist of a single ellipsis.
+            let allow_newline_after_block_open =
+                is_allow_newline_after_block_open_enabled(f.context())
+                    && matches!(self.kind, SuiteKind::Function)
+                    && matches!(first, SuiteChildStatement::Other(_))
+                    && !comments.has_leading(first)
+                    && !contains_only_an_ellipsis(statements, f.context().comments());
```
MichaReiser: This elipsis check should exactly match the check we do in the function formatting (also test for trailing comments)

https://github.com/astral-sh/ruff/blob/c4506f0e94882fa1118eb2a13339f9be7bf0bcaa/crates/ruff_python_formatter/src/statement/clause.rs#L451-L453

crates/ruff_python_formatter/src/statement/suite.rs
```diff
@@ -169,6 +171,21 @@ impl FormatRule<Suite, PyFormatContext<'_>> for FormatSuite {
...
+
+            if allow_newline_after_block_open
+                && lines_before(first.start(), f.context().source()) > 1
```
MichaReiser: I think this is correct but maybe worth adding a test that we collapse any empty line before a comment:

```py
def test():

    # comment

    a

```

gets formatted to 

```py
def test():
    # comment

    a
```

crates/ruff_python_formatter/tests/snapshots/format@range_formatting__indent.py.snap
```diff
@@ -149,6 +149,21 @@ def test6 ():
...
+--- Stable
++++ Preview
+@@ -55,6 +55,7 @@
+ 
+ 
+ def test6 ():
++
+     print("Format")
+     print(3 + 4)
+     print("Format to fix indentation" )
```
MichaReiser, line 175: Can we add another test like this. As I don't think the original I wrote captured what I wanted to test

```py
def test6 ():
    <RANGE_START>print("Format" )
    print(3 +  4)<RANGE_END>
    print("Format to fix indentation" )
```

crates/ruff_python_formatter/src/statement/suite.rs
ntBre: Is that different from the new test here?

https://github.com/astral-sh/ruff/blob/abb54b32b0bbf5becfdf58fb9aae5e10b951141a/crates/ruff_python_formatter/resources/test/fixtures/ruff/newlines.py#L358-L362

It looks like the formatting got slightly mangled in your comment.

crates/ruff_python_formatter/src/statement/suite.rs
ntBre: I added two new tests trying to exercise this, and they both seemed to work as I expected without any additional check:

```py
def preserved5():

    ...
    # trailing comment prevents collapsing the stub


def removed4():

    ...  # trailing same-line comment does not prevent collapsing the stub
```

and in fact when I updated this condition to:

```rust
            let will_collapse_stub_function =
                contains_only_an_ellipsis(statements, f.context().comments())
                    && !first_comments.has_trailing();
```

I caused another syntax error for the `removed4` case. `contains_only_an_ellipsis` already has one kind of check for trailing comments:

https://github.com/astral-sh/ruff/blob/d38a5292d2a9c8509da6868046da0caa49ed5e46/crates/ruff_python_formatter/src/statement/suite.rs#L730-L739

but I'm very likely to be missing something. Is [`first_comments`](https://github.com/astral-sh/ruff/blob/d38a5292d2a9c8509da6868046da0caa49ed5e46/crates/ruff_python_formatter/src/statement/suite.rs#L151) not what I should be checking?

ntBre:
> I went through the original issue and I noticed that we don't preserve the empty line for

Should we try preserving the empty line for any `if`, any `if` with a "long"/compound condition, and/or where the body starts with a comment? Just checking which part of the example to focus on.

One reason I was wary of the conditionals is that they don't seem much different from loops. For [example](https://play.ruff.rs/de50a314-446f-41c3-894e-48e793ddcd0e?secondary=Format):

```py
for i in some_generator_function(
    with_many_arguments=[1, 2, 3], another=(4, 5, 6), one_more=None
):

    # a comment
    print("body of the loop")
```

Should we handle those too?

Functions seemed like a good place to draw the line to me, but I can see an argument for more consistency too.

MichaReiser:
> Should we try preserving the empty line for any if, any if with a "long"/compound condition, and/or where the body starts with a comment? Just checking which part of the example to focus on.

No sorry. I think I tested it with a function header but then copied the original example from the issue, but maybe I'm just jetlagged and tested it with a function header?

So what I think we should preserve is:


```py
def foo():

    # A pretty long comment documenting what the block does
    print("Do stuff")
```

ntBre:
Ohh okay, that makes sense! So preserve lines before comments instead of removing them. I think that does make sense.

At least one person on one of the Black issues even wanted to preserve empty lines before docstrings. I don't feel too strongly about either comments or docstrings, I was just trying to narrow the scope even further.

MichaReiser:
> At least one person on one of the Black issues even wanted to preserve empty lines before docstrings. I don't feel too strongly about either comments or docstrings, I was just trying to narrow the scope even further.

The docstrings make sense to me (not allowing an empty line) and is sort of named exactly in the rule. The comments just were a surprise to me (but I don't feel strongly about it)

ntBre:
Oops, last commit was too simplistic. We're now adding blank lines before comments rather than just preserving them. That looks like the majority of new ecosystem hits.

crates/ruff_python_formatter/src/statement/suite.rs
ntBre: Ah I see, this is what's coming up in the ecosystem check now. I accidentally skirted around this by not preserving cases with any leading comments.

ntBre: `lines_before` may not be what I want here, if we want to preserve blank lines before comments. All of these are returning 2 `lines_before`:

```py
def untouched1():

    # comment
    # more lines in the comment
    # and one more

    return 0


def untouched2():

    # comment

    return 0


def untouched3():
    # comment

    return 0
```

MichaReiser: That, or you'd have to pass the position of the first leading comment as the offset

MichaReiser: Note, we do have a `lines_after_ignore_trivia` (or similar) that uses the tokenizer over doing a string search

ntBre: Ah I didn't consider that I could pass a comment offset there. I assumed it had to be a statement node. Thanks!

I saw `lines_after_ignore_trivia`, but not a `before` variant. Happy to try that approach if it would be better, but using the comment offset seems to be working well.

ntBre:
The ecosystem check looks good to me again. I didn't click through to every change, but I downloaded the full output from the Actions log and checked more than the subset in the comment.

~~This might be expected, but most of the links don't point to the exact line of the diff, which makes it a bit more of a manual process.~~ I checked the script and this is expected.

crates/ruff_python_formatter/src/statement/suite.rs
MichaReiser: Maybe it's not relevant. You have to trace through `FormatStmtFunctionDef`. 

But yes, `first_comments` is the wrong thing to test. `trailing_comments` in `FormatClauseBody` are the `trailing_definition_comments` in `FormatStmtFunctionDef`. So I think it's more about something like 

```py
def test(): # comment


	...
```

ntBre: Ah okay, I see now. Thanks! Yes, your example comment prevents collapsing the stub in `FormatClauseBody`, but we remove the newline because the logic is different.

Is there a way to pass this `should_collapse_stub` information into the suite formatting (maybe another `with_options` implementation?)? Or alternatively a way to retrieve the parent function definition so that we could recompute it?

ntBre: As we discussed on Discord, I added a variant of `contains_only_an_ellipsis` that returns the ellipsis statement, which we can then format directly in `FormatClauseBody` instead of calling into the suite formatting at all. Now the blank line in this example is correctly preserved.

MichaReiser (approved):

astralcai:
Thank you for making this change! I see that this feature is only available in preview. How do I enable this behaviour without enabling all other rules in preview?

MichaReiser:
You can set `format.preview` which enables all preview formatting changes. There's no feature-specific flag. You'll have to wait for our next minor where we plan on stabilizing this feature.

edu-nixon:
> You can set `format.preview` which enables all preview formatting changes. There's no feature-specific flag. You'll have to wait for our next minor where we plan on stabilizing this feature.

Hi @MichaReiser, thanks for approving this rule. Quick question, does the flag exist now? I tried to search it myself with no luck

Thank you!

PR #21111, opened 2025-10-28
[ty] Reformat constraint set mdtests

dcreager:
This PR updates the mdtests that test how our generics solver interacts with our new constraint set implementation. Because the rendering of a constraint set can get long, this standardizes on putting the `revealed` assertion on a separate line. We also add a `static_assert` test for each constraint set to verify that they are all coerced into simple `bool`s correctly.

This is a pure reformatting (not even a refactoring!) that changes no behavior. I've pulled it out of #20093 to reduce the amount of effort that will be required to review that PR.


dcreager:
Rebased this onto `main` — it doesn't actually depend on anything in #21068, and since it's got a :heavy_check_mark: I'll go ahead and merge this

PR #21114, opened 2025-10-28
[ty] Add error counter to the ty benchmark package

alex-gregory-ds:
## Summary

Related to https://github.com/astral-sh/ty/issues/241.

This adds a counter to count the number of errors that are produced by ty on each of the benchmark repositories in `ty_benchmark`. For simplicity, we count the numer of lines produce by ty (when using `--output-format=concise`) as a proxy for the number of errors instead of counting the actual number of errors.

## Test Plan

This is a draft so it's still a work in progress.

### DIFF
--- a/scripts/ty_benchmark/src/benchmark/error_count.py
+++ b/scripts/ty_benchmark/src/benchmark/error_count.py
@@ -88,14 +88,11 @@ class ErrorCountSummary:
             ],
             widths,
         )
-        self._print_line(widths)
 
         old_datetime = old.date_time.strftime("%Y-%m-%d %H:%M:%S")
         new_datetime = new.date_time.strftime("%Y-%m-%d %H:%M:%S")
-        datetime_difference = str(new.date_time - old.date_time)
-        self._print_row(
-            ["datetime", old_datetime, new_datetime, datetime_difference], widths
-        )
+        self._print_row(["", old_datetime, new_datetime, ""], widths)
+        self._print_line(widths)
 
         all_benchmark_names = list(old.error_counts.keys()) + list(
             new.error_counts.keys()


alex-gregory-ds:
Can anyone give me some guidance on where to go from here please? I have a function that counts the number of errors (number of lines) produced by ty. I assume I need to log the number of errors somewhere but:

* I cannot find where `ty_benchmark` is run in the CI/CD pipeline.
* I cannot see how the current Hyperfine benchmark results are logged.

MichaReiser:
The `ty_benchmark` doesn't run on CI. It's a script that we run manually when comparing different type checkers. 

>  I assume I need to log the number of errors somewhere but:

I think what we want is to hard code a rough number of expected diagnostics and compare them with the diagnostics count when running ty. Logging the diagnostic count when there's a mismatch would certainly be useful.

However, I'm not entirely convinced that this is indeed necessary. We could also run ty with `--exit-zero` and have `hyperfine` assert that the exit code is 0.

alex-gregory-ds:
Thanks for the suggestions. This PR includes an `error_count.json` file and every time the benchmark is run, a new file is created called `error_count_new.json`. At the end of the benchmark run, the difference between the two JSON files is printed to the terminal. Here is an example:

```
+--------------------------------------------------------------------------------------+
|                                 Error Count Summary                                  |
+-----------------+----------------------+----------------------+----------------------+
|                 | Old (6a92f1d9)       | New (3ffe3eea)       | Difference           |
|                 | 2025-11-08 11:55:29  | 2025-11-08 12:09:53  |                      |
+-----------------+----------------------+----------------------+----------------------+
| black (cold)    | 44                   | 44                   | 0                    |
| isort (cold)    | 21                   | 21                   | 0                    |
| jinja (cold)    | 81                   | 81                   | 0                    |
+-----------------+----------------------+----------------------+----------------------+
```

Let me know if this is right direction or if it's too much for what you need. If this isn't the right direction, I am happy to close this and try again :).

Currently, I am not running `ty` with `--exit-zero`, and I am not doing this with `hyperfine` but I am happy to make those changes if you'd prefer.

MichaReiser:
CC: @AlexWaygood

MichaReiser:
Thank you @alex-gregory-ds for working on this. 

I realized while working on adding and updating our benchmarks that a much easier approach is just to snapshot the output, and I implemented this as part of https://github.com/astral-sh/ruff/pull/21536. It's certainly a bit verbose, but it has the added benefit that it allows comparison between versions. 

I'm sorry that I misled you towards comparing the error counts instead. I might use your approach if the team decides that my solution is too verbose. But for now, I'll close this PR

alex-gregory-ds:
> I'm sorry that I misled you towards comparing the error counts instead. I might use your approach if the team decides that my solution is too verbose. But for now, I'll close this PR.

No problem at all. Thank you for getting back to me.

Issue #16802, opened 2025-03-17
[Infinite loop] I002 and UP035


dscorbett:
### Summary

Ruff fails to converge when [`missing-required-import` (I002)](https://docs.astral.sh/ruff/rules/missing-required-import/) and [`deprecated-import` (UP035)](https://docs.astral.sh/ruff/rules/deprecated-import/) are selected and a deprecated import is required.

```console
$ echo 1 | ruff --isolated check - --select I002,UP035 --config 'lint.isort.required-imports=["from collections import Sequence"]' --fix

error: Failed to converge after 100 iterations.

This indicates a bug in Ruff. If you could open an issue at:

    https://github.com/astral-sh/ruff/issues/new?title=%5BInfinite%20loop%5D

...quoting the contents of `-`, the rule codes I002, along with the `pyproject.toml` settings and executed command, we'd be very appreciative!

from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
from collections.abc import Sequence
1
-:1:1: I002 Missing required import: `from collections import Sequence`
[...]

PR #21115, opened 2025-10-28
[`ruff`] Prevent PYI025 from clashing with isort-required check

### LINKED ISSUE #20891: [Infinite loop] F401, I002, and PYI025
### Summary

Ruff fails to converge when [`unused-import` (F401)](https://docs.astral.sh/ruff/rules/unused-import/), [`missing-required-import` (I002)](https://docs.astral.sh/ruff/rules/missing-required-import/), and [`unaliased-collections-abc-set-import` (PYI025)](https://docs.astral.sh/ruff/rules/unaliased-collections-abc-set-import/) are selected and `from collections.abc import Set` is a required import.
```console
$ echo 1 | ruff --isolated check --select F401,I002,PYI025 --config 'lint.isort.required-imports = ["from collections.abc import Set"]' - --unsafe-fixes --diff

...quoting the contents of `-`, the rule codes PYI025, along with the `pyproject.toml` settings and executed command, we'd be very appreciative!

@@ -1 +1,2 @@
+from collections.abc import Set
 1

Would fix 100 errors.
```

### Version

TaKO8Ki:
## Summary

Fixes #20891

## Test Plan

Added `required_import_skips_pyi025`.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_pyi/rules/unaliased_collections_abc_set_import.rs
+++ b/crates/ruff_linter/src/rules/flake8_pyi/rules/unaliased_collections_abc_set_import.rs
@@ -1,10 +1,12 @@
 use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::{Stmt, StmtImportFrom};
 use ruff_python_semantic::Imported;
 use ruff_python_semantic::{Binding, BindingKind, Scope};
 use ruff_text_size::Ranged;
 
 use crate::checkers::ast::Checker;
 use crate::renamer::Renamer;
+use crate::rules::pyupgrade::rules::is_import_required_by_isort;
 use crate::{Applicability, Fix, FixAvailability, Violation};
 
 /// ## What it does
@@ -70,6 +72,20 @@ pub(crate) fn unaliased_collections_abc_set_import(checker: &Checker, binding: &
 
     let name = binding.name(checker.source());
     if name == "AbstractSet" {
+        return;
+    }
+
+    // Skip if this import is required by isort to prevent infinite loops with I002 and F401
+    if let Some(stmt @ Stmt::ImportFrom(StmtImportFrom { names, .. })) =
+        binding.statement(checker.semantic())
+        && names.iter().any(|alias| {
+            is_import_required_by_isort(
+                &checker.settings().isort.required_imports,
+                stmt.into(),
+                alias,
+            )
+        })
+    {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


ntBre (commented):
Thank you!

I confirmed locally that this fixes the issue, but this test doesn't reproduce the error from the issue when I revert the rule change. I think you need to include a flag like `--fix` or `--diff` in the `Command` to reproduce the infinite loop.

crates/ruff_linter/src/rules/flake8_pyi/rules/unaliased_collections_abc_set_import.rs
```diff
@@ -73,6 +75,20 @@ pub(crate) fn unaliased_collections_abc_set_import(checker: &Checker, binding: &
...
+        && names.iter().any(|alias| {
+            is_import_required_by_isort(
+                &checker.settings().isort.required_imports,
+                stmt.into(),
+                alias,
+            )
+        })
+    {
+        return;
+    }
```
AlexWaygood, line 90: I don't think this is the best fix. To me I think it would make more sense to reject this configuration entirely at an earlier stage. Fundamentally, it just doesn't make sense to tell Ruff that you simultaneously want `from collections.abc import Set` to always appear in every module, and also `from collections.abc import Set` to never appear in any module.

That implies that we'd want to edit this part of the code: https://github.com/astral-sh/ruff/blob/aca8ba76a489f2ec43958868254de2ce14c8e9d2/crates/ruff_workspace/src/configuration.rs#L1614-L1651

ntBre: I thought about that too. This seemed like a pretty niche issue, so I didn't feel too strongly, but this also makes sense to me.

AlexWaygood: I generally have a pretty strong preference towards rejecting invalid configurations rather than silently accepting them and assuming we understand what the user _actually_ wanted 😄

ntBre: We've done a bit of both with these conflicts, I think. https://github.com/astral-sh/ruff/issues/16802 is one example, but maybe that's different :slightly_smiling_face:

TaKO8Ki: Ok. I will check it.

MichaReiser:
Thanks for working on this. I'll close this PR due to inactivity. Feel free to resubmit the changes with the feedback addressed.

Issue #19933, opened 2025-08-16
Performance lint for missed CONTAINS_OP LOAD_CONST optimization


Zaczero:
### Summary

I've recently found out that some people are unaware that **CONTAINS_OP**'s **LOAD_CONST** optimization applies only to containers with trivial values. As long as the RHS is non-trivial, Python will not apply **LOAD_CONST** optimization and will build the container each time we do **CONTAINS_OP**. This is a *major performance bottlneck* that many people may accidentally stumble upon. I think there's great potential in having a new rule that checks for this kind of inefficiency.

I prepared an example that hopefully explains my idea and labels which cases should be considered for a warning:
https://godbolt.org/z/radT91rdn

For example, `0 in {'foo': 1, 'bar': 2}` is much more efficient in the form of `0 == 'foo' or 0 == 'bar'`, so manual iteration. An alternative would be to declare a variable for computing RHS once but that may not always be feasible.

There is one potential issue: in the original code, all of the RHS is run, which can cause some side effects. After rewriting it to a sequence of `or...or` operations, Python may not run all the cases, potentially changing runtime behavior. Although I think such code would be of very poor quality in the first place.

ntBre:
Thanks for the suggestion! This is cool, and I didn't know godbolt had a Python option.

One thing that concerns me about making this a lint rule is that it seems very dependent upon CPython implementation details that might vary between CPython versions and especially other implementations of Python.

This also seems partially covered by two of our other rules:
- [literal-membership (PLR6201)](https://docs.astral.sh/ruff/rules/literal-membership/#literal-membership-plr6201)
- [single-item-membership-test (FURB171)](https://docs.astral.sh/ruff/rules/single-item-membership-test/#single-item-membership-test-furb171)

but not the dict case in your example, as far as I can tell.

Zaczero:
> This also seems partially covered by two of our other rules

Not quite, my issue concerns the set/list being constructed from scratch whenever it's being tested. Some people are unaware that this LOAD_CONST optimization only applies to Python trivial values and when such optimization is missed, there's a significant performance degradation.

---

> One thing that concerns me about making this a lint rule is that it seems very dependent upon CPython implementation details that might vary between CPython versions and especially other implementations of Python.

[...]

PR #21116, opened 2025-10-28
[`ruff`] Implement `inefficient-membership-test` (`RUF066`)

Zaczero:
Closes https://github.com/astral-sh/ruff/issues/19933

## Summary

Add new rule to detect membership tests against list, tuple, or set literals containing complex elements that prevent Python's `LOAD_CONST` bytecode optimization.

When containers have complex elements (nested lists, dicts, function calls, operations), Python reconstructs the container on every membership test, causing performance degradation in hot code paths.

The rule detects:
- Lists/tuples with complex elements (nested containers, dicts, function calls, operations, lambdas)
- Sets with complex elements (function calls, operations, etc.)
- Both `in` and `not in` operators

Provides unsafe autofix converting to equality comparisons:
- `x in [a, b]` → `x == a or x == b`
- `x not in [a, b]` → `x != a and x != b`

Fix is marked unsafe because it changes evaluation semantics (short-circuit behavior and potential side effects in custom `__eq__`).

## Test Plan

Implementing RUFF066.py test case.

I verified locally whether this performance optimization applies to non-trivial lists and sets of any size, and it appears it does. There's always a performance win, but the smaller the collection, the better.


Zaczero:
I wanted to push forward the idea of https://github.com/astral-sh/ruff/issues/19933 and arrogantly implemented a functional rule. I am really new to ruff and have fairly basic Rust knowledge so it may need improvement.

cc @ntBre since you interacted

crates/ruff_linter/src/rules/ruff/rules/inefficient_membership_test.rs
```diff
@@ -0,0 +1,197 @@
...
+/// if item in VALID_ITEMS:
+///     ...
+/// ```
+///
+/// ## Fix safety
+/// The fix is marked as unsafe because:
+/// - Converting `x in container` to `x == a or x == b` changes short-circuit
+///   evaluation behavior (original evaluates all elements before membership test).
+/// - Custom `__eq__` implementations may have side effects.
+/// - Comments within the expression might be lost during the transformation.
```
dscorbett, line 53: More reasons:
* `in` also checks for identity: compare `math.nan in [math.nan]` to `math.nan == math.nan`.
* The left argument of `in` may have a side effect.
* `in` always returns a `bool` but `==` can return anything.

crates/ruff_linter/src/rules/ruff/rules/inefficient_membership_test.rs
```diff
@@ -0,0 +1,197 @@
...
+    match expr {
+        // Literals are const-folded
+        _ if expr.is_literal_expr() => false,
+
+        // Tuples of literals are const-folded
+        Expr::Tuple(ast::ExprTuple { elts, .. }) => elts.iter().any(is_complex_element),
+
+        // Operations on literals are const-folded (e.g., `1+2` → `3`)
+        Expr::BinOp(ast::ExprBinOp { left, right, .. }) => {
+            is_complex_element(left) || is_complex_element(right)
```
dscorbett, line 138: Not all operations on literals are const-folded, e.g. `"%s" % 3`.

ntBre:
Thanks for your work on this! I quickly skimmed the implementation, and it looks good overall :)

However, we're not really prioritizing adding new rules at the moment, so I'd like to hold off a bit longer to see if there's more interest in this rule. I'll approve the workflow run so that we can get an ecosystem report, though.

crates/ruff_linter/src/rules/ruff/rules/inefficient_membership_test.rs
Zaczero: TYVM for very nice findings! I'll resolve it when and if ruff decides to include such rule.

MichaReiser:
We've been hesitant about adding rules that encode specific optimizations that Python does because that can change between Python versions and is out of our control or might even become irrelevant. That's why this rule will require some more discussion before being added and I'll close this PR for now, as there's no clear path forward. 

But thanks for your work. It's a great starting point if we decide to implement this rule.

Zaczero:
No prob, I was surprised by how convenient it was to add a new rule and test it in Ruff. Good job on that!

MichaReiser:
Thanks for the kind words and for being so understanding.

Issue #5841, opened 2023-07-17
Implement flake8-import-restrictions


justinchuby:
When following the Google style guide and to avoid cyclic imports, we would really like to see the rule from https://github.com/atollk/flake8-import-restrictions#imr241

(import modules only and not the functions and classes from the modules)

> IMR241
> When using the from syntax, only submodules are imported, not module elements.

```
# Bad
from os.path import join

# Good
from os import path
```

justinchuby:
So it’s a duplicate of https://github.com/astral-sh/ruff/issues/3045, but I found flake8-import-restrictions which has a good collection of related rules.

edgarrmondragon:
It's not exactly equivalent, but the [`banned-from`](https://beta.ruff.rs/docs/settings/#flake8-import-conventions-banned-from) setting can get a similar violation with this config:

```toml
# ruff.toml
[flake8-import-conventions]
banned-from = ["os.path"]
```

```console
$ ruff check --select ICN myfile.py
myfile.py:2:1: ICN003 Members of `os.path` should not be imported explicitly
Found 1 error.
```

justinchuby:
True but it is impossible to configure the list for all possible packages. It would be good to recognize an imported entity is not a module and error on that.

charliermarsh:
We would need to integrate our Python module resolver to support this (doable).

QuentinSoubeyranAqemia:
I would also be interested in a rule to disallow importing members, and only allow importing module (or packages, but they are module too technically speaking), per the Google Code Style.
Having a config option for exceptions (like the `typing`, `typing_extensions` and `collections.abc` defined in the google style guide) would be most useful too.

I'm curious what is this "Python module resolver"? Is this an upcoming feature? Some internal that would need to be exposed to rules? Something devs are thinking of doing?

segoon:
Hello,

I might miss something, but AFACS the technique used in https://github.com/atollk/flake8-import-restrictions/blob/master/flake8_import_restrictions/imports_submodule.py#L35 is vulnerable to arbitrary code execution due to the loading of python module from the working directory. I'd not run such a check on my system account with a random foreign code.

[...]

Issue #3045, opened 2023-02-20
Rule to enforce import module only


justinchuby:
The google style guide recommends importing modules only. https://google.github.io/styleguide/pyguide.html#22-imports. It would be great if there’s a rule to enforce that.

Avasam:
If I understand correctly, you'd want https://docs.astral.sh/ruff/rules/banned-import-from/ but for all imports ? (which could be done if globbing is supported, see https://github.com/astral-sh/ruff/issues/2660 )

justinchuby:
Not exactly. From import should still be allowed if the imported name is a module. 

E.g.

from torch import n

should be allowed because nn is a module.

qthequartermasterman:
https://github.com/Enforcer/pylint_google_style_guide_imports_enforcing exists as a reference implementation. It is a pylint plugin.

revmischa:
I opened a PR to enforce the google style guide rules #21118

qthequartermasterman:
It's tragic the PR was closed. The [pylint plugin](https://github.com/Enforcer/pylint_google_style_guide_imports_enforcing) enforcing this feature is the only things keeping us on pylint still.

serjflint:
@ntBre Hello! Is it possible to contribute this feature as on opt-in option to isort section of the formatter instead of a separate rule for the linter?

ntBre:
Hi @serjflint! Thanks for checking! But this doesn't seem like it has much overlap with isort to me. This is about linting against particular types of imports, not about grouping or sorting them in a certain way, unless I'm misunderstanding something.

serjflint:
> Hi @serjflint! Thanks for checking! But this doesn't seem like it has much overlap with isort to me. This is about linting against particular types of imports, not about grouping or sorting them in a certain way, unless I'm misunderstanding something.

I meant forcing imports to import only modules. But knowing which is a module or an object inside a module is a bit outside of scope of the formatter.

paveldikov:
Is it possible to reconsider the closure of the previous PR? It seemed like a very simple implementation, and it being gated behind preview made sense.

Lots of enterprise 'Python house styles' (not just Google's!) recommend/mandate module imports (reduction of naming tautologies and namespace pollution being some of the key benefits).

At least in my world, currently we have to police this by hand -- or hope that AI assistants/reviewers will respect the rule. Obviously that is rather futile. I'd much rather rely on the linter instead! :-)

PR #21118, opened 2025-10-29
[`ruff`] Implement `direct-member-import` (RUF066)

revmischa:
## Summary

Implements **RUF066** (`direct-member-import`) to enforce importing modules instead of specific members, following [section 2.2 of the Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#22-imports).

This rule helps make code origins more explicit by flagging `from module import member` patterns and encouraging `import module` instead.

## Rule Behavior

**Flags:**
```python
from os import path  # RUF066
from pathlib import Path  # RUF066
from collections import defaultdict  # RUF066
from json import loads, dumps  # RUF066 (reports twice)
```

**Exemptions** (per Google Style Guide 2.2.4.1):
- `typing` and `typing_extensions` modules - type hints are commonly imported directly
- `collections.abc` module - abstract base classes  
- `six.moves` module - Python 2/3 compatibility
- Relative imports (e.g., `from . import foo`)
- Wildcard imports (e.g., `from module import *`)

**Example violations:**
```python
# Bad - violates RUF066
from os import path
result = path.join("a", "b")

# Good - passes RUF066  
import os
result = os.path.join("a", "b")
```

## Implementation Details

- **Rule location:** `crates/ruff_linter/src/rules/ruff/rules/direct_member_import.rs`
- **Rule code:** RUF066
- **Preview mode:** `preview_since = "0.15.0"`
- **Integration:** Checks `StmtImportFrom` nodes in AST analysis
- **Test fixture:** `crates/ruff_linter/resources/test/fixtures/ruff/RUF066.py`

## Test Plan

- ✅ Compiles successfully with `cargo build`
- ✅ Added comprehensive test fixture covering violations and exemptions
- 🔲 Snapshot tests (require `cargo insta test --accept`)
- 🔲 Full validation suite (clippy, schema update, pre-commit)

## References

- Google Python Style Guide: https://google.github.io/styleguide/pyguide.html#22-imports
- Exemptions follow section 2.2.4.1 exactly


ntBre:
Thanks for working on this! I believe this is related to issues https://github.com/astral-sh/ruff/issues/5841 and https://github.com/astral-sh/ruff/issues/3045, although one of the examples in the summary here is more restrictive than those in the issues.

However, these are both marked as `needs-decision` and we aren't prioritizing adding new rules at the moment, so I don't think we can move this forward currently.

crates/ruff_linter/resources/test/fixtures/ruff/RUF066.py
```diff
@@ -0,0 +1,26 @@
+# Test cases for RUF066 - DirectMemberImport
+
+# Should trigger violations
+from os import path  # RUF066
+from pathlib import Path  # RUF066
+from collections import defaultdict  # RUF066
+from json import loads, dumps  # RUF066 (twice)
+
+# Should NOT trigger - typing imports are exempt (per Google Style Guide 2.2.4.1)
```
justinchuby, line 9: Looks like `__future__` needs to be included

revmischa: I don't mind to continue working on this if it's something that could get accepted one day but the PR is closed for now

justinchuby: Yeah a bit disappointing to see

PR #21125, opened 2025-10-29
[`flake8-type-checking`] Fix `TC003` false positive with `future-annotations`

### LINKED ISSUE #21121: TC003 false-positive for dataclass.KW_ONLY when future-annotations is true
### Summary

playground: https://play.ruff.rs/10ce19b3-ce84-44e2-be98-19ece9acdf9d

---

pyproject.toml
```toml
[project]
name = "TC-TEST"
version = "0.1.0"
description = "TC-TEST"
readme = "README.md"
requires-python = ">=3.13"
dependencies = ["ruff>=0.14.2"]

[tool.ruff.lint]
select = ["TC"]
extend-safe-fixes = ["TC"]
future-annotations = true
```

```python
from dataclasses import KW_ONLY, dataclass


@dataclass
class DataClass:
    a: int
    _: KW_ONLY
    b: int
```

ruff (v0.14.2) check `TC003` for importing `KW_ONLY` in python 3.13.

<br>

Auto (unsafe) fix result:
```python
from typing import TYPE_CHECKING
from dataclasses import KW_ONLY
```

<br>

`KW_ONLY` import in `TYPE_CHECKING` block cause `TC004`.
```python

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from dataclasses import KW_ONLY


@dataclass
class DataClass:
    a: int
    _: KW_ONLY
    b: int
```

Relative to #12859

### Version

ntBre:
Summary
--

Fixes #21121 by upgrading `RuntimeEvaluated` annotations like
`dataclasses.KW_ONLY` to `RuntimeRequired`. We already had special handling for
`TypingOnly` annotations in this context but not `RuntimeEvaluated`. Combining
that with the `future-annotations` setting, which allowed ignoring the
`RuntimeEvaluated` flag, led to the reported bug where we would try to move
`KW_ONLY` into a `TYPE_CHECKING` block.

Test Plan
--

A new test based on the issue

### DIFF
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -1408,6 +1408,9 @@ impl<'a> Visitor<'a> for Checker<'a> {
                     {
                         self.visit_runtime_required_annotation(annotation);
                     }
+                    AnnotationContext::RuntimeEvaluated => {
+                        self.visit_runtime_evaluated_annotation(annotation);
+                    }
                     AnnotationContext::TypingOnly
                         if flake8_type_checking::helpers::is_dataclass_meta_annotation(
                             annotation,
@@ -1422,9 +1425,6 @@ impl<'a> Visitor<'a> for Checker<'a> {
                             // Ex) `InitVar`
                             self.visit_runtime_required_annotation(annotation);
                         }
-                    }
-                    AnnotationContext::RuntimeEvaluated => {
-                        self.visit_runtime_evaluated_annotation(annotation);
                     }
                     AnnotationContext::TypingOnly => self.visit_annotation(annotation),
                 }


ntBre:
I should have waited a second longer before taking it out of draft. I think the ecosystem hit is a false positive, taking a closer look now.

ntBre:
It looks like we have 4 pre-existing UP037 false positives on this project:
[`src/latch/ldata/_transfer/upload.py`](https://github.com/latchbio/latch/blob/main/src/latch/ldata/_transfer/upload.py):121:27: UP037 `[*]` Remove quotes from type annotation
[`src/latch/ldata/_transfer/upload.py`](https://github.com/latchbio/latch/blob/main/src/latch/ldata/_transfer/upload.py):122:33: UP037 `[*]` Remove quotes from type annotation
[`src/latch/ldata/_transfer/upload.py`](https://github.com/latchbio/latch/blob/main/src/latch/ldata/_transfer/upload.py):132:28: UP037 `[*]` Remove quotes from type annotation
[`src/latch/types/directory.py`](https://github.com/latchbio/latch/blob/main/src/latch/types/directory.py):162:36: UP037 `[*]` Remove quotes from type annotation

but this PR adds a fifth.

ntBre:
Now the ecosystem check looks good! We needed to visit the whole `RuntimeEvaluated` annotation as `RuntimeRequired` for the special `dataclass` types, unlike the existing `TypingOnly` branch, which could treat the annotation and slice separately.

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -1418,6 +1423,9 @@ impl<'a> Visitor<'a> for Checker<'a> {
                             self.visit_runtime_required_annotation(annotation);
                         }
                     }
+                    AnnotationContext::RuntimeEvaluated => {
+                        self.visit_runtime_evaluated_annotation(annotation);
+                    }
```
MichaReiser: Nit: I would move this up right below the other `AnnotationContext::RuntimeEvaluated` arm. Not only reduces it the size of the diff, it also makes it easier to "match" the different arms when reading the code because you don't need to keep track of the only partially matched arms (or, you have but only up to the next arm)

PR #21127, opened 2025-10-29
[ty] Dont provide goto definition for definitions which are not reexported in builtins

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1451

## Test Plan

Add test that previously fails that shows that imports in builtins that are not reexported are not given goto definitions.


MichaReiser (approved):
Nice, I didn't know we already have a method to test exactly this

MatthewMckee4:
Took me a while to find 😃

PR #21129, opened 2025-10-29
[ty] Add and test when constraint sets are satisfied by their typevars

dcreager:
This PR adds a new `satisfied_by_all_typevar` method, which implements one of the final steps of actually using these dang constraint sets. Constraint sets exist to help us check assignability and subtyping of types in the presence of typevars. We construct a constraint set describing the conditions under which assignability holds between the two types. Then we check whether that constraint set is satisfied for the valid specializations of the relevant typevars (which is this new method).

We also add a new `ty_extensions.ConstraintSet` method so that we can test this method's behavior in mdtests, before hooking it up to the rest of the specialization inference machinery.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,60 @@
 # Changelog
 
+## 0.14.3
+
+Released on 2025-10-30.
+
+### Preview features
+
+- Respect `--output-format` with `--watch` ([#21097](https://github.com/astral-sh/ruff/pull/21097))
+- \[`pydoclint`\] Fix false positive on explicit exception re-raising (`DOC501`, `DOC502`) ([#21011](https://github.com/astral-sh/ruff/pull/21011))
+- \[`pyflakes`\] Revert to stable behavior if imports for module lie in alternate branches for `F401` ([#20878](https://github.com/astral-sh/ruff/pull/20878))
+- \[`pylint`\] Implement `stop-iteration-return` (`PLR1708`) ([#20733](https://github.com/astral-sh/ruff/pull/20733))
+- \[`ruff`\] Add support for additional eager conversion patterns (`RUF065`) ([#20657](https://github.com/astral-sh/ruff/pull/20657))
+
+### Bug fixes
+
+- Fix finding keyword range for clause header after statement ending with semicolon ([#21067](https://github.com/astral-sh/ruff/pull/21067))
+- Fix syntax error false positive on nested alternative patterns ([#21104](https://github.com/astral-sh/ruff/pull/21104))
+- \[`ISC001`\] Fix panic when string literals are unclosed ([#21034](https://github.com/astral-sh/ruff/pull/21034))
+- \[`flake8-django`\] Apply `DJ001` to annotated fields ([#20907](https://github.com/astral-sh/ruff/pull/20907))
+- \[`flake8-pyi`\] Fix `PYI034` to not trigger on metaclasses (`PYI034`) ([#20881](https://github.com/astral-sh/ruff/pull/20881))
+- \[`flake8-type-checking`\] Fix `TC003` false positive with `future-annotations` ([#21125](https://github.com/astral-sh/ruff/pull/21125))
+- \[`pyflakes`\] Fix false positive for `__class__` in lambda expressions within class definitions (`F821`) ([#20564](https://github.com/astral-sh/ruff/pull/20564))
+- \[`pyupgrade`\] Fix false positive for `TypeVar` with default on Python \<3.13 (`UP046`,`UP047`) ([#21045](https://github.com/astral-sh/ruff/pull/21045))
+
+### Rule changes
+
+- Add missing docstring sections to the numpy list ([#20931](https://github.com/astral-sh/ruff/pull/20931))
+- \[`airflow`\] Extend `airflow.models..Param` check (`AIR311`) ([#21043](https://github.com/astral-sh/ruff/pull/21043))
+- \[`airflow`\] Warn that `airflow....DAG.create_dagrun` has been removed (`AIR301`) ([#21093](https://github.com/astral-sh/ruff/pull/21093))
+- \[`refurb`\] Preserve digit separators in `Decimal` constructor (`FURB157`) ([#20588](https://github.com/astral-sh/ruff/pull/20588))
+
+### Server
+
+- Avoid sending an unnecessary "clear diagnostics" message for clients supporting pull diagnostics ([#21105](https://github.com/astral-sh/ruff/pull/21105))
+
+### Documentation
+
+- \[`flake8-bandit`\] Fix correct example for `S308` ([#21128](https://github.com/astral-sh/ruff/pull/21128))
+
+### Other changes
+
+- Clearer error message when `line-length` goes beyond threshold ([#21072](https://github.com/astral-sh/ruff/pull/21072))
+
+### Contributors
+
+- [@danparizher](https://github.com/danparizher)
+- [@jvacek](https://github.com/jvacek)
+- [@ntBre](https://github.com/ntBre)
+- [@augustelalande](https://github.com/augustelalande)
+- [@prakhar1144](https://github.com/prakhar1144)
+- [@TaKO8Ki](https://github.com/TaKO8Ki)
+- [@dylwil3](https://github.com/dylwil3)
+- [@fatelei](https://github.com/fatelei)
+- [@ShaharNaveh](https://github.com/ShaharNaveh)
+- [@Lee-W](https://github.com/Lee-W)
+
 ## 0.14.2
 
 Released on 2025-10-23.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,7 +243,7 @@ dependencies = [
  "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -633,7 +633,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -642,7 +642,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1007,7 +1007,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.0",
 ]
 
 [[package]]
@@ -1093,7 +1093,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.0",
 ]
 
 [[package]]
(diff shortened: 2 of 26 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,158 @@
...
+
+If a typevar has an upper bound, then it must specialize to a type that is a subtype of that bound.
+For an inferable typevar, that means we need a single type that satisfies both the constraint set
+and the upper bound.
+
+```py
+def bounded[T: Base]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(T))
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(T))
+    static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(T))
```
MichaReiser: I don't think I understand why `T <= Super` satisfies `T <= Base`, given that `Super` isn't a subtype of `Base`. Or am I reading this the wrong way around? Or is the `Never` part important?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,158 @@
...
+    static_assert(ConstraintSet.range(Never, T, Base).satisfied_by_all_typevars(T))
+    static_assert(ConstraintSet.range(Never, T, Sub).satisfied_by_all_typevars(T))
+
+    # This succeeds because T can specialize to Never
```
MichaReiser: I don't think I understand the why but I trust you on this

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,158 @@
...
+If a typevar has constraints, then it must specialize to one of those specific types. (Not to a
+subtype of one of those types!) For an inferable typevar, that means we need the constraint set to
+be satisfied by any one of the constraints.
+
+```py
+def constrained[T: (Base, Unrelated)]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(T))
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(T))
+    static_assert(ConstraintSet.range(Never, T, Unrelated).satisfied_by_all_typevars(T))
+    static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(T))
```
MichaReiser: Hmm, same here. Why is `Super` satisfying the constraint, given that it isn't `Base` or `Unrelated`?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,158 @@
...
+@final
+class Unrelated: ...
+```
+
+If a typevar has no bound or constraints, then it can specialize to any type. For a non-inferable
+typevar, that means the constraint set must be satisfied for every possible type.
+
+```py
+def unbounded[T]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars())
```
MichaReiser: Gosh... it took me way too long to notice the difference to the unbounded example on line 40... We don't pass any type var here...

Should we use a different method name for this? It seems very subtle, but maybe it's obvious for someone more familiar with all this

crates/ty_vendored/ty_extensions/ty_extensions.pyi
```diff
@@ -67,6 +67,12 @@ class ConstraintSet:
         .. _subtype: https://typing.python.org/en/latest/spec/concepts.html#subtype-supertype-and-type-equivalence
         """
 
+    def satisfied_by_all_typevars(self, *inferable: Any) -> bool:
+        """
+        Returns whether this constraint set is satisfied by all of the typevars
+        that it mentions.
+        """
```
MichaReiser, line 78: I'd find it helpful to document the behavior when `inferable` is empty. Naively, I'd assume it always returns `true` or `false` but the actual behavior seems to be different?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: I can make it a keyword-only parameter, which I think will call out the difference more clearly

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: We discussed this sync — the key point is that the implication is in the other direction. (The typevar bounds/constraints need to imply the constraint set being checked.) So here, `T` must specialize to a subtype of `Base`. `Base` is a subtype of `Super`, so every valid specialization of `T` is also a subtype of `Super`. (Technically, this test is in the inferable section, so we only need _one_ valid specialization of `T` to satisfy the constraint set. In this case, _every_ valid specialization does, so you'll see that this same test also holds down in the non-inferable section.)

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: This is trying to show that `Never` is sneaky, especially for inferable typevars. Since `T` is inferable, we only need _one_ specialization to satisfy the constraint set. `Never` is a valid specialization, since `Never ≤ Base`. And `Never ≤ Unrelated`, so the constraint set is satisfied for the `T = Never` case. That's enough for an inferable typevar.

The corresponding test in the non-inferable section fails, though, since there we need _all_ valid specializations to satisfy the constraint set. And `T = Base` is a counter-example where it doesn't.

This tells me that it might help to change the structure of this file? Right now I have inferable/non-inferable as the top-level sections, and different kinds of typevar bounds/constraints as (unlabeled) subsections. Maybe instead I should have "unbounded", "bound", and "constrained" as the top-level sections, to more clearly call out how the behavior is different for inferable vs non-inferable?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: Same as above. This is in inferable position, so we only need one specialization of `T` to satisfy the constraint set. `T = Base` does, so the check passes. Down below in the non-inferable section, the check fails, because `T = Unrelated` is a valid specialization that doesn't satisfy the constraint set.

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: > This tells me that it might help to change the structure of this file?

I did this, and added more explanatory comments for each example. Hopefully that helps clarify the logic a bit.

crates/ty_vendored/ty_extensions/ty_extensions.pyi
dcreager: Done

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,220 @@
...
+specializations of the relevant typevars. This file tests that final step.
+
+## Inferable vs non-inferable typevars
+
+Typevars can appear in _inferable_ or _non-inferable_ positions.
+
+When a typevar is in an inferable position, the constraint set only needs to be satisfied for _some_
+valid specialization. The most common inferable position occurs when invoking a generic function:
+all of the function's typevars are inferable, because we want to use the argument types to infer
+which specialization is being invoked.
```
MichaReiser, line 20: This is great. Thanks for adding it

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,220 @@
...
+class Super: ...
+class Base(Super): ...
+class Sub(Base): ...
+
+@final
+class Unrelated: ...
+
+def unbounded[T]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(inferable=tuple[()]))
```
MichaReiser: I've a slight preference towards making `inferable` optional over this somewhat magical `()`

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -0,0 +1,220 @@
...
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(inferable=tuple[()]))
+
+    # (T = Unrelated) is a valid specialization, which satisfies (T ≤ Unrelated).
+    static_assert(ConstraintSet.range(Never, T, Unrelated).satisfied_by_all_typevars(inferable=tuple[T]))
+    # (T = Base) is a valid specialization, which does not satisfy (T ≤ Unrelated).
+    static_assert(not ConstraintSet.range(Never, T, Unrelated).satisfied_by_all_typevars(inferable=tuple[()]))
+
+    # (T = Base) is a valid specialization, which satisfies (T ≤ Super).
+    static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(inferable=tuple[T]))
```
MichaReiser, line 175: I'm still somewhat confused by the DSL here but I (maybe?) finally figured out how to read this?

Is my understanding correct that this results in:

```
Never <= any(Base, Unrelated) <= Super
```

which is true because `Base` satisfies this constraint.

And the next example is:

```
Never < forall(Base, Unrelated) < Super
```

which is false, because `Unrelated` doesn't satisfy this constraint.


The part I'm struggling with right now is what a real-world example of `static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(inferable=tuple[T]))` would look like?

MichaReiser (approved):
Thanks again for taking the time to explain this to me during our phone call. The reasoning in the mdtests make sense to me. Altough I'm still finding the DSL slightly confusing. But maybe that just is because I fail to map it to an example using Python

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: > I'm still somewhat confused by the DSL here but I (maybe?) finally figured out how to read this?

I would formalize it differently, but I think your notation leads to the right understanding:

> Is my understanding correct that this results in:
> 
> ```
> Never <= any(Base, Unrelated) <= Super
> ```
> 
> which is true because `Base` satisfies this constraint.

Yes

> And the next example is:
> 
> ```
> Never < forall(Base, Unrelated) < Super
> ```
> 
> which is false, because `Unrelated` doesn't satisfy this constraint.

This should use `<=` instead of `<` like the first example, but otherwise yes.

> The part I'm struggling with right now is what a real-world example of `static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(inferable=tuple[T]))` would look like?

I'm actually not sure what Python I could write that would result in this check either! We need something that wants to check that an instance of `T` is assignable to `Super`. The non-inferable case is easy:

```py
def f[T: (Base, Unrelated)](t: T):
    x: Super = t
```

But in the `T` inferable case, we would be invoking this function. And I'm not sure what Python code would lead to a `T ≤ Super` check.

The problem is that I can't limit myself to implementing these algorithms for constraint set checks that have obvious Python analogues. 😅

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: Done

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: > But in the `T` inferable case, we would be invoking this function. And I'm not sure what Python code would lead to a `T ≤ Super` check.

Ah maybe via the return type:

```py
def f[T: (Base, Unrelated)]() -> T:
    raise NotImplementedError

x: Super = f()
```

(Disregard the fact that there is no real function body that we could write that would satisfy that signature!)

PR #21130, opened 2025-10-29
[ty] Reachability and narrowing for enum methods

sharkdp:
## Summary

Adds proper type narrowing and reachability analysis for matching on non-inferable type variables bound to enums. For example:

```py
from enum import Enum

class Answer(Enum):
    NO = 0
    YES = 1

    def is_yes(self) -> bool:  # no error here!
        match self:
            case Answer.YES:
                return True
            case Answer.NO:
                return False
```

closes https://github.com/astral-sh/ty/issues/1404

## Test Plan

Added regression tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/match.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/match.md
@@ -252,3 +252,51 @@ match x:
 
 reveal_type(x)  # revealed: object
 ```
+
+## Narrowing on `Self` in `match` statements
+
+When performing narrowing on `self` inside methods on enums, we take into account that `Self` might
+refer to a subtype of the enum class, like `Literal[Answer.YES]`. This is why we do not simplify
+`Self & ~Literal[Answer.YES]` to `Literal[Answer.NO, Answer.MAYBE]`. Otherwise, we wouldn't be able
+to return `self` in the `assert_yes` method below:
+
+```py
+from enum import Enum
+from typing_extensions import Self, assert_never
+
+class Answer(Enum):
+    NO = 0
+    YES = 1
+    MAYBE = 2
+
+    def is_yes(self) -> bool:
+        reveal_type(self)  # revealed: Self@is_yes
+
+        match self:
+            case Answer.YES:
+                reveal_type(self)  # revealed: Self@is_yes
+                return True
+            case Answer.NO | Answer.MAYBE:
+                reveal_type(self)  # revealed: Self@is_yes & ~Literal[Answer.YES]
+                return False
+            case _:
+                assert_never(self)  # no error
+
+    def assert_yes(self) -> Self:
+        reveal_type(self)  # revealed: Self@assert_yes
+
+        match self:
+            case Answer.YES:
+                reveal_type(self)  # revealed: Self@assert_yes
+                return self
+            case _:
+                reveal_type(self)  # revealed: Self@assert_yes & ~Literal[Answer.YES]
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/narrow/match.md
```diff
@@ -252,3 +252,53 @@ match x:
...
+
+Answer.YES.is_yes()
+
+try:
+    reveal_type(Answer.MAYBE.assert_yes())  # revealed: Literal[Answer.MAYBE]
+except ValueError:
+    pass
+```
+
+We do
```
AlexWaygood: We do indeed

sharkdp: Glad you like it, I was planning on leaving this in there.

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -780,6 +780,37 @@ impl<'db> IntersectionBuilder<'db> {
...
+                    enum_metadata(self.db, enum_class).expect("Class of enum literal is an enum");
+
+                let enum_members_in_negative_part = self
+                    .intersections
+                    .iter()
+                    .flat_map(|intersection| &intersection.negative)
+                    .filter_map(|ty| ty.as_enum_literal())
+                    .filter(|lit| lit.enum_class(self.db) == enum_class)
+                    .map(|lit| lit.name(self.db).clone())
+                    .chain(std::iter::once(enum_literal.name(self.db).clone()))
```
AlexWaygood: nit: it seems to compile fine if we just use references here?

```suggestion
                    .map(|lit| lit.name(self.db))
                    .chain(std::iter::once(enum_literal.name(self.db))
```

crates/ty_python_semantic/src/types/builder.rs
```diff
@@ -780,6 +780,37 @@ impl<'db> IntersectionBuilder<'db> {
...
+                    .collect::<FxOrderSet<_>>();
```
AlexWaygood: I think this can be an `FxHashSet`, which is a slighlty simpler datastructure and possibly slightly faster? I don't think deterministic ordering is necessary here

```suggestion
                    .collect::<FxHashSet<_>>();
```

crates/ty_python_semantic/src/semantic_index/reachability_constraints.rs
```diff
@@ -802,10 +802,27 @@ impl ReachabilityConstraints {
...
+            .add_negative(type_excluded_by_previous_patterns(db, predicate));
+
+        let narrowed_subject_ty = narrowed_subject.clone().build();
+
+        // Consider a case where we match on a subject type of `Self` with an upper bound of `Answer`, where
+        // `Answer` is a {YES, NO} enum. After a previous pattern matching on `NO`, the narrowed subject
+        // type is `Self & ~Literal[NO]`. This type is *not* equivalent to `Literal[YES]`, because `Self`
+        // could also specialize to `Literal[NO]` or `Never`, making the intersection empty. However, if the
+        // current pattern matches on `YES`, the *next* narrowed subject type will be `Self & ~Literal[NO] &
+        // ~Literal[YES]`, which *is* equivalent to `Never`. This means that subsequent patterns can never
```
AlexWaygood:
```suggestion
        // ~Literal[YES]`, which *is* always equivalent to `Never`. This means that subsequent patterns can never
```

PR #21131, opened 2025-10-29
[`flake8-bugbear`] Ignore non-NFKC attribute names in B009 and B010 (`B009`, `B010`)

### LINKED ISSUE #21126: B009 and B010 should ignore non-NFKC attribute names
### Summary

[`get-attr-with-constant` (B009)](https://docs.astral.sh/ruff/rules/get-attr-with-constant/) and [`set-attr-with-constant` (B010)](https://docs.astral.sh/ruff/rules/set-attr-with-constant/) should ignore non-NFKC attribute names because the usual attribute syntax would change the program’s behavior. [Example](https://play.ruff.rs/8beed5a5-02f2-40df-9533-71d04176ac50):
```console
$ cat >b009_b010.py <<'# EOF'
from types import SimpleNamespace
ns = SimpleNamespace(s=0)
setattr(ns, "ſ", 1)
print(ns.s, getattr(ns, "ſ"))
# EOF

$ python b009_b010.py
0 1

$ ruff check --isolated b009_b010.py --select B009,B010 --fix

$ cat b009_b010.py
from types import SimpleNamespace
ns = SimpleNamespace(s=0)
ns.ſ = 1
print(ns.s, ns.ſ)

$ python b009_b010.py
1 1
```

### Version

danparizher:
## Summary

This PR updates the `B009` (`get-attr-with-constant`) and `B010` (`set-attr-with-constant`) rules to ignore non-NFKC attribute names, preventing refactoring suggestions that could change program behavior.

Fixes #21126

## Problem Analysis

Python normalizes identifiers using NFKC (Normalization Form KC) normalization. When using attribute access syntax (e.g., `obj.attr`), Python automatically normalizes the identifier. However, when using `getattr` or `setattr` with a string literal, the identifier is not normalized.

The issue occurs when an attribute name contains characters that normalize differently under NFKC. For example:
- The long s character `"ſ"` normalizes to `"s"` in NFKC
- Using `setattr(ns, "ſ", 1)` creates a distinct attribute from `ns.s`
- If Ruff suggests replacing `setattr(ns, "ſ", 1)` with `ns.ſ = 1`, Python would normalize `ns.ſ` to `ns.s`, which changes the program's behavior

The bug was that Ruff's B009 and B010 rules didn't account for this normalization difference, potentially suggesting unsafe refactorings.

## Approach

The fix adds a check in both `B009` and `B010` rule implementations to detect non-NFKC attribute names. Before suggesting a refactoring, the code now:

1. Normalizes the attribute name using NFKC
2. Compares the normalized form with the original
3. If they differ, the rule skips the check (returns early) to avoid suggesting an unsafe refactoring

This ensures that:
- Rules only suggest refactorings when it's safe (i.e., when NFKC normalization wouldn't change the attribute name)
- The fix is minimal and doesn't affect other rule functionality
- Edge cases are handled correctly (non-NFKC names are silently ignored)

Test cases were added to verify that non-NFKC attribute names (like `"ſ"`) are correctly ignored by both rules.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_bugbear/rules/getattr_with_constant.rs
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/rules/getattr_with_constant.rs
@@ -30,6 +30,21 @@ use crate::{AlwaysFixableViolation, Edit, Fix};
 /// obj.foo
 /// ```
 ///
+/// ## Fix safety
+/// The fix is marked as unsafe for attribute names that are not in NFKC (Normalization Form KC)
+/// normalization. Python normalizes identifiers using NFKC when using attribute access syntax
+/// (e.g., `obj.attr`), but does not normalize string arguments passed to `getattr`. Rewriting
+/// `getattr(obj, "ſ")` to `obj.ſ` would be interpreted as `obj.s` at runtime, changing behavior.
+///
+/// For example, the long s character `"ſ"` normalizes to `"s"` under NFKC, so:
+/// ```python
+/// # This accesses an attribute with the exact name "ſ" (if it exists)
+/// value = getattr(obj, "ſ")
+///
+/// # But this would normalize to "s" and access a different attribute
+/// obj.ſ  # This is interpreted as obj.s, not obj.ſ
+/// ```
+///
 /// ## References
 /// - [Python documentation: `getattr`](https://docs.python.org/3/library/functions.html#getattr)
 #[derive(ViolationMetadata)]
--- a/crates/ruff_linter/src/rules/flake8_bugbear/rules/setattr_with_constant.rs
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/rules/setattr_with_constant.rs
@@ -29,6 +29,23 @@ use crate::{AlwaysFixableViolation, Edit, Fix};
 /// obj.foo = 42
 /// ```
 ///
+/// ## Fix safety
+/// The fix is marked as unsafe for attribute names that are not in NFKC (Normalization Form KC)
+/// normalization. Python normalizes identifiers using NFKC when using attribute access syntax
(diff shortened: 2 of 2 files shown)


crates/ruff_linter/src/rules/flake8_bugbear/rules/getattr_with_constant.rs
```diff
@@ -65,6 +66,13 @@ pub(crate) fn getattr_with_constant(checker: &Checker, expr: &Expr, func: &Expr,
     if is_mangled_private(value.to_str()) {
         return;
     }
+    // Ignore non-NFKC attribute names. Python normalizes identifiers using NFKC, so using
+    // attribute syntax (e.g., `obj.attr`) would normalize the name and potentially change
+    // program behavior.
+    let attr_name = value.to_str();
+    if attr_name.nfkc().collect::<String>() != attr_name {
+        return;
+    }
```
MichaReiser: Let's move this after line 78 as this comparison is more expensive than testing if it is the builtin getattr

crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B009_B010.py
```diff
@@ -70,3 +70,8 @@
 
 # Regression test for: https://github.com/astral-sh/ruff/issues/18353
 setattr(foo, "__debug__", 0)
+
+# Regression test for: https://github.com/astral-sh/ruff/issues/21126
+# Non-NFKC attribute names should be ignored (e.g., "ſ" normalizes to "s")
```
MichaReiser: I suggest including a short explanation **why** they should be ignored

amyreese:
+1 to just marking it unsafe

PR #21138, opened 2025-10-30
[ty] Use `range` instead of custom `IntIterable`

sharkdp:
## Summary

We previously didn't understand `range` and wrote these custom `IntIterable`/`IntIterator` classes for tests. We can now remove them and make the tests shorter in some places.


AlexWaygood (approved):
🚀

PR #21139, opened 2025-10-30
[ty] Understand legacy and PEP 695 `ParamSpec`

dhruvmanila:
## Summary

This PR adds support for understanding the legacy definition and PEP 695 definition for `ParamSpec`.

This is still very initial and doesn't really implement any of the semantics.

Part of https://github.com/astral-sh/ty/issues/157

## Test Plan

Add mdtest cases.

## Ecosystem analysis

Most of the diagnostics in `starlette` are due to the fact that ty now understands `ParamSpec` is not a `Todo` type, so the assignability check fails. The code looks something like:

```py
class _MiddlewareFactory(Protocol[P]):
    def __call__(self, app: ASGIApp, /, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover

class Middleware:
    def __init__(
        self,
        cls: _MiddlewareFactory[P],
        *args: P.args,
        **kwargs: P.kwargs,
    ) -> None:
        self.cls = cls
        self.args = args
        self.kwargs = kwargs

# ty complains that `ServerErrorMiddleware` is not assignable to `_MiddlewareFactory[P]`
Middleware(ServerErrorMiddleware, handler=error_handler, debug=debug)
```

There are multiple diagnostics where there's an attribute access on the `Wrapped` object of `functools` which Pyright also raises:
```py
from functools import wraps

def my_decorator(f):
    @wraps(f)
    def wrapper(*args, **kwds):
        return f(*args, **kwds)

	# Pyright: Cannot access attribute "__signature__" for class "_Wrapped[..., Unknown, ..., Unknown]"
      Attribute "__signature__" is unknown [reportAttributeAccessIssue]
	# ty: Object of type `_Wrapped[Unknown, Unknown, Unknown, Unknown]` has no attribute `__signature__` [unresolved-attribute]
    wrapper.__signature__
    return wrapper
```

There are additional diagnostics that is due to the assignability checks failing because ty now infers the `ParamSpec` instead of using the `Todo` type which would always succeed. This results in a few `no-matching-overload` diagnostics because the assignability checks fail.

There are a few diagnostics related to https://github.com/astral-sh/ty/issues/491 where there's a variable which is either a bound method or a variable that's annotated with `Callable` that doesn't contain the instance as the first parameter.

Another set of (valid) diagnostics are where the code hasn't provided all the type variables. ty is now raising diagnostics for these because we include `ParamSpec` type variable in the signature. For example, `staticmethod[Any]` which contains two type variables.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/paramspec.md
+++ b/crates/ty_python_semantic/resources/mdtest/paramspec.md
@@ -118,7 +118,8 @@ def foo1[**P]() -> None:
 def foo2[**P = ...]() -> None:
     reveal_type(P)  # revealed: typing.ParamSpec
 
-def foo3[**P = [int, str]]() -> None: ...
+def foo3[**P = [int, str]]() -> None:
+    reveal_type(P)  # revealed: typing.ParamSpec
 
 def foo4[**P, **Q = P]():
     reveal_type(P)  # revealed: typing.ParamSpec


dhruvmanila:
> ```diff
> -callables_protocol.py:179:62: error[invalid-return-type] Function always implicitly returns `None`, which is not assignable to return type `R@__call__`
> +callables_protocol.py:186:5: error[invalid-assignment] Object of type `Literal["str"]` is not assignable to attribute `other_attribute` of type `int`
> +callables_protocol.py:187:5: error[unresolved-attribute] Unresolved attribute `xxx` on type `Proto9[P@decorator1, R@decorator1]`.
> +callables_protocol.py:197:7: error[unresolved-attribute] Object of type `Proto9[Unknown, Unknown]` has no attribute `other_attribute2`
> ```

These are correct!

> ```diff
> +generics_defaults.py:80:32: error[too-many-positional-arguments] Too many positional arguments to class `Class_ParamSpec`: expected 1, got 2
> ```

This will require adding special support for allowing both `Class_ParamSpec[int, str]` and `Class_ParamSpec[[int, str]]` when the class is generic over a single type variable which is a `ParamSpec`.

> ```diff
> +generics_paramspec_basic.py:23:20: error[invalid-return-type] Function always implicitly returns `None`, which is not assignable to return type `P@func1`
> ```

Still need to enforce rules about where a `ParamSpec` can be used.

dhruvmanila:
I think this is a good checkpoint to get initial feedback. Follow-up work would be to:
1. Update the `Parameters` structure to accommodate `ParamSpec`
2. Representing `ParamSpecArgs`, `ParamSpecKwargs`
3. Validating the usage of `ParamSpec` e.g., it cannot be used as annotation
4. Validating the usage of `P.args` and `P.kwargs`

For (1) and (2), I think I have a good idea on how to represent them based on the following:
> A function declared as `def inner(*args: P.args, **kwargs: P.kwargs) -> X` has type `Callable[P, X]`.

And,
> The `P.args` attribute of a `ParamSpec` is an instance of `ParamSpecArgs`, and `P.kwargs` is an instance of `ParamSpecKwargs`. They are intended for runtime introspection and have no special meaning to static type checkers.

I'm thinking to create `Parameters` as `*args: P.args, **kwargs: P.kwargs` where `P.args` and `P.kwargs` would be `KnownInstanceType::ParamSpecArgs` and `KnownInstanceType::ParamSpecKwargs` respectively. Both of these variants would have a link back to the original `ParamSpec`.

AlexWaygood:
> Most of the diagnostics in `starlette` are due to the fact that ty now understands `ParamSpec` is not a `Todo` type, so the assignability check fails

There might be a short-term hack we could add to `Type::has_relation_to_impl` so that a generic type is considered to be specialized with `Unknown` if one of the typevarlikes in its generic context is a `ParamSpec`? That might reduce the false positives for users while we're still working on the full semantics for `ParamSpec`.

There aren't _that_ many new diagnostics, though, so this also might be fine for now considering that we're planning on addressing these new diagnostics shortly. Still, it's nice to avoid adding false positives for existing users where possible.

crates/ty_python_semantic/resources/mdtest/annotations/callable.md
```diff
@@ -307,8 +307,9 @@ Using a `ParamSpec` in a `Callable` annotation:
 from typing_extensions import Callable
 
 def _[**P1](c: Callable[P1, int]):
-    reveal_type(P1.args)  # revealed: @Todo(ParamSpec)
-    reveal_type(P1.kwargs)  # revealed: @Todo(ParamSpec)
+    # TODO: Should reveal `ParamSpecArgs` and `ParamSpecKwargs`
+    reveal_type(P1.args)  # revealed: Unknown
+    reveal_type(P1.kwargs)  # revealed: Unknown
```
AlexWaygood: _Ideally_ I feel like these would still be `@Todo` types, so that it's clear to users that we're aware of this limitation and that this isn't our intended final behaviour for this. But it's also fine to leave this as-is if it would be a lot of code to make these still be `@Todo` types.

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_forms.md
```diff
@@ -21,8 +21,9 @@ def f(*args: Unpack[Ts]) -> tuple[Unpack[Ts]]:
 
 def g() -> TypeGuard[int]: ...
 def i(callback: Callable[Concatenate[int, P], R_co], *args: P.args, **kwargs: P.kwargs) -> R_co:
-    reveal_type(args)  # revealed: tuple[@Todo(Support for `typing.ParamSpec`), ...]
-    reveal_type(kwargs)  # revealed: dict[str, @Todo(Support for `typing.ParamSpec`)]
+    # TODO: Should reveal a type representing `P.args` and `P.kwargs`
+    reveal_type(args)  # revealed: tuple[Unknown, ...]
+    reveal_type(kwargs)  # revealed: dict[str, Unknown]
```
AlexWaygood: same here

crates/ty_python_semantic/resources/mdtest/paramspec.md
```diff
@@ -0,0 +1,154 @@
...
+T = ParamSpec("T")
+# error: [invalid-paramspec]
+P1: ParamSpec = ParamSpec("P1")
+
+# error: [invalid-paramspec]
+tuple_with_typevar = ("foo", ParamSpec("P2"))
+reveal_type(tuple_with_typevar[1])  # revealed: ParamSpec
+```
+
+### `TypeVar` parameter must match variable name
```
AlexWaygood:
```suggestion
### `ParamSpec` parameter must match variable name
```

crates/ty_python_semantic/resources/mdtest/paramspec.md
```diff
@@ -0,0 +1,154 @@
...
+#### Valid
+
+```py
+def foo1[**P]() -> None:
+    reveal_type(P)  # revealed: typing.ParamSpec
+
+def foo2[**P = ...]() -> None:
+    reveal_type(P)  # revealed: typing.ParamSpec
+
+def foo2[**P = [int, str]]() -> None: ...
```
AlexWaygood: is it worth also explicitly testing the PEP-695 syntax where one `ParamSpec` has another `ParamSpec` as a default?

```suggestion
def foo1[**P]() -> None:
    reveal_type(P)  # revealed: typing.ParamSpec

def foo2[**P = ...]() -> None:
    reveal_type(P)  # revealed: typing.ParamSpec

def foo2[**P = [int, str]]() -> None: ...

def foo3[**P, **Q = P]():
    reveal_type(P)  # revealed: typing.ParamSpec
    reveal_type(Q)  # revealed: typing.ParamSpec
```

AlexWaygood (approved):
nice

crates/ty_python_semantic/resources/mdtest/annotations/callable.md
dhruvmanila: I believe this shouldn't be too hard, let me try...

dhruvmanila: Yeah, this wasn't hard, I've updated these to use `@Todo` types.

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_forms.md
dhruvmanila: And, this as well. Thanks for the suggestion!

crates/ty_python_semantic/resources/mdtest/paramspec.md
dhruvmanila: Yeah, I agree.

PR #21143, opened 2025-10-30
[ty] Use declared attribute types as type context

ibraheemdev:
## Summary

For example:
```py
class X:
    x: list[Literal[1]]

def _(x: X):
    x.x = [1]
```

Resolves https://github.com/astral-sh/ty/issues/1375.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -4207,7 +4207,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         }
     }
 
-    #[allow(clippy::type_complexity)]
+    #[expect(clippy::type_complexity)]
     fn infer_target_impl(
         &mut self,
         target: &ast::Expr,
@@ -10299,7 +10299,7 @@ enum MultiInferenceState {
 }
 
 impl MultiInferenceState {
-    fn is_panic(self) -> bool {
+    const fn is_panic(self) -> bool {
         matches!(self, MultiInferenceState::Panic)
     }
 }


crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -200,6 +200,41 @@ def f() -> list[Literal[1]]:
...
+    class Meta(type):
+        if flag:
+            x: DataDescriptor = DataDescriptor()
+
+    class C(metaclass=Meta):
+        x: list[int | None]
+
+    def _(c: C):
+        c.x = lst(1)
+        C.x = lst(1)
```
ibraheemdev, line 235: Hover tests would be nice here...

crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -200,6 +200,41 @@ def f() -> list[Literal[1]]:
...
+    def __set__(self, instance: object, value: list[Literal[1]]) -> None:
+        pass
+
+def lst[T](x: T) -> list[T]:
+    return [x]
+
+def _(flag: bool):
+    class Meta(type):
+        if flag:
+            x: DataDescriptor = DataDescriptor()
```
sharkdp: Is it necessary to partially shadow `x` on `C` here, or could this just be an attribute with a different name? (and a separate assignment test below)?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3607,8 +3651,17 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
         match object_ty {
             Type::Union(union) => {
+                // TODO: We could perform multi-inference here with each element of the union as type context.
```
AlexWaygood, line 3654: should we open an issue for this, with an example of code that this would help us infer better?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4098,22 +4207,35 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         }
     }
 
+    #[allow(clippy::type_complexity)]
```
```suggestion
    #[expect(clippy::type_complexity)]
```

or we could also just give in to Clippy here and add a type alias ;)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -10155,16 +10291,16 @@ enum MultiInferenceState {
...
 
+    /// Overwrite the previously inferred value.
+    Overwrite,
+
     /// Store the intersection of all types inferred for the expression.
     Intersect,
 }
 
 impl MultiInferenceState {
     fn is_panic(self) -> bool {
```
```suggestion
    const fn is_panic(self) -> bool {
```

PR #21145, opened 2025-10-30
Fix bug where classmethods and staticmethods were considered subtypes of `types.FunctionType`

AlexWaygood:
Fixes https://github.com/astral-sh/ty/issues/1452


AlexWaygood:
the minimal repro of the corpus panic is

```py
from typing_extensions import ParamSpec

ParamSpec("P")
```

...

AlexWaygood:
Going to abandon this for now: it doesn't seem high-priority; I've already spent too much time on it; and this isn't really an attractive fix. A better fix would probably be to add separate `Type::ClassMethod` and `Type::StaticMethod` variants.

PR #21146, opened 2025-10-30
[ty] Don't provide completions when in class or function definition

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1457

We now check if we are in a situation where we are in the name of class or function definition like `def f` or `class f`. We check if the second last token is `def` or `class`.

## Test Plan

Added test shown in that issue.

Also add test for situation where we have not started typing the name like `def ` and `class `. My changes do not change any behavior there, but i think it is good to show we still dont provide completions there

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -859,12 +859,17 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
         .checked_sub(2)
         .and_then(|i| tokens.get(i))
         .is_some_and(|t| {
-            matches!(
+            if matches!(
                 t.kind(),
                 TokenKind::Def | TokenKind::Class | TokenKind::Type
-            ) || file.read_to_string(db).is_ok_and(|source| {
-                source[t.range().start().to_usize()..t.range().end().to_usize()] == *"type"
-            })
+            ) {
+                true
+            } else if t.kind() == TokenKind::Name {
+                let source = source_text(db, file);
+                &source[t.range()] == "type"
+            } else {
+                false
+            }
         })
 }
 


crates/ty_ide/src/completion.rs
```diff
@@ -212,7 +212,11 @@ pub fn completion<'db>(
     offset: TextSize,
 ) -> Vec<Completion<'db>> {
     let parsed = parsed_module(db, file).load(db);
-    if is_in_comment(&parsed, offset) || is_in_string(&parsed, offset) {
+
+    if is_in_comment(&parsed, offset)
+        || is_in_string(&parsed, offset)
+        || is_in_definition_place(&parsed, offset)
```
MichaReiser: Should we pull out the `tokens.before()`? It seems every one of those functions now searches the tokens before `offset`.

crates/ty_ide/src/completion.rs
```diff
@@ -849,6 +853,19 @@ fn is_in_string(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
...
+/// If the tokens end with `class` or `def`, we return false.
+/// This is fine because we don't provide completions anyway.
+fn is_in_definition_place(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
+    let tokens = tokens_start_before(parsed.tokens(), offset);
+
+    tokens
+        .len()
+        .checked_sub(2)
+        .and_then(|i| tokens.get(i))
+        .is_some_and(|t| matches!(t.kind(), TokenKind::Def | TokenKind::Class))
```
MichaReiser: I forgot to mention this in my PR but we should probably do this too in `type <CURSOR>.

MichaReiser (approved):
Thank you

crates/ty_ide/src/completion.rs
MatthewMckee4: Good idea

crates/ty_ide/src/completion.rs
MatthewMckee4: cool, will cover that, though i think there is a case where we can't catch it:

image

`type` here is just `Name` in the ast.

What we can catch is this:

```py
type f<CURSOR> = int
```

MatthewMckee4: (unless i look directly at the string, but I'm not sure if this is right or not.

MichaReiser: Oh right, because it is a contextual keyword. I think inspecting the source would be fine here because I don't see any case where `type <name>` is valid.

crates/ty_ide/src/completion.rs
```diff
@@ -849,6 +850,24 @@ fn is_in_string(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
...
+    tokens
+        .len()
+        .checked_sub(2)
+        .and_then(|i| tokens.get(i))
+        .is_some_and(|t| {
+            matches!(
+                t.kind(),
+                TokenKind::Def | TokenKind::Class | TokenKind::Type
+            ) || file.read_to_string(db).is_ok_and(|source| {
+                source[t.range().start().to_usize()..t.range().end().to_usize()] == *"type"
```
MatthewMckee4: This does not seem like the right approach, but i can't seem to find any other apis to use to do this.

MichaReiser: You want something like:

```rust
if matches!(t.kind(), TokenKind::Def | TokenKind::Class | TokenKind::Type) {
	true
} else if t.kind() == TokenKind::Name {
	let source = source_text(db, file);
	
 	&source[t.range()] == "type"
} else {
	false 
}
```

MatthewMckee4: cool, thanks

crates/ty_ide/src/completion.rs
```diff
@@ -839,8 +841,7 @@ fn is_in_comment(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
 ///
 /// Note that this will return `false` when positioned within an
 /// interpolation block in an f-string or a t-string.
```
BurntSushi, line 843: The doc comments for this function (and the one above) should be updated.

crates/ty_ide/src/completion.rs
```diff
@@ -4058,6 +4082,83 @@ def f[T](x: T):
...
+    #[test]
+    fn no_completions_in_type_def_empty_name() {
+        let builder = completion_test_builder(
+            "\
+type <CURSOR>
+        ",
+        );
+
+        builder.auto_import().build().not_contains("fabs");
+    }
```
BurntSushi, line 4160: I don't think these tests need to be enabling `auto_import()`. That's a bit of a big hammer. You can instead add a local item in the source file and test completions on that.

BurntSushi (commented):
I realize this PR was already merged, but I wanted to leave feedback anyway.

I can address these in a follow-up PR if you don't want though. :-)

crates/ty_ide/src/completion.rs
```diff
@@ -849,6 +850,29 @@ fn is_in_string(parsed: &ParsedModuleRef, offset: TextSize) -> bool {
     })
 }
 
+/// If the tokens end with `class f` or `def f` we return true.
+/// If the tokens end with `class` or `def`, we return false.
+/// This is fine because we don't provide completions anyway.
```
BurntSushi, line 855: I think it would be better to not repeat the implementation in the doc comment where possible. e.g., Your doc comment doesn't include `type`. Instead, something like, "Returns true when the tokens indicate that the definition of a new name is being introduced at the end."

MatthewMckee4:
Thanks for the feedback, ill address this in a PR

PR #21150, opened 2025-10-30
[ty] Use the top materialization of classes for narrowing in class-patterns for `match` statements

AlexWaygood:
Fixes https://github.com/astral-sh/ty/issues/1458

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/match.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/match.md
@@ -118,6 +118,32 @@ def f(x: Covariant[int]):
             assert_never(x)
 ```
 
+## Class patterns where the class pattern does not resolve to a class
+
+In general this does not allow for narrowing, but we make an exception for `Any`. This is to support
+[real ecosystem code](https://github.com/jax-ml/jax/blob/d2ce04b6c3d03ae18b145965b8b8b92e09e8009c/jax/_src/pallas/mosaic_gpu/lowering.py#L3372-L3387)
+found in `jax`.
+
+```py
+from typing import Any
+
+X = Any
+
+def f(obj: object):
+    match obj:
+        case int():
+            reveal_type(obj)  # revealed: int
+        case X():
+            reveal_type(obj)  # revealed: Any & ~int
+
+def g(obj: object, Y: Any):
+    match obj:
+        case int():
+            reveal_type(obj)  # revealed: int
+        case Y():
+            reveal_type(obj)  # revealed: Any & ~int
+```
+
 ## Value patterns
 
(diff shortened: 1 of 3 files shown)


carljm (approved):

MeGaGiGaGon:
That's fun in the primer seeing a bunch of Expression Result diagnostics going away, since my issue came from also running into this making my own Result type :)

crates/ty_python_semantic/resources/mdtest/narrow/match.md
```diff
@@ -69,6 +69,81 @@ match x:
...
+            reveal_type(obj)  # revealed: int
+        case X():
+            reveal_type(obj)  # revealed: Any & ~int
+
+def g(obj: object, Y: Any):
+    match obj:
+        case int():
+            reveal_type(obj)  # revealed: int
+        case Y():
+            reveal_type(obj)  # revealed: Any & ~int
```
abhijeetbodas2001, line 144: Alex, I did not understand this test case. The first test with `X()` gives a runtime error if the `Any()` match arm is checked. And in the real-code example you linked, I don't think any match arm is checking for `Any()`?
```py
[nav] In [1]: from typing import Any
         ...:
         ...: X = Any
         ...:
         ...: def f(obj: object):
         ...:     match obj:
         ...:         case int():
         ...:             ...
         ...:         case X():
         ...:             ...

[ins] In [2]: f("")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 f("")

Cell In[1], line 9, in f(obj)
      7 case int():
      8     ...
----> 9 case X():
     10     ...

File ~/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/typing.py:515, in _AnyMeta.__instancecheck__(self, obj)
    513 def __instancecheck__(self, obj):
    514     if self is Any:
--> 515         raise TypeError("typing.Any cannot be used with isinstance()")
    516     return super().__instancecheck__(obj)

TypeError: typing.Any cannot be used with isinstance()
```

AlexWaygood: I think I did this because it fixed a bunch of unexpected ecosystem diagnostics on one project that had this pattern, which showed up in an early version of this PR. But it's possible that I reacted too rashly to the unexpected diagnostics; maybe we just shouldn't support that pattern. I'm open to changing this!

abhijeetbodas2001: I opened https://github.com/astral-sh/ruff/pull/23011 to remove this special case. Can you take a look?

PR #21151, opened 2025-10-30
[ty] Fix playground crash when file name includes path separator

sharkdp:


playground/ty/src/Editor/SecondaryPanel.tsx
```diff
@@ -103,6 +103,8 @@ function Content({
   }
 }
 
+const PYODIDE_HOME = "/home/pyodide/";
```
MichaReiser: To be less dependent on pyodide:

We could set the home path when loading pyodide, see https://pyodide.org/en/stable/usage/api/js-api.html#exports.PyodideConfig.env

sharkdp: Like this?

MichaReiser: yes

PR #21154, opened 2025-10-30
[`pydocstyle`] Refactor `D417` to use the `pydoclint` docstring processing

augustelalande:
## Summary

This PR refactors `D417` to use the parameters detected in the `pydoclint` scope, the goal is to have a `D417` implementation which is equivalent to the previous one, but uses the `pydoclint` docstring processing. Later on `D417` can be renamed to `DOC101` and `D417` deprecated. This will make deploying `DOC101` easier.

## Test Plan

The refactor is tested on the old fixtures, and the new snap was manually diffed. The following changes were found between the old and new implementation.

In the following the new implementation reports an error which the old didn't. Notice that the code mixes google and numpy style elements which makes the docstring invalid for both.

```python
def f(x):
    """Do something with valid description.

    Args:
    ----
        x: the value

    Returns:
    -------
        the value
    """
    return x
```

In the following, the old implementation reports an error but the new one does not. This is because the old implementation enforces argument order in the docstring which the new one does not. Pydoclint has a seperate rule for checking documentation order which can be implemented later.
```python
    @classmethod
    @expect("D417: Missing argument descriptions in the docstring "
            "(argument(s) test, y, z are missing descriptions in "
            "'test_missing_args_class_method' docstring)", arg_count=4)
    def test_missing_args_class_method(cls, test, x, y, z=3):  # noqa: D213, D407
        """Test a valid args section.

        Parameters
        ----------
        z
        x
            Another parameter. The parameters y, test below are
            missing descriptions. The parameter z above is also missing
            a description.
        y
        test

        """
```

### DIFF
--- a/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
+++ b/crates/ruff_linter/src/rules/pydoclint/rules/check_docstring.rs
@@ -528,7 +528,7 @@ impl<'a> ParametersSection<'a> {
         }
     }
 
-    fn extends_from_section(&mut self, section: &SectionContext<'a>, style: Option<SectionStyle>) {
+    fn extend_from_section(&mut self, section: &SectionContext<'a>, style: Option<SectionStyle>) {
         let mut new_entries = parse_parameters(
             section.following_lines_str(),
             section.following_range().start(),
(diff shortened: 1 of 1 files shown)


augustelalande:
The ecosystem changes are minimal, with the new implementation removing some old false positives. It introduces some new positives, caused by improper formatting which are debatably true positives in my opinion.

ntBre:
Thanks for the merge and taking a look at the ecosystem results. This is in my inbox to have a look!

PR #21155, opened 2025-10-31
[ty] Remove mentions of VS Code from server logs

MatthewMckee4:
## Summary

Some of the logs i received in zed mentioned "VS Code", and one log displayed site packages paths like: `SitePackagesPaths({"..", ".."})`, which doesn't look great.

### DIFF
--- a/crates/ty_python_semantic/src/program.rs
+++ b/crates/ty_python_semantic/src/program.rs
@@ -113,7 +113,10 @@ pub enum PythonVersionSource {
     /// long argument (`--extra-paths`) or `--config key=value`.
     Cli,
 
-    /// The value comes from the user's editor (the selected interpreter).
+    /// The value comes from the user's editor,
+    /// while it's left open if specified as a setting
+    /// or if the value was auto-discovered by the editor
+    /// (e.g., the Python environment)
     Editor,
 
     /// We fell back to a default value because the value was not specified via the CLI or a config file.


crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -111,6 +111,12 @@ impl SitePackagesPaths {
     }
 }
 
+impl fmt::Debug for SitePackagesPaths {
```
MichaReiser: I think we should implement `Display` instead of `Debug` if we care about how this is formatted. It more clearly communicates that the formatting is user-facing.

MichaReiser (commented):
Thank you. Yeah, Zed started using my "VS code only" feature 😆 

I think we should also change `ValueSource::VSCodeExtension` to `ValueSource::Editor` or similar.

MatthewMckee4:
We also have `PythonVersionSource::PythonVSCodeExtension` and `SysPrefixPathOrigin::PythonVSCodeExtension`. Should these be updated too?

MichaReiser:
Oh yes, those should be updated too

MatthewMckee4:
does `ValueSource::LSPConfiguration` sound better?

To me `LSPConfiguration` seems like a generalization of `Editor`, but I'm not sure what is better, its not like anyone is using the lsp not through an editor, so perhaps editor is better. I think all 3 of these should match too.

crates/ty_project/src/metadata/value.rs
```diff
@@ -28,16 +28,16 @@ pub enum ValueSource {
     /// long argument (`--extra-paths`) or `--config key=value`.
     Cli,
 
-    /// The value comes from an LSP client configuration.
-    PythonVSCodeExtension,
+    /// The value comes from the user's editor settings.
+    Editor,
```
MichaReiser:
```suggestion
    /// The value comes from the user's editor,
    /// while it's left open if specified as a setting
    /// or if the value was auto-discovered by the editor
    /// (e.g., the Python environment)
    Editor,
```

crates/ty_python_semantic/src/diagnostic.rs
```diff
@@ -88,10 +88,10 @@ pub fn add_inferred_python_version_hint_to_diagnostic(
                 or in a configuration file",
             );
         }
-        crate::PythonVersionSource::PythonVSCodeExtension => {
+        crate::PythonVersionSource::Editor => {
             diagnostic.info(format_args!(
                 "Python {version} was assumed when {action} \
-                because it's the version of the selected Python interpreter in the VS Code Python extension",
+                because it's the version of the selected Python interpreter in the user's editor",
```
MichaReiser:
```suggestion
                because it's the version of the selected Python interpreter in your editor",
```

crates/ty_python_semantic/src/program.rs
```diff
@@ -113,8 +113,8 @@ pub enum PythonVersionSource {
     /// long argument (`--extra-paths`) or `--config key=value`.
     Cli,
 
-    /// The value comes from the Python VS Code extension (the selected interpreter).
-    PythonVSCodeExtension,
+    /// The value comes from the user's editor (the selected interpreter).
```
MichaReiser:
```suggestion
    /// The value comes from the user's editor,
    /// while it's left open if specified as a setting
    /// or if the value was auto-discovered by the editor
    /// (e.g., the Python environment)
```

crates/ty_python_semantic/src/site_packages.rs
```diff
@@ -1622,9 +1626,7 @@ impl std::fmt::Display for SysPrefixPathOrigin {
             Self::CondaPrefixVar => f.write_str("`CONDA_PREFIX` environment variable"),
             Self::DerivedFromPyvenvCfg => f.write_str("derived `sys.prefix` path"),
             Self::LocalVenv => f.write_str("local virtual environment"),
-            Self::PythonVSCodeExtension => {
-                f.write_str("selected interpreter in the VS Code Python extension")
-            }
+            Self::Editor => f.write_str("selected interpreter in the user's editor"),
```
MichaReiser:
```suggestion
            Self::Editor => f.write_str("selected interpreter in your editor"),
```

MichaReiser:
Thank you

PR #21157, opened 2025-10-31
[ty] implement `typing.NewType` as `Type::NewTypeInstance`

oconnor663:
Fixes https://github.com/astral-sh/ty/issues/742. This PR rewrites and supersedes my previous `NewType` PR, https://github.com/astral-sh/ruff/pull/20126. The high-level changes:

- I've taken @AlexWaygood's advice and added `Type::NewTypeInstance` instead of expanding `NominalInstanceInner`.
- Rebasing on top of https://github.com/astral-sh/ruff/pull/20598 required a rewrite of how the bindings are wired up. Now it's a new special case alongside `infer_legacy_typevar` in `TypeInferenceBuilder::infer_assignment_definition_impl`.
- Related to that, inference of the base type is now deferred. This makes it possible to support cyclic newtypes, and there's a section of test cases for this in the new mdtests.

~~There are several `// REVIEWERS:` questions inline, and I expect I've gone about some things the wrong way (particularly cycle detection?), so I'm leaving this in Draft status for now.~~ I've moved this out of Draft.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
@@ -1,7 +1,5 @@
 # NewType
 
-Currently, ty doesn't support `typing.NewType` in type annotations.
-
 ## Valid forms
 
 ```py
@@ -12,13 +10,389 @@ X = GenericAlias(type, ())
 A = NewType("A", int)
 # TODO: typeshed for `typing.GenericAlias` uses `type` for the first argument. `NewType` should be special-cased
 # to be compatible with `type`
-# error: [invalid-argument-type] "Argument to function `__new__` is incorrect: Expected `type`, found `NewType`"
+# error: [invalid-argument-type] "Argument to function `__new__` is incorrect: Expected `type`, found `<NewType pseudo-class 'A'>`"
 B = GenericAlias(A, ())
 
 def _(
     a: A,
     b: B,
 ):
-    reveal_type(a)  # revealed: @Todo(Support for `typing.NewType` instances in type expressions)
+    reveal_type(a)  # revealed: A
     reveal_type(b)  # revealed: @Todo(Support for `typing.GenericAlias` instances in type expressions)
+```
+
+## Subtyping
+
+The basic purpose of `NewType` is that it acts like a subtype of its base, but not the exact same
+type (i.e. not an alias).
+
+```py
+from typing_extensions import NewType
+from ty_extensions import static_assert, is_subtype_of, is_equivalent_to
+
+Foo = NewType("Foo", int)
+Bar = NewType("Bar", Foo)
+
+static_assert(is_subtype_of(Foo, int))
+static_assert(not is_equivalent_to(Foo, int))
+
+static_assert(is_subtype_of(Bar, Foo))
+static_assert(is_subtype_of(Bar, int))
+static_assert(not is_equivalent_to(Bar, Foo))
+
+Foo(42)
+Foo(Foo(42))  # allowed: `Foo` is a subtype of `int`.
+Foo(Bar(Foo(42)))  # allowed: `Bar` is a subtype of `int`.
+Foo(True)  # allowed: `bool` is a subtype of `int`.
+Foo("forty-two")  # error: [invalid-argument-type] "Argument is incorrect: Expected `int`, found `Literal["forty-two"]`"
+
+def f(_: int): ...
+def g(_: Foo): ...
+def h(_: Bar): ...
+
+f(42)
+f(Foo(42))
+f(Bar(Foo(42)))
+
+g(42)  # error: [invalid-argument-type] "Argument to function `g` is incorrect: Expected `Foo`, found `Literal[42]`"
+g(Foo(42))
+g(Bar(Foo(42)))
+
+h(42)  # error: [invalid-argument-type] "Argument to function `h` is incorrect: Expected `Bar`, found `Literal[42]`"
+h(Foo(42))  # error: [invalid-argument-type] "Argument to function `h` is incorrect: Expected `Bar`, found `Foo`"
+h(Bar(Foo(42)))
+```
+
+## Member and method lookup work
+
+```py
+from typing_extensions import NewType
+
+class Foo:
+    foo_member: str = "hello"
+    def foo_method(self) -> int:
+        return 42
+
+Bar = NewType("Bar", Foo)
+Baz = NewType("Baz", Bar)
+baz = Baz(Bar(Foo()))
+reveal_type(baz.foo_member)  # revealed: str
+reveal_type(baz.foo_method())  # revealed: int
+```
+
+We also infer member access on the `NewType` pseudo-type itself correctly:
+
+```py
+reveal_type(Bar.__supertype__)  # revealed: type | NewType
+reveal_type(Baz.__supertype__)  # revealed: type | NewType
+```
+
+## `NewType` wrapper functions are `Callable`
+
+```py
+from collections.abc import Callable
+from typing_extensions import NewType
+from ty_extensions import CallableTypeOf
+
+Foo = NewType("Foo", int)
+
+def _(obj: CallableTypeOf[Foo]):
+    reveal_type(obj)  # revealed: (int, /) -> Foo
+
+def f(_: Callable[[int], Foo]): ...
+
+f(Foo)
+map(Foo, [1, 2, 3])
+
+def g(_: Callable[[str], Foo]): ...
+
+g(Foo)  # error: [invalid-argument-type]
+```
+
+## `NewType` instances are `Callable` if the base type is
+
+```py
+from typing import NewType, Callable, Any
+from ty_extensions import CallableTypeOf
+
+N = NewType("N", int)
+i = N(42)
+
+y: Callable[..., Any] = i  # error: [invalid-assignment] "Object of type `N` is not assignable to `(...) -> Any`"
+
+# error: [invalid-type-form] "Expected the first argument to `ty_extensions.CallableTypeOf` to be a callable object, but got an object of type `N`"
+def f(x: CallableTypeOf[i]):
+    reveal_type(x)  # revealed: Unknown
+
+class SomethingCallable:
+    def __call__(self, a: str) -> bytes:
+        raise NotImplementedError
+
+N2 = NewType("N2", SomethingCallable)
+j = N2(SomethingCallable())
+
+z: Callable[[str], bytes] = j  # fine
+
+def g(x: CallableTypeOf[j]):
+    reveal_type(x)  # revealed: (a: str) -> bytes
+```
+
+## The name must be a string literal
+
+```py
+from typing_extensions import NewType
+
+def _(name: str) -> None:
+    _ = NewType(name, int)  # error: [invalid-newtype] "The first argument to `NewType` must be a string literal"
+```
+
+However, the literal doesn't necessarily need to be inline, as long as we infer it:
+
+```py
+name = "Foo"
+Foo = NewType(name, int)
+reveal_type(Foo)  # revealed: <NewType pseudo-class 'Foo'>
+```
+
+## The second argument must be a class type or another newtype
+
+Other typing constructs like `Union` are not allowed.
+
+```py
+from typing_extensions import NewType
+
+# error: [invalid-newtype] "invalid base for `typing.NewType`"
+Foo = NewType("Foo", int | str)
+```
+
+We don't emit the "invalid base" diagnostic for `Unknown`, because that typically results from other
+errors that already have a diagnostic, and there's no need to pile on. For example, this mistake
+gives you an "Int literals are not allowed" error, and we'd rather not see an "invalid base" error
+on top of that:
+
+```py
+# error: [invalid-type-form] "Int literals are not allowed in this context in a type expression"
+Foo = NewType("Foo", 42)
+```
+
+## A `NewType` definition must be a simple variable assignment
+
+```py
+from typing import NewType
+
+N: NewType = NewType("N", int)  # error: [invalid-newtype] "A `NewType` definition must be a simple variable assignment"
+```
+
+## Newtypes can be cyclic in various ways
+
+Cyclic newtypes are kind of silly, but it's possible for the user to express them, and it's
+important that we don't go into infinite recursive loops and crash with a stack overflow. In fact,
+this is *why* base type evaluation is deferred; otherwise Salsa itself would crash.
+
+```py
+from typing_extensions import NewType, reveal_type, cast
+
+# Define a directly cyclic newtype.
+A = NewType("A", "A")
+reveal_type(A)  # revealed: <NewType pseudo-class 'A'>
+
+# Typechecking still works. We can't construct an `A` "honestly", but we can `cast` into one.
+a: A
+a = 42  # error: [invalid-assignment] "Object of type `Literal[42]` is not assignable to `A`"
+a = A(42)  # error: [invalid-argument-type] "Argument is incorrect: Expected `A`, found `Literal[42]`"
+a = cast(A, 42)
+reveal_type(a)  # revealed: A
+
+# A newtype cycle might involve more than one step.
+B = NewType("B", "C")
+C = NewType("C", "B")
+reveal_type(B)  # revealed: <NewType pseudo-class 'B'>
+reveal_type(C)  # revealed: <NewType pseudo-class 'C'>
+b: B = cast(B, 42)
+c: C = C(b)
+reveal_type(b)  # revealed: B
+reveal_type(c)  # revealed: C
+# Cyclic types behave in surprising ways. These assignments are legal, even though B and C aren't
+# the same type, because each of them is a subtype of the other.
+b = c
+c = b
+
+# Another newtype could inherit from a cyclic one.
+D = NewType("D", C)
+reveal_type(D)  # revealed: <NewType pseudo-class 'D'>
+d: D
+d = D(42)  # error: [invalid-argument-type] "Argument is incorrect: Expected `C`, found `Literal[42]`"
+d = D(c)
+d = D(b)  # Allowed, the same surprise as above. B and C are subtypes of each other.
+reveal_type(d)  # revealed: D
+```
+
+Normal classes can't inherit from newtypes, but generic classes can be parametrized with them, so we
+also need to detect "ordinary" type cycles that happen to involve a newtype.
+
+```py
+E = NewType("E", list["E"])
+reveal_type(E)  # revealed: <NewType pseudo-class 'E'>
+e: E = E([])
+reveal_type(e)  # revealed: E
+reveal_type(E(E(E(E(E([]))))))  # revealed: E
+reveal_type(E([E([E([]), E([E([])])]), E([])]))  # revealed: E
+E(["foo"])  # error: [invalid-argument-type]
+E(E(E(["foo"])))  # error: [invalid-argument-type]
+```
+
+## `NewType` wrapping preserves singleton-ness and single-valued-ness
+
+```py
+from typing_extensions import NewType
+from ty_extensions import is_singleton, is_single_valued, static_assert
+from types import EllipsisType
+
+A = NewType("A", EllipsisType)
+static_assert(is_singleton(A))
+static_assert(is_single_valued(A))
+reveal_type(type(A(...)) is EllipsisType)  # revealed: Literal[True]
+# TODO: This should be `Literal[True]` also.
+reveal_type(A(...) is ...)  # revealed: bool
+
+B = NewType("B", int)
+static_assert(not is_singleton(B))
+static_assert(not is_single_valued(B))
+```
+
+## `NewType`s of tuples can be iterated/unpacked
+
+```py
+from typing import NewType
+
+N = NewType("N", tuple[int, str])
+
+a, b = N((1, "foo"))
+
+reveal_type(a)  # revealed: int
+reveal_type(b)  # revealed: str
+```
+
+## `isinstance` of a `NewType` instance and its base class is inferred as `Literal[True]`
+
+```py
+from typing import NewType
+
+N = NewType("N", int)
+
+def f(x: N):
+    reveal_type(isinstance(x, int))  # revealed: Literal[True]
+```
+
+However, a `NewType` isn't a real class, so it isn't a valid second argument to `isinstance`:
+
+```py
+def f(x: N):
+    # error: [invalid-argument-type] "Argument to function `isinstance` is incorrect"
+    reveal_type(isinstance(x, N))  # revealed: bool
+```
+
+Because of that, we don't generate any narrowing constraints for it:
+
+```py
+def f(x: N | str):
+    if isinstance(x, N):  # error: [invalid-argument-type]
+        reveal_type(x)  # revealed: N | str
+    else:
+        reveal_type(x)  # revealed: N | str
+```
+
+## Trying to subclass a `NewType` produces an error matching CPython
+
+<!-- snapshot-diagnostics -->
+
+```py
+from typing import NewType
+
+X = NewType("X", int)
+
+class Foo(X): ...  # error: [invalid-base]
+```
+
+## Don't narrow `NewType`-wrapped `Enum`s inside of match arms
+
+`Literal[Foo.X]` is actually disjoint from `N` here:
+
+```py
+from enum import Enum
+from typing import NewType
+
+class Foo(Enum):
+    X = 0
+    Y = 1
+
+N = NewType("N", Foo)
+
+def f(x: N):
+    match x:
+        case Foo.X:
+            reveal_type(x)  # revealed: N
+        case Foo.Y:
+            reveal_type(x)  # revealed: N
+        case _:
+            reveal_type(x)  # revealed: N
+```
+
+## We don't support `NewType` on Python 3.9
+
+We implement `typing.NewType` as a `KnownClass`, but in Python 3.9 it's actually a function, so all
+we get is the `Any` annotations from typeshed. However, `typing_extensions.NewType` is always a
+class. This could be improved in the future, but Python 3.9 is now end-of-life, so it's not
+high-priority.
+
+```toml
+[environment]
+python-version = "3.9"
+```
+
+```py
+from typing import NewType
+
+Foo = NewType("Foo", int)
+reveal_type(Foo)  # revealed: Any
+reveal_type(Foo(42))  # revealed: Any
+
+from typing_extensions import NewType
+
+Bar = NewType("Bar", int)
+reveal_type(Bar)  # revealed: <NewType pseudo-class 'Bar'>
+reveal_type(Bar(42))  # revealed: Bar
+```
+
+## The base of a `NewType` can't be a protocol class or a `TypedDict`
+
+<!-- snapshot-diagnostics -->
+
+```py
+from typing import NewType, Protocol, TypedDict
+
+class Id(Protocol):
+    code: int
+
+UserId = NewType("UserId", Id)  # error: [invalid-newtype]
+
+class Foo(TypedDict):
+    a: int
+
+Bar = NewType("Bar", Foo)  # error: [invalid-newtype]
+```
+
+## TODO: A `NewType` cannot be generic
+
+```py
+from typing import Any, NewType, TypeVar
+
+# All of these are allowed.
+A = NewType("A", list)
+B = NewType("B", list[int])
+B = NewType("B", list[Any])
+
+# But a free typevar is not allowed.
+T = TypeVar("T")
+C = NewType("C", list[T])  # TODO: should be "error: [invalid-newtype]"
 ```
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -66,6 +66,7 @@ use crate::types::generics::{
 use crate::types::infer::infer_unpack_types;
 use crate::types::mro::{Mro, MroError, MroIterator};
 pub(crate) use crate::types::narrow::infer_narrowing_constraint;
+use crate::types::newtype::NewType;
 use crate::types::signatures::{ParameterForm, walk_signature};
 use crate::types::tuple::{TupleSpec, TupleSpecBuilder};
 pub(crate) use crate::types::typed_dict::{TypedDictParams, TypedDictType, walk_typed_dict_type};
@@ -98,6 +99,7 @@ mod instance;
 mod member;
 mod mro;
 mod narrow;
+mod newtype;
 mod protocol_class;
 mod signatures;
 mod special_form;
@@ -783,6 +785,13 @@ pub enum Type<'db> {
     TypedDict(TypedDictType<'db>),
     /// An aliased type (lazily not-yet-unpacked to its value type).
     TypeAlias(TypeAliasType<'db>),
+    /// The set of Python objects that belong to a `typing.NewType` subtype. Note that
+    /// `typing.NewType` itself is a `Type::ClassLiteral` with `KnownClass::NewType`, and the
+    /// identity callables it returns (which behave like subtypes in type expressions) are of
+    /// `Type::KnownInstance` with `KnownInstanceType::NewType`. This `Type` refers to the objects
+    /// wrapped/returned by a specific one of those identity callables, or by another that inherits
+    /// from it.
+    NewTypeInstance(NewType<'db>),
 }
 
 #[salsa::tracked]
@@ -1420,6 +1429,13 @@ impl<'db> Type<'db> {
                 self
             }
             Type::TypeAlias(alias) => alias.value_type(db).normalized_impl(db, visitor),
+            Type::NewTypeInstance(newtype) => {
+                visitor.visit(self, || {
+                    Type::NewTypeInstance(newtype.map_base_class_type(db, |class_type| {
+                        class_type.normalized_impl(db, visitor)
+                    }))
+                })
+            }
             Type::LiteralString
             | Type::AlwaysFalsy
             | Type::AlwaysTruthy
@@ -1482,7 +1498,8 @@ impl<'db> Type<'db> {
             | Type::BoundSuper(_)
             | Type::TypeIs(_)
             | Type::TypedDict(_)
-            | Type::TypeAlias(_) => false,
+            | Type::TypeAlias(_)
+            | Type::NewTypeInstance(_) => false,
         }
     }
 
@@ -1520,6 +1537,10 @@ impl<'db> Type<'db> {
 
             Type::GenericAlias(alias) => Some(ClassType::Generic(alias).into_callable(db)),
 
+            Type::NewTypeInstance(newtype) => {
+                Type::instance(db, newtype.base_class_type(db)).try_upcast_to_callable(db)
+            }
+
             // TODO: This is unsound so in future we can consider an opt-in option to disable it.
             Type::SubclassOf(subclass_of_ty) => match subclass_of_ty.subclass_of() {
                 SubclassOfInner::Class(class) => Some(class.into_callable(db)),
@@ -1548,6 +1569,15 @@ impl<'db> Type<'db> {
                 CallableSignature::from_overloads(wrapper_descriptor.signatures(db)),
                 false,
             ))),
+
+            Type::KnownInstance(KnownInstanceType::NewType(newtype)) => Some(CallableType::single(
+                db,
+                Signature::new(
+                    Parameters::new([Parameter::positional_only(None)
+                        .with_annotated_type(newtype.base(db).instance_type(db))]),
+                    Some(Type::NewTypeInstance(newtype)),
+                ),
+            )),
 
             Type::Never
             | Type::DataclassTransformer(_)
@@ -2429,6 +2459,22 @@ impl<'db> Type<'db> {
                 })
             }
 
+            (Type::NewTypeInstance(self_newtype), Type::NewTypeInstance(target_newtype)) => {
+                self_newtype.has_relation_to_impl(db, target_newtype)
+            }
+
+            (
+                Type::NewTypeInstance(self_newtype),
+                Type::NominalInstance(target_nominal_instance),
+            ) => self_newtype.base_class_type(db).has_relation_to_impl(
+                db,
+                target_nominal_instance.class(db),
+                inferable,
+                relation,
+                relation_visitor,
+                disjointness_visitor,
+            ),
+
             (Type::PropertyInstance(_), _) => {
                 KnownClass::Property.to_instance(db).has_relation_to_impl(
                     db,
@@ -2448,14 +2494,15 @@ impl<'db> Type<'db> {
                 disjointness_visitor,
             ),
 
-            // Other than the special cases enumerated above, `Instance` types and typevars are
-            // never subtypes of any other variants
+            // Other than the special cases enumerated above, nominal-instance types,
+            // newtype-instance types, and typevars are never subtypes of any other variants
             (Type::TypeVar(bound_typevar), _) => {
                 // All inferable cases should have been handled above
                 assert!(!bound_typevar.is_inferable(db, inferable));
                 ConstraintSet::from(false)
             }
             (Type::NominalInstance(_), _) => ConstraintSet::from(false),
+            (Type::NewTypeInstance(_), _) => ConstraintSet::from(false),
         }
     }
 
@@ -2527,6 +2574,10 @@ impl<'db> Type<'db> {
                 visitor.visit((self, other_alias_ty), || {
                     self.is_equivalent_to_impl(db, other_alias_ty, inferable, visitor)
                 })
+            }
+
+            (Type::NewTypeInstance(self_newtype), Type::NewTypeInstance(other_newtype)) => {
+                ConstraintSet::from(self_newtype.is_equivalent_to_impl(db, other_newtype))
             }
 
             (Type::NominalInstance(first), Type::NominalInstance(second)) => {
@@ -3288,6 +3339,19 @@ impl<'db> Type<'db> {
                     )
                 }),
 
+            (Type::NewTypeInstance(left), Type::NewTypeInstance(right)) => {
+                left.is_disjoint_from_impl(db, right)
+            }
+            (Type::NewTypeInstance(newtype), other) | (other, Type::NewTypeInstance(newtype)) => {
+                Type::instance(db, newtype.base_class_type(db)).is_disjoint_from_impl(
+                    db,
+                    other,
+                    inferable,
+                    disjointness_visitor,
+                    relation_visitor,
+                )
+            }
+
             (Type::PropertyInstance(_), other) | (other, Type::PropertyInstance(_)) => {
                 KnownClass::Property.to_instance(db).is_disjoint_from_impl(
                     db,
@@ -3432,6 +3496,9 @@ impl<'db> Type<'db> {
             Type::TypeIs(type_is) => type_is.is_bound(db),
             Type::TypedDict(_) => false,
             Type::TypeAlias(alias) => alias.value_type(db).is_singleton(db),
+            Type::NewTypeInstance(newtype) => {
+                Type::instance(db, newtype.base_class_type(db)).is_singleton(db)
+            }
         }
     }
 
@@ -3482,6 +3549,9 @@ impl<'db> Type<'db> {
             }
 
             Type::NominalInstance(instance) => instance.is_single_valued(db),
+            Type::NewTypeInstance(newtype) => {
+                Type::instance(db, newtype.base_class_type(db)).is_single_valued(db)
+            }
 
             Type::BoundSuper(_) => {
                 // At runtime two super instances never compare equal, even if their arguments are identical.
@@ -3645,7 +3715,8 @@ impl<'db> Type<'db> {
             | Type::ProtocolInstance(_)
             | Type::PropertyInstance(_)
             | Type::TypeIs(_)
-            | Type::TypedDict(_) => None,
+            | Type::TypedDict(_)
+            | Type::NewTypeInstance(_) => None,
         }
     }
 
@@ -3732,6 +3803,7 @@ impl<'db> Type<'db> {
             Type::Dynamic(_) | Type::Never => Place::bound(self).into(),
 
             Type::NominalInstance(instance) => instance.class(db).instance_member(db, name),
+            Type::NewTypeInstance(newtype) => newtype.base_class_type(db).instance_member(db, name),
 
             Type::ProtocolInstance(protocol) => protocol.instance_member(db, name),
 
@@ -4404,6 +4476,7 @@ impl<'db> Type<'db> {
 
             Type::NominalInstance(..)
             | Type::ProtocolInstance(..)
+            | Type::NewTypeInstance(..)
             | Type::BooleanLiteral(..)
             | Type::IntLiteral(..)
             | Type::StringLiteral(..)
@@ -4842,6 +4915,8 @@ impl<'db> Type<'db> {
                     .value_type(db)
                     .try_bool_impl(db, allow_short_circuit, visitor)
             })?,
+            Type::NewTypeInstance(newtype) => Type::instance(db, newtype.base_class_type(db))
+                .try_bool_impl(db, allow_short_circuit, visitor)?,
         };
 
         Ok(truthiness)
@@ -5528,7 +5603,7 @@ impl<'db> Type<'db> {
                 SubclassOfInner::Class(class) => Type::from(class).bindings(db),
             },
 
-            Type::NominalInstance(_) | Type::ProtocolInstance(_) => {
+            Type::NominalInstance(_) | Type::ProtocolInstance(_) | Type::NewTypeInstance(_) => {
                 // Note that for objects that have a (possibly not callable!) `__call__` attribute,
                 // we will get the signature of the `__call__` attribute, but will pass in the type
                 // of the original object as the "callable type". That ensures that we get errors
@@ -5580,6 +5655,16 @@ impl<'db> Type<'db> {
             Type::SpecialForm(_) => CallableBinding::not_callable(self).into(),
 
             Type::EnumLiteral(enum_literal) => enum_literal.enum_class_instance(db).bindings(db),
+
+            Type::KnownInstance(KnownInstanceType::NewType(newtype)) => Binding::single(
+                self,
+                Signature::new(
+                    Parameters::new([Parameter::positional_only(None)
+                        .with_annotated_type(newtype.base(db).instance_type(db))]),
+                    Some(Type::NewTypeInstance(newtype)),
+                ),
+            )
+            .into(),
 
             Type::KnownInstance(known_instance) => {
                 known_instance.instance_fallback(db).bindings(db)
@@ -5716,6 +5801,7 @@ impl<'db> Type<'db> {
 
             match ty {
                 Type::NominalInstance(nominal) => nominal.tuple_spec(db),
+                Type::NewTypeInstance(newtype) => non_async_special_case(db, Type::instance(db, newtype.base_class_type(db))),
                 Type::GenericAlias(alias) if alias.origin(db).is_tuple(db) => {
                     Some(Cow::Owned(TupleSpec::homogeneous(todo_type!(
                         "*tuple[] annotations"
@@ -6346,6 +6432,9 @@ impl<'db> Type<'db> {
             Type::ClassLiteral(class) => Some(Type::instance(db, class.default_specialization(db))),
             Type::GenericAlias(alias) => Some(Type::instance(db, ClassType::from(alias))),
             Type::SubclassOf(subclass_of_ty) => Some(subclass_of_ty.to_instance(db)),
+            Type::KnownInstance(KnownInstanceType::NewType(newtype)) => {
+                Some(Type::NewTypeInstance(newtype))
+            }
             Type::Union(union) => union.to_instance(db),
             // If there is no bound or constraints on a typevar `T`, `T: object` implicitly, which
             // has no instance type. Otherwise, synthesize a typevar with bound or constraints
@@ -6376,7 +6465,8 @@ impl<'db> Type<'db> {
             | Type::AlwaysTruthy
             | Type::AlwaysFalsy
             | Type::TypeIs(_)
-            | Type::TypedDict(_) => None,
+            | Type::TypedDict(_)
+            | Type::NewTypeInstance(_) => None,
         }
     }
 
@@ -6455,6 +6545,7 @@ impl<'db> Type<'db> {
 
             Type::KnownInstance(known_instance) => match known_instance {
                 KnownInstanceType::TypeAliasType(alias) => Ok(Type::TypeAlias(*alias)),
+                KnownInstanceType::NewType(newtype) => Ok(Type::NewTypeInstance(*newtype)),
                 KnownInstanceType::TypeVar(typevar) => {
                     let index = semantic_index(db, scope_id.file(db));
                     Ok(bind_typevar(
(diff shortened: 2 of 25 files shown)


crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,305 @@
...
+            // Note that NewTypeBaseIter does *not* catch this kind of cycle, because it stops
+            // after returning `NewTypeBase::ClassType` for `list`.
+            let mut is_first_encounter = false;
+            let base_type = Type::instance(db, base_class_type);
+            _ = visitor.visit(base_type, || {
+                // If we hit in a cycle, this closure won't run.
+                is_first_encounter = true;
+                base_type
+            });
+            if is_first_encounter {
```
oconnor663: @AlexWaygood One of the downsides of using the restricted `NewTypeBase` enum instead of just making the base a `Type`, is that calling `visitor.visit()` here isn't "natural". We need to synthesize a `Type` just to be able to make this call for its side effects, and then we discard the return value. But if we omit this (and of course I omitted it at first), the last line of the mdtests crashes with a stack overflow. I wonder if that's an argument for making our visitor machinery more flexible somehow?

(Incidentally `normalized_impl` is in the exact same situation, but I don't think I have a test case to provoke a stack overflow for it. Any suggestions?)

AlexWaygood: Ah I think you just want to do the `visitor.visit()` thing in the `match` in `types.rs`? All tests pass if I make this change to your branch, anyway :-)

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 2c65727ee2..0cfc92c5dd 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6893,7 +6893,7 @@ impl<'db> Type<'db> {
             },
 
             Type::NewTypeInstance(newtype) => {
-                Type::NewTypeInstance(newtype.apply_type_mapping_impl(db, type_mapping, tcx, visitor))
+                visitor.visit(self, || Type::NewTypeInstance(newtype.apply_type_mapping_impl(db, type_mapping, tcx, visitor)))
             },
 
             Type::ProtocolInstance(instance) => {
diff --git a/crates/ty_python_semantic/src/types/newtype.rs b/crates/ty_python_semantic/src/types/newtype.rs
index 57c879343e..f9dda5ec78 100644
--- a/crates/ty_python_semantic/src/types/newtype.rs
+++ b/crates/ty_python_semantic/src/types/newtype.rs
@@ -214,30 +214,7 @@ impl<'db> NewType<'db> {
         visitor: &ApplyTypeMappingVisitor<'db>,
     ) -> Self {
         self.map_base_class_type(db, |base_class_type| {
-            // We don't really want to work with `Type` directly (`NewTypeBase` is much more
-            // restricted), but we need to let the `ApplyTypeMappingVisitor` see the `Type` of the
-            // base class, because this is its only chance to detect an indirect cycle like:
-            //
-            // ```py
-            // Foo = NewType("Foo", list["Foo"])
-            // ```
-            //
-            // Note that NewTypeBaseIter does *not* catch this kind of cycle, because it stops
-            // after returning `NewTypeBase::ClassType` for `list`.
-            let mut is_first_encounter = false;
-            let base_type = Type::instance(db, base_class_type);
-            _ = visitor.visit(base_type, || {
-                // If we hit in a cycle, this closure won't run.
-                is_first_encounter = true;
-                base_type
-            });
-            if is_first_encounter {
-                base_class_type.apply_type_mapping_impl(db, type_mapping, tcx, visitor)
-            } else {
-                // Cycle detected. Fall back to `object`.
-                // REVIEWERS: Is this correct?
-                ClassType::object(db)
-            }
+            base_class_type.apply_type_mapping_impl(db, type_mapping, tcx, visitor)
         })
     }
 }
```

oconnor663: Ohhhh that does make sense.

AlexWaygood:
This change fixes the panic when type-checking py-fuzzer in CI (and I assume also the panics mypy_primer is surfacing) without breaking any of your tests. I _think_ it's the correct solution... in general, as soon as you do anything to a newtype, it's meant to lose its "newtypeness", so it makes sense to me that obtaining the meta-type of a newtype instance would just return the underlying nominal class for that newtype instance

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 2c65727ee2..13918d56ca 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -6744,9 +6744,7 @@ impl<'db> Type<'db> {
             // understand a more specific meta type in order to correctly handle `__getitem__`.
             Type::TypedDict(typed_dict) => SubclassOfType::from(db, typed_dict.defining_class()),
             Type::TypeAlias(alias) => alias.value_type(db).to_meta_type(db),
-            Type::NewTypeInstance(newtype) => {
-                Type::KnownInstance(KnownInstanceType::NewType(newtype))
-            }
+            Type::NewTypeInstance(newtype) => Type::from(newtype.base_class_type(db)),
         }
     }
 ```

File: crates/ty_python_semantic/src/types/infer/builder.rs [hunk 2]
```diff
@@ -4496,11 +4506,129 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        if !arguments.keywords.is_empty() {
+            return error(
+                &self.context,
+                "Keyword arguments are not supported in `NewType` creation",
+                call_expr,
+            );
+        }
+
+        // REVIEWERS: Is it worth trying to do full argument parsing here? If so is there some
+        // machinery we might want to reuse or factor out somewhere else?
```
AlexWaygood: I think this is probably fine for now? We could always try to factor out the common bits between this and TypeVar parsing as a followup. But it's not _that_ much duplication

crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,305 @@
...
+                }
+            }
+        }
+        false
+    }
+
+    pub(crate) fn has_relation_to_impl(self, db: &'db dyn Db, other: Self) -> ConstraintSet<'db> {
+        // REVIEWERS: Is this correct?
+        ConstraintSet::from(self.is_subtype_of(db, other))
+    }
```
AlexWaygood: For consistency with our other modules, I think it would be better to write `is_subtype_of` as a wrapper around `has_relation_to(_impl)`, rather than vice versa. But I think this is correct, yes -- subtyping/assignability/redundancy should all give the same result when the left-hand side and the right-hand side are both newtypes.

crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,305 @@
...
+            _ = visitor.visit(base_type, || {
+                // If we hit in a cycle, this closure won't run.
+                is_first_encounter = true;
+                base_type
+            });
+            if is_first_encounter {
+                base_class_type.normalized_impl(db, visitor)
+            } else {
+                // Cycle detected. Fall back to `object`.
+                // REVIEWERS: Is this correct?
```
AlexWaygood: (Same response as https://github.com/astral-sh/ruff/pull/21157/files#r2482511640)

crates/ty_python_semantic/src/types/visitor.rs
```diff
@@ -156,7 +156,9 @@ impl<'db> From<Type<'db>> for TypeKind<'db> {
             | Type::ModuleLiteral(_)
             | Type::ClassLiteral(_)
             | Type::SpecialForm(_)
-            | Type::Dynamic(_) => TypeKind::Atomic,
+            | Type::Dynamic(_)
+            // REVIEWERS: is this correct?
+            | Type::NewTypeInstance(_) => TypeKind::Atomic,
```
AlexWaygood: ummm you could probably argue it both ways. I think this is fine for now; if we find a reason in the future why it isn't fine we can always change it

crates/ty_python_semantic/src/types.rs
```diff
@@ -7413,7 +7524,9 @@ impl<'db> VarianceInferable<'db> for Type<'db> {
             | Type::BoundSuper(_)
             | Type::TypeVar(_)
             | Type::TypedDict(_)
-            | Type::TypeAlias(_) => TypeVarVariance::Bivariant,
+            | Type::TypeAlias(_)
+            // REVIEWERS: Is this correct?
+            | Type::NewTypeInstance(_) => TypeVarVariance::Bivariant,
```
AlexWaygood: I _think_ so? You can never have a "generic NewType" -- if it has a generic type as its underlying nominal class, that generic type _must_ have always been specialised. Therefore, it stands to reason that the typevar can never be relevant to this newtype, so must be bivariant with respect to this newtype.

AlexWaygood (commented):
(I just looked at the `// REVIEWER` comments for now, since it's still in draft, but happy to look at the other bits if you'd like me to!)

crates/ty_python_semantic/src/types/newtype.rs
oconnor663: f3b6c6794d43ed97ce09707ce457b6c62c646335

oconnor663:
Relevant to the error formatting tweak I just pushed in daeb800c4c5f3972cc6899dda7f0d210c647eecb, do we like the example in that commit message:

```py
from typing import NewType
from nonexistent_module import Foo
Bar = NewType("Bar", Foo)
```

```
error[invalid-newtype]: invalid base for `typing.NewType`
 --> test.py:3:22
  |
1 | from typing import NewType
2 | from nonexistent_module import Foo
3 | Bar = NewType("Bar", Foo)
  |                      ^^^ type `Unknown`
info: The base of a `NewType` must be a class type or another `NewType`.
info: rule `invalid-newtype` is enabled by default
```

That's a result of the check that says a `NewType` base should be either a class type or another newtype. `Unknown` is neither of those things. Is this error what we expect? Will this sort of thing by noisy around unresolved imports? (I haven't played much with how those work yet.)

AlexWaygood:
> Diagnostic diff on [typing conformance tests](https://github.com/python/typing/tree/d4f39b27a4a47aac8b6d4019e1b0b5b3156fabdc/conformance)

Nice, these changes all look like true positives -- places where we should have been emitting diagnostics, but we previously weren't!

AlexWaygood:
> That's a result of the check that says a `NewType` base should be either a class type or another newtype. `Unknown` is neither of those things. Is this error what we expect? Will this sort of thing by noisy around unresolved imports? (I haven't played much with how those work yet.)

I think it would be good to suppress the diagnostic here if the second argument has a dynamic type. The user will get a diagnostic from the unresolved import in your example; there's not much use in having cascading diagnostics later on in the module due to the same issue.

crates/ty_python_semantic/src/types.rs
```diff
@@ -1475,6 +1492,8 @@ impl<'db> Type<'db> {
 
             Type::GenericAlias(alias) => Some(ClassType::Generic(alias).into_callable(db)),
 
+            Type::NewTypeInstance(newtype) => Some(newtype.base_class_type(db).into_callable(db)),
```
AlexWaygood: this isn't quite right. The question this branch is asking is "What is the `Callable` supertype for an instance of a given newtype?". I.e., given this snippet:

```py
N = NewType("N", int)
i = N(42)
```

The type of `i` is `N`, but what is the `Callable` supertype of `i`? The correct answer should be that it has no `Callable` supertype (the object `i` is not callable), but on your branch currently we say it is in some situations:

```py
from typing import NewType, Callable, Any
from ty_extensions import CallableTypeOf

N = NewType("N", int)
i = N(42)

# no diagnostic here, but there should be one!
y: Callable[..., Any] = i

def f(x: CallableTypeOf[i]):
    # revealed: Overload[(x: @Todo(Support for `typing.TypeAlias`) = Literal[0], /) -> int, (x: str | bytes | bytearray, /, base: SupportsIndex) -> int]
    reveal_type(x)
```

The fix for the bug here is simple:

```suggestion
            Type::NewTypeInstance(newtype) => {
                Type::instance(db, newtype.base_class_type(db)).try_upcast_to_callable(db)
            }
```

you were answering the question by delegating to the `Callable` supertype of the newtype's underlying class, but what you need to do is to delegate to the `Callable` supertype of an _instance_ of the newtype's underlying class

crates/ty_python_semantic/src/types.rs
```diff
@@ -2397,6 +2441,7 @@ impl<'db> Type<'db> {
                 ConstraintSet::from(false)
             }
             (Type::NominalInstance(_), _) => ConstraintSet::from(false),
+            (Type::NewTypeInstance(_), _) => ConstraintSet::from(false),
```
AlexWaygood, line 2505: nit: can you adjust the comment a few lines above?

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 3ce82eebc9..0370e86e60 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -2433,8 +2433,8 @@ impl<'db> Type<'db> {
                 disjointness_visitor,
             ),
 
-            // Other than the special cases enumerated above, `Instance` types and typevars are
-            // never subtypes of any other variants
+            // Other than the special cases enumerated above, nominal-instance types,
+            // newtype-instance types and typevars are never subtypes of any other variants
             (Type::TypeVar(bound_typevar), _) => {
                 // All inferable cases should have been handled above
                 assert!(!bound_typevar.is_inferable(db, inferable));
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -3373,6 +3431,9 @@ impl<'db> Type<'db> {
             Type::TypeIs(type_is) => type_is.is_bound(db),
             Type::TypedDict(_) => false,
             Type::TypeAlias(alias) => alias.value_type(db).is_singleton(db),
+            Type::NewTypeInstance(newtype) => {
+                Type::instance(db, newtype.base_class_type(db)).is_singleton(db)
+            }
```
AlexWaygood, line 3501: nit: could you add a test for this? E.g. a newtype of `types.EllipsisType` should be recognized as a singleton type, since we recognize `types.EllipsisType` as being a singleton type. You can add a test using our `ty_extensions.is_singleton` function

crates/ty_python_semantic/src/types.rs
```diff
@@ -3423,6 +3484,9 @@ impl<'db> Type<'db> {
             }
 
             Type::NominalInstance(instance) => instance.is_single_valued(db),
+            Type::NewTypeInstance(newtype) => {
+                Type::instance(db, newtype.base_class_type(db)).is_single_valued(db)
+            }
```
AlexWaygood, line 3554: same here -- this looks correct, but it would be great to have an explicit test for this

crates/ty_python_semantic/src/types.rs
```diff
@@ -5155,6 +5224,24 @@ impl<'db> Type<'db> {
...
+                                Parameter::positional_or_keyword(Name::new_static("tp")),
+                            ]),
+                            // The actual return type is a KnownInstanceType::NewType(_), but each
+                            // callsite gets a unique payload, including a `Definition` that points
+                            // to that callsite. See `struct NewType` and `infer_newtype_expression`.
+                            None,
+                        ),
+                    )
+                    .into()
+                }
```
AlexWaygood: one issue here is that on Python 3.9, `NewType` is a function rather than a class. I don't know if this really matters this much TBH, since Python 3.9 is now end-of-life, and supporting both the pre-3.9 version of `NewType` (where it was a function) and the 3.10-version of `NewType` (where it's a class) is going to lead to complications for us.

At the very least, it might be good to add an mdtest with `python-version = "3.9"` so that we document that we don't support `NewType` on Python <=3.9 currently. But I would certainly be inclined to leave <=3.9 support out of the first PR adding support for `NewType`.

crates/ty_python_semantic/src/types.rs
```diff
@@ -5631,6 +5728,7 @@ impl<'db> Type<'db> {
 
             match ty {
                 Type::NominalInstance(nominal) => nominal.tuple_spec(db),
+                Type::NewTypeInstance(newtype) => non_async_special_case(db, Type::instance(db, newtype.base_class_type(db))),
```
AlexWaygood, line 5804: it would be great to add a test for this -- e.g.

```py
from typing import NewType

N = NewType("N", tuple[int, str])

a, b = N((1, "foo"))

reveal_type(a)  # revealed: int
reveal_type(b)  # revealed: str
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7049,6 +7163,15 @@ impl<'db> Type<'db> {
...
             }
 
+            Type::NewTypeInstance(newtype) => {
+                newtype.base_class_type(db).find_legacy_typevars_impl(
+                    db,
+                    binding_context,
+                    typevars,
+                    visitor,
+                );
+            }
```
AlexWaygood: if we do the proper validation elsewhere so that you're unable to construct a `NewType` from an unspecialized generic class, I think we can optimize this a bit and just skip it?

```suggestion
            Type::NewTypeInstance(newtype) => {
                // A newtype can never be constructed from an unspecialized generic class,
                // so it is impossible that we could ever find any legacy typevars
                // in a newtype instance or its underlying class
            }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -7614,6 +7740,10 @@ pub enum KnownInstanceType<'db> {
     /// A constraint set, which is exposed in mdtests as an instance of
     /// `ty_extensions.ConstraintSet`.
     ConstraintSet(TrackedConstraintSet<'db>),
+
+    /// An identity function created with `typing.NewType(name, base)`, which behaves like a
```
AlexWaygood: nit: when you call it, it has an identity-function signature, but it is not actually an instance of `types.FunctionType` at runtime (it's an instance of `typing.NewType`)

```suggestion
    /// An identity callable created with `typing.NewType(name, base)`, which behaves like a
```

crates/ty_python_semantic/src/types/bound_super.rs
```diff
@@ -404,6 +404,9 @@ impl<'db> BoundSuperType<'db> {
                         .to_specialized_instance(db, [key_builder.build(), value_builder.build()]),
                 );
             }
+            Type::NewTypeInstance(newtype) => {
+                return delegate_to(Type::instance(db, newtype.base_class_type(db)));
+            }
```
AlexWaygood, line 409: nit: could you add a test for this to https://github.com/astral-sh/ruff/blob/bff32a41dc440b30764e9414767794e01d25c265/crates/ty_python_semantic/resources/mdtest/class/super.md?plain=1#L63-L121 ?

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -161,7 +161,8 @@ impl<'db> ClassBase<'db> {
             | Type::AlwaysFalsy
             | Type::AlwaysTruthy
             | Type::TypeIs(_)
-            | Type::TypedDict(_) => None,
+            | Type::TypedDict(_)
+            | Type::NewTypeInstance(_) => None,
```
AlexWaygood: I'd feel slightly safer here if we delegated:

```suggestion
            | Type::TypedDict(_) => None,
            
            Type::NewTypeInstance(newtype) => {
                ClassBase::try_from_type(db, Type::instance(db, newtype.base_class_type(db)), subclass)
            }
```

It's unlikely that we'll ever accept any `NominalInstance` types as base classes, but there _is_ a TODO above the `NominalInstance` branch above which could _plausibly_ be fixed, and it seems important to keep the behaviour for a given `NewType` here consistent with the behaviour of that `NewType`'s nominal-instance supertype

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -1093,6 +1093,11 @@ fn is_instance_truthiness<'db>(
 
         Type::NominalInstance(..) => always_true_if(is_instance(&ty)),
 
+        Type::NewTypeInstance(newtype) => always_true_if(is_instance(&Type::instance(
+            db,
+            newtype.base_class_type(db),
+        ))),
```
AlexWaygood, line 1107: It would be great to add a test for this -- something like this?

```py
from typing import NewType

N = NewType("N", int)

def f(x: N):
    # TODO: we should emit an error here (requires PEP-613 support)
    reveal_type(isinstance(x, N))  # revealed: bool

    reveal_type(isinstance(x, int))  # revealed: Literal[True]
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4345,8 +4348,15 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                             matches!(cls, KnownClass::TypeVar | KnownClass::ExtensionsTypeVar)
                         });
 
+                    let is_newtype = callable_type
+                        .as_class_literal()
+                        .and_then(|cls| cls.known(self.db()))
+                        .is_some_and(|cls| matches!(cls, KnownClass::NewType));
```
AlexWaygood: I think you could do this -- and a similar bit below -- slightly more efficiently, e.g.

```diff
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index b6a9df7d11..95f7369c67 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -4341,24 +4341,20 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                         TypeContext::default(),
                     );
 
-                    let typevar_class = callable_type
+                    let known_class = callable_type
                         .as_class_literal()
-                        .and_then(|cls| cls.known(self.db()))
-                        .filter(|cls| {
-                            matches!(cls, KnownClass::TypeVar | KnownClass::ExtensionsTypeVar)
-                        });
-
-                    let is_newtype = callable_type
-                        .as_class_literal()
-                        .and_then(|cls| cls.known(self.db()))
-                        .is_some_and(|cls| matches!(cls, KnownClass::NewType));
+                        .and_then(|cls| cls.known(self.db()));
 
-                    let ty = if let Some(typevar_class) = typevar_class {
-                        self.infer_legacy_typevar(target, call_expr, definition, typevar_class)
-                    } else if is_newtype {
-                        self.infer_newtype_expression(target, call_expr, definition)
-                    } else {
-                        self.infer_call_expression_impl(call_expr, callable_type, tcx)
+                    let ty = match known_class {
+                        Some(
+                            typevar_class @ (KnownClass::TypeVar | KnownClass::ExtensionsTypeVar),
+                        ) => {
+                            self.infer_legacy_typevar(target, call_expr, definition, typevar_class)
+                        }
+                        Some(KnownClass::NewType) => {
+                            self.infer_newtype_expression(target, call_expr, definition)
+                        }
+                        _ => self.infer_call_expression_impl(call_expr, callable_type, tcx),
                     };
 
                     self.store_expression_type(value, ty);
@@ -4743,21 +4739,16 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         let callable_type =
             self.infer_maybe_standalone_expression(func.as_ref(), TypeContext::default());
 
-        let is_typevar = callable_type
-            .as_class_literal()
-            .and_then(|cls| cls.known(self.db()))
-            .filter(|cls| matches!(cls, KnownClass::TypeVar | KnownClass::ExtensionsTypeVar))
-            .is_some();
-
-        let is_newtype = callable_type
+        let known_class = callable_type
             .as_class_literal()
-            .and_then(|cls| cls.known(self.db()))
-            .is_some_and(|cls| matches!(cls, KnownClass::NewType));
+            .and_then(|cls| cls.known(self.db()));
 
-        if is_typevar {
-            self.infer_typevar_assignment_deferred(arguments);
-        } else if is_newtype {
-            self.infer_newtype_assignment_deferred(arguments);
+        match known_class {
+            Some(KnownClass::TypeVar | KnownClass::ExtensionsTypeVar) => {
+                self.infer_typevar_assignment_deferred(arguments);
+            }
+            Some(KnownClass::NewType) => self.infer_newtype_assignment_deferred(arguments),
+            _ => {}
         }
     }
```

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -235,7 +235,8 @@ impl ClassInfoConstraintFunction {
             | Type::TypeIs(_)
             | Type::WrapperDescriptor(_)
             | Type::DataclassTransformer(_)
-            | Type::TypedDict(_) => None,
+            | Type::TypedDict(_)
+            | Type::NewTypeInstance(_) => None,
```
AlexWaygood, line 256: could we add an explicit test for this? e.g.

```py
from typing import NewType

N = NewType("N", int)

def f(x: N | str):
    # TODO: we should emit an error here (requires PEP-613 support)
    if isinstance(x, N):
        # we do not narrow the type here,
        # since the `isinstance()` call will actually fail at runtime!
        reveal_type(x)  # revealed: N | str
    else:
        reveal_type(x)  # revealed: N | str
```

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -170,7 +171,11 @@ impl<'db> ClassBase<'db> {
                 | KnownInstanceType::TypeVar(_)
                 | KnownInstanceType::Deprecated(_)
                 | KnownInstanceType::Field(_)
-                | KnownInstanceType::ConstraintSet(_) => None,
+                | KnownInstanceType::ConstraintSet(_)
+                // A class inheriting from a newtype would make intuitive sense, but newtype
+                // wrappers are just identity functions at runtime, so this sort of inheritance
+                // doesn't work and isn't allowed.
+                | KnownInstanceType::NewType(_) => None,
```
AlexWaygood, line 182: the runtime actually has a special-cased error message specifically for this, which we could consider emulating:

```pycon
>>> from typing import NewType
>>> X = NewType("X", int)
>>> class Foo(X): ...
... 
    class Foo(X): ...
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 3428, in __init_subclass__
    raise TypeError(
    ...<2 lines>...
    )
TypeError: Cannot subclass an instance of NewType. Perhaps you were looking for: `Foo = NewType('Foo', X)`
```

But that obviously doesn't need to be done in this PR!

crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,242 @@
...
+
+use crate::Db;
+use crate::semantic_index::definition::{Definition, DefinitionKind};
+use crate::types::constraints::ConstraintSet;
+use crate::types::{ClassType, Type, definition_expression_type};
+use ruff_db::parsed::parsed_module;
+use ruff_python_ast as ast;
+
+/// A `typing.NewType` declaration, either from the perspective of the
+/// identity-function-that-acts-like-a-subtype-in-type-expressions returned by the call to
```
AlexWaygood:
```suggestion
/// identity-callable-that-acts-like-a-subtype-in-type-expressions returned by the call to
```

crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,242 @@
...
+/// `typing.NewType`, or from the perspective of instances of that subtype returned by the
```
AlexWaygood:
```suggestion
/// `typing.NewType()`, or from the perspective of instances of that subtype returned by the
```

crates/ty_python_semantic/src/types/newtype.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Foo = typing.NewType("Foo", int)
+/// x = Foo(42)
+/// ```
+///
+/// The revealed types there are:
+/// - `typing.NewType`: `Type::ClassLiteral(ClassLiteral)` with `KnownClass::NewType`.
+/// - `Foo`: `Type::KnownInstance(KnownInstanceType::NewType(NewType { .. }))`
+/// - `x`: `Type::NewTypeInstance(NewType { .. })`
+#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
```
AlexWaygood, line 30: nit: could you add a note similar to this to the doc-comment for the struct?

https://github.com/astral-sh/ruff/blob/bff32a41dc440b30764e9414767794e01d25c265/crates/ty_python_semantic/src/types.rs#L481-L485

AlexWaygood (commented):
Here's my full review. Overall this looks excellent!! This definitely looks like right overall approach IMO.

My main feedback is that it would be great to have some more tests for various parts of the behaviour that you're implementing here. I left some comments inline about things that look correct, but that it would be great to explicitly test. It might also be worth looking through the typing conformance-suite tests for newtype, and/or [mypy's tests for newtype](https://github.com/python/mypy/blob/master/test-data/unit/check-newtype.test), to see if there are any other things that would be worth testing for explicitly.

crates/ty_python_semantic/src/types.rs
oconnor663: What's your recommendation for this diagnostic? It doesn't look like I have an `InferContext` at this point to directly emit something. Is there a type that I can return that indirectly forces a diagnostic? (edit: ah you're not asking for a diagnostic)

crates/ty_python_semantic/src/types.rs
oconnor663: What's the story with unspecified type vars? Do they get replaced with `object` at some point, such that the call to `.base_class_type(db).find_legacy_typevars_impl(...)` is definitely not going to find anything?

AlexWaygood: yes, they _should_ always be substituted with `Unknown` if they're not specified: we see here that `list[Unknown]` appears in the MRO rather than `list[T]` or anything else:

https://play.ty.dev/89863b06-3878-484c-a393-bb9562c70efc

crates/ty_python_semantic/src/types/function.rs
oconnor663: When I try this, I _do_ get an error there:

> Expected `type | UnionType | tuple[Unknown, ...]`, found `<NewType pseudo-class 'N'>`.

That seems like what we want?

AlexWaygood: Aha, yes! The missing piece here was added in https://github.com/astral-sh/ruff/pull/21195. That's definitely the behaviour we want!

AlexWaygood: (It would still be good to add a test!)

crates/ty_python_semantic/src/types.rs
```diff
@@ -783,6 +785,13 @@ pub enum Type<'db> {
     TypedDict(TypedDictType<'db>),
     /// An aliased type (lazily not-yet-unpacked to its value type).
     TypeAlias(TypeAliasType<'db>),
+    /// The set of Python objects that belong to a `typing.NewType` subtype. Note that
+    /// `typing.NewType` itself is a `Type::ClassLiteral` with `KnownClass::NewType`, and the
+    /// identity functions it returns (which behave like subtypes in type expressions) are of
+    /// `Type::KnownInstance` with `KnownInstanceType::NewType`. This `Type` refers to the objects
+    /// wrapped/returned by a specific one of those identity functions, or by another that inherits
```
```suggestion
    /// identity callables it returns (which behave like subtypes in type expressions) are of
    /// `Type::KnownInstance` with `KnownInstanceType::NewType`. This `Type` refers to the objects
    /// wrapped/returned by a specific one of those identity callables, or by another that inherits
```

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -171,7 +177,11 @@ impl<'db> ClassBase<'db> {
                 | KnownInstanceType::Deprecated(_)
                 | KnownInstanceType::Field(_)
                 | KnownInstanceType::ConstraintSet(_)
-                | KnownInstanceType::UnionType(_) => None,
+                | KnownInstanceType::UnionType(_)
+                // A class inheriting from a newtype would make intuitive sense, but newtype
+                // wrappers are just identity functions at runtime, so this sort of inheritance
```
AlexWaygood:
```suggestion
                // wrappers are just identity callables at runtime, so this sort of inheritance
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,257 @@ X = GenericAlias(type, ())
...
+
+N = NewType("N", int)
+i = N(42)
+
+y: Callable[..., Any] = i  # error: [invalid-assignment] "Object of type `N` is not assignable to `(...) -> Any`"
+
+# error: [invalid-type-form] "Expected the first argument to `ty_extensions.CallableTypeOf` to be a callable object, but got an object of type `N`"
+def f(x: CallableTypeOf[i]):
+    reveal_type(x)  # revealed: Unknown
+```
```
AlexWaygood, line 138: we could also add a test here for a `NewType` of a type that _is_ callable, e.g.

````suggestion
```py
from typing import NewType, Callable, Any
from ty_extensions import CallableTypeOf

N = NewType("N", int)
i = N(42)

y: Callable[..., Any] = i  # error: [invalid-assignment] "Object of type `N` is not assignable to `(...) -> Any`"

# error: [invalid-type-form] "Expected the first argument to `ty_extensions.CallableTypeOf` to be a callable object, but got an object of type `N`"
def f(x: CallableTypeOf[i]):
    reveal_type(x)  # revealed: Unknown

class SomethingCallable:
    def __call__(self, a: str) -> bytes:
        raise NotImplementedError

N2 = NewType("N2", SomethingCallable)
j = N2(SomethingCallable())

z: Callable[[str], bytes] = j  # fine

def g(x: CallableTypeOf[j]):
    reveal_type(x)  # revealed: (str) -> bytes
```
````

**Review by AlexWaygood — COMMENTED** [review]
Looking really good -- a couple more small comments!

File: crates/ty_python_semantic/src/types/infer/builder.rs [hunk 27]
```diff
@@ -4880,34 +4885,149 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
         };
+        let func_ty = self
+            .try_expression_type(func)
+            .unwrap_or_else(|| self.infer_expression(func, TypeContext::default()));
+        let known_class = func_ty
+            .as_class_literal()
+            .and_then(|cls| cls.known(self.db()));
+        if let Some(KnownClass::NewType) = known_class {
+            self.infer_newtype_assignment_deferred(arguments);
+            return;
```
oconnor663, line 5006: I just pushed a big rebase (addressing most of @AlexWaygood's comments, but now I see there are more :-D ), and this part ended up being tricky. It looks like if I don't short-circuit here, I run into a panic about inferring the same expression (the second arg) twice. Please let me know if this looks like the right way to avoid that, and whether it's maybe worthy of a block comment saying why this is sensitive?

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
oconnor663: In retrospect clear that I should have done that :sweat_smile:

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+## Subtyping
+
+The basic purpose of `NewType` is that it acts like a subtype of its base, but not the exact same
+type (i.e. not an alias).
+
+```py
+from typing_extensions import NewType
+
+Foo = NewType("Foo", int)
+Bar = NewType("Bar", Foo)
```
AlexWaygood, line 34: we could add some more assertions here that directly check the subtype relationship:

```suggestion
from typing_extensions import NewType
from ty_extensions import static_assert, is_subtype_of, is_equivalent_to

Foo = NewType("Foo", int)
Bar = NewType("Bar", Foo)

static_assert(is_subtype_of(Foo, int))
static_assert(not is_equivalent_to(Foo, int))

static_assert(is_subtype_of(Bar, Foo))
static_assert(is_subtype_of(Bar, int))
static_assert(not is_equivalent_to(Bar, Foo))
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
AlexWaygood, line 1: One pathological thing I wondered while reviewing this PR was "What happens if you have a newtype of an enum?" I think our behaviour makes more sense than mypy/pyright here, so it might be worth adding a test for it. (I think it's correct _not_ to narrow the type of `x` in the first `match` case to `Literal[Foo.X]`, for example, because `Literal[Foo.X]` is actually disjoint from `N`:

```py
from enum import Enum
from typing import NewType, reveal_type

class Foo(Enum):
    X = 0
    Y = 1

N = NewType("N", Foo)

def f(x: N):
    match x:
        case Foo.X:
            reveal_type(x)  # revealed: N
        case Foo.Y:
            reveal_type(x)  # revealed: N
        case _:
            reveal_type(x)  # revealed: N
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+reveal_type(baz.foo_method())  # revealed: int
+```
+
+## `NewType` wrapper functions are `Callable`
+
+```py
+from collections.abc import Callable
+from typing_extensions import NewType
+
+Foo = NewType("Foo", int)
```
AlexWaygood, line 97: again here, it could be worth adding a direct assertion for the property we're trying to test (the callable supertype of the NewType pseudo-class) as well as the existing assertions you have for the impact this has in user code:

```suggestion
from collections.abc import Callable
from typing_extensions import NewType
from ty_extensions import CallableTypeOf

Foo = NewType("Foo", int)

def _(obj: CallableTypeOf[Foo]):
    reveal_type(obj)  # revealed: (int) -> Foo
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+
+def _(name: str) -> None:
+    _ = NewType(name, int)  # error: [invalid-newtype] "The first argument to `NewType` must be a string literal"
+```
+
+However, the literal doesn't necessarily need to be inline, as long as we infer it:
+
+```py
+name = "Foo"
+Foo = NewType(name, int)  # allowed
```
AlexWaygood: maybe demonstrate here that we infer the name just fine?

```suggestion
name = "Foo"
Foo = NewType(name, int)
reveal_type(Foo)  # revealed: <NewType pseudo-class 'Foo'>
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+Foo = NewType("Foo", int | str)
+```
+
+We don't emit the "invalid base" diagnostic for `Unknown`, because that typically results from other
+errors that already have a diagnostic, and there's no need to pile on. For example:
+
+```py
+# error: [invalid-type-form] "Int literals are not allowed in this context in a type expression"
+Foo = NewType("Foo", 42)
+```
```
AlexWaygood, line 176: I had to read this twice because the prose describes how we _aren't_ going to emit a diagnostic but then the example shows us emitting a diagnostic 😆

````suggestion
```py
# Here we only emit one `invalid-type-form` diagnostic,
# rather than one `invalid-type-form` diagnostic and another `invalid-base` diagnostic:
#
# error: [invalid-type-form] "Int literals are not allowed in this context in a type expression"
Foo = NewType("Foo", 42)
```
````

File: crates/ty_python_semantic/resources/mdtest/annotations/new_types.md [hunk 32]
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+
+Normal classes can't inherit from newtypes, but generic classes can be parametrized with them, so we
+also need to detect "ordinary" type cycles that happen to involve a newtype.
+
+```py
+E = NewType("E", list["E"])
+reveal_type(E)  # revealed: <NewType pseudo-class 'E'>
+e: E = E([])
+reveal_type(e)  # revealed: E
```
AlexWaygood, line 237: We could add some more stress tests here to show that we really do understand the recursive nature of the newtype (which is really cool!)

````suggestion
Normal classes can't inherit from newtypes, but generic classes can be parametrized with them, so we
also need to detect "ordinary" type cycles that happen to involve a newtype.

```py
E = NewType("E", list["E"])
reveal_type(E)  # revealed: <NewType pseudo-class 'E'>
e: E = E([])
reveal_type(e)  # revealed: E
reveal_type(E(E(E(E(E([]))))))  # revealed: E
reveal_type(E([E([E([]), E([E([])])]), E([])]))  # revealed: E
E(["foo"])  # error: [invalid-argument-type]
E(E(E(["foo"])))  # error: [invalid-argument-type]
````

File: crates/ty_python_semantic/resources/mdtest/annotations/new_types.md [hunk 33]
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+## `NewType` wrapping preserves singleton-ness and single-valued-ness
+
+```py
+from typing_extensions import NewType
+from ty_extensions import is_singleton, is_single_valued, static_assert
+from types import EllipsisType
+
+A = NewType("A", EllipsisType)
+static_assert(is_singleton(A))
+static_assert(is_single_valued(A))
```
AlexWaygood, line 253: and these also reveal the correct result, which is pretty cool!

```suggestion
A = NewType("A", EllipsisType)
static_assert(is_singleton(A))
static_assert(is_single_valued(A))
reveal_type(type(A(...)) is EllipsisType)  # revealed: Literal[True]
reveal_type(A(...) is ...)  # revealed: Literal[True]
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+
+B = NewType("B", int)
+static_assert(not is_singleton(B))
+static_assert(not is_single_valued(B))
+```
+
+## `NewType` tuples can be iterated/unpacked
```
AlexWaygood:
```suggestion
## `NewType`s of tuples can be iterated/unpacked
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,269 @@ X = GenericAlias(type, ())
...
+
+N = NewType("N", tuple[int, str])
+
+a, b = N((1, "foo"))
+
+reveal_type(a)  # revealed: int
+reveal_type(b)  # revealed: str
+```
+
+## `isinstance` of a `NewType` instance and its base class is inferred `True`
```
AlexWaygood:
```suggestion
## `isinstance` of a `NewType` instance and its base class is inferred as `Literal[True]`
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
AlexWaygood: I think we're also still missing a test that demonstrates our `NewType` behaviour on Python 3.9, too (https://github.com/astral-sh/ruff/pull/21157#discussion_r2484100928)

It's fine to leave support for 3.9 `NewType` out of this PR, I think (and possibly never add it, since Python 3.9 is end-of-life). But I'd love a test that just demonstrates what our behaviour is when you use `typing.NewType` on Python 3.9

AlexWaygood: [One thing the typing conformance suite points out](https://github.com/python/typing/blob/9f6d8ced7cd1c8d92687a4e9c96d7716452e471e/conformance/tests/aliases_newtype.py#L37-L65) is that we need to emit an error if you try to create a generic `NewType` -- e.g. we should emit an error on this snippet, but that's not currently implemented in this PR:

```py
from typing import NewType, TypeVar

T = TypeVar("T")
BadNewType2 = NewType("BadNewType2", list[T])
```

For this as well, I think it's fine to leave it out of this PR, but it would be great to add a test with a TODO comment saying that we should try to emit a diagnostic on this in the future

AlexWaygood: it would also be nice to add a test somewhere that demonstrates that we infer member access on the pseudo-class _itself_ correctly, e.g.

```py
from typing import NewType

N = NewType("N", int)
reveal_type(N.__supertype__)  # revealed: type | NewType
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4880,34 +4885,149 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
             }
         }
     }
 
+    // Infer the deferred base type of a NewType.
+    fn infer_newtype_assignment_deferred(&mut self, arguments: &ast::Arguments) {
+        match self.infer_type_expression(&arguments.args[1]) {
+            Type::NominalInstance(_) | Type::NewTypeInstance(_) => {}
+            // `Unknown` is likely to be the result of an unresolved import or a typo, which will
+            // already get a diagnostic, so don't pile on an extra diagnostic here.
```
AlexWaygood, line 5028: this causes us to emit an error if you try to use a `Protocol` class as a `NewType` base. [According to the spec](https://typing.python.org/en/latest/spec/protocol.html#newtype-and-type-aliases), this is correct! However, we currently emit quite a confusing diagnostic for it -- the diagnostic currently implies that we haven't passed a class object, but we have (the issue is that it's a `Protocol` class object, which means that the `self.infer_type_expression()` call has returned a `Type::ProtocolInstance()` variant rather than a `Type::NominalInstance()` variant).

```sh
~/dev/ruff (jack/newtype4)⚡ % cat foo.py                                        
from typing import NewType, Protocol

class Id(Protocol):
    code: int

UserId = NewType('UserId', Id)
~/dev/ruff (jack/newtype4)⚡ % cargo run -p ty check foo.py --python-version=3.13
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/ty check foo.py --python-version=3.13`
error[invalid-newtype]: invalid base for `typing.NewType`
 --> foo.py:6:28
  |
4 |     code: int
5 |
6 | UserId = NewType('UserId', Id)
  |                            ^^ type `Id`
  |
info: The base of a `NewType` must be a class type or another `NewType`.
info: rule `invalid-newtype` is enabled by default

Found 1 diagnostic
```


It would be great if we could say something a bit clearer here, that says directly that the issue is that it's a `Protocol` class, which isn't allowed.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4880,34 +4885,149 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         )))
     }
 
+    fn infer_newtype_expression(
+        &mut self,
+        target: &ast::Expr,
+        call_expr: &ast::ExprCall,
+        definition: Definition<'db>,
+    ) -> Type<'db> {
```
AlexWaygood, line 4905: for type variables we explicitly reject the assignment unless it's a simple `ast::Assign` node (`ast::AnnAssign` nodes are rejected). I think we should probably do the same for `NewType` definitions:

```py
from typing import TypeVar, NewType

T: TypeVar = TypeVar("T")  # error[invalid-legacy-type-variable] "A `TypeVar` definition must be a simple variable assignment"
N: NewType = NewType("N", int)  # no error here currently

File: crates/ty_python_semantic/src/types/infer/builder.rs (hunk 27 above)
Comment (AlexWaygood) [reply crates/ty_python_semantic/src/types/infer/builder.rs:5006]: this looks correct to me. I think the issue is that type variables can have an arbitrary number of positional arguments when they're constructed, but a `NewType` can only have 2. I don't think it needs a comment!

File: crates/ty_python_semantic/resources/mdtest/annotations/new_types.md (hunk 33 above)
Comment (oconnor663) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: Hmm, I'm getting a `Revealed type: bool` failure for the last line there:

```
  crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:247 unmatched assertion: revealed: Literal[True]
  crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:247 unexpected error: 13 [revealed-type] "Revealed type: `bool`"
```

Did you see something different? Is that unexpected?

Comment (oconnor663) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: Digging into this, looks like `is_equivalent_to` is returning false in this case. Probably a bug worth fixing here...

Edit: Ooo the `match` at the bottom of that has a default arm, so my compiler-error-driven-development approach never saw it. Makes me worry that there might be a lot more cases like that.

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: The two types aren't equivalent (one's a subtype of the other), so I think that's correct

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: > Hmm, I'm getting a `Revealed type: bool` failure for the last line there:

I could've sworn I tried this out on your branch the other day and got `Literal[True]` 😆 but I don't think it matters too much either way; this is very much an edge case. The main thing I'd like is a test that records what our current behaviour is here, so we can see if it changes in the future.

Comment (oconnor663) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: Err, hmm, I was kind of confusing `is_equivalent_to` ("these types are the exact same set of values") with `is_single_valued` ("all values compare equal"). A newtype isn't equivalent to any other newtype besides itself (which I should still fix probably?). Maybe `infer_binary_type_comparison` needs to be tweaked. Thinking about this...

Comment (AlexWaygood) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: > A newtype isn't equivalent to any other newtype besides itself (which I should still fix probably?).

Again, I think that's arguably correct as it currently is. I wouldn't spend too much time looking at this; this all really feels very much like an extreme edge case. I'm sort-of regretting bringing it up at this point 😄

Comment (oconnor663) [reply crates/ty_python_semantic/resources/mdtest/annotations/new_types.md:253]: Ok I'll make the second one a TODO.

File: crates/ty_python_semantic/src/types/infer/builder.rs (hunk 37 above)
Comment (oconnor663) [reply crates/ty_python_semantic/src/types/infer/builder.rs:4905]: Ah neat, as part of copy/pasting `TypeVar`'s handling of this, I noticed that there was a dead special case for `NewType` in the call-binding machinery, added in an earlier version of this PR. That special case was preventing control from reaching the new error. Removing.

File: crates/ty_python_semantic/resources/mdtest/annotations/new_types.md [hunk 38]
```diff
@@ -12,13 +10,368 @@ X = GenericAlias(type, ())
...
+            reveal_type(x)  # revealed: N
+        case _:
+            reveal_type(x)  # revealed: N
+```
+
+## We don't support `NewType` on Python 3.9
+
+We implement `typing.NewType` as a `KnownClass`, but in Python 3.9 it's actually a function, so all
+we get is the `Any` annotations from typeshed. However, `typing_extensions.NewType` is always a
+class.
```
AlexWaygood:
```suggestion
We implement `typing.NewType` as a `KnownClass`, but in Python 3.9 it's actually a function, so all
we get is the `Any` annotations from typeshed. However, `typing_extensions.NewType` is always a
class.

This could be improved in the future, but Python 3.9 is now end-of-life, so it's not high-priority.
```

crates/ty_python_semantic/resources/mdtest/snapshots/new_types.md_-_NewType_-_The_base_of_a_`NewTy…_(8e2d1e1f951b7fc4).snap
```diff
@@ -0,0 +1,37 @@
...
+  |
+4 |     code: int
+5 |
+6 | UserId = NewType('UserId', Id)  # error: [invalid-newtype]
+  |                            ^^ type `Id`
+  |
+info: The base of a `NewType` is not allowed to be a protocol class.
+info: rule `invalid-newtype` is enabled by default
+
+```
```
AlexWaygood: great diagnostic!

AlexWaygood (approved):
https://github.com/astral-sh/ruff/pull/21157#discussion_r2504575209 is the only remaining significant thing I think that's missing from our `NewType` implementation now! And it's absolutely fine to defer that to a followup IMO

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
carljm, line 1: This test suite is fantastic. So thorough, so clear, all the prose is on-point. Thank you!!

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4892,34 +4897,157 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                &self.context,
+                format_args!(
+                    "The name of a `NewType` (`{name}`) must match \
+                    the name of the variable it is assigned to (`{target_name}`)"
+                ),
+                target,
+            );
+        }
+
+        // Inference of `tp` must be deferred, to avoid cycles.
```
carljm, line 4979: Not clear what the name `tp` refers to here?
```suggestion
        // Inference of the base type must be deferred, to avoid cycles.
```

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -12,13 +10,389 @@ X = GenericAlias(type, ())
...
+from typing import Any, NewType, TypeVar
+
+# All of these are allowed.
+A = NewType("A", list)
+B = NewType("B", list[int])
+B = NewType("B", list[Any])
+
+# But a free typevar is not allowed.
+T = TypeVar("T")
+C = NewType("C", list[T])  # TODO: should be "error: [invalid-newtype]"
```
oconnor663, line 397: @AlexWaygood do we like to open issues for these, or is the comment itself sufficient?

AlexWaygood: This is something mandated by the spec, so it would be great to open an issue for it. I don't see it as particularly high priority right now, though!

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
oconnor663: A *lot* of those cases are copy/pasted from @AlexWaygood's suggestions, to be clear :sweat_smile: Obviously all the bits that are like

`from ty_extensions import is_bivariant_endofunctor`

AlexWaygood: No no, bivariant endofunctors are your assignment for the GA release

crates/ty_python_semantic/src/types/infer/builder.rs
oconnor663: Ah, the names of `NewType`'s arguments are `name` and `tp`. I guess I spent enough time swimming in this that I forgot how non-obvious that is. `NewType` technically supports keyword args, though I didn't include that in this PR, and given all the "only simple assignments are allowed" behavior here and in other typecheckers, I'm not sure whether it would be a good idea to use them...

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
oconnor663: https://github.com/astral-sh/ty/issues/1519

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: I ended up supporting keyword args for `TypeVar`. I think it's totally fine to base our decision for `NewType` on whether we ever get a user report. (For reference, it looks like pyright does support keyword arguments with `NewType`, but neither mypy nor pyrefly does.)

PR #21159, opened 2025-10-31
[ty] Fix generic inference for non-dataclass inheriting from generic dataclass

saada:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1427

This PR fixes a regression introduced in alpha.24 where non-dataclass children of generic dataclasses lost generic type parameter information during `__init__` synthesis.

The issue occurred because when looking up inherited members in the MRO, the child class's `inherited_generic_context` was correctly passed down, but `own_synthesized_member()` (which synthesizes dataclass `__init__` methods) didn't accept this parameter. It only used `self.inherited_generic_context(db)`, which returned the parent's context instead of the child's.

The fix threads the child's generic context through to the synthesis logic, allowing proper generic type inference for inherited dataclass constructors.

## Test Plan

- Added regression test for non-dataclass inheriting from generic dataclass
- Verified the exact repro case from the issue now works
- All 277 mdtest tests passing
- Clippy clean
- Manually verified with Python runtime, mypy, and pyright - all accept this code pattern

## Verification

Tested against multiple type checkers:
- ✅ Python runtime: Code works correctly
- ✅ mypy: No issues found
- ✅ pyright: 0 errors, 0 warnings
- ✅ ty alpha.23: Worked (before regression)
- ❌ ty alpha.24: Regression
- ✅ ty with this fix: Works correctly

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
@@ -858,15 +858,13 @@ class ChildOfParentDataclass[T](ParentDataclass[T]): ...
 def uses_dataclass[T](x: T) -> ChildOfParentDataclass[T]:
     return ChildOfParentDataclass(x)
 
-# Verify the __init__ signatures
 # TODO: ParentDataclass.__init__ should show generic types, not Unknown
-reveal_type(ParentDataclass.__init__)  # revealed: (self: ParentDataclass[Unknown], value: Unknown) -> None
-# TODO: self type should be ChildOfParentDataclass, not ParentDataclass
-reveal_type(
-    ChildOfParentDataclass.__init__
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -838,6 +838,40 @@ class WrappedIntAndExtraData[T](Wrap[int]):
...
+class ParentDataclass[T]:
+    value: T
+
+# Non-dataclass inheriting from generic dataclass
+class ChildOfParentDataclass[T](ParentDataclass[T]): ...
+
+def uses_dataclass[T](x: T) -> ChildOfParentDataclass[T]:
+    return ChildOfParentDataclass(x)
+
+# TODO: ParentDataclass.__init__ should show generic types, not Unknown
```
sharkdp, line 861: Thank you for highlighting this. I opened https://github.com/astral-sh/ty/issues/1461 to track this.

sharkdp (approved):
Thank you!

PR #21160, opened 2025-10-31
Update etcetera to 0.11.0

musicinmybrain:
## Summary

This just bumps `etcetera` to the latest version. This has the pleasant effect of removing the `home` crate from the dependency tree. See: https://github.com/lunacookies/etcetera/releases/tag/v0.11.0

## Test Plan

Tested as a patch for the `ruff` package in Fedora with a proposed `rust-etcetera-0.11.0` update in [COPR](https://copr.fedorainfracloud.org/coprs/music/etcetera/packages/). (Normally I would just compile this locally from a git checkout and run the tests, but my machine with enough memory to compile ruff without workarounds is currently not available.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,7 +243,7 @@ dependencies = [
  "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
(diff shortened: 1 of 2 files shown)


MichaReiser:
Thank you

Issue #21099, opened 2025-10-27
`ruff analyze graph` seems to skip notebooks


gwdekker:
### Summary

It seems like `ruff analyze graph` only includes .py files but skips notebooks (.ipynb files) while plain ruff nowadays supports notebooks (https://docs.astral.sh/ruff/configuration/#jupyter-notebook-discovery).

MichaReiser:
I've to jump into my next meeting so I can't try it right now but does ruff not list notebook files or are they listed but without any imports?

gwdekker:
They are not listed at all.

MichaReiser:
I was surprised this behavior because analyze uses the same logic to discover all python files as all other commands but it then also does this:

https://github.com/astral-sh/ruff/blob/d9cab4d242a18102617d2a9eea1fc10a70c67a50/crates/ruff/src/commands/analyze_graph.rs#L130-L133

I'm not sure why it's there. Do you want to try to remove it and see if it just works?

It seems reasonable to have a way to exclude notebooks for people who don't care about them but that's what `analyze.exclude` can be used for.

gwdekker:
I tried. As a result after commenting out the lines and running `cargo run -p ruff -- analyze graph --python ...`
shows that now the notebook files show up as keys in the resulting mapping - however, the values are all empty lists. So I think a bit more work is needed to make this actually work.

Disclaimer: this is the very first thing I ever did with rust and cargo so no guarantees i did everything right.

PR #21161, opened 2025-10-31
[ruff]: Make `ruff analyze graph` work with jupyter notebooks

gauthsvenkat:
## Summary
 `ruff analyze graph` skips jupyter notebooks. This MR addresses that issue so that jupyter notebooks are also considered in the output. 

Closes: #21099 

### Detailed Summary
`ruff analyze graph` has the following lines to skip if the source is a jupyter notebook.

https://github.com/astral-sh/ruff/blob/d9cab4d242a18102617d2a9eea1fc10a70c67a50/crates/ruff/src/commands/analyze_graph.rs#L130-L133

Removing these lines alone is not enough (see [this comment](https://github.com/astral-sh/ruff/issues/21099#issuecomment-3453135503) on the issue). The reason is because in `detect()` the file under consideration is read and parsed to be a `.py` file. 

https://github.com/astral-sh/ruff/blob/d9cab4d242a18102617d2a9eea1fc10a70c67a50/crates/ruff_graph/src/lib.rs#L25-L33

For notebooks, you'd expect this parsing to fail as a `.ipynb` and `.py` are different. However that is not the case as under the hood, contents of a jupyter notebook are JSON. What instead is happening is that the jupyter notebooks get parsed as  
a giant nested python dictionary.

### Fix

- `detect()` now takes two more arguments; `source`: the raw content and `source_type` to correctly parse both `.py` and `.ipynb` content.
- in `analyze_graph.rs` we extract the raw content correctly for both `.py` and `.ipynb`.

## Test Plan

All existing unit tests pass and added two new tests to make sure the jupyter notebook flow also works.

### DIFF
--- a/crates/ruff/src/commands/analyze_graph.rs
+++ b/crates/ruff/src/commands/analyze_graph.rs
@@ -148,7 +148,7 @@ pub(crate) fn analyze_graph(
                     let source_kind = match SourceKind::from_path(path.as_std_path(), source_type) {
                         Ok(Some(source_kind)) => source_kind,
                         Ok(None) => {
-                            debug!("Skipping non-Python notebook: {}", path);
+                            debug!("Skipping non-Python notebook: {path}");
                             return;
                         }
                         Err(err) => {


crates/ruff/src/commands/analyze_graph.rs
amyreese: Does this part still need to be removed?

gauthsvenkat: Damn, I'm dumb. Yeah I think it should be removed.

What's wild is it seems to work even with that code present. I'll take a look soon.


EDIT: Spoke too soon! I think I messed up the cherry picking or something. It is removed now

PR #21164, opened 2025-10-31
[ty] Do not promote literals in contravariant position

sharkdp:
## Summary

closes https://github.com/astral-sh/ty/issues/1463

## Test Plan

Regression tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/literal_promotion.md
+++ b/crates/ty_python_semantic/resources/mdtest/literal_promotion.md
@@ -1,0 +1,32 @@
+# Literal promotion
+
+There are certain places where we promote literals to their common supertype:
+
+```py
+reveal_type([1, 2, 3])  # revealed: list[Unknown | int]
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/literal_promotion.md
```diff
@@ -0,0 +1,32 @@
...
+```py
+from typing import Callable, Literal
+
+def in_negated_position(non_zero_number: int):
+    if non_zero_number == 0:
+        raise ValueError()
+
+    reveal_type(non_zero_number)  # revealed: int & ~Literal[0]
+
+    reveal_type([non_zero_number])  # revealed: list[Unknown | (int & ~Literal[0])]
```
carljm, line 21: I suspect that this should better be `list[Unknown | int]`. That is, if we are in a context where we are promoting literals, the contravariant version of "promotion" is just to eliminate the negative literal types. This will help with the same kinds of invariance problems that literal promotion is intended to help with in the first place.

PR #21167, opened 2025-10-31
[ty] Fix `is_disjoint_from` with `@final` classes

ibraheemdev:
## Summary

We currently perform a subtyping check instead of the intended subclass check (and the subtyping check is confusingly named `is_subclass_of`). This showed up in https://github.com/astral-sh/ruff/pull/21070.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/is_disjoint_from.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/is_disjoint_from.md
@@ -87,6 +87,31 @@ static_assert(is_disjoint_from(memoryview, Foo))
 static_assert(is_disjoint_from(type[memoryview], type[Foo]))
 ```
 
+## Specialized `@final` types
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+from typing import final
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/is_disjoint_from.md
```diff
@@ -87,6 +87,25 @@ static_assert(is_disjoint_from(memoryview, Foo))
...
+```py
+from typing import final
+from ty_extensions import static_assert, is_disjoint_from
+
+@final
+class Foo[T]:
+    def get(self) -> T:
+        raise NotImplementedError
+
+static_assert(not is_disjoint_from(Foo[int], Foo[str]))
```
AlexWaygood, line 112: `int` and `str` are disjoint bases, so these two actually _should_ be disjoint (we're missing a check somewhere else). So this would be a better test here:

```suggestion
class A: ...
class B: ...

static_assert(not is_disjoint_from(Foo[A], Foo[B]))
```

AlexWaygood (approved):
Thanks!

PR #21168, opened 2025-10-31
[ty] prefer declared type on invalid TypedDict creation

carljm:
## Summary

In general, when we have an invalid assignment (inferred assigned type is not assignable to declared type), we fall back to inferring the declared type, since the declared type is a more explicit declaration of the programmer's intent. This also maintains the invariant that our inferred type for a name is always assignable to the declared type for that same name. For example:

```py
x: str = 1
reveal_type(x)  # revealed: str
```

We weren't following this pattern for dictionary literals inferred (via type context) as a typed dictionary; if the literal was not valid for the annotated TypedDict type, we would just fall back to the normal inferred type of the dict literal, effectively ignoring the annotation, and resulting in inferred type not assignable to declared type.

## Test Plan

Added mdtest assertions.


AlexWaygood (approved):

carljm:
Conformance suite change is positive (we are supposed to error on that line).

Ecosystem change looks like we have a TypedDict type where we previously didn't.

PR #21172, opened 2025-10-31
[ty] Improve exhaustiveness analysis for type variables with bounds or constraints

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1456.

This PR backs out the changes made in https://github.com/astral-sh/ruff/pull/21130 and replaces them with a more generalised implementation of exhaustiveness checking for `match`es over type variables with upper bounds or constraints. The approach in #21130 works for type variables with an enum upper bound, and could have been adapted to work for type variables with a `bool` upper bound. But it would _not_ work for a type variable with a union upper bound, because of the fact that we distribute over a union (to create a union of intersections) when adding a union type to an intersection.

Instead, the approach this PR takes is to speculatively solve all type variables in the positive part of an intersection to their upper bound/constraints when finally building the intersection. If that speculative intersection simplifies to `Never`, so must the original intersection.

## Test Plan

New mdtests added.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/exhaustiveness_checking.md
+++ b/crates/ty_python_semantic/resources/mdtest/exhaustiveness_checking.md
@@ -417,3 +417,55 @@ class Answer(Enum):
             case Answer.NO:
                 return False
 ```
+
+## Exhaustiveness checking for type variables with bounds or constraints
+
+```toml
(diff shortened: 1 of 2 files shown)


carljm (approved):
Very nice!

PR #21173, opened 2025-10-31
[ty] introduce local variables for `from` imports of submodules in `__init__.py(i)`

Gankra:
This rips out the previous implementation in favour of a new implementation with 3 rules:

- **froms are locals**: a `from..import` can only define locals, it does not have global
    side-effects. Specifically any submodule attribute `a` that's implicitly introduced by either
    `from .a import b` or `from . import a as b` (in an `__init__.py(i)`) is a local and not a
    global. If you do such an import at the top of a file you won't notice this. However if you do
    such an import in a function, that means it will only be function-scoped (so you'll need to do
    it in every function that wants to access it, making your code less sensitive to execution
    order).

- **first from first serve**: only the *first* `from..import` in an `__init__.py(i)` that imports a
    particular direct submodule of the current package introduces that submodule as a local.
    Subsequent imports of the submodule will not introduce that local. This reflects the fact that
    in actual python only the first import of a submodule (in the entire execution of the program)
    introduces it as an attribute of the package. By "first" we mean "the first time in this scope
    (or any parent scope)". This pairs well with the fact that we are specifically introducing a
    local (as long as you don't accidentally shadow or overwrite the local).

- **dot re-exports**: `from . import a` in an `__init__.pyi` is considered a re-export of `a`
    (equivalent to `from . import a as a`). This is required to properly handle many stubs in the
    wild. Currently it must be *exactly* `from . import ...`.
    
This implementation is intentionally limited/conservative (notably, often requiring a from import to be relative). I'm going to file a ton of followups for improvements so that their impact can be evaluated separately.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
@@ -26,9 +26,14 @@ This file currently covers the following details:
     (equivalent to `from . import a as a`). This is required to properly handle many stubs in the
     wild. Currently it must be *exactly* `from . import ...`.
 
+Note: almost all tests in here have a stub and non-stub version, because we're interested in both
+defining symbols *at all* and re-exporting them.
+
 ## Relative `from` Import of Direct Submodule in `__init__`
 
 We consider the `from . import submodule` idiom in an `__init__.pyi` an explicit re-export.
+
+### In Stub
 
 `mypackage/__init__.pyi`:
 
@@ -58,7 +63,7 @@ reveal_type(mypackage.imported.X)  # revealed: int
 reveal_type(mypackage.fails.Y)  # revealed: Unknown
 ```
 
-## Relative `from` Import of Direct Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -94,6 +99,8 @@ If an absolute `from...import` happens to import a submodule (i.e. it's equivale
 `from . import y`) we do not treat it as a re-export. We could, but we don't. (This is an arbitrary
 decision and can be changed!)
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -124,7 +131,7 @@ reveal_type(mypackage.imported.X)  # revealed: Unknown
 reveal_type(mypackage.fails.Y)  # revealed: Unknown
 ```
 
-## Absolute `from` Import of Direct Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -159,6 +166,8 @@ reveal_type(mypackage.fails.Y)  # revealed: Unknown
 An `import` that happens to import a submodule does not expose the submodule as an attribute. (This
 is an arbitrary decision and can be changed!)
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -181,7 +190,7 @@ import mypackage
 reveal_type(mypackage.imported.X)  # revealed: Unknown
 ```
 
-## Import of Direct Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -209,6 +218,8 @@ reveal_type(mypackage.imported.X)  # revealed: Unknown
 
 `from .submodule import nested` in an `__init__.pyi` does not re-export `mypackage.submodule`,
 `mypackage.submodule.nested`, or `nested`.
+
+### In Stub
 
 `mypackage/__init__.pyi`:
 
@@ -244,7 +255,7 @@ reveal_type(mypackage.nested)  # revealed: Unknown
 reveal_type(mypackage.nested.X)  # revealed: Unknown
 ```
 
-## Relative `from` Import of Nested Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `from .submodule import nested` in an `__init__.py` exposes `mypackage.submodule` and `nested`.
 
@@ -285,6 +296,8 @@ reveal_type(mypackage.nested.X)  # revealed: int
 `from mypackage.submodule import nested` in an `__init__.pyi` does not re-export
 `mypackage.submodule`, `mypackage.submodule.nested`, or `nested`.
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -320,7 +333,7 @@ reveal_type(mypackage.nested)  # revealed: Unknown
 reveal_type(mypackage.nested.X)  # revealed: Unknown
 ```
 
-## Absolute `from` Import of Nested Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `from mypackage.submodule import nested` in an `__init__.py` only creates `nested`.
 
@@ -362,6 +375,8 @@ reveal_type(mypackage.nested.X)  # revealed: int
 `import mypackage.submodule.nested` in an `__init__.pyi` does not re-export `mypackage.submodule` or
 `mypackage.submodule.nested`.
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -392,7 +407,7 @@ reveal_type(mypackage.submodule.nested)  # revealed: Unknown
 reveal_type(mypackage.submodule.nested.X)  # revealed: Unknown
 ```
 
-## Import of Nested Submodule in `__init__` (Non-Stub Check)
+### In Non-Stub
 
 `import mypackage.submodule.nested` in an `__init__.py` does not define `mypackage.submodule` or
 `mypackage.submodule.nested` outside the package.
@@ -432,6 +447,8 @@ reveal_type(mypackage.submodule.nested.X)  # revealed: Unknown
 
 Renaming the submodule to something else disables the `__init__.pyi` idiom.
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -455,7 +472,7 @@ reveal_type(mypackage.imported.X)  # revealed: Unknown
 reveal_type(mypackage.imported_m.X)  # revealed: Unknown
 ```
 
-## Relative `from` Import of Direct Submodule in `__init__`, Mismatched Alias (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -485,6 +502,8 @@ reveal_type(mypackage.imported_m.X)  # revealed: int
 The `__init__.pyi` idiom should definitely always work if the submodule is renamed to itself, as
 this is the re-export idiom.
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -505,7 +524,7 @@ import mypackage
 reveal_type(mypackage.imported.X)  # revealed: int
 ```
 
-## Relative `from` Import of Direct Submodule in `__init__`, Matched Alias (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -532,6 +551,8 @@ reveal_type(mypackage.imported.X)  # revealed: int
 Even if the `__init__` idiom is in effect, star imports do not pick it up. (This is an arbitrary
 decision that mostly fell out of the implementation details and can be changed!)
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -556,7 +577,7 @@ reveal_type(imported.X)  # revealed: Unknown
 reveal_type(Z)  # revealed: int
 ```
 
-## Star Import Unaffected (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -586,6 +607,8 @@ reveal_type(Z)  # revealed: int
 A `from` import that imports a non-submodule isn't currently a special case here (various
 proposed/tested approaches did treat this specially).
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -607,7 +630,7 @@ import mypackage
 reveal_type(mypackage.imported.X)  # revealed: Unknown
 ```
 
-## `from` Import of Non-Submodule (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -634,6 +657,8 @@ reveal_type(mypackage.imported.X)  # revealed: int
 `from mypackage import submodule` from outside the package is not modeled as a side-effect on
 `mypackage`, even in the importing file (this could be changed!).
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -658,7 +683,7 @@ reveal_type(imported.X)  # revealed: int
 reveal_type(mypackage.imported.X)  # revealed: Unknown
 ```
 
-## `from` Import of Other Package's Submodule (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -687,6 +712,8 @@ reveal_type(mypackage.imported.X)  # revealed: Unknown
 
 `from . import submodule` from a sibling module is not modeled as a side-effect on `mypackage` or a
 re-export from `submodule`.
+
+### In Stub
 
 `mypackage/__init__.pyi`:
 
@@ -719,7 +746,7 @@ reveal_type(imported.fails.Y)  # revealed: Unknown
 reveal_type(mypackage.fails.Y)  # revealed: Unknown
 ```
 
-## `from` Import of Sibling Module (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
@@ -768,6 +795,8 @@ We avoid this by ensuring that the imported name (the right-hand `funcmod` in
 `from .funcmod import funcmod`) overwrites the submodule attribute (the left-hand `funcmod`), as it
 does at runtime.
 
+### In Stub
+
 `mypackage/__init__.pyi`:
 
 ```pyi
@@ -800,7 +829,7 @@ from mypackage import funcmod
 x = funcmod(1)
 ```
 
-## Fractal Re-export Nameclash Problems (Non-Stub Check)
+### In Non-Stub
 
 `mypackage/__init__.py`:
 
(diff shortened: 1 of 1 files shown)


Gankra:
Preserving for future comparison: this is the "submodule-attributes" approach's results. I am now pushing up a different WIP implementation.

Screenshot 2025-11-03 at 2 41 32 PM

sharkdp:
> Preserving for future comparison

Just FYI, you can also access the full history here, I do that quite often after pushing a change.

image

Gankra:
> Just FYI, you can also access the full history here, I do that quite often after pushing a change.

Good callout, but I've done so many ecosystem runs on this PR to evaluate different approaches it would be very easy to lose track of which run is which.

Gankra:
I have cleaned up the implementation and added some more tests. I've made a few edge-cases a bit more strict. I am hoping this is a complete implementation, except for some code/mdtest comments need to be updated to reflect the current semantics.

For posterity, this is the results that we agreed seemed shippable *before* my cleanup. Hopefully the stricter implementation isn't a huge change!

Screenshot 2025-11-04 at 3 41 00 PM

Gankra:
I believe the TL;DR of this PR is now:

* I have ripped out my previous work with submodule-attributes
* I have introduced a re-export rule that treats `from . import x` in `__init__.pyi` as if it was `x as x` (must be *only* from `.`, cannot be `.y`)
* I have introduced a new `ImportFromImplicitDefinitionKind` that expresses the implicit `x = <thispackage.x>` that results from the first `from .x.y import z` in an `__init__.py(i)` 
  * This value is *never* re-exported, as there wasn't a good way to "turn it off" (you could do an absolute import in the current implementation), and you can just do `from . import x` if you wanted to re-export it.
  * Because it's not a submodule-attribute, it essentially introduces a local variable, which means it implicitly models semantics that imports in a function should probably only be visible in that function, safety-wise. While top-level imports would be visible to all functions.
  * Because it's not a submodule-attribute, it can be overwritten and we won't complain (`from .myfunc import myfunc`, `from .myfunc import other; myfunc = 0`)
  * Only the first import of a particular direct-submodule name will create an assignment
    * Known bug I'm shipping and will file an issue for: the "first" check is file-global and not per-scope, which means we get confused if e.g. two functions do the same import (only the first will see the symbol). 


As a reminder, "re-exported" is only an applicable concern for definitions in .pyi files, so both kinds of definitions will be fully visible in .py files.

Gankra:
See https://github.com/astral-sh/ty/issues/133#issuecomment-3492576578 for a list of followups that I am intentionally punting on in this PR (or the sea of references right above this comment...)

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
-    at runtime)
-- Extensions should ideally not make `.pyi` files more permissive than `.py` files (if it works in a
-    stub, it works in an impl)
-
-To that end we define the following extension:
-
-> If an `__init__.pyi` for `mypackage` contains a `from...import` targetting a direct submodule of
-> `mypackage`, then that submodule should be available as an attribute of `mypackage`.
+[standard import conventions](https://typing.python.org/en/latest/spec/distributing.html#import-conventions),
+and other intentional deviations from actual python semantics.
```
carljm, line 5: The current text below buries (or eliminates) the lede: we dive straight into the edgiest of edge cases (like submodule imports inside a function) without any discussion of the general/common cases and how we choose to model these "submodule attributes" in general.

I think it probably makes most sense to do an update here separately, after a few more PRs in this area land. I'm happy to do the prose update myself since I have opinions about how we present it.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
+
+This file currently covers the following details:
+
+- **froms are locals**: a `from..import` can only define locals, it does not have global
+    side-effects. Specifically any submodule attribute `a` that's implicitly introduced by either
+    `from .a import b` or `from . import a as b` (in an `__init__.py(i)`) is a local and not a
+    global. If you do such an import at the top of a file you won't notice this. However if you do
+    such an import in a function, that means it will only be function-scoped (so you'll need to do
+    it in every function that wants to access it, making your code less sensitive to execution
+    order).
```
carljm, line 15: I assume that in the upcoming PR you mentioned in person, this whole paragraph goes away to be replaced by something that says "we only model submodule imports adding the submodule attribute for global-scope submodule imports"

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
+
+- **first from first serve**: only the *first* `from..import` in an `__init__.py(i)` that imports a
+    particular direct submodule of the current package introduces that submodule as a local.
+    Subsequent imports of the submodule will not introduce that local. This reflects the fact that
+    in actual python only the first import of a submodule (in the entire execution of the program)
+    introduces it as an attribute of the package. By "first" we mean "the first time in this scope
+    (or any parent scope)". This pairs well with the fact that we are specifically introducing a
+    local (as long as you don't accidentally shadow or overwrite the local).
```
carljm, line 23: I like this paragraph, very clear. (Last sentence I also expect to go away in a future PR.)

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
+
+- **dot re-exports**: `from . import a` in an `__init__.pyi` is considered a re-export of `a`
+    (equivalent to `from . import a as a`). This is required to properly handle many stubs in the
+    wild. Currently it must be *exactly* `from . import ...`.
```
carljm, line 27: And it sounds like you have a PR in the works that will also eliminate the last sentence here?

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
 
 ## Relative `from` Import of Direct Submodule in `__init__`
 
-The `from . import submodule` idiom in an `__init__.pyi` is fairly explicit and we should definitely
-support it.
+The `from . import submodule` idiom in an `__init__.pyi` should definitely be considered an explicit
+re-export.
```
carljm: Nit: "should definitely be" seems like quite strong language here, given that all of the stub-file rules are kind of arbitrary (there's no runtime behavior to model it on) and we're just trying to find the right balance between "what the spec explicitly says" and "what stub files in the wild rely on." I'd give this a more descriptive tone: "We consider ... to be an explicit re-export."

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -1,39 +1,35 @@
...
+    in actual python only the first import of a submodule (in the entire execution of the program)
+    introduces it as an attribute of the package. By "first" we mean "the first time in this scope
+    (or any parent scope)". This pairs well with the fact that we are specifically introducing a
+    local (as long as you don't accidentally shadow or overwrite the local).
+
+- **dot re-exports**: `from . import a` in an `__init__.pyi` is considered a re-export of `a`
+    (equivalent to `from . import a as a`). This is required to properly handle many stubs in the
+    wild. Currently it must be *exactly* `from . import ...`.
 
 ## Relative `from` Import of Direct Submodule in `__init__`
```
carljm, line 32: Nit: the fact that all test headers / names in this test file assume "in a stub file" by default (and then suffix "(Non-stub Check)" to a subsequent equivalent-except-not-in-stub test) may have made sense before this PR, when this file was really only about tweaking stub file rules, but it feels weird after this PR, now that this file is really about submodule attributes in general, stub or non-stub.

I would just have a single test heading describing each case, and then use sub-headings `### In py file` and `### In stub file` under that, rather than repeating the whole test description.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
carljm, line 679: I would find it useful to future-us to document here what we mean by "dangerous", and why it's dangerous

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -752,9 +764,8 @@ Can easily result in the typechecker getting "confused" and thinking imports of
 top-level package are referring to the subpackage and not the function/class. This issue can be
 found with the `lobpcg` function in `scipy.sparse.linalg`.
 
-This kind of failure mode is why the rule is restricted to *direct* submodule imports, as anything
-more powerful than that in the current implementation strategy quickly gets the functions and
-submodules mixed up.
+We avoid this by ~desugarring the imports into two definitions, with the RHS occuring second, so
+that it can freely overwrite the LHS.
```
carljm:
```suggestion
We avoid this by ensuring that the imported name (the right-hand `funcmod` in
`from .funcmod import funcmod`) overwrites the submodule attribute (the left-hand
`funcmod`), as it does at runtime.
```

crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
```diff
@@ -607,23 +607,33 @@ module:
 `module2.py`:
 
 ```py
-import importlib
-import importlib.abc
+import imported
+import imported.abc
+```
+
+`imported/__init__.pyi`:
```
carljm, line 614: I think this is a good change, in the interest of simpler and more self-contained tests, but I assume (hope?) the previous version of this test (using importlib) still passes after this PR? It's just that the simpler version was easier to debug in development?

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1453,6 +1449,43 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
+                // * And this is the first time we've seen `from .x` in this module
+                //
+                // We introduce a local definition `x = <module 'thispackage.x'>` that occurs
+                // before the `z = ...` declaration the import introduces. This models the fact
+                // that the *first* time that you import 'thispackage.x' the python runtime creates
+                // `x` as a variable in the global scope of `thispackage`.
+                //
+                // This is not a perfect simulation of actual runtime behaviour for *various*
+                // reasons but it works well for most practical purposes. In particular it's nice
+                // that `x` can be freely overwritten, and that we don't assume that an import
```
carljm, line 1465: Great explanation, very helpful.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1453,6 +1449,43 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
+                    && !self.seen_submodule_imports.contains(direct_submodule)
+                {
+                    self.seen_submodule_imports
+                        .insert(direct_submodule.to_owned());
+
+                    let direct_submodule_name = Name::new(direct_submodule);
+                    let symbol = self.add_symbol(direct_submodule_name);
+                    self.add_definition(
+                        symbol.into(),
+                        ImportFromImplicitDefinitionNodeRef { node, submodule },
```
carljm: I don't know that it matters, since the from-import node itself is otherwise unused as a definition NodeKey, and there can only be one implicit definition per from-import, but I'm curious whether we could attach this definition to the Identifier for the submodule itself rather than to the entire import-from node? It seems slightly more intuitive/clearer. Maybe we can't do this because `Identifier` isn't a real AST node? (I don't recall...)

crates/ty_python_semantic/src/semantic_index/definition.rs
```diff
@@ -209,6 +210,7 @@ impl<'db> DefinitionState<'db> {
 pub(crate) enum DefinitionNodeRef<'ast, 'db> {
     Import(ImportDefinitionNodeRef<'ast>),
     ImportFrom(ImportFromDefinitionNodeRef<'ast>),
+    ImportFromImplicit(ImportFromImplicitDefinitionNodeRef<'ast>),
```
carljm: nit: I think I would slightly prefer if this (and all the related stuff) used the name `ImportFromSubmodule` rather than `ImportFromImplicit`? This is always used for "implicit attachment of submodule as attribute to parent package" -- of all the words in that verbose description, I think "submodule", more than "implicit", gestures in a clarifying direction.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5475,6 +5483,99 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        // By introducing it as a local variable in the scope the import occurs in,
+        // we effectively require the developer to either do the import at the start of
+        // the file where it belongs, or to repeat the import in every function that
+        // wants to use it, which "definitely" works.
+        //
+        // (It doesn't actually "definitely" work because only the first import of `thispackage.x`
+        // will ever set `x`, and any subsequent overwrites of it will permanently clobber it.
+        // Also, a local variable `x` in a function should always shadow the submodule because
+        // the submodule is defined at file-scope. However, both of these issues are much more
+        // narrow, so this approach seems to work well in practice!)
```
carljm, line 5509: As discussed in person, I don't think accepting these two bugs (submodule-attribute works only in first function, and can wrongly clobber a function local) is the right tradeoff, vs just not modeling submodule-attribute side-effects at all for imports inside functions. The latter is a clearly delineated and explainable choice of what to model and what not to model; the former two are confusing edge-case bugs where something sometimes works but then sometimes fails in surprising ways.

Assuming this text will be updated in upcoming PR!

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -822,3 +833,305 @@ from mypackage import funcmod
...
+```py
+def funcmod(x: int) -> int:
+    return x
+```
+
+## LHS `from` Imports In Functions (Non-Stub Check)
+
+If a `from` import occurs in a function, LHS symbols should only be visible in that function. This
+very blatantly is not runtime-accurate, but exists to try to force you to write "obviously
+deterministically correct" imports instead of relying on execution order.
```
carljm: This should change in an upcoming PR.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5475,6 +5483,99 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            ret
+        });
+        // And try to import it
+        if let Some(submodule_type) = full_submodule_name
+            .as_ref()
+            .and_then(|submodule_name| self.module_type_from_name(submodule_name))
+        {
+            // Success, introduce a binding!
+            //
+            // We explicitly don't introduce a *definition* because it's actual ok
```
carljm:
```suggestion
            // We explicitly don't introduce a *declaration* because it's actual ok
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5475,6 +5483,99 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            self.add_binding(import_from.into(), definition, |_, _| submodule_type);
+            return;
+        }
+
+        // That didn't work, try to produce diagnostics
+        self.add_binding(import_from.into(), definition, |_, _| Type::unknown());
+
+        if !self.is_reachable(import_from) {
+            return;
+        }
```
carljm, line 5556: I assume this was caught be ecosystem? Do we have any test covering this case?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5475,6 +5483,99 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+
+        let Some(builder) = self
+            .context
+            .report_lint(&UNRESOLVED_IMPORT, AnyNodeRef::StmtImportFrom(import_from))
+        else {
+            return;
+        };
+
+        let diagnostic = builder.into_diagnostic(format_args!(
+            "Module `{module_name}` has no member `{name}`"
```
carljm: Nit: we are trying to create the "member" here -- what we've failed to do is find the submodule that it should point to. So I would expect something more like:
```suggestion
            "Module `{module_name}` has no submodule `{name}`"
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -11029,29 +10982,23 @@ impl<'db> ModuleLiteralType<'db> {
...
     ///   *should* shadow any non-submodule of the same name.
     ///
-    /// * If the module is an `__init__.pyi` for `mypackage`, and it contains a `from...import`
-    ///   that normalizes to `from mypackage import submodule`, then `mypackage.submodule` is
-    ///   defined in all files. This supports the `from . import submodule` idiom. Critically,
-    ///   we do *not* allow `from mypackage.nested import submodule` to affect `mypackage`.
-    ///   The idea here is that `from mypackage import submodule` *from mypackage itself* can
-    ///   only ever reasonably be an import of a submodule. It doesn't make any sense to import
-    ///   a function or class from yourself! (You *can* do it but... why? Don't? Please?)
+    /// `import x.y from z` instances are currently ignored because the `x.y` part may not be a
```
```suggestion
    /// `from x.y import z` instances are currently ignored because the `x.y` part may not be a
```

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
Gankra: You're extremely right and I was hoping no one would notice. *rolls up sleeves*

crates/ty_python_semantic/resources/mdtest/type_properties/is_equivalent_to.md
Gankra: Different versions of this PR have variously broken and unbroken the test because of the line `from importlib.abc import Loader` in typeshed's `importlib/__init__.pyi`. I would rather check in this not-relying-on-a-random-library version.

crates/ty_python_semantic/src/semantic_index/builder.rs
Gankra: I was definitely under the impression that the `Identifier` couldn't be attached to, yeah.
(I agree, I would've if I thought I could've)

crates/ty_python_semantic/src/types/infer/builder.rs
Gankra: No this is copy-paste from `infer_from_import_definition`, I have no insights. I just assumed it was a common "don't do diagnostics about dead code" idiom.

carljm: Yeah that's exactly what it is, makes sense that it would just be copied over. Don't even feel strongly about having a test for it, but it seems like a good idea to avoid future regression?

Gankra: Followup it is!

PR #21175, opened 2025-10-31
[ty] Support for notebooks in VS Code

MichaReiser:
## Summary

This PR adds support for notebook synchronization to ty's LSP.
ty now supports LSP operations across LSP cell boundaries and no longer incorrectly
reports diagnostics when using variables that were defined in earlier cells. 

Most of the changes are integrating notebooks into ty's LSP range to text-range conversion. However, this PR also introduces a new `system.source_type` method to ensure the editor and ty agree on what's considered a notebook. 

Fixes https://github.com/astral-sh/ty/issues/173

## Known limitations

The LSP doesn't support cross-cell edits for completions. This is a problem for auto-imports because the importer will try to insert the import in the first cell, rather than the current cell. Given that auto import is still experimental, I suggest tackling this in a follow up PR.

Another limitation is that we can't show diagnostics for notebooks or references within notebooks unless they're open in the editor. This is because ty can't know the URI for a cell unless the notebook is open, which prevents us from mapping the diagnostic to a meaningful location. The best we could do is to map the location to the start of the notebook but this also seems bad (and notebooks don't support pull or workspace diagnostics...). For now, we match what pylance does and only handle open notebooks.

## Test Plan

Added E2E tests

https://github.com/user-attachments/assets/cbd9b873-a840-4e86-b280-d5d0d5a7d739

### DIFF
--- a/crates/ty_server/src/system.rs
+++ b/crates/ty_server/src/system.rs
@@ -324,7 +324,7 @@ fn document_revision(document: &Document, index: &Index) -> FileRevision {
 
             // Use higher 64 bits for notebook version and lower 64 bits for cell revisions
             let notebook_version_high = (notebook.version() as u128) << 64;
-            let cell_versions_low = (hasher.finish() as u128) & 0xFFFF_FFFF_FFFF_FFFF;
+            let cell_versions_low = u128::from(hasher.finish()) & 0xFFFF_FFFF_FFFF_FFFF;
             let combined_revision = notebook_version_high | cell_versions_low;
 
             FileRevision::new(combined_revision)


crates/ty_ide/src/semantic_tokens.rs
```diff
@@ -226,7 +226,12 @@ impl<'db> SemanticTokenVisitor<'db> {
...
         // Only emit tokens that intersect with the range filter, if one is specified
         if let Some(range_filter) = self.range_filter {
-            if range.intersect(range_filter).is_none() {
+            // Only include ranges that intersect. Ranges that only touch (are empty)
+            // should be excluded.
+            if range
+                .intersect(range_filter)
+                .is_none_or(ruff_text_size::TextRange::is_empty)
+            {
                 return;
```
MichaReiser, line 241: This fixes a bug where ty highlighted a token that started right at the end of the provided range. This resulted in a panic for notebooks because ty tried to highlight a token from the next cell, which isn't allowed

crates/ty_server/src/document/notebook.rs
```diff
@@ -334,64 +244,62 @@ mod tests {
...
-            "cell = 1
-cell = 0
-cell = 2
-"
-        );
-    }
+    //     /// This test case checks that for a notebook with three code cells, when the client sends a
+    //     /// change request to swap the first two cells, the notebook document is updated correctly.
+    //     ///
+    //     /// The swap operation as a change request is represented as deleting the first two cells and
```
MichaReiser: TODO, delete

crates/ty_server/src/document/range.rs
```diff
@@ -1,148 +1,383 @@
...
+        {
+            let notebook_index = notebook.index();
+
+            let start_in_concatenated =
+                index.source_location(self.range.start(), &source, self.encoding.into());
+            let cell_index = notebook_index.cell(start_in_concatenated.line);
+
+            let cell_range = notebook
+                .cell_range(cell_index.unwrap().to_zero_indexed())
+                .unwrap();
```
MichaReiser: TODO: Handle more gracefully

crates/ty_server/src/server/api/notifications/did_change_watched_files.rs
```diff
@@ -96,7 +96,6 @@ impl SyncNotificationHandler for DidChangeWatchedFiles {
                 publish_diagnostics(session, key.url(), client);
             }
         }
-        // TODO: always publish diagnostics for notebook files (since they don't use pull diagnostics)
```
MichaReiser, line 99: We can't publish diagnostics for notebooks that aren't open because we don't know the URLs of the cells. But VS code uses `didChangeNotebook` for open notebooks and not `didChangeWatchedFiles`. That's why this comment isn't relevant.

crates/ty_server/src/server/api/requests/diagnostic.rs
```diff
@@ -27,6 +27,24 @@ impl BackgroundDocumentRequestHandler for DocumentDiagnosticRequestHandler {
...
+    // a document that's supposed to be a notebook. Resulting in a mismatch between
+    // So how can we channel through the entire notebook? But also make this work
+    // If we kept the cells within notebook, would it then be easier?
+    // Maybe, we could snapshot the entire notebook and call ruff notebook on it.
+    // But that only works for the current notebook and not for **any** notebook.
+    // In the end, we need a way to retrieve the metadata from any notebook.
+    // A method on lsp system doesn't sound that horrible?
+    // what happens if we don't report the diagnostics on the cell?
+    // The issue with that is that a notebook has no rows?
+    // We can't import other notebooks, so related should only be limited to the current file.
```
MichaReiser: Delete

crates/ruff_db/src/source.rs
```diff
@@ -40,17 +41,19 @@ pub fn source_text(db: &dyn Db, file: File) -> SourceText {
...
 }
 
-fn is_notebook(path: &FilePath) -> bool {
-    match path {
-        FilePath::System(system) => system.extension().is_some_and(|extension| {
+fn is_notebook(system: &dyn System, path: &FilePath) -> bool {
+    let source_type = match path {
+        FilePath::System(path) => system.source_type(path),
+        FilePath::SystemVirtual(system_virtual) => system.virtual_path_source_type(system_virtual),
+        FilePath::Vendored(_) => return false,
```
Gankra, line 48: wow not very forward thinking, micha, what if typeshed adds stubs for the most popular notebooks in the python ecosystem!!!! /silly

crates/ruff_db/src/system.rs
```diff
@@ -66,6 +66,27 @@ pub trait System: Debug + Sync + Send {
...
 
+    /// Returns the source type for `path` if known or `None`.
+    ///
+    /// This is primarily used for the LSP integration to respect
+    /// the chosen language (or the fact that it is a notebook) in
+    /// the editor.
+    fn source_type(&self, path: &SystemPath) -> Option<PySourceType> {
+        let _ = path;
+        None
+    }
```
Gankra, line 81: It's probably worth adding a comment to these two functions explaining that we're defaulting to "system has no additional info" (and so the caller will defer to e.g. the file extension)

crates/ruff_notebook/src/notebook.rs
```diff
@@ -388,6 +388,21 @@ impl Notebook {
...
+    /// text document.
+    pub fn cell_offset(&self, cell: OneIndexed) -> Option<TextSize> {
+        self.cell_offsets.get(cell.to_zero_indexed()).copied()
+    }
+
+    /// Returns the text range in the concatenated document of the cell
+    /// with index `cell`.
+    pub fn cell_range(&self, cell: OneIndexed) -> Option<TextRange> {
+        let start = self.cell_offsets.get(cell.to_zero_indexed()).copied()?;
+        let end = self.cell_offsets.get(cell.to_zero_indexed() + 1).copied()?;
```
Gankra, line 401: Checking that this actually handles the last cell properly? We store a trailing offset?
(I assume this will be elucidated elsewhere...)

crates/ty_server/src/document/notebook.rs
```diff
@@ -1,23 +1,24 @@
...
+use crate::session::index::Index;
 
-pub(super) type CellId = usize;
-
-/// The state of a notebook document in the server. Contains an array of cells whose
-/// contents are internally represented by [`TextDocument`]s.
+/// A notebook document.
+///
+/// The document only stores the metadata about its cell.
+/// Each cell is stored as a [`super::TextDocument`].
```
Gankra: I can't understand what this is saying

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
...
-                                cell_text_document.version,
-                            );
-                        }
-                    }
-                }
-            }
+        self.cells.splice(
+            deleted_range.clone(),
+            new_cells.into_iter().map(NotebookCell::new),
+        );
```
Gankra, line 142: Obscure collections API being useful!!! 🎊 

(Had to actually check that `1..1` works the way you want, it does)

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
...
-            }
+        // Re-build the cell-index if new cells were added, deleted or removed
+        if !deleted_range.is_empty() || added > 0 {
+            self.cell_index.clear();
+            self.cell_index.extend(
+                self.cells
+                    .iter()
+                    .enumerate()
+                    .map(|(i, cell)| (cell.url.clone(), i)),
+            );
```
Gankra, line 152: Slightly unfortunate that this logic is duplicated but it's so trivial it's whatever

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
...
-                    if let Some(cell) = self.cell_by_uri_mut(content_change.document.uri.as_str()) {
-                        cell.document
-                            .apply_changes(content_change.changes, version, encoding);
-                    }
-                }
+        for cell in cells {
+            if let Some(existing_cell_index) = self.cell_index.get(&cell.document).copied() {
+                self.cells[existing_cell_index].kind = cell.kind;
             }
         }
```
Gankra: This code profoundly confused me (the key here is that `cells` isn't `self.cells` or `new_cells`). Either it should have a comment or some of these cells need more descriptive names.

crates/ty_server/src/document/notebook.rs
```diff
@@ -221,177 +162,30 @@ impl NotebookDocument {
...
-            document: TextDocument::new(cell.document.clone(), contents, version),
             url: cell.document,
             kind: cell.kind,
+            execution_summary: cell.execution_summary,
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
Gankra, line 281: I wonder if I'll ever find the new home for these tests... (or are they meaningless?)

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -17,15 +17,18 @@ use crate::document::{FileRangeExt, ToRangeExt};
 use crate::session::DocumentSnapshot;
 use crate::session::client::Client;
 use crate::system::{AnySystemPath, file_to_url};
-use crate::{NotebookDocument, PositionEncoding, Session};
+use crate::{PositionEncoding, Session};
 
-pub(super) struct Diagnostics<'a> {
+// THe real challenge here is that `to_lsp_range` now needs to
```
Gankra:
```suggestion
// The real challenge here is that `to_lsp_range` now needs to
```

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -17,15 +17,18 @@ use crate::document::{FileRangeExt, ToRangeExt};
...
+// map to the cell document instead of the document. This is tricky.
+
+pub(super) struct Diagnostics {
```
Gankra: ...also this is a really weird comment outside of the context of reviewing this PR. Was this a note to yourself during hacking?

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -151,7 +155,13 @@ pub(super) fn publish_diagnostics(session: &Session, url: &lsp_types::Url, clien
         }
     };
 
-    let db = session.project_db(&snapshot.to_file_path());
+    if !snapshot.document().is_cell_or_notebook()
+        && session.client_capabilities().supports_pull_diagnostics()
+    {
+        return;
+    }
```
Gankra: Might be good to add a comment explaining why notebooks need to bypass this

crates/ty_server/src/server/api/notifications/did_change_notebook.rs
```diff
@@ -0,0 +1,39 @@
+use lsp_server::ErrorCode;
+use lsp_types as types;
+use lsp_types::notification as notif;
+
+use crate::server::Result;
+use crate::server::api::LSPResult;
+use crate::server::api::notifications::did_change::file_changed;
```
Gankra: Slightly odd for notifications to yoink code from eachother like this

crates/ty_server/src/server/api/notifications/did_close.rs
```diff
@@ -31,55 +32,67 @@ impl SyncNotificationHandler for DidCloseTextDocumentHandler {
             .document_handle(&uri)
             .with_failure_code(ErrorCode::InternalError)?;
 
-        let path = document.to_file_path().into_owned();
+        if !document.is_cell() {
+            close_document(&document, session, client);
+        }
+
         let url = document.url().clone();
+        let is_cell = document.is_cell();
```
Gankra: Compute is_cell earlier so you can use it in the check above

crates/ty_server/src/server/api/notifications/did_close.rs
```diff
@@ -31,55 +32,67 @@ impl SyncNotificationHandler for DidCloseTextDocumentHandler {
...
 
         document
             .close(session)
             .with_failure_code(ErrorCode::InternalError)?;
 
-        let db = session.project_db_mut(&path);
+        if is_cell {
+            // Notebooks only support publish diagnostics.
```
Gankra:
```suggestion
            // Notebooks only support push diagnostics.
```

(Genuinely forget, but I am very sad if the terminology is really "publish" and "pull" and not "push" and "pull")

crates/ty_server/src/server/api/notifications/did_open.rs
```diff
@@ -39,45 +39,49 @@ impl SyncNotificationHandler for DidOpenTextDocumentHandler {
...
-                let event = if is_maybe_new_system_file {
-                    ChangeEvent::Created {
-                        path: system_path.clone(),
-                        kind: CreatedKind::File,
-                    }
-                } else {
-                    ChangeEvent::Opened(system_path.clone())
-                };
-                session.apply_changes(&path, vec![event]);
+pub(super) fn opened_document(document: &DocumentHandle, session: &mut Session, client: &Client) {
```
Gankra: `open_document`?

crates/ty_server/src/server/api/requests/semantic_tokens.rs
```diff
@@ -33,14 +36,26 @@ impl BackgroundDocumentRequestHandler for SemanticTokensRequestHandler {
...
+
+        if snapshot.document().is_cell() {
+            if let Some(notebook_document) = db.notebook_document(file)
+                && let Some(notebook) = source_text(db, file).as_notebook()
+            {
+                let cell_index = notebook_document.cell_index_by_uri(snapshot.url());
+
+                range = cell_index.and_then(|index| notebook.cell_range(index));
+            }
+        }
```
Gankra: This can be one big && chain instead of nested ifs

crates/ty_server/src/document/notebook.rs
```diff
@@ -33,7 +34,7 @@ struct NotebookCell {
     /// > <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#notebookDocument_synchronization>
     url: lsp_types::Url,
     kind: NotebookCellKind,
-    document: TextDocument,
```
dhruvmanila, line 36: Can you update the documentation for `NotebookCell` now that it doesn't contain the `TextDocument` in it?

crates/ty_server/src/document/notebook.rs
```diff
@@ -33,7 +34,7 @@ struct NotebookCell {
...
+    execution_summary: Option<lsp_types::ExecutionSummary>,
```
dhruvmanila, line 39: I'm curious to know why is this being added?

crates/ty_server/src/document/notebook.rs
```diff
@@ -79,29 +66,38 @@ impl NotebookDocument {
...
-                    ruff_notebook::Cell::Markdown(ruff_notebook::MarkdownCell {
-                        attachments: None,
+            .map(|cell| {
+                let source = ruff_notebook::SourceValue::String(
+                    index
+                        .document(&DocumentKey::from_url(&cell.url))
+                        .ok()
+                        .and_then(|document| document.as_text())
+                        .map(|text| text.contents().to_string())
+                        .unwrap_or_default(),
```
dhruvmanila: Would it be useful to add a debug log if the index doesn't contain the text document representing the cell?

crates/ty_server/src/document/text_document.rs
```diff
@@ -16,15 +16,16 @@ pub struct TextDocument {
 
     /// The string contents of the document.
     contents: String,
-    /// A computed line index for the document. This should always reflect
-    /// the current version of `contents`. Using a function like [`Self::modify`]
-    /// will re-calculate the line index automatically when the `contents` value is updated.
-    index: LineIndex,
```
dhruvmanila, line 22: IIUC, this is removed because the server will always need to re-compute the index whenever the content in a text document changes so there's no benefit of computing this ahead of time?

crates/ty_server/src/server/api/diagnostics.rs
dhruvmanila: Agree, I think we should update the function documentation where "This function is a no-op..." is mentioned.

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -139,10 +147,6 @@ pub(super) fn clear_diagnostics(session: &Session, uri: &lsp_types::Url, client:
 ///
 /// [publish diagnostics notification]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics
 pub(super) fn publish_diagnostics(session: &Session, url: &lsp_types::Url, client: &Client) {
-    if session.client_capabilities().supports_pull_diagnostics() {
-        return;
-    }
```
dhruvmanila, line 144: Is there a specific reason this has been moved down after the document snapshotting? This seems like a small optimization to avoid taking the snapshot if the client uses pull diagnostics.

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -115,16 +115,24 @@ impl LspDiagnostics {
     }
 }
 
+pub(super) fn clear_diagnostics_if_needed(
```
dhruvmanila, line 116: Why is this (`*_if_needed` variant of `clear_diagnostics`) required when a single `publish_diagnostics` is sufficient?

It might be worth adding a brief documentation for what does "if needed" mean here.

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -17,15 +17,18 @@ use crate::document::{FileRangeExt, ToRangeExt};
 use crate::session::DocumentSnapshot;
 use crate::session::client::Client;
 use crate::system::{AnySystemPath, file_to_url};
-use crate::{NotebookDocument, PositionEncoding, Session};
+use crate::{PositionEncoding, Session};
 
-pub(super) struct Diagnostics<'a> {
+// THe real challenge here is that `to_lsp_range` now needs to
+// map to the cell document instead of the document. This is tricky.
```
dhruvmanila: Is this a leftover comment? Or, which part of the code is this referring to?

crates/ty_server/src/server/api/diagnostics.rs
```diff
@@ -139,10 +147,6 @@ pub(super) fn clear_diagnostics(session: &Session, uri: &lsp_types::Url, client:
 ///
 /// [publish diagnostics notification]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics
 pub(super) fn publish_diagnostics(session: &Session, url: &lsp_types::Url, client: &Client) {
```
dhruvmanila: I think we could rename this to `publish_diagnostics_if_needed` to be consistent with the `clear_diagnostics_*` version.

crates/ty_server/src/server/api/diagnostics.rs
dhruvmanila: Oh, is the reason that a document representing the cell doesn't really exists on the ty side and it only exists in the LSP index which means that when closing a document representing a notebook cell, it needs to take a different logic (as is done using the `document.is_cell` check)?

crates/ty_server/src/server/api/notifications/did_close.rs
```diff
@@ -31,55 +32,67 @@ impl SyncNotificationHandler for DidCloseTextDocumentHandler {
             .document_handle(&uri)
             .with_failure_code(ErrorCode::InternalError)?;
 
-        let path = document.to_file_path().into_owned();
+        if !document.is_cell() {
+            close_document(&document, session, client);
+        }
```
dhruvmanila: It might be worth adding a comment for posterity to explain why this is not being done for a document representing a notebook cell. Refer to my other comment in `diagnostics.rs` as I got confused why cell document needs to be handled uniquely but I understand it now.

crates/ty_server/src/server/api/notifications/did_change_watched_files.rs
dhruvmanila: We have the logic for the Ruff language server to do this, is that not relevant here? The Ruff server will update the diagnostics for all the _open_ notebooks.

Edit: Ok, I just tested this out and VS Code does send the `didChangeNotebook` notification but it is also sending the `didChangeWatchedFiles` notification which I guess is irrelevant here.

crates/ty_server/src/session/index.rs
```diff
@@ -83,26 +59,88 @@ impl Index {
...
+        let (structure, data, text_content) = cells
+            .map(|cells| {
+                let lsp_types::NotebookDocumentCellChange {
+                    structure,
+                    data,
+                    text_content,
+                } = cells;
+                (structure, data, text_content)
+            })
+            .unwrap_or_default();
```
dhruvmanila, line 76: Is there a specific reason for using `unwrap_or_default` here instead of `if let ...` ?

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
...
-                }
-            }
+        // Re-build the cell-index if new cells were added, deleted or removed
+        if !deleted_range.is_empty() || added > 0 {
+            self.cell_index.clear();
+            self.cell_index.extend(
+                self.cells
+                    .iter()
+                    .enumerate()
+                    .map(|(i, cell)| (cell.url.clone(), i)),
```
dhruvmanila, line 151:
```suggestion
                    .map(|(index, cell)| (cell.url.clone(), index)),
```

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
...
+        );
 
-            if let Some(cell_data) = data {
-                for cell in cell_data {
-                    if let Some(existing_cell) = self.cell_by_uri_mut(cell.document.as_str()) {
-                        existing_cell.kind = cell.kind;
-                    }
-                }
-            }
+        // Re-build the cell-index if new cells were added, deleted or removed
```
dhruvmanila, line 144:
```suggestion
        // Re-build the cell-index if new cells were added or deleted
```

crates/ty_server/src/document/notebook.rs
```diff
@@ -118,93 +114,38 @@ impl NotebookDocument {
 
     pub(crate) fn update(
```
dhruvmanila, line 124: 💯 I really like how the `update` method is simplified given that it has been the source of multiple hard to find bugs in the Ruff server !!

crates/ty_server/src/document/range.rs
```diff
@@ -127,12 +131,32 @@ impl PositionExt for lsp_types::Position {
...
+            let cell_index = notebook_document.cell_index_by_uri(url)?;
+
+            let cell_start_offset = notebook.cell_offset(cell_index).unwrap_or_default();
+            let cell_relative_line = OneIndexed::from_zero_indexed(u32_index_to_usize(self.line));
+
+            let cell_start_location =
+                index.source_location(cell_start_offset, source.as_str(), encoding.into());
+            assert_eq!(cell_start_location.character_offset, OneIndexed::MIN);
+
+            let absolute_start = SourceLocation {
```
dhruvmanila: This seems more like an `absolute_position` instead of the absolute start?

crates/ty_server/src/document/range.rs
```diff
@@ -165,6 +189,19 @@ impl TextSizeExt for TextSize {
         let source = source_text(db, file);
         let index = line_index(db, file);
 
+        if let Some(notebook) = source.as_notebook() {
```
dhruvmanila, line 193: Unrelated to this PR, but there seems to be a some typo in the documentation for `TextSizeExt::to_lsp_position` and `ToRangeExt::to_lsp_range`:

```rs
    /// Converts self into a position into an LSP text document (can be a cell or regular document).
```

Should it be "Converts self into a position for an LSP text ..."?

crates/ty_server/src/document/range.rs
```diff
@@ -252,6 +289,38 @@ impl ToRangeExt for TextRange {
...
+            let notebook_index = notebook.index();
+            let notebook_document = db.notebook_document(file)?;
+
+            let start_in_concatenated =
+                index.source_location(self.start(), &source, encoding.into());
+            let cell_index = notebook_index.cell(start_in_concatenated.line)?;
+            let cell_range = notebook.cell_range(cell_index)?;
+
+            // Clamp the end offset to the end of the cell.
+            let end_offset = self.end().min(cell_range.end());
```
dhruvmanila: I think it might be useful to expand this comment with a reason on why this is necessary (I'm not exactly sure why).

crates/ty_server/src/server/api/requests/semantic_tokens.rs
```diff
@@ -33,14 +36,26 @@ impl BackgroundDocumentRequestHandler for SemanticTokensRequestHandler {
             return Ok(None);
         }
 
-        let Some(file) = snapshot.to_file(db) else {
+        let Some(file) = snapshot.to_notebook_or_file(db) else {
             return Ok(None);
         };
 
+        let mut range = None;
```
dhruvmanila: It might be worth renaming this to `cell_range` instead with a comment stating that this is required to limit the semantic token generation for the cell as `file` would represent an entire notebook.

crates/ty_server/tests/e2e/main.rs
```diff
@@ -433,6 +434,33 @@ impl TestServer {
...
     }
 
+    /// Collects `N` publish diagnostics into a map, indexed by the document url.
+    ///
+    /// ## Panics
+    /// If there are multiple publish diagnostics notifications for the same document.
+    pub(crate) fn collect_publish_diagnostics(
+        &mut self,
+        count: usize,
+    ) -> Result<BTreeMap<lsp_types::Url, Vec<lsp_types::Diagnostic>>> {
```
dhruvmanila: At first, I thought this was collecting `N` number of diagnostics but that's not true, maybe we can rename this to `send_publish_diagnostics_requests` with documentation as "Send `N` publish diagnostics requests and collect the diagnostics into ..."?

crates/ty_server/src/document/notebook.rs
dhruvmanila: I think this has been ported over to use the E2E test framework, is that correct @MichaReiser ?

MichaReiser: Yes, I ported this to an E2E test (which I'm very glad that I did because I made the same mistake as we had in ruff's first notebook implementation 😆)

crates/ty_server/src/document/text_document.rs
MichaReiser: The main reason we stored the `index` in Ruff is because we use `document.index` to map positions from text ranges to LSP ranges (and vice versa). However, this isn't the case for ty. In ty, we use `line_index` which has its own caching. 

The only remaining place where `index` was used was in the `document.update`, which we only need to compute when a document changes and the index is outdated immediately after. I didn't bother trying to use `line_index` in `document.update`. We could, but it's probably not worth the complexity.

dhruvmanila: That makes sense, thanks for the explanation!

crates/ty_server/src/server/api/notifications/did_change_watched_files.rs
MichaReiser: I'm surprised that it sends both. That seemms like bug in VS Code. For non-notebooks, VS Code sends a `didChange` event when the file is open and a `didChangeWatchedFiles` event otherwise. 

`didChangeWatchedFiles` isn't relevant for notebooks because we can't do anything useful with them when they're closed. That's why we skip them here.

crates/ruff_db/src/source.rs
MichaReiser: Lol. I'll update it when you change our module resolver to work with notebooks ;)

crates/ruff_notebook/src/notebook.rs
MichaReiser: `CellOffset` contains:

* the offset `0` (even if the notebook has no cells)
* the end offset of every cell

Let's assume the notebook has one cell where the cell's length is 20 bytes. The build cell offsets then contains `[0, 20]`. Retrieving the range for cell 0 returns `0..20`. 

The function returns `None` if called with an out-of-bounds cell. E.g. `cell_range(1)` succeeds to look up the start offset (20), but it fails to retrieve the end offset and returns `None` instead

crates/ty_server/src/document/notebook.rs
MichaReiser: Yes, I was very pleased when I found it. I first did some obscure reverse insertion in combination with `insert`. Not only did I get it wrong in the first place, it also would have had horrible performance. 

However, it took me a surprising long time to find it. Surprising because this is/was the main array manipulation API in JavaScript. So I should have tried this first.

crates/ty_server/src/server/api/diagnostics.rs
MichaReiser: Oops, sorry. Yes, this is just me taking notes and forgetting to clean up.... Am I an AI?

crates/ty_server/src/server/api/notifications/did_close.rs
MichaReiser: I'm sorry that I have to break this to you but...

`textDocument/publishDiagnostics`

crates/ty_server/src/server/api/notifications/did_open.rs
MichaReiser: Hmm, maybe. I used `opened` to indicate that it should be called **after** the document was opened. Similar to the Reacts callbacks `onClicked` or `onChanged`. But let me think if there's a better name

crates/ty_server/src/server/api/requests/semantic_tokens.rs
MichaReiser: It will take me forever to get used to let-chains... But you can see, I tried.

crates/ty_server/src/document/notebook.rs
MichaReiser: It's so that we can map it to `ruff_notebook::CodeCell::execution_count` (the naming here is a bit weird but I'm 99% sure it's the same)

crates/ty_server/src/server/api/diagnostics.rs
MichaReiser: We can't use pull diagnostics if the document is a cell or notebook and we need to snapshot the document to make this decision.

dhruvmanila: Right, but we can still make the decision about whether to even take the snapshot if the client doesn't support pull diagnostics at all.

MichaReiser: I'm not sure I follow. We only early exit if the client supports pull diagnostics **and** it's not a notebook cell and we need the document to make this decision. 

Also, taking a snapshot is fairly cheap (a few clones and bumping a few `Arc` counters)

crates/ty_server/src/session/index.rs
MichaReiser: It avoids very deeply nested `if let` chains because there are so many fields that can be `None`.

crates/ty_server/tests/e2e/main.rs
MichaReiser: I can rename it to `collect_publish_diagnostic_notifications`. Renaming it to `send` seems incorrect to me as it isn't sending any messages.

PR #21176, opened 2025-10-31
Avoid extra parentheses for long `match` patterns with `as` captures

### LINKED ISSUE #17796: Formatting adds unnecessary parentheses to long patterns with `as` captures
### Summary

See this playground link, where I've put the expected formatting going in, and you can see the buggy formatting coming out:

https://play.ruff.rs/477afb90-ce93-47a6-a5a1-ed528ab0ef8f?secondary=Format

It appears that "grouping" characters in patterns like `()`, `[]`, and `{}` aren't being used correctly when they are followed by a capture (`as name`). Instead of breaking on these, the entire pattern (including the capture) is wrapped in parentheses and indented.

First encountered in https://github.com/python/cpython/pull/133123/files#r2071981099.

### Version

0.11.8

ntBre:
Summary
--

This PR fixes #17796 by taking the approach mentioned in https://github.com/astral-sh/ruff/issues/17796#issuecomment-2847943862 of simply recursing into the `MatchAs` patterns when checking if we need parentheses. This allows us to reuse the parentheses in the inner pattern before also breaking the `MatchAs` pattern itself:

```diff
 match class_pattern:
     case Class(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx) as capture:
         pass
-    case (
-        Class(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx) as capture
-    ):
+    case Class(
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ) as capture:
         pass
-    case (
-        Class(
-            xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
-        ) as capture
-    ):
+    case Class(
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ) as capture:
         pass
     case (
         Class(
@@ -685,13 +683,11 @@
 match sequence_pattern_brackets:
     case [xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] as capture:
         pass
-    case (
-        [xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] as capture
-    ):
+    case [
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ] as capture:
         pass
-    case (
-        [
-            xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
-        ] as capture
-    ):
+    case [
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+    ] as capture:
         pass
```

I haven't really resolved the question of whether or not it's okay always to recurse, but I'm hoping the ecosystem check on this PR might shed some light on that.

Test Plan
--

New tests based on the issue and then reviewing the ecosystem check here

### DIFF
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/match.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/match.py
@@ -642,3 +642,29 @@ match sequence_pattern_brackets:
         xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
     ] as capture:
         pass
+
+
+match class_pattern:
+    # 1
+    case Class(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx) as capture:  # 2
+        # 3
+        pass # 4
+    # 5
+    case Class( # 6
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx # 7
+    ) as capture: # 8
+        pass
+    case Class( # 9
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx # 10
+    ) as capture: # 11
+        pass
+    case Class( # 12
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx # 13
+    ) as really_really_really_really_really_really_really_really_really_really_really_really_long_capture: # 14
+        pass
+    case Class( # 0
+            # 1
+            xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx # 2
+            # 3
+    ) as capture:
+        pass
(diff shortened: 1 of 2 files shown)


MichaReiser:
> I haven't really resolved the question of whether or not it's okay always to recurse, but I'm hoping the ecosystem check on this PR might shed some light on that.

😆

dylwil3:
To save you future pain, or rather to bring future pain into the present, I recommend adding some fixtures with absurd comment placement 😄

ntBre:
> To save you future pain, or rather to bring future pain into the present, I recommend adding some fixtures with absurd comment placement 😄

Thanks for the suggestion! I added some more test cases, but I'm definitely open to additional suggestions. I'm not sure I captured the full absurdity available here.

I guess I'll tentatively call it okay always to recurse since nothing showed up in the ecosystem check. From what I remember with the syntax errors, many projects still support 3.9 and don't use much pattern matching in general, so I guess this isn't too surprising. The main case I though to worry about was this test case:

```py
match class_pattern:
    case Class(
        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    ) as really_really_really_really_really_really_really_really_really_really_really_really_long_capture:
        pass
```

but it both wraps the `Class` and the whole `MatchAs` pattern, as I hoped.

Open to suggestions there as well!

MichaReiser (approved):

Issue #21162, opened 2025-10-31
PLR1708 false positives on nested functions


dscorbett:
### Summary

[`stop-iteration-return` (PLR1708)](https://docs.astral.sh/ruff/rules/stop-iteration-return/) reports false positives when a non-generator function raises `StopIteration` and contains a nested generator function, or a generator function contains a nested non-generator function that raises `StopIteration`. [Example](https://play.ruff.rs/2218cd6c-d01f-422b-a5a3-469d1b9c0cbf):
```console
$ cat >plr1708.py <<'# EOF'
def f():
    def g():
        yield 1
    raise StopIteration

def g():
    def f():
        raise StopIteration
    yield 1
# EOF

$ ruff --isolated check plr1708.py --select PLR1708 --preview --output-format concise -q
plr1708.py:4:5: PLR1708 Explicit `raise StopIteration` in generator
plr1708.py:8:9: PLR1708 Explicit `raise StopIteration` in generator
```

### Version

ruff 0.14.3 (8737a2d5f 2025-10-30)

PR #21177, opened 2025-10-31
[`pylint`] Fix `PLR1708` false positives on nested functions

chirizxc:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/21162

## Test Plan

`cargo nextest run pylint`

### DIFF
--- a/crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
+++ b/crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
@@ -1,6 +1,7 @@
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::{
     self as ast,
+    helpers::map_callable,
     visitor::{Visitor, walk_expr, walk_stmt},
 };
 use ruff_text_size::Ranged;
@@ -59,9 +60,7 @@ pub(crate) fn stop_iteration_return(checker: &Checker, function_def: &ast::StmtF
         stop_iteration_raises: Vec::new(),
     };
 
-    for stmt in &function_def.body {
-        analyzer.visit_stmt(stmt);
-    }
+    analyzer.visit_body(&function_def.body);
 
     if analyzer.has_yield {
         for raise_stmt in analyzer.stop_iteration_raises {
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -50,65 +52,65 @@ impl Violation for StopIterationReturn {
...
-            checker.semantic().match_builtin_expr(func, "StopIteration")
-        }
-        expr => checker.semantic().match_builtin_expr(expr, "StopIteration"),
-    };
-
-    if !is_stop_iteration {
-        return;
+    for stmt in &function_def.body {
+        analyzer.visit_stmt(stmt);
     }
```
ntBre: nit:
```suggestion
    analyzer.visit_body(&function_def.body);
```

crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs
```diff
@@ -50,65 +52,65 @@ impl Violation for StopIterationReturn {
...
             }
+            _ => walk_expr(self, expr),
         }
     }
+}
+
+fn is_stop_iteration_raise(stmt_raise: &ast::StmtRaise, checker: &Checker) -> bool {
+    let Some(exc) = &stmt_raise.exc else {
+        return false;
+    };
```
ntBre: I'm not sure how helpful this helper function is. I think we could just destructure the `StmtRaise` in the `match` on line 83 and use `map_callable`:

https://github.com/astral-sh/ruff/blob/192c37d5401786b3043aafbfb657b4f5e7df8ac8/crates/ruff_python_ast/src/helpers.rs#L724-L735

For example,

```rust
    fn visit_stmt(&mut self, stmt: &'a ast::Stmt) {
        match stmt {
            ast::Stmt::FunctionDef(_) => {}
            ast::Stmt::Raise(ast::StmtRaise { exc: Some(exc), .. }) => {
                if self.checker.semantic().match_builtin_expr(map_callable(exc), "StopIteration") {
                    self.stop_iteration_raises.push(stmt_raise);
                }
                walk_stmt(self, stmt);
            }
            _ => walk_stmt(self, stmt),
        }
    }
```

I was going to suggest removing the `walk_stmt` call in the `Stmt::Raise` arm, but I think that's actually correct, and B901 has a bug in it by only checking for statements with a single yield expression:

https://github.com/astral-sh/ruff/blob/192c37d5401786b3043aafbfb657b4f5e7df8ac8/crates/ruff_linter/src/rules/flake8_bugbear/rules/return_in_generator.rs#L118-L123

That means it misses cases like this:

```py
def foo():
    return (yield 1)
```

Feel free to follow up on that if you're interested!

This also applies here, where your code will correctly flag something like this:

```py
def foo():
    raise StopIteration((yield 1))
```

That might be a nice test case to add.

ntBre (commented):
Thanks! This looks good to me now, I just had a couple more minor simplification suggestions.

ntBre (approved):
Thank you!

PR #21185, opened 2025-11-01
Adjust own-line comment placement between branches

dylwil3:
This PR attempts to improve the placement of own-line comments between branches in the setting where the comment is more indented than the preceding node.

There are two main changes.

### First change: Preceding node has leading content

If the preceding node has leading content, we now regard the comment as automatically _less_ indented than the preceding node, and format accordingly.

For example, 

```python
if True: preceding_node
# leading on `else`, not trailing on `preceding_node`
else: ...
```

This is more compatible with `black`, although there is a (presumably very uncommon) edge case:

```python
if True:
    this;that
    # leading on `else`, but trailing in `black`
else: ...
```

I'm sort of okay with this - presumably if one wanted a comment for those semi-colon separated statements, one should have put it _above_ them, and one wanted a comment only for `that` then it ought to have been on the same line?

### Second change: searching for last child in body

While searching for the (recursively) last child in the body of the preceding _branch_, we implicitly assumed that the preceding node had to have a body to begin the recursion. But actually, in the base case, the preceding node _is_ the last child in the body of the preceding branch. So, for example:

```python
if True:
    something
    last_child_but_no_body
    # leading on else for `main` but trailing in this PR
else: ...
```

### More examples

The table below is an attempt to summarize the changes in behavior. The rows alternate between an example snippet with `while` and the same example with `if` - in the former case we do _not_ have an `else` node and in the latter we do.

Notice that:

1. On `main` our handling of `if` vs. `while` is not consistent, whereas it is consistent in the present PR
2. We disagree with `black` in all cases except that last example on `main`, but agree in all cases for the present PR (though see above for a wonky edge case where we disagree).

<table>
<tr>
<th>Original &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</th>
<th><code>main</code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</th>
<th>This PR&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</th>
<th><code>black</code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</th>
</tr>
<tr>
<td>

<pre lang="python">
while True: 
    pass
        # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
else:
    # comment
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
    # comment
else:
    pass
</pre>

</td>
</tr>
<tr>
<td>

<pre lang="python">
if True:
    pass
        # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
    # comment
else:
    pass
</pre>

</td>
</tr>
<tr>
<td>

<pre lang="python">
while True: pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
# comment
else:
    pass
</pre>

</td>
</tr>
<tr>
<td>

<pre lang="python">
if True: pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
</tr>
<tr>
<td>

<pre lang="python">
while True: pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
else:
    # comment
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
while True:
    pass
# comment
else:
    pass
</pre>

</td>
</tr>
<tr>
<td>

<pre lang="python">
if True: pass
    # comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
<td>

<pre lang="python">
if True:
    pass
# comment
else:
    pass
</pre>

</td>
</tr>
</table>

### DIFF
--- a/crates/ruff_python_formatter/tests/snapshots/format@statement__while.py.snap
+++ b/crates/ruff_python_formatter/tests/snapshots/format@statement__while.py.snap
@@ -36,36 +36,36 @@ while (
     print("Do something")
 
 while True: pass
-    # comment
+    # 1
 else:
     pass
 
 while True:
     pass
-        # comment
+        # 2
 else:
     pass
 
 while True: pass
-# comment
+# 3
 else:
     pass
 
 while True: pass
-    # comment
+    # 4
 else:
     pass
 
 def foo():
     while True:
         pass
-# comment
+# 5
     else:
         pass
 
 while True:
     first;second
-    # comment
+    # 6
 else:
     pass
 ```
@@ -106,25 +106,25 @@ while (
 
 while True:
     pass
-# comment
+# 1
 else:
     pass
 
 while True:
     pass
-    # comment
+    # 2
 else:
     pass
 
 while True:
     pass
-# comment
+# 3
 else:
     pass
 
 while True:
     pass
-# comment
+# 4
 else:
     pass
 
@@ -132,7 +132,7 @@ else:
 def foo():
     while True:
         pass
-    # comment
+    # 5
     else:
         pass
 
@@ -140,7 +140,7 @@ def foo():
 while True:
     first
     second
-# comment
+# 6
 else:
     pass
 ```
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/if.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/if.py
@@ -298,35 +298,35 @@ if parent_body:
 # Compare behavior with `while`/`else` comment placement
 
 if True: pass
-    # comment
+    # 1
 else:
     pass
 
 if True:
     pass
-        # comment
+        # 2
 else:
     pass
 
 if True: pass
-# comment
+# 3
 else:
     pass
 
 if True: pass
-    # comment
+    # 4
 else:
     pass
 
 def foo():
     if True:
         pass
-# comment
+# 5
     else:
         pass
 
 if True:
     first;second
-    # comment
+    # 6
 else:
     pass
--- a/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/while.py
+++ b/crates/ruff_python_formatter/resources/test/fixtures/ruff/statement/while.py
@@ -30,35 +30,35 @@ while (
     print("Do something")
 
 while True: pass
-    # comment
+    # 1
 else:
     pass
 
 while True:
     pass
-        # comment
+        # 2
 else:
     pass
 
 while True: pass
-# comment
+# 3
 else:
     pass
 
 while True: pass
-    # comment
+    # 4
 else:
     pass
 
 def foo():
     while True:
         pass
-# comment
+# 5
     else:
         pass
 
 while True:
     first;second
-    # comment
+    # 6
 else:
     pass
--- a/crates/ruff_python_formatter/tests/snapshots/format@statement__if.py.snap
+++ b/crates/ruff_python_formatter/tests/snapshots/format@statement__if.py.snap
@@ -304,36 +304,36 @@ if parent_body:
 # Compare behavior with `while`/`else` comment placement
 
 if True: pass
-    # comment
+    # 1
 else:
     pass
 
 if True:
     pass
-        # comment
+        # 2
 else:
     pass
 
 if True: pass
-# comment
+# 3
 else:
     pass
 
 if True: pass
-    # comment
+    # 4
 else:
     pass
 
 def foo():
     if True:
         pass
-# comment
+# 5
     else:
         pass
 
 if True:
     first;second
-    # comment
+    # 6
 else:
     pass
 ```
@@ -647,25 +647,25 @@ if parent_body:
 
 if True:
     pass
-# comment
+# 1
 else:
     pass
 
 if True:
     pass
-    # comment
+    # 2
 else:
     pass
 
 if True:
     pass
-# comment
+# 3
 else:
     pass
 
 if True:
     pass
-# comment
+# 4
 else:
     pass
 
@@ -673,7 +673,7 @@ else:
 def foo():
     if True:
         pass
-    # comment
+    # 5
     else:
         pass
 
@@ -681,7 +681,7 @@ def foo():
 if True:
     first
     second
-# comment
+# 6
 else:
     pass
 ```


crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -569,6 +569,15 @@ fn handle_own_line_comment_between_statements<'a>(
...
     if max_empty_lines(&source[TextRange::new(comment.end(), following.start())]) == 0 {
+        for token in SimpleTokenizer::new(source, TextRange::new(comment.end(), following.start()))
+        {
+            match token.kind() {
+                SimpleTokenKind::Else | SimpleTokenKind::Finally => {
+                    return CommentPlacement::dangling(comment.enclosing_node(), comment);
+                }
+                _ => {}
+            }
+        }
```
dylwil3: Note that `max_empty_lines` also uses a `SimpleTokenizer`. We could combine that into a single pass here, I think, but it would be less readable. Is it worth it?

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -569,6 +569,15 @@ fn handle_own_line_comment_between_statements<'a>(
     // y = 2
     // ```
     if max_empty_lines(&source[TextRange::new(comment.end(), following.start())]) == 0 {
+        for token in SimpleTokenizer::new(source, TextRange::new(comment.end(), following.start()))
```
MichaReiser: Can you say more why we're handling this case here? I'd expect this to be handled by `handle_own_line_comment_between_branches`

dylwil3: After our offline discussion I've tried a hopefully more reasonable spot in the logic for the fix

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -602,9 +602,30 @@ fn handle_own_line_comment_between_branches<'a>(
...
+        // Note we even do this if the comment is very indented
+        // (which matches `black`'s behavior as of 2025.11.11)
+        //
+        // ```python
+        // if True: pass
+        //          # leading on `else`
+        // else:
+        //     pass
+        // ```
+        || comment_indentation + TextSize::new(1),
```
MichaReiser: What's the `+1` for? I like to use `char.text_len`() (e.g. `' '.text_len()`) to more clearly what character we're skipping.

Is it important what value we use for `comment_indentation` or can it be any value for as long as it is larger than `comment_indentation` (so that we hit the `Less` case?)

crates/ruff_python_formatter/tests/snapshots/format@statement__while.py.snap
```diff
@@ -35,6 +34,40 @@ while (
     and anotherCondition or aThirdCondition  # trailing third condition
 ):  # comment
     print("Do something")
+
+while True: pass
+    # comment
```
MichaReiser: I often find it useful to number the comments. It makes it easier to match the input/output (and debug panics if we forgot to format a comment)

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -678,9 +699,31 @@ fn handle_own_line_comment_after_branch<'a>(
...
+    //
+    // ```python
+    // if True:
+    //     something
+    //     last_child
+    //         # comment
+    // else:
+    //     pass
+    // ```
+    let last_child = preceding.last_child_in_body().unwrap_or(preceding);
```
MichaReiser: This seems risky. There isn't really anything enforcing that we're within a block statement. I think we, at least, want to assert that the enclosing node is a block statement.

MichaReiser (changes requested):
This looks much better to me and it's nice that this matches black in almost all cases.

However, there are a couple of tests that now fail because of instable formatting (the invariant `format(source) == format(format(source))` doesn't hold anymore). I think it's because the fallback to `preceding` is too liberal.

crates/ruff_python_formatter/tests/snapshots/format@statement__while.py.snap
dylwil3: Here I didn't do that because there is only one comment in each example - do you want me to the number them across the whole fixture? Among these examples? Or just make them all `#1`?

crates/ruff_python_formatter/src/comments/placement.rs
dylwil3: We are just trying to hit the `Less` case - it can be any positive number. I've clarified with a comment.

dylwil3:
Sorry about the failing tests! I made the beginner mistake of running tests, fixing snapshots, but then not re-running tests after that. Whoops!

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -678,8 +704,49 @@ fn handle_own_line_comment_after_branch<'a>(
...
+            comment.enclosing_node(),
+            AnyNodeRef::StmtIf(_)
+                | AnyNodeRef::StmtWhile(_)
+                | AnyNodeRef::StmtFor(_)
+                | AnyNodeRef::StmtMatch(_)
+                | AnyNodeRef::ElifElseClause(_)
+                | AnyNodeRef::StmtTry(_)
+                | AnyNodeRef::MatchCase(_)
+                | AnyNodeRef::ExceptHandlerExceptHandler(_)
+        ) =>
```
MichaReiser: Could we use `if following.is_first_statement_in_body()` or even `is_first_statement_in_alternate_body`

I'm surprised that we need `MatchCase` and `ExceptHandler` here, as those nodes don't have alternate branches? How did you pick the nodes listed here? Did you add tests demonstrating that they're all necessary?

crates/ruff_python_formatter/tests/snapshots/format@statement__while.py.snap
MichaReiser: I would give them unique numbers within the `# Compare behavior with `while`/`else` comment placement` section

crates/ruff_python_formatter/src/comments/placement.rs
dylwil3: Yes, sorry - I added all of those because that was my interpretation of

> There isn't really anything enforcing that we're within a block statement. I think we, at least, want to assert that the enclosing node is a block statement.

But you're right that `is_first_statement_in_alternate_body` is more targeted, so I've changed it to that, thanks!

PR #21187, opened 2025-11-01
[`flake8-simplify`] Fix SIM222 false positive for `tuple(generator) or None` (`SIM222`)

### LINKED ISSUE #21136: SIM222 overstepping
### Summary

Hello!
I was updating to v0.14.2, I had only one unsafe fix with SIM222. I think the rule is overstepping with this example:

```py
def get_items() -> tuple[Item, ...] | None:
    return tuple(item for item in Item.objects.all()) or None
```

suggested fix:
```py
def get_items() -> tuple[Item, ...] | None:
    return tuple(item for item in Item.objects.all())
```

I'm not saying the code has the best design, but the suggestion changes the function behavior (of course not critical, this is an unsafe fix)

By the way, the fix not suggested in this case (no comprehension)
```py
def get_items() -> tuple[Item, ...] | None:
    return tuple(Item.objects.all()) or None
```

danparizher:
## Summary

Fixes #21136. The rule was incorrectly treating `tuple(generator)` as always truthy, but `tuple(empty_generator)` is falsy.

## Problem Analysis

SIM222 was incorrectly suggesting to remove `or None` from `tuple(item for item in items) or None` because it was treating the tuple call as always truthy. However, when a generator expression is empty, `tuple(empty_generator)` produces an empty tuple `()`, which is falsy. Removing `or None` would change the runtime behavior.

## Approach

Updated `Truthiness::from_expr` to return `Unknown` for `tuple(generator)`, `tuple(list_comp)`, and `tuple(set_comp)` calls, since we cannot statically determine if the resulting tuple will be empty or not. This prevents SIM222 from flagging these cases incorrectly.

### DIFF
--- a/crates/ruff_python_ast/src/helpers.rs
+++ b/crates/ruff_python_ast/src/helpers.rs
@@ -1318,16 +1318,18 @@ impl Truthiness {
                         if arguments.is_empty() {
                             // Ex) `list()`
                             Self::Falsey
-                        } else if arguments.args.len() == 1 && arguments.keywords.is_empty() {
+                        } else if let [argument] = &*arguments.args
+                            && arguments.keywords.is_empty()
+                        {
                             // Ex) `list([1, 2, 3])`
                             // For tuple(generator), we can't determine statically if the result will
                             // be empty or not, so return Unknown. The generator itself is truthy, but
                             // tuple(empty_generator) is falsy. ListComp and SetComp are handled by
                             // recursing into Self::from_expr below, which returns Unknown for them.
-                            if matches!(&arguments.args[0], Expr::Generator(_)) {
+                            if argument.is_generator_expr() {
                                 Self::Unknown
                             } else {
-                                Self::from_expr(&arguments.args[0], is_builtin)
+                                Self::from_expr(argument, is_builtin)
                             }
                         } else {
                             Self::Unknown


crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1320,7 +1320,18 @@ impl Truthiness {
...
                         } else if arguments.args.len() == 1 && arguments.keywords.is_empty() {
                             // Ex) `list([1, 2, 3])`
-                            Self::from_expr(&arguments.args[0], is_builtin)
+                            // For tuple(generator), tuple(list_comp), tuple(set_comp), we can't
+                            // determine statically if the result will be empty or not, so return Unknown.
+                            // The generator/comprehension itself is truthy, but tuple(empty_generator)
+                            // is falsy.
+                            if matches!(
+                                &arguments.args[0],
+                                Expr::Generator(_) | Expr::ListComp(_) | Expr::SetComp(_)
```
ntBre: I think this only needs to include `Generator`, not the comprehension types. Recursing into the `Self::from_expr` arm below should already return `Self::Unknown` for other comprehensions. Generators are the only special case on 1223, if I'm following correctly.

The last two new tests already suppress the rule, for example: https://play.ruff.rs/a211c51a-b6f3-4dae-b259-431a75788464

PR #21189, opened 2025-11-01
[`flake8-bugbear`] Flag mutable defaults passed via named variables, excluding UPPER_CASE (`B006`)

### LINKED ISSUE #21142: B006 doesn’t flag mutable defaults when passed via a named constant
Ruff’s B006 (mutable-argument-default) correctly flags obvious mutable defaults written inline (e.g., {}, [], set()), but it does not warn when the default is a reference to a module-level named constant that is a mutable object (e.g., a set). This yields a false negative and can lead to shared mutable state across calls/instances without any lint warning.

This appears to be a false negative due to B006’s syntax-local detection not resolving identifiers to determine mutability.

# Environment
Ruff version: ruff 0.9.6
Python version: Python 3.13.7
pyproject.toml file Config has: select = ["ALL"]; B006 not ignored
```
[tool.ruff.lint]
select = ["ALL"]
exclude = [".github/**"]
```

# Minimal Reproducer
```
Python
# Module-level mutable “constant”
MY_SET = {"ABC", "DEF"}  # plain set

# NOT triggering B006
def func_A(s: set[str] = MY_SET):
    return s

# Triggering B006
def func_B(s: set[str] = {"ABC", "DEF"}):
    return s
```
# Observed Behavior

 - Ruff flags `func_B` with B006 (as expected).
 - Ruff does not flag `func_A` even though they share the same mutable default object via `MY_SET`.

# Expected Behavior

- B006 should flag mutable default arguments even when the default is provided via a referenced identifier bound at module scope to a mutable object (list/set/dict).

danparizher:
## Summary

Fixes #21142

B006 now flags mutable defaults passed via module-level named constants. Previously, it only flagged inline mutable defaults (e.g., `{}`, `[]`, `set()`), missing cases where the default references a module-level constant bound to a mutable object (e.g., `MY_SET = {"ABC", "DEF"}`).

## Problem Analysis

B006’s detection was syntax-local and didn’t resolve identifiers to check mutability. When a default used a name like `MY_SET`, it wasn’t resolved to the bound mutable object, causing false negatives.

**Example:**
```python
MY_SET = {"ABC", "DEF"}  # module-level mutable constant

def func_A(s: set[str] = MY_SET):  # Not flagged (should be flagged)
    return s

def func_B(s: set[str] = {"ABC", "DEF"}):  # Correctly flagged
    return s
```

This is a false negative because `func_A` and `func_B` share the same mutable default object.

## Approach

Extended `is_guaranteed_mutable_expr` (used in preview mode) to handle `Expr::Name`:

1. Resolve the identifier using `semantic.only_binding()` to get the binding.
2. Require an assignment binding (not imports, parameters, etc.).
3. Extract the assigned value via `find_binding_value()`.
4. Recursively check if that value is mutable using `is_guaranteed_mutable_expr()`.

This enables detecting cases like:
- `MY_SET = {"ABC", "DEF"}` → flagged when used as a default
- `MY_LIST = [1, 2, 3]` → flagged when used as a default  
- `MY_DICT = {"key": "value"}` → flagged when used as a default

The fix is enabled only in preview mode. Non-preview mode preserves existing behavior (only flags inline mutable defaults).

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
@@ -3,7 +3,7 @@ MY_SET = {"ABC", "DEF"}  # plain set
 MY_LIST = [1, 2, 3]  # plain list
 MY_DICT = {"key": "value"}  # plain dict
 
-# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
+# NOT triggering B006 (correct - PEP 8 specifies that UPPER_CASE variables are considered constants)
 def func_A(s: set[str] = MY_SET):
     return s
 
@@ -11,11 +11,11 @@ def func_A(s: set[str] = MY_SET):
 def func_B(s: set[str] = {"ABC", "DEF"}):
     return s
 
-# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
+# NOT triggering B006 (correct - PEP 8 specifies that UPPER_CASE variables are considered constants)
 def func_C(items: list[int] = MY_LIST):
     return items
 
-# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
+# NOT triggering B006 (correct - PEP 8 specifies that UPPER_CASE variables are considered constants)
 def func_D(data: dict[str, str] = MY_DICT):
     return data
 
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
@@ -3,7 +3,7 @@ MY_SET = {"ABC", "DEF"}  # plain set
 MY_LIST = [1, 2, 3]  # plain list
 MY_DICT = {"key": "value"}  # plain dict
 
-# NOT triggering B006 (should trigger)
+# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
 def func_A(s: set[str] = MY_SET):
     return s
 
@@ -11,11 +11,11 @@ def func_A(s: set[str] = MY_SET):
 def func_B(s: set[str] = {"ABC", "DEF"}):
     return s
 
-# Should trigger B006
+# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
 def func_C(items: list[int] = MY_LIST):
     return items
 
-# Should trigger B006
+# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
 def func_D(data: dict[str, str] = MY_DICT):
     return data
 
--- a/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
@@ -1,0 +1,21 @@
+# Module-level mutable "constant"
+MY_SET = {"ABC", "DEF"}  # plain set
+MY_LIST = [1, 2, 3]  # plain list
+MY_DICT = {"key": "value"}  # plain dict
+
+# NOT triggering B006 (should trigger)
+def func_A(s: set[str] = MY_SET):
+    return s
+
+# Triggering B006 (correct)
+def func_B(s: set[str] = {"ABC", "DEF"}):
+    return s
+
+# Should trigger B006
+def func_C(items: list[int] = MY_LIST):
+    return items
+
+# Should trigger B006
+def func_D(data: dict[str, str] = MY_DICT):
+    return data
+
--- a/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
+++ b/crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
@@ -10,6 +10,7 @@ use ruff_python_semantic::analyze::function_type::is_stub;
 use ruff_python_semantic::analyze::typing::{
     find_binding_value, is_immutable_annotation, is_mutable_expr,
 };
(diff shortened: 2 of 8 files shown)


BlindChickens:
This brings the rule more in line with the pylint rule it is supposed to represent/replace. `dangerous-default-value (W0102)`
Lets get it merged.

MichaReiser:
Hmm, I'm not sure I agree with this change. The `UPPER_CASE` indicates a constant:

> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.
> https://peps.python.org/pep-0008/#constants


I think it makes sense to assume that those variables are read-only by default.

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
```diff
@@ -142,6 +144,23 @@ fn is_guaranteed_mutable_expr(expr: &Expr, semantic: &SemanticModel) -> bool {
             elts.iter().any(|e| is_guaranteed_mutable_expr(e, semantic))
         }
         Expr::Named(ast::ExprNamed { value, .. }) => is_guaranteed_mutable_expr(value, semantic),
+        Expr::Name(name) => {
+            // Resolve module-level constants that are bound to mutable objects
+            let Some(binding_id) = semantic.only_binding(name) else {
+                return false;
+            };
+            let binding = semantic.binding(binding_id);
+            // Only check assignments (not imports, function parameters, etc.)
```
MichaReiser, line 159: This is interesting. That means using `a = imported.MY_LIST` is fine but using `MY_LIST` in the module itself isn't.

MichaReiser:
I'm not sure if I agree with this change. Especially because the proposed fix doesn't really help.

`ALL_UPPERCASE` names in Python are considered constants. That's why I think using them in a default value position is fine. The only risk really is that the function body might change the value, which in turn changes the constant. 

```py
A = [1, 2, 3]

def test(a = A): a.append(3) # changes `A`
```

But this is still the case even when applying the proposed fix:

```py
>>> def test(a):
...     if a is None:
...         a = A
...     a.append(4) # changes `A`
...
```

We should at least update the fix to something that actually fixes the underlying problem, e.g. by changing the type annotation to `typing.Sequence` or making a deep copy (but that also seems undesired). Overall, this seems to require more design

MichaReiser (changes requested):
See my other comment

danparizher:
Thanks for the feedback, I've updated the implementation to exclude `UPPER_CASE` constants from being flagged,

Since we're no longer flagging that, no fix is suggested for them, which avoids the underlying mutation problem. For inline mutable defaults, the current fix works because it creates a new object per call.

Maybe what we can do is when flagging a constant like `MY_LIST`, we could suggest using `MY_LIST.copy()` or `copy.deepcopy(MY_LIST)` to create a new instance, or suggest changing the type annotation to `Sequence[int]` to indicate immutability. 

We could also analyze the function body to detect if the argument is mutated and suggest appropriate fixes—if it's only read, suggest `Sequence`/`Mapping` types; if it's mutated, suggest copies or immutable alternatives like `tuple` or `frozenset`. 

This would require more design work as you noted, but could be a good follow-up. For now, I think excluding `UPPER_CASE` constants seems like the right pragmatic approach.

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
```diff
@@ -142,6 +145,27 @@ fn is_guaranteed_mutable_expr(expr: &Expr, semantic: &SemanticModel) -> bool {
             elts.iter().any(|e| is_guaranteed_mutable_expr(e, semantic))
         }
         Expr::Named(ast::ExprNamed { value, .. }) => is_guaranteed_mutable_expr(value, semantic),
+        Expr::Name(name) => {
+            // Exclude UPPER_CASE constants (PEP 8 convention - meant to be read-only)
+            if str::is_cased_uppercase(&name.id) {
+                return false;
+            }
+            // Resolve module-level constants that are bound to mutable objects
```
MichaReiser: I think this comment here is inaccurate as it doesn't just resolve module level constants:

* It resolves constants and non constants
* It resolves any name in the current scope

We should add at least one test demonstrating the behavior with a nested function that reads a value from the outer scope.

crates/ruff_linter/src/rules/flake8_bugbear/rules/mutable_argument_default.rs
MichaReiser: Can you explain the motivation for this restriction?

We should also add tests that demonstrate this behavior

crates/ruff_linter/resources/test/fixtures/flake8_bugbear/B006_10.py
```diff
@@ -0,0 +1,21 @@
+# Module-level mutable "constant"
+MY_SET = {"ABC", "DEF"}  # plain set
+MY_LIST = [1, 2, 3]  # plain list
+MY_DICT = {"key": "value"}  # plain dict
+
+# NOT triggering B006 (correct - UPPER_CASE constants are excluded per PEP 8)
```
MichaReiser: The comment here is a bit misleading. PEP8 doesn't specify that this rule should exclude constants. Instead, PEP8 specifies that UPPER_CASE variables are considered constants.

MichaReiser (commented):
I've a few more inline comments. Could you also update the PR title and summary to reflect your most recent changes?

danparizher:
The assignment restriction ensures we only flag cases where mutable objects are explicitly assigned at definition time (which are then shared across calls), avoiding false positives from imports (external dependencies) and function parameters (runtime values evaluated per call).

k4lizen:
@MichaReiser hello, what is left to be done for this to be merged, i had to fix some buggy behaviour in prod today that would have been caught by this lint :p

PR #21190, opened 2025-11-01
[`refurb`] Fix false negative for underscores before sign in `Decimal` constructor (`FURB157`)

### LINKED ISSUE #21186: FURB157 misses `decimal.Decimal("_-1")`
### Summary

[`verbose-decimal-constructor` (FURB157)](https://docs.astral.sh/ruff/rules/verbose-decimal-constructor/) has a false negative when the sign character follows an underscore. [Example](https://play.ruff.rs/34abd766-bcd1-4e22-9a21-aa2d340f859f):
```console
$ cat >furb157.py <<'# EOF'
from decimal import Decimal
Decimal("_-1")
# EOF

$ ruff --isolated check furb157.py --select FURB157
All checks passed!
```
This was a true positive in version 0.14.2.

### Version

danparizher:
## Summary

Fixes FURB157 false negative where `Decimal("_-1")` was not flagged as verbose when underscores precede the sign character. This fixes #21186.

## Problem Analysis

The `verbose-decimal-constructor` (FURB157) rule failed to detect verbose `Decimal` constructors when the sign character (`+` or `-`) was preceded by underscores. For example, `Decimal("_-1")` was not flagged, even though it can be simplified to `Decimal(-1)`.

The bug occurred because the rule checked for the sign character at the start of the string before stripping leading underscores. According to Python's `Decimal` parser behavior (as documented in CPython's `_pydecimal.py`), underscores are removed before parsing the sign. The rule's logic didn't match this behavior, causing a false negative for cases like `"_-1"` where the underscore came before the sign.

This was a regression introduced in version 0.14.3, as these cases were correctly flagged in version 0.14.2.

## Approach

The fix updates the sign extraction logic to:
1. Strip leading underscores first (matching Python's Decimal parser behavior)
2. Extract the sign from the underscore-stripped string
3. Preserve the string after the sign for normalization purposes

This ensures that cases like `Decimal("_-1")`, `Decimal("_+1")`, and `Decimal("_-1_000")` are correctly detected and flagged. The normalization logic was also updated to use the string after the sign (without underscores) to avoid double signs in the replacement output.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/refurb/FURB157.py
+++ b/crates/ruff_linter/resources/test/fixtures/refurb/FURB157.py
@@ -85,3 +85,9 @@ Decimal("1234_5678")    # Safe fix: preserves non-thousands separators
 Decimal("0001_2345")
 Decimal("000_1_2345")
 Decimal("000_000")
+
+# Test cases for underscores before sign
+# https://github.com/astral-sh/ruff/issues/21186
+Decimal("_-1")      # Should flag as verbose
+Decimal("_+1")      # Should flag as verbose
+Decimal("_-1_000")  # Should flag as verbose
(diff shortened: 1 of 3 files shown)


ntBre (approved):
Thanks, this makes sense to me!

PR #21191, opened 2025-11-01
FURB142: Parenthesize generator expression arguments in fixer to preserve scoping

### LINKED ISSUE #21098: FURB142 fix should parenthesize generator expressions
### Summary

The fix for [`for-loop-set-mutations` (FURB142)](https://docs.astral.sh/ruff/rules/for-loop-set-mutations/) should parenthesize unparenthesized generator expressions to avoid changing the program’s behavior. [Example](https://play.ruff.rs/b619adff-1744-42bc-a4d3-1cc9bef1032e):
```console
$ cat >furb142.py <<'# EOF'
s = set()
for x in ("abc", "def"):
    s.add(c for c in x)
# EOF

$ python furb142.py; echo $?
0

$ ruff check --isolated furb142.py --select FURB142 --preview --fix

$ python furb142.py 2>&1 | tail -n 3
    s.update(c for c in x for x in ("abc", "def"))
                        ^
```

### Version

sohamukute:
## Summary

Fixes a bug in the FURB142 (for-loop-set-mutations) fixer where an unparenthesized generator expression used as the `.add(...)` argument would be inlined into the rewritten comprehension without parentheses, changing variable scoping and causing a runtime NameError.

Before (problematic after fix):
```python
s = set()
for x in ("abc", "def"):
    s.add(c for c in x)
# fixed to:
s.update(c for c in x for x in ("abc", "def"))  # NameError: x
```

After (correct, parenthesized):
```python
s = set()
for x in ("abc", "def"):
    s.add(c for c in x)
# fixed to:
s.update((c for c in x) for x in ("abc", "def"))
```

Implementation details:
- In the “comprehension” rewrite path of FURB142, detect `Expr::Generator` arguments that are not already parenthesized and wrap them with parentheses in the generated replacement string.
- Leaves already-parenthesized generator expressions unchanged.
- Leaves non-generator arguments unchanged.
- Maintains existing applicability handling and comment-safety behavior.

References:
- Fixes #21098
- Rule docs: FURB142 (for-loop-set-mutations)

## Test Plan

- Repro the original failing case and confirm the new fix:

  1) Create the file:
     ```python
     s = set()
     for x in ("abc", "def"):
         s.add(c for c in x)
     ```

  2) Verify Ruff flags it:
     ```
     cargo run -p ruff_cli -- check --isolated --select FURB142 --preview furb142.py
     ```

  3) Apply the fix:
     ```
     cargo run -p ruff_cli -- check --isolated --select FURB142 --preview --fix furb142.py
     ```

  4) Inspect output (should contain parentheses):
     ```
     s.update((c for c in x) for x in ("abc", "def"))
     ```

  5) Run the file:
     ```
     python furb142.py
     ```
     Expect no NameError.

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/for_loop_set_mutations.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/for_loop_set_mutations.rs
@@ -111,13 +111,24 @@ pub(crate) fn for_loop_set_mutations(checker: &Checker, for_stmt: &StmtFor) {
                 parenthesize_loop_iter_if_necessary(for_stmt, checker, IterLocation::Call),
             )
         }
-        (for_target, arg) => format!(
-            "{}.{batch_method_name}({} for {} in {})",
-            set.id,
-            locator.slice(arg),
-            locator.slice(for_target),
-            parenthesize_loop_iter_if_necessary(for_stmt, checker, IterLocation::Comprehension),
-        ),
+        (for_target, arg) => {
+            // Parenthesize an unparenthesized generator expression argument to preserve semantics,
+            // e.g. `s.add(c for c in x)` -> `s.update((c for c in x) for x in ...)`.
+            let arg_text = match arg {
+                Expr::Generator(generator) if !generator.parenthesized => {
+                    format!("({})", locator.slice(arg))
+                }
+                _ => locator.slice(arg).to_string(),
+            };
+
+            format!(
+                "{}.{batch_method_name}({} for {} in {})",
+                set.id,
+                arg_text,
+                locator.slice(for_target),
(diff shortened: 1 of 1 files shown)


ntBre:
Thanks! Could you add some [snapshot tests](https://github.com/astral-sh/ruff/blob/main/CONTRIBUTING.md#rule-testing-fixtures-and-snapshots) too? Basically just including your manual tests in the repo to avoid future regressions.

The code looks good to me otherwise.

sohamukute:
> Thanks! Could you add some [snapshot tests](https://github.com/astral-sh/ruff/blob/main/CONTRIBUTING.md#rule-testing-fixtures-and-snapshots) too? Basically just including your manual tests in the repo to avoid future regressions.
> 
> The code looks good to me otherwise.

okay will add them.

PR #21194, opened 2025-11-01
[ty] Favor in scope completions

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1464

We sort the completions before we add the unimported ones, meaning that imported completions show up before unimported ones.

This is also spoken about in https://github.com/astral-sh/ty/issues/1274, and this is probably a duplicate of that.

@AlexWaygood mentions this [here](https://github.com/astral-sh/ty/issues/1274#issuecomment-3345942698) too.

## Test Plan

Add a test showing even if an unimported completion "should" (alphabetically before) come first, we favor the imported one.

### DIFF
--- a/crates/ty_completion_eval/completion-evaluation-tasks.csv
+++ b/crates/ty_completion_eval/completion-evaluation-tasks.csv
@@ -10,7 +10,7 @@ import-deprioritizes-type_check_only,main.py,1,1
 import-deprioritizes-type_check_only,main.py,2,1
 import-deprioritizes-type_check_only,main.py,3,2
 import-deprioritizes-type_check_only,main.py,4,3
-internal-typeshed-hidden,main.py,0,8
+internal-typeshed-hidden,main.py,0,5
 none-completion,main.py,0,11
 numpy-array,main.py,0,
 numpy-array,main.py,1,1
@@ -22,5 +22,5 @@ scope-prioritize-closer,main.py,0,2
 scope-simple-long-identifier,main.py,0,1
 tstring-completions,main.py,0,1
 ty-extensions-lower-stdlib,main.py,0,8
-type-var-typing-over-ast,main.py,0,4
-type-var-typing-over-ast,main.py,1,429
+type-var-typing-over-ast,main.py,0,3
+type-var-typing-over-ast,main.py,1,277
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -260,7 +260,6 @@ pub fn completion<'db>(
     if scoped.is_some() {
         add_keyword_value_completions(db, &typed_query, &mut completions);
     }
-
     if settings.auto_import {
         if let Some(scoped) = scoped {
             add_unimported_completions(
@@ -273,10 +272,8 @@ pub fn completion<'db>(
             );
         }
     }
-
     completions.sort_by(compare_suggestions);
     completions.dedup_by(|c1, c2| (&c1.name, c1.module_name) == (&c2.name, c2.module_name));
-
     completions
 }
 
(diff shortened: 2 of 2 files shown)


MatthewMckee4:
@BurntSushi I have seen that you are planning to change the sorting of completions based on scope anyway, so perhaps this isn't needed

crates/ty_ide/src/completion.rs
```diff
@@ -357,6 +358,15 @@ fn add_unimported_completions<'db>(
             documentation: None,
         });
     }
+
+    sort_completions(&mut completions);
```
MichaReiser: Is it now possible that `dedupe_by` won't deduplicate completions that are available both locally and as import or is it guaranteed that local suggestions and unimported completions never overlap?

MatthewMckee4: Yes, good point, I'll move that back to the main function

crates/ty_ide/src/completion.rs
```diff
@@ -272,8 +273,10 @@ pub fn completion<'db>(
             );
         }
     }
+
     completions.sort_by(compare_suggestions);
     completions.dedup_by(|c1, c2| (&c1.name, c1.module_name) == (&c2.name, c2.module_name));
+
```
BurntSushi: Delete these extraneous newlines?

crates/ty_completion_eval/completion-evaluation-tasks.csv
```diff
@@ -10,17 +10,17 @@ import-deprioritizes-type_check_only,main.py,1,1
...
-scope-existing-over-new-import,main.py,0,474
+scope-existing-over-new-import,main.py,0,13
 scope-prioritize-closer,main.py,0,2
 scope-simple-long-identifier,main.py,0,1
 tstring-completions,main.py,0,1
 ty-extensions-lower-stdlib,main.py,0,8
-type-var-typing-over-ast,main.py,0,3
-type-var-typing-over-ast,main.py,1,270
+type-var-typing-over-ast,main.py,0,4
+type-var-typing-over-ast,main.py,1,429
```
BurntSushi: Some of these might worth investigating.

I think we can merge as-is, particularly given the improvement here. But I should probably add snapshots as part of the eval so that changes like this are easier to review.

crates/ty_ide/src/completion.rs
```diff
@@ -887,9 +890,16 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
 /// This has the effect of putting all dunder attributes after "normal"
 /// attributes, and all single-underscore attributes after dunder attributes.
 fn compare_suggestions(c1: &Completion, c2: &Completion) -> Ordering {
-    let (kind1, kind2) = (NameKind::classify(&c1.name), NameKind::classify(&c2.name));
+    fn key<'a>(completion: &'a Completion) -> (Option<&'a ModuleName>, NameKind, bool, &'a Name) {
+        (
+            completion.module_name,
```
BurntSushi: Hmmmm. This is sorting by module name rather than the _existence_ of module name. I think we might just want `completion.module_name.is_none()` here? (That should make the `None` case sort before any `Some` case.)

BurntSushi (changes requested):

crates/ty_ide/src/completion.rs
MatthewMckee4: Do we not want `is_some`, so that None < Some (imported < unimported)?

My tests fail if i do `is_none`

BurntSushi: Yes!

BurntSushi (approved):
LGTM, thank you!

PR #21195, opened 2025-11-01
[ty] Implicit type aliases: Support for PEP 604 unions

sharkdp:
## Summary

Add support for implicit type aliases that use PEP 604 unions:
```py
IntOrStr = int | str

reveal_type(IntOrStr)  # UnionType

def _(int_or_str: IntOrStr):
    reveal_type(int_or_str)  # int | str
```

## Typing conformance

The changes are either removed false positives, or new diagnostics due to known limitations unrelated to this PR.

## Ecosystem impact

Spot checked, a mix of true positives and known limitations.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -7651,7 +7651,7 @@ fn walk_known_instance_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
         }
         KnownInstanceType::UnionType(union_type) => {
             for element in union_type.elements(db) {
-                visitor.visit_type(db, *element);
+                visitor.visit_type(db, element);
             }
         }
     }
@@ -8907,27 +8907,22 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
 #[salsa::interned(debug)]
 #[derive(PartialOrd, Ord)]
 pub struct UnionTypeInstance<'db> {
-    #[returns(deref)]
-    elements: Box<[Type<'db>; 2]>,
+    left: Type<'db>,
+    right: Type<'db>,
 }
 
 impl get_size2::GetSize for UnionTypeInstance<'_> {}
 
 impl<'db> UnionTypeInstance<'db> {
-    pub(crate) fn from_elements(
-        db: &'db dyn Db,
-        left: Type<'db>,
-        right: Type<'db>,
-    ) -> UnionTypeInstance<'db> {
-        UnionTypeInstance::new(db, Box::new([left, right]))
+    pub(crate) fn elements(self, db: &'db dyn Db) -> [Type<'db>; 2] {
+        [self.left(db), self.right(db)]
     }
 
     pub(crate) fn normalized_impl(self, db: &'db dyn Db, visitor: &NormalizedVisitor<'db>) -> Self {
-        let elements = self.elements(db);
-        UnionTypeInstance::from_elements(
+        UnionTypeInstance::new(
             db,
-            elements[0].normalized_impl(db, visitor),
-            elements[1].normalized_impl(db, visitor),
+            self.left(db).normalized_impl(db, visitor),
+            self.right(db).normalized_impl(db, visitor),
         )
     }
 }
--- a/crates/ty_python_semantic/resources/mdtest/mro.md
+++ b/crates/ty_python_semantic/resources/mdtest/mro.md
@@ -291,6 +291,20 @@ class Foo(x): ...
 reveal_mro(Foo)  # revealed: (<class 'Foo'>, Unknown, <class 'object'>)
 ```
 
+## `UnionType` instances are now allowed as a base
+
+This is not legal:
+
+```py
+class A: ...
+class B: ...
+
+EitherOr = A | B
+
+# error: [invalid-base] "Invalid class base with type `UnionType`"
+class Foo(EitherOr): ...
+```
+
 ## `__bases__` is a union of a dynamic type and valid bases
 
 If a dynamic type such as `Any` or `Unknown` is one of the elements in the union, and all other
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -8474,7 +8474,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                     Some(left_ty)
                 } else {
                     Some(Type::KnownInstance(KnownInstanceType::UnionType(
-                        UnionTypeInstance::from_elements(
+                        UnionTypeInstance::new(
                             self.db(),
                             convert_none_type(left_ty),
                             convert_none_type(right_ty),


crates/ruff_benchmark/benches/ty_walltime.rs
```diff
@@ -146,7 +146,7 @@ static FREQTRADE: Benchmark = Benchmark::new(
         max_dep_date: "2025-06-17",
         python_version: PythonVersion::PY312,
     },
-    400,
+    500,
```
sharkdp, line 149: freqtrade [makes use of a pattern](https://github.com/freqtrade/freqtrade/blob/3cfa366ec9c676abf9da0b3c40d8f3eb43d1e2d2/freqtrade/rpc/telegram.py#L549C1-L549C100) where a union of typeddicts is narrowed by using a special tag (`msg["type"] == RPCMessageType.ENTRY`). We don't support this yet, leading to ~100 new diagnostics in a single file (but nowhere else in the ecosystem).

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -17,6 +17,157 @@ def f(x: MyInt):
...
+MyNone = None
+
+# TODO: this should not be an error
+# error: [invalid-type-form] "Variable of type `None` is not allowed in a type expression"
+def g(x: MyNone):
+    # TODO: this should be `None`
+    reveal_type(x)  # revealed: Unknown
+
+g(None)
+```
```
sharkdp, line 32: This is unrelated to unions, just something I thought of while writing the tests below. Probably not super important to support :smile:

crates/ty_python_semantic/src/types.rs
```diff
@@ -6572,9 +6580,6 @@ impl<'db> Type<'db> {
                 Some(KnownClass::GenericAlias) => Ok(todo_type!(
                     "Support for `typing.GenericAlias` instances in type expressions"
                 )),
-                Some(KnownClass::UnionType) => Ok(todo_type!(
-                    "Support for `types.UnionType` instances in type expressions"
-                )),
```
sharkdp, line 6594: These are now hard errors. Opaque instances of `UnionType` can not be used in a type expression (see corresponding test)

crates/ty_python_semantic/src/types.rs
```diff
@@ -7763,6 +7779,7 @@ impl<'db> KnownInstanceType<'db> {
                             constraints.display(self.db)
                         )
                     }
+                    KnownInstanceType::UnionType(_) => f.write_str("UnionType"),
```
sharkdp, line 7782: For consistency with the branches above, I would like to change this to `types.UnionType`, but I will do that in a separate step, because it causes many downstream changes.

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -170,7 +170,8 @@ impl<'db> ClassBase<'db> {
                 | KnownInstanceType::TypeVar(_)
                 | KnownInstanceType::Deprecated(_)
                 | KnownInstanceType::Field(_)
-                | KnownInstanceType::ConstraintSet(_) => None,
+                | KnownInstanceType::ConstraintSet(_)
+                | KnownInstanceType::UnionType(_) => None,
```
sharkdp, line 174: This is different from the `Type::Union` branch above. `UnionType` instances are not allowed as bases.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
AlexWaygood: It's a one-line fix 😆

````diff
diff --git a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
index 904921e7b3..7e7f2fd59f 100644
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -22,11 +22,8 @@ f(1)
 ```py
 MyNone = None
 
-# TODO: this should not be an error
-# error: [invalid-type-form] "Variable of type `None` is not allowed in a type expression"
 def g(x: MyNone):
-    # TODO: this should be `None`
-    reveal_type(x)  # revealed: Unknown
+    reveal_type(x)  # revealed: None
 
 g(None)
 ```
index 7d4bf07d97..aaa120f053 100644
@@ -6565,6 +6565,7 @@ impl<'db> Type<'db> {
             Type::Dynamic(_) => Ok(*self),
 
             Type::NominalInstance(instance) => match instance.known_class(db) {
+                Some(KnownClass::NoneType) => Ok(Type::none(db)),
                 Some(KnownClass::TypeVar) => Ok(todo_type!(
                     "Support for `typing.TypeVar` instances in type expressions"
                 )),
````

may as well just fix it in this PR?

Comment (sharkdp) [reply crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md:32]: I was worried about a case like
```py
def _(n: None):
    x: int | n = 1
```
which shouldn't actually be allowed? I'll propose this as a separate change once this lands.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -17,6 +17,209 @@ def f(x: MyInt):
...
+
+## Stringified annotations?
+
+From the [typing spec on type aliases](https://typing.python.org/en/latest/spec/aliases.html):
+
+> Type aliases may be as complex as type hints in annotations – anything that is acceptable as a
+> type hint is acceptable in a type alias
+
+However, no other type checker seems to support stringified annotations in implicit type aliases. We
+currently also do not support them:
```
AlexWaygood, line 197: correct -- this was one of the primary motivations for PEP 613

crates/ty_python_semantic/src/types.rs
```diff
@@ -8882,6 +8899,39 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
...
+/// An instance of `types.UnionType`.
+///
+/// # Ordering
+/// Ordering is based on the context's salsa-assigned id and not on its values.
+/// The id may change between runs, or when the context was garbage collected and recreated.
+#[salsa::interned(debug)]
+#[derive(PartialOrd, Ord)]
+pub struct UnionTypeInstance<'db> {
+    #[returns(deref)]
+    elements: Box<[Type<'db>; 2]>,
```
AlexWaygood: what's the motivation for using a boxed array rather than just two separate fields?

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 7d4bf07d97..a7dc98d47c 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -7651,7 +7651,7 @@ fn walk_known_instance_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
         }
         KnownInstanceType::UnionType(union_type) => {
             for element in union_type.elements(db) {
-                visitor.visit_type(db, *element);
+                visitor.visit_type(db, element);
             }
         }
     }
@@ -8907,27 +8907,22 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
 #[salsa::interned(debug)]
 #[derive(PartialOrd, Ord)]
 pub struct UnionTypeInstance<'db> {
-    #[returns(deref)]
-    elements: Box<[Type<'db>; 2]>,
+    left: Type<'db>,
+    right: Type<'db>,
 }
 
 impl get_size2::GetSize for UnionTypeInstance<'_> {}
 
 impl<'db> UnionTypeInstance<'db> {
-    pub(crate) fn from_elements(
-        db: &'db dyn Db,
-        left: Type<'db>,
-        right: Type<'db>,
-    ) -> UnionTypeInstance<'db> {
-        UnionTypeInstance::new(db, Box::new([left, right]))
+    pub(crate) fn elements(self, db: &'db dyn Db) -> [Type<'db>; 2] {
+        [self.left(db), self.right(db)]
     }
 
     pub(crate) fn normalized_impl(self, db: &'db dyn Db, visitor: &NormalizedVisitor<'db>) -> Self {
-        let elements = self.elements(db);
-        UnionTypeInstance::from_elements(
+        UnionTypeInstance::new(
             db,
-            elements[0].normalized_impl(db, visitor),
-            elements[1].normalized_impl(db, visitor),
+            self.left(db).normalized_impl(db, visitor),
+            self.right(db).normalized_impl(db, visitor),
         )
     }
 }
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index faa3b3cac8..5330b2830b 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -8474,7 +8474,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                     Some(left_ty)
                 } else {
                     Some(Type::KnownInstance(KnownInstanceType::UnionType(
-                        UnionTypeInstance::from_elements(
+                        UnionTypeInstance::new(
                             self.db(),
                             convert_none_type(left_ty),
                             convert_none_type(right_ty),
```

crates/ty_python_semantic/src/types/instance.rs
```diff
@@ -95,6 +95,14 @@ impl<'db> Type<'db> {
         }
     }
 
+    /// Return `true` if `self` is a nominal instance of the given known class.
+    pub(crate) fn is_instance_of(self, db: &'db dyn Db, known_class: KnownClass) -> bool {
+        match self {
+            Type::NominalInstance(instance) => instance.class(db).is_known(db, known_class),
+            _ => false,
+        }
+    }
```
AlexWaygood, line 104: do we also need to add a branch for type aliases here?

crates/ty_python_semantic/src/types/class_base.rs
AlexWaygood: can we add an explicit test for that?

AlexWaygood (commented):
awesome!

crates/ty_python_semantic/src/types.rs
sharkdp: This struct was initially not interned, so I needed a `Box` anyway. But now that its interned anyway, I should go back to what I had previously, thanks.

crates/ty_python_semantic/src/types/instance.rs
sharkdp: I was mainly trying to replace the previous `ty.as_nominal_instance().is_some_and(|instance| instance.class(db).is_known(…))` pattern, so.. no?

AlexWaygood: yes... I have been wondering if we need to update all of our `as_*` methods along these lines, though... e.g.

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index e79282a35d..f094763fc5 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1013,9 +1013,10 @@ impl<'db> Type<'db> {
         any_over_type(db, self, &|ty| matches!(ty, Type::TypeVar(_)), false)
     }
 
-    pub(crate) const fn as_class_literal(self) -> Option<ClassLiteral<'db>> {
+    pub(crate) const fn as_class_literal(self, db: &'db dyn Db) -> Option<ClassLiteral<'db>> {
         match self {
             Type::ClassLiteral(class_type) => Some(class_type),
+            Type::TypeAlias(alias) => alias.value_type(db).as_class_literal(),
             _ => None,
         }
     }
```

Since in general, you should always treat a type alias to `T` the same as you treat `T` itself.

But yes, maybe we should just change all these together.

AlexWaygood:
Quite a few new diagnostics on this branch have the same cause as those I analysed as happening on `kornia` in https://github.com/astral-sh/ruff/pull/20107#issuecomment-3411347073. I.e., we now emit a diagnostic on things like this, because we now correctly understand typeshed's annotation for `isinstance()` and we correctly do not view the symbol `typing.Dict` as being an instance of `type`:

```py
import typing

isinstance({}, typing.Dict)
```

I think we should fix this, but that fixing it might need some deep special casing of `isinstance()`. We cannot simply treat `typing.Dict` as an alias to `dict`, because it is not, and has different behaviour in some situations:

```pycon
>>> import typing
>>> typing.Dict()
  File "<python-input-7>", line 1, in <module>
    typing.Dict()
    ~~~~~~~~~~~^^
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 1315, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
                    f"use {self.__origin__.__name__}() instead")
TypeError: Type Dict cannot be instantiated; use dict() instead
```

This doesn't need to be fixed in this PR, but we might want to prioritise fixing it as a followup.

AlexWaygood:
The new `invalid-type-form` errors on `trio` and `zulip` also suggest that we might want to prioritise adding support for `type[]` in implicit type aliases, e.g.

```py
X = type[int] | type[str]
y: X
```

AlexWaygood (approved):

sharkdp:
> adding support for `type[]` in implicit type aliases

Yes, I saw that. I will update the https://github.com/astral-sh/ty/issues/221 ticket with a list of things that we need to support, and put that near the top of the list.

PR #21196, opened 2025-11-02
[ty] Increase timeout-minutes to 10 for py-fuzzer job

AlexWaygood:
In https://github.com/astral-sh/ruff/pull/20937, we reduced this from 20 to 5, so that it would become obvious if ty was now taking a pathological amount of time to type-check a given seed. 5 minutes is easily long enough if there are no new bugs, but it turns out to be too low if there _are_ new bugs, because on encountering a new bug the fuzzer must repeatedly run ty on a smaller and smaller snippet in order to try to minimize the reproducible example. That means that the fuzzer job is timing out on https://github.com/astral-sh/ruff/pull/20962, not because of pathological performance issues but because we're not giving the script enough time to minimize the bugs that it's found.

Hopefully 10 should be a good compromise where it still becomes obvious if we introduce pathological performance issues on certain fuzzer seeds, but we give the fuzzer enough time to minimize examples when it finds new bugs.

A better solution to find new pathological performance issues on fuzzer seeds might be to just compare the execution time between the baseline-executable run on a given seed and the test-executable run on the same seed. This is just a quick-and-easy fix for now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -277,8 +277,8 @@ jobs:
         run: cargo test -p ty_python_semantic --test mdtest || true
       - name: "Run tests"
         run: cargo insta test --all-features --unreferenced reject --test-runner nextest
-      # Dogfood ty on py-fuzzer
-      - run: uv run --project=./python/py-fuzzer cargo run -p ty check --project=./python/py-fuzzer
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):

sharkdp:
Thanks

PR #21197, opened 2025-11-02
Use system jemalloc

WhyNotHugo:
The version compiled by tikv_jemallocator by default assumes 16K pages, and doesn't work on systems which have 64K pages.

Link the system jemalloc to ensure we ruff always uses a compatible jemalloc which aligns with local system conventions.

We're currently shipping this patch downstream as otherwise builds simply don't work on systems with 64K pages.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,3 +8,7 @@ benchmark = "bench -p ruff_benchmark --bench linter --bench formatter --"
 # See: https://github.com/astral-sh/ruff/issues/11503
 [target.'cfg(all(target_env="msvc", target_os = "windows"))']
 rustflags = ["-C", "target-feature=+crt-static"]
+
+# Use system jemalloc (which may or may not be compiled for 64K pages).
+[target.'cfg(all(target_env = "musl", target_os = "linux"))'.jemalloc]
+rustc-link-lib = ["jemalloc"]


MichaReiser:
I'm far from an expert on this but the way we solve this in our own built pipeline is by setting `JEMALLOC_SYS_WITH_LG_PAGE=16`. We haven't had any issues since we increased the page size.

https://github.com/astral-sh/ruff/blob/bff32a41dc440b30764e9414767794e01d25c265/.github/workflows/build-binaries.yml#L280

WhyNotHugo:
That workaround only affects builds made by CI, not builds made downstream with release tarballs.

WhyNotHugo:
Configuring the build system in releases to use `JEMALLOC_SYS_WITH_LG_PAGE=16` would also work, but most distributions prefer using the system jemalloc regardless, so doing that is likely a better fit.

MichaReiser:
but what if the target system doesn't have jemalloc installed. Wouldn't the build fail in that case?

WhyNotHugo:
Hang on though, looks like this final syntax isn't working the same as my previous approach.

WhyNotHugo:
The following works:

Okay, this works:

    [target.x86_64-alpine-linux-musl.jemalloc]

But this does not:

    [target.'cfg(all(target_env = "musl", target_os = "linux"))'.jemalloc]

I'm not entirely sure why. Maybe we'll need another approach to address this.

WhyNotHugo:
> but what if the target system doesn't have jemalloc installed. Wouldn't the build fail in that case?

Right, jemalloc would need to be installed like all other build dependencies. I don't think there's any musl-based distribution which doesn't ship jemalloc.

MichaReiser:
I'll close this as the PR has been stale for a while. Please feel free to submit a new PR if you'd like to keep working on this.

Issue #14453, opened 2024-11-19
B901: Misses `yield` sub-expresisons


MichaReiser:
A very contrived example but `B901` misses yield expressions where they're not the expression of a `ExprStmt`. 

```py
def get_file_paths(file_types: Iterable[str] | None = None) -> Iterable[Path]:
    dir_path = Path(".")
    if file_types is None:
        return dir_path.glob("*")

    for file_type in file_types:
        (yield a for a in dir_path.glob(f"*.{file_type}"))
```

Note: we should make sure that we don't traverse into any lambda expressions...

MichaReiser:
hmm, I only tested with our parser. nested yield expressions are definitely valid :)

MichaReiser:
Okay, found one

```
async def main():
    await (yield x)
```

AlexWaygood:
Here's another function that is a generator function with a `return` statement and is not flagged by B901:

```py
def f():
    x = yield
    print(x)
    return 42
```

This is a generator function that you can `send` values into:

```pycon
>>> def f():
...     x = yield
...     print(x)
...     return 42
...     
>>> y = f()
>>> next(y)
>>> y.send('foo')
foo
Traceback (most recent call last):
  File "<python-input-13>", line 1, in <module>
    y.send('foo')
    ~~~~~~^^^^^^^
StopIteration: 42
```

`Send`ing valued _into_ a generator is, like `return`ing values from a generator, quite an advanced feature. So if I saw a function like this in code review, I would assume the user knew what they were doing. That doesn't feel like a very principled reason not to emit B901 on it, though; maybe we should. Not sure.

rabelmervin:
Hi @MichaReiser @AlexWaygood  can I help you fix this issue ?

MichaReiser:
Sure. You can find the code for this rule in this file

https://github.com/astral-sh/ruff/blob/b7e32b0a18487e8c9813e0d189d7a5d625ff5c3f/crates/ruff_linter/src/rules/flake8_bugbear/rules/return_in_generator.rs

rabelmervin:
Thanks @MichaReiser excited to work on this !

MichaReiser:
We decided to leave this as is for now because there are some valid patterns that would be flagged if we change the rule to detect all yield usages. See the linked PR

ntBre:
[...]

PR #15254, opened 2025-01-04
[flake8-bugbear] Catch yield in subexpressions (B901) (#14453)

kaspell:
## Summary

Currently, the B901 rule misses yield expressions that are not top-of-tree in a function body. Refactor the rule to find such yield expressions, but otherwise try to follow the flake8-bugbear implementation. Do not search for yield or yield from expressions in assignment statements, or in lambda or function call expressions.

Closes #14453.

## Test Plan

`cargo test`. Didn't yet manage to get the ecosystem tests to run locally.

PR #15101, opened 2024-12-22
B901 catches yields in subexprs and more sensitive B901

guptaarnav:
## Summary
Closes #14453 by updating the implementation of the B901 lint rule.  The changes ensure that:
	1.	The visitor now traverses all expressions within statements, ensuring yield expressions embedded are correctly flagged even if they are not the expression of an `ExprStmt`.
	2.     The rule now correctly identifies yield and yield from expressions when they appear as the right-hand side of assignment statements (e.g., x = yield or x = yield from []); this catches the case mentioned in the original issue by @AlexWaygood.
	3.     We still do not recurse into lambda expressions or nested function defs as these are evaluated separately.

Possible point of contention:
For the purposes of deciding whether or not to flag B901, the previous code basically treats `yield from []` as if it isn't a yield statement.  e.g. the following two examples did not flag B901 in the B901 snapshot tests:
```python
def broken3():
    return (yield from [])


def broken4():
    x = yield from []
    return x
```
In my opinion, code that yields from an empty iterable (along with a return) should flag B901, and I have modified it so that it does (and accordingly updated the tests).  The description of B901 in the docs say the following:

> The combination of yield and return can lead to confusing behavior, as the return statement will cause the generator to raise StopIteration with the value provided, rather than returning the value to the caller.

To me, this implies that regardless of the fact that `yield from []` is very predictable given the explicit empty iterable, that doesn't sidestep the lack of clarity present given both `return` and `yield from` being used in a single function.  If a user really wants to do this, they should have to suppress B901 imo.  

## Test Plan
I added the tests mentioned by @MichaReiser and @AlexWaygood in the original issue into `test/fixtures/flake8_bugbear/B901.py` and updated the related snapshot.

PR #21200, opened 2025-11-02
[`flake8-bugbear`] Catch `yield` expressions within other statements (`B901`)

11happy:
## Summary

This PR re-implements [return-in-generator (B901)](https://docs.astral.sh/ruff/rules/return-in-generator/#return-in-generator-b901) for async generators as a semantic syntax error. This is not a syntax error for sync generators, so we'll need to preserve both the lint rule and the syntax error in this case.

It also updates B901 and the new implementation to catch cases where the generator's `yield` or `yield from` expression is part of another statement, as in:

```py
def foo():
    return (yield)
```

These were previously not caught because we only looked for `Stmt::Expr(Expr::Yield)` in `visit_stmt` instead of visiting `yield` expressions directly. I think this modification is within the spirit of the rule and safe to try out since the rule is in preview.

## Test Plan

I have written tests as directed in #17412


11happy:
Hii @ntBre would appreciate your views on some design decisions for this PR , 
- I am currently detecting return statements like this, 
  ```
  body.iter().any(|stmt| matches!(stmt, Stmt::Return(_)))
  ```
  I think it migh miss the nested ones, should I also use visitor pattern like in original `B901` 
- also if you have any suggestions for the error message, currently kept it short.
- right now its pointing to yield should it point to return ?

Thank you : )

ntBre:
> Hii @ntBre would appreciate your views on some design decisions for this PR ,
> 
> * I am currently detecting return statements like this,
>   ```
>   body.iter().any(|stmt| matches!(stmt, Stmt::Return(_)))
>   ```
>   
>   I think it migh miss the nested ones, should I also use visitor pattern like in original `B901`
> * also if you have any suggestions for the error message, currently kept it short.
> * right now its pointing to yield should it point to return ?
> 
> Thank you : )


Yes, I think we should use a visitor like in B901. We can't fully replace B901 since it also applies to sync functions, but we should make sure we catch all of the async cases B901 handles.

I think the error message looks good currently and matches CPython, but I think it should point to `return` rather than `yield`, also like CPython and B901.

11happy:
its ready for review : )
Thank you

crates/ruff_linter/resources/test/fixtures/syntax_errors/return_in_generator.py
ntBre, line 1: Could we add a few more test cases here? As a couple of examples, this is not a syntax error for non-async functions:

```py
def gen():
    yield 1
    return 42
```

and it's also not a syntax error for `return` without a value:

```py
async def gen():
    yield 1
    return
```

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -838,6 +864,24 @@ impl SemanticSyntaxContext for Checker<'_> {
         )
     }
 
+    fn has_return(&self) -> Option<TextRange> {
```
ntBre: Does this need to be a `SemanticSyntaxContext` method? From looking at B901, it seems that it runs on `StmtFunctionDef`s, not on `yield` statements. I think we could do the same in `SemanticSyntaxChecker::visit_stmt`. That would be a bit simpler and we wouldn't have to implement methods for Ruff and ty separately.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1572,6 +1590,8 @@ pub enum SemanticSyntaxErrorKind {
 
     /// Represents a nonlocal statement for a name that has no binding in an enclosing scope.
     NonlocalWithoutBinding(String),
+
+    ReturnInAsyncGenerator,
```
ntBre: Could we add some docs here? I think it's especially important to capture the "with value" part since it's not in the name.

ntBre:
Oh, one other thing I forgot to mention: I think we need to make B901 only run on sync generators and this syntax error only run on async generators. Then we'll need to convert the syntax error into a B901 error for now to preserve backwards compatibility.

crates/ruff_python_parser/src/semantic_errors.rs

crates/ruff_linter/src/checkers/ast/mod.rs

crates/ruff_linter/resources/test/fixtures/syntax_errors/return_in_generator.py
11happy: done added more cases

crates/ruff_linter/src/rules/flake8_bugbear/rules/return_in_generator.rs
```diff
@@ -96,6 +96,10 @@ pub(crate) fn return_in_generator(checker: &Checker, function_def: &StmtFunction
         return;
     }
 
+    if function_def.is_async {
```
ntBre, line 99: small nit: It might be worth a comment here like:


```suggestion
    // Async functions are flagged by the `ReturnInAsyncGenerator` semantic syntax error.
    if function_def.is_async {
```

crates/ruff_linter/src/linter.rs
```diff
@@ -980,6 +980,7 @@ mod tests {
     #[test_case(Path::new("write_to_debug.py"), PythonVersion::PY310)]
     #[test_case(Path::new("invalid_expression.py"), PythonVersion::PY312)]
     #[test_case(Path::new("global_parameter.py"), PythonVersion::PY310)]
+    #[test_case(Path::new("return_in_async_generator.py"), PythonVersion::PY310)]
```
ntBre: nit: I think we should maybe use `PythonVersion::default()` or `PythonVersion::latest()` unless we have a specific reason to choose a version. I think most of our tests run with `latest`.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -822,6 +837,7 @@ impl SemanticSyntaxChecker {
                     // def f(): yield *x
                     Self::invalid_star_expression(value, ctx);
                 }
+
```
ntBre: nit: can we revert this?


```suggestion
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1688,6 +1710,36 @@ impl Visitor<'_> for ReboundComprehensionVisitor<'_> {
...
+impl StatementVisitor<'_> for ReturnVisitor {
+    fn visit_stmt(&mut self, stmt: &Stmt) {
+        match stmt {
+            Stmt::Expr(ast::StmtExpr { value, .. }) => match **value {
+                Expr::Yield(_) | Expr::YieldFrom(_) => {
+                    self.has_yield = true;
+                }
+                _ => {}
+            },
+            Stmt::FunctionDef(_) | Stmt::ClassDef(_) => {}
```
ntBre, line 1769: I think it's probably worth copying the comment from B901:


```suggestion
            // Do not recurse into nested functions; they're evaluated separately.
            Stmt::FunctionDef(_) | Stmt::ClassDef(_) => {}
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1688,6 +1710,36 @@ impl Visitor<'_> for ReboundComprehensionVisitor<'_> {
...
+            Stmt::Return(ast::StmtReturn {
+                value: Some(_),
+                range,
+                ..
+            }) => {
+                if self.return_range.is_none() {
+                    self.return_range = Some(*range);
+                }
```
ntBre: I like that this would take the first return like CPython does rather than the last return like B901, but I think for the sake of compatibility with B901, we should just take the last return and always overwrite:


```suggestion
                self.return_range = Some(*range);
```

crates/ruff_linter/src/checkers/ast/mod.rs
```diff
@@ -738,6 +738,7 @@ impl SemanticSyntaxContext for Checker<'_> {
             | SemanticSyntaxErrorKind::InvalidExpression(..)
             | SemanticSyntaxErrorKind::GlobalParameter(_)
             | SemanticSyntaxErrorKind::DuplicateMatchKey(_)
+            | SemanticSyntaxErrorKind::ReturnInAsyncGenerator
```
ntBre: We need to convert this into a B901 error right? It looks like we actually don't have any B901 test cases with async functions (that trigger the lint), otherwise I would have expected tests to fail. We should probably add one of those as well.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -3,17 +3,18 @@
...
-    StmtImportFrom,
+    StmtFunctionDef, StmtImportFrom,
     comparable::ComparableExpr,
     helpers,
     visitor::{Visitor, walk_expr},
 };
 use ruff_text_size::{Ranged, TextRange, TextSize};
 use rustc_hash::{FxBuildHasher, FxHashSet};
 use std::fmt::Display;
-
```
ntBre, line 16: Let's keep this empty line.

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -3,17 +3,18 @@
 //! This checker is not responsible for traversing the AST itself. Instead, its
 //! [`SemanticSyntaxChecker::visit_stmt`] and [`SemanticSyntaxChecker::visit_expr`] methods should
 //! be called in a parent `Visitor`'s `visit_stmt` and `visit_expr` methods, respectively.
+use ruff_python_ast::statement_visitor;
+use ruff_python_ast::statement_visitor::StatementVisitor;
```
ntBre: nit:

```suggestion
use ruff_python_ast::statement_visitor::{self, StatementVisitor};
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1688,6 +1710,35 @@ impl Visitor<'_> for ReboundComprehensionVisitor<'_> {
...
+
+impl StatementVisitor<'_> for ReturnVisitor {
+    fn visit_stmt(&mut self, stmt: &Stmt) {
+        match stmt {
+            Stmt::Expr(ast::StmtExpr { value, .. }) => match **value {
+                Expr::Yield(_) | Expr::YieldFrom(_) => {
+                    self.has_yield = true;
+                }
+                _ => {}
+            },
```
ntBre: I noticed this while reviewing a different PR (https://github.com/astral-sh/ruff/pull/21177#discussion_r2539702728), but I think this logic is actually incorrect here and in B901 itself. This only catches `yield` and `yield from` in standalone statements, but the syntax error applies even when these expressions are nested within other statements:

```pycon
>>> async def foo():
...     return (yield 1)
...
  File "<python-input-1>", line 2
    return (yield 1)
    ^^^^^^^^^^^^^^^^
SyntaxError: 'return' with value in async generator
```

That's one compact example that won't be flagged here.

The fix is to implement both `visit_stmt` and `visit_expr` and to perform the `yield` checks in `visit_expr`, as in PLR1708:

https://github.com/astral-sh/ruff/blob/0e651b50b7d0c2614c5958e78525213e759e5b41/crates/ruff_linter/src/rules/pylint/rules/stop_iteration_return.rs#L99-L100

We also have to call `walk_stmt` in the `Return` branch, or this case will still not be caught.

crates/ruff_linter/src/linter.rs
```diff
@@ -1042,6 +1041,7 @@ mod tests {
         Rule::YieldFromInAsyncFunction,
         Path::new("yield_from_in_async_function.py")
     )]
+    #[test_case(Rule::ReturnInGenerator, Path::new("return_in_generator.py"))]
```
ntBre, line 1046: I think it's okay to leave this in the `test_semantic_errors` test case. It's a bit confusing to see the non-async errors in the snapshot here, especially with the `ok` comments.

I think we should just add 1 or 2 `async` test cases to `B901.py` to make sure the syntax error is being converted correctly for now.

crates/ruff_python_parser/src/semantic_errors.rs

crates/ruff_linter/src/linter.rs
11happy: done

ntBre: Oh I see, the error doesn't trigger at all in `test_semantic_errors` because we map it back to the rule. I will restore the `test_syntax_errors` version that you had initially in that case and just clarify the `ok` comments.

I think it's still nice to have the B901 tests you added :+1:

crates/ruff_python_parser/src/semantic_errors.rs
ntBre: After this change, we didn't need the `Stmt::Expr` part anymore. I also copied this change over to B901 proper.

For future reference, I also checked the blame for the "not broken" B901 tests that this changed. They were part of the original PR adding the rule (https://github.com/astral-sh/ruff/pull/11644) and were copied from the upstream linter. However, I think the upstream tests were added incorrectly in https://github.com/PyCQA/flake8-bugbear/commit/05002b5cc7b56f06ff2f84a133339eedb0f470f2 when the real fix was to avoid flagging the error in `__await__` methods, as we already do. As I noted in one of my commit messages, these cases _are_ syntax errors in async generators, so I think we should be flagging them.

ntBre (approved):
Thanks! I pushed a few commits. Just waiting for CI and then I'll merge!

ntBre:
Hmm, there are actually 35 new ecosystem hits. They all look like true positives to me, and this is a preview rule, so I think it's okay to accept this, but @amyreese, do you know if code like this is common enough that we should allow it?

```py
@pytest.hookimpl(wrapper=True, tryfirst=True)
def pytest_collection(session: Session) -> Generator[None, object, object]:
    config = session.config
    with catch_warnings_for_item(
        config=config, ihook=config.hook, when="collect", item=None
    ):
        return (yield)  # <--
```

It popped up many times in `pytest`, but maybe they just wouldn't want to enable this rule.

It seems to me that the motivation for the rule applies whether the yield is a standalone expression or not. But we should probably include this in the release notes, either by splitting out the visitor change/bug fix or at least by updating the title and labels.

For a bit more context, my suggestion to start flagging these was in https://github.com/astral-sh/ruff/pull/21200#discussion_r2578909298.

amyreese:
I've literally never seen `return (yield)` in the wild before today. 😅

Since most of the ecosystem hits seem related to pytest hooks, I think it would be ok as-is — it's a really weird pattern and seems to negate most of the "purpose" of a generator function. 

I presume that pytest is doing something like sending a value back into a generator and then capturing that in the `StopIteration` when it returns, and I suspect they would already have B901 disabled.

amyreese:
2.2k hits on github, some of which at least seem mildly more useful than what I saw in pytest: https://github.com/search?q=%22return+%28yield%29%22&type=code

I can't imagine people are actually using the actual value from their yield expressions as a return value, probably more as shorthand for `yield X; return`.

I guess as exceptions to a rule go, it's not crazy to allow it, just weird.

ntBre:
Thanks! Sounds like it's probably reasonable to land this, especially since the rule is still in preview. I'll try to update the title and labels so that it shows up in the release notes.

ntBre:
Hmm, these benchmarks don't seem right. I'll try to merge main again.

AlexWaygood:
> Hmm, these benchmarks don't seem right. I'll try to merge main again.

The pydantic benchmark seems to fairly regularly vary around 5% or so since we landed https://github.com/astral-sh/ruff/pull/21467, unfortunately :/

ntBre:
Hmm maybe I should have searched around a bit more. I stumbled across https://github.com/astral-sh/ruff/issues/14453 and https://github.com/astral-sh/ruff/pull/15101 and https://github.com/astral-sh/ruff/pull/15254 for unrelated reasons this afternoon. Apparently this change has already been considered and rejected.

I still think it makes sense to flag these, including after checking the upstream implementation (https://github.com/astral-sh/ruff/pull/21200#discussion_r2583138174), but I guess we should either revert the visitor changes (not the syntax error part) like #15254 or close #14453 as now completed. As I say in my other comment, we at least have to visit sub-expressions in the async case because that's a syntax error:

```pycon
>>> async def foo():
...     return (yield 1)
...
  File "<python-input-0>", line 2
    return (yield 1)
    ^^^^^^^^^^^^^^^^
SyntaxError: 'return' with value in async generator
```

but I think that bolsters the sync case too.

cc @MichaReiser who reviewed the earlier PR

PR #21207, opened 2025-11-02
[ty] support subscripting typing.Literal with a type alias

carljm:
Fixes https://github.com/astral-sh/ty/issues/1368

## Summary

Add support for patterns like this, where a type alias to a literal type (or union of literal types) is used to subscript `typing.Literal`:

```py
type MyAlias = Literal[1]
def _(x: Literal[MyAlias]): ...
```

This shows up in the ecosystem report for PEP 613 type alias support.

One interesting case is an alias to `bool` or an enum type. `bool` is an equivalent type to `Literal[True, False]`, which is a union of literal types. Similarly an enum type `E` is also equivalent to a union of its member literal types. Since (for explicit type aliases) we infer the RHS directly as a type expression, this makes it difficult for us to distinguish between `bool` and `Literal[True, False]`, so we allow either one to (or an alias to either one) to appear inside `Literal`, where other type checkers allow only the latter.

I think for implicit type aliases it may be simpler to support only types derived from actually subscripting `typing.Literal`, though, so I didn't make a TODO-comment commitment here.

## Test Plan

Added mdtests, including TODO-filled tests for PEP 613 and implicit type aliases.

### Conformance suite

All changes here are positive -- we now emit errors on lines that should be errors. This is a side effect of the new implementation, not the primary purpose of this PR, but it's still a positive change.

### Ecosystem

Eliminates one ecosystem false positive, where a PEP 695 type alias for a union of literal types is used to subscript `typing.Literal`.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -259,7 +259,7 @@ class Color(Enum):
     RED = "red"
 
 f: dict[list[Literal[1]], list[Literal[Color.RED]]] = {[1]: [Color.RED, Color.RED]}
-reveal_type(f)  # revealed: dict[list[Literal[1]], list[Literal[Color.RED]]]
+reveal_type(f)  # revealed: dict[list[Literal[1]], list[Color]]
 
 class X[T]:
     def __init__(self, value: T): ...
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -259,7 +259,7 @@ class Color(Enum):
     RED = "red"
 
 f: dict[list[Literal[1]], list[Literal[Color.RED]]] = {[1]: [Color.RED, Color.RED]}
-reveal_type(f)  # revealed: dict[list[Literal[1]], list[Literal[Color.RED]]]
+reveal_type(f)  # revealed: dict[list[Literal[1]], list[Color]]
```
carljm, line 262: Our previous support for enum member literals in `typing.Literal` did not implement the simplification of a single-member enum literal type to the overall enum type. Since we do implement this simplification otherwise, I thought it more consistent to do it in this case also. (And it falls out from the implementation, which now simply uses general type inference rather than special handling for enum members.)

AlexWaygood (approved):

PR #21210, opened 2025-11-03
[ty] Improve generic call expression inference

ibraheemdev:
## Summary

Implements https://github.com/astral-sh/ty/issues/1356 and https://github.com/astral-sh/ty/issues/136#issuecomment-3413669994.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -456,7 +456,7 @@ j: int | str = f2(True)
 reveal_type(j)  # revealed: Literal[True]
 ```
 
-The function arguments are also inferred using the type context:
+A function's arguments are also inferred using the type context:
 
 `typed_dict.py`:
 
@@ -571,6 +571,57 @@ g: list[Any] | dict[Any, Any] = f3(1)
 reveal_type(g)  # revealed: list[Literal[1]] | dict[Literal[1], Literal[1]]
 ```
 
+We currently prefer the generic declared type regardless of its variance:
+
+```py
+class Bivariant[T]:
+    pass
+
+class Covariant[T]:
+    def pop(self) -> T:
+        raise NotImplementedError
+
+class Contravariant[T]:
+    def push(self, value: T) -> None:
+        pass
+
+class Invariant[T]:
+    x: T
+
+def bivariant[T](x: T) -> Bivariant[T]:
+    return Bivariant()
+
+def covariant[T](x: T) -> Covariant[T]:
+    return Covariant()
+
+def contravariant[T](x: T) -> Contravariant[T]:
+    return Contravariant()
+
+def invariant[T](x: T) -> Invariant[T]:
+    return Invariant()
+
+x1 = bivariant(1)
+x2 = covariant(1)
+x3 = contravariant(1)
+x4 = invariant(1)
+
+reveal_type(x1)  # revealed: Bivariant[Literal[1]]
+reveal_type(x2)  # revealed: Covariant[Literal[1]]
+reveal_type(x3)  # revealed: Contravariant[Literal[1]]
+reveal_type(x4)  # revealed: Invariant[Literal[1]]
+
+x5: Bivariant[Any] = bivariant(1)
+x6: Covariant[Any] = covariant(1)
+x7: Contravariant[Any] = contravariant(1)
+x8: Invariant[Any] = invariant(1)
+
+# TODO: This could reveal `Bivariant[Any]`.
+reveal_type(x5)  # revealed: Bivariant[Literal[1]]
+reveal_type(x6)  # revealed: Covariant[Any]
+reveal_type(x7)  # revealed: Contravariant[Any]
+reveal_type(x8)  # revealed: Invariant[Any]
+```
+
 ## Narrow generic unions
 
 ```toml
@@ -581,30 +632,30 @@ python-version = "3.12"
 ```py
 from typing import reveal_type, TypedDict
 
-def id[T](x: T) -> T:
+def identity[T](x: T) -> T:
     return x
 
 def _(narrow: dict[str, str], target: list[str] | dict[str, str] | None):
-    target = id(narrow)
+    target = identity(narrow)
     reveal_type(target)  # revealed: dict[str, str]
 
 def _(narrow: list[str], target: list[str] | dict[str, str] | None):
-    target = id(narrow)
+    target = identity(narrow)
     reveal_type(target)  # revealed: list[str]
 
 def _(narrow: list[str] | dict[str, str], target: list[str] | dict[str, str] | None):
-    target = id(narrow)
+    target = identity(narrow)
     reveal_type(target)  # revealed: list[str] | dict[str, str]
 
 class TD(TypedDict):
     x: int
 
 def _(target: list[TD] | dict[str, TD] | None):
-    target = id([{"x": 1}])
+    target = identity([{"x": 1}])
     reveal_type(target)  # revealed: list[TD]
 
 def _(target: list[TD] | dict[str, TD] | None):
-    target = id({"x": {"x": 1}})
+    target = identity({"x": {"x": 1}})
     reveal_type(target)  # revealed: dict[str, TD]
 ```
 
@@ -616,7 +667,7 @@ python-version = "3.12"
 ```
 
 ```py
-def id[T](x: T) -> T:
+def identity[T](x: T) -> T:
     return x
 
 def lst[T](x: T) -> list[T]:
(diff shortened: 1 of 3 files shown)


ibraheemdev:
I imagine we could avoid the performance regression by special casing `x: ... | None = f()` to avoid the inference attempt using only `None` as the type context, but I'm a little hesitant to make that change, as there are (rare) cases where we might want to narrow to `None`. Most of the regressions seem relatively reasonable, other than `colour`.

Also note that a lot of the diagnostic improvements won't be seen in this PR directly, but https://github.com/astral-sh/ruff/pull/20933 (which is currently stacked on top of this PR).

ibraheemdev:
Most of the performance regressions have been resolved. There is an unavoidable regression when assigning the result of a generic call to a union, as we have to attempt inference multiple times for each element of the union.

carljm:
Can you take a look at the new diagnostics in the ecosystem report? Just taking a quick glance, I see in HomeAssistant a few occurrences of "object of type `Any` is not assignable ..." which definitely doesn't look right.

ibraheemdev:
It looks like those failures are due to `TypedDict` being assignable to `None`, which confuses overload evaluation. The weird diagnostic should be fixed by https://github.com/astral-sh/ruff/pull/21267, the actual type we infer (with type context) is `Any | None` there, not `Any`.

I added a failing test to track those.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -417,6 +417,8 @@ reveal_type(x) # revealed: Literal[1]
 python-version = "3.12"
 ```
 
+`generic_list.py`:
```
carljm, line 420: Nit: should these just be another heading layer? Our usual approach is to reserve filenames for tests that actually need to be multi-file (because they are testing cross-module behavior / imports) and use headings when we are just trying to group related tests together.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -455,7 +456,164 @@ j: int | str = f2(True)
...
+    return [x]
+
+def f2[T](x: T) -> list[T] | None:
+    return [x]
+
+def f3[T](x: T) -> list[T] | dict[T, T]:
+    return [x]
+
+a = f(1)
+reveal_type(a)  # revealed: list[Literal[1]]
```
carljm, line 551: We should probably do literal promotion here?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -455,7 +456,164 @@ j: int | str = f2(True)
...
+    reveal_type(x6)  # revealed: Any
+
+    x7: TD2 | None = dt.get(key, {})
+    reveal_type(x7)  # revealed: Any
+
+    x8: TD2 | None = dt.get(key, {"x": 0})
+    reveal_type(x8)  # revealed: Any
+```
+
+## Prefer the declared type of generic classes
```
AlexWaygood, line 531: is this for all generic classes? or just invariant ones? All the examples in this section use invariant ones

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -455,7 +456,164 @@ j: int | str = f2(True)
 reveal_type(j)  # revealed: Literal[True]
 ```
 
-Types are not widened unnecessarily:
+The function arguments are also inferred using the type context:
```
AlexWaygood:
```suggestion
A function's arguments are also inferred using the type context:
```

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -455,7 +456,164 @@ j: int | str = f2(True)
...
+
+```toml
+[environment]
+python-version = "3.12"
+```
+
+```py
+from typing import reveal_type, TypedDict
+
+def id[T](x: T) -> T:
```
AlexWaygood: nit: there's a builtin function `id()` in Python which you're shadowing here; maybe call this `identity` to avoid confusion?

crates/ty_python_semantic/src/types.rs
```diff
@@ -884,21 +884,31 @@ impl<'db> Type<'db> {
             _ => false,
         }
     }
-
     // If the type is a specialized instance of the given `KnownClass`, returns the specialization.
```
AlexWaygood: nit

```suggestion

    /// If the type is a specialized instance of the given `KnownClass`, returns the specialization.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -884,21 +884,31 @@ impl<'db> Type<'db> {
...
         &self,
         db: &'db dyn Db,
         known_class: KnownClass,
     ) -> Option<Specialization<'db>> {
         let class_literal = known_class.try_to_class_literal(db)?;
-        self.specialization_of(db, Some(class_literal))
+        self.specialization_of(db, class_literal)
+    }
+
+    // If this type is a class instance, returns its specialization.
```
```suggestion
    /// If this type is a class instance, returns its specialization.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -884,21 +884,31 @@ impl<'db> Type<'db> {
...
+    pub(crate) fn class_specialization(self, db: &'db dyn Db) -> Option<Specialization<'db>> {
+        self.specialization_of_optional(db, None)
     }
 
     // If the type is a specialized instance of the given class, returns the specialization.
```
```suggestion
    /// If the type is a specialized instance of the given class, returns the specialization.
```

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2738,9 +2753,21 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
             return;
         };
 
+        let return_with_tcx = self
+            .signature
+            .return_ty
+            .zip(self.call_expression_tcx.annotation);
```
AlexWaygood, line 2759: TIL about `Option::zip()`

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5961,27 +5973,156 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            // Infer the argument types again with diagnostics enabled.
+            self.context.set_multi_inference(was_in_multi_inference);
+
+            self.infer_all_argument_types(
+                ast_arguments,
+                argument_types,
+                bindings,
+                narrowed_tcx,
+                MultiInferenceState::Intersect,
+            );
```
AlexWaygood: can we skip the re-inference if `was_in_multi_inference == true`? Because that means that we wouldn't actually have diagnostics enabled for this later `infer_all_argument_types()` call either. No tests fail if I apply this edit to your branch:

```suggestion
            // Successfully narrowed to an element of the union.
            self.context.set_multi_inference(was_in_multi_inference);

            // If necessary, infer the argument types again with diagnostics enabled.
            if !was_in_multi_inference {
                self.infer_all_argument_types(
                    ast_arguments,
                    argument_types,
                    bindings,
                    narrowed_tcx,
                    MultiInferenceState::Intersect,
                );
            }
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5961,27 +5973,156 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
         bindings: &Bindings<'db>,
+        call_expression_tcx: TypeContext<'db>,
+        multi_inference_state: MultiInferenceState,
     ) {
         debug_assert!(
-            ast_arguments.len() == arguments.len()
-                && arguments.len() == bindings.argument_forms().len()
+            ast_arguments.len() == arguments_types.len()
+                && arguments_types.len() == bindings.argument_forms().len()
         );
```
AlexWaygood: nit: I'd prefer to have this as two `debug_assert_eq!()` calls so it's easy to identify which condition is failing (if one of them fails)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -10763,8 +10922,15 @@ enum MultiInferenceState {
     Panic,
 
     /// Overwrite the previously inferred value.
+    ///
+    // Note that `MultiInferenceState::Overwrite` does not play well with nested
+    // expressions. Note that `Overwrite` does not interact well with nested inferences,
+    // overwriting values that were written with `MultiInferenceState::Intersect`.
```
```suggestion
    /// Note that `Overwrite` does not interact well with nested inferences: it
    /// overwrites values that were written with `MultiInferenceState::Intersect`.
```

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
ibraheemdev: This currently applies to all generic classes. I'm open to changing it, but see https://github.com/astral-sh/ruff/pull/21070#issuecomment-3445659355 for my reasoning there.

crates/ty_python_semantic/src/types/infer/builder.rs
ibraheemdev: Great idea!

carljm (commented):
Oops, I had a couple draft comments from a review I started on Friday but hadn't completed yet. Nothing important, just things to consider for future.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
AlexWaygood: that might be fixed in https://github.com/astral-sh/ruff/pull/21320 :-)

PR #21220, opened 2025-11-03
[`airflow`] extend deprecated argument `concurrency` in `airflow..DAG` (`AIR301`)

Lee-W:
## Summary

* extend AIR301 to include deprecated argument `concurrency` in `airflow....DAG`

## Test Plan

update the existing test fixture in the first commit and then reorganize in the second one


ashb (commented):
Changed LGTM from the airflow side.

PR #21230, opened 2025-11-03
[ty] Refactor `Range` to/from `TextRange` conversion as prep for notebook support

MichaReiser:
## Summary

This PR changes the signature of the `TextSize` <-> `Position` and `TextRange` <-> `Range` conversion methods in preparation for notebook support.

According to the LSP specification, each Notebook consists of many `TextDocument`s. However, ty internally concatenates all cells to a single text document. That means, offsets coming from ty need to be mapped back to their corresponding cell, and then to the offset within the cell. The same applies for LSP offsets: The cell-relative offset needs to be converted to an absolute offset into the concatenated notebook document.

This PR doesn't implement the logic itself for mapping between the different offsets but it refactors the call-sites and the methods used for those conversions to add the required logic in a follow-up PR.

The most important change is that we now pass `db` and `File` to the conversion methods. This will allow us to query the notebook to resolve the cell metadata (offset per cell, etc).

## Test Plan

Existing tests. I tested different LSP features and they seem to work fine

### DIFF
--- a/crates/ty_server/src/document/range.rs
+++ b/crates/ty_server/src/document/range.rs
@@ -175,26 +175,26 @@ impl PositionExt for lsp_types::Position {
 pub(crate) trait TextSizeExt {
     /// Converts this position to an `LspPosition`, which then requires an explicit
     /// decision about how to use it (as a local position or as a location).
-    fn as_lsp_position(
-        self,
-        db: &dyn Db,
+    fn as_lsp_position<'db>(
+        &self,
+        db: &'db dyn Db,
         file: File,
         encoding: PositionEncoding,
-    ) -> LspPosition<'_>
+    ) -> LspPosition<'db>
     where
         Self: Sized;
 }
 
 impl TextSizeExt for TextSize {
-    fn as_lsp_position(
-        self,
-        db: &dyn Db,
+    fn as_lsp_position<'db>(
+        &self,
+        db: &'db dyn Db,
         file: File,
         encoding: PositionEncoding,
-    ) -> LspPosition<'_> {
+    ) -> LspPosition<'db> {
         LspPosition {
             file,
-            position: self,
+            position: *self,
             db,
             encoding,
         }


crates/ty_server/src/server/api/requests/completion.rs
```diff
@@ -49,11 +48,10 @@ impl BackgroundDocumentRequestHandler for CompletionRequestHandler {
...
         };
 
-        let source = source_text(db, file);
-        let line_index = line_index(db, file);
         let offset = params.text_document_position.position.to_text_size(
-            &source,
-            &line_index,
+            db,
+            file,
+            snapshot.url(),
```
MichaReiser, line 54: It's a bit unfortunate that we need the `url` here but the file isn't unique enough (the file maps to the entire notebook). We need the `url` to lookup the corresponding cell within the notebook.

crates/ty_server/src/server/api/requests/completion.rs
```diff
@@ -73,9 +71,10 @@ impl BackgroundDocumentRequestHandler for CompletionRequestHandler {
                 let kind = comp.kind(db).map(ty_kind_to_lsp_kind);
                 let type_display = comp.ty.map(|ty| ty.display(db).to_string());
                 let import_edit = comp.import.as_ref().map(|edit| {
-                    let range =
-                        edit.range()
-                            .to_lsp_range(&source, &line_index, snapshot.encoding());
+                    let range = edit
+                        .range()
+                        .as_lsp_range(db, file, snapshot.encoding())
+                        .to_local_range();
```
MichaReiser, line 77: Using `to_local_range` here is technically incorrect. In fact, it's very likely that the importer adds the import to the very first cell within the notebook, in which case the edit is not local to a single cell. 

However, fixing this requires changes to the importer and they're substantial enough to justify a separate PR

crates/ty_server/src/server/api/requests/doc_highlights.rs
```diff
@@ -58,7 +56,8 @@ impl BackgroundDocumentRequestHandler for DocumentHighlightRequestHandler {
             .map(|target| {
                 let range = target
                     .range()
-                    .to_lsp_range(&source, &line_index, snapshot.encoding());
+                    .as_lsp_range(db, file, snapshot.encoding())
+                    .to_local_range();
```
MichaReiser, line 60: We'll need to limit the `document_highlights` to the current cell. This seems very unfortunate as the highlights are very likely to be cross cell but the LSP protocol doesn't allow us to specify a URI. Again, I think this is best done in a separate PR

crates/ty_server/src/server/api/requests/document_symbols.rs
```diff
@@ -112,15 +108,19 @@ fn convert_to_lsp_document_symbol(
         tags: None,
         #[allow(deprecated)]
         deprecated: None,
-        range: symbol.full_range.to_lsp_range(source, line_index, encoding),
-        selection_range: symbol.name_range.to_lsp_range(source, line_index, encoding),
+        range: symbol
+            .full_range
+            .as_lsp_range(db, file, encoding)
+            .to_local_range(),
```
MichaReiser, line 108: Same here: We need to limit the document symbol request to the current cell. But I'd prefer doing that in a separate PR

crates/ty_server/src/server/api/semantic_tokens.rs
```diff
@@ -25,7 +24,10 @@ pub(crate) fn generate_semantic_tokens(
     let mut prev_start = 0u32;
 
     for token in &*semantic_token_data {
-        let lsp_range = token.range().to_lsp_range(&source, &line_index, encoding);
+        let lsp_range = token
```
MichaReiser, line 27: We'll need to limit the semantic tokens to a single range but I'll do this in a separate PR

sharkdp:
Seems like something that maybe @dhruvmanila would be a good candidate to review?

crates/ty_server/src/server/api/requests/doc_highlights.rs
dhruvmanila: This isn't something that requires to be investigated right now but I tested Pylance and it seems to be able to highlight a symbol across multiple cells in a notebook.

https://github.com/user-attachments/assets/ef8dd35d-9898-40a5-8c22-edf2374c030b

dhruvmanila (approved):
Looks good!

crates/ty_server/src/server/api/requests/doc_highlights.rs
MichaReiser: I'm not sure this is something Pylance does (although it could because it isn't an LSP).

But I don't see a highlight request for cells at all and the highlighting is also incorrect, suggesting it's just a simple name matching:




https://github.com/user-attachments/assets/1b323d9f-bfa3-4f4b-bc3d-6e2095d649b8

MichaReiser: It could mean that we should return `Option` from `to_location` but I've to revisit this when being further along with the notebook integration

dhruvmanila: That makes sense.

PR #21238, opened 2025-11-03
[ty] `dict` is not assignable to `TypedDict`

ibraheemdev:
## Summary

A lot of the bidirectional inference work relies on `dict` not being assignable to `TypedDict`, so I think it makes sense to add this before fully implementing https://github.com/astral-sh/ty/issues/1387.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/typed_dict.md
+++ b/crates/ty_python_semantic/resources/mdtest/typed_dict.md
@@ -83,7 +83,6 @@ CAPITALIZED_NAME = "Name"
 
 # error: [invalid-key] "Invalid key for TypedDict `Person`: Unknown key "Name" - did you mean "name"?"
 # error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `Person` constructor"
-# error: [invalid-assignment]
 dave: Person = {CAPITALIZED_NAME: "Dave", "age": 20}
 
 def age() -> Literal["age"] | None:
@@ -96,7 +95,6 @@ The construction of a `TypedDict` is checked for type correctness:
 
 ```py
 # error: [invalid-argument-type] "Invalid argument to key "name" with declared type `str` on TypedDict `Person`"
-# error: [invalid-assignment]
 eve1a: Person = {"name": b"Eve", "age": None}
 
 # error: [invalid-argument-type] "Invalid argument to key "name" with declared type `str` on TypedDict `Person`"
@@ -106,7 +104,6 @@ reveal_type(eve1a)  # revealed: Person
 reveal_type(eve1b)  # revealed: Person
 
 # error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `Person` constructor"
-# error: [invalid-assignment]
 eve2a: Person = {"age": 22}
 
 # error: [missing-typed-dict-key] "Missing required key 'name' in TypedDict `Person` constructor"
@@ -116,7 +113,6 @@ reveal_type(eve2a)  # revealed: Person
 reveal_type(eve2b)  # revealed: Person
 
 # error: [invalid-key] "Invalid key for TypedDict `Person`: Unknown key "extra""
-# error: [invalid-assignment]
 eve3a: Person = {"name": "Eve", "age": 25, "extra": True}
 
 # error: [invalid-key] "Invalid key for TypedDict `Person`: Unknown key "extra""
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/resources/mdtest/comprehensions/basic.md
```diff
@@ -162,10 +162,13 @@ The type context is propagated down into the comprehension:
 class Person(TypedDict):
     name: str
 
+# TODO: This should not error.
+# error: [invalid-assignment]
 persons: list[Person] = [{"name": n} for n in ["Alice", "Bob"]]
```
ibraheemdev, line 167: @sharkdp it looks like the type context is not being propagated correctly here, and only seemed so because of `dict[str, str]` being assignable to `Person`.

ibraheemdev:
I'm not sure if we want to be emitting `invalid-assignment` errors if we have already emitted a `invalid-key` error. Maybe we should special case `TypedDict` annotated assignments to avoid this? Note that we *do* want to infer a `dict` for the RHS in that case (see https://github.com/astral-sh/ruff/pull/21169), but the diagnostics are effectively being duplicated.

crates/ty_python_semantic/resources/mdtest/comprehensions/basic.md
sharkdp: Yes, absolutely. I was aware about this, but should have added a comment. But this is why I added the other test below (where it's supposed to be an error).

sharkdp: See also this comment: https://github.com/astral-sh/ruff/pull/20962#discussion_r2482437556

sharkdp:
> I'm not sure if we want to be emitting `invalid-assignment` errors if we have already emitted a `invalid-key` error. Maybe we should special case `TypedDict` annotated assignments to avoid this? Note that we _do_ want to infer a `dict` for the RHS in that case (see #21169), but the diagnostics are effectively being duplicated.

I was about to add a comment to this PR that we should maybe try to avoid emitting the additional `invalid-assignment` diagnostic, and only then saw your comment. I think it would be great if we could attempt to do this, if it's not too complicated.

crates/ty_python_semantic/src/types.rs
```diff
@@ -1975,11 +1975,14 @@ impl<'db> Type<'db> {
                 ConstraintSet::from(false)
             }
 
-            (Type::TypedDict(_), _) | (_, Type::TypedDict(_)) => {
+            (Type::TypedDict(_), _) => {
                 // TODO: Implement assignability and subtyping for TypedDict
                 ConstraintSet::from(relation.is_assignability())
             }
 
+            // A non-`TypedDict` cannot subtype a `TypedDict`
```
sharkdp, line 1983: I guess intersections (of a `TypedDict` and something else), typevars, and `Never` would be types that can subtype `TypedDict`, but they were all handled above...

crates/ty_python_semantic/resources/mdtest/typed_dict.md
```diff
@@ -201,6 +205,8 @@ reveal_type(alice["inner"]["age"])  # revealed: int | None
 reveal_type(alice["inner"]["non_existing"])  # revealed: Unknown
 
 # error: [invalid-key] "Invalid key for TypedDict `Inner`: Unknown key "extra""
+# error: [invalid-argument-type]
```
sharkdp: I'm confused. Where does `invalid-argument-type` come from?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -6211,7 +6205,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         } = set;
 
         let elts = elts.iter().map(|elt| [Some(elt)]);
-        self.infer_collection_literal(elts, tcx, KnownClass::Set)
+        let infer_elt_ty = |builder: &mut Self, elt, tcx| builder.infer_expression(elt, tcx);
```
sharkdp, line 6208: I know you're not the one to invent/introduce this, but I think I would prefer expanding `elt`/`elts` to `element`/`elements`

crates/ty_python_semantic/resources/mdtest/typed_dict.md
ibraheemdev: We emit `invalid-argument-type` when a value is not assignable to its declared key type, but this should be removed now.

ibraheemdev:
I added a special case to avoid the duplicated diagnostics for simple annotated assignments, but attribute assignment and function arguments are trickier, because we may hide typed-dict diagnostics during multi-inference and rely on the assignability check to fail.

There is one correct new typing comformance diagnostic, but also two regressions because we do not correctly handle `extra_items` (and were hiding diagnostics by treating a `dict` as a `TypedDict`).

ibraheemdev:
Most of the diagnostics seem to be duplicated diagnostics (in the not-so-simple special case). I opened https://github.com/astral-sh/ty/issues/1472 to track that issue, but I think landing this is high priority because of how much bidirectional-inference relies on the fallback untyped `dict`.

PR #21241, opened 2025-11-03
[ty] Add `ty_server::Db` trait

MichaReiser:
## Summary

For notebooks, we'll need a way to query the metadata of an arbitrary notebook when mapping 
e.g. the range of a find reference operation to the reference's location. 

This PR introduces a new `Db` trait and implements it for `ProjectDatabase`. 
The new `Db` trait allows quering the document metadata by path, assuming
that the `ProjectDatabase` uses the `LSPSystem` (which is always true).

### DIFF
--- a/crates/ty_server/src/db.rs
+++ b/crates/ty_server/src/db.rs
@@ -7,11 +7,14 @@ use ty_project::{Db as ProjectDb, ProjectDatabase};
 
 #[salsa::db]
 pub(crate) trait Db: ProjectDb {
+    /// Returns the LSP [`Document`] corresponding to `File` or
+    /// `None` if the file isn't open in the editor.
     fn document(&self, file: File) -> Option<&Document>;
 
+    /// Returns the LSP [`NotebookDocument`] corresponding to `File` or
+    /// `None` if the file isn't open in the editor or if it isn't a notebook.
     fn notebook_document(&self, file: File) -> Option<&NotebookDocument> {
-        let document = self.document(file)?;
-        document.as_notebook()
+        self.document(file)?.as_notebook()
     }
 }
 


crates/ty/src/lib.rs
```diff
@@ -450,12 +450,12 @@ impl ty_project::ProgressReporter for IndicatifReporter {
         self.bar.set_draw_target(self.printer.progress_target());
     }
 
-    fn report_checked_file(&self, db: &dyn Db, file: File, diagnostics: &[Diagnostic]) {
+    fn report_checked_file(&self, db: &ProjectDatabase, file: File, diagnostics: &[Diagnostic]) {
```
MichaReiser, line 453: I had to change the signature of `ProgressReporter` to take a `ProjectDatabase` as argument because we now need a `ty_server::Db` instead of a `ty_project::Db` in the LSP.

crates/ty_server/src/db.rs
```diff
@@ -0,0 +1,30 @@
+use crate::NotebookDocument;
+use crate::session::index::Document;
+use crate::system::LSPSystem;
+use ruff_db::Db as _;
+use ruff_db::files::{File, FilePath};
+use ty_project::{Db as ProjectDb, ProjectDatabase};
+
+#[salsa::db]
+pub(crate) trait Db: ProjectDb {
```
sharkdp, line 9: I find this name shadowing here confusing. Is there another name that we could use for the trait here? I assume you named it `Db` to avoid all kinds of downstream changes?

crates/ty_server/src/db.rs
```diff
@@ -0,0 +1,30 @@
...
+    fn document(&self, file: File) -> Option<&Document>;
```
sharkdp, line 12: Maybe
```suggestion
    fn document_from_file(&self, file: File) -> Option<&Document>;
```

and similar for `notebook_document` below?

crates/ty_server/src/db.rs
MichaReiser: The idea is that you can always import `Db` without having to think within which crate you're in (this is also how all our other `Db` traits are defined)

sharkdp: Makes sense, thank you.

crates/ty_server/src/db.rs
MichaReiser: I find `from` more confusing because `From` implies that there's a conversion at play. But that's not the case. The operation here is closer to a `map.get(file)` where `map` is a `HashMap<File, Document>`.

PR #21255, opened 2025-11-03
[ty] Use "cannot" consistently over "can not"

AlexWaygood:
## Summary

This is mostly an internal change, but does impact a few ty diagnostics

## Test Plan

`cargo test`


MichaReiser (commented):
As a non native speaker: Why prefer "cannot" over "can not". We should probably also start writing down some guidance on how to write diagnostic messages

AlexWaygood:
> As a non native speaker: Why prefer "cannot" over "can not". We should probably also start writing down some guidance on how to write diagnostic messages

There's not any difference in meaning, but "cannot" is far more commonly used:
- https://www.merriam-webster.com/grammar/cannot-vs-can-not-is-there-a-difference
- https://www.grammarly.com/blog/commonly-confused-words/cannot-or-can-not/
- https://www.microsoft.com/en-us/microsoft-365-life-hacks/writing/when-to-use-cannot-vs-can-not

As a result, "can not" ends up looking odd to my eyes: I trip up slightly every time I read it. You should basically always use "cannot" rather than "can not", I think

Issue #21257, opened 2025-11-03
FURB164 fixes should be safe in a few cases


dscorbett:
### Summary

Some fixes of [`unnecessary-from-float` (FURB164)](https://docs.astral.sh/ruff/rules/unnecessary-from-float/) are marked unsafe but are actually safe: when the argument is specified as a keyword argument but would otherwise be marked safe, and when the argument to `Fraction.from_decimal` is a `Decimal`. [Example](https://play.ruff.rs/9f908efb-7488-4848-b1d0-9b8a1553c01b):

```console
$ cat >furb164.py <<'# EOF'
from decimal import Decimal
from fractions import Fraction
print(Fraction.from_float(f=4.2))
print(Fraction.from_decimal(dec=4))
print(Fraction.from_decimal(Decimal("4.2")))
print(Fraction.from_decimal(dec=Decimal("4.2")))
# EOF

$ python furb164.py
4728779608739021/1125899906842624
4
21/5
21/5

$ ruff --isolated check furb164.py --select FURB164 --preview --output-format concise -q
furb164.py:3:7: FURB164 Verbose method `from_float` in `Fraction` construction
furb164.py:4:7: FURB164 Verbose method `from_decimal` in `Fraction` construction
furb164.py:5:7: FURB164 Verbose method `from_decimal` in `Fraction` construction
furb164.py:6:7: FURB164 Verbose method `from_decimal` in `Fraction` construction

$ ruff --isolated check furb164.py --select FURB164 --preview --unsafe-fixes --fix
Found 4 errors (4 fixed, 0 remaining).

$ cat furb164.py
from decimal import Decimal
from fractions import Fraction
print(Fraction(4.2))
print(Fraction(4))
print(Fraction(Decimal("4.2")))
print(Fraction(Decimal("4.2")))

$ python furb164.py
4728779608739021/1125899906842624
4
21/5
21/5
```

### Version

ruff 0.14.3 (8737a2d5f 2025-10-30)

PR #21259, opened 2025-11-03
[`refurb`] Expand fix safety for keyword arguments and `Decimal`s (`FURB164`)

chirizxc:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/21257

## Test Plan

`cargo nextest run furb164`


ntBre (approved):
Thank you! It would be kind of nice if we could use our limited [type inference](https://github.com/astral-sh/ruff/blob/f64bbb45b9d7249138d65cc9b9ac6b87f74eef1b/crates/ruff_python_semantic/src/analyze/typing.rs#L1) infrastructure to check `Decimal` types for annotated variables and not just the direct constructor calls, but this seems like a good step in the right direction.

PR #21260, opened 2025-11-03
[ty] prefer submodule over module __getattr__ in from-imports

carljm:
Fixes https://github.com/astral-sh/ty/issues/1053

## Summary

Other type checkers prioritize a submodule over a package `__getattr__` in `from mod import sub`, even though the runtime precedence is the other direction. In effect, this is making an implicit assumption that a module `__getattr__` will not handle (that is, will raise `AttributeError`) for names that are also actual submodules, rather than shadowing them. In practice this seems like a realistic assumption in the ecosystem? Or at least the ecosystem has adapted to it, and we need to adapt this precedence also, for ecosystem compatibility.

The implementation is a bit ugly, precisely because it departs from the runtime semantics, and our implementation is oriented toward modeling runtime semantics accurately. That is, `__getattr__` is modeled within the member-lookup code, so it's hard to split "member lookup result from module `__getattr__`" apart from other member lookup results. I did this via a synthetic `TypeQualifier::FROM_MODULE_GETATTR` that we attach to a type resulting from a member lookup, which isn't beautiful but it works well and doesn't introduce inefficiency (e.g. redundant member lookups).

## Test Plan

Updated mdtests.

Also added a related mdtest formalizing our support for a module `__getattr__` that is explicitly annotated to accept a limited set of names. In principle this could be an alternative (more explicit) way to handle the precedence problem without departing from runtime semantics, if the ecosystem would adopt it.

### Ecosystem analysis

Lots of removed diagnostics which are an improvement because we now infer the expected submodule.

Added diagnostics are mostly unrelated issues surfaced now because we previously had an earlier attribute error resulting in `Unknown`; now we correctly resolve the module so that earlier attribute error goes away, we get an actual type instead of `Unknown`, and that triggers a new error.

In scipy and sklearn, the module `__getattr__` which we were respecting previously is un-annotated so returned a forgiving `Unknown`; now we correctly see the actual module, which reveals some cases of https://github.com/astral-sh/ty/issues/133 that were previously hidden (`scipy/optimize/__init__.py` [imports `from ._tnc`](https://github.com/scipy/scipy/blob/eff82ca575668d2d7a4bc12b6afba98daaf6d5d0/scipy/optimize/__init__.py#L429).)

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/module_getattr.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/module_getattr.md
@@ -95,7 +95,7 @@ import mod.sub
 reveal_type(mod.sub)  # revealed: <module 'mod.sub'>
 ```
 
-If you `from mod import sub`, at runtime `sub` will be the return value from the module
+If you `from mod import sub`, at runtime `sub` will be the value returned by the module
 `__getattr__`, but other type checkers do not model the precedence this way. They will always prefer
 a submodule over a package `__getattr__`, and thus this is the current expectation in the ecosystem.
 Effectively, this assumes that a well-implemented package `__getattr__` will always raise


crates/ty_python_semantic/resources/mdtest/import/module_getattr.md
```diff
@@ -86,10 +84,53 @@ import mod
...
+
 `test_import_mod_sub.py`:
 
 ```py
 import mod.sub
 
 reveal_type(mod.sub)  # revealed: <module 'mod.sub'>
 ```
+
+If you `from mod import sub`, at runtime `sub` will be the return value from the module
```
AlexWaygood:
```suggestion
If you `from mod import sub`, at runtime `sub` will be the value returned by the module
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5418,6 +5426,24 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                alias.into(),
+                definition,
+                &DeclaredAndInferredType::MightBeDifferent {
+                    declared_ty: TypeAndQualifiers {
+                        inner: ty,
+                        origin: TypeOrigin::Declared,
+                        qualifiers,
+                    },
+                    inferred_ty: ty,
+                },
```
AlexWaygood, line 5442: why are we using `DeclaredAndInferredType::MightBeDifferent` here? It looks like the type always comes from the return annotation of `__getattr__` (i.e., a declaration) and that they're always the same? Intuitively, `DeclaredAndInferredType::AreTheSame` feels like it would be a better fit here

carljm: `AreTheSame` implies "there are no type qualifiers on the declared type", because you provide only a type and internally it builds a no-qualifiers declared type from it. If there are potential qualifiers to pass through, we can't use it.

AlexWaygood: Ah I missed that, thanks

PR #21261, opened 2025-11-03
Run codspeed benchmarks with `profiling` profile

ibraheemdev:
## Summary

The codspeed action currently takes >15m to run. Let's see how much this improves it..

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -949,7 +949,7 @@ jobs:
           tool: cargo-codspeed
 
       - name: "Build benchmarks"
-        run: cargo codspeed build --features "codspeed,instrumented" --no-default-features --profile profiling -p ruff_benchmark --bench formatter --bench lexer --bench linter --bench parser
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -949,7 +949,7 @@ jobs:
           tool: cargo-codspeed
 
       - name: "Build benchmarks"
-        run: cargo codspeed build --features "codspeed,instrumented" --no-default-features -p ruff_benchmark --bench formatter --bench lexer --bench linter --bench parser
+        run: cargo codspeed build --features "codspeed,instrumented" --no-default-features --profile profiling -p ruff_benchmark --bench formatter --bench lexer --bench linter --bench parser
```
MichaReiser: You probably want to change the profile for other jobs too? Because this is only for the formatter.

ibraheemdev:
This reduces the walltime benchmarks from 15m to 10m, and we should see an even bigger improvement once build caching kicks in, so I think it's worth the downsides.

PR #21266, opened 2025-11-03
Run py-fuzzer with `--profile=profiling` locally and in CI

AlexWaygood:
## Summary

Following #20962 we're seeing the fuzzer take much longer in CI on seed 742 than on any other seed (https://github.com/astral-sh/ruff/pull/20962#issuecomment-3477968074). But this problem is only reproducible in debug builds, so this doesn't seem like an issue that's worth worrying much about.

This PR switches the fuzzer CI job to use `--profile=profiling`, and also makes this the new default if you run the fuzzer locally. This should mean that the build the fuzzer uses in CI has more similar performance characteristics to a release build. This in turn means we don't have to constantly wonder whether the fuzzer taking twice as long is indicative of a "real" issue or just an artifact of debug-build weirdness.

As well this, this also reduces the amount of time the fuzzer CI job takes (4min30s -> 3min40s) and reduces the amount of time we need to wait for the fuzzer job to start. (Prior to this PR, you had to wait for the `cargo-test-linux` job to finish before it would even start, because it reused the build uploaded from that job.)

## Test Plan

CI on this PR

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -437,6 +437,8 @@ jobs:
           workspaces: "fuzz -> target"
       - name: "Install Rust toolchain"
         run: rustup show
+      - name: "Install mold"
+        uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
       - name: "Install cargo-binstall"
         uses: cargo-bins/cargo-binstall@b3f755e95653da9a2d25b99154edfdbd5b356d0a # v1.15.10
       - name: "Install cargo-fuzz"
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):

.github/workflows/ci.yaml
```diff
@@ -645,45 +647,45 @@ jobs:
...
+          git rev-list --format=%s --max-count=1 "$GITHUB_SHA"
+          cargo build --profile=profiling --bin=ty
+          mv target/profiling/ty ty-new
+
+          MERGE_BASE="$(git merge-base "$GITHUB_SHA" "origin/$GITHUB_BASE_REF")"
+          git checkout -b old_commit "$MERGE_BASE"
+          echo "old commit (merge base)"
+          git rev-list --format=%s --max-count=1 old_commit
+          cargo build --profile=profiling --bin=ty
+          mv target/profiling/ty ty-old
```
AlexWaygood, line 679: this is all basically copied-and-pasted from https://github.com/astral-sh/ruff/blob/79a02711c12d5381a640c7b07639e058917a3234/.github/workflows/typing_conformance.yaml#L63-L78

and is also similar to what mypy_primer does.

PR #21275, opened 2025-11-04
[ty] Simplify unions containing multiple type variables during inference

ibraheemdev:
## Summary

Splitting this one out from https://github.com/astral-sh/ruff/pull/21210. This is also something that should be made obselete by the new constraint solver, but is easy enough to fix now.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
@@ -474,7 +474,7 @@ def g(x: str):
     f(prefix=x, suffix=".tar.gz")
 ```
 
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
```diff
@@ -474,6 +474,16 @@ def g(x: str):
     f(prefix=x, suffix=".tar.gz")
 ```
 
+If the type variables is present multiple times in the union, we choose the correct union element to
```
carljm:
```suggestion
If the type variable is present multiple times in the union, we choose the correct union element to
```

PR #21276, opened 2025-11-04
[ty] Make special cases for `UnionType` slightly narrower

AlexWaygood:
Fixes https://github.com/astral-sh/ty/issues/1478

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -8474,11 +8474,6 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 | Type::GenericAlias(..)
                 | Type::SpecialForm(_)
                 | Type::KnownInstance(KnownInstanceType::UnionType(_)),
-                _,
-                ast::Operator::BitOr,
-            )
-            | (
-                _,
                 Type::ClassLiteral(..)
                 | Type::SubclassOf(..)
                 | Type::GenericAlias(..)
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Nearly all the new diagnostics are false positive due to things like `X = type[str] | int`. We already have false positives like this on `main` (https://github.com/astral-sh/ruff/pull/21195#issuecomment-3482524821) and I think @sharkdp is planning on prioritising adding support for this kind of pattern in implicit type aliases

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -8474,42 +8474,73 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                | Type::GenericAlias(..)
+                | Type::KnownInstance(..)
+                | Type::SpecialForm(..),
+                ast::Operator::BitOr,
+            ) if Program::get(self.db()).python_version(self.db()) >= PythonVersion::PY310
+                && instance.has_known_class(self.db(), KnownClass::NoneType) =>
+            {
+                Some(Type::KnownInstance(KnownInstanceType::UnionType(
+                    UnionTypeInstance::new(self.db(), left_ty, right_ty),
+                )))
```
sharkdp, line 8514: I'm slightly confused why we don't need the `convert_none_type` anymore?

sharkdp (approved):
Thank you!

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: Because of https://github.com/astral-sh/ruff/pull/21263!

sharkdp: :see_no_evil:

PR #21278, opened 2025-11-05
[`refurb`] Auto-fix annotated assignments (`FURB101`)

### LINKED ISSUE #21274: Annotations block the FURB101 fix
### Summary

Annotations block the fix for [`read-whole-file` (FURB101)](https://docs.astral.sh/ruff/rules/read-whole-file/) but shouldn’t. [Example](https://play.ruff.rs/8ab9fa5b-48fc-474c-ae0e-b71b5112dcb3):
```console
$ cat >furb101.py <<'# EOF'
with open("file.txt", encoding="utf-8") as f:
    contents: str = f.read()
# EOF

$ ruff --isolated check furb101.py --select FURB101 --preview --fix --output-format concise
furb101.py:1:6: FURB101 `open` and `read` should be replaced by `Path("file.txt").read_text(encoding="utf-8")`
Found 1 error.
```

The fix should produce:
```python
import pathlib
contents: str = pathlib.Path("file.txt").read_text(encoding="utf-8")
```

### Version

danparizher:
## Summary

Fixed FURB101 (`read-whole-file`) to handle annotated assignments. Previously, the rule would detect violations in code like `contents: str = f.read()` but fail to generate a fix. Now it correctly generates fixes that preserve type annotations (e.g., `contents: str = Path("file.txt").read_text(encoding="utf-8")`).

Fixes #21274

## Problem Analysis

The FURB101 rule was only checking for `Stmt::Assign` statements when determining whether a fix could be applied. When encountering annotated assignments (`Stmt::AnnAssign`) like `contents: str = f.read()`, the rule would:

1. Correctly detect the violation (the diagnostic was reported)
2. Fail to generate a fix because:
   - The `visit_expr` method only matched `Stmt::Assign`, not `Stmt::AnnAssign`
   - The `generate_fix` function only accepted `Stmt::Assign` in its body validation
   - The replacement code generation didn't account for type annotations

This occurred because Python's AST represents annotated assignments as a different node type (`StmtAnnAssign`) with separate fields for the target, annotation, and value, unlike regular assignments which use a list of targets.

## Approach

The fix extends the rule to handle both assignment types:

1. **Updated `visit_expr` method**: Now matches both `Stmt::Assign` and `Stmt::AnnAssign`, extracting:
   - Variable name from the target expression
   - Type annotation code (when present) using the code generator

2. **Updated `generate_fix` function**:
   - Added `annotation: Option<String>` parameter to accept annotation code
   - Updated body validation to accept both `Stmt::Assign` and `Stmt::AnnAssign`
   - Modified replacement code generation to preserve annotations: `{var}: {annotation} = {binding}({filename_code}).{suggestion}`

3. **Added test case**: Added an annotated assignment test case to verify the fix works correctly.

The implementation maintains backward compatibility with regular assignments while adding support for annotated assignments, ensuring type annotations are preserved in the generated fixes.

### DIFF
--- a/crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
@@ -182,40 +182,11 @@ fn generate_fix(
     with_stmt: &ast::StmtWith,
     suggestion: &str,
 ) -> Option<Fix> {
-    if !(with_stmt.items.len() == 1
-        && matches!(
-            with_stmt.body.as_slice(),
-            [Stmt::Assign(_) | Stmt::AnnAssign(_)]
-        ))
-    {
+    if with_stmt.items.len() != 1 {
         return None;
     }
 
     let locator = checker.locator();
-
-    let (target, annotation) = match with_stmt.body.first() {
-        Some(Stmt::Assign(assign)) if assign.value.range().contains_range(expr.range()) => {
-            match assign.targets.first() {
-                Some(Expr::Name(name)) => (Some(name.id.as_str()), None),
-                _ => (None, None),
-            }
-        }
-        Some(Stmt::AnnAssign(ann_assign))
-            if ann_assign
-                .value
-                .as_ref()
-                .is_some_and(|v| v.range().contains_range(expr.range())) =>
-        {
-            match ann_assign.target.as_ref() {
-                Expr::Name(name) => {
-                    let annotation_code = locator.slice(ann_assign.annotation.range());
-                    (Some(name.id.as_str()), Some(annotation_code.to_string()))
-                }
-                _ => (None, None),
-            }
-        }
-        _ => (None, None),
-    };
 
     let filename_code = locator.slice(open.filename.range());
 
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -129,21 +129,43 @@ impl<'a> Visitor<'a> for ReadMatcher<'a, '_> {
                     return;
                 }
 
-                let target = match self.with_stmt.body.first() {
+                let (target, annotation) = match self.with_stmt.body.first() {
```
ntBre: I should have noticed this in the earlier PR, but this actually looks quite repetitive with the check inside of `generate_fix`. I think we can combine this check and this check:

```rust
matches!(with_stmt.body.as_slice(), [Stmt::Assign(_)]))
```

`generate_fix` seems like the right place to do that to me, which should allow us to decrease the number of arguments since we're already passing the `with_stmt` itself.

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -129,21 +129,43 @@ impl<'a> Visitor<'a> for ReadMatcher<'a, '_> {
...
+                            .value
+                            .as_ref()
+                            .is_some_and(|v| v.range().contains_range(expr.range())) =>
+                    {
+                        match ann_assign.target.as_ref() {
+                            Expr::Name(name) => {
+                                let annotation_code = self
+                                    .checker
+                                    .generator()
+                                    .expr(ann_assign.annotation.as_ref());
```
ntBre: Do we need to use the generator here, or can we just slice the `&str` out of the input?


```suggestion
                                let annotation_code = self
                                    .checker
                                    .locator()
                                    .slice(ann_assign.annotation.range());
```

(untested but that's what I had in mind)

crates/ruff_linter/src/rules/refurb/rules/read_whole_file.rs
```diff
@@ -194,15 +182,45 @@ fn make_suggestion(open: &FileOpen<'_>, generator: Generator) -> String {
...
+    expr: &Expr,
     with_stmt: &ast::StmtWith,
     suggestion: &str,
 ) -> Option<Fix> {
-    if !(with_stmt.items.len() == 1 && matches!(with_stmt.body.as_slice(), [Stmt::Assign(_)])) {
+    if !(with_stmt.items.len() == 1
+        && matches!(
+            with_stmt.body.as_slice(),
+            [Stmt::Assign(_) | Stmt::AnnAssign(_)]
+        ))
```
ntBre: I think we should combine this check with the `match` below. We currently have code for non-assignments (`target = None` and `annotation = None` in the `match` and the third `replacement` arm), but these are unreachable because of this check.

ntBre:
I merged the conflicting test changes I made in another PR, simplified the implementation a bit further, and then fixed a couple of preexisting bugs I noticed. The ecosystem comment doesn't seem to be updating, but I downloaded the artifact and checked that instead. It's showing -188 fixes, and from the ones I've clicked on this seems correct. The two bugs I fixed were that the `assign.targets.first()` check wasn't exactly right because it could drop other elements in an assignment like:

```py
content, x = f.read(), 2
```

and the `contains_range` checks also weren't correct. `contains_range` is true for an expression like:

```py
content = process_contents(f.read())
```

but we were then replacing the entire assignment statement and discarding the `process_contents` call. All of the ecosystem changes I've seen so far fit into this latter pattern.

The fix is now fairly restricted. We may want to consider expanding it in the future by doing a more targeted replacement of the `read` call expression itself instead of re-building the whole assignment as a string. But I think the current state of the PR is a substantial enough improvement to merge as-is.

ntBre (approved):

PR #21281, opened 2025-11-05
[ty] Update salsa

MichaReiser:
## Summary

Update Salsa to pull in https://github.com/salsa-rs/salsa/pull/1021, which is needed for https://github.com/astral-sh/ruff/pull/20566

## Test Plan

`cargo test`

### DIFF
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -114,17 +114,15 @@ pub(crate) fn infer_definition_types<'db>(
 fn definition_cycle_recover<'db>(
     db: &'db dyn Db,
     _id: salsa::Id,
-    _last_provisional_value: &DefinitionInference<'db>,
-    _value: &DefinitionInference<'db>,
+    last_provisional_value: &DefinitionInference<'db>,
+    value: DefinitionInference<'db>,
     count: u32,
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/types/infer.rs
```diff
@@ -114,17 +114,15 @@ pub(crate) fn infer_definition_types<'db>(
...
+    value: DefinitionInference<'db>,
     count: u32,
     definition: Definition<'db>,
-) -> salsa::CycleRecoveryAction<DefinitionInference<'db>> {
-    if count == ITERATIONS_BEFORE_FALLBACK {
-        salsa::CycleRecoveryAction::Fallback(DefinitionInference::cycle_fallback(
-            definition.scope(db),
-        ))
+) -> DefinitionInference<'db> {
+    if &value == last_provisional_value || count != ITERATIONS_BEFORE_FALLBACK {
```
MichaReiser, line 122: It would be correcter for us to use `count < ITERATIONS_BEFORE_FALLBACK` but one existing mdtest fails if we do so (but it fixes a corpus test panic).

I decided not to spend any time on debugging this further, given that https://github.com/astral-sh/ruff/pull/20566 is about to solve this holistically

carljm (approved):

PR #21285, opened 2025-11-05
[ty] Favour imported symbols over builtin symbols

MatthewMckee4:
## Summary

Raised by @AlexWaygood.

We previously did not favour imported symbols, when we probably should've 

## Test Plan

Add test showing that we favour imported symbol even if it is alphabetically after other symbols that are builtin.


BurntSushi (approved):
Nice!

PR #21287, opened 2025-11-05
[ty] Clarify behavior of constraint sets for gradual upper bounds and constraints

dcreager:
When checking whether a constraint set is satisfied, if a typevar has a non-fully-static upper bound or constraint, we are free to choose any materialization that makes the check succeed.

In non-inferable positions, we have to show that the constraint set is satisfied for all valid specializations, so it's best to choose the most restrictive materialization, since that minimizes the set of valid specializations that have to pass.

In inferable positions, we only have to show that the constraint set is satisfied for _some_ valid specializations, so it's best to choose the most permissive materialization, since that maximizes our chances of finding a specialization that passes.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
@@ -328,8 +328,8 @@ def constrained_by_gradual[T: (Base, Any)]():
     static_assert(not ConstraintSet.never().satisfied_by_all_typevars(inferable=tuple[T]))
     static_assert(not ConstraintSet.never().satisfied_by_all_typevars())
 
-    # If we choose Unrelated as the materialization, then (T = Unrelated) is a valid specialization,
-    # which satisfies (T ≤ Unrelated).
+    # If we choose Unrelated as the materialization of the gradual constraint, then (T = Unrelated)
+    # is a valid specialization, which satisfies (T ≤ Unrelated).
     static_assert(ConstraintSet.range(Never, T, Unrelated).satisfied_by_all_typevars(inferable=tuple[T]))
     # No matter which materialization we choose, (T = Base) is a valid specialization, which does
     # not satisfy (T ≤ Unrelated).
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -141,6 +141,116 @@ def bounded[T: Base]():
     static_assert(not constraints.satisfied_by_all_typevars())
 ```
 
+If the upper bound is a gradual type, we are free to choose any materialization of the upper bound
+that makes the test succeed. In non-inferable positions, it is most helpful to choose an upper bound
+that is as restrictive as possible, since that minimizes the number of valid specializations that
+must satisfy the constraint set. (That means we will almost always choose `Never` — or more
+precisely, the bottom specialization — as the upper bound.) In inferable positions, the opposite is
```
AlexWaygood: I think that only holds true if the upper bound of a typevar is a gradual specialization of an _invariant_ type? The bottom materialization of `list[Any]` is `Never`, but the bottom materialization of `Sequence[Any]` is `Sequence[Never]`, since `Sequence[Never]` is a subtype of `Sequence[int]` and `Sequence[str]`, and is still an inhabited type (`[]`, `()` can both inhabit `Sequence[Never]`, for example). Similarly for contravariance, the bottom materialization of `Callable[[Any], int]` is `Callable[[object], int]`.

dcreager: Which part doesn't hold true? This might be poor wording on my part — I am not trying to claim that the bottom materialization is always `Never`. I am (attempting to) say that for `T: Any`, we pick `T: Never`, but it's more precise to say that we pick the bottom specialization, because for `T: list[Any]`, we pick `T: Bottom[list[Any]]`.

sharkdp: > might be poor wording on my part

I think I understood it the way you explained in your comment now, but I also stumbled over it.

AlexWaygood: > This might be poor wording on my part — I am not trying to claim that the bottom materialization is always `Never`.

Ah, okay -- yes, I think the wording can be improved there a little!

dcreager: Reworded this, lmkwyt

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -141,6 +141,116 @@ def bounded[T: Base]():
...
+
+def bounded_by_gradual[T: Any]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars())
+
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars())
+
+    # If we choose Super as the materialization, then (T = Super) is a valid specialization, which
+    # satisfies (T ≤ Super).
```
sharkdp: Maybe just for the first of these: "If we choose Super as the materialization *for the upper bound on T* …"?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -141,6 +141,116 @@ def bounded[T: Base]():
...
+    static_assert(ConstraintSet.range(Never, T, Super).satisfied_by_all_typevars(inferable=tuple[T]))
+    # If we choose Never as the materialization, then (T = Never) is the only valid specialization,
+    # which satisfies (T ≤ Super).
```
sharkdp: I was slightly confused by the description here and above, because it sounded like there would be any proof here for the statement that `T = Never` is the only valid specialization. I think I would have expected something like:

A non-inferable `T: Any` satisfies the constraint set `Never ≤ T ≤ Super`, because we are free to assume any materialization for the upper bound and can therefore chose `T: Never`.

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -141,6 +141,116 @@ def bounded[T: Base]():
...
+    static_assert(constraints.satisfied_by_all_typevars())
+
+    # If we choose Unrelated as the materialization, then (T = Unrelated) is a valid specialization,
+    # which satisfies (T ≤ Unrelated ∧ T ≠ Never).
+    constraints = constraints & ~ConstraintSet.range(Never, T, Never)
+    static_assert(constraints.satisfied_by_all_typevars(inferable=tuple[T]))
+    # There is no upper bound that we can choose to satisfy this constraint set in non-inferable
+    # position. (T = Never) will be a valid assignment no matter what, and that does not satisfy
+    # (T ≤ Unrelated ∧ T ≠ Never).
+    static_assert(not constraints.satisfied_by_all_typevars())
```
sharkdp, line 188: Would it be worth demonstrating that this is only true for `UnrelatedFinal`, but not for a general `Unrelated` class? I would find that more useful compared to checking `Super`, `Sub` and `Base` above, which do not change anything?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
```diff
@@ -141,6 +141,116 @@ def bounded[T: Base]():
...
+```py
+def bounded_by_gradual[T: list[Any]]():
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(ConstraintSet.always().satisfied_by_all_typevars())
+
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars(inferable=tuple[T]))
+    static_assert(not ConstraintSet.never().satisfied_by_all_typevars())
+
+    # If we choose Super as the materialization, then (T = list[Super]) is a valid specialization,
+    # which satisfies (T ≤ list[Super]).
```
sharkdp: Minor: I think it should either say "if we choose `Super` as the materialization of the `Any` in `list[Any]`", or just "if we choose `list[Super]` as the materialization of the upper bound"?

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: Done

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: I was trying to be careful to use "materialization" to only refer to what the `Any` turns into, even when it's part of a larger type. But `list[Any]` is a type that can be materialized, so maybe that is more confusing than helpful. Updated to line up with the changes suggested above

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: I think this will hold for `UnrelatedButNotFinal` as well — and for any type at all, actually. It's the `∧ T ≠ Never` part in the constraint set that causes this to fail — `T = Never` is a valid specialization no matter what materialization we choose, and the extra clause in the constraint set prohibits that specialization no matter what else appears in the constraint set.

(That said, I did follow your suggestion to remove the redundant checks above for `Super` and `Base`)

crates/ty_python_semantic/resources/mdtest/type_properties/satisfied_by_all_typevars.md
dcreager: Reworded

PR #21290, opened 2025-11-06
[`pylint`] Fix suppression for empty dict without tuple key annotation (`PLE1141`)

### LINKED ISSUE #21289: PLE1141 fix is suppressed for empty `dict`
### Summary

The fix for [`dict-iter-missing-items` (PLE1141)](https://docs.astral.sh/ruff/rules/dict-iter-missing-items/) is suppressed when the relevant `dict`’s keys are all 2-tuples in its initial binding. This usually makes sense, but it doesn’t make sense when the `dict` is defined as empty. In that case, there is no reason to assume that any keys added later will be 2-tuples, and that is a rare pattern so they probably won’t be, so PLE1141 should allow the fix. If the empty `dict` is annotated to have 2-tuple keys, as in `dict[tuple[int, str], bool]`, though, it is still appropriate to suppress the fix. [Example](https://play.ruff.rs/e6181e08-ba24-4009-a49d-33a13ad03e0e):
```console
$ cat >ple1141.py <<'# EOF'
data = {}
data["x"] = 1
for k, v in data:
    print(f"{k=} {v=}")
# EOF

$ ruff --isolated check ple1141.py --select PLE1141 --preview
All checks passed!
```

### Version

danparizher:
## Summary

Fixes the PLE1141 (`dict-iter-missing-items`) rule to allow fixes for empty dictionaries unless they have type annotations indicating 2-tuple keys. Previously, the fix was incorrectly suppressed for all empty dicts due to vacuous truth in the `all()` function.

Fixes #21289

## Problem Analysis

The `is_dict_key_tuple_with_two_elements` function was designed to suppress the fix when a dictionary's keys are all 2-tuples, as unpacking tuple keys directly would change runtime behavior.

However, for empty dictionaries, `iter_keys()` returns an empty iterator, and `all()` on an empty iterator returns `true` (vacuous truth). This caused the function to incorrectly suppress fixes for empty dicts, even when there was no indication that future keys would be 2-tuples.

## Approach

1. **Detect empty dictionaries**: Added a check to identify when a dict literal has no keys.

2. **Handle annotated empty dicts**: For empty dicts with type annotations:
   - Parse the annotation to check if it's `dict[tuple[T1, T2], ...]` where the tuple has exactly 2 elements
   - Support both PEP 484 (`typing.Dict`, `typing.Tuple`) and PEP 585 (`dict`, `tuple`) syntax
   - If tuple keys are detected, suppress the fix (correct behavior)
   - Otherwise, allow the fix

3. **Handle unannotated empty dicts**: For empty dicts without annotations, allow the fix since there's no indication that keys will be 2-tuples.

4. **Preserve existing behavior**: For non-empty dicts, the original logic is unchanged - check if all existing keys are 2-tuples.

The implementation includes helper functions:
- `is_annotation_dict_with_tuple_keys()`: Checks if a type annotation specifies dict with tuple keys
- `is_tuple_type_with_two_elements()`: Checks if a type expression represents a 2-tuple

Test cases were added to verify:
- Empty dict without annotation triggers the error
- Empty dict with `dict[tuple[int, str], bool]` suppresses the error
- Empty dict with `dict[str, int]` triggers the error
- Existing tests remain unchanged

### DIFF
--- a/crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
+++ b/crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
@@ -127,13 +127,8 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
 
     if is_empty {
         // For empty dicts, check type annotation
-        // If there's an annotation indicating tuple keys, suppress the fix (return true)
-        // Otherwise, allow the fix (return false)
-        if let Some(annotation) = annotation {
-            return is_annotation_dict_with_tuple_keys(annotation, semantic);
-        }
-        // No annotation - allow the fix
-        return false;
+        return annotation
+            .is_some_and(|annotation| is_annotation_dict_with_tuple_keys(annotation, semantic));
     }
 
     // For non-empty dicts, check if all keys are 2-tuples


crates/ruff_linter/resources/test/fixtures/pylint/dict_iter_missing_items.py
```diff
@@ -30,3 +30,19 @@
...
     pass
+
+# Empty dict cases
+empty_dict = {}
+empty_dict["x"] = 1
+for k, v in empty_dict:
+    pass
+
+empty_dict_annotated_tuple_keys: dict[tuple[int, str], bool] = {}
+empty_dict_annotated_tuple_keys[("x", "y")] = True
```
ntBre: I think it would be interesting to have these two lines in separate test cases, i.e.

```py
empty_dict_annotated_tuple_keys: dict[tuple[int, str], bool] = {}
for k, v in empty_dict_annotated_tuple_keys:
    pass

empty_dict_annotated_tuple_keys = {}
empty_dict_annotated_tuple_keys[("x", "y")] = True
for k, v in empty_dict_annotated_tuple_keys:
    pass
```

I don't think we have to skip the second case even if it's technically incorrect because it's likely to be rare, as mentioned on the issue. But it seems useful to capture our behavior in this case.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+        Stmt::AnnAssign(ann_assign_stmt) => {
+            let Some(value) = ann_assign_stmt.value.as_ref() else {
+                return false;
+            };
+            let Expr::Dict(dict_expr) = value.as_ref() else {
+                return false;
+            };
+            (dict_expr, Some(ann_assign_stmt.annotation.as_ref()))
+        }
+        _ => return false,
```
ntBre:
```suggestion
    let (value, annotation) = match statement {
        Stmt::Assign(assign_stmt) => {
            (assign_stmt.value, None)
        }
        Stmt::AnnAssign(ann_assign_stmt) => {
            let Some(value) = ann_assign_stmt.value.as_ref() else {
                return false;
            };
            (value, Some(ann_assign_stmt.annotation.as_ref()))
        }
        _ => return false,
```

It might even be nicer with something like:

```rust
        Stmt::AnnAssign(ast::StmtAnnAssign{ value: Some(value), annotation, .. }) => {
            (value, Some(annotation))
        }
```

and then we can keep/share the `Expr::Dict` check below instead of duplicating it in both match arms.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
     };
 
-    let Expr::Dict(dict_expr) = &*assign_stmt.value else {
+    // Check if dict is empty
+    let is_empty = dict_expr.iter_keys().next().is_none();
```
ntBre: `ExprDict` has an `is_empty` method, which I think would work here.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+
+    if is_empty {
+        // For empty dicts, check type annotation
+        if let Some(annotation) = annotation {
+            // Check if annotation is dict[tuple[...], ...] where tuple has 2 elements
+            return is_annotation_dict_with_tuple_keys(annotation, semantic);
+        }
+        // Empty dict without annotation should allow the fix
         return false;
```
ntBre: I think this is all just:

```suggestion
        return annotation.is_some_and(|annotation| is_annotation_dict_with_tuple_keys(annotation, semantic));
```

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+fn is_annotation_dict_with_tuple_keys(annotation: &Expr, semantic: &SemanticModel) -> bool {
+    // Handle stringized annotations
+    let annotation = if let Expr::StringLiteral(_) = annotation {
+        // For stringized annotations, we'd need to parse them, but for now,
+        // we'll be conservative and return false (allow the fix)
+        return false;
+    } else {
+        annotation
+    };
+
```
ntBre: If we're not handling strings, which I think is fine, we can just omit this check since it's redundant with the one below.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+    // Check if it's a subscript: dict[...]
+    let Expr::Subscript(subscript) = annotation else {
+        return false;
+    };
+
+    // Check if the value is `dict`
+    let value_name = match subscript.value.as_ref() {
+        Expr::Name(name) => name.id.as_str(),
+        _ => return false,
+    };
```
ntBre: Should this just be a let-else as well?

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+
+    // Check if it's dict or typing.Dict
+    if value_name != "dict" && !semantic.match_typing_expr(subscript.value.as_ref(), "Dict") {
+        return false;
+    }
```
ntBre: I think we should check that `dict` has its builtin binding with `match_builtin_expr`.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+
+    // Extract the slice (should be a tuple: (key_type, value_type))
+    let Expr::Tuple(tuple) = subscript.slice.as_ref() else {
+        return false;
+    };
+
+    // dict[K, V] format - check if K is tuple with 2 elements
+    if let Some(key_type) = tuple.elts.first() {
+        return is_tuple_type_with_two_elements(key_type, semantic);
+    }
```
ntBre, line 162: It might make sense to do something like:

```suggestion
    // dict[K, V] format - check if K is tuple with 2 elements
    if let [key, _value] = tuple.elts.as_slice() {
        return is_tuple_type_with_two_elements(key, semantic);
    }
```

just to make sure there's the expected number of elements.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+fn is_tuple_type_with_two_elements(expr: &Expr, semantic: &SemanticModel) -> bool {
+    // Handle tuple[...] subscript
+    if let Expr::Subscript(subscript) = expr {
+        let value_name = match subscript.value.as_ref() {
+            Expr::Name(name) => name.id.as_str(),
+            _ => return false,
+        };
+
+        // Check if it's tuple or typing.Tuple
+        if value_name == "tuple" || semantic.match_typing_expr(subscript.value.as_ref(), "Tuple") {
```
ntBre: Again I think we need to check for builtin bindings to `tuple` to make sure it's not shadowed.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,109 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
+            // Check the slice - tuple[T1, T2] or tuple[T1, T2, ...]
+            if let Expr::Tuple(tuple_slice) = subscript.slice.as_ref() {
+                // For PEP 484: tuple[T1, T2, ...], the last element might be ...
```
ntBre: I don't think this is a valid annotation, but I could be wrong. I think the ellipsis can only come after the first element. I would probably just check the `tuple.len()` here.

crates/ruff_linter/src/rules/pylint/rules/dict_iter_missing_items.rs
```diff
@@ -108,15 +108,77 @@ fn is_dict_key_tuple_with_two_elements(binding: &Binding, semantic: &SemanticMod
...
         return false;
     };
 
+    // Check if dict is empty
+    let is_empty = dict_expr.is_empty();
+
+    if is_empty {
+        // For empty dicts, check type annotation
+        return annotation
+            .is_some_and(|annotation| is_annotation_dict_with_tuple_keys(annotation, semantic));
```
ntBre: I may have gotten this backwards, sorry. Should it be `is_none_or` instead of `is_some_and`? I'm wondering because I was expecting a diagnostic at line 46 of the new tests, but we're not getting one.

danparizher: Thanks for catching that. The logic was correct (`is_some_and` is right), but I made it explicit for clarity.

For empty dicts:
- With annotation indicating tuple keys → suppress fix (return `true`)
- Without annotation → allow fix (return `false`)

The `is_some_and` chain was correct, but I refactored it to an explicit `if let Some` pattern to make the behavior clearer.

ntBre: Oh I see, the two test cases were interfering because they both referred to the same annotation above. I think the `is_some_and` is clear then, I was mostly just confused about the end result. I'll put it back. Thanks for fixing the test!

ntBre (approved):
Thanks!

PR #21291, opened 2025-11-06
[ty] provide `import` completion when in `from <name> <name>` statement

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1494

## Test Plan

Add a test showing if we are in `from <name> <name> ` we provide the keyword completion "import"

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -160,6 +160,20 @@ impl<'db> Completion<'db> {
                 .and_then(|ty| imp(db, ty, &CompletionKindVisitor::default()))
         })
     }
+
+    fn keyword(name: &str) -> Self {
+        Completion {
+            name: name.into(),
+            insert: None,
+            ty: None,
+            kind: Some(CompletionKind::Keyword),
+            module_name: None,
+            import: None,
+            builtin: false,
+            is_type_check_only: false,
+            documentation: None,
+        }
+    }
 }
 
 /// The "kind" of a completion.
@@ -212,14 +226,16 @@ pub fn completion<'db>(
     offset: TextSize,
 ) -> Vec<Completion<'db>> {
     let parsed = parsed_module(db, file).load(db);
-
     let tokens = tokens_start_before(parsed.tokens(), offset);
+    let typed = find_typed_text(db, file, &parsed, offset);
 
-    if is_in_comment(tokens) || is_in_string(tokens) || is_in_definition_place(db, tokens, file) {
+    if is_in_no_completions_place(db, tokens, file) {
         return vec![];
     }
+    if let Some(completions) = only_keyword_completion(tokens, typed.as_deref()) {
+        return vec![completions];
+    }
 
-    let typed = find_typed_text(db, file, &parsed, offset);
     let typed_query = typed
         .as_deref()
         .map(QueryPattern::new)
@@ -307,6 +323,17 @@ fn add_keyword_value_completions<'db>(
             documentation: None,
         });
     }
+}
+
+/// When the tokens indicate that the last token should be precisely one
+/// possible keyword, we provide a single completion for it.
+///
+/// `typed` should be the text that we think the user has typed so far.
+fn only_keyword_completion<'db>(tokens: &[Token], typed: Option<&str>) -> Option<Completion<'db>> {
+    if is_import_from_incomplete(tokens, typed) {
+        return Some(Completion::keyword("import"));
+    }
+    None
 }
 
 /// Adds completions not in scope.
@@ -801,6 +828,67 @@ fn import_tokens(tokens: &[Token]) -> Option<(&Token, &Token)> {
     None
 }
 
+/// Looks for the start of a `from module <CURSOR>` statement.
+///
+/// If found, `true` is returned.
+///
+/// `typed` should be the text that we think the user has typed so far.
+fn is_import_from_incomplete(tokens: &[Token], typed: Option<&str>) -> bool {
+    // N.B. The implementation here is very similar to
+    // `from_import_tokens`. The main difference is that
+    // we're just looking for whether we should suggest
+    // the `import` keyword. So this is a little simpler.
+
+    use TokenKind as TK;
+
+    const LIMIT: usize = 1_000;
+
+    /// A state used to "parse" the tokens preceding the user's cursor,
+    /// in reverse, to detect a "from import" statement.
+    enum S {
+        Start,
+        ImportKeyword,
+        ModulePossiblyDotted,
+        ModuleOnlyDotted,
+    }
+
+    let mut state = S::Start;
+    if typed.is_none() {
+        state = S::ImportKeyword;
+    }
+    // Move backward through the tokens until we get to
+    // the `from` token.
+    for token in tokens.iter().rev().take(LIMIT) {
+        state = match (state, token.kind()) {
+            // Match an incomplete `import` keyword.
+            //
+            // It's okay to pop off a newline token here initially,
+            // since it may occur before the user starts typing
+            // `import` but after the module name.
+            (S::Start, TK::Newline | TK::Name | TK::Import) => S::ImportKeyword,
+            // We are a bit more careful with how we parse the module
+            // here than in `from_import_tokens`. In particular, we
+            // want to make sure we don't incorrectly suggest `import`
+            // for `from os.i<CURSOR>`. If we aren't careful, then
+            // `i` could be considered an incomplete `import` keyword
+            // and `os.` is the module. But of course, ending with a
+            // `.` (unless the entire module is dots) is invalid.
+            (S::ImportKeyword, TK::Dot | TK::Ellipsis) => S::ModuleOnlyDotted,
+            (S::ImportKeyword, TK::Name | TK::Case | TK::Match | TK::Type | TK::Unknown) => {
+                S::ModulePossiblyDotted
+            }
+            (S::ModuleOnlyDotted, TK::Dot | TK::Ellipsis) => S::ModuleOnlyDotted,
+            (
+                S::ModulePossiblyDotted,
+                TK::Name | TK::Dot | TK::Ellipsis | TK::Case | TK::Match | TK::Type | TK::Unknown,
+            ) => S::ModulePossiblyDotted,
+            (S::ModulePossiblyDotted | S::ModuleOnlyDotted, TK::From) => return true,
+            _ => return false,
+        };
+    }
+    false
+}
+
 /// Looks for the text typed immediately before the cursor offset
 /// given.
 ///
@@ -815,7 +903,10 @@ fn find_typed_text(
     let source = source_text(db, file);
     let tokens = tokens_start_before(parsed.tokens(), offset);
     let last = tokens.last()?;
-    if !matches!(last.kind(), TokenKind::Name) {
+    // It's odd to include `TokenKind::Import` here, but it
+    // indicates that the user has typed `import`. This is
+    // useful to know in some contexts.
+    if !matches!(last.kind(), TokenKind::Name | TokenKind::Import) {
         return None;
     }
     // This one's weird, but if the cursor is beyond
@@ -828,6 +919,11 @@ fn find_typed_text(
         return None;
     }
     Some(source[last.range()].to_string())
+}
+
+/// Whether the last token is in a place where we should not provide completions.
+fn is_in_no_completions_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
+    is_in_comment(tokens) || is_in_string(tokens) || is_in_definition_place(db, tokens, file)
 }
 
 /// Whether the last token is within a comment or not.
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Only showing `import` in a `from <NAME> <NAME2>` position does make sense to me. I was surprised to see that Pylance shows all completions, but gives `import` a higher ranking

MatthewMckee4:
For some reason i dont get completions in playground locally for `import` when i have just `a`
image

This could be a playground thing though?

MichaReiser:
I just checked out this PR and built the playground and this works as expected when you have `from types i` but it doesn't work when you have `from types aa`. 

This could either be because the playground applies its own filtering (not sure if there's a way for us to disable it, do you want to take a look) or because our own "text prefix matching"

BurntSushi:
Yeah we may not be able to fully control whether `import` is shown for `from module import aa<CURSOR>`. LSP clients can and will do their own filtering AIUI. I'm not sure if there is a foolproof way to override that.

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,19 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+    is_in_import_statement(tokens)
+}
+
+fn is_in_import_statement<'db>(tokens: &[Token]) -> Option<Completion<'db>> {
+    if token_suffix_by_kinds(tokens, [TokenKind::From, TokenKind::Name, TokenKind::Name]).is_some()
+    {
+        return Some(Completion::keyword("import"));
+    }
+    None
+}
```
AlexWaygood: this works well for the simple case of `from <NAME> <CURSOR>`, but doesn't work for:
- `from <NAME>.<NAME> <CURSOR>`
- `from .<NAME> <CURSOR>`
- `from ...<NAME>.<NAME> <CURSOR>`

etc.

image

crates/ty_ide/src/completion.rs
```diff
@@ -215,10 +229,14 @@ pub fn completion<'db>(
...
     let tokens = tokens_start_before(parsed.tokens(), offset);
 
-    if is_in_comment(tokens) || is_in_string(tokens) || is_in_definition_place(db, tokens, file) {
+    if is_in_no_completions_place(db, tokens, file) {
         return vec![];
     }
 
+    if let Some(completions) = keyword_completions(tokens) {
+        return vec![completions];
+    }
```
AlexWaygood, line 237: I don't think it's true in the long term that all locations where we might want to offer a keyword completion are places where we definitely won't want to offer any non-keyword completions. For example `pass` (a keyword) and `print` (a non-keyword) should both be included in our completion suggestions in this example:

```py
match x:
    case int():
        p<CURSOR>
```

so the semantics you're adding in this PR are totally correct, but that might imply a different code structure here that would be more extensible in the future

crates/ty_ide/src/completion.rs
AlexWaygood: the full grammar for import statements given at https://docs.python.org/3/reference/grammar.html is:

```
import_stmt:
    | import_name
    | import_from

# Import statements
# -----------------

import_name: 'import' dotted_as_names 
# note below: the ('.' | '...') is necessary because '...' is tokenized as ELLIPSIS
import_from:
    | 'from' ('.' | '...')* dotted_name 'import' import_from_targets 
    | 'from' ('.' | '...')+ 'import' import_from_targets 
import_from_targets:
    | '(' import_from_as_names [','] ')' 
    | import_from_as_names !','
    | '*' 
import_from_as_names:
    | ','.import_from_as_name+ 
import_from_as_name:
    | NAME ['as' NAME ] 

dotted_as_names:
    | ','.dotted_as_name+ 
dotted_as_name:
    | dotted_name ['as' NAME ] 

dotted_name:
    | dotted_name '.' NAME 
    | NAME
```

MatthewMckee4:
Thank you, I will revisit this in a few days, when I am back home.

crates/ty_ide/src/completion.rs
MatthewMckee4: Good point, it seems we need some indication of when keywords are the only completions we should provide.

I will think more about this

BurntSushi: Right. What we'll want it some way to identify something about the context in which completions are being offered. There are other considerations in play here, like `raise <CURSOR>` should have the suggestions altered to things that are appropriate for `raise`. There are likely other examples.

crates/ty_ide/src/completion.rs
BurntSushi: You might be able to copy this routine and tweak it a bit:

https://github.com/astral-sh/ruff/blob/0c7b93363e6b5b43b8f36adbd6e3adf1a13374c2/crates/ty_ide/src/completion.rs#L667-L670

crates/ty_ide/src/completion.rs
```diff
@@ -4216,6 +4252,32 @@ type <CURSOR>
...
+
+        completions_test.contains("import");
+
+        let completions = completions_test.completions();
+
+        assert!(completions.len() == 1);
+    }
+
+    #[test]
+    fn from_import_random_name_suggests_import() {
```
BurntSushi, line 4360: Can you add a comment noting that this might not work the same way in real LSP clients (since they might do their own filtering based on what has been typed)?

crates/ty_ide/src/completion.rs
MatthewMckee4: I can implement something like this in another PR

crates/ty_ide/src/completion.rs
MatthewMckee4: I have done my best to mirror this.

crates/ty_ide/src/completion.rs
```diff
@@ -4216,6 +4302,95 @@ type <CURSOR>
...
+            .completion_test_builder();
+
+        let completions_test = builder.build();
+
+        completions_test.contains("import");
+
+        let completions = completions_test.completions();
+
+        assert!(completions.len() == 1);
+    }
```
AlexWaygood, line 4445: A version of this test with _three_ preceding dots doesn't pass. The reason is that `from ...foo` doesn't parse as `[<FROM>, <DOT>, <DOT>, <DOT>, <NAME("foo")>]`, it parses as `[<FROM>, <ELLIPSIS>, <NAME("foo")>]`.

You want to make this change:

```diff
diff --git a/crates/ty_ide/src/completion.rs b/crates/ty_ide/src/completion.rs
index d6896f3804..adf107d7c4 100644
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -919,7 +919,8 @@ fn from_import_statement_keyword_completions<'db>(tokens: &[Token]) -> Option<Co
     // Skip over '('.' | '...')*'
     let dots_count = tokens_after_from
         .iter()
-        .take_while(|t| t.kind() == TokenKind::Dot)
+        .map(Token::kind)
+        .take_while(|t| matches!(t, TokenKind::Dot | TokenKind::Ellipsis))
         .count();
 
     let tokens_after_dots = &tokens_after_from[dots_count..];
@@ -4391,6 +4392,23 @@ type <CURSOR>
         assert!(completions.len() == 1);
     }
 
+    #[test]
+    fn from_import_nested_very_dotted_name_relative_import_suggests_import() {
+        let builder = CursorTest::builder()
+            // N.B. the `...` tokenizes as `TokenKind::Ellipsis`
+            .source("src/main.py", "from ...foo i<CURSOR>")
+            .source("foo.py", "")
+            .completion_test_builder();
+
+        let completions_test = builder.build();
+
+        completions_test.contains("import");
+
+        let completions = completions_test.completions();
+
+        assert!(completions.len() == 1);
+    }
+
     /// A way to create a simple single-file (named `main.py`) completion test
     /// builder.
     ///
```

MatthewMckee4: Ah, thank you.

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,70 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+    if tokens_len >= 3 {
+        let last_token = tokens.last().unwrap();
+        let second_last_token = &tokens[tokens_len - 2];
+        let tokens_to_third_last = &tokens[0..tokens_len - 2];
+        return is_dotted_name(tokens_to_third_last)
+            && second_last_token.kind() == TokenKind::Dot
+            && last_token.kind() == TokenKind::Name;
+    }
+    false
+}
```
AlexWaygood: I find this a bit easier to understand:

```suggestion
fn is_dotted_name(tokens: &[Token]) -> bool {
    match tokens {
        [] | [_, _] => false,
        [tok] => tok.kind() == TokenKind::Name,
        [dotted_name @ .., penultimate, last] => {
            last.kind() == TokenKind::Name
                && penultimate.kind() == TokenKind::Dot
                && is_dotted_name(dotted_name)
        }
    }
}
```

AlexWaygood (approved):
Thank you!

crates/ty_ide/src/completion.rs
MatthewMckee4: Very nice, my functional programming class clearly hasn't sunk in enough.

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,66 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+/// Checks whether the tokens indicate that the last token is in the `import` section
+/// of the `import_from` definition from <https://docs.python.org/3/reference/grammar.html>.
+/// Being in this section means we match all of the parts of the definition up to `import`
+/// and we have a `Name` token in place of the `import`.
+fn from_import_statement_keyword_completions<'db>(tokens: &[Token]) -> Option<Completion<'db>> {
+    if tokens.is_empty() {
+        return None;
+    }
+
+    let from_index = tokens.iter().position(|t| t.kind() == TokenKind::From)?;
```
MichaReiser: The search here seems to be too permissive to me. What's preventing it from going across logical lines? 

Can you add some tests with multiple statements? I suspect that we now incorrectly match on import completions if there's **any** import statement before `<CURSOR>`.

MatthewMckee4: Good point, will add that

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,66 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+/// Checks whether the tokens indicate that the last token is in the `import` section
+/// of the `import_from` definition from <https://docs.python.org/3/reference/grammar.html>.
+/// Being in this section means we match all of the parts of the definition up to `import`
+/// and we have a `Name` token in place of the `import`.
+fn from_import_statement_keyword_completions<'db>(tokens: &[Token]) -> Option<Completion<'db>> {
+    if tokens.is_empty() {
+        return None;
+    }
+
+    let from_index = tokens.iter().rposition(|t| t.kind() == TokenKind::From)?;
```
MichaReiser: I still think this is incorrect as it cross logical line boundaries. E.g. what if

```py
from module impo

a<CURSOR>
```

We would want to show completions for `a` and not suggest `import`

MatthewMckee4: I am maybe misunderstanding you. we do still provide completions for `a` here.
image

MichaReiser: I haven't checked out the PR. What I'm concerned with the current approach is that `tokens` is anything before `<CURSOR>` (unless this isn't the case?). That means, it's possible for `tokens.iter().rposition()` to find the `from` even though it belongs to an entirely different statement.

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,66 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+
+    let tokens_after_dots = &tokens_after_from[dots_count..];
+
+    let (last_token, tokens_after_dots_before_last_name) = tokens_after_dots.split_last()?;
+
+    if last_token.kind() == TokenKind::Name {
+        // If we have `from .('.' | '...')* <name>` we can suggest `import`.
+        // We check that we must have at least one `.` because we don't want to suggest `import` for `from <name>`.
+        //
+        // If we have `from ('.' | '...')* dotted_name <name>` we can suggest `import`.
```
BurntSushi: Can you make sure comments are wrapped to 99 columns (inclusive) please? There are some other comments below that need to be wrapped too.

crates/ty_ide/src/completion.rs
```diff
@@ -871,6 +894,66 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
...
+        if (tokens_after_dots_before_last_name.is_empty() && dots_count > 0)
+            || is_dotted_name(tokens_after_dots_before_last_name)
+        {
+            return Some(Completion::keyword("import"));
+        }
+    }
+
+    None
+}
```
BurntSushi: I think this routine should be modeled on this existing function:

https://github.com/astral-sh/ruff/blob/f44598dc113a7380ca52d6c969431ca929a93473/crates/ty_ide/src/completion.rs#L667-L756

Maybe there is a way to factor out the common logic, but even if there isn't, just copying the _approach_ seems like the right move here. I agree with @MichaReiser above. And notice that the function above is careful to not do too much look-behind on the token sequence. It also handles things like ellipsis, non-logical newlines and so on.

crates/ty_ide/src/completion.rs
```diff
@@ -4216,6 +4299,126 @@ type <CURSOR>
...
+
+        let completions = completions_test.completions();
+
+        assert!(completions.len() == 1);
+    }
+
+    #[test]
+    fn from_import_two_imports_suggests_import() {
+        let builder =
+            completion_test_builder("from collections.abc import Sequence\nfrom typing i<CURSOR>");
```
BurntSushi: For something really short, I'm fine with the `\n` here. But this starts to make very long lines. I'd rather see this written out using actual new lines:

```
        let builder = completion_test_builder(
            "from collections.abc import Sequence
             from typing i<CURSOR>"
        );
```

The test code should automatically handle dedenting.

BurntSushi:
@MatthewMckee4 I'm going to take over this PR since we'd like to get this fixed. And I'm planning to work on more stuff around keyword completions this week, which will really want to build on top of this PR. Thank you. :-)

MatthewMckee4:
Cool, apologies for the delays, I've been a little busy. Thank you for the feedback too!

AlexWaygood:
> Cool, apologies for the delays, I've been a little busy.

never feel like you need to apologise for unpaid volunteer work :-) we really appreciate the contribution!!

PR #21292, opened 2025-11-06
Require ignore 0.4.24 in `Cargo.toml`

musicinmybrain:
## Summary

Since 4c4ddc8c29e, ruff uses the `WalkBuilder::current_dir` API [introduced in `ignore` version 0.4.24](https://diff.rs/ignore/0.4.23/0.4.24/src%2Fwalk.rs), so it should explicitly depend on this minimum version.

See also https://github.com/astral-sh/ruff/pull/20979.

## Test Plan

Source inspection verifies this version is necessary; no additional testing is required since `Cargo.lock` already has (at least) this version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,7 @@ hashbrown = { version = "0.16.0", default-features = false, features = [
     "inline-more",
 ] }
 heck = "0.5.0"
-ignore = { version = "0.4.22" }
+ignore = { version = "0.4.24" }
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

T-256:
I wonder if there is anyway to prevent this misconfiguration in future. IOW, could test against minimum declared dependencies versions set in `Cargo.toml`?

Issue #10070, opened 2024-02-21
Ability to add a reason along with `--add-noqa`


ashrub-holvi:
Hi!

[Ruff can automatically add `noqa` directives](https://docs.astral.sh/ruff/linter/#inserting-necessary-suppression-comments), but they don't have any info about how they were added:
```
  # noqa: S308
```
and it may be difficult to distinguish intentionally/manually added `noqa` and those automatically added.

I think, main use case is migration to ruff for existing codebase without fixing everything at once, but it should be clear that those `noqa` added automatically and should be handled at some point, something like that:
```
  # noqa: S308 FIXME when you are here
```
Downside of it - adding longer comment will produce more formatting changes, so, perhaps
```
  # noqa: S308 auto
```
or even ruff may have some argument for provide the text of comment.

kaddkaka:
Do you really want to `# noqa` them if you intend to fix them?

In our CI we have say 10000 ruff lint violations, and we accept if this number shrinks but fail if it ever increases.

kaddkaka:
Thanks now I understand the situation much better. For me #1149 would be a more valuable feature. 

If you really want to do this it can be done using ruff+other tool. 

Example on how to achieve this today with ruff+git+vim:
```
$ ruff check --add-noqa
$ git jump diff 
:cdo normal! A (added automatically by ruff) 
```

Perhaps this eliminates the feature need in ruff?

MichaReiser:
This should be fairly easy to implement and shouldn't even need a new CLI flag if we make `--add-noqa` accept a value.

engnatha:
This isn't working for me on `0.14.10`. `--add-noqa` works fine, but providing a reason silently exits without fixing any errors.

Edit: Ah I was missing the `requires_equals`

PR #21294, opened 2025-11-06
Add option to provide a reason to `--add-noqa`

MichaReiser:
## Summary

This PR adds an optional reason to `--add-noqa` that will be added to every inserted `noqa` comment. 

How to use:

```
ruff check --add-noqa "Disallow usage of method"
```

Fixes https://github.com/astral-sh/ruff/issues/10070

## Test Plan

Added CLI tests

### DIFF
--- a/crates/ruff/src/args.rs
+++ b/crates/ruff/src/args.rs
@@ -405,8 +405,13 @@ pub struct CheckCommand {
     )]
     pub statistics: bool,
     /// Enable automatic additions of `noqa` directives to failing lines.
+    /// Optionally provide a reason to append after the codes.
     #[arg(
         long,
+        value_name = "REASON",
+        default_missing_value = "",
+        num_args = 0..=1,
+        require_equals = true,
         // conflicts_with = "add_noqa",
         conflicts_with = "show_files",
         conflicts_with = "show_settings",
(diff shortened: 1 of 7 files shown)


crates/ruff/src/args.rs
```diff
@@ -405,8 +405,13 @@ pub struct CheckCommand {
     )]
     pub statistics: bool,
     /// Enable automatic additions of `noqa` directives to failing lines.
+    /// Optionally provide a reason to append after the codes.
     #[arg(
         long,
+        value_name = "REASON",
+        default_missing_value = "",
+        num_args = 0..=1,
+        require_equals = true,
```
ntBre, line 414: Do we want `require_equals = true`? I usually prefer not to use the `=` (like in the PR summary), but I'm fine either way.

MichaReiser: We might have to but I'm curious to hear your opinion too. I, honestly, didn't realize that claude added it until you pointed it out. So I tried to remove it but then run into issues when using `--add-noqa` with stdin filenames:

```
ruff check --add-noqa -
```

clap no incorrectly parses the `-` as the noqa comment. 

Or running

```
ruff cehck --add-noqa file.py
```

Will use `file.py` as the noqa reason rather than only adding noqa comments to `file.py`. That's why I think we should keep it as is.

ntBre: Ah that makes sense! Seems fine as-is then.

PR #21296, opened 2025-11-06
[ty] Add support for `Literal`s in implicit type aliases

sharkdp:
## Summary

Add support for `Literal` types in implicit type aliases.

## Ecosystem analysis

This looks good to me, true positives and known problems.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -313,10 +313,14 @@ We reject invalid uses:
 
 ```py
 # error: [invalid-type-form] "Type arguments for `Literal` must be `None`, a literal value (int, bool, str, or bytes), or an enum member"
-def _(weird: Literal[int]):
+LiteralInt = Literal[int]
+
+reveal_type(LiteralInt)  # revealed: Unknown
+
+def _(weird: LiteralInt):
     reveal_type(weird)  # revealed: Unknown
 
-# error: [invalid-type-form] "`typing.Literal` instances are not allowed in type expressions"
+# error: [invalid-type-form] "`Literal[26]` is not a generic class"
 def _(weird: IntLiteral1[int]):
     reveal_type(weird)  # revealed: Unknown
 ```
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -814,11 +814,12 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     self.infer_type_expression(slice);
                     todo_type!("Generic specialization of types.UnionType")
                 }
-                KnownInstanceType::Literal(_) => {
+                KnownInstanceType::Literal(ty) => {
                     self.infer_type_expression(slice);
                     if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, subscript) {
                         builder.into_diagnostic(format_args!(
-                            "`typing.Literal` instances are not allowed in type expressions",
+                            "`{ty}` is not a generic class",
+                            ty = ty.to_union(self.db()).display(self.db())
                         ));
                     }
                     Type::unknown()


crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -255,6 +263,64 @@ def _(list_or_tuple: ListOrTuple[int]):
...
+    reveal_type(bool_literal)  # revealed: Literal[True]
+    reveal_type(mixed_literals)  # revealed: Literal[1, "a", True] | None
+    reveal_type(enum_literal)  # revealed: Literal[Color.RED]
+```
+
+We reject invalid uses:
+
+```py
+# error: [invalid-type-form] "Type arguments for `Literal` must be `None`, a literal value (int, bool, str, or bytes), or an enum member"
+def _(weird: Literal[int]):
```
carljm: This case doesn't involve an implicit type alias, and it is testing a behavior which is unchanged by this PR, so it's not clear why it appears in this PR or in this test file?

The new version of this diagnostic added in this PR would I think only trigger if `Literal[int]` appeared in a value expression, but here it's in an annotation.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -9643,6 +9647,29 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 );
             }
         }
+        if value_ty == Type::SpecialForm(SpecialFormType::Literal) {
+            match self.infer_literal_parameter_type(slice) {
+                Ok(result) => {
+                    return Type::KnownInstance(KnownInstanceType::Literal(TypeList::singleton(
+                        self.db(),
+                        result,
+                    )));
```
sharkdp, line 9656: I think the `TypeList` type is useful, but `Literal` should probably use a simpler wrapper that only contains a single type. `Optional` and `Annotated` also just need a single element. I will change this in a follow-up PR.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -255,6 +263,64 @@ def _(list_or_tuple: ListOrTuple[int]):
...
+    reveal_type(weird)  # revealed: Unknown
+
+# error: [invalid-type-form] "`typing.Literal` instances are not allowed in type expressions"
+def _(weird: IntLiteral1[int]):
```
carljm: I have two thoughts about this error message.

1. It's a little weird to talk about "`typing.Literal` instances", because at runtime and in typeshed, `typing.Literal` is not a class and there are no instances of it. It's a special form, and indexing it returns an instance of `typing._LiteralGenericAlias`. I think maybe a term that would hew a bit closer to runtime, and align more with the user's likely intuition that `T = Literal[1]` defines a type alias, would be to call them "`typing.Literal` aliases", maybe?

2) The error message doesn't quite match what I see happening in the code? What you are describing as a "`typing.Literal` instance" (and I suggest calling a "`typing.Literal` alias") is `IntLiteral1` -- and that most certainly is allowed in a type expression! What isn't allowed is to subscript it again. So the error I'd expect to see here would be more like "`typing.Literal` aliases cannot be subscripted." (For reference, pyright gives the somewhat obscure "Expected no type arguments for class int", where I think "class int" results from an implicit promotion from the literal type, and mypy gives "Bad number of arguments for type alias, expected 0, got 1"). Both of these indicate that the mistake is not `IntLiteral1` appearing in a type expression, it's the attempt to subscript it (or pass it type arguments).

crates/ty_python_semantic/src/types.rs
```diff
@@ -8949,32 +8956,46 @@ impl<'db> TypeVarBoundOrConstraints<'db> {
     }
 }
 
-/// An instance of `types.UnionType`.
+/// A salsa-interned list of types.
```
AlexWaygood, line 8959: I'm not wild about the fact that -- depending on context -- this type can sometimes hold a list of value-expression types (if it's the wrapped data inside a `KnownInstance::UnionType` variant) and sometimes hold a list of type-expression types (if it's the wrapped data inside a `KnownInstance::Literal` variant). That feels a little bug-prone? E.g. it would be incorrect to call the `.to_union()` method on a `TypeList` wrapped inside a `KnownInstance::Union` variant, because you need to first convert them all to type expressions. But it would be incorrect to convert them all to type expressions on the elements in a `TypeList` inside a `KnownInstance::Literal` variant, because those have all already been interpreted as type expressions eagerly when building the type list.

I think I'd prefer to have different structs for the wrapped data inside the two variants, since they have pretty different semantics.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -814,6 +814,15 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     self.infer_type_expression(slice);
                     todo_type!("Generic specialization of types.UnionType")
                 }
+                KnownInstanceType::Literal(_) => {
+                    self.infer_type_expression(slice);
+                    if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, subscript) {
+                        builder.into_diagnostic(format_args!(
+                            "`typing.Literal` instances are not allowed in type expressions",
```
AlexWaygood: If I understand correctly, we only reach this branch for cases like these:

```py
from typing import Literal

X = Literal[1, 2]

def f(a: X[int], b: Literal["foo"][str]):
    ...
```

In which case, I think I'd prefer something a bit closer to the runtime error message, which feels clearer to me:

```pytb
>>> from typing import Literal
>>> Literal[2][int]
    Literal[2][int]
    ~~~~~~~~~~^^^^^
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 432, in inner
    return func(*args, **kwds)
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 1473, in __getitem__
    raise TypeError(f"{self} is not a generic class")
TypeError: typing.Literal[2] is not a generic class
```

carljm: Discussed above -- doesn't seem like the right error message for this case

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: Because in a case like `Literal[1, 2, 3]`, it's still just a "single" union type?

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
sharkdp: Yes, thank you. I changed it and the error message now reads:

> [invalid-type-form] "`Literal[26]` is not a generic class"

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
sharkdp: See above, the error message now reads:

> [invalid-type-form] "`Literal[26]` is not a generic class"

crates/ty_python_semantic/src/types.rs
sharkdp: I agree. I will change this in major ways in my next PR, and will keep this in mind.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -814,6 +814,16 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     self.infer_type_expression(slice);
                     todo_type!("Generic specialization of types.UnionType")
                 }
+                KnownInstanceType::Literal(ty) => {
+                    self.infer_type_expression(slice);
+                    if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, subscript) {
+                        builder.into_diagnostic(format_args!(
+                            "`{ty}` is not a generic class",
+                            ty = ty.to_union(self.db()).display(self.db())
```
sharkdp, line 822: This looks very strange, but will turn into `ty.to_inner(…).display(…)` soon.

PR #21299, opened 2025-11-06
Fix main by using `infer_expression`

dhruvmanila:


dhruvmanila:
Yikes, the `main` commit fails which is why there are failures in the CI as it tries to build a `ty` executable from `main` as well as this PR.

MichaReiser:
Are those checks failing because main isn't building?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4895,7 +4895,9 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             self.infer_type_expression(&bound.value);
         }
         if let Some(default) = arguments.find_keyword("default") {
-            let func_ty = self.get_or_infer_expression(func, TypeContext::default());
+            let func_ty = self
+                .try_expression_type(func)
+                .unwrap_or_else(|| self.infer_expression(func, TypeContext::default()));
             if func_ty.as_class_literal().is_some_and(|class_literal| {
                 class_literal.is_known(self.db(), KnownClass::ParamSpec)
             }) {
```
dhruvmanila, line 4903: I think at this point, the expression should've already been inferred because this is in the deferred region and the function symbol should've already been inferred. So, I could directly use `self.expression_type` instead.

dhruvmanila:
> Are those checks failing because main isn't building?

Yes!

MichaReiser:
Let's merge it. We can easily change it to not use `try` if we feel confident in a follow up PR

PR #21302, opened 2025-11-06
chore: bump dist, remove old commenting workflows

woodruffw:
This removes the four `workflow_run`-based commenting workflows now that astral-sh-bot can track `workflow_run` events.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -627,18 +627,8 @@ jobs:
           cat ecosystem-result-format-preview >> ecosystem-result
           echo "" >> ecosystem-result
 
-      - name: Export pull request number
-        run: |
-          echo "${PR_NUMBER}" > pr-number
-        env:
-          PR_NUMBER: ${{ github.event.number }}
-
-      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
-        name: Upload PR Number
-        with:
-          name: pr-number
-          path: pr-number
-
+      # NOTE: astral-sh-bot uses this artifact to post comments on PRs.
+      # Make sure to update the bot if you rename the artifact.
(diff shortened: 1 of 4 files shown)


MichaReiser (commented):
Thank you.

woodruffw:
Thanks both! I'll land this in a moment, but it can be reverted if it causes any disruption.

UnknownPlatypus:
Out of curiosity, where can I see how the bot is setup ?
I would also like to drop these commenting workflows in my projects

woodruffw:
> Out of curiosity, where can I see how the bot is setup ?

It's not public at the moment, but we may consider making it public once it's bit more mature 🙂

For reference however, it's very similar to CPython's "bedevere" bot: https://github.com/python/bedevere

Mariatta Wijaya has a really great guide on building these, which our bot also follows pretty closely: https://github-app-tutorial.readthedocs.io/en/latest/creating-github-app.html

UnknownPlatypus:
Awesome, very cool. Thanks for the insights, this is very valuable :raised_hands:

j178:
I built a similar bot for commenting benchmark results on PRs for [prek](https://github.com/j178/prek): https://github.com/j178/prek-ci-bot. It listens to the `workflow_run` `completed` event, downloads the artifacts from that run, unpacks them, and then posts the results as a comment on the PR.

While working on it, I ran into the issue that for PRs coming from forks, the `workflow_run` event comes with an empty  `pull_requests` payload. So I ended up listing all the open PRs in the repo and matching them by head SHA to figure out which PR the run belonged to. It works, but I’m not really a fan of this approach. I’m also curious how `astral-sh-bot` handles this.

woodruffw:
That's exactly what astral-sh-bot does too 😅. I agree about it feeling wrong, but GitHub seemingly doesn't give us any better options here.

(You can do the PR-to-SHA matching with basically no rate limit usage by using `git ls-remote` instead of querying GitHub's API for PRs.)

PR #20224, opened 2025-09-04
[`flake8-logging-format`] fix handles grouping parens and preserves r-prefix/quotes

TaKO8Ki:
## Summary

Fixes some parts of #20151

Handles grouping parens and preserves r-prefix/quotes

- Iterate over `FStringPart` to build the format string, handling implicit string literal concatenation.
- Preserve the original quote style and `r` prefix in the generated message string.
- Replace over the expression’s parenthesized range to avoid tuple args, e.g. `logging.warning((f"{Ellipsis}"))` -> `logging.warning("%s", Ellipsis)`.
- Update tests and snapshots for nested parentheses and backslash escape case.

## Test Plan

I have updated `crates/ruff_linter/resources/test/fixtures/flake8_logging_format/G004.py`.

Issue #20151, opened 2025-08-29
G004 fix can change behavior and introduce syntax errors


dscorbett:
### Summary

The fix for [`logging-f-string` (G004)](https://docs.astral.sh/ruff/rules/logging-f-string/) transforms the argument incorrectly in various ways, changing the program’s behavior. [Example](https://play.ruff.rs/14641467-b12d-4186-bfd0-ca4dc1f99fcd):
```console
$ cat >g004_1.py <<'# EOF'
import logging
logging.warning(fr"\'{Ellipsis}")
logging.warning(f"{Ellipsis}" "!")
logging.warning((f"{Ellipsis}"))
logging.warning(f"\x5c{Ellipsis}")
# EOF

$ python g004_1.py
WARNING:root:\'Ellipsis
WARNING:root:Ellipsis!
WARNING:root:Ellipsis
WARNING:root:\Ellipsis

$ ruff --isolated check g004_1.py --select G004 --preview --fix
Found 4 errors (4 fixed, 0 remaining).

$ cat g004_1.py
import logging
logging.warning("\'%s", Ellipsis)
logging.warning("%s", Ellipsis)
logging.warning(("%s", Ellipsis))
logging.warning("\%s", Ellipsis)

$ python g004_1.py
g004_1.py:5: SyntaxWarning: invalid escape sequence '\%'
  logging.warning("\%s", Ellipsis)
WARNING:root:'Ellipsis
WARNING:root:Ellipsis
WARNING:root:('%s', Ellipsis)
WARNING:root:\Ellipsis
```

It can also introduce syntax errors. [Example](https://play.ruff.rs/4b8cff4f-804f-47f7-a0ef-7213fb4a9e2b):
```console
$ cat >g004_2.py <<'# EOF'
import logging
logging.warning(f"\n{Ellipsis}")
logging.warning(f"\x22{Ellipsis}")
# EOF

$ python g004_2.py
WARNING:root:
Ellipsis
WARNING:root:"Ellipsis

$ ruff --isolated check g004_2.py --select G004 --preview --diff 2>&1 | grep error:
error: Fix introduced a syntax error. Reverting all changes.
```

### Version

ruff 0.12.11 (c2bc15bc1 2025-08-28)

TaKO8Ki:
I’ll take this one next.

ericrbg-harmonic:
another one is something like:
```
x = 1
logging.info(f"{x=}")
```

will go to

```
logging.info("%s", x)
```

PR #21310, opened 2025-11-07
[`flake8-logging-format`] Fix G004 preview mode breaking on f-strings with control characters (`G004`)

### LINKED ISSUE #21295: [Fix error] G004 preview fix breaks if string contains newline
Minimal example:

(file: dummy.py)
```python
import logging

bar = "bar"
logging.getLogger().error(f"Lorem ipsum \n {bar}")
```

Run command `ruff check path/to/dummy.py --select G004 --fix --preview` (with nothing in config files). It seemingly tries to refactor the log line to

```python
logging.getLogger().error("Lorem ipsum 
%s", foo)
```
 and gets a syntax error.

Ruff version: 0.14.3

danparizher:
## Summary

Fixes #21295. The G004 rule's preview mode fix was generating invalid Python syntax when converting f-strings containing control characters (newlines, tabs, carriage returns) to `%` formatting in logging calls. This fix adds proper escaping of control characters and quotes when building the replacement string literal.

## Problem Analysis

When G004's preview mode converts f-strings to `%` formatting, it directly appends literal text from the f-string parts to the format string without escaping control characters. This causes syntax errors when the generated code contains unescaped newlines, tabs, or other control characters in string literals.

The root cause was in the `logging_f_string` function in `crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs`, where literal text from f-string parts was appended directly to the format string without escaping control characters (lines 60 and 72 in the original code).

## Approach

1. **Added escape helper function**: Created `escape_string_for_literal()` that properly escapes control characters (`\n`, `\t`, `\r`), backslashes (`\`), and quotes matching the quote style for use in Python string literals.

2. **Updated literal handling**: Modified the `logging_f_string` function to use the escape function when appending literal text from both:
   - `FStringPart::Literal` parts (line 83)
   - `InterpolatedStringElement::Literal` elements (lines 95-98)

3. **Added test cases**: Added test cases covering newlines, tabs, and carriage returns to ensure the fix works correctly and prevents regressions.

The fix ensures that all control characters are properly escaped in the generated Python code, making it syntactically valid while preserving the intended behavior of the original f-string.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_logging_format/G004.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_logging_format/G004.py
@@ -64,3 +64,12 @@ import logging
 
 x = 1
 logging.error(f"{x} -> %s", x)
+
+# Test cases for control characters in f-strings
+bar = "bar"
+logging.getLogger().error(f"Lorem ipsum \n {bar}")
+
+value = "test"
+logging.info(f"Tab\t{value}")
+
+logging.info(f"CR\r{value}")
--- a/crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
+++ b/crates/ruff_linter/src/rules/flake8_logging_format/rules/logging_call.rs
@@ -14,6 +14,26 @@ use crate::rules::flake8_logging_format::violations::{
 };
 use crate::{Edit, Fix};
 
+/// Escape control characters and quotes for use in a Python string literal.
+/// Escapes newlines, tabs, carriage returns, backslashes, and quotes that match the quote style.
+fn escape_string_for_literal(text: &str, quote_char: char) -> String {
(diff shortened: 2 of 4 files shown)


ntBre:
Hmm, I'm not sure this is the right approach. There's another open PR and issue related to this rule in https://github.com/astral-sh/ruff/pull/20224 and https://github.com/astral-sh/ruff/issues/20151, respectively. Many of the issues here seem to be that we're performing unnecessary string manipulations on the contents of the f-string when it would be better only to very narrowly replace the interpolations with `%s`.

Actually now that I look at it again, I think #21295 is a duplicate of #20151 too. If @TaKO8Ki plans to get back to the other PR, it might be better to close this in favor of that one (or vice versa if not). Sorry for the confusion!

PR #21312, opened 2025-11-07
[ty] name is parameter and global is a syntax error

11happy:
## Summary

This PR implements `is_bound_parameter` for ty for syntax error name is parameter and global, this is a follow up to this https://github.com/astral-sh/ruff/pull/20426

## Test Plan

mdtest and snapshots

## CC
- @ntBre

### DIFF
--- a/crates/ruff_linter/src/checkers/ast/mod.rs
+++ b/crates/ruff_linter/src/checkers/ast/mod.rs
@@ -860,23 +860,17 @@ impl SemanticSyntaxContext for Checker<'_> {
     }
 
     fn is_bound_parameter(&self, name: &str) -> bool {
-        for scope in self.semantic.current_scopes() {
-            match scope.kind {
-                ScopeKind::Class(_) => return false,
-                ScopeKind::Function(ast::StmtFunctionDef { parameters, .. })
-                | ScopeKind::Lambda(ast::ExprLambda {
-                    parameters: Some(parameters),
-                    ..
-                }) => return parameters.includes(name),
-                ScopeKind::Lambda(_)
-                | ScopeKind::Generator { .. }
-                | ScopeKind::Module
-                | ScopeKind::Type
-                | ScopeKind::DunderClassCell => {}
+        match self.semantic.current_scope().kind {
+            ScopeKind::Function(ast::StmtFunctionDef { parameters, .. }) => {
+                parameters.includes(name)
             }
+            ScopeKind::Class(_)
+            | ScopeKind::Lambda(_)
+            | ScopeKind::Generator { .. }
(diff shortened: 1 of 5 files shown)


11happy:
```
crates/ty_python_semantic/resources/mdtest/diagnostics/semantic_syntax_errors.md:386 unexpected error: 12 [unresolved-global] "Invalid global declaration of `a`: `a` has no declarations or bindings in the global scope"

crates/ty_python_semantic/resources/mdtest/diagnostics/semantic_syntax_errors.md:386 unexpected error: 12 [invalid-syntax] "name `a` is used prior to global declaration"
```

actually two more errors are also emitted along with the expected one , can we supress others ? as the changes in this PR are not related to other two errors ,

ntBre:
For the first error, I think you can just add a global `a` binding before the function definition. The second case looks interesting, I think it's basically a conflicting error with the error you're adding in this PR. I think we might want to suppress the `used prior to global declaration` error in this case so that we only report the more specific `parameter and global` error.

I checked how we handle this in [ruff](https://play.ruff.rs/b0d433c7-568d-48f2-8288-0ff9ef72c0b1) too, and it looks like we only emit the `parameter and global` syntax error, even if [PLE0118](https://docs.astral.sh/ruff/rules/load-before-global-declaration/) is also enabled.

11happy:
> For the first error, I think you can just add a global `a` binding before the function definition. The second case looks interesting, I think it's basically a conflicting error with the error you're adding in this PR. I think we might want to suppress the `used prior to global declaration` error in this case so that we only report the more specific `parameter and global` error.
> 
> I checked how we handle this in [ruff](https://play.ruff.rs/b0d433c7-568d-48f2-8288-0ff9ef72c0b1) too, and it looks like we only emit the `parameter and global` syntax error, even if [PLE0118](https://docs.astral.sh/ruff/rules/load-before-global-declaration/) is also enabled.

what do you think of this approach I have added a new symbol flag _IS_PARAMETER ? is working as expected now

ntBre:
Nice! This does seem to work. My only hesitation is that I suspect there is an existing way to obtain this information without adding adding a new `SymbolFlags` variant. I poked around a bit and didn't find anything, but it's probably worth someone on the ty team taking a look if they get the chance.

Maybe @AlexWaygood and I can have a look tomorrow during our meeting :)

carljm:
FWIW I think it's probably _possible_ to do it another way, but this way looks pretty clean and straightforward, and IMO is very much in line with the intended use of `SymbolFlags`, so no objection from me.

AlexWaygood (approved):
thank you!

PR #21317, opened 2025-11-07
[ty] fix global symbol lookup from eager scopes

mtshiba:
## Summary

cf. https://github.com/astral-sh/ruff/pull/20962

In the following code, `foo` in the comprehension was not reported as unresolved:

```python
# error: [unresolved-reference] "Name `foo` used when not defined"
foo
foo = [
    # no error!
    # revealed: Divergent
    reveal_type(x) for _ in () for x in [foo]
]

baz = [
    # error: [unresolved-reference] "Name `baz` used when not defined"
    # revealed: Unknown
    reveal_type(x) for _ in () for x in [baz]
]
```

In fact, this is a more serious bug than it looks: for `foo`, [`explicit_global_symbol` is called](https://github.com/astral-sh/ruff/blob/6cc3393ccd9059439d9c1325e0e041db1d7481af/crates/ty_python_semantic/src/types/infer/builder.rs#L8052), causing a symbol that should actually be `Undefined` to be reported as being of type `Divergent`.

This PR fixes this bug. As a result, the code in `mdtest/regression/pr_20962_comprehension_panics.md` no longer panics.

## Test Plan

`corpus\cyclic_symbol_in_comprehension.py` is added.
New tests are added in `mdtest/comprehensions/basic.md`.

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -315,6 +315,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
     // Records snapshots of the place states visible from the current eager scope.
     fn record_eager_snapshots(&mut self, popped_scope_id: FileScopeId) {
         let popped_scope = &self.scopes[popped_scope_id];
+        let popped_scope_is_annotation_scope = popped_scope.kind().is_annotation();
 
         // If the scope that we just popped off is an eager scope, we need to "lock" our view of
         // which bindings reach each of the uses in the scope. Loop through each enclosing scope,
@@ -330,6 +331,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
         // ```
         for enclosing_scope_info in self.scope_stack.iter().rev() {
             let enclosing_scope_id = enclosing_scope_info.file_scope_id;
+            let is_immediately_enclosing_scope = popped_scope.parent() == Some(enclosing_scope_id);
             let enclosing_scope_kind = self.scopes[enclosing_scope_id].kind();
             let enclosing_place_table = &self.place_tables[enclosing_scope_id];
 
@@ -352,14 +354,12 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
                     nested_scope: popped_scope_id,
                     nested_laziness: ScopeLaziness::Eager,
                 };
-                let is_immediately_enclosing_scope = popped_scope.kind().is_annotation()
-                    && popped_scope.parent() == Some(enclosing_scope_id);
                 let eager_snapshot = self.use_def_maps[enclosing_scope_id]
                     .snapshot_enclosing_state(
                         enclosing_place_id,
                         enclosing_scope_kind,
                         enclosing_place,
-                        is_immediately_enclosing_scope,
+                        popped_scope_is_annotation_scope && is_immediately_enclosing_scope,
                     );
                 self.enclosing_snapshots.insert(key, eager_snapshot);
             }
--- a/crates/ty_python_semantic/src/semantic_index/use_def.rs
+++ b/crates/ty_python_semantic/src/semantic_index/use_def.rs
@@ -1188,7 +1188,7 @@ impl<'db> UseDefMapBuilder<'db> {
         enclosing_place: ScopedPlaceId,
         enclosing_scope: ScopeKind,
         enclosing_place_expr: PlaceExprRef,
-        is_immediately_enclosing_scope: bool,
+        is_parent_of_annotation_scope: bool,
     ) -> ScopedEnclosingSnapshotId {
         let bindings = match enclosing_place {
             ScopedPlaceId::Symbol(symbol) => self.symbol_states[symbol].bindings(),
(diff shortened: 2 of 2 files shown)


mtshiba:
`ty_server::e2e pull_diagnostics::document_diagnostic_caching_changed` seems to hang non-deterministically.

https://github.com/astral-sh/ruff/actions/runs/19173434027/job/54811898939?pr=21317

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -8022,6 +8022,9 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                     FileScopeId::global(),
                                     ConstraintKey::NarrowingConstraint(constraint),
                                 ));
+                                // Reaching here means that no bindings are found in any scope.
+                                // Since `explicit_global_symbol` may return a cycle initial value, we return `Place::Undefined` here.
+                                return Place::Undefined.into();
```
carljm, line 8494: If returning here is correct, then pushing the constraint into `constraint_keys` immediately above this cannot have any effect, so should we remove that as well? (Removing it doesn't seem to fail any test.)

Why do we hit the `FoundConstraint` case in this situation in the first place? It's not clear to me where the "constraint" comes from.

mtshiba: Since place includes member as well as symbol, narrowing constraints must still be tracked even if there are no bindings, for example:

```python
from typing import Literal

class Foo:
    bar: Literal[1, 2]

foo = Foo()
if foo.bar == 1:
    # `foo.bar` has no explicit binding, but the constraint is `foo.bar == 1`
    [reveal_type(foo.bar) for _ in ()]  # Literal[1]
```

But as you pointed out, there was no need to add the constraint to `constraint_keys` here, because we already added it when iterating over `ancestor_scopes` above. Fixed.

mtshiba:
I found another example of a too-many-cycle-iteration panic.

```python
if C:
    class C[_T](C): ...
```

The panic occurred because `C` in the base class list was being resolved circularly.
It should evaluate to `Undefined` rather than throwing a cyclic-class-definition error.

mtshiba:
I don't have time right now (I originally found this bug while working on #20566), but I think the logic in `infer_place_load` is clearly too complicated and could be a breeding ground for bugs like this one.
I think it should be refactored at some point.

crates/ty_python_semantic/src/types/infer/builder.rs
mtshiba: A more fundamental issue is that the global scope should be excluded from the scopes of enclosing scope lookup.
See the comments below.

crates/ty_python_semantic/resources/corpus/cyclic_symbol_in_comprehension.py
```diff
@@ -1,13 +1,7 @@
-# Documentation of two fuzzer panics involving comprehensions
+# Regression test for https://github.com/astral-sh/ruff/pull/20962
```
MichaReiser, line 1: Can you say more about why you converted this file to a corpus test? It's not clear to me why that would be preferred or required.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -7866,6 +7866,12 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             let mut nonlocal_union_builder = UnionBuilder::new(db);
             let mut found_some_definition = false;
             for (enclosing_scope_file_id, _) in self.index.ancestor_scopes(file_scope_id).skip(1) {
+                // If the current enclosing scope is global, no place lookup is performed here,
+                // instead falling back to the module's explicit global lookup below.
+                if enclosing_scope_file_id == FileScopeId::global() {
```
MichaReiser: Nit
```suggestion
                if enclosing_scope_file_id.is_global() {
```

crates/ty_python_semantic/resources/corpus/cyclic_symbol_in_comprehension.py
AlexWaygood: The previous test asserted that we panicked; the new test asserts that the snippet does not cause us to panic

AlexWaygood: I guess it could have stayed as an mdtest, but corpus tests are what we usually use when all we care about is checking that we didn't panic; it makes sense to me

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: > Fixed.

It looks like the push to `constraint_keys` is still there?

mtshiba: Pushing to `constraint_keys` is now necessary because iteration over `ancestor_scopes` now skips the global scope lookup.
For example in the following case:

```python
class C:
    x: int | None = None

c = C()

if c.x is not None:
    class _:
        reveal_type(c.x)  # revealed: int
```

carljm:
The new diagnostic on the conformance suite looks incorrect. We are now throwing an error on the reference to `Private` at https://github.com/python/typing/blob/main/conformance/tests/generics_syntax_scoping.py#L74 but we should not.

carljm:
This is looking great, thanks! Just as a heads-up: I am making some (minor: code organization and naming) changes to this PR and then I will merge it.

PR #21319, opened 2025-11-07
Reduce notebook memory footprint

MichaReiser:
## Summary

Reduce the memory footprint of `NotebookIndex` from `O(lines)` to `O(cells)`. 

The `NotebookIndex` stores a mapping from rows in the concatenated notebook document
to the relative row numbers within each cell and from absolute line number to the corresponding cell.

The way this is implemented today is by having a vector with one entry for every absolute row number
where the value is the index of the cell. While this allows `O(1)` lookup, it does require an entry for every single line.

This PR rewrites our representation (and uses one `Vec` instead of two) to only store the 
start line number per cell and use a binary search to find to which cell a line number belongs. 
This makes lookups slightly slower (from `O(1)` to `O(log(n))` but reduces memory usage
from `O(rows)` to `O(cells)`.

I noticed this improvement when working on ty's notebook support but decided to extract it into its own PR to make reviewing easier.

## Test Plan

Existing tests

### DIFF
--- a/crates/ruff_notebook/src/index.rs
+++ b/crates/ruff_notebook/src/index.rs
@@ -8,37 +8,40 @@ use ruff_source_file::{LineColumn, OneIndexed, SourceLocation};
 /// [`ruff_text_size::TextSize`] to jupyter notebook cell/row/column.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct NotebookIndex {
-    /// Enter a row (1-based), get back the cell (1-based)
-    pub(super) row_to_cell: Vec<OneIndexed>,
-    /// Enter a row (1-based), get back the row in cell (1-based)
-    pub(super) row_to_row_in_cell: Vec<OneIndexed>,
+    /// Stores the starting row and the absolute cell index for every Python (valid) cell.
+    ///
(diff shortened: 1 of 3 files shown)


crates/ruff_notebook/src/index.rs
```diff
@@ -8,37 +8,40 @@ use ruff_source_file::{LineColumn, OneIndexed, SourceLocation};
...
 
-    /// Returns an iterator over the row:cell-number pairs (both 1-based).
-    pub fn iter(&self) -> impl Iterator<Item = (OneIndexed, OneIndexed)> {
-        self.row_to_cell
-            .iter()
-            .enumerate()
-            .map(|(row, cell)| (OneIndexed::from_zero_indexed(row), *cell))
+    /// Returns an iterator over the starting rows of each cell (1-based).
+    ///
+    /// This yields one entry per Python cell (skipping over Makrdown cell).
```
dhruvmanila, line 42:
```suggestion
    /// This yields one entry per Python cell (skipping over Markdown cell).
```

MichaReiser: Thanks. I pushed the fix to https://github.com/astral-sh/ruff/pull/21175

PR #21320, opened 2025-11-07
[ty] Promote literals in invariant return position

ibraheemdev:
## Summary

We currently unconditionally promote literals in generic class constructors. We should do this for any literal in an invariant position in the return type. Resolves https://github.com/astral-sh/ty/issues/1357.

This PR is stacked on https://github.com/astral-sh/ruff/pull/20933.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -368,7 +368,7 @@ class Z[T]:
 
 z1: Z[Any] = Z(1)
 # TODO: This should reveal `Z[Any]`.
-reveal_type(z1)  # revealed: Z[int]
+reveal_type(z1)  # revealed: Z[Literal[1]]
 ```
 
 ## PEP-604 annotations are supported
@@ -669,7 +669,7 @@ x4 = invariant(1)
 reveal_type(x1)  # revealed: Bivariant[Literal[1]]
 reveal_type(x2)  # revealed: Covariant[Literal[1]]
 reveal_type(x3)  # revealed: Contravariant[Literal[1]]
-reveal_type(x4)  # revealed: Invariant[Literal[1]]
+reveal_type(x4)  # revealed: Invariant[int]
 
 x5: Bivariant[Any] = bivariant(1)
 x6: Covariant[Any] = covariant(1)


AlexWaygood:
The primer report looks excellent but it looks like there's a new type-assertion failure in the conformance test suite -- have you checked out if that's expected?

crates/ty_python_semantic/resources/mdtest/generics/pep695/classes.md
```diff
@@ -315,27 +315,31 @@ If either method comes from a generic base class, we don't currently use its inf
 to specialize the class.
 
 ```py
+from typing import Literal
+
```
AlexWaygood: this doesn't seem to be used?

```suggestion
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/classes.md
```diff
@@ -315,27 +315,31 @@ If either method comes from a generic base class, we don't currently use its inf
...
 
-reveal_type(D(1))  # revealed: D[int]
+reveal_type(D(1))  # revealed: D[Literal[1]]
 ```
 
 ### Generic class inherits `__init__` from generic base class
 
 ```py
+from typing import Literal
+
```
AlexWaygood: same here

```suggestion
```

crates/ty_python_semantic/resources/mdtest/type_compendium/tuple.md
```diff
@@ -514,10 +514,8 @@ For covariant types, such as `frozenset`, the ideal behaviour would be to not pr
 types to their instance supertypes: doing so causes more false positives than it fixes:
 
 ```py
-# TODO: better here would be `frozenset[Literal[1, 2, 3]]`
-reveal_type(frozenset((1, 2, 3)))  # revealed: frozenset[int]
-# TODO: better here would be `frozenset[tuple[Literal[1], Literal[2], Literal[3]]]`
-reveal_type(frozenset(((1, 2, 3),)))  # revealed: frozenset[tuple[int, int, int]]
+reveal_type(frozenset((1, 2, 3)))  # revealed: frozenset[Literal[1, 2, 3]]
+reveal_type(frozenset(((1, 2, 3),)))  # revealed: frozenset[tuple[Literal[1], Literal[2], Literal[3]]]
```
AlexWaygood, line 518: wonderful!

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2762,6 +2763,51 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
 
+        let tcx_specialization = self
+            .call_expression_tcx
+            .annotation
+            .and_then(|annotation| annotation.class_specialization(self.db));
+
+        let promote_literals = |typevar: GenericContextTypeVar<'db>, ty: Type<'db>| -> Type<'db> {
+            let bound_typevar = typevar.bound_typevar();
+
+            if typevar.is_inherited() && bound_typevar.variance(self.db).is_invariant() {
```
AlexWaygood, line 2774: What does `is_inherited` mean here? Why does it matter whether a typevar is inherited or not?

(It would be great if `is_inherited` could have a doc-comment explaining what it does 😃)

AlexWaygood: Ah, I think I see -- you're using the term to mean that the typevar is associated with a class's generic context rather than the generic context of a function or method? So it doesn't have any thing to do with the generic context of a class's base classes etc.?

sharkdp:
> The primer report looks excellent but it looks like there's a new type-assertion failure in the conformance test suite -- have you checked out if that's expected?

Note that both the primer and the conformance checks here include changes from the previous PRs on which this is stacked. We compare against main.

crates/ty_python_semantic/src/types/call/bind.rs
ibraheemdev: That's right, [though I didn't invent that terminology](https://github.com/astral-sh/ruff/blob/47cdf494d5b6317a5d111ba1fdf3d3b820619f22/crates/ty_python_semantic/src/types/class.rs#L1476).

Ideally we wouldn't need to special case inherited type variables, but we don't currently have a robust way to get access to the bound type of a constructor method (the [`synthesized_constructor_return_ty`](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types.rs#L1115C19-L1115C52)).

ibraheemdev:
Hmm this is more tricky than I realized. We can only perform the literal promotion of a given type variable if it does not lead to argument assignability errors. We may need to resort to a simple heuristic of "promote every type variable in invariant position", and fallback to not performing any promotion if type-checking fails, as I suspect attempting partial promotion would be too expensive.

ibraheemdev:
Superseded by https://github.com/astral-sh/ruff/pull/21439.

PR #21322, opened 2025-11-07
[ty] Respect notebook cell boundaries when adding an auto import

MichaReiser:
## Summary

The text edits for auto-completion are constrained to within the same document (because there's only a range but no URI field). Given that the LSP represents each cell as a separate document, this means that auto import edits must be limited to the same cell (we can't change an import from another cell or add an import to the first cell). 

This PR does just that.

## Test plan

Added e2e tests

### DIFF
--- a/crates/ty_server/tests/e2e/snapshots/e2e__notebook__auto_import.snap
+++ b/crates/ty_server/tests/e2e/snapshots/e2e__notebook__auto_import.snap
@@ -6,7 +6,7 @@ expression: completions
   {
     "label": "Literal (import typing)",
     "kind": 6,
-    "sortText": " 44",
+    "sortText": " 43",
     "insertText": "Literal",
     "additionalTextEdits": [
       {
@@ -27,7 +27,7 @@ expression: completions
   {
     "label": "Literal (import typing_extensions)",
     "kind": 6,
-    "sortText": " 45",
+    "sortText": " 44",
     "insertText": "Literal",
     "additionalTextEdits": [
       {
@@ -48,7 +48,7 @@ expression: completions
   {
     "label": "LiteralString (import typing)",
     "kind": 6,
-    "sortText": " 46",
+    "sortText": " 45",
     "insertText": "LiteralString",
     "additionalTextEdits": [
       {
@@ -69,7 +69,7 @@ expression: completions
   {
     "label": "LiteralString (import typing_extensions)",
     "kind": 6,
-    "sortText": " 47",
+    "sortText": " 46",
     "insertText": "LiteralString",
     "additionalTextEdits": [
       {
--- a/crates/ty_server/tests/e2e/snapshots/e2e__notebook__auto_import_docstring.snap
+++ b/crates/ty_server/tests/e2e/snapshots/e2e__notebook__auto_import_docstring.snap
@@ -6,7 +6,7 @@ expression: completions
   {
     "label": "Literal (import typing)",
     "kind": 6,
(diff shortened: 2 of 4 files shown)


.config/nextest.toml
```diff
@@ -7,6 +7,10 @@ serial = { max-threads = 1 }
 filter = 'binary(file_watching)'
 test-group = 'serial'
 
+[[profile.default.overrides]]
+filter = 'binary(e2e)'
+test-group = 'serial'
+
```
Gankra, line 13: Was this intended to be checked in?

crates/ruff_python_importer/src/insertion.rs
```diff
@@ -49,7 +49,21 @@ impl<'a> Insertion<'a> {
     ///
     /// The insertion returned will begin at the start of the `import os` statement, and will
     /// include a trailing newline.
```
BurntSushi, line 51: Could you add something to the docs here saying something about `within_range` and what it's for?

BurntSushi (approved):
Makes sense to me!

.config/nextest.toml
MichaReiser: Sort of. My hope was that it makes the lsp `e2e` tests less flaky

crates/ty_ide/src/importer.rs
```diff
@@ -145,10 +147,15 @@ impl<'a> Importer<'a> {
...
         let Some(response) = self.find(&request, members.at) else {
-            let insertion = if let Some(future) = self.find_last_future_import() {
+            let insertion = if let Some(future) = self.find_last_future_import(members.at) {
                 Insertion::end_of_statement(future.stmt, self.source, self.stylist)
             } else {
-                Insertion::start_of_file(self.parsed.suite(), self.source, self.stylist)
+                // What we want here: Given an offset, return the range of that cell. I guess that's what containing range is
+                let range = source_text(self.db, self.file)
+                    .as_notebook()
+                    .and_then(|notebook| notebook.cell_offsets().containing_range(members.at));
```
dhruvmanila: Yes, this comment is correct. The `containing_range` will return the cell range that contains this offset.

crates/ruff_notebook/src/cell.rs
```diff
@@ -294,19 +294,17 @@ impl CellOffsets {
     }
 
     /// Returns `true` if the given range contains a cell boundary.
+    ///
+    /// A range starting at the cell boundary isn't considered to contain a cell boundary
+    /// as it starts right after it. A range starting before a cell boundary
+    /// and ending exactly at the boundary is considered to contain a cell boundary.
     pub fn has_cell_boundary(&self, range: TextRange) -> bool {
```
dhruvmanila, line 317: I'm not exactly sure I follow this - are you saying that if the cell boundary is at `range.start`, then this method should return `false`?

MichaReiser: Yes. Otherwise, testing any range from the start of the cell always returns `true`. 

Cell 1:

```py
import c
```

Cell 2:

```py
import os
```

When checking the range between `import c`..`import os`, then `start` is before the cell boundary but it ends at the cell boundary. That means, this range contains the cell boundary.

When checking the range `import os`, then this shouldn't contain the cell boundary because it starts at the cell boundary, but it never crosses the boundary

MichaReiser:
I marked this as internal because notebook support hasn't shipped yet. So this isn't a user visible change

PR #21327, opened 2025-11-07
Enable perf profile upload for walltime benchmarks

MichaReiser:
Enable the walltime performance profile upload for codspeed walltime benchmarks. 

This does increase the walltime benchmark runtime by ~2min (from ~8 to ~10 and ~9 to ~11min). I do find the profile's useful when investigating a regression (to get a first impression of where it might be coming from).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1033,11 +1033,6 @@ jobs:
 
       - name: "Run benchmarks"
         uses: CodSpeedHQ/action@bb005fe1c1eea036d3894f02c049cb6b154a1c27 # v4.3.3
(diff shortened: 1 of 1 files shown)


MichaReiser:
I used my watch to stop the time, it's 1min and 20s. I agree, that's somewhat significant. 

> I've found they've given me any useful information regarding a ty performance regression -- and all our walltime benchmarks are ty benchmarks. A 2-minute increase in CI times is a pretty significant increase, meanwhile.

I agree that they're less useful but they often give me at least an idea where a regression comes from when reviewing another PR without having to clone it and do some local profiling.

PR #21330, opened 2025-11-07
[`ruff`] Ignore `str()` when not used for simple conversion (`RUF065`)

### LINKED ISSUE #21315: RUF065 should ignore `str` when not used for conversion
### Summary

[`logging-eager-conversion` (RUF065)](https://docs.astral.sh/ruff/rules/logging-eager-conversion/) should only treat `str` as an unnecessary conversion when it has exactly one argument and that argument isn’t starred. Otherwise, the `str` call is doing something else and can’t be so straightforwardly removed. [Example](https://play.ruff.rs/d4bd231e-6994-449a-9639-b344f7a2ed78):
```console
$ cat >ruf065.py <<'# EOF'
import logging
logging.warning("%s", str())
logging.warning("%s", str(b"\xe2\x9a\xa0", "utf-8"))
logging.warning("%s", str(*(b"\xf0\x9f\x9a\xa7", "utf-8")))
logging.warning("%s", str(**{"object": b"\xf0\x9f\x9a\xa8", "encoding": "utf-8"}))
# EOF

$ ruff --isolated check ruf065.py --select RUF065 --preview --output-format concise -q
ruf065.py:2:23: RUF065 Unnecessary `str()` conversion when formatting with `%s`
ruf065.py:3:23: RUF065 Unnecessary `str()` conversion when formatting with `%s`
ruf065.py:4:23: RUF065 Unnecessary `str()` conversion when formatting with `%s`
ruf065.py:5:23: RUF065 Unnecessary `str()` conversion when formatting with `%s`
```

### Version

danparizher:
## Summary

Fixed RUF065 (`logging-eager-conversion`) to only flag `str()` calls when they perform a simple conversion that can be safely removed. The rule now ignores `str()` calls with no arguments, multiple arguments, starred arguments, or keyword unpacking, preventing false positives.

Fixes #21315

## Problem Analysis

The RUF065 rule was incorrectly flagging all `str()` calls in logging statements, even when `str()` was performing actual conversion work beyond simple type coercion. Specifically, the rule flagged:

- `str()` with no arguments - which returns an empty string
- `str(b"data", "utf-8")` with multiple arguments - which performs encoding conversion
- `str(*args)` with starred arguments - which unpacks arguments
- `str(**kwargs)` with keyword unpacking - which passes keyword arguments

These cases cannot be safely removed because `str()` is doing meaningful work (encoding conversion, argument unpacking, etc.), not just redundant type conversion.

The root cause was that the rule only checked if the function was `str()` without validating the call signature. It didn't distinguish between simple `str(value)` conversions (which can be removed) and more complex `str()` calls that perform actual work.

## Approach

The fix adds validation to the `str()` detection logic in `logging_eager_conversion.rs`:

1. **Check argument count**: Only flag `str()` calls with exactly one positional argument (`str_call_args.args.len() == 1`)
2. **Check for starred arguments**: Ensure the single argument is not starred (`!str_call_args.args[0].is_starred_expr()`)
3. **Check for keyword arguments**: Ensure there are no keyword arguments (`str_call_args.keywords.is_empty()`)

This ensures the rule only flags cases like `str(value)` where `str()` is truly redundant and can be removed, while ignoring cases where `str()` performs actual conversion work.

The fix maintains backward compatibility - all existing valid test cases continue to be flagged correctly, while the new edge cases are properly ignored.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/ruff/RUF065_0.py
+++ b/crates/ruff_linter/resources/test/fixtures/ruff/RUF065_0.py
@@ -68,3 +68,4 @@ log(logging.INFO, "Octal: %s", oct(255))
 
 info("Hex: %s", hex(42))
 log(logging.INFO, "Hex: %s", hex(255))
+


crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -138,16 +138,25 @@ pub(crate) fn logging_eager_conversion(checker: &Checker, call: &ast::ExprCall)
...
             // Check for various eager conversion patterns
             match format_conversion {
                 // %s with str() - remove str() call
+                // Only flag if str() has exactly one non-starred positional argument and no keyword arguments
                 FormatConversion::Str
-                    if checker.semantic().match_builtin_expr(func.as_ref(), "str") =>
+                    if checker.semantic().match_builtin_expr(func.as_ref(), "str")
+                        && str_call_args.args.len() == 1
+                        && !str_call_args.args[0].is_starred_expr()
+                        && str_call_args.keywords.is_empty() =>
```
dscorbett: `logging.warning("%s", str(object="!"))`, with one keyword argument, should still be flagged.

danparizher:
I added a test case for `str(object="!")`, and the new logic seems to work in isolation, but for some reason isn't being detected in the test file. I would appreciate it if someone could help me diagnose the issue!

crates/ruff_linter/src/rules/ruff/rules/logging_eager_conversion.rs
```diff
@@ -138,16 +138,28 @@ pub(crate) fn logging_eager_conversion(checker: &Checker, call: &ast::ExprCall)
...
+                // Only flag if str() has exactly one argument (positional or keyword) that is not unpacked
                 FormatConversion::Str
-                    if checker.semantic().match_builtin_expr(func.as_ref(), "str") =>
+                    if checker.semantic().match_builtin_expr(func.as_ref(), "str")
+                        && ((str_call_args.args.len() == 1
+                            && !str_call_args.args[0].is_starred_expr()
+                            && str_call_args.keywords.is_empty())
+                            || (str_call_args.args.is_empty()
+                                && str_call_args.keywords.len() == 1
+                                && str_call_args.keywords[0].arg.is_some())) =>
```
ntBre: This looks pretty complicated. Can we just use `find_argument` and a check that `str_call_args.len() == 1`?


```suggestion
                    if checker.semantic().match_builtin_expr(func.as_ref(), "str")
                        && str_call_args.len() == 1
                        && str_call_args.find_argument("object", 0).is_some_and(|arg| !arg.is_variadic()) => 
```

with 

```rust
impl<'a> ArgOrKeyword<'a> {
    pub const fn is_variadic(self) -> bool {
        match self {
            ArgOrKeyword::Arg(expr) => expr.is_starred_expr(),
            ArgOrKeyword::Keyword(keyword) => keyword.arg.is_none(),
        }
    }
}
````

This comes up a lot, so that might be a nice helper to have around.

Hopefully this will help with the `object` case, but I'm not sure.

ntBre (changes requested):
Thanks! I had a suggestion for simplification, which will hopefully help with the test case too.

danparizher:
Thanks! I pushed those changes, but it looks like we're running into the same issue as before. The fix works in isolation but not within the test.

ntBre:
Oh, `str` is shadowed on line 37 of the test file, so none of the new tests are actually exercising the new logic. I think we should add the new tests to a separate file.

PR #21335, opened 2025-11-08
[ty] Add support for properties that return `Self`

sharkdp:
## Summary

Detect usages of implicit `self` in property getters, which allows us to treat their signature as being generic.

closes https://github.com/astral-sh/ty/issues/1502

## Typing conformance

Two new type assertions that are succeeding.

## Ecosystem results

Mostly look good. There are a few new false positives related to a bug with constrained typevars that is unrelated to the work here. I reported this as https://github.com/astral-sh/ty/issues/1503.

## Test Plan

Added regression tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/self.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/self.md
@@ -139,7 +139,7 @@ The first parameter of instance methods always has type `Self`, if it is not exp
 The name `self` is not special in any way.
 
 ```py
-def some_decorator(f: Callable) -> Callable:
+def some_decorator[**P, R](f: Callable[P, R]) -> Callable[P, R]:
     return f
 
 class B:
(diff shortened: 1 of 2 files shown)


AlexWaygood (approved):
How extensible will this approach be for other decorated methods, like e.g. methods decorated with `functools.lru_cache`? (Mind you, I don't know why a method decorated with `@lru_cache` would ever be annotated as returning `Self`, but you get the idea.)

Even if this approach won't work in other cases, though, properties are common enough that I think even a narrow fix is worthwhile for now!

sharkdp:
> How extensible will this approach be for other decorated methods, like e.g. methods decorated with `functools.lru_cache`? (Mind you, I don't know why a method decorated with `@lru_cache` would ever be annotated as returning `Self`, but you get the idea.)

I thought about this, but there was no way to test this yet. I should have commented on that, though. I have now pushed up a fix that is not specific to properties and *should* work for other decorators as well. The only way I can decorate a method right now without being restricted by our generics solver is to use `def identity[T](t: T) -> T`, but that worked with the previous fix as well, as it keeps the FunctionType intact. But I modified an existing test and added a TODO.

PR #21337, opened 2025-11-08
[ruff+ty] Add colour to `--help`

hugovk:
## Summary

The output of `uv --help` uses green and cyan to improve readability. But ruff and ty do not:

image

Let's add similar styling to ruff and ty. This is the main bit in uv:

https://github.com/astral-sh/uv/blob/85c5d32284522eb958d9cfc10ff62f0956eab113/crates/uv-cli/src/lib.rs#L88-L93

## Test Plan

Compare and check they have similarly styled output:

```sh
./target/debug/ruff --help
./target/debug/ty --help
uv --help
```

image

Also run with `NO_COLOR=1` to confirm there's no colour or bold text:

```sh
NO_COLOR=1 ./target/debug/ruff --help
NO_COLOR=1 ./target/debug/ty --help
NO_COLOR=1 uv --help
```

image

### DIFF
--- a/crates/ty/src/args.rs
+++ b/crates/ty/src/args.rs
@@ -1,5 +1,7 @@
 use crate::logging::Verbosity;
 use crate::python_version::PythonVersion;
+use clap::builder::Styles;
+use clap::builder::styling::{AnsiColor, Effects};
 use clap::error::ErrorKind;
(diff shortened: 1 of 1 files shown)


MichaReiser:
Thank you, this is great

Issue #21119, opened 2025-10-29
Update stdlibs with 3.15


amyreese:
Not sure when we normally do this, but at some point we'll need to update the known stdlib list, and I just pushed a `stdlibs` release earlier today with a list of modules from 3.15.0a1.

MichaReiser:
We do have a versioned list of all builtins in ruff (see [builtins.rs](https://github.com/astral-sh/ruff/blob/c4506f0e94882fa1118eb2a13339f9be7bf0bcaa/crates/ruff_python_stdlib/src/builtins.rs#L14)). I'm not aware of any place where we check for builtin modules, but I might be mistaken. Do you have a reference where Ruff checks for built-in modules?

amyreese:
It's part of isort, so that it knows which modules are first-party/stdlib for sorting into categories.

MichaReiser:
Oh this one https://github.com/astral-sh/ruff/blob/bc7b30364dcd196f9d660218b852870b1d978e76/crates/ruff_python_stdlib/src/sys/known_stdlib.rs#L1-L0

I think we can update it anytime, as the imports are by Python version. We might just need it to do multiple times. The first step is to add Python 3.15 to the supported python versions

ntBre:
I believe this was resolved by https://github.com/astral-sh/ruff/pull/25103 and https://github.com/astral-sh/ruff/pull/25273.

PR #21338, opened 2025-11-08
Update known stdlibs with python `3.15`

ShaharNaveh:
Follow up on https://github.com/astral-sh/ruff/issues/21119#issuecomment-3462875240


scripts/generate_known_standard_library.py
```diff
@@ -4,7 +4,8 @@
 
 from stdlibs import stdlib_module_names
 
-PATH = Path("crates") / "ruff_python_stdlib" / "src" / "sys" / "known_stdlib.rs"
+CRATES_PATH = Path(__file__).parents[1] / "crates"
+PATH = CRATES_PATH / "ruff_python_stdlib" / "src" / "sys" / "known_stdlib.rs"
```
ShaharNaveh, line 8: I got an error when running from the `scripts` directory.
I can revert this if it's an issue.

MichaReiser:
Thank you.

I'm not sure if it makes sense to update the stdlib if Ruff doesn't have support for Python 3.15. I think I'd prefer to wait for when we add Python 3.15 support once the beta is closer, to ensure that we don't ship with a stale stdlib for 3.15

ShaharNaveh:
> Thank you.
> 
> I'm not sure if it makes sense to update the stdlib if Ruff doesn't have support for Python 3.15. I think I'd prefer to wait for when we add Python 3.15 support once the beta is closer, to ensure that we don't ship with a stale stdlib for 3.15

Sure, feel free to close this PR if you think it's just noise in  the PR pool ATM.

---

What is the approach when supporting a new Python version? When will users can set `python-version = "PY315"` for example? (even if it's still considered experimental)

Does ruff have a contributer's guide on when  to implement support for new Python  versions?

MichaReiser:
> What is the approach when supporting a new Python version? When will users can set python-version = "PY315" for example? (even if it's still considered experimental)

In the past, we waited for the new Python version to be closer to beta to avoid implementing features that don't end up being stabilized (or have to change the semantics multiple times). 

> Does ruff have a contributer's guide on when to implement support for new Python versions?

We don't have a contributor guide for adding support for a new Python version but you can search for the [`python314`](https://github.com/astral-sh/ruff/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aclosed+label%3Apython314) label to see the relevant PRs for adding Python 3.14. That should give you a good idea on how we go about it.

PR #21344, opened 2025-11-08
Allow making a naive date with datetime.strptime()

### LINKED ISSUE #13359: DTZ007 - arguably a false positive
A not uncommon pattern to parse a date string to a Python date uses similar code to the below. Times/time-zones shouldn't matter.

```
Ruff: DTZ007 Naive datetime constructed using `datetime.datetime.strptime()` without %z
  |
1 |         # Parse the date parameter
2 |         try:
3 |             input_date = datetime.strptime(date, "%Y-%m-%d").date()
  |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ DTZ007
4 |         except ValueError:
5 |             print("Invalid date format. Use YYYY-MM-DD.")
  |
    = help: Call `.replace(tzinfo=<timezone>)` or `.astimezone()` to convert to an aware datetime
```

axxeny:
## Summary

Resolves #13359 if decision is made to approve the rule change.

## Test Plan

Untested.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_datetimez/DTZ007.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_datetimez/DTZ007.py
@@ -29,6 +29,9 @@ datetime.datetime.strptime("something", something).astimezone()
 # OK
 datetime.datetime.strptime("something", something).replace(tzinfo=datetime.timezone.utc)
 
+# OK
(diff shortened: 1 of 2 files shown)


MichaReiser:
I'll close this PR as we can track the decision on #13359 and it's easy to reopen the PR if we decide to add the rule. 

Thanks for taking the time to work on this rule.

PR #21348, opened 2025-11-09
[`refurb`] Detect empty f-strings (`FURB105`)

### LINKED ISSUE #21346: FURB105 should detect empty f-strings
### Summary

It would be helpful for [`print-empty-string` (FURB105)](https://docs.astral.sh/ruff/rules/print-empty-string/) to detect empty f-strings via [`is_empty_f_string`](https://github.com/astral-sh/ruff/blob/0.14.4/crates/ruff_python_ast/src/helpers.rs#L1414). [Example](https://play.ruff.rs/210c3a04-8fb6-43d3-a39c-4caed9335c1d):
```console
$ cat >furb105.py <<'# EOF'
print(f"")
# EOF

$ ruff --isolated check furb105.py --select FURB105
All checks passed!
```

### Version

danparizher:
## Summary

Fixes FURB105 (`print-empty-string`) to detect empty f-strings in addition to regular empty strings. Previously, the rule only flagged `print("")` but missed `print(f"")`. This fix ensures both cases are detected and can be automatically fixed.

Fixes #21346

## Problem Analysis

The FURB105 rule checks for unnecessary empty strings passed to `print()` calls. The `is_empty_string` helper function was only checking for `Expr::StringLiteral` with empty values, but did not handle `Expr::FString` (f-strings). As a result, `print(f"")` was not being flagged as a violation, even though it's semantically equivalent to `print("")` and should be simplified to `print()`.

The issue occurred because the function used a `matches!` macro that only checked for string literals:

```rust
fn is_empty_string(expr: &Expr) -> bool {
    matches!(
        expr,
        Expr::StringLiteral(ast::ExprStringLiteral { value, .. }) if value.is_empty()
    )
}
```

## Approach

1. **Import the helper function**: Added `is_empty_f_string` to the imports from `ruff_python_ast::helpers`, which already provides logic to detect empty f-strings.

2. **Update `is_empty_string` function**: Changed the implementation from a `matches!` macro to a `match` expression that handles both string literals and f-strings:

   ```rust
   fn is_empty_string(expr: &Expr) -> bool {
       match expr {
           Expr::StringLiteral(ast::ExprStringLiteral { value, .. }) => value.is_empty(),
           Expr::FString(f_string) => is_empty_f_string(f_string),
           _ => false,
       }
   }
   ```

The fix leverages the existing `is_empty_f_string` helper function which properly handles the complexity of f-strings, including nested f-strings and interpolated expressions. This ensures the detection is accurate and consistent with how empty strings are detected elsewhere in the codebase.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/refurb/FURB105.py
+++ b/crates/ruff_linter/resources/test/fixtures/refurb/FURB105.py
@@ -19,6 +19,9 @@ print("", *args, sep="")
 print("", **kwargs)
 print(sep="\t")
 print(sep=print(1))
+print(f"")
+print(f"", sep=",")
+print(f"", end="bar")
 
 # OK.
 
@@ -33,3 +36,4 @@ print("foo", "", sep=",")
 print("foo", "", "bar", "", sep=",")
 print("", "", **kwargs)
 print(*args, sep=",")
+print(f"foo")
--- a/crates/ruff_linter/src/rules/refurb/rules/print_empty_string.rs
+++ b/crates/ruff_linter/src/rules/refurb/rules/print_empty_string.rs
@@ -1,5 +1,5 @@
 use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::helpers::contains_effect;
+use ruff_python_ast::helpers::{contains_effect, is_empty_f_string};
 use ruff_python_ast::{self as ast, Expr};
 use ruff_python_codegen::Generator;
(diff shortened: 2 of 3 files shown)


ntBre (approved):
Thank you!

I'm just going to close and reopen to try to trigger the ecosystem comment.

ntBre:
Still no comment, but I downloaded the artifact:

> ### Linter (stable)
> ✅ ecosystem check detected no linter changes.
> 
> ### Linter (preview)
> ✅ ecosystem check detected no linter changes.

PR #21361, opened 2025-11-10
Rebuild ruff binary instead of sharing it across jobs

MichaReiser:
## Summary

Similar to ty's ecosystem check, build ruff from source during the ecosytem check intsead of downloading the binary from the `test` job.

Building ruff from source is relatively fast and waiting for the test job before starting the ecosystem check also only delays it unnecessarily and has the downside that the job fails if `main` hasn't uploaded the `ruff` binary yet (because the test job is still running or failed).

This also simplifies the CI pipeline.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -531,7 +531,9 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
+          ref: ${{ github.event.pull_request.base.ref }}
           persist-credentials: false
+
       - uses: astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41 # v7.1.2
         with:
           python-version: ${{ env.PYTHON_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -474,26 +465,23 @@ jobs:
...
-          chmod +x "${DOWNLOAD_PATH}/ruff"
-
           (
             uv run \
             --python="${PYTHON_VERSION}" \
             --project=./python/py-fuzzer \
             --locked \
             fuzz \
-            --test-executable="${DOWNLOAD_PATH}/ruff" \
+            --test-executable=target/debug/ruff \
```
AlexWaygood, line 484: if you run the script without the `--test-executable` flag set, it defaults to building a binary itself using `cargo build --profile=profiling`. So you could possibly simplify this job even more by omitting this flag, and not even building Ruff as a separate step.

.github/workflows/ci.yaml
```diff
@@ -535,53 +523,57 @@ jobs:
...
-      - determine_changes
+    needs: determine_changes
     # Only runs on pull requests, since that is the only we way we can find the base version for comparison.
     # Ecosystem check needs linter and/or formatter changes.
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-test') && github.event_name == 'pull_request' && needs.determine_changes.outputs.code == 'true' }}
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
+          ref: ${{ github.event.pull_request.base.ref }}
```
AlexWaygood, line 534: Is this checkout for the baseline version? For mypy_primer, the typing conformance suite diff, and the ty fuzzer we checkout the baseline version using this logic:

https://github.com/astral-sh/ruff/blob/ada3208de50d3dce3aacb393ea23b35416bed593/.github/workflows/ci.yaml#L660-L661

Not sure if there's any difference between the two, but I have a lot of confidence in the logic we're using elsewhere because we've been using it for a while and I haven't observed any problems with it 😄

MichaReiser: Possibly, The idea was to migrate the workflow while keeping the same semantics. However, using `MERGE_BASE` would fix the issue where someone merges changes to main when your PR isn't updated yet

MichaReiser: I'll leave this for now

.github/workflows/ci.yaml
```diff
@@ -474,28 +465,20 @@ jobs:
...
           persist-credentials: false
       - uses: astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41 # v7.1.2
-      - uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
-        name: Download Ruff binary to test
-        id: download-cached-binary
+      - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
         with:
-          name: ruff
-          path: ruff-to-test
+          shared-key: ruff-linux-debug
```
AlexWaygood, line 470: now that the script is doing the build itself (and using `--profile=profiling` internally), the debug build won't be a good cache key anymore

MichaReiser: Yeah, I just wanted to say that I revert this change as it also switches to profiling builds

AlexWaygood: sounds good

.github/workflows/ci.yaml
MichaReiser: I won't make this change because it also results in switching from debug to profiling builds. The idea of this PR is to not change behavior other than switching to freshly built ruff binaries over downloading them

AlexWaygood:
Cool that the ecosystem check now completes before some of the tests 😃

PR #21362, opened 2025-11-10
Fix syntax error false positive on alternative `match` patterns

### LINKED ISSUE #21360: False positive: `invalid-syntax` alternative patterns bind different names
### Summary

This file produces an `invalid-syntax` error, but it contains valid syntax:

```python3
import ast

match None:
    case ast.Subscript(n, ast.Constant() | ast.Slice()) | ast.Attribute(n):
        pass
```

https://play.ruff.rs/a88bae2d-d65a-4495-a33d-384a09fbbb4f

If I remove `| ast.Slice()`, the error goes away, so it seems like state is incorrectly shared between different alternative patterns.

### Version

ruff 0.14.4

ntBre:
Summary
--

Fixes #21360 by using the union of names instead of overwriting them, as Micha suggested originally on #21104.

This avoids overwriting the `n` name in the `Subscript` by the empty set of names visited in the nested OR pattern before visiting the other arm of the outer OR pattern.

Test Plan
--

A new inline test case taken from the issue


crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -1868,14 +1868,16 @@ impl<'a, Ctx: SemanticSyntaxContext> MatchPatternVisitor<'a, Ctx> {
...
+                        //     case ast.Subscript(n, ast.Constant() | ast.Slice()) | ast.Attribute(n): ...
                         SemanticSyntaxChecker::add_error(
                             self.ctx,
                             SemanticSyntaxErrorKind::DifferentMatchPatternBindings,
                             *range,
                         );
                         break;
                     }
-                    self.names = visitor.names;
+                    self.names.extend(visitor.names);
```
MichaReiser, line 1880: Can you say a bit more about the fix. My understanding (and why I recommended it) was that using union or intersection would only matter for the error recovery case. But it seems that using union over intersection is important for valid cases too or are we only fixing a symptom here?

ntBre: I think the union is correct because there can be existing patterns in `self.names` before visiting nested alternative patterns. We need to compare all of the names we visit in that arm to all of the names we visit in the next arm, not just the set of names visited in the last, innermost alternative pattern.

That's what's happening in this case because we first visit the outer class pattern, which captures `n`, then we were overwriting that `n` with the empty set when visiting the `Constant | Slice` part of the pattern. Instead we need the union of `n` and the empty set to compare to the `Attribute(n)` branch.

```py
match 42:
    case ast.Subscript(n, ast.Constant() | ast.Slice()) 
       | ast.Attribute(n): ...
```

PR #21363, opened 2025-11-10
[ty] Implicit type aliases: Add support for `typing.Union`

sharkdp:
## Summary

Add support for `typing.Union` in implicit type aliases / in value position.

## Typing conformance tests

Two new tests are passing

## Ecosystem impact

* The 2k new `invalid-key` diagnostics on pydantic are caused by https://github.com/astral-sh/ty/issues/1479#issuecomment-3513854645.
* Everything else I've checked is either a known limitation (often related to type narrowing, because union types are often narrowed down to a subset of options), or a true positive.

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -33,7 +33,7 @@ g(None)
 We also support unions in type aliases:
 
 ```py
-from typing_extensions import Any, Never, Literal, LiteralString, Tuple, Annotated, Optional
+from typing_extensions import Any, Never, Literal, LiteralString, Tuple, Annotated, Optional, Union
 from ty_extensions import Unknown
 
 IntOrStr = int | str
@@ -41,6 +41,8 @@ IntOrStrOrBytes1 = int | str | bytes
 IntOrStrOrBytes2 = (int | str) | bytes
 IntOrStrOrBytes3 = int | (str | bytes)
 IntOrStrOrBytes4 = IntOrStr | bytes
+IntOrStrOrBytes5 = int | Union[str, bytes]
+IntOrStrOrBytes6 = Union[int, str] | bytes
 BytesOrIntOrStr = bytes | IntOrStr
 IntOrNone = int | None
 NoneOrInt = None | int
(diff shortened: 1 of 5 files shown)


AlexWaygood:
The pydantic benchmark is also panicking:

```
Expected between 1 and 1000 diagnostics but got 2716
```

AlexWaygood:
@woodruffw, I see a huge diff in the mypy_primer logs [here](https://github.com/astral-sh/ruff/actions/runs/19261360735/job/55066696930?pr=21363#step:6:1063) but the comment is reporting

> No ecosystem changes detected ✅

which looks to me like a bug in astral-sh-bot? Possibly in the truncation logic? Previously we used to have it post as much of the diff as it was able to in the comment, but have the last line be something like

```
[...full diff truncated]
```

to indicate that we weren't able to post the full thing there

woodruffw:
Hmmm yeah, could be a truncation bug or maybe a mismatch in how the artifact is pulled down -- I can look at it today.

woodruffw:
> Hmmm yeah, could be a truncation bug or maybe a mismatch in how the artifact is pulled down -- I can look at it today.

Yeah, this looks like exactly it -- I'm deploying a fix to the bot momentarily 🙂 

Edit: Deploying now.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -457,13 +465,90 @@ def _(
...
+JustInt = Union[int]
+
+reveal_type(JustInt)  # revealed: <class 'int'>
+
+def _(just_int: JustInt):
+    reveal_type(just_int)  # revealed: int
+```
+
+An empty `typing.Union` leads to a `TypeError` at runtime, so we emit an error. We still infer
+`Never` when used as a type expression, which seems reasonable for an empty union:
```
carljm, line 578: 👍 I don't know why this raises an error at runtime, seems silly to me.

carljm (approved):
Nice! Personally I think we can land this even with the big perf regression on pydantic, especially if that regression still leaves us on par with pyrefly. But of course if we can optimize, that's better!

PR #21366, opened 2025-11-10
[ty] Implement constraint implication for compound types

dcreager:
This PR updates the constraint implication type relationship to work on compound types as well. (A compound type is a non-atomic type, like `list[T]`.)

The goal of constraint implication is to check whether the requirements of a constraint imply that a particular subtyping relationship holds. Before, we were only checking atomic typevars. That would let us verify that the constraint set `T ≤ bool` implies that `T` is always a subtype of `int`. (In this case, the lhs of the subtyping check, `T`, is an atomic typevar.)

But we weren't recursing into compound types, to look for nested occurrences of typevars. That means that we weren't able to see that `T ≤ bool` implies that `Covariant[T]` is always a subtype of `Covariant[int]`.

Doing this recursion means that we have to carry the constraint set along with us as we recurse into types as part of `has_relation_to`, by adding constraint implication as a new `TypeRelation` variant. (Before it was just a method on `ConstraintSet`.)

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1627,18 +1627,18 @@ impl<'db> Type<'db> {
     /// all of the restrictions in `constraints` hold.
     ///
     /// See [`TypeRelation::SubtypingAssuming`] for more details.
-    fn when_subtype_of_given(
+    fn when_subtype_of_assuming(
         self,
         db: &'db dyn Db,
         target: Type<'db>,
-        constraints: ConstraintSet<'db>,
+        assuming: ConstraintSet<'db>,
         inferable: InferableTypeVars<'_, 'db>,
     ) -> ConstraintSet<'db> {
         self.has_relation_to(
             db,
             target,
             inferable,
-            TypeRelation::SubtypingAssuming(constraints),
+            TypeRelation::SubtypingAssuming(assuming),
         )
     }
 
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -1216,7 +1216,7 @@ impl<'db> Bindings<'db> {
                             continue;
                         };
 
-                        let result = ty_a.when_subtype_of_given(
+                        let result = ty_a.when_subtype_of_assuming(
                             db,
                             *ty_b,
                             tracked.constraints(db),


AlexWaygood:
Are the primer hits here expected?

dcreager:
> Are the primer hits here expected?

They were not! Latest patch disappears them

sharkdp:
I am planning to review this, but postponing to tomorrow, since Doug is out today anyway.

dcreager:
> I am planning to review this, but postponing to tomorrow, since Doug is out today anyway.

Thank you! No rush

crates/ty_python_semantic/src/types.rs
```diff
@@ -10309,9 +10340,46 @@ pub(crate) enum TypeRelation {
...
+    /// assuming that the constraints in a particular constraint set hold.
+    ///
+    /// For concrete types (types that do not contain typevars), this relationship is the same as
+    /// [subtyping][Self::Subtyping]. (Constraint sets place restrictions on typevars, so if you
+    /// are not comparing typevars, the constraint set can have no effect on whether subtyping
+    /// holds.)
+    ///
+    /// If you're comparing a typevar, we have to consider what restrictions the constraint set
+    /// places on that typevar to determine if subtyping holds. For instance, if you want to check
+    /// whether `T ≤ int`, then answer will depend on what constraint set you are considering:
```
sharkdp:
```suggestion
    /// whether `T ≤ int`, then the answer will depend on what constraint set you are considering:
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -10309,9 +10340,46 @@ pub(crate) enum TypeRelation {
...
+    ///
+    /// - The "always true" constraint set does not place any restrictions on any typevar. In this
+    ///   case, `when_subtype_of_given` will return the same result as `when_subtype_of`, even if
+    ///   you're comparing against a typevar.
+    ///
+    /// - The "always false" constraint set represents an impossible situation. In this case, every
+    ///   subtype check will be vacuously true, even if you're comparing two concrete types that
+    ///   are not actually subtypes of each other. (That is,
+    ///   `when_subtype_of_given(false, int, str)` will return true!)
+    ConstraintImplication(ConstraintSet<'db>),
```
sharkdp: Is "constraint implication" the name in the literature? I think I would prefer something like
```suggestion
    SubtypingGivenPremises(ConstraintSet<'db>),
```
or
```suggestion
    SubtypingUnderAssumptions(ConstraintSet<'db>),
```
or
```suggestion
    SubtypingAssuming(ConstraintSet<'db>),
```
but I'm obviously also fine with what you have.

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
```diff
@@ -191,4 +190,155 @@ def mutually_constrained[T, U]():
...
+    # For a covariant typevar, (T ≤ int) implies that (Covariant[T] ≤ Covariant[int]).
+    given_int = ConstraintSet.range(Never, T, int)
+    static_assert(given_int.implies_subtype_of(Covariant[T], Covariant[int]))
+    static_assert(not given_int.implies_subtype_of(Covariant[T], Covariant[bool]))
+    static_assert(not given_int.implies_subtype_of(Covariant[T], Covariant[str]))
+
+    given_bool = ConstraintSet.range(Never, T, bool)
+    static_assert(given_bool.implies_subtype_of(Covariant[T], Covariant[int]))
+    static_assert(given_bool.implies_subtype_of(Covariant[T], Covariant[bool]))
+    static_assert(not given_bool.implies_subtype_of(Covariant[T], Covariant[str]))
```
sharkdp, line 224: Would it be useful to add tests with the typevar on the right hand side? For example
```suggestion
    static_assert(not given_bool.implies_subtype_of(Covariant[T], Covariant[str]))

    given_bool_le_t_le_int = ConstraintSet.range(bool, T, int)
    static_assert(not given_bool_le_t_le_int.implies_subtype_of(Covariant[int], Covariant[T]))
    static_assert(given_bool_le_t_le_int.implies_subtype_of(Covariant[bool], Covariant[T]))
    static_assert(not given_bool_le_t_le_int.implies_subtype_of(Covariant[str], Covariant[T]))
```

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
```diff
@@ -191,4 +190,155 @@ def mutually_constrained[T, U]():
...
+    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[int]))
+    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[bool]))
+    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[str]))
+
+    # If (T = U ∧ U = int), then (T = int) must be true as well. That is an equality constraint, so
+    # even though T is invariant, it does imply that (Invariant[T] ≤ Invariant[int]).
+    given_int = ConstraintSet.range(U, T, U) & ConstraintSet.range(int, U, int)
+    static_assert(given_int.implies_subtype_of(Invariant[T], Invariant[int]))
+    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[bool]))
+    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[str]))
```
sharkdp, line 348: Maybe test the reverse direction too?
```suggestion
    static_assert(given_int.implies_subtype_of(Invariant[T], Invariant[int]))
    static_assert(given_int.implies_subtype_of(Invariant[int], Invariant[T]))
    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[bool]))
    static_assert(not given_int.implies_subtype_of(Invariant[bool], Invariant[T]))
    static_assert(not given_int.implies_subtype_of(Invariant[T], Invariant[str]))
    static_assert(not given_int.implies_subtype_of(Invariant[str], Invariant[T]))
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -1618,6 +1618,25 @@ impl<'db> Type<'db> {
...
 
+    /// Return the constraints under which this type is a subtype of type `target`, assuming that
+    /// all of the restrictions in `constraints` hold.
+    ///
+    /// See [`TypeRelation::ConstraintImplication`] for more details.
+    fn when_subtype_of_given(
+        self,
+        db: &'db dyn Db,
+        target: Type<'db>,
+        constraints: ConstraintSet<'db>,
```
sharkdp: Maybe
```suggestion
        premises: ConstraintSet<'db>,
```
or
```suggestion
        assumptions: ConstraintSet<'db>,
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -1679,6 +1698,14 @@ impl<'db> Type<'db> {
             return ConstraintSet::from(true);
         }
 
+        // Handle constraint implication first. If either `self` or `target` is a typevar, check
+        // the constraint set to see if the corresponding constraint is satisfied.
+        if let TypeRelation::ConstraintImplication(constraints) = relation
+            && (self.is_type_var() || target.is_type_var())
+        {
+            return constraints.when_subtype_of_given(db, self, target);
```
sharkdp: I think I missed a few of your PRs, so I'm not sure if this came up before, but the naming here is a bit confusing to me. I think I would prefer something like below, but certainly no need to change anything here if this has been discussed before and/or if you prefer the current naming
```suggestion
            return premises.imply_subtype_of(db, self, target);
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -1481,7 +1446,18 @@ impl<'db> InteriorNode<'db> {
...
+            // other typevars, producing constraints on this typevar that have concrete lower/upper
+            // bounds. That means we can skip the simplifications below if any bound is another
+            // typevar.
+            if left_constraint.lower(db).is_type_var()
+                || left_constraint.upper(db).is_type_var()
+                || right_constraint.lower(db).is_type_var()
+                || right_constraint.upper(db).is_type_var()
+            {
+                continue;
+            }
```
sharkdp, line 1460: This sounds like an optimization, but seems to be required for correctness?

dcreager: Yep, as described (and clarified/fixed) in much more detail in https://github.com/astral-sh/ruff/pull/21463

crates/ty_python_semantic/src/types.rs
dcreager: Yes that also lines up more nicely with the name of the mdtest function. (That would change the tense slightly to `implies_subtype_of` which _technically_ doesn't agree with a plural `premises`, but I am going to choose not to worry about that.)

crates/ty_python_semantic/src/types.rs
dcreager: > Is "constraint implication" the name in the literature?

Nope! I do like `SubtypingAssuming`, that also calls out the relationship to `TypeRelation::Subtyping` more clearly.

crates/ty_python_semantic/src/types.rs
dcreager: Done

crates/ty_python_semantic/src/types.rs
dcreager: Done

crates/ty_python_semantic/src/types.rs
dcreager: Done

PR #21367, opened 2025-11-10
[ty] supress some trivial expr inlay hints

Gankra:
I'm not 100% sold on this implementation, but it's a strict improvement and it adds a ton of snapshot tests for future iteration.

Part of https://github.com/astral-sh/ty/issues/494

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -849,7 +849,14 @@ mod tests {
 
     #[test]
     fn test_disabled_variable_types() {
-        let test = inlay_hint_test("x = 1");
+        let test = inlay_hint_test(
+            "
+            def i(x: int, /) -> int:
+                return x
+
+            x = i(1)
+            ",
+        );
 
         assert_snapshot!(
             test.inlay_hints_with_settings(&InlayHintSettings {
@@ -857,7 +864,10 @@ mod tests {
                 ..Default::default()
             }),
             @r"
-        x = 1
+        def i(x: int, /) -> int:
+            return x
+
+        x = i(1)
         "
         );
     }


crates/ty_ide/src/inlay_hints.rs
```diff
@@ -415,47 +441,183 @@ mod tests {
...
+            ",
+        );
 
         assert_snapshot!(test.inlay_hints(), @r"
-        x[: Literal[1]] = 1
+        def i(x: int, /) -> int:
+            return x
+
+        x = 1
+        y[: Literal[1]] = x
```
Gankra, line 461: This case is why this approach is mediocre. The type-directed approach would catch this trivially.

AlexWaygood: I don't think this is actually so bad? It tells the user that we inferred the type of `y` as `Literal[1]`, but mypy would infer it as `int` here -- that's potentially useful information. The `y = x` assignment could also happen much lower down in the function, far away from the initial `x = 1` assignment

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -415,47 +441,183 @@ mod tests {
...
+        );
+
+        assert_snapshot!(test.inlay_hints(), @r#"
+        def i(x: int, /) -> int:
+            return x
+        def s(x: str, /) -> str:
+            return x
+
+        x1, y1 = (1, 'abc')
+        x2[: Literal[1]], y2[: Literal["abc"]] = (x1, y1)
```
AlexWaygood, line 490: this, on the other hand, feels kinda yucky to me 😆

AlexWaygood (approved):

PR #21368, opened 2025-11-10
[ty] supress inlay hints for `+1` and `-1`

Gankra:
It's everyone's favourite language corner case!

Also having kicked the tires on it, I'm pretty happy to call this (in conjunction with #21367):

Fixes https://github.com/astral-sh/ty/issues/494

There's cases where you can make noisy Literal hints appear, so we can always iterate on it, but this handles like, 98% of the cases in the wild, which is great.

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -346,7 +346,7 @@ fn type_hint_is_excessive_for_expr(expr: &Expr) -> bool {
         | Expr::TString(_)=> true,
 
         // You too `+1 and `-1`, get back here  
-        | Expr::UnaryOp(ExprUnaryOp { op: UnaryOp::UAdd | UnaryOp::USub, operand, .. }) => matches!(**operand, Expr::NumberLiteral(_)),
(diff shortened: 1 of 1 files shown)


crates/ty_ide/src/inlay_hints.rs
```diff
@@ -345,6 +345,9 @@ fn type_hint_is_excessive_for_expr(expr: &Expr) -> bool {
         // This one expands to `Template` which isn't verbose but is redundant
         | Expr::TString(_)=> true,
 
+        // You too `+1 and `-1`, get back here  
+        | Expr::UnaryOp(ExprUnaryOp { op: UnaryOp::UAdd | UnaryOp::USub, operand, .. }) => matches!(**operand, Expr::NumberLiteral(_)),
```
sharkdp: Wow, how did I not know that this is legal syntax? :smile:

In general, I think I understand the reasoning behind this syntactic feature (better diffs, easier editing), but for this particular case, I think I would prefer the version that seems more readable to me (it seems unlikely that we add more patterns to this exact case), but definitely no strong opinion.
```suggestion
        Expr::UnaryOp(ExprUnaryOp { op: UnaryOp::UAdd | UnaryOp::USub, operand, .. }) => matches!(**operand, Expr::NumberLiteral(_)),
```

PR #21372, opened 2025-11-11
[ty] support absolute `from` imports introducing local submodules in `__init__.py` files

Gankra:
By resolving `.` and the LHS of the from import during semantic indexing, we can check if the LHS is a submodule of `.`, and handle `from whatever.thispackage.x.y import z` exactly like we do `from .x.y import z`.

Fixes https://github.com/astral-sh/ty/issues/1484

### DIFF
--- a/crates/ty_python_semantic/src/module_name.rs
+++ b/crates/ty_python_semantic/src/module_name.rs
@@ -295,6 +295,7 @@ impl ModuleName {
         Self::from_identifier_parts(db, importing_file, module.as_deref(), *level)
     }
 
+    /// Computes the absolute module name from the LHS components of `from LHS import RHS`
     pub(crate) fn from_identifier_parts(
         db: &dyn Db,
         importing_file: File,
@@ -308,6 +309,16 @@ impl ModuleName {
                 .and_then(Self::new)
                 .ok_or(ModuleNameResolutionError::InvalidSyntax)
         }
+    }
+
+    /// Computes the absolute module name for the package this file belongs to.
+    ///
+    /// i.e. this resolves `.`
+    pub(crate) fn package_for_file(
+        db: &dyn Db,
+        importing_file: File,
+    ) -> Result<Self, ModuleNameResolutionError> {
+        Self::from_identifier_parts(db, importing_file, None, 1)
     }
 }
 
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -1465,8 +1465,8 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                 // reasons but it works well for most practical purposes. In particular it's nice
                 // that `x` can be freely overwritten, and that we don't assume that an import
                 // in one function is visible in another function.
-                if let Some(submodule) = &node.module
-                    && self.current_scope() == FileScopeId::global()
+                if node.module.is_some()
+                    && self.current_scope().is_global()
                     && self.file.is_package(self.db)
                     && let Ok(module_name) = ModuleName::from_identifier_parts(
                         self.db,
@@ -1474,8 +1474,7 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                         node.module.as_deref(),
                         node.level,
                     )
-                    && let Ok(thispackage) =
-                        ModuleName::from_identifier_parts(self.db, self.file, None, 1)
+                    && let Ok(thispackage) = ModuleName::package_for_file(self.db, self.file)
                     && let Some(relative_submodule) = module_name.relative_to(&thispackage)
                     && let Some(direct_submodule) = relative_submodule.components().next()
                     && !self.seen_submodule_imports.contains(direct_submodule)
@@ -1487,7 +1486,7 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                     let symbol = self.add_symbol(direct_submodule_name);
                     self.add_definition(
                         symbol.into(),
-                        ImportFromSubmoduleDefinitionNodeRef { node, submodule },
+                        ImportFromSubmoduleDefinitionNodeRef { node },
                     );
                 }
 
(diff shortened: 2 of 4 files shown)


Gankra:
(This implementation isn't the prettiest, especially in the infer side... something I should iterate on tomorrow probably)

Gankra:
Some of these results are perplexing... some completely randomly added missing import errors.

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1465,14 +1465,19 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
                 // in one function is visible in another function.
-                //
-                // TODO: Also support `from thispackage.x.y import z`?
-                if self.current_scope() == FileScopeId::global()
-                    && node.level == 1
-                    && let Some(submodule) = &node.module
-                    && let Some(parsed_submodule) = ModuleName::new(submodule.as_str())
-                    && let Some(direct_submodule) = parsed_submodule.components().next()
+                if let Some(submodule) = &node.module
+                    && self.current_scope() == FileScopeId::global()
```
MichaReiser: Nit:

```suggestion
                    && self.current_scope().is_global()
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5916,13 +5916,14 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         // this type of Definition is only created when a `from..import` is in global scope.)
 
         // Get this package's module by resolving `.`
-        let Ok(module_name) = ModuleName::from_identifier_parts(self.db(), self.file(), None, 1)
+        let Ok(thispackage_name) =
+            ModuleName::from_identifier_parts(self.db(), self.file(), None, 1)
```
MichaReiser: Nit: I had to lookup the definition of `from_identifier_parts` to understanding what we're doing here. Should we expose a `ModuleName::package_for(self.db(), self.file())` that hides the `tail` and `level` stuff (which I find more confusing here than useful)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -5932,12 +5933,26 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            return;
+        };
+        let Some(relative_submodule_name) = submodule_name.relative_to(&thispackage_name) else {
+            self.add_binding(import_from.into(), definition, |_, _| Type::unknown());
+            return;
+        };
+        let Some(name) = relative_submodule_name.components().next() else {
+            self.add_binding(import_from.into(), definition, |_, _| Type::unknown());
+            return;
+        };
```
MichaReiser: Nit. To reduce the duplication between the `else` branches
```suggestion
        let Ok(name) = ModuleName::from_identifier_parts(
            self.db(),
            self.file(),
            import_from.module.as_deref(),
            import_from.level,
        ).and_then(|submodule_name| submodule_name.relative_to(&thispackage_name))
         .and_then(|relative_submodule_name| relative_submodule_name.components().next()) else {
            self.add_binding(import_from.into(), definition, |_, _| Type::unknown());
            return;
        };
```

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1465,14 +1465,19 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
                     && self.file.is_package(self.db)
+                    && let Ok(module_name) = ModuleName::from_identifier_parts(
+                        self.db,
+                        self.file,
+                        node.module.as_deref(),
+                        node.level,
+                    )
+                    && let Ok(thispackage) =
+                        ModuleName::from_identifier_parts(self.db, self.file, None, 1)
```
MichaReiser: I wonder if you see the new failing imports because this operation can fail if the path to the next search root contains any non-valid-identifier parts (e.g. looking up `core-db/root` will fail if the only search path is at `/`)

AlexWaygood:
> Some of these results are perplexing... some completely randomly added missing import errors.

Which diagnostics are you talking about specifically here? I only see one added `possibly-missing-import` diagnostic in the ecosystem-analyzer summary comment, and no new `unresolved-import` diagnostics

MichaReiser:
> Some of these results are perplexing... some completely randomly added missing import errors.

This one might actually be correct because it's a re-export and the import is within a `try...except`

Gankra:
Screenshot 2025-11-11 at 7 54 44 AM

stuff like this -- maybe this is just telling me that we had an Unknown here before and now we don't?

AlexWaygood:
> maybe this is just telling me that we had an Unknown here before and now we don't?

Yeah that seems quite likely to me! Would be good to spot-check one or two and spelunk through them to confirm, though 👍

Gankra:
Ok yeah the ecosystem results look reasonable, great. Just need to cleanup.

Gankra:
Cleaned up impl is up.

crates/ty_python_semantic/src/types/infer/builder.rs
Gankra: I was gonna complain that I need some of these temporaries in scope but you actually perfectly identified that and only factored out the part that doesn't need it, incredible.

MichaReiser (approved):

crates/ty_python_semantic/src/module_name.rs
```diff
@@ -309,6 +310,16 @@ impl ModuleName {
...
     }
+
+    /// Computes the absolute module name for the package this file belongs to.
+    ///
+    /// i.e. this resolves `.`
+    pub(crate) fn package_for_file(
+        db: &dyn Db,
+        importing_file: File,
+    ) -> Result<Self, ModuleNameResolutionError> {
+        Self::from_identifier_parts(db, importing_file, None, 1)
```
AlexWaygood, line 321: it looks like this will always take the `if let Some(level) = NonZeroU32::new(level)` branch in `ModuleName::from_identifier_parts()`, because `NonZeroU32::new(1)` will always return `Some()`. So can't this be done with less indirection as

```suggestion
        relative_module_name(db, importing_file, None, NonZeroU32::new(1).unwrap())
```

?

Gankra: Ah true, good catch!

PR #15951, opened 2025-02-04
[`flake8-builtins`] Make strict module name comparison optional (`A005`)

ntBre:
## Summary

This PR adds the configuration option `lint.flake8-builtins.builtins-strict-checking`, which is used in A005 to determine whether the fully-qualified module name (relative to the project root or source directories) should be checked instead of just the final component as is currently the case.

As discussed in https://github.com/astral-sh/ruff/issues/15399#issuecomment-2587017147, the default value of the new option is `false`, so modules like `utils.logging` from the initial report are no longer flagged by default.

## Test Plan

New A005 test module with the structure reported in #15399.

PR #12838, opened 2024-08-12
Change default for PT001 and PT023

MichaReiser:
## Summary

This PR promotes the changed defaults for PT001 and PT023 to not require parentheses for fixtures and marker decorators without arguments to stable. 

The default were changed in preview mode back in June (https://github.com/astral-sh/ruff/pull/12106). 

See https://github.com/astral-sh/ruff/issues/8796

## Test Plan

Updated snapshot tests

PR #1098, opened 2022-12-06
Implement import conventions

edgarrmondragon:
Closes #1004

PR #21373, opened 2025-11-11
[`flake8-import-conventions`] Add missing conventions from upstream (`ICN001`, `ICN002`)

### LINKED ISSUE #21300: ICN rules miss some conventions from flake8-import-conventions
### Summary

[`unconventional-import-alias` (ICN001)](https://docs.astral.sh/ruff/rules/unconventional-import-alias/) and possibly [`banned-import-alias` (ICN002)](https://docs.astral.sh/ruff/rules/banned-import-alias/) are missing some conventions enforced by the upstream [flake8-import-conventions](https://github.com/joaopalmeiro/flake8-import-conventions).

For ICN001, `plotly.graph_objects` should be imported as `go` and `statsmodels.api` should be imported as `sm`.

For ICN002, `geopandas` should not be imported as `gpd`. Actually, flake8-import-conventions enforces that it not be aliased at all, but Ruff doesn’t support that. The specific alias that prompted the upstream rule is `gpd`, banning which is the closest Ruff can get.

[Example](https://play.ruff.rs/63025674-955a-482b-b786-23badd5686a1):
```console
$ cat >icn.py <<'# EOF'
import geopandas as gpd
import plotly.graph_objects
import statsmodels.api
# EOF

$ ruff --isolated check icn.py --select ICN001,ICN002
All checks passed!

$ flake8 --select IC icn.py
icn.py:1:1: IC002 geopandas should be imported as `import geopandas`
icn.py:2:1: IC008 plotly.graph_objects should be imported as `import plotly.graph_objects as go`
icn.py:3:1: IC010 statsmodels.api should be imported as `import statsmodels.api as sm`
```

### Version

danparizher:
## Summary

Adds missing import conventions from upstream flake8-import-conventions plugin. Specifically:
- `ICN001`: Enforces `plotly.graph_objects` → `go` and `statsmodels.api` → `sm`
- `ICN002`: Bans `geopandas` → `gpd` alias

Fixes #21300

## Problem Analysis

Ruff's `ICN001` (unconventional-import-alias) and `ICN002` (banned-import-alias) rules were missing some conventions enforced by the upstream [flake8-import-conventions](https://github.com/joaopalmeiro/flake8-import-conventions) plugin:

1. **ICN001 missing conventions:**
   - `plotly.graph_objects` should be imported as `go` (IC008 in upstream)
   - `statsmodels.api` should be imported as `sm` (IC010 in upstream)

2. **ICN002 missing convention:**
   - `geopandas` should not be imported as `gpd` (IC002 in upstream)

The root cause was that these conventions were not included in Ruff's default configuration. The `CONVENTIONAL_ALIASES` constant was missing the two ICN001 entries, and there was no default banned aliases configuration for ICN002.

## Approach

1. **Added missing ICN001 conventions:**
   - Added `("plotly.graph_objects", "go")` and `("statsmodels.api", "sm")` to `CONVENTIONAL_ALIASES` in `settings.rs`
   - Updated the default option string in `options.rs` to include these new aliases

2. **Added default banned aliases for ICN002:**
   - Created `default_banned_aliases()` function returning `geopandas` → `["gpd"]`
   - Updated `Settings::default()` to use `default_banned_aliases()`
   - Updated `try_into_settings()` in `options.rs` to use `default_banned_aliases()` when `banned_aliases` is None
   - Updated the default option string for `banned_aliases` in `options.rs`

3. **Added comprehensive tests:**
   - Created `missing_conventions.py` test fixture with cases for all three new conventions
   - Added `missing_conventions` test in `mod.rs` to verify the new conventions work correctly

## Test Plan

Added new snapshot test `missing_conventions` that verifies:
- ICN001 correctly flags `plotly.graph_objects` without alias and requires `go`
- ICN001 correctly flags `statsmodels.api` without alias and requires `sm`
- ICN002 correctly flags `geopandas as gpd` as banned
- All existing tests continue to pass (10/10 tests passing)

The fix has been manually verified to match the behavior of upstream flake8-import-conventions.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_import_conventions/settings.rs
+++ b/crates/ruff_linter/src/rules/flake8_import_conventions/settings.rs
@@ -81,8 +81,7 @@ pub fn default_aliases(preview: PreviewMode) -> FxHashMap<String, String> {
         aliases.extend(
             PREVIEW_ALIASES
                 .iter()
-                .map(|(k, v)| ((*k).to_string(), (*v).to_string()))
-                .collect::<FxHashMap<_, _>>(),
+                .map(|(k, v)| ((*k).to_string(), (*v).to_string())),
         );
     }
 


crates/ruff_linter/resources/test/fixtures/flake8_import_conventions/missing_conventions.py
ntBre, line 1: Should we just add these to one of the existing test files? `defaults.py` seems promising. `missing_conventions` seems a bit misleading since they're not missing anymore after your PR.

ntBre (commented):
On second thought, could we make this a preview change? Initially I assumed this was an unintentional omission from the original PR adding the rules (https://github.com/astral-sh/ruff/pull/1098), but I see now that these were changed upstream _after_ our rule was added (https://github.com/joaopalmeiro/flake8-import-conventions/issues/3, https://github.com/joaopalmeiro/flake8-import-conventions/issues/4). These issues were opened upstream before our rule but closed after. And there are a few hits in the ecosystem report for `geopandas` too.

crates/ruff_linter/src/rules/flake8_import_conventions/rules/banned_import_alias.rs
```diff
@@ -56,7 +57,17 @@ pub(crate) fn banned_import_alias(
...
+    let banned_aliases = if is_import_conventions_preview_enabled(checker.settings()) {
+        banned_conventions.get(name).cloned().or_else(|| {
+            let preview_banned = preview_banned_aliases();
+            preview_banned.get(name).cloned()
+        })
+    } else {
+        banned_conventions.get(name).cloned()
+    };
+
+    if let Some(banned_aliases) = banned_aliases.as_ref() {
```
ntBre: Do we need to change this code? I thought we could handle this just by adjusting the preview behavior of the default settings.

crates/ruff_linter/src/rules/flake8_import_conventions/settings.rs
```diff
@@ -75,18 +76,37 @@ pub fn default_aliases() -> FxHashMap<String, String> {
...
     FxHashMap::from_iter([(
         "geopandas".to_string(),
         BannedAliases::from_iter(["gpd".to_string()]),
     )])
 }
 
+impl Settings {
+    /// Merge preview aliases and banned aliases into the settings if preview mode is enabled.
+    #[must_use]
+    pub fn with_preview(mut self, preview_enabled: bool) -> Self {
```
ntBre: I think the way we usually handle this is by defining a method that takes a `preview` argument instead of `impl Default` (which can't take arguments). This is kind of the opposite since it was a stabilization PR, but you can see an example here: https://github.com/astral-sh/ruff/pull/12838/files#diff-0e4daa1a461406d9fd7af87b85ff4ecc89bd61f748d2b7cb24b88b25c7d5ba59

Possibly a better example: https://github.com/astral-sh/ruff/pull/15951

crates/ruff_workspace/src/options.rs
```diff
@@ -1672,6 +1673,11 @@ impl Flake8ImportConventionsOptions {
             );
         }
 
+        // Merge preview aliases if preview mode is enabled
+        if preview.is_enabled() {
+            aliases.extend(flake8_import_conventions::settings::preview_aliases());
+        }
+
```
ntBre: I don't think we should set this here. If I understand correctly, this will extend the list of aliases when preview is enabled even if the user provided a set of aliases. We only want to extend the _default_ set in the case where `self.aliases` is `None` above. I think it would make sense for the `default_aliases` helper function to take the `preview` argument and apply it there.

crates/ruff_workspace/src/options.rs
```diff
@@ -1683,9 +1689,16 @@ impl Flake8ImportConventionsOptions {
             normalized_aliases.insert(module, normalized_alias);
         }
 
+        let mut banned_aliases = self.banned_aliases.unwrap_or_default();
+
+        // Merge preview banned aliases if preview mode is enabled
+        if preview.is_enabled() {
+            banned_aliases.extend(flake8_import_conventions::settings::preview_banned_aliases());
+        }
```
ntBre: Same as above, we should only do this in the `or_default` case.

ntBre (commented):
Thanks, this is looking good. I fixed a few nits, reverting some unrelated changes and tweaking the test setup, but I also wanted to get your thoughts on the `try_into_settings` changes. See my inline comments for that.

ntBre:
Not sure why the ecosystem check dropped a diagnostic after the last push, I'm still seeing 4 locally. I think I'll go ahead and land this anyway once CI finishes again.

PR #21375, opened 2025-11-11
[`pyupgrade`] Skip UP007 for dynamic Union creation (`UP007`)

### LINKED ISSUE #21347: UP007 emitted for dynamic `Union` creation
### Summary

It's possible to pass a tuple of types to `Union` at the run time:

```py
from typing import Union


def f(types: tuple[type, ...]):
    return Union[types]


if __name__ == "__main__":
    u = f((int, str, float))
    print(u)  # typing.Union[int, str, float]
    print(type(u))  # <class 'typing._UnionGenericAlias'>
```

ruff emits ```UP007 Use `X | Y` for type annotations``` for this, but I don't see how can that be done.

Playground link: https://play.ruff.rs/b9d9c214-7cf5-473e-914f-264a6da9a92d

### Version

0.14.4

danparizher:
## Summary

Fix UP007 to skip dynamic `Union` creation where the slice is a runtime variable (fixes #21347).

The rule was incorrectly emitting "Use `X | Y` for type annotations" for cases like `Union[types]` where `types` is a variable, which cannot be converted to PEP 604 syntax since the types are determined at runtime.

## Problem Analysis

UP007 checks for `Union[...]` annotations that can be rewritten using PEP 604's `|` operator syntax. However, the rule was triggering even when the slice argument was a variable (e.g., `Union[types]`), which represents dynamic Union creation at runtime. This is invalid because:

1. The types are not known statically
2. PEP 604 syntax (`X | Y`) requires compile-time type literals
3. The suggested fix cannot be applied

The rule should only apply to static type annotations with literal tuples or type literals that can be statically analyzed.

## Approach

Added an early return check in `non_pep604_annotation` for the `Union` operator branch to skip emitting diagnostics when the slice is an `Expr::Name` (variable). This prevents false positives for dynamic Union creation while preserving correct behavior for static type annotations.

The fix is minimal and targeted - it only adds a guard clause before the diagnostic is emitted, ensuring that dynamic Union creation cases are filtered out early.

## Test Plan

Added a regression test case in `UP007.py` that demonstrates dynamic Union creation (`Union[types]` where `types` is a variable) should not trigger UP007. All existing pyupgrade tests pass (151 tests), confirming the fix doesn't break existing functionality.


dylwil3:
Thanks for working on this! I'm not sure that this is fixing the root of the problem - for example it wouldn't catch `Union[foo()]` where `foo` returns a tuple of types. Instead, I think we should only be applying this lint in type annotation positions, as @ntBre suggested in the linked issue.

On an unrelated note: I personally do not find the AI-generated summaries in your PRs to be helpful. First, they are unnecessarily verbose and often include overly-confident or misleading statements. Second, they do not give me much confidence that a human being has reviewed the code changes. I would much prefer to read a concise summary that you wrote yourself.

danparizher:
Thanks for the feedback @dylwil3, I'll write the PR descriptions myself going forward and be more attentive to the code changes that AI generates.

Regarding this PR, I caught myself up in the linked issue and made a few changes following the discussion. Let me know if this revision is on the right track!

ntBre:
Sorry for the delay in review here. Are you still interested in working on this? Another related issue came in today with another interesting test case: https://github.com/astral-sh/ruff/issues/27238.

PR #21376, opened 2025-11-11
[ty] Make `CycleDetector::visit` take `&mut self`

MichaReiser:
I have a dislike for `RefCell` but I'm not sure if this is worth it. 

Curious to hear what other people think (very happy to close, doesn't have to be a long explanation why)

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -109,7 +109,7 @@ impl<'db> Completion<'db> {
         fn imp<'db>(
             db: &'db dyn Db,
             ty: Type<'db>,
-            visitor: &CompletionKindVisitor<'db>,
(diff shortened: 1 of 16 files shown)


AlexWaygood:
I think this is basically a straight revert of https://github.com/astral-sh/ruff/pull/19871? I weakly prefer it the way it is just for concision, and I do also sorta agree with Carl's reasoning in https://github.com/astral-sh/ruff/pull/19871#issue-3311732446 :-)

carljm:
Yeah, this is just going back to what we originally had. We changed it to use interior mutability because it was a pain to deal with the borrow checker. E.g. it introduces the need for otherwise-unnecessary `collect()` in `tuple.rs`.

I would prefer to leave this as it is.

PR #21378, opened 2025-11-11
[ty] Faster subscript assignment checks for (unions of) `TypedDict`s

sharkdp:
## Summary

We synthesize a (potentially large) set of `__setitem__` overloads for every item in a `TypedDict`. Previously, validation of subscript assignments on `TypedDict`s relied on actually calling `__setitem__` with the provided key and value types, which implied that we needed to do the full overload call evaluation for this large set of overloads. This PR improves the performance of subscript assignment checks on `TypedDict`s by validating the assignment directly instead of calling `__setitem__`.

This PR also adds better handling for assignments to subscripts on union and intersection types (but does not attempt to make it perfect). It achieves this by distributing the check over unions and intersections, instead of calling `__setitem__` on the union/intersection directly. We already do something similar when validating *attribute* assignments.

## Ecosystem impact

* A lot of diagnostics change their rule type, and/or split into multiple diagnostics. The new version is more verbose, but easier to understand, in my opinion
* Almost all of the invalid-key diagnostics come from pydantic, and they should all go away (including many more) when we implement https://github.com/astral-sh/ty/issues/1479
* Everything else looks correct to me. There may be some new diagnostics due to the fact that we now check intersections.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Possibly_missing_`__…_(efd3f0c02e9b89e9).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Possibly_missing_`__…_(efd3f0c02e9b89e9).snap
@@ -13,19 +13,20 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
 
 ```
 1 | def _(config: dict[str, int] | None) -> None:
-2 |     config["retries"] = 3  # error: [possibly-missing-implicit-call]
+2 |     config["retries"] = 3  # error: [invalid-assignment]
 ```
 
 # Diagnostics
 
 ```
-warning[possibly-missing-implicit-call]: Method `__setitem__` of type `dict[str, int] | None` may be missing
+error[invalid-assignment]: Cannot assign to a subscript on an object of type `None` with no `__setitem__` method
  --> src/mdtest_snippet.py:2:5
   |
 1 | def _(config: dict[str, int] | None) -> None:
-2 |     config["retries"] = 3  # error: [possibly-missing-implicit-call]
+2 |     config["retries"] = 3  # error: [invalid-assignment]
   |     ^^^^^^
   |
-info: rule `possibly-missing-implicit-call` is enabled by default
+info: The full type of the subscripted object is `dict[str, int] | None`
+info: rule `invalid-assignment` is enabled by default
 
 ```
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
@@ -22,19 +22,39 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
  8 |     legs: int
  9 | 
 10 | def _(being: Person | Animal) -> None:
-11 |     being["surname"] = "unknown"  # error: [invalid-assignment]
+11 |     # error: [invalid-key]
+12 |     # error: [invalid-key]
+13 |     being["surname"] = "unknown"
 ```
 
 # Diagnostics
 
 ```
-error[invalid-assignment]: Method `__setitem__` of type `(key: Literal["name"], value: str, /) -> None` cannot be called with a key of type `Literal["surname"]` and a value of type `Literal["unknown"]` on object of type `Person | Animal`
-  --> src/mdtest_snippet.py:11:5
+error[invalid-key]: Invalid key for TypedDict `Person`
+  --> src/mdtest_snippet.py:13:5
    |
-10 | def _(being: Person | Animal) -> None:
-11 |     being["surname"] = "unknown"  # error: [invalid-assignment]
-   |     ^^^^^
+11 |     # error: [invalid-key]
+12 |     # error: [invalid-key]
+13 |     being["surname"] = "unknown"
+   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     |
+   |     TypedDict `Person` in union type `Person | Animal`
    |
-info: rule `invalid-assignment` is enabled by default
+info: rule `invalid-key` is enabled by default
+
+```
+
+```
+error[invalid-key]: Invalid key for TypedDict `Animal`
+  --> src/mdtest_snippet.py:13:5
+   |
+11 |     # error: [invalid-key]
+12 |     # error: [invalid-key]
+13 |     being["surname"] = "unknown"
+   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     |
+   |     TypedDict `Animal` in union type `Person | Animal`
+   |
+info: rule `invalid-key` is enabled by default
 
 ```
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
@@ -89,7 +89,7 @@ info: rule `invalid-key` is enabled by default
 ```
 
 ```
-error[invalid-key]: Invalid key for TypedDict `Person` of type `str`
+error[invalid-key]: Invalid key of type `str` for TypedDict `Person`
   --> src/mdtest_snippet.py:16:12
    |
 15 | def access_with_str_key(person: Person, str_key: str):
--- a/crates/ty_python_semantic/resources/mdtest/typed_dict.md
+++ b/crates/ty_python_semantic/resources/mdtest/typed_dict.md
@@ -69,7 +69,7 @@ def name_or_age() -> Literal["name", "age"]:
 carol: Person = {NAME: "Carol", AGE: 20}
 
 reveal_type(carol[NAME])  # revealed: str
-# error: [invalid-key] "Invalid key for TypedDict `Person` of type `str`"
+# error: [invalid-key] "Invalid key of type `str` for TypedDict `Person`"
 reveal_type(carol[non_literal()])  # revealed: Unknown
 reveal_type(carol[name_or_age()])  # revealed: str | int | None
 
@@ -553,7 +553,7 @@ def _(
     # error: [invalid-key] "Invalid key for TypedDict `Person`: Unknown key "non_existing""
     reveal_type(person["non_existing"])  # revealed: Unknown
 
-    # error: [invalid-key] "Invalid key for TypedDict `Person` of type `str`"
+    # error: [invalid-key] "Invalid key of type `str` for TypedDict `Person`"
     reveal_type(person[str_key])  # revealed: Unknown
 
     # No error here:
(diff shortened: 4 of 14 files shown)


crates/ty_python_semantic/resources/mdtest/typed_dict.md
```diff
@@ -580,13 +602,33 @@ def _(person: Person, literal_key: Literal["age"]):
...
 
+def _(being: Person | Animal):
+    being["name"] = "Being"
+
+    # error: [invalid-assignment] "Invalid assignment to key "name" with declared type `str` on TypedDict `Person`: value of type `Literal[1]`"
+    # error: [invalid-assignment] "Invalid assignment to key "name" with declared type `str` on TypedDict `Animal`: value of type `Literal[1]`"
+    being["name"] = 1
+
+    # TODO: this could be improved
+    # error: [invalid-assignment] "Method `__setitem__` of type `((key: Literal["name", "surname", "age"], value: Any, /) -> None) | ((key: Literal["name", "legs"], value: Any, /) -> None)` cannot be called with a key of type `Literal["surname"]` and a value of type `Literal["unknown"]` on object of type `Person | Animal`"
```
sharkdp: The message on `main` is not better either, I did not attempt to improve this here.

AlexWaygood:
I wondered about how this might affect assignability and subtyping of `TypedDict`s to protocols, e.g.

```py
from ty_extensions import static_assert, is_subtype_of, is_assignable_to
from typing import Protocol, Literal, TypedDict

class Foo(Protocol):
    def __setitem__(self, key: Literal["x"], value: int, /) -> None: ...

class Bar(TypedDict):
    x: int

static_assert(is_assignable_to(Bar, Foo))
static_assert(is_subtype_of(Bar, Foo))
```

but I suppose it's impossible to test that right now, because `TypedDict`s are currently considered assignable to everything and subtypes of nothing, until https://github.com/astral-sh/ty/issues/1387 is fixed?

AlexWaygood:
If need be, we could always add a special case for `TypedDict -> Protocol` assignability/subtyping (we could synthesize the required method on the fly during the assignability/subtyping check?), but it's nice to keep that kind of special case to a minimum.

sharkdp:
> I wondered about how this might affect assignability and subtyping of `TypedDict`s to protocols, e.g.

Did you steal that from my comment yesterday? :smile:

I guess the PR would complicate the situation, but no other type checker generates these precise overloads either. I guess we could maybe keep these overloads and then prevent calling `__setitem__` in `validate_subscript_assignment`? But I'd rather like to postpone this discussion for now, unless you think it's important.

AlexWaygood:
> I guess we could maybe keep these overloads and then prevent calling `__setitem__` in `validate_subscript_assignment`?

I think I would prefer that? It'll probably make Jack's life easier, considering that he intends to start working on https://github.com/astral-sh/ty/issues/1387 today, I think. It doesn't seem like it'll be too hard to make that change in this PR, and it seems silly to get rid of the overloads here if there's a strong chance we'll only add them back in one or two days to fix this edge case.

And yes, other type checkers don't do great on this edge case either, but I hope that we can make ty better than other type checkers on things like this 😄

sharkdp:
> I think I would prefer that? It'll probably make Jack's life easier, considering that he intends to start working on [astral-sh/ty#1387](https://github.com/astral-sh/ty/issues/1387) today, I think. It doesn't seem like it'll be too hard to make that change in this PR, and it seems silly to get rid of the overloads here if there's a strong chance we'll only add them back in one or two days to fix this edge case.

I made the change, and I like it better. It lets us keep the overloads and improves the diagnostics in one case. But the optimization doesn't work anymore on pydantic :weary:. Looking into it.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3538,142 +3538,298 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            object_ty,
+            slice.as_ref(),
+            slice_ty,
+            rhs_value,
+            rhs_value_ty,
+            true,
+        )
+    }
+
+    #[expect(clippy::too_many_arguments)]
```
sharkdp, line 3571: No, I am not proud, but it feels like there are more important things. Will add this as a personal TODO note.

crates/ruff_benchmark/benches/ty_walltime.rs
```diff
@@ -181,7 +181,7 @@ static PYDANTIC: Benchmark = Benchmark::new(
         max_dep_date: "2025-06-17",
         python_version: PythonVersion::PY39,
     },
-    3000,
+    5000,
```
sharkdp, line 184: The increase here is caused by the fact that we now split diagnostic on unions into N_elements diagnostics.

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Possibly_missing_`__…_(efd3f0c02e9b89e9).snap
```diff
@@ -13,19 +13,20 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
...
  --> src/mdtest_snippet.py:2:5
   |
 1 | def _(config: dict[str, int] | None) -> None:
-2 |     config["retries"] = 3  # error: [possibly-missing-implicit-call]
+2 |     config["retries"] = 3  # error: [invalid-assignment]
   |     ^^^^^^
   |
-info: rule `possibly-missing-implicit-call` is enabled by default
+info: The full type of the subscripted object is `dict[str, int] | None`
+info: rule `invalid-assignment` is enabled by default
```
AlexWaygood, line 30: This is definitely an improvement, but he summary line here still feels too long to me -- for a blanket "cannot assign to a subscript" diagnostic, there's no other reason why this would be emitted except for a missing `__setitem__` method. Ideally I'd go for something like this:

```
error[invalid-assignment]: Cannot assign to a subscript on an object of type `None`
 --> src/mdtest_snippet.py:2:5
  |
1 | def _(config: dict[str, int] | None) -> None:
2 |     config["retries"] = 3  # error: [invalid-assignment]
  |     ^^^^^^
  |
info: Subscript assignment implicitly calls `__setitem__`, but `None` has no `__setitem__` method
info: `config` has type `dict[str, int] | None`
info: rule `invalid-assignment` is enabled by default
```

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Possibly_missing_`__…_(efd3f0c02e9b89e9).snap
```diff
@@ -13,19 +13,20 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
...
 
 ```
-warning[possibly-missing-implicit-call]: Method `__setitem__` of type `dict[str, int] | None` may be missing
+error[invalid-assignment]: Cannot assign to a subscript on an object of type `None` with no `__setitem__` method
  --> src/mdtest_snippet.py:2:5
   |
 1 | def _(config: dict[str, int] | None) -> None:
-2 |     config["retries"] = 3  # error: [possibly-missing-implicit-call]
+2 |     config["retries"] = 3  # error: [invalid-assignment]
   |     ^^^^^^
```
AlexWaygood, line 27: the highlighted range here still looks incorrect -- shouldn't it be

```suggestion
2 |     config["retries"] = 3  # error: [invalid-assignment]
  |     ^^^^^^^^^^^^^^^^^^^^^
```

?

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
```diff
@@ -22,19 +22,39 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
...
-11 |     being["surname"] = "unknown"  # error: [invalid-assignment]
-   |     ^^^^^
+11 |     # error: [invalid-key]
+12 |     # error: [invalid-key]
+13 |     being["surname"] = "unknown"
+   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     |
+   |     TypedDict `Person` in union type `Person | Animal`
+   |
+info: rule `invalid-key` is enabled by default
```
AlexWaygood, line 43: What about including the name of the unknown key in the summary line here and taking it out of the primary-annotation message?

```
error[invalid-key]: Invalid key "surname" for TypedDict `Person`
  --> src/mdtest_snippet.py:13:5
   |
11 |     # error: [invalid-key]
12 |     # error: [invalid-key]
13 |     being["surname"] = "unknown"
   |     ----- ^^^^^^^^^ Did you mean "name"?
   |     |
   |     TypedDict `Person` in union type `Person | Animal`
   |
info: rule `invalid-key` is enabled by default
```

We could also consider linking back to the definition of the `Person` typeddict in a subdiagnostic, though I guess that would be _very_ verbose if you had a union of typeddicts.

crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
```diff
@@ -89,7 +89,7 @@ info: rule `invalid-key` is enabled by default
 ```
 
 ```
-error[invalid-key]: Invalid key for TypedDict `Person` of type `str`
+error[invalid-key]: Invalid key of type `str` for TypedDict `Person`
```
AlexWaygood, line 92: should we have an `info: TypedDicts can only be subscripted with string literals` subdiagnostic here? This might be a bit opaque for users unfamiliar with the rules around typeddicts

crates/ty_python_semantic/src/types/typed_dict.rs
```diff
@@ -143,30 +143,57 @@ impl TypedDictAssignmentKind {
 pub(super) fn validate_typed_dict_key_assignment<'db, 'ast>(
```
AlexWaygood, line 143: Calls to this function are becoming quite difficult to read because of how many arguments there are. Especially the third argument, where `None` is often passed, and the last argument which is a `bool`. We could consider refactoring this into a struct that holds all the options for better readability, e.g.

```rs
let validator = TypedDictValidator {
    context: &self.context,
    typed_dict,
    full_object_ty: None,
    key,
    value_ty,
    typed_dict_node,
    key_node,
    value_node,
    assignment_kind,
    emit_diagnostics: true
};

validator.validate();
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3538,142 +3538,305 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            break;
+                        }
+                    }
+
+                    valid
+                };
+
+                // Perform an initial check of all elements. If the assignment is valid
+                // for at least one element, we do not emit any diagnostics. Otherwise,
+                // we re-run the check and emit a diagnostic on the first failing element.
```
AlexWaygood, line 3659: Your logic here looks correct to me.

The question I had when reading this was "But surely it wouldn't be valid to assign a key of type `str` to an object of type `list[int] & Unknown` -- surely we know that all subtypes of `list[int]` only accept `int` keys? But no, we don't know that; this class is fine:

```py
from typing import overload, Iterable, SupportsIndex

class F(list[int]):
    @overload
    def __setitem__(self, key: str | SupportsIndex, value: int) -> None: ...
    @overload
    def __setitem__(self, key: slice, value: Iterable[int]) -> None: ...
    def __setitem__(self, key: str | SupportsIndex | slice, value: int | Iterable[int]) -> None: ...

F()["foo"] = 42  # no diagnostic here
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3538,142 +3538,305 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            let diagnostic = builder.into_diagnostic(format_args!(
+                                "Cannot assign value of type `{assigned_d}` to key of type `{}` on TypedDict `{value_d}`",
+                                slice_ty.display(db)
+                            ));
+                            attach_original_type_info(diagnostic);
+                        }
+                    } else {
+                        if let Some(builder) = self.context.report_lint(&INVALID_KEY, slice_node) {
+                            let diagnostic = builder.into_diagnostic(format_args!(
+                                "Cannot access `{value_d}` with a key of type `{}`. Only string literals are allowed as keys on TypedDicts.",
```
AlexWaygood, line 3703: nit: rustfmt gives up formatting regions if lines in string literals are too long

```suggestion
                                "Cannot access `{value_d}` with a key of type `{}`. \
                                Only string literals are allowed as keys on TypedDicts.",
```

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Possibly_missing_`__…_(efd3f0c02e9b89e9).snap
sharkdp: Yes, absolutely. I haven't forgotten the comment, and absolutely agree with it. I just wanted to postpone this.

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
sharkdp: > What about including the name of the unknown key in the summary line here and taking it out of the primary-annotation message?

I can try that, but it's always a bit tricky to get both the full diagnostic and the concise diagnostic to look good. I still wish we had an API to set a separate concise message.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Will also move this to a follow-up.

PR #21379, opened 2025-11-11
[ty] Allow PEP-604 unions in stubs and `TYPE_CHECKING` blocks prior to 3.10

AlexWaygood:
## Summary

We currently only support implicit PEP-604 type aliases if `python-version` is set to Python 3.10+. But we should also support them on Python <=3.9 if it's a stub file, or some other suite that we know will not be executed at runtime such as an `if TYPE_CHECKING` block.

## Test Plan

Mdtests added.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -276,7 +276,7 @@ def g(
 
 In runtime contexts, `|` unions are only permitted on Python 3.10+. But in suites of code that are
 never executed at runtime (stub files, `if TYPE_CHECKING` blocks, and stringified annotations), they
-are permitted even if the target version is set to Python \<=3.9.
+are permitted even if the target version is set to Python 3.9 or earlier.
 
 ```toml
 [environment]


AlexWaygood:
The ecosystem impact looks like basically what I'd expect.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -272,6 +272,54 @@ def g(
 ): ...
 ```
 
+## `|` unions in stubs and `TYPE_CHECKING` blocks
+
+In runtime contexts, `|` unions are only permitted on Python 3.10+. But in suites of code that are
+never executed at runtime (stub files, `if TYPE_CHECKING` blocks, and stringified annotations), they
+are permitted even if the target version is set to Python \<=3.9.
```
sharkdp:
```suggestion
are permitted even if the target version is set to Python 3.9 or earlier.
```

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -272,6 +272,54 @@ def g(
...
+`foo.py`:
+
+```py
+from typing import TYPE_CHECKING
+from bar import GLOBAL_CONSTANT
+
+reveal_type(GLOBAL_CONSTANT)  # revealed: int | str
+
+if TYPE_CHECKING:
+    class ItsQuiteCloudyInManchester:
```
sharkdp, line 302: :eyes:

Are you trying to test if we are actually reviewing your code? :smile:

AlexWaygood: A man gets bored of naming everything `Foo` and `Bar`

PR #21382, opened 2025-11-11
[`pylint`] Detect subclasses of builtin exceptions (`PLW0133`)

### LINKED ISSUE #17347: [`pylint`] Extend `PLW0133` to also detect user-defined exceptions
### Summary

From the [rule's documentation](https://docs.astral.sh/ruff/rules/useless-exception-statement/#useless-exception-statement-plw0133) (PLW0133):

> Known problems
>
> This rule only detects built-in exceptions, like ValueError, and does not catch user-defined exceptions.

So currently this is caught:
```py
def foo():
    ValueError("...")
```
But not this. But I would love if it would, it would have spared me 1hr of debugging this morning :)
```py
class MyCustomError(Exception):
    ...

def foo():
    MyCustomError("...")
```

LoicRiegel:
## Summary

Closes #17347

Goal is to detect the useless exception statement not just for builtin exceptions but also custom (user defined) ones.

## Test Plan

I added test cases in the rule fixture and updated the insta snapshot.
Note that I first moved up a test case case which was at the bottom to the correct "violation category".
I wasn't sure if I should create new test cases or just insert inside those tests. I know that ideally each test case should test only one thing, but here, duplicating twice 12 test cases seemed very verbose, and actually less maintainable in the future. The drawback is that the diff in the snapshot is hard to review, sorry. But you can see that the snapshot gives 38 diagnostics, which is what we expect.

Alternatively, I also created this file for manual testing.
```py
# tmp/test_error.py

class MyException(Exception):
    ...
class MyBaseException(BaseException):
    ...
class MyValueError(ValueError):
    ...
class MyExceptionCustom(Exception):
    ...
class MyBaseExceptionCustom(BaseException):
    ...
class MyValueErrorCustom(ValueError):
    ...
class MyDeprecationWarning(DeprecationWarning):
    ...
class MyDeprecationWarningCustom(MyDeprecationWarning):
    ...
class MyExceptionGroup(ExceptionGroup):
    ...
class MyExceptionGroupCustom(MyExceptionGroup):
    ...
class MyBaseExceptionGroup(ExceptionGroup):
    ...
class MyBaseExceptionGroupCustom(MyBaseExceptionGroup):
    ...


def foo():
    Exception("...")
    BaseException("...")
    ValueError("...")
    RuntimeError("...")
    DeprecationWarning("...")
    GeneratorExit("...")
    SystemExit("...")
    ExceptionGroup("eg", [ValueError(1), TypeError(2), OSError(3), OSError(4)])
    BaseExceptionGroup("eg", [ValueError(1), TypeError(2), OSError(3), OSError(4)])
    MyException("...")
    MyBaseException("...")
    MyValueError("...")
    MyExceptionCustom("...")
    MyBaseExceptionCustom("...")
    MyValueErrorCustom("...")
    MyDeprecationWarning("...")
    MyDeprecationWarningCustom("...")
    MyExceptionGroup("...")
    MyExceptionGroupCustom("...")
    MyBaseExceptionGroup("...")
    MyBaseExceptionGroupCustom("...")

```

and you can run this to check the PR:
```sh
target/debug/ruff check tmp/test_error.py --select PLW0133 --unsafe-fixes --diff --no-cache --isolated --target-version py310
target/debug/ruff check tmp/test_error.py --select PLW0133 --unsafe-fixes --diff --no-cache --isolated --target-version py314
```

### DIFF
--- a/crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
+++ b/crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
@@ -17,8 +17,12 @@ use ruff_python_ast::PythonVersion;
 /// `ValueError("...")` on its own will have no effect (unlike
 /// `raise ValueError("...")`) and is likely a mistake.
 ///
-/// Without [preview], this rule only detects built-in exceptions, like `ValueError`, and does
+/// ## Known problems
+/// This rule only detects built-in exceptions, like `ValueError`, and does
 /// not catch user-defined exceptions.
+///
+/// In [preview], this rule will also detect user-defined exceptions, but only
+/// the ones defined in the file being checked.
 ///
 /// ## Example
 /// ```python


crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
```diff
@@ -56,7 +52,9 @@ pub(crate) fn useless_exception_statement(checker: &Checker, expr: &ast::StmtExp
         return;
     };
 
-    if is_builtin_exception(func, checker.semantic(), checker.target_version()) {
+    if is_builtin_exception(func, checker.semantic(), checker.target_version())
+        || is_custom_exception(func, checker.semantic(), checker.target_version())
```
ntBre: I think we should make this a preview change since the rule itself is stable. The ecosystem report was clear, but I feel like this could still apply to a lot more situations.

You can see [`preview.rs`](https://github.com/astral-sh/ruff/blob/c0fb235a703684bca1cc1767015bb7e19ab1c22f/crates/ruff_linter/src/preview.rs#L16-L19) for some examples.

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
```diff
@@ -2,15 +2,34 @@
 from contextlib import suppress
 
 
+class MyError(Exception):
+    ...
```
ntBre, line 6: Should we try one test case with another link in the inheritance chain like:

```py
class MySubError(MyError): ...
```

I think `any_base_class` is recursive, so we'll handle that too.

ntBre (commented):
Thank you! This looks great overall. I just think we should make it a preview change and maybe add a recursive test case (which I think your implementation will already handle well).

LoicRiegel:
Thank you for the review! No problem to do the changes, I've just been very busy with work, but I'll find the time to finish this PR in the next few days :)

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
LoicRiegel: Done, yes this was already handled well

crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
LoicRiegel: - I added a new function with the link of this PR in preview.rs
- documentation updated

⚠️ now the cargo insta snapshots don't detect the new test cases I added. They don't run with preview enabled, do they? How should I fix that ?

ntBre: I would just add a new `preview_rules` function that's a copy of `rules`:

https://github.com/astral-sh/ruff/blob/d7b2d261da5b03d211e1c19b4a49e3451454082b/crates/ruff_linter/src/rules/pylint/mod.rs#L237-L239

with preview enabled and `preview` added in the snapshot name. Here's a random example from another rule:

https://github.com/astral-sh/ruff/blob/d7b2d261da5b03d211e1c19b4a49e3451454082b/crates/ruff_linter/src/rules/flake8_bugbear/mod.rs#L97-L103

or we could use the relatively new `assert_diagnostics_diff` macro:

https://github.com/astral-sh/ruff/blob/6e7a2dce0fbccd749f4dc503a50b61bc50d88b80/crates/ruff_linter/src/rules/flake8_bandit/mod.rs#L116-L127

LoicRiegel: Thank you very much for the hint! ``assert_diagnotics_diff`` is really nice! I used and now both snapshots are good!
I also indicated in the fixture file which lines are supposed to be caught only in preview

Hope everything is alright now :)

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
```diff
@@ -2,15 +2,40 @@
...
+
+
+# Violation test cases with builtin errors: PLW0133
+
+
 # Test case 1: Useless exception statement
 def func():
     AssertionError("This is an assertion error")  # PLW0133
+    MyError("This is a custom error")  # PLW0133 (review)
```
ntBre: Should we remove these `(review)` suffixes? Happy to delete them myself, I was just curious if they were needed for some reason.

crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
```diff
@@ -16,8 +17,7 @@ use ruff_python_ast::PythonVersion;
 /// `ValueError("...")` on its own will have no effect (unlike
 /// `raise ValueError("...")`) and is likely a mistake.
 ///
-/// ## Known problems
-/// This rule only detects built-in exceptions, like `ValueError`, and does
+/// Without [preview], this rule only detects built-in exceptions, like `ValueError`, and does
```
ntBre: I think we should probably keep the `## Known problems` heading and the old text, with the new preview behavior following it. There will still be a known problem, even in preview, in that we can't resolve user-defined exceptions defined in other files.

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
LoicRiegel: I thought those comments made it a little easier when reviewing cargo insta snapshots, but I totally understand if you prefer not to have them, it'll be one less thing to worry about when making the preview behavior stable.
I did it in a specific commit on purpose, so I'll just revert it :)

crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
LoicRiegel: Oh right, it won't work if the exception is define in a different file.
Yes I'll put back the Known problem section, and maybe include in the Preview section that it only works for exceptions defined in the file being checked.

Waw this makes me think, custom exceptions are usually imported and raised in lots of files, mist of the time not the file where they are defined... This PR is still an improvement, but with limited impact... 
Maybe we should keep the issue open...

Is there a way I could continue this work and make it work for custom exceptions in different files?

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
ntBre: Ohhh, are they supposed to say `preview` instead of `review`? That does make some sense to me! I thought they were related to the code review, but it may have just been a typo.

crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
ntBre: Yeah, this is why the issue is labeled `type-inference`. That label used to be `multi-file-analysis`, but we combined them since our type checker has both. Ruff currently can't do any multi-file analysis, so this is the best we can do for now. I think that means it's okay to close the issue too.

This is still a nice improvement, as you said!

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
LoicRiegel: Oh my god yes I meant to write "preview*! So sorry for the typo 
Okay, so let me know if you want me to revert, otherwise I'll fix the typo :)

ntBre: I'd probably still lean toward reverting now that we're using `assert_diagnostics_diff`, but either works for me.

crates/ruff_linter/src/rules/pylint/rules/useless_exception_statement.rs
LoicRiegel: I updated the documentation like we said

crates/ruff_linter/resources/test/fixtures/pylint/useless_exception_statement.py
LoicRiegel: Absolutely, it's reverted :)

ntBre (approved):
Thank you!

PR #21383, opened 2025-11-11
[ty] Add more keywords to scope-based completions

BurntSushi:
This PR adds more keyword completions when asking for scoped-based
completions. This makes things like `pas<CURSOR>` return `pass`, among
other things.

We do this in a somewhat naive way. This means we might suggest
keywords even when they aren't really appropriate. I think I originally
wanted to use more context to be more precise here. And perhaps we
still should. But I think it's better to just offer these as choices
for now even if they aren't always correct.

More to the point, it seems that pylance does the same thing here.

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -1004,9 +1004,18 @@ fn is_in_definition_place(db: &dyn Db, tokens: &[Token], file: File) -> bool {
 /// This has the effect of putting all dunder attributes after "normal"
 /// attributes, and all single-underscore attributes after dunder attributes.
 fn compare_suggestions(c1: &Completion, c2: &Completion) -> Ordering {
-    fn key<'a>(completion: &'a Completion) -> (bool, bool, NameKind, bool, &'a Name) {
+    fn key<'a>(completion: &'a Completion) -> (bool, bool, bool, NameKind, bool, &'a Name) {
         (
             completion.module_name.is_some(),
+            // At time of writing (2025-11-11), keyword completions
+            // are classified as builtins, which makes them sort after
+            // everything else. But we probably want keyword completions
+            // to sort *before* anything else since they are so common.
+            // Moreover, it seems VS Code forcefully does this sorting.
+            // By doing it ourselves, we make our natural sorting match
+            // VS Code's, and thus our completion evaluation framework
+            // should be more representative of real world conditions.
+            completion.kind != Some(CompletionKind::Keyword),
             completion.builtin,
             NameKind::classify(&completion.name),
             completion.is_type_check_only,
--- a/crates/ty_completion_eval/completion-evaluation-tasks.csv
+++ b/crates/ty_completion_eval/completion-evaluation-tasks.csv
@@ -12,12 +12,12 @@ import-deprioritizes-type_check_only,main.py,3,2
 import-deprioritizes-type_check_only,main.py,4,3
 import-keyword-completion,main.py,0,1
 internal-typeshed-hidden,main.py,0,5
-none-completion,main.py,0,11
+none-completion,main.py,0,2
 numpy-array,main.py,0,
 numpy-array,main.py,1,1
 object-attr-instance-methods,main.py,0,1
 object-attr-instance-methods,main.py,1,1
-pass-keyword-completion,main.py,0,7
+pass-keyword-completion,main.py,0,1
 raise-uses-base-exception,main.py,0,2
 scope-existing-over-new-import,main.py,0,1
 scope-prioritize-closer,main.py,0,2


crates/ty_ide/src/completion.rs
```diff
@@ -170,7 +170,21 @@ impl<'db> Completion<'db> {
             kind: Some(CompletionKind::Keyword),
             module_name: None,
             import: None,
-            builtin: false,
+            builtin: true,
```
AlexWaygood: this means that they will be sorted above auto-import completions but below local-scope completions, correct?

EDIT: ah, just got to your big comment lower down in the `key()` method, which explains that this is not the case. Your reasoning there makes sense. But if these are special cased in the `key()` method, what difference does it make to mark them as "builtins"?

crates/ty_ide/src/completion.rs
```diff
@@ -294,35 +308,39 @@ pub fn completion<'db>(
...
-        });
+        completions.push(Completion::keyword_value(name, ty));
+    }
+
+    let keywords = [
+        "and", "as", "assert", "async", "await", "break", "class", "continue", "def", "del",
+        "elif", "else", "except", "false", "finally", "for", "from", "global", "if", "import",
+        "in", "is", "lambda", "nonlocal", "not", "or", "pass", "raise", "return", "try", "while",
+        "with", "yield", "case", "match", "type",
+    ];
```
AlexWaygood, line 344: we'll have to remember to add `lazy` here when we add support for Python 3.15.

`type` is also a builtin symbol -- is it going to cause problems for it to be included twice?

crates/ty_ide/src/completion.rs
```diff
@@ -4463,10 +4721,12 @@ type <CURSOR>
     /// of completions from a cursor position in source code. Most of
     /// the options involve some kind of filtering or adjustment to
     /// apply to the snapshots, depending on what one wants to test.
+    #[allow(clippy::struct_excessive_bools)] // free the bools!
```
AlexWaygood: nit

```suggestion
    #[expect(clippy::struct_excessive_bools)] // free the bools!
```

crates/ty_ide/src/completion.rs
AlexWaygood, line 1: it doesn't look like you've added any unit tests that show that we _do_ offer keyword completions in some situations. (I do see that you added some ranking integration tests, and they seem probably sufficient, just wanted to double-check that not adding any unit tests was deliberate!)

crates/ty_ide/src/completion.rs
```diff
@@ -170,7 +170,21 @@ impl<'db> Completion<'db> {
...
             module_name: None,
             import: None,
-            builtin: false,
+            builtin: true,
+            is_type_check_only: false,
+            documentation: None,
+        }
+    }
+
+    fn keyword_value(name: &str, ty: Type<'db>) -> Completion<'db> {
```
MichaReiser: Took me a while to understand what a `keyword_value` is but I also don't have a great suggestion on how to name it instead. Maybe `value_keyword` is slightly better in the sense that it specifies what kind of keyword it is (and it's not a keyword value). I'm not sure if my explanation makes sense 😅

AlexWaygood (approved):
I had a quick muck around in a local playground build, and this is much improved, thank you ❤️

It's a _tiny_ bit annoying that `raise` is now suggested above `range` for `for x in ra<CURSOR>`, but `range` is the next suggestion, and it's fairly easy to select that one instead of `raise`

crates/ty_ide/src/completion.rs
AlexWaygood: I suppose ideally here we'd also not include entries that aren't keywords on your selected Python version. E.g. `match` and `case` are not keywords on Python <=3.9, and `type` is not a keyword on Python <=3.11. (Doesn't need to be done in this PR obviously, just a possible future improvement.)

AlexWaygood: It would be _possible_ to generate the list of keywords here, ensuring that it's kept up to date with the parser:

<details>

```diff
diff --git a/Cargo.lock b/Cargo.lock
index d9f831125d..831abb6d78 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3355,6 +3355,8 @@ dependencies = [
  "serde",
  "serde_json",
  "static_assertions",
+ "strum",
+ "strum_macros",
  "unicode-ident",
  "unicode-normalization",
  "unicode_names2",
@@ -4395,6 +4397,7 @@ dependencies = [
  "rustc-hash",
  "salsa",
  "smallvec",
+ "strum",
  "tracing",
  "ty_project",
  "ty_python_semantic",
diff --git a/crates/ruff_python_parser/Cargo.toml b/crates/ruff_python_parser/Cargo.toml
index ae45871866..199534a91c 100644
--- a/crates/ruff_python_parser/Cargo.toml
+++ b/crates/ruff_python_parser/Cargo.toml
@@ -24,6 +24,8 @@ get-size2 = { workspace = true }
 memchr = { workspace = true }
 rustc-hash = { workspace = true }
 static_assertions = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
 unicode-ident = { workspace = true }
 unicode_names2 = { workspace = true }
 unicode-normalization = { workspace = true }
diff --git a/crates/ruff_python_parser/src/token.rs b/crates/ruff_python_parser/src/token.rs
index a5790a9597..06b592cf97 100644
--- a/crates/ruff_python_parser/src/token.rs
+++ b/crates/ruff_python_parser/src/token.rs
@@ -124,7 +124,18 @@ impl fmt::Debug for Token {
 }
 
 /// A kind of a token.
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord, get_size2::GetSize)]
+#[derive(
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    Debug,
+    PartialOrd,
+    Ord,
+    get_size2::GetSize,
+    strum_macros::EnumIter,
+)]
 pub enum TokenKind {
     /// Token kind for a name, commonly known as an identifier.
     Name,
@@ -613,9 +624,9 @@ impl From<Operator> for TokenKind {
     }
 }
 
-impl fmt::Display for TokenKind {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let value = match self {
+impl TokenKind {
+    pub const fn as_str(self) -> &'static str {
+        match self {
             TokenKind::Unknown => "Unknown",
             TokenKind::Newline => "newline",
             TokenKind::NonLogicalNewline => "NonLogicalNewline",
@@ -722,8 +733,13 @@ impl fmt::Display for TokenKind {
             TokenKind::Case => "`case`",
             TokenKind::With => "`with`",
             TokenKind::Yield => "`yield`",
-        };
-        f.write_str(value)
+        }
+    }
+}
+
+impl fmt::Display for TokenKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
     }
 }
 
diff --git a/crates/ty_ide/Cargo.toml b/crates/ty_ide/Cargo.toml
index 4be5d49fe2..54453a9ea8 100644
--- a/crates/ty_ide/Cargo.toml
+++ b/crates/ty_ide/Cargo.toml
@@ -34,6 +34,7 @@ regex = { workspace = true }
 rustc-hash = { workspace = true }
 salsa = { workspace = true, features = ["compact_str"] }
 smallvec = { workspace = true }
+strum = { workspace = true }
 tracing = { workspace = true }
 
 [dev-dependencies]
diff --git a/crates/ty_ide/src/completion.rs b/crates/ty_ide/src/completion.rs
index c13ec95de9..c69cde7d0d 100644
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -14,6 +14,8 @@ use ty_python_semantic::{
     types::{CycleDetector, Type},
 };
 
+use strum::IntoEnumIterator;
+
 use crate::docstring::Docstring;
 use crate::find_node::covering_node;
 use crate::goto::DefinitionsOrTargets;
@@ -330,12 +332,11 @@ fn add_keyword_completions<'db>(
         completions.push(Completion::keyword_value(name, ty));
     }
 
-    let keywords = [
-        "and", "as", "assert", "async", "await", "break", "class", "continue", "def", "del",
-        "elif", "else", "except", "false", "finally", "for", "from", "global", "if", "import",
-        "in", "is", "lambda", "nonlocal", "not", "or", "pass", "raise", "return", "try", "while",
-        "with", "yield", "case", "match", "type",
-    ];
+    let keywords: Vec<&'static str> = TokenKind::iter()
+        .filter(|token| token.is_keyword())
+        .map(TokenKind::as_str)
+        .collect();
+
     for name in keywords {
         if !query.is_match_symbol_name(name) {
             continue;
```

</details>

crates/ty_ide/src/completion.rs
BurntSushi: Sure, I'm fine with that. Done.

crates/ty_ide/src/completion.rs
BurntSushi: I guess `None`, `True` and `False` are all actually in `builtins`, so presumably those should be marked as `builtin`. But the language keywords aren't. So I updated the code to reflect that.

AlexWaygood: > I guess `None`, `True` and `False` are all actually in `builtins`, so presumably those should be marked as `builtin`.

huh, yeah, TIL:

```pycon
>>> import builtins
>>> builtins.__dict__["True"]
True
>>> builtins.__dict__["False"]
False
>>> builtins.__dict__["None"] is None
True
```

crates/ty_ide/src/completion.rs
BurntSushi: Yeah, but: this will include value keywords twice... Which is easy enough to work-around. But the string representation for some of the keywords also includes backticks, which is kind of annoying.

Anyway, I took your patch, but changed it so that the code here trips on an assertion if the number of expected keywords doesn't match the number we suggest here. It doesn't avoid duplication per se, but should at least trip if someone forgets to add a new keyword here.

crates/ty_ide/src/completion.rs
BurntSushi: Oh, there is a `keywords` test. But it was wrong because of a late change where I flipped `builtin` to `true` for keyword completions. But I swapped `builtin` back to `false` for those as per discussion above.

crates/ty_ide/src/completion.rs
AlexWaygood: > Yeah, but: this will include value keywords twice...

`type` is also included defined as a class in typeshed's `builtins.pyi` file, so I think that means that `type` will be included twice, since you've also included it in the list of keywords?

crates/ruff_python_parser/Cargo.toml
```diff
@@ -24,6 +24,8 @@ get-size2 = { workspace = true }
 memchr = { workspace = true }
 rustc-hash = { workspace = true }
 static_assertions = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
```
MichaReiser: Lol no, I sort of hate strum 😆

crates/ty_ide/src/completion.rs
```diff
@@ -294,36 +310,51 @@ pub fn completion<'db>(
...
+    // If this assert trips, it probably means that a new keyword
+    // was added to `TokenKind` but wasn't added above. Just add it
+    // to the end of `keywords` (or `keyword_values` if the keyword
+    // is also a value expression). ---AG
+    debug_assert_eq!(
+        keyword_values.len() + keywords.len(),
+        TokenKind::iter().filter(|token| token.is_keyword()).count(),
+        "Expected actual number of keyword tokens to match \
+         number of keyword completion tokens",
+    );
```
MichaReiser: I personally would probably just keep what you had. It seemed very reasonable to me. There's a risk that we forget about a token but this wouldn't lead to a catastrophic bug. We just wait for a user to report it and all is good :). But maybe that's just me disliking strum :)

If we go the `strum` approach, I'd probably go all in and iterate over all tokens and add a `is_value_keyword` to `TokenKind` instead (similar to `is_contextual_keyword`). We could then also add methods like `is_keyword_in(python_version)`.

If you want to keep your manual list, then I suggest moving this assertion into a test that asserts that every token in `TokenKind::iter` is contained within the completions.

AlexWaygood: > Hmm, I'm not sure this warants adding `strum` to `TokenKind`.

yeah, I was in two minds about even making the suggestion, since it involves adding `strum` as a dependency to the parser crate 😄

If we moved this to become a test, then I suppose it could be a test-dependency only for the parser crate.

crates/ty_ide/src/completion.rs
BurntSushi: Those are deduped. I thought I checked in VS Code and only one is shown.

crates/ty_ide/src/completion.rs
BurntSushi: OK, I'll just drop this commit entirely and keep what I had.

crates/ruff_python_parser/Cargo.toml
BurntSushi: Macro magic adding methods? Yeah I hate that too, but I feel like we have a lot of it already so I kind of just shrug my shoulders lol.

crates/ty_ide/src/completion.rs
BurntSushi: I doubled checked, and no, this is not deduped! I've fixed this by omitting `type` from keyword completions.

PR #21386, opened 2025-11-11
[ty] Silence false-positive diagnostics when using `typing.Dict` or `typing.Callable` as the second argument to `isinstance()`

AlexWaygood:
## Summary

Typeshed annotates the builtin functions `isinstance()` and `issubclass()` like so:

https://github.com/astral-sh/ruff/blob/44b0c9ebac8b63d39632233959ff3ec2694c3ef6/crates/ty_vendored/vendor/typeshed/stdlib/builtins.pyi#L3741-L3760

That leads us to currently reject code such as `isinstance({}, typing.Dict)`, since `typing.Dict` isn't an instance of `type`, an instance of `types.UnionType`, or a tuple thereof:

```pycon
>>> import typing
>>> type(typing.Dict)
<class 'typing._SpecialGenericAlias'>
```

However, you _can_ still nonetheless use it in `isinstance()` checks at runtime, due to some magic the typing module does to make these special forms behave similarly to the stdlib clases that they're aliasing in some ways:

```pycon
>>> isinstance({}, typing.Dict)
True
```

This PR implements that special case.

## Test Plan

Mdtests added

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/call/builtins.md
+++ b/crates/ty_python_semantic/resources/mdtest/call/builtins.md
@@ -163,7 +163,7 @@ def _(x: A | B, y: list[int]):
         reveal_type(isinstance(x, B))  # revealed: Literal[True]
 ```
 
-Certain special forms in the typing module are not instances of `type`, so are strictly-spekaing
+Certain special forms in the typing module are not instances of `type`, so are strictly-speaking
 disallowed as the second argument to `isinstance()` according to typeshed's annotations. However, at
 runtime they work fine as the second argument, and we implement that special case in ty:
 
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/resources/mdtest/call/builtins.md
```diff
@@ -162,3 +162,32 @@ def _(x: A | B, y: list[int]):
         reveal_type(x)  # revealed: B & ~A
         reveal_type(isinstance(x, B))  # revealed: Literal[True]
 ```
+
+Certain special forms in the typing module are not instances of `type`, so are strictly-spekaing
```
carljm:
```suggestion
Certain special forms in the typing module are not instances of `type`, so are strictly-speaking
```

AlexWaygood:
> I wonder if people are also relying on these isinstance/issubclass calls for type narrowing? I think we'd need to do something further to make that work.

Yes, I agree! But that can be its own PR ;)

crates/ty_python_semantic/resources/mdtest/call/builtins.md
AlexWaygood: it me, the professional jorunlaist

PR #21388, opened 2025-11-11
[ty] Fix goto for `float` and `complex` in type annotation positions

MichaReiser:
## Summary

I looked into fixing https://github.com/astral-sh/ty/issues/1470, specifically that `float` and `complex` are highlighted differently (as Variables instead of classes) compared to `int`, `bool`. 

The reason for the difference in highlighting is that ty (following the typing spec) expands `float` and `complex` in type annotation positions to `float | int` and `complex | float | int`, a union. 

However, this difference wasn't obvious to me as a user because clicking on` float` or `complex` jumped right to `class float` and `class complex` in `builtins.pyi` rather than showing all definitions. 

This PR fixes that which, IMO, makes the experience more consistent:

* Hovering `float` in type annotation positions shows `float | int`
* We show `float | int` within type annotations over just `float` or `int`

I think we might still want to fix issue https://github.com/astral-sh/ty/issues/1470 separately by highlighting `float` and `complex` the same as we would any other type alias.

## Test

I added a test for goto definition. I also added a test for semantic highlighting just so that it's already in place.

### DIFF
--- a/crates/ty_ide/src/semantic_tokens.rs
+++ b/crates/ty_ide/src/semantic_tokens.rs
@@ -1203,15 +1203,15 @@ result = check(None)
         "str" @ 43..46: Class
         "d" @ 51..52: Variable
         "float" @ 54..59: Variable
-        "e" @ 64..65: Variable
-        "list" @ 67..71: Class
-        "int" @ 72..75: Class
-        "f" @ 81..82: Variable
-        "list" @ 84..88: Class
-        "float" @ 89..94: Variable
-        "g" @ 100..101: Variable
-        "int" @ 103..106: Class
-        "float" @ 109..114: Variable
+        "e" @ 89..90: Variable
+        "list" @ 92..96: Class
+        "int" @ 97..100: Class
+        "f" @ 106..107: Variable
+        "list" @ 109..113: Class
+        "float" @ 114..119: Variable
+        "g" @ 150..151: Variable
+        "int" @ 153..156: Class
+        "float" @ 159..164: Variable
         "#);
     }
 
--- a/crates/ty_python_semantic/src/types/ide_support.rs
+++ b/crates/ty_python_semantic/src/types/ide_support.rs
@@ -634,13 +634,10 @@ pub fn definitions_for_name<'db>(
                 .elements(db)
                 .iter()
                 .filter_map(|ty| ty.as_nominal_instance())
-                .flat_map(|instance| {
-                    resolve_definition(
-                        db,
-                        instance.class_literal(db).definition(db),
-                        Some(name_str),
-                        ImportAliasResolution::ResolveAliases,
-                    )
+                .map(|instance| {
+                    let definition = instance.class_literal(db).definition(db);
+                    let parsed = parsed_module(db, definition.file(db));
+                    ResolvedDefinition::FileWithRange(definition.focus_range(db, &parsed.load(db)))
                 })
                 .collect();
         }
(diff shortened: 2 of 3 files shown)


crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -619,6 +619,33 @@ pub fn definitions_for_name<'db>(
...
+        // We don't know whether we're in a type annotation position, so we'll just ask `Name`'s type,
+        // which resolves to `int | float` or `int | float | complex` if `float` or `complex` is used in
+        // a type annotation position and `float` or `complex` otherwise.
+        //
+        // https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex
+        if matches!(name_str, "float" | "complex")
+            && let Some(union) = name.inferred_type(&SemanticModel::new(db, file)).as_union()
+        {
+            if is_float_or_complex_annotation(db, union, name_str) {
+                return union
```
MichaReiser: This feels very silly but I'm not sure if there's a better way of doing this.

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -619,6 +619,33 @@ pub fn definitions_for_name<'db>(
...
+        // Special cases for `float` and `complex` in type annotation positions.
+        // We don't know whether we're in a type annotation position, so we'll just ask `Name`'s type,
+        // which resolves to `int | float` or `int | float | complex` if `float` or `complex` is used in
+        // a type annotation position and `float` or `complex` otherwise.
+        //
+        // https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex
+        if matches!(name_str, "float" | "complex")
+            && let Some(union) = name.inferred_type(&SemanticModel::new(db, file)).as_union()
+        {
+            if is_float_or_complex_annotation(db, union, name_str) {
```
Gankra: could && chain these together?

Gankra (approved):
I agree this is kinda silly but I guess it works fine? Could you add a hover test too? I feel like hover and goto might want to diverge here, so it would be nice to track them both.

MichaReiser:
> I feel like hover and goto might want to diverge here, so it would be nice to track them both.

What's the reason why you think that hover should diverge? Is it because we want to show the documentation for `float` (I'm pretty sure this breaks with this change)

crates/ty_ide/src/semantic_tokens.rs
```diff
@@ -1178,6 +1178,43 @@ result = check(None)
...
+        assert_snapshot!(test.to_snapshot(&test.highlight_file()), @r#"
+        "Test" @ 7..11: Class [definition]
+        "a" @ 17..18: Variable
+        "int" @ 20..23: Class
+        "b" @ 28..29: Variable
+        "bool" @ 31..35: Class
+        "c" @ 40..41: Variable
+        "str" @ 43..46: Class
+        "d" @ 51..52: Variable
+        "float" @ 54..59: Variable
```
carljm, line 1205: Do we want to mark this (and the other cases of `float: Variable`) in the test as ideally TODOs?

carljm: It seems to me like it would be reasonable (and not hard) to unconditionally mark any union of ClassLiteral types as `Class` for purposes of semantic tokens?

Gankra:
I'm kinda 🤷‍♀️ on literally any behaviour here, I just want it written down somewhere

crates/ty_ide/src/semantic_tokens.rs
MichaReiser: While this works, it's just a work around for fixing how semantic tokens looks up symbols by name. I rather invest time in the proper fix.

PR #21390, opened 2025-11-11
[ty] make literally any import from any submodule a re-export unless renamed

Gankra:
*turning a big dial taht says "Non-standard Import Conventions" on it and constantly looking back at the audience for approval like a contestant on the price is right.*

There's no way this one is a good idea, right? Right?

But I gotta see what the impact is.

Fixes https://github.com/astral-sh/ty/issues/1486


Gankra:
WOW I am honestly just shocked how *insignificant* this change is? I figured at least it would break a lot of stuff! Ok well there's no point in shipping this if it doesn't do anything other than expose like 2 internal types that shouldn't be exposed. Hooray for science!

PR #21391, opened 2025-11-11
[ty] Support `isinstance()` and `issubclass()` narrowing when the second argument is a `typing.py` stdlib alias

AlexWaygood:
## Summary

A followup to https://github.com/astral-sh/ruff/pull/21386

## Test Plan

New mdtests added

### DIFF
--- a/crates/ty_python_semantic/src/types/narrow.rs
+++ b/crates/ty_python_semantic/src/types/narrow.rs
@@ -11,9 +11,9 @@ use crate::types::enums::{enum_member_literals, enum_metadata};
 use crate::types::function::KnownFunction;
 use crate::types::infer::infer_same_file_expression_type;
 use crate::types::{
-    ClassLiteral, ClassType, IntersectionBuilder, KnownClass, KnownInstanceType, SpecialFormType,
-    SubclassOfInner, SubclassOfType, Truthiness, Type, TypeContext, TypeVarBoundOrConstraints,
-    UnionBuilder, infer_expression_types,
+    CallableType, ClassLiteral, ClassType, IntersectionBuilder, KnownClass, KnownInstanceType,
+    SpecialFormType, SubclassOfInner, SubclassOfType, Truthiness, Type, TypeContext,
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Not a huge ecosystem impact, but I think this is worth it for consistency as much as anything else.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -229,6 +229,18 @@ impl ClassInfoConstraintFunction {
                 )
             }
 
+            // We don't have a good meta-type for `Callable`s right now,
+            // so only apply `isinstance()` narrowing, not `issubclass()`
```
carljm, line 233: This comment suggests this is a ty limitation, but at runtime `issubclass(..., typing.Callable)` just raises `TypeError`, so I don't think this is something we should support.

We should probably revise your previous PR so that we emit a diagnostic on that again?

AlexWaygood: Oh it does? Huh, I'll fix that, thanks

AlexWaygood: > at runtime `issubclass(..., typing.Callable)` just raises `TypeError`

not on my machine?

```pycon
Python 3.13.1 (main, Jan  3 2025, 12:04:03) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> import typing as t
>>> issubclass(int, t.Callable)
False
>>> issubclass(type(lambda: 42), t.Callable)
True
```

AlexWaygood: Did you literally pass in a `...` for the first argument? If so then the runtime was probably complaining about your first argument not being an instance of `type` rather than anything about the second argument 😄

carljm: Yeah my mistake, sorry!

carljm: Looks good as is. I guess in principle we could have a protocol as the meta-type of Callable...

PR #21392, opened 2025-11-11
[ty] Constraint sets compare generic callables correctly

dcreager:
Constraint sets can now track subtyping/assignability/etc of generic callables correctly. For instance:

```py
def identity[T](t: T) -> T:
    return t

constraints = ConstraintSet.always()
static_assert(constraints.implies_subtype_of(TypeOf[identity], Callable[[int], int]))
static_assert(constraints.implies_subtype_of(TypeOf[identity], Callable[[str], str]))
```

A generic callable can be considered an intersection of all of its possible specializations, and an assignability check with an intersection as the lhs side succeeds of _any_ of the intersected types satisfies the check. Put another way, if someone expects to receive any function with a signature of `(int) -> int`, we can give them `identity`.

Note that the corresponding check using `is_subtype_of` directly does not yet work, since #20093 has not yet hooked up the core typing relationship logic to use constraint sets:

```py
# These currently fail
static_assert(is_subtype_of(TypeOf[identity], Callable[[int], int]))
static_assert(is_subtype_of(TypeOf[identity], Callable[[str], str]))
```

To do this, we add a new _existential quantification_ operation on constraint sets. This takes in a list of typevars and _removes_ those typevars from the constraint set. Conceptually, we return a new constraint set that evaluates to `true` when there was _any_ assignment of the removed typevars that caused the old constraint set to evaluate to `true`.

When comparing a generic constraint set, we add its typevars to the `inferable` set, and figure out whatever constraints would allow any specialization to satisfy the check. We then use the new existential quantification operator to remove those new typevars, since the caller doesn't (and shouldn't) know anything about them.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
@@ -420,7 +420,6 @@ def unrelated[T]():
     static_assert(not constraints.implies_subtype_of(Callable[[str], int], TypeOf[identity]))
     static_assert(not constraints.implies_subtype_of(GenericIdentity[int], TypeOf[identity]))
     static_assert(not constraints.implies_subtype_of(GenericIdentity[str], TypeOf[identity]))
-
 ```
 
 The generic callable's typevar _also_ does not affect whether the subtyping check succeeds or fails!


crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
```diff
@@ -341,4 +341,79 @@ def mutually_constrained[T, U]():
...
+```py
+def identity2[T](t: T) -> T:
+    # This constraint set refers to the same typevar as the generic function types below!
+    constraints = ConstraintSet.range(bool, T, int)
+
+    static_assert(constraints.implies_subtype_of(TypeOf[identity2], Callable[[int], int]))
+    static_assert(constraints.implies_subtype_of(TypeOf[identity2], Callable[[str], str]))
+    # TODO: no error
+    # error: [static-assert-error]
+    static_assert(not constraints.implies_subtype_of(TypeOf[identity2], Callable[[str], int]))
```
carljm, line 436: This seems like a significant issue -- is there a next step / plan for how to fix this TODO?

crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
```diff
@@ -1227,6 +1232,46 @@ from ty_extensions import static_assert, is_assignable_to
...
+
+```py
+from typing import Callable
+from ty_extensions import CallableTypeOf, TypeOf, is_assignable_to, static_assert
+
+def identity[T](t: T) -> T:
+    return t
+
+static_assert(is_assignable_to(TypeOf[identity], Callable[[int], int]))
+static_assert(is_assignable_to(TypeOf[identity], Callable[[str], str]))
```
carljm, line 1250: Hmm, why do we do better in these tests than in the subtype-of tests below? I don't see any gradual forms, so I would think they should have the same results.

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of.md
```diff
@@ -2207,6 +2207,50 @@ static_assert(is_subtype_of(CallableTypeOf[overload_ab], CallableTypeOf[overload
...
+
+```py
+from typing import Callable
+from ty_extensions import CallableTypeOf, TypeOf, is_subtype_of, static_assert
+
+def identity[T](t: T) -> T:
+    return t
+
+# TODO: no error
+# error: [static-assert-error]
```
carljm, line 2229: I assume that (to the extent the constraint-set versions of these tests pass correctly) we would expect these TODOs to be fixed by https://github.com/astral-sh/ruff/pull/20093 ?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -145,10 +145,24 @@ impl<'db> BoundTypeVarInstance<'db> {
...
-            None => *self,
+    pub(crate) fn merge(&'a self, other: &'a InferableTypeVars<'a, 'db>) -> Self {
+        match (self, other) {
+            (InferableTypeVars::None, other) | (other, InferableTypeVars::None) => *other,
+            _ => InferableTypeVars::Two(self, other),
+        }
+    }
+
+    // This is not an IntoIterator implementation because I have no desire to try to name the
+    // iterator type.
```
carljm, line 156: I think probably the move would be to define your own iterator type in that case (in which case naming it is easy! `InferableTypeVarsIterator`), rather than trying to name a type composed of double-nested `Either` and boxed/chained iterators :) But I think what you've got here is just fine.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -638,15 +645,6 @@ impl<'db> Signature<'db> {
...
     ) -> ConstraintSet<'db> {
-        // The typevars in self and other should also be considered inferable when checking whether
-        // two signatures are equivalent.
-        let self_inferable =
-            (self.generic_context).map(|generic_context| generic_context.inferable_typevars(db));
-        let other_inferable =
-            (other.generic_context).map(|generic_context| generic_context.inferable_typevars(db));
-        let inferable = inferable.merge(self_inferable.as_ref());
-        let inferable = inferable.merge(other_inferable.as_ref());
-
```
carljm, line 649: Was this just wrong? Or unnecessary

I don't see any new/changed tests above regarding equivalence of callables, but apparently removing this doesn't break any tests?

carljm (approved):
Nice!

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
dcreager: We will need to alpha-rename the typevars that we add to the inferable set when checking generic signatures. I'll add a TODO

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
```diff
@@ -349,4 +349,79 @@ def mutually_constrained[T, U]():
...
+constraints = ConstraintSet.always()
+
+static_assert(constraints.implies_subtype_of(TypeOf[identity], Callable[[int], int]))
+static_assert(constraints.implies_subtype_of(TypeOf[identity], Callable[[str], str]))
+static_assert(not constraints.implies_subtype_of(TypeOf[identity], Callable[[str], int]))
+
+static_assert(constraints.implies_subtype_of(CallableTypeOf[identity], Callable[[int], int]))
+static_assert(constraints.implies_subtype_of(CallableTypeOf[identity], Callable[[str], str]))
+static_assert(not constraints.implies_subtype_of(CallableTypeOf[identity], Callable[[str], int]))
+```
```
sharkdp, line 383: I tried to see if tests like the following would pass here …
```py
type GenericIdentity[T] = Callable[[T], T]
type GenericFunction[T, R] = Callable[[T], R]

static_assert(constraints.implies_subtype_of(TypeOf[identity], GenericIdentity))
static_assert(constraints.implies_subtype_of(CallableTypeOf[identity], GenericFunction))
```
… but I get a panic due to a [failing debug assertion](https://github.com/astral-sh/ruff/blob/9269d9a36e04b4908a379ff535e0574d5f906e40/crates/ty_python_semantic/src/types/constraints.rs#L377)

crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
dcreager: It's because of these lines:

https://github.com/astral-sh/ruff/blob/9269d9a36e04b4908a379ff535e0574d5f906e40/crates/ty_python_semantic/src/types.rs#L2042-L2045

I can remove the `is_assignable` part of the test, and it does cause the subtype checks to line up with the assignability checks here. But it causes other tests to start failing, because this interacts poorly with the union specialization heuristics here:

https://github.com/astral-sh/ruff/blob/9269d9a36e04b4908a379ff535e0574d5f906e40/crates/ty_python_semantic/src/types/generics.rs#L1445-L1455

Did I mention that the current setup is very brittle? 😅 

I can add a TODO in the code and here calling out that this is a known inconsistency we're working towards removing

crates/ty_python_semantic/resources/mdtest/type_properties/is_subtype_of.md
dcreager: Exactly, that's the patch that would update the "normal" `is_assignable_to` / `is_subtype_of` calls over to using the constraint set machinery for typevars

crates/ty_python_semantic/src/types/signatures.rs
dcreager: Bad copy/pasta. I did at one point think that we only needed to force-to-inferable the typevars of one side of the callable comparison, thinking we needed that to handle a co/contravariance thing. But that's already being handled correctly by how we perform the recursive subtyping calls below, so I think it's correct to mark both sides as inferable. (This is a mumbly explanation since much of that detail has paged out of my memory at this point)

But in the world where that hunch was correct, then we wouldn't do _any_ force-to-inferable marking here, because if we only mark one side in `has_relation_to`, and `is_equivalent_to` is akin to `has_relation_to` in both directions, then neither side is always forced-to-inferable. (That inconsistency was another hint that made me think my first hunch was wrong.)

crates/ty_python_semantic/resources/mdtest/type_properties/implies_subtype_of.md
dcreager: I traced this a bit, and the issue is that [`type_alias.value_type`](https://github.com/astral-sh/ruff/blob/9269d9a36e04b4908a379ff535e0574d5f906e40/crates/ty_python_semantic/src/types.rs#L11629) is returning a dynamic callable for these type aliases. I think that's a separate issue from this PR, though I will add some logic here to make this not cause a panic!

dcreager: If you do

```py
static_assert(constraints.implies_subtype_of(TypeOf[identity], GenericIdentity[int]))
```

the test passes. `value_type` is eagerly specializing the type alias to `GenericIdentity[Unknown]`. So other than the panic, this might be the correct expected behavior?

carljm: I think it is correct/expected that a not-explicitly-specialized use of a generic type alias in a type expression is implicitly default-specialized.

dcreager: Fixed the panic and added these as additional tests

PR #21394, opened 2025-11-12
[ty] support PEP 613 type aliases

carljm:
Refs https://github.com/astral-sh/ty/issues/544

## Summary

Takes a more incremental approach to PEP 613 type alias support (vs https://github.com/astral-sh/ruff/pull/20107). Instead of eagerly inferring the RHS of a PEP 613 type alias as a type expression, infer it as a value expression, just like we do for implicit type aliases, taking advantage of the same support for e.g. unions and other type special forms.

The main reason I'm following this path instead of the one in https://github.com/astral-sh/ruff/pull/20107 is that we've realized that people do sometimes use PEP 613 type aliases as values, not just as types (because they are just a normal runtime assignment, unlike PEP 695 type aliases which create an opaque `TypeAliasType`).

This PR doesn't yet provide full support for recursive type aliases (they don't panic, but they just fall back to `Unknown` at the recursion point). This is probably post-beta work.

## Test Plan

Added mdtests.

Many new ecosystem diagnostics, probably mostly just because we understand new types in lots of places, but I'll spot-check for potential problems.

Conformance suite changes are correct.

Performance regression is also just due to understanding lots of new types; nothing we do in this PR is inherently expensive.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
@@ -46,6 +46,30 @@ except MyExc as e:
     reveal_type(e)  # revealed: Exception
 ```
 
+## Can inherit from an alias
+
+```py
+from typing import TypeAlias
+from ty_extensions import is_subtype_of, static_assert
+
+MyList: TypeAlias = list["int"]
+
+class Foo(MyList): ...
+
+static_assert(is_subtype_of(Foo, list[int]))
+```
+
+## Cannot inherit from a stringified alias
+
+```py
+from typing import TypeAlias
+
+MyList: TypeAlias = "list[int]"
+
+# error: [invalid-base] "Invalid class base with type `str`"
+class Foo(MyList): ...
+```
+
 ## Unknown type in PEP 604 union
 
 If we run into an unknown type in a PEP 604 union in the right-hand side of a PEP 613 type alias, we
--- a/crates/ty_python_semantic/src/types/infer/builder/annotation_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/annotation_expression.rs
@@ -10,6 +10,12 @@ use crate::types::{
     KnownClass, SpecialFormType, Type, TypeAndQualifiers, TypeContext, TypeQualifiers, todo_type,
 };
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum PEP613Policy {
+    Allowed,
+    Disallowed,
+}
+
 /// Annotation expressions.
 impl<'db> TypeInferenceBuilder<'db, '_> {
     /// Infer the type of an annotation expression with the given [`DeferredExpressionState`].
@@ -18,7 +24,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         annotation: &ast::Expr,
         deferred_state: DeferredExpressionState,
     ) -> TypeAndQualifiers<'db> {
-        self.infer_annotation_expression_inner(annotation, deferred_state, false)
+        self.infer_annotation_expression_inner(annotation, deferred_state, PEP613Policy::Disallowed)
     }
 
     /// Infer the type of an annotation expression with the given [`DeferredExpressionState`],
@@ -28,7 +34,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         annotation: &ast::Expr,
         deferred_state: DeferredExpressionState,
     ) -> TypeAndQualifiers<'db> {
-        self.infer_annotation_expression_inner(annotation, deferred_state, true)
+        self.infer_annotation_expression_inner(annotation, deferred_state, PEP613Policy::Allowed)
     }
 
     /// Similar to [`infer_annotation_expression`], but accepts an optional annotation expression
@@ -47,7 +53,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         &mut self,
         annotation: &ast::Expr,
         deferred_state: DeferredExpressionState,
-        allow_pep_613: bool,
+        pep_613_policy: PEP613Policy,
     ) -> TypeAndQualifiers<'db> {
         // `DeferredExpressionState::InStringAnnotation` takes precedence over other deferred states.
         // However, if it's not a stringified annotation, we must still ensure that annotation expressions
@@ -61,7 +67,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         };
 
         let previous_deferred_state = std::mem::replace(&mut self.deferred_state, state);
-        let annotation_ty = self.infer_annotation_expression_impl(annotation, allow_pep_613);
+        let annotation_ty = self.infer_annotation_expression_impl(annotation, pep_613_policy);
         self.deferred_state = previous_deferred_state;
         annotation_ty
     }
@@ -72,13 +78,13 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
     fn infer_annotation_expression_impl(
         &mut self,
         annotation: &ast::Expr,
-        allow_pep_613: bool,
+        pep_613_policy: PEP613Policy,
     ) -> TypeAndQualifiers<'db> {
         fn infer_name_or_attribute<'db>(
             ty: Type<'db>,
             annotation: &ast::Expr,
             builder: &TypeInferenceBuilder<'db, '_>,
-            allow_pep_613: bool,
+            pep_613_policy: PEP613Policy,
         ) -> TypeAndQualifiers<'db> {
             match ty {
                 Type::SpecialForm(SpecialFormType::ClassVar) => TypeAndQualifiers::new(
@@ -106,13 +112,15 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     TypeOrigin::Declared,
                     TypeQualifiers::READ_ONLY,
                 ),
-                Type::SpecialForm(SpecialFormType::TypeAlias) if allow_pep_613 => {
+                Type::SpecialForm(SpecialFormType::TypeAlias)
+                    if pep_613_policy == PEP613Policy::Allowed =>
+                {
                     TypeAndQualifiers::declared(ty)
                 }
                 // Conditional import of `typing.TypeAlias` or `typing_extensions.TypeAlias` on a
                 // Python version where the former doesn't exist.
                 Type::Union(union)
-                    if allow_pep_613
+                    if pep_613_policy == PEP613Policy::Allowed
                         && union.elements(builder.db()).iter().all(|ty| {
                             matches!(
                                 ty,
@@ -185,7 +193,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     self.infer_attribute_expression(attribute),
                     annotation,
                     self,
-                    allow_pep_613,
+                    pep_613_policy,
                 ),
                 ast::ExprContext::Invalid => TypeAndQualifiers::declared(Type::unknown()),
                 ast::ExprContext::Store | ast::ExprContext::Del => TypeAndQualifiers::declared(
@@ -198,7 +206,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     self.infer_name_expression(name),
                     annotation,
                     self,
-                    allow_pep_613,
+                    pep_613_policy,
                 ),
                 ast::ExprContext::Invalid => TypeAndQualifiers::declared(Type::unknown()),
                 ast::ExprContext::Store | ast::ExprContext::Del => TypeAndQualifiers::declared(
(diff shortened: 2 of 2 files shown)


AlexWaygood:
~~If you rebase on `main`, I think the number of new `invalid-type-form` diagnostics should go down drastically due to the temporary hack for literal-strings-in-type-expressions that David added in #21363~~

Ah, no, David's corrected me -- the final version of that PR added no such hack. @sharkdp figured out a way of implementing #21363 without it. So don't mind me 🙃

carljm:
From the ecosystem, most diagnostics look like either true positives or unrelated existing issues, exposed by now understanding a bunch of new types.

But there are quite a few `invalid-type-form` resulting from failing to understand stringified annotations in these aliases. It looks like it may not be a good idea to land this without adding some kind of support or workaround for those.

sharkdp:
> But there are quite a few `invalid-type-form` resulting from failing to understand stringified annotations in these aliases. It looks like it may not be a good idea to land this without adding some kind of support or workaround for those.

Given that this PR implements PEP 613 aliases as if they were implicit aliases, a simple first-iteration version might be to add a new `Type::LiteralString(_) => todo_type!("Stringified annotation in type alias")` branch in `Type::in_type_expression`?

MichaReiser:
> Performance regression is also just due to understanding lots of new types; nothing we do in this PR is inherently expensive.

While true, I think it's still worth looking at the performance regression to see if anything stands out. 50% is way more than even the type-of-self PR introduced and I suspect what we're seeing is some bad run-away, similar to what we've seen with pydantic the other day. I'm not suggesting that we necessearily need to fix the performance regression, but we should at least try to understand where it's coming from and if there's something that clearly stands out.

carljm:
Adding support for stringified RHS eliminated the false positives from that -- but just added a bunch more diagnostics from understanding even more type aliases!

Some major categories of new diagnostics that I'm seeing:
1. `invalid-type-form` due to seeing a type of `object` or `UnionType` in a type expression. It looks like these are mostly due to specialized generic aliases or callable types or other things we don't understand yet. Probably need to either delay this PR until after support for those lands, or find a way to silence the false positives.
2. A ton of new `invalid-key` diagnostics from pydantic, because we now understand more type aliases for unions of typed dicts, and we don't support tagged-union discrimination. (We will also need to support aliased conditional expressions to get rid of some of these, because in some places pydantic does `schema_type = schema['type']` and then performs checks on the `schema_type` variable.) These we just have to live with until we implement those features.

Trying to debug the first category is painful, because you just get this opaque error at the usage site of the type alias and have to dig back to its definition and then figure out what we didn't understand, sometimes several layers of type aliases deep. I'm a bit worried that even once we support all the "correct" forms of type aliases, users will still get confusing and unhelpful error messages if they write bad type aliases. Maybe one way we could mitigate this, at least for `TypeAlias` if not for implicit type aliases, would be to always create a known-instance union type that is valid in type expressions for any vertical-pipe operators, even if some of the components weren't valid/known.

AlexWaygood:
The pydantic benchmark is panicking FYI, so the codspeed results aren't up to date:

```
  Expected between 1 and 5000 diagnostics on project 'pydantic' but got 6334
```

carljm:
Added a Todo type for subscripting `UnionType` known instance, and added specialized handling for PEP 604 unions in a `TypeAlias`, and that looks a lot better. Still a few added `invalid-type-form` diagnostics to debug.

carljm:
> I think it's still worth looking at the performance regression to see if anything stands out. 50% is way more than even the type-of-self PR introduced and I suspect what we're seeing is some bad run-away, similar to what we've seen with pydantic the other day.

I spent some time looking at it and can't find any evidence of runaway or pathological performance. Altair is a relatively small project which checks very fast both before and after this PR. On CI it is showing 50% regression multi-threaded and 30% single-threaded; locally it shows even less regression for a single-threaded check (0.22s to 0.27s). `-vvv` traces for old and new look remarkably similar when diffed; there are just more types resolved and more diagnostics emitted in the new version. There are no slow files reported in either version.

It still appears to me that the regression here is explained mostly by resolving more types and emitting more diagnostics.

I do think that we can be somewhat more efficient about how we handle implicit and PEP 613 type aliases, especially if they are large union type aliases used in many places, but @sharkdp is already fixing this in https://github.com/astral-sh/ruff/pull/21531

So I will mark this ready for review.

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -1,8 +1,124 @@
...
+
+## Callable type in union
+
+```py
+from typing import TypeAlias, Callable
+
+MyAlias: TypeAlias = int | Callable[[str], int]
+
+def _(x: MyAlias):
+    # TODO: int | (str) -> int
```
AlexWaygood: This todo seems done?

```suggestion
```

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -1,8 +1,124 @@
...
+    reveal_type(x)  # revealed: int | ((str, /) -> int)
```
AlexWaygood: This TODO seems done?

```suggestion
    reveal_type(x)  # revealed: int | ((str, /) -> int)
```

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -46,7 +182,31 @@ f(stub.B())
...
+
+# error: [invalid-type-form]
+def _(x: TypeAlias):
+    reveal_type(x)  # revealed: Unknown
+
+# error: [invalid-type-form]
+y: list[TypeAlias] = []
+```
+
+## RHS is required
```
AlexWaygood:
```suggestion
## Right-hand side is required
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -863,6 +863,10 @@ impl<'db> Type<'db> {
             .is_some()
     }
 
+    fn is_typealias_special_form(&self) -> bool {
+        matches!(self, Type::SpecialForm(SpecialFormType::TypeAlias))
+    }
+
```
AlexWaygood, line 869: nit: this type is simple enough that it seems like it's less code overall without this helper method?

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 16abbbb191..91ec7b4585 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -863,10 +863,6 @@ impl<'db> Type<'db> {
             .is_some()
     }
 
-    fn is_typealias_special_form(&self) -> bool {
-        matches!(self, Type::SpecialForm(SpecialFormType::TypeAlias))
-    }
-
     /// Return true if this type overrides __eq__ or __ne__ methods
     fn overrides_equality(&self, db: &'db dyn Db) -> bool {
         let check_dunder = |dunder_name, allowed_return_value| {
diff --git a/crates/ty_python_semantic/src/types/infer.rs b/crates/ty_python_semantic/src/types/infer.rs
index 94d2ca6115..03a65ccbc3 100644
--- a/crates/ty_python_semantic/src/types/infer.rs
+++ b/crates/ty_python_semantic/src/types/infer.rs
@@ -379,8 +379,7 @@ impl<'db> TypeContext<'db> {
     }
 
     pub(crate) fn is_typealias(&self) -> bool {
-        self.annotation
-            .is_some_and(|ty| ty.is_typealias_special_form())
+        self.annotation == Some(Type::SpecialForm(SpecialFormType::TypeAlias))
     }
 }
 
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index 56c5f421c9..e8bfa4b9a7 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -5461,7 +5461,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
         // Check if this is a PEP 613 `TypeAlias`. (This must come below the SpecialForm handling
         // immediately below, since that can overwrite the type to be `TypeAlias`.)
-        let is_pep_613_type_alias = declared.inner_type().is_typealias_special_form();
+        let is_pep_613_type_alias =
+            declared.inner_type() == Type::SpecialForm(SpecialFormType::TypeAlias);
 
         // Handle various singletons.
         if let Some(name_expr) = target.as_name_expr() {
```

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -1,8 +1,124 @@
...
+`main.py`:
+
+```py
+from alias import MyAlias
+
+def _(x: MyAlias):
+    reveal_type(x)  # revealed: int | str
+```
+
+## String literal in RHS
```
AlexWaygood:
```suggestion
## String literal on the right-hand side
```

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -1,8 +1,124 @@
...
+```py
+from typing import TypeAlias
+
+MyExc: TypeAlias = Exception
+
+try:
+    raise MyExc("error")
+except MyExc as e:
+    reveal_type(e)  # revealed: Exception
+```
```
AlexWaygood, line 47: could you also add a test that makes sure that we _disallow_ this if it's a string literal on the right-hand side? E.g. this will obviously pass at runtime:

```py
from typing import TypeAlias

MyList: TypeAlias = list["int"]

class Foo(MyList): ...
reveal_mro(Foo)
```

but this obviously won't:

```py
from typing import TypeAlias

MyList: TypeAlias = "list[int]"

class Foo(MyList): ...
reveal_mro(Foo)
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -9246,6 +9281,44 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+        if *op != ast::Operator::BitOr {
+            if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, node) {
+                let mut diag = builder.into_diagnostic(format_args!(
+                    "Invalid binary operator `{}` in type alias",
+                    op.as_str()
+                ));
+                diag.info("Did you mean to use `|`?");
+            }
+            return Type::unknown();
+        }
```
AlexWaygood, line 9304: have you added any tests for this new diagnostic? It looks like we now emit both `unsupported-operator` and `invalid-type-form` if you use `+` in a type expression, which feels a bit excessive?

image

If we keep the new diagnostic, I also think the subdiagnostic would probably be better as a `help` subdiagnostic rather than an `info` one:

```suggestion
        if *op != ast::Operator::BitOr {
            if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, node) {
                let mut diag = builder.into_diagnostic(format_args!(
                    "Invalid binary operator `{}` in type alias",
                    op.as_str()
                ));
                diag.help("Did you mean to use `|`?");
            }
            return Type::unknown();
        }
```

Though I'm not sure it really is obvious that the user meant to use `|` if they have something like `int + str` or `int - str` TBH. For the first of these, I have no idea _what_ they meant; for the second, my best guess would be that they're new to Python typing and they're trying to express a negation type somehow. This subdiagnostic really feels more suitable for something like `int or str` in a type expression? For that case, I think it really is pretty obvious that they meant to use `|`

crates/ty_python_semantic/src/types/infer/builder/annotation_expression.rs
```diff
@@ -17,6 +17,37 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
...
+    }
+
+    /// Infer the type of an annotation expression with the given [`DeferredExpressionState`],
+    /// allowing a PEP 613 `typing.TypeAlias` annotation.
+    pub(super) fn infer_annotation_expression_allow_pep_613(
+        &mut self,
+        annotation: &ast::Expr,
+        deferred_state: DeferredExpressionState,
+    ) -> TypeAndQualifiers<'db> {
+        self.infer_annotation_expression_inner(annotation, deferred_state, true)
```
AlexWaygood: blegh, I hate having functions that take boolean arguments in Rust 😆 unless you open the code up in an IDE, it makes callsites impossible to read -- you have to navigate to the function signature to figure out which parameter is being passed `true` here, and what that implies. I'd create a custom enum here:

```rs
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
enum PEP613Policy {
    Allowed,
    Disallowed,
}

crates/ty_python_semantic/src/types.rs
carljm: There's barely a difference in overall code size either way, but I don't think that's what the decision should be based on. If the representation changes, there's only one place we have to change it. The explicit version isn't super complex but still reads quite repetitively (three occurrences of `Type`, two of `SpecialForm`).

I agree only two usage sites makes this marginal; there were more before I added `TypeContext::is_typealias` to wrap it. But it doesn't seem like an improvement to me to remove it.

crates/ty_python_semantic/src/types/infer/builder.rs
carljm: I think in the interests of getting this PR landed, I will just remove this diagnostic. The value seems likely low, and it would be better designed in a dedicated effort.

MichaReiser:
@carljm would you mind updating https://github.com/astral-sh/ty/issues/445, now that type aliases are supported?

PR #21395, opened 2025-11-12
[`flake8-simplify`] Apply `SIM113` when index variable is of type `int`

### LINKED ISSUE #21393: SIM113 only applies when the index variable is initialized to `0`
### Summary

[`enumerate-for-loop` (SIM113)](https://docs.astral.sh/ruff/rules/enumerate-for-loop/) applies only when the index variable is initialized to `0`, but `enumerate` takes an optional argument specifying its starting point, so SIM113 should only require that the variable start as a strict instance of `int`. For reference, `unnecessary_cast_to_int::call_applicability` and `ResolvedPythonType::from` include various ways to determine whether an expression is an `int`. [Example](https://play.ruff.rs/35196ea9-0532-4208-9fca-f618c3424080):
```console
$ cat >sim113.py <<'# EOF'
fruits = ["apple", "banana", "cherry"]
i = 1
for fruit in fruits:
    print(f"{i}. {fruit}")
    i += 1
# EOF

$ python sim113.py
1. apple
2. banana
3. cherry

$ ruff --isolated check sim113.py --select SIM113
All checks passed!
```

In the spirit of SIM113, that code should be:
```python
fruits = ["apple", "banana", "cherry"]
for i, fruit in enumerate(fruits, 1):
    print(f"{i}. {fruit}")
```

### Version

njhearp:
## Summary

Fixes #21393

Now the rule checks if the index variable is initialized as an `int` type rather than only flagging if the index variable is initialized to `0`. I used `ResolvedPythonType` to check if the index variable is an `int` type.

## Test Plan

Updated snapshot test for `SIM113`.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_simplify/rules/enumerate_for_loop.rs
+++ b/crates/ruff_linter/src/rules/flake8_simplify/rules/enumerate_for_loop.rs
@@ -13,7 +13,8 @@ use crate::checkers::ast::Checker;
 /// Checks for `for` loops with explicit loop-index variables that can be replaced
 /// with `enumerate()`.
 ///
-/// In [preview], this rule checks for index variable initialized as an instance of `int`.
+/// In [preview], this rule checks for index variables initialized with any integer rather than only
+/// a literal zero.
 ///
 /// ## Why is this bad?
 /// When iterating over a sequence, it's often desirable to keep track of the


dscorbett:
SIM113’s recommendation is only safe when the index variable’s initial value is a strict instance of `int`, whereas something annotated `int` could be an instance of e.g. `bool` or a custom subclass, which would make the recommendation unsafe because the program’s behavior could change.

ntBre:
Ah, thanks for the clarification! Disregard my last paragraph then, but I'd still lean toward making it a preview change.

njhearp:
Thanks for the feedback! I updated the rule to be gated behind preview for strict instances of `int`. I also updated the documentation and comments. It looks like some actions failed due to timeouts, so I suspect some GitHub services are temporarily down and they will pass later.

crates/ruff_linter/src/rules/flake8_simplify/rules/enumerate_for_loop.rs
```diff
@@ -11,6 +13,8 @@ use crate::checkers::ast::Checker;
 /// Checks for `for` loops with explicit loop-index variables that can be replaced
 /// with `enumerate()`.
 ///
+/// In [preview], this rule checks for index variable initialized as an instance of `int`.
```
ntBre:
```suggestion
/// In [preview], this rule checks for index variables initialized with any integer rather than only
/// a literal zero.
```

Having to say this is making me rethink the preview suggestion 😅 But I guess it's better to be safe. Thanks for updating the docs!

PR #21398, opened 2025-11-12
[ty] Better assertion message for benchmark diagnostics check

sharkdp:
I don't know why, but it always takes me an eternity to find the failing project name a few lines below in the output. So I'm suggesting we just add the project name to the assertion message.

### DIFF
--- a/crates/ruff_benchmark/benches/ty_walltime.rs
+++ b/crates/ruff_benchmark/benches/ty_walltime.rs
@@ -71,16 +71,13 @@ impl Display for Benchmark<'_> {
     }
 }
 
-fn check_project(db: &ProjectDatabase, max_diagnostics: usize) {
+fn check_project(db: &ProjectDatabase, project_name: &str, max_diagnostics: usize) {
     let result = db.check();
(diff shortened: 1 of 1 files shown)


crates/ruff_benchmark/benches/ty_walltime.rs
```diff
@@ -71,16 +71,13 @@ impl Display for Benchmark<'_> {
     }
 }
 
-fn check_project(db: &ProjectDatabase, max_diagnostics: usize) {
+fn check_project(db: &ProjectDatabase, project_name: &str, max_diagnostics: usize) {
```
sharkdp, line 74: Was thinking about using `db.project().name(db)` instead of passing in the `project_name`...

MichaReiser: but? (not that I mind much either way)

sharkdp: My guess was that this would have taken the actual project name from the `pyproject.toml`, which might be different from "our" name for the benchmark?

PR #21399, opened 2025-11-12
[ty] Improve semantic token classification for names

MichaReiser:
## Summary

This PR addresses the first TODO in https://github.com/astral-sh/ty/issues/791

> Improve classification for name tokens that are imported from other modules. Currently, these are classified based on their types, which often means they're classified as variables when they should be classes, etc.

The semantic token provider was written before we added the more involved `definitions_for_name` that resolves definitions across modules. This PR changes the semantic token provider to use `definitions_for_name` instead of rolling its own, very limited definition lookup (that only performed a lookup in the same scope).

This should fix many miss classifications, like that `float` was classified as a `Variable` instead of a `Class`. 

This PR further fixes some more misclasification:

* Classify `self` in `self.x` within a method as `SelfParamter`
* Set the `Definition` modifier for assignments: `x` in `x = 10` is a definition
* Classify legacy type aliases as `Class`es
* Classify names referencing a parameter as `Parameter`

## Test plan

Added tests, I compared all classification changes with what Pylance does

### DIFF
--- a/crates/ty_ide/src/semantic_tokens.rs
+++ b/crates/ty_ide/src/semantic_tokens.rs
@@ -1,3 +1,37 @@
+//! This module walks the AST and collects a set of "semantic tokens" for a file
+//! or a range within a file. Each semantic token provides a "token type" and zero
+//! or more "modifiers". This information can be used by an editor to provide
+//! color coding based on semantic meaning.
+//!
+//! Visual Studio has a very useful debugger that allows you to inspect the
+//! semantic tokens for any given position in the code. Not only is this useful
+//! to debug our semantic highlighting, it also allows easy comparison with
+//! how Pylance (or other LSPs) highlight a certain token. You can open the scope inspector,
+//! with the Command Palette (Command/Ctrl+Shift+P), then select the
+//!  `Developer: Inspect Editor Tokens and Scopes` command.
+//!
+//! Current limitations and areas for future improvement:
+//!
+//! TODO: Need to handle semantic tokens within quoted annotations.
+//!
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -633,11 +618,11 @@ pub fn definitions_for_name<'db>(
             return union
                 .elements(db)
                 .iter()
+                .rev()
                 .filter_map(|ty| ty.as_nominal_instance())
                 .map(|instance| {
                     let definition = instance.class_literal(db).definition(db);
-                    let parsed = parsed_module(db, definition.file(db));
-                    ResolvedDefinition::FileWithRange(definition.focus_range(db, &parsed.load(db)))
```
MichaReiser, line 640: I realized right after landing the *go to definition for float and complex* PR that using `FileWithRange` here is a stupid idea. I switched to using `FileWithRange` to suppress the docstring (we always pick the first docstring which ended up to be `int` which is rather confusing). However, returning `FileWithRange` here breaks semantic highlighting because a `FileWithRange` has no definition.

That's why I opted in this PR to keep using `Definition` but return the definition for `float` first (or `complex`) so that we show the docstring of `float`. 

This is not 100% correct because `float` is a union in this position, but it also doesn't feel entirely wrong (and sort of what I'd expect as a user).

Gankra: Could you add a comment to the `.rev()` explaining that it makes `float` / `complex` come first in their own listings?

PR #21401, opened 2025-11-12
[ty] Ensure annotation/type expressions in stub files are always deferred

AlexWaygood:
## Summary

Inference of type expressions should always be deferred when analysing a stub file: stubs are allowed to (and often do) do things like this:

```py
from typing import TypeVar

T = TypeVar("T", bound=Foo)

class Foo: ...
```

`bound=Foo` here would cause a `NameError` if you tried to run this snippet with an interpreter at runtime, since `Foo` is not yet defined at the point when the `TypeVar` is assigned. But it's nonetheless permitted in a stub file, because stub files are never intended to be executed at runtime. They're just "data files" for the type checker.

This pattern is used in quite a few typeshed stub files. Among other things, the fact that we don't yet support it means that we aren't able to correctly infer the type of the `default=` argument for this `TypeVar` (`int` has not yet been defined at that point in `builtins.pyi`):

https://github.com/astral-sh/ruff/blob/9dd666d67758c9d25a361c7c23d7f8b883b0b05a/crates/ty_vendored/vendor/typeshed/stdlib/builtins.pyi#L93

And therefore we think that a bare `memoryview` annotation should signify `memoryview[Unknown]`, whereas in fact it should signify `memoryview[int]`.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
@@ -396,3 +396,34 @@ B = NewType("B", list[Any])
 T = TypeVar("T")
 C = NewType("C", list[T])  # TODO: should be "error: [invalid-newtype]"
 ```
+
+## Forward references in stub files
+
+Stubs natively support forward references, so patterns that would raise `NameError` at runtime are
+allowed in stub files:
+
+`stub.pyi`:
+
+```pyi
+from typing import NewType
+
+N = NewType("N", A)
+
+class A: ...
+```
+
+`main.py`:
+
+```py
+from stub import N, A
+
+n = N(A())  # fine
+
+def f(x: A): ...
+
+f(n)  # fine
+
+class Invalid: ...
+
+bad = N(Invalid())  # error: [invalid-argument-type]
+```
--- a/crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/legacy/variables.md
@@ -266,7 +266,48 @@ from typing import TypeVar
 
 # error: [invalid-legacy-type-variable]
 T = TypeVar("T", invalid_keyword=True)
+```
 
+### Forward references in stubs
+
+Stubs natively support forward references, so patterns that would raise `NameError` at runtime are
+allowed in stub files:
+
+`stub.pyi`:
+
+```pyi
+from typing import TypeVar
+
+T = TypeVar("T", bound=A, default=B)
+U = TypeVar("U", C, D)
+
+class A: ...
+class B(A): ...
+class C: ...
+class D: ...
+
+def f(x: T) -> T: ...
+def g(x: U) -> U: ...
+```
+
+`main.py`:
+
+```py
+from stub import f, g, A, B, C, D
+
+reveal_type(f(A()))  # revealed: A
+reveal_type(f(B()))  # revealed: B
+reveal_type(g(C()))  # revealed: C
+reveal_type(g(D()))  # revealed: D
+
+# TODO: one diagnostic would probably be sufficient here...?
+#
+# error: [invalid-argument-type] "Argument type `C` does not satisfy upper bound `A` of type variable `T`"
+# error: [invalid-argument-type] "Argument to function `f` is incorrect: Expected `B`, found `C`"
+reveal_type(f(C()))  # revealed: B
+
+# error: [invalid-argument-type]
+reveal_type(g(A()))  # revealed: Unknown
 ```
 
 ### Constructor signature versioning
--- a/crates/ty_python_semantic/resources/mdtest/paramspec.md
+++ b/crates/ty_python_semantic/resources/mdtest/paramspec.md
@@ -102,6 +102,20 @@ Other values are invalid.
 P4 = ParamSpec("P4", default=int)
 ```
 
+### Forward references in stub files
+
+Stubs natively support forward references, so patterns that would raise `NameError` at runtime are
+allowed in stub files:
+
+```pyi
+from typing_extensions import ParamSpec
+
+P = ParamSpec("P", default=[A, B])
+
+class A: ...
+class B: ...
+```
+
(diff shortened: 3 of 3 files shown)


AlexWaygood:
This new `dedupe` diagnostic is very weird:

```diff
+ dedupe/convenience.py:88:9: error[invalid-assignment] Object of type `property` is not assignable to `int`
```

Here's a minimal repro:

````md

`stub.pyi`:

```pyi
from typing_extensions import TypeVar, Generic, Any

_IntegerT_co = TypeVar("_IntegerT_co", bound=integer)

class integer: ...

class iinfo(Generic[_IntegerT_co]):
    @property
    def max(self) -> int: ...
    def __new__(cls, dtype: _IntegerT_co | Any) -> iinfo[_IntegerT_co]: ...
```

`foo.py`:

```py
# This should be `int`!
reveal_type(iinfo("int").max)  # revealed: property
```
````

Hmmm, okay, this is reproducible on `main`, without even any stubs or stringized annotations: https://play.ty.dev/ecc94b2c-e353-4733-bfd9-bf457251324b

**Comment by AlexWaygood** [comment]
All the new `jax` diagnostics, and many of the new `scipy`/`pandas` diagnostics, appear to be due to the same pre-existing bug as the `dedupe` diagnostic.

**Comment by AlexWaygood** [comment]
## Ecosystem analysis

Overall this gets rid of a lot of `unresolved-reference` and `unresolved-attribute` errors; it appears to significantly improve our understanding of types from stub files and reduce the number of variables inferred as `Unknown`.

There are lots of new `no-matching-overload` diagnostics on scipy, e.g.

```diff
+ scipy/interpolate/tests/test_bary_rational.py:173:42: error[no-matching-overload] No overload of function `assert_allclose` matches arguments
```

These appear due to a bug in numpy's stubs. That `scipy` call is [here](https://github.com/scipy/scipy/blob/aa4eddcab554bdd41320952c5eb3050989bb9ca0/scipy/interpolate/tests/test_bary_rational.py#L173), and we can see that it passes `atol=TOL` to `assert_allclose`. Adding a `reveal_type` call to that line indicates that we infer the type of `TOL` to be `floating[Any] | float64`. That seems to be accurate:

```
% uvx --with=scipy --with=pytest python3.14
Installed 7 packages in 35ms
Python 3.14.0 (main, Oct 10 2025, 12:54:13) [Clang 20.1.4 ] on darwin
>>> from scipy.interpolate.tests.test_bary_rational import TOL
>>> type(TOL)
<class 'numpy.float64'>
```

But none of [numpy's overloads for `assert_allclose`](https://github.com/numpy/numpy/blob/5d4dc2ab393e4e4d4cd0a2b77e3fe9f84487e409/numpy/testing/_private/utils.pyi#L325-L348) permit a type of `float64` or `floating` to be passed to the `atol` or `rtol` parameters; all overloads for the function use `float` for those parameters.

AlexWaygood:
I believe the performance regression is mostly due to the fact that we now accurately understand a lot more typeshed types that use forward references. Note that the regression is most significant on microbenchmarks and tomllib, i.e., the smaller codebases in which inferring typeshed types will take up a greater proportion of the overall type-checking time.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -20,6 +20,18 @@ use crate::types::{
...
+        // However, if it's not a stringified annotation, we must still ensure that annotation expressions
+        // are always deferred in stub files.
+        match self.deferred_state {
+            DeferredExpressionState::None => {
+                if self.in_stub() {
+                    self.deferred_state = DeferredExpressionState::Deferred;
+                }
+            }
+            DeferredExpressionState::InStringAnnotation(_) | DeferredExpressionState::Deferred => {}
+        }
```
carljm, line 33: In general, setting the deferred state without a paired revert back to the previous deferred state is not right; it will cause the new deferred state to leak back up the stack, outside of this type expression inference.

I feel like there might be some refactor called for here between this method and `infer_type_expression_with_state`, but I haven't quite sorted yet what it should be...

carljm (approved):
Thank you!

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
AlexWaygood: > In general, setting the deferred state without a paired revert back to the previous deferred state is not right; it will cause the new deferred state to leak back up the stack, outside of this type expression inference.

Yes, but it can never leak so far as leak into a different file, correct? And when you're inferring types for one scope in a stub file, you know that all other scopes in that stub file will also be in a stub file. So I think this is fine for this specific scenario.

My first draft of this PR just switched a bunch of callsites from `self.infer_type_expression(expr)` to `self.infer_type_expression_with_state(expr, state)` where `state` was `DeferredExpressionState::Deferred` if we were in a stub file. But that doesn't feel like a robust way of doing it at all, because you have to remember to do that every time, and there were _loads_ of callsites where we were forgetting to do that.

AlexWaygood: > I feel like there might be some refactor called for here between this method and `infer_type_expression_with_state`, but I haven't quite sorted yet what it should be...

Yeah, I also looked briefly at whether a refactor would be possible, but it seemed nontrivial...

carljm: > Yes, but it can never leak so far as leak into a different file, correct? And when you're inferring types for one scope in a stub file, you know that all other scopes in that stub file will also be in a stub file. So I think this is fine for this specific scenario.

I don't think it's fine, because the updated `DeferredExpressionState::Deferred` that is set here can leak into non-type-expressions in that stub file. And (at least in this PR) it is not the intent that _all_ name references in stub files are deferred, only those in type expressions.

AlexWaygood: Oh, good point. Argh, that seems serious enough that I maybe wouldn't have approved if I'd been reviewing 😆 Will fix in a followup!

PR #21405, opened 2025-11-12
[ty] remove erroneous canonicalize

Gankra:
Alternative implementation to https://github.com/astral-sh/ruff/pull/21052

### DIFF
--- a/crates/ty/tests/cli/python_environment.rs
+++ b/crates/ty/tests/cli/python_environment.rs
@@ -465,105 +465,85 @@ import colorama
 
     assert_cmd_snapshot!(case.command()
         .current_dir(case.root().join("project"))
-        .arg("--python").arg(case.root().join("opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/bin/python3")), @r#"
+        .arg("--python").arg(case.root().join("opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/bin/python3")), @r"
     success: false
-    exit_code: 101
+    exit_code: 1
     ----- stdout -----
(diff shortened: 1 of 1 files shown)


Gankra:
I'm claiming this is erroneous based on the fact that this canonicalize appears from seemingly nowhere in this commit:

https://github.com/astral-sh/ruff/commit/ddd4bab67c6d94eea312e4951fcc2c7bc8299fa4

And removing it seems to fix the homebrew crasher without breaking anything else. I otherwise have no good intuition about it.

cc @BurntSushi

Gankra:
(diagnostic change is the flake that's going around... memory usage change might be Why burntsushi made this change)

MichaReiser (approved):
Removing the `canonicalize` here looks correct to me. If it's needed for correctness (for whatever reason), than it should be moved here instead:

https://github.com/astral-sh/ruff/blob/376571eed196e05f42e271fbe8357e96fafa5bd3/crates/ty_python_semantic/src/module_resolver/resolver.rs#L289-L292


But I'm inclined to just ship this as is

Gankra:
Do we have any reasonable way to test this..?

MichaReiser:
You could add a CLI test. We also have some file watching tests exercising symlinks. 

https://github.com/astral-sh/ruff/blob/376571eed196e05f42e271fbe8357e96fafa5bd3/crates/ty/tests/file_watching.rs#L1395

Gankra:
Awesome thank you micha!

That was *frustratingly* hard to get the right minimal conditions for the thing to reproduce in tests but I got it working, and this fix indeed fixes it.

BurntSushi:
Thanks! I don't remember why I added this. I'm usually pretty skeptical of `canonicalize`.

Issue #10102, opened 2024-02-23
Support Pycharm via creating Astral Plugins 


Pixel-Minions:
Hi,

I was wondering if there are any plans for creating plugins for Pycharm for the formatter and linter Ruff.

Most of the people I have ever worked use Pycharm as the main driver for dev and we find ourselves using the terminal within to use Ruff.

best,

MichaReiser:
Hey. 

PyCharm support is something that's on our minds and we agree that it's important. For now, you can give the unofficial pycharm plugin a try https://docs.astral.sh/ruff/integrations/#pycharm-unofficial

Pixel-Minions:
Thank you for your insights, @MichaReiser. We've been utilizing the plugin you recommended, primarily due to the lack of alternatives. However, we've encountered some challenges with it. Specifically, the plugin doesn't seem to perform linting as effectively as needed. We've noticed that certain linting issues aren't identified during its use, leading to unexpected feedback from our CI/CD pipeline. To address this, we're considering the implementation of pre-commit hooks as a means to automate and enhance the linting process.

MichaReiser:
@Pixel-Minions, have you tried creating an issue in https://github.com/koxudaxi/ruff-pycharm-plugin to see if your problem can get fixed or is there already an existing issue?

anmolratn-qure:
PyCharm is one of the most used Python IDEs, specially in a professional setting. A lot of people at my company use PyCharm but find ourselves running ruff from the command line. I believe this should be taken as a priority. The unofficial plugin has linting bugs and sometimes work unexpectedly.

MichaReiser:
@snowsignal plans to look into this once we've rewritten our LSP in rust.

zanieb:
We are continuing to consider the best path forward here and discussing with both the JetBrains team and the author of the existing plugin. We're very interested in providing a better experience in PyCharm and we'll definitely be dedicating resources to it. I think most of the remaining uncertainty is how to use our LSP to provide the same experience in both paid and free PyCharm versions.

Kaelten:
Please consider prioritizing this. 

The plugin maintained by @koxudaxi routinely get's confused about what version of ruff it should be using in a given project and most recently has started just erroring all the time and trying to use a version of ruff from a project that's not even open.

This is creating a horrible development experience and there's many of us that have over a decade of experience with Pycharm and moving to VSCode is a huge ask.

koxudaxi:
[...]

PR #21409, opened 2025-11-12
Update PyCharm setup instructions

charliecloudberry:
## Summary

Ruff support was recently added to PyCharm and I want to update Ruff's documentation accordingly.

## Test Plan

n/a

### DIFF
--- a/docs/editors/setup.md
+++ b/docs/editors/setup.md
@@ -361,6 +361,22 @@ and [LSP-ruff](https://github.com/sublimelsp/LSP-ruff) package.
 
 ## PyCharm
 
+Starting with version 2025.3, PyCharm supports Ruff out of the box:
+
+1. Go to **Python | Tools | Ruff** in the Settings dialog.
+
+1. Select the **Enable** checkbox.
+
+1. In the Execution mode setting, select how PyCharm should search for the executable:
+
+    **Interpreter** mode: PyCharm searches for an executable installed in your interpreter. To install the Ruff package for the selected interpreter, click _Install Ruff_.
+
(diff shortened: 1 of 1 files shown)


ntBre:
My understanding based on https://github.com/astral-sh/ruff/issues/10102#issuecomment-3488551126 was that the Ruff support was still in preview, and this seems to be supported by the warning at the top of the [docs](https://www.jetbrains.com/help/pycharm/2025.3/lsp-tools.html#ruff) you linked:

> Because PyCharm 2025.3 is still in development, this documentation may not be entirely accurate and is subject to change.

Should we hold off until the Ruff support is stabilized, or do most users have access to the preview features?

charliecloudberry:
@ntBre Hi! Ruff support is already available in the EAP version of PyCharm (anyone can install and use this version if they want to), and the major release is planned before the end of 2025. 
Also I was not sure how long it would take for the changes to be merged and published, so I decided to submit the PR now.
If you think we should wait, I can submit my changes later.

ntBre:
Ah okay, I wasn't sure exactly how preview worked, thanks!

And thank you for opening this by the way, we're excited for the official support too!

If we merge this today it would go out in the weekly patch release tomorrow, though, so I would probably lean toward waiting or at least keeping the old setup instructions too until stabilization.

But I'm curious what others think too (cc @MichaReiser)

MichaReiser:
Nice! I agree with @ntBre. I'd prefer to keep the old documentation around for now and mention that the builtin Ruff integration is only available with Pycharm verison X or newer

charliecloudberry:
Hi! Thank you for the review and suggestions. I’ve restored the old documentation and updated the text to specify the PyCharm version.
I’ll also submit a new PR to update the link to the PyCharm docs after the major release.

MichaReiser:
Thank you

PR #21411, opened 2025-11-12
[ty] Further improve subscript assignment diagnostics

sharkdp:
## Summary

Further improve subscript assignment diagnostics, especially for `dict`s:

```py
config: dict[str, int] = {}

config["retries"] = "three"
```

image

I have many more ideas, but this looks like a reasonable first step. Thank you @AlexWaygood for some of the suggestions here.

## Test Plan

Update tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_No_`__setitem__`_met…_(468f62a3bdd1d60c).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_No_`__setitem__`_met…_(468f62a3bdd1d60c).snap
@@ -30,7 +30,7 @@ error[invalid-assignment]: Cannot assign to a subscript on an object of type `Re
 6 | config["retries"] = 3  # error: [invalid-assignment]
   | ^^^^^^^^^^^^^^^^^
   |
-info: Consider adding a `__setitem__` method to `ReadOnlyDict`.
+help: Consider adding a `__setitem__` method to `ReadOnlyDict`.
 info: rule `invalid-assignment` is enabled by default
 
 ```
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -3884,7 +3884,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                         object_ty.display(db),
                                     ));
                                 } else {
-                                    diagnostic.info(format_args!(
+                                    diagnostic.help(format_args!(
                                         "Consider adding a `__setitem__` method to `{}`.",
                                         object_ty.display(db),
                                     ));


sharkdp:
~~Hm, there seems to be a bug in ecosystem-analyzer's diagnostic parsing. This should only show changed diagnostics.  Instead, some changes are shown as a "changed" + a "new" diagnostic. I guess it happens if there were two diagnostics of type A previously, and we now have two new diagnostics of type B.~~  Fixed

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Invalid_key_type_(d3d47de65fb3bad).snap
```diff
@@ -19,12 +19,14 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
 # Diagnostics
 
 ```
-error[invalid-assignment]: Method `__setitem__` of type `bound method dict[str, int].__setitem__(key: str, value: int, /) -> None` cannot be called with a key of type `Literal[0]` and a value of type `Literal[3]` on object of type `dict[str, int]`
+error[invalid-assignment]: Cannot assign to `dict[str, int]` with key of type `Literal[0]` and a value of type `Literal[3]`
```
carljm: I notice that in the "not subscriptable" case we take care to say "Cannot assign to a subscript on object of type ...", but here we shorthand it to "Cannot assign to ...".

In general we've taken care in our diagnostics to say "object of type ..." rather than just conflating types and inhabitants, though tbh I'm not clear if that's worth the extra words.

I do think it's a bit odd to say "Cannot assign to `dict[str, int]`" when what we mean is "Cannot assign to subscript of `dict[str, int]`". But I also don't feel strongly if you prefer the concision.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3824,13 +3865,25 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-                                    "Cannot assign to a subscript on an object of type `{}` with no `__setitem__` method",
+                                let mut diagnostic = builder.into_diagnostic(format_args!(
+                                    "Cannot assign to a subscript on an object of type `{}`",
                                     object_ty.display(db),
                                 ));
-                                attach_original_type_info(diagnostic);
+                                attach_original_type_info(&mut diagnostic);
+
+                                // Use `KnownClass` as a crude proxy for checking if this is not a user-defined class. Otherwise,
+                                // we end up suggesting things like "`None` may be missing a `__setitem__` method."
```
carljm: I don't think saying "`None` may be missing a `__setitem__` method." would actually be the worst thing -- I wondered why it wasn't there when reading the tests. It is in fact the case that the reason for the diagnostic is that `None` has no `__setitem__` method! But I'm also not opposed to what you do here.

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Invalid_key_type_(d3d47de65fb3bad).snap
sharkdp: I wanted it to be more concise, yes. But I agree that the slightly longer form "Cannot assign to a subscript on object of type …" is better. Changed.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: I chose the wording *"`Xyz` may be missing a `__setitem__` method"* because it suggests to me that the user may want to consider adding such a method to `Xyz`, and that is certainly not a helpful hint for stdlib types. I changed it now to say *"`None` does not have a `__setitem__` method"* for known classes.

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Invalid_key_type_(d3d47de65fb3bad).snap
```diff
@@ -19,12 +19,14 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
 # Diagnostics
 
 ```
-error[invalid-assignment]: Method `__setitem__` of type `bound method dict[str, int].__setitem__(key: str, value: int, /) -> None` cannot be called with a key of type `Literal[0]` and a value of type `Literal[3]` on object of type `dict[str, int]`
+error[invalid-assignment]: Cannot assign to a subscript on object of type `dict[str, int]` with key of type `Literal[0]` and a value of type `Literal[3]`
```
AlexWaygood: This is the same number of characters but fewer words, so it feels slightly more terse -- WDYT?

```suggestion
error[invalid-assignment]: Invalid subscript assignment on object of type `dict[str, int]` with key of type `Literal[0]` and value of type `Literal[3]`
```

or alternatively, we could put the key/value first, which feel maybe more important than the type of the object being subscripted:

```suggestion
error[invalid-assignment]: Cannot assign to a subscript with a key of type `Literal[0]` and a value of type `Literal[3]` on an object of type `dict[str, int]`
```

or

```suggestion
error[invalid-assignment]: Invalid subscript assignment with key of type `Literal[0]` and value of type `Literal[3]` on object of type `dict[str, int]`
```

the last suggestion is maybe my favourite 😄

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Invalid_key_type_for…_(815dae276e2fd2b7).snap
```diff
@@ -24,7 +24,7 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
 # Diagnostics
 
 ```
-error[invalid-key]: Cannot access `Config` with a key of type `Literal[0]`. Only string literals are allowed as keys on TypedDicts.
+error[invalid-key]: TypedDict `Config` can only be subscripted with string literal keys, got key type `Literal[0]`.
```
AlexWaygood: "key type `Literal[0]`" isn't quite grammatical, I don't think

```suggestion
error[invalid-key]: TypedDict `Config` can only be subscripted with a string literal key, got key of type `Literal[0]`.
```

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_No_`__setitem__`_met…_(468f62a3bdd1d60c).snap
```diff
@@ -23,13 +23,14 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/subscript/assignment_dia
...
-error[invalid-assignment]: Cannot assign to a subscript on an object of type `ReadOnlyDict` with no `__setitem__` method
+error[invalid-assignment]: Cannot assign to a subscript on an object of type `ReadOnlyDict`
  --> src/mdtest_snippet.py:6:1
   |
 5 | config = ReadOnlyDict()
 6 | config["retries"] = 3  # error: [invalid-assignment]
-  | ^^^^^^
+  | ^^^^^^^^^^^^^^^^^
   |
+info: `ReadOnlyDict` may be missing a `__setitem__` method.
```
AlexWaygood: "may be" sounds like we're unsure here -- it's the language we use elsewhere if we're emitting a diagnostic on a union, for example, where some elements of the union are fine but other elements of the union are invalid. In this situation, I think we're very confident that it doesn't have a `__getitem__` method, so I think this should be

```suggestion
info: `ReadOnlyDict` has no `__setitem__` method.
```

AlexWaygood (approved):
Fantastic, thank you!

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3885,7 +3885,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                     ));
                                 } else {
                                     diagnostic.info(format_args!(
```
AlexWaygood: ah, if this is a "here's how to fix this:" subdiagnostic, it might be better as

```suggestion
                                    diagnostic.help(format_args!(
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3107,9 +3107,10 @@ pub(crate) fn report_invalid_key_on_typed_dict<'db>(
             }
             _ => {
                 let mut diagnostic = builder.into_diagnostic(format_args!(
-                    "Invalid key of type `{}` for TypedDict `{}`",
-                    key_ty.display(db),
+                    "TypedDict `{}` can only be subscripted with string literal keys, \
```
AlexWaygood, line 3110: for consistency with the other message, which now uses the phrase "a string literal key" rather than "string literal keys"

```suggestion
                    "TypedDict `{}` can only be subscripted with a string literal key, \
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3778,44 +3777,86 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                                            if let Some([expected_key_ty, expected_value_ty]) =
+                                                object_ty
+                                                    .known_specialization(db, KnownClass::Dict)
+                                                    .map(|s| s.types(db))
+                                            {
+                                                let mut diagnostic = builder.into_diagnostic(format_args!(
+                                                    "Invalid subscript assignment with key of type `{}` and value of \
+                                                     type `{assigned_d}` on object of type `{object_d}`",
+                                                    slice_ty.display(db),
+                                                ));
```
AlexWaygood, line 3807: I see that we can only give the nice subdiagnostic if it's a dict, but the top-level diagnostic summary here seems like it would be an improvement for non-dictionaries as well as dictionaries? The terms "key", "value", and "subscript assignment" aren't specific to dictionaries; they apply to any object with a `__setitem__` method.

But I suppose that can be done as a separate change

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3824,13 +3865,30 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
                                     object_ty.display(db),
                                 ));
-                                attach_original_type_info(diagnostic);
+                                attach_original_type_info(&mut diagnostic);
+
+                                // Use `KnownClass` as a crude proxy for checking if this is not a user-defined class. Otherwise,
+                                // we end up suggesting things like "Consider adding a `__setitem__` method to `None`".
+                                if object_ty
+                                    .as_nominal_instance()
+                                    .is_some_and(|instance| instance.class(db).known(db).is_some())
```
AlexWaygood, line 3880: A more exact way of checking whether a class is user-defined might be:

```suggestion
                                // If it's a user-defined class, suggest adding a `__setitem__` method.
                                if object_ty
                                    .as_nominal_instance()
                                    .and_then(|instance| {
                                        file_to_module(
                                            db,
                                            instance.class(db).class_literal(db).0.file(db),
                                        )
                                    })
                                    .and_then(|module| module.search_path(db))
                                    .is_some_and(crate::SearchPath::is_first_party)
```

PR #21412, opened 2025-11-12
RFC [ruff][ext-lint] external lint rule registry, type definitions and serde config

pieterh-oai:
## Summary

> [!Note]
> This is PR 1/3 in an initial 3-PR stack. The two follow-on PRs are #21413 and #21415 .

The goal of this stack is, more or less, to explore what it looks like if we hook up PyO3 and use it to run out-of-tree lint checks written in Python (focusing only on AST linters and excluding fixers). This is based on some of the discussion (on Discord and Notion) regarding #283. 

This PR adds a registry, the config datatypes, TOML file parsing, and some initial plumbing to support external linters, with the following constraints:

- We want to minimize changes to the existing Ruff plumbing for how ‘internal’ Rust-based linters and rules are defined, selected, and run.
    - This PR adds a single rule,  `RUF300`, that is used to represent all external linters. This is used mostly internally (though it can be `--select`ed). This makes it easy to disable all external linters in one place; folks who don’t use this functionality can be reasonably sure that it isn’t costing them performance.
    - The downside is that we will need our own machinery for selecting/ignoring/extend-*-ing external linters, supporting noqa, and so on. This is fairly complex, so I made it a separate PR.

```bash
ruff.toml / pyproject.toml 1--->n my_external_linter.toml 1--->n my_rule.py
```

- The ‘schema’ ^ expects each external linter to live in its own TOML file, defining one or more rules. Each rule has a few TOML properties and is defined in a separate Python file. Rules ‘subscribe’ to the node types they want to examine. We store that mapping in a registry on the Rust side so that the AST visitor can locate and run the correct rules reasonably quickly/efficiently.

## Test Plan

```bash
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo test --workspace --exclude ty

# add RUF300 to ruff.schema.json
cargo dev generate-all
```


crates/ruff_linter/src/external/ast/registry.rs
```diff
@@ -0,0 +1,163 @@
...
+        let linter = self.linters.get(locator.linter_index)?;
+        let rule = linter.rules.get(locator.rule_index)?;
+        Some((rule, linter))
+    }
+}
+
+impl ruff_cache::CacheKey for ExternalLintRegistry {
+    fn cache_key(&self, key: &mut ruff_cache::CacheKeyHasher) {
+        key.write_usize(self.linters.len());
+        for linter in &self.linters {
```
MichaReiser, line 137: This might be more tricky. Just hashing the linter metadata isn't enough. We'd also need to hash the linter implementation, e.g. by hasing the linter file content and all its dependencies. 


I suspect that we want the linter to provide its own hash during initialization.

crates/ruff_linter/src/external/ast/registry.rs
```diff
@@ -0,0 +1,163 @@
...
+        Self {
+            linter_index,
+            rule_index,
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct ExternalLintRegistry {
+    linters: Vec<ExternalAstLinter>,
```
MichaReiser: Nit: You could use an `IndexVec` here to get a type safe `LinterIndex` type (over just an `usize` where it's unclear into what the value is indexing into)

crates/ruff_linter/src/external/ast/registry.rs
```diff
@@ -0,0 +1,163 @@
...
+    index_by_code: FxHashMap<String, RuleLocator>,
+    stmt_index: FxHashMap<StmtKind, Vec<RuleLocator>>,
+    expr_index: FxHashMap<ExprKind, Vec<RuleLocator>>,
```
MichaReiser, line 32: Nit: These could probably become `Vec`'s if we add a method to convert from `StmtKind` to `usize`.

MichaReiser (commented):
I started skimming the code. I'll leave comments as I go.

crates/ruff_linter/src/external/ast/registry.rs
pieterh-oai: Ack. For transparency, I haven't worried or tested caching yet in trying to get this to just work. This is currently used to select interpreter+environments for the runtime. It's not really ready for actual caching yet, but I think it can be made to work with a few more fields included.

pieterh-oai:
(rebase)

pieterh-oai:
(push: Hopefully satify the mkdoc CI job).

pieterh-oai:
(check_docs_formatted.py)

crates/ruff_linter/src/external/ast/registry.rs
pieterh-oai: Changed this to IndexVec with a newtype equivalent.

pieterh-oai:
(Rebase; IndexVec per @MichaReiser )

PR #21413, opened 2025-11-12
RFC [ruff][ext-lint] 2: external lint rule selection

pieterh-oai:
## Summary

> [!Note]
> This PR 2 in  a stack of 3. This PR contains 2 commits; the base commit is PR'ed in  #21412 . The follow-on PR is #21415  . The `[ruff][ext-lint] external lint rule selection` commit is the actual delta for this PR.

This PR implements external lint rule selection in both config and from the CLI, supporting the usual `select`, `ignore`, `extend-select`, and `extend-ignore`.

To ensure that internal rules are selected and ignored exactly as before, the flags for external rules are separate (`select-external`, `ignore-external`, etc.). In addition to avoiding weird corner cases, having separate flags means the use of external linters is always visible/explicit.

Based on the configuration and flags, we trim an `ExternalLintRegistry` (introduced in #21412) to contain the correct rules for each configuration scope. The assumption is that building registries will be relatively infrequent, while quickly finding and executing the correct rules is common and more performance sensitive.

Based on how we’ve set up the config (in #21412), the available set of external lint rules is not constant across the workspace. Any non-root `ruff.toml` that we encounter could define new linters+rules that don’t exist in other parts of the same codebase. In addition, linter .toml files can flag a linter as enabled/disabled.

I’d say this is fairly gnarly, but I’m not sure there is a great way to better share code with the internal linter selection path. 

## Test Plan

```bash
cargo clippy --workspace --all-targets --all-features -- -D warnings
RUFF_UPDATE_SCHEMA=1 cargo test --workspace --exclude ty
```


MichaReiser:
Hmm, this is a bit painful to review because I can't change the baseline to your first PR.

MichaReiser:
> To ensure that internal rules are selected and ignored exactly as before, the flags for external rules are separate (select-external, ignore-external, etc.). In addition to avoiding weird corner cases, having separate flags means the use of external linters is always visible/explicit.

Can you say more about the decision to use separate configuration options? The part that I'm worried about is that this wont' scale well. We'll need an `external` variant anywhere where we accept rule codes (fixable, unfixable, and possibly other places in the future)

pieterh-oai:
> Hmm, this is a bit painful to review because I can't change the baseline to your first PR.

Sorry about that. I'm used to stacking and I had no idea while setting this up that I'd have to base each PR off of `main`. I use this to look at these, fwiw:
Screenshot 2025-11-13 at 8 59 31 AM
Screenshot 2025-11-13 at 8 59 54 AM

I'm not sure that fits into how you do reviews generally, but hopefully it's helpful.

> Can you say more about the decision to use separate configuration options?

My main concern, at least when I started out, was around overlapping rule names. Suppose someone writes RLI001 as an external linter somewhere in a private repo. Later, someone coincidentally writes a built-in rule with the same code. That sort of thing.

I think it depends a little on the 'policies' we want to set around external linters:
* Should it be possible to define new external linters in nested TOML config files?
* If a nested TOML file defines an external linter, should I be able to enable it / disable it from a 'parent' TOML?

On the plus side, I think most alternatives are probably simpler than this current implementation, so in some sense this PR represents the worst case scenario for external rule selection :)

pieterh-oai:
(rebase)

pieterh-oai:
(rebase)

Issue #10084, opened 2024-02-22
EXE001 and EXE002 do not run on WSL


MusicalNinjaRandInt:
Background:
I have just set up ruff for the first time, and loved it, so fast and so complete! I fixed all the errors, added the github action to my pipeline, pushed the commit and ... boom, pipeline fail due to EXE002 ;(

The opening sentence of Microsoft's documentation on [Working across file systems](https://learn.microsoft.com/en-us/windows/wsl/filesystems#file-storage-and-performance-across-file-systems) is:

> We recommend against working across operating systems with your files, unless you have a specific reason for doing so.

Executing checks for EXE001 and EXE002 should therefore be possible on WSL in most situations. These have been specifically disabled with #5735 

Suggestion:
Add a configuration option `lint.ignore-exe-wsl`; ideally defaulting to `false` to allow users to configure this for their situation

I would be happy to support a PR with documentation and/or testing but have zero rust experience and the code-base is huge (I'd need too many pointers on how to add the config option)

MusicalNinjaRandInt:
Hi @MichaReiser , thanks. Yes, the issue was that I had no complaints when running locally but did in the CI. And the CI was correct - I had a file with 0755 permissions and no shebang.

The files on my local setup are in the WSL filesystem (ext4). This aligns with the recommendations from MS - there is a huge performance and functionality difference vs. mounting the windows system

MusicalNinjaRandInt:
Sorry, @MichaReiser , just saw your comment now.
Maybe the presence or response of https://docs.rs/filesystem/latest/filesystem/trait.UnixFileSystem.html could be used? Then the implementation wouldn't be Windows / Linux / WSL / non-WSL / etc... but based on the FS capabilities, which is probably more future-proof and clear.
The question is - how does this trait respond on a Windows filesystem?

mikeleppane:
@MusicalNinjaRandInt I've also had the same issue when running tests locally. I also use WSL and files are stored in WSL file system. Please fix this! 

> Suggestion:
Add a configuration option lint.ignore-exe-wsl; ideally defaulting to false to allow users to configure this for their situation

Is this the way to tackle this?

mikeleppane:
```bash
failures:

---- rules::flake8_executable::tests::rules::path_new_exe002_1_py_expects stdout ----
[...]

PR #21416, opened 2025-11-12
[ty] Fix panic for unmatched trailing revealed assertions in mdtests

Hugo-Polloli:
## Summary

- Prevent mdtest from panicking when an inline assertion refers to a non-existent line (e.g., a trailing `# revealed: ...` with no matching `reveal_type`)
- `EmbeddedFileSourceMap::to_absolute_line_number` now returns `Result` with a clear diagnostic, plus a helper to report the last valid embedded code line
- Mdtest CLI/GitHub output surfaces both the original assertion failure and the new explanation, and parser unit tests cover both the success and error paths

Fixes [ty#1512](https://github.com/astral-sh/ty/issues/1512).

## Test Plan

```shell
cargo clippy --workspace --all-targets --all-features -- -D warnings  # Rust linting
cargo test  # Rust testing
uvx pre-commit run --all-files --show-diff-on-failure  # Rust and Python formatting, Markdown and Python linting, etc...
``` 
Everything ok except the 2 mentioned in #10084 but on WSL so no issue here

### DIFF
--- a/crates/ty_test/src/lib.rs
+++ b/crates/ty_test/src/lib.rs
@@ -86,28 +86,34 @@ pub fn run(
                     EmbeddedFileSourceMap::new(&md_index, test_failures.backtick_offsets);
 
                 for (relative_line_number, failures) in test_failures.by_line.iter() {
-                    let (absolute_line_number, line_number_error, emit_failures) = match source_map
-                        .to_absolute_line_number(relative_line_number)
-                    {
-                        Ok(line_number) => (Some(line_number), None, true),
-                        Err(last_line_number) => {
-                            let message = "Found a trailing assertion comment (e.g., `# revealed:` or `# error:`) \
-                                not followed by any statement.";
-                            (Some(last_line_number), Some(message), false)
-                        }
-                    };
-
                     let file = match output_format {
                         OutputFormat::Cli => relative_fixture_path.as_str(),
                         OutputFormat::GitHub => absolute_fixture_path.as_str(),
                     };
-                    if emit_failures {
-                        for failure in failures {
-                            output_format.print_error(file, absolute_line_number, failure);
+
+                    let absolute_line_number = match source_map
+                        .to_absolute_line_number(relative_line_number)
+                    {
+                        Ok(line_number) => line_number,
+                        Err(last_line_number) => {
+                            print!("{}",
+                                output_format.display_error(
+                                    file,
+                                    last_line_number,
+                                    "Found a trailing assertion comment (e.g., `# revealed:` or `# error:`) \
+                                    not followed by any statement."
+                                )
+                            );
+
+                            continue;
                         }
-                    }
-                    if let Some(err) = line_number_error {
-                        output_format.print_error(file, absolute_line_number, err);
+                    };
+
+                    for failure in failures {
+                        print!(
+                            "{}",
+                            output_format.display_error(file, absolute_line_number, failure)
+                        );
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/ty_test/src/lib.rs
```diff
@@ -86,20 +86,49 @@ pub fn run(
...
+
+                    match output_format {
+                        OutputFormat::Cli => {
+                            for failure in failures {
+                                println!("  {} {failure}", cli_line_info.as_str().cyan());
+                            }
+                            if let Some(err) = line_number_error.as_deref() {
+                                println!("  {} {err}", cli_line_info.as_str().cyan());
+                            }
+                        }
```
MichaReiser: Nit: to reduce code duplication, I suggest adding a method to `OutputFormat` to print an error


```rust
impl OutputFormat {
	fn print_error(self, file: &str, line: Option<OneIndexed>, failure: impl Display) {
		match self {
			Self::Cli => {
				let cli_line_info = match absolute_line_number_value {
		        Some(line) => format!("{relative_fixture_path}:{line}"),
		        None => format!("{relative_fixture_path}:?"),
		    };
				println!(" {} {failure}", cli_line_info.as_str().cyan());
			}
			Self::GitHub => {
				if let Some(line) = absolute_line_number_value {
					println!("::error file={absolute_fixture_path},line={line}::{err}");
				} else {
					println!("::error file={absolute_fixture_path}::{err}");
				}
			}
		}
	}
}

File: crates/ty_test/src/parser.rs [hunk 2]
```diff
@@ -273,20 +273,35 @@ impl EmbeddedFileSourceMap {
         }
     }
 
-    pub(crate) fn to_absolute_line_number(&self, relative_line_number: OneIndexed) -> OneIndexed {
-        let mut absolute_line_number = 0;
+    pub(crate) fn to_absolute_line_number(
+        &self,
+        relative_line_number: OneIndexed,
+    ) -> anyhow::Result<OneIndexed> {
```
MichaReiser: I think I would change this to be an `std::result::Result<OneIndexed, OneIndexed>` where
the error case returns the last line number (`last_absolute_line_number`). 

This leaves deciding on what error message to use up to the caller

crates/ty_test/src/lib.rs
```diff
@@ -86,20 +86,49 @@ pub fn run(
...
+                        OutputFormat::GitHub => {
+                            for failure in failures {
+                                if let Some(line) = absolute_line_number_value {
+                                    println!(
+                                        "::error file={absolute_fixture_path},line={line}::{failure}"
+                                    );
+                                } else {
+                                    println!("::error file={absolute_fixture_path}::{failure}");
+                                }
+                            }
```
MichaReiser: I'm slightly leaning towards skipping the `failures` if the line number can't be mapped because I suspect it leads to slightly simpler code but I'm curious to hear why you decide to emit them even if the file mapping failed.

crates/ty_test/src/lib.rs
```diff
@@ -86,22 +86,34 @@ pub fn run(
...
-                            ),
+                    let (absolute_line_number, line_number_error, emit_failures) = match source_map
+                        .to_absolute_line_number(relative_line_number)
+                    {
+                        Ok(line_number) => (Some(line_number), None, true),
+                        Err(last_line_number) => {
+                            let total = source_map.total_relative_line_count();
+                            let message = format!(
+                                "Relative line {} exceeds total embedded code lines ({total}). \
+                                     This usually means a trailing `# revealed:` without a matching \
```
MichaReiser: Can you help me understand how `total_relative_line_count` is different from `last_absolute_line_number`?

crates/ty_test/src/lib.rs
Hugo-Polloli: Thanks for the implementation, I used it as is

crates/ty_test/src/parser.rs
Hugo-Polloli: Oh yes, much cleaner, done

crates/ty_test/src/lib.rs
Hugo-Polloli: No real reason haha, to be honest I kinda emitted them "by default" and did not think about just not skipping them, done

crates/ty_test/src/lib.rs
Hugo-Polloli: This was supposed to be complementary information, `total_relative_line_count` showing the number of lines in the embedded snippet, `relative_line_number` being in our case equal to `total_relative_line_count + 1`, showing that the assertion engine tried to find a line that just does not exist due to the trailing assertion comment.
Decided to discard it because `last_absolute_line_number` and the error message are plenty enough information for the user to find what's happening

(Separate process question: when I’m still iterating on feedback and not ready for another review yet, is it preferable to convert the PR to a draft?)

btw the changes in [suggestions follow-up](https://github.com/astral-sh/ruff/pull/21416/commits/4170fa988fbc3a7c9f9a0cc3074d7a7b05b3a4d1) commit are, this time, ready for review, I changed the error message to be more generic than just about revealed, because this happens for any type of assertion comment

PR #18452, opened 2025-06-04
[ty] Surface matched overload diagnostic directly

dhruvmanila:
## Summary

This PR resolves the way diagnostics are reported for an invalid call to an overloaded function.

If any of the steps in the overload call evaluation algorithm yields a matching overload but it's type checking that failed, the `no-matching-overload` diagnostic is incorrect because there is a matching overload, it's the arguments passed that are invalid as per the signature. So, this PR improves that by surfacing the diagnostics on the matching overload directly.

It also provides additional context, specifically the matching overload where this error occurred and other non-matching overloads. Consider the following example:

```py
from typing import overload


@overload
def f() -> None: ...
@overload
def f(x: int) -> int: ...
@overload
def f(x: int, y: int) -> int: ...
def f(x: int | None = None, y: int | None = None) -> int | None:
    return None


f("a")
```

We get:

Screenshot 2025-06-18 at 11 07 10

## Test Plan

Update test cases, resolve existing todos and validate the updated snapshots.

PR #21417, opened 2025-11-13
[ty] Resolve overloads for hovers

Gankra:
This is a very conservative minimal implementation of applying overloads to resolve a callable-type-being-called down to a single function signature on hover. If we ever encounter a situation where the answer doesn't simplify down to a single function call, we bail out to preserve prettier printing of non-raw-Signatures.

The resulting Signatures are still a bit bare, I'm going to try to improve that in a followup to improve our Signature printing in general.

Fixes https://github.com/astral-sh/ty/issues/73

### DIFF
--- a/crates/ty_ide/src/hover.rs
+++ b/crates/ty_ide/src/hover.rs
@@ -6,7 +6,6 @@ use ruff_db::parsed::parsed_module;
 use ruff_text_size::{Ranged, TextSize};
 use std::fmt;
 use std::fmt::Formatter;
-use ty_python_semantic::types::ide_support::CallSignatureDetails;
 use ty_python_semantic::types::{KnownInstanceType, Type, TypeVarVariance};
 use ty_python_semantic::{DisplaySettings, SemanticModel};
 
@@ -31,7 +30,7 @@ pub fn hover(db: &dyn Db, file: File, offset: TextSize) -> Option<RangedValue<Ho
         .map(HoverContent::Docstring);
 
     let mut contents = Vec::new();
-    if let Some(signature) = goto_target.signature(&model) {
+    if let Some(signature) = goto_target.call_type_simplified_by_overloads(&model) {
         contents.push(HoverContent::Signature(signature));
     } else if let Some(ty) = goto_target.inferred_type(&model) {
         tracing::debug!("Inferred type of covering node is {}", ty.display(db));
@@ -118,7 +117,7 @@ impl fmt::Display for DisplayHover<'_, '_> {
 
 #[derive(Debug, Clone)]
 pub enum HoverContent<'db> {
-    Signature(Vec<CallSignatureDetails<'db>>),
+    Signature(String),
     Type(Type<'db>, Option<TypeVarVariance>),
     Docstring(Docstring),
 }
@@ -142,14 +141,8 @@ pub(crate) struct DisplayHoverContent<'a, 'db> {
 impl fmt::Display for DisplayHoverContent<'_, '_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.content {
-            HoverContent::Signature(signatures) => {
-                for signature in signatures {
-                    self.kind
-                        .fenced_code_block(&signature.label, "python")
-                        .fmt(f)?;
-                    self.kind.horizontal_line().fmt(f)?;
-                }
-                Ok(())
+            HoverContent::Signature(signature) => {
+                self.kind.fenced_code_block(&signature, "python").fmt(f)
             }
             HoverContent::Type(ty, variance) => {
                 let variance = match variance {
@@ -973,16 +966,12 @@ def ab(a: str): ...
         assert_snapshot!(test.hover(), @r"
         (a: int) -> Unknown
         ---------------------------------------------
-
-        ---------------------------------------------
         the int overload
 
         ---------------------------------------------
         ```python
         (a: int) -> Unknown
         ```
-        ---
-
         ---
         ```text
         the int overload
@@ -1040,16 +1029,12 @@ def ab(a: str):
         assert_snapshot!(test.hover(), @r#"
         (a: str) -> Unknown
         ---------------------------------------------
-
-        ---------------------------------------------
         the int overload
 
         ---------------------------------------------
         ```python
         (a: str) -> Unknown
         ```
-        ---
-
         ---
         ```text
         the int overload
@@ -1105,18 +1090,20 @@ def ab(a: int):
             .build();
 
         assert_snapshot!(test.hover(), @r"
-        (a: int, b: int) -> Unknown
-        ---------------------------------------------
-
+        (
+            a: int,
+            b: int
+        ) -> Unknown
         ---------------------------------------------
         the two arg overload
 
         ---------------------------------------------
         ```python
-        (a: int, b: int) -> Unknown
+        (
+            a: int,
+            b: int
+        ) -> Unknown
         ```
-        ---
-
         ---
         ```text
         the two arg overload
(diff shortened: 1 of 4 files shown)


Gankra:
~~Most of these snapshot changes are bad and the consequence of this logic firing too eagerly. I need to cleanup that part, but I wanted to get the Types and Semantics part up sooner than later for correction.~~

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -973,6 +973,65 @@ pub fn call_signature_details<'db>(
...
+        // Only use the Ok
+        .iter()
+        .flatten()
+        .flat_map(|binding| {
+            binding.matching_overloads().map(|(_, overload)| {
+                overload
+                    .signature
+                    .display_with(db, DisplaySettings::default().multiline())
+                    .to_string()
+            })
```
Gankra, line 1023: I spent an *unconscionable* amount of time trying to have this return a Type or something more useful than a String but the types on `Binding` are haunted by the overloads when you try to render them (meaning printing them still returns the full overload set!). 

Stringifying the raw Signature is the only thing that seems to be safe.

Gankra: (An early version returned a full CallSignatureDetails but this usecase has no use of that so I dropped it as a waste of effort. Also this version that just returns a String is easier to iterate on (I kept trying to Not have a String and had to refactor like 4 calls and 2 types over and over as I experimented between `Type` vs `Vec<Type>` vs `CallSignatureDetails` vs...))

crates/ty_ide/src/hover.rs
```diff
@@ -93,13 +93,13 @@ impl<'a, 'db> IntoIterator for &'a Hover<'db> {
...
 }
 
-pub struct DisplayHover<'a> {
-    db: &'a dyn Db,
-    hover: &'a Hover<'a>,
+pub struct DisplayHover<'db, 'a> {
+    db: &'db dyn Db,
+    hover: &'a Hover<'db>,
     kind: MarkupKind,
 }
```
Gankra, line 100: This is debris from versions where this was being asked to store `CallSignatureDetails`, which is actually invariant over `'db`. Leaving it like this just simplifies any potential future work.

crates/ty_ide/src/hover.rs
```diff
@@ -30,9 +29,10 @@ pub fn hover(db: &dyn Db, file: File, offset: TextSize) -> Option<RangedValue<Ho
         .and_then(|definitions| definitions.docstring(db))
         .map(HoverContent::Docstring);
 
-    // TODO: Render the symbol's signature instead of just its type.
     let mut contents = Vec::new();
-    if let Some(ty) = ty {
+    if let Some(signature) = goto_target.call_type_simplified_by_overloads(&model) {
```
MichaReiser, line 33: Is there a reason we only apply overload simplification for hover mode? For example, TypeScript jumps to the matching overload if you use "go to definition," which matches my expectation. https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABFApgZygCgIYC5FggC2ARigE4CU+hpFA3AFCiSwLLpZ4HFnkA0iEvgzkYYAObVEo8RKYto8JKgw4avCoJIB+EVDGTp2MAE9EAb0aMAvtcaqsARgAMggESP3le0A

I'm surprised that Pylance doesn't seem to do that

MichaReiser (commented):
I'll do a full review on Monday but I've a question why we limit overload matching to hover only

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -973,6 +973,65 @@ pub fn call_signature_details<'db>(
...
+/// succeeds in are those where we would print a complicated/ugly type anyway.
+pub fn call_type_simplified_by_overloads<'db>(
+    db: &'db dyn Db,
+    model: &SemanticModel<'db>,
+    call_expr: &ast::ExprCall,
+) -> Option<String> {
+    let func_type = call_expr.func.inferred_type(model);
+
+    // Use into_callable to handle all the complex type conversions
+    let callable_type = func_type.try_upcast_to_callable(db)?;
```
MichaReiser, line 995: Is my understanding correct that this adds another "todo" call-site to https://github.com/astral-sh/ty/issues/1086?

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -973,6 +973,65 @@ pub fn call_signature_details<'db>(
...
+        .iter()
+        .flatten()
+        .flat_map(|binding| {
+            binding.matching_overloads().map(|(_, overload)| {
+                overload
+                    .signature
+                    .display_with(db, DisplaySettings::default().multiline())
+                    .to_string()
+            })
+        })
```
dhruvmanila, line 1024: I think I remember now facing a similar problem regarding using the `signature_type` on `Binding` struct that we saw during our call last week. For context, the `signature_type` would represent the `FunctionLiteral` which means the display would contain all the overloads and this is the main reason we need to use the display of `Signature` to display the _matching_ overload which leads to requiring https://github.com/astral-sh/ruff/pull/21438.

I faced this issue while implementing https://github.com/astral-sh/ruff/pull/18452 and the way I resolved it is by capturing the required information (`FunctionType`, `BoundMethodType`, etc. and the matching overload index) in [`MatchingOverloadLiteral`](https://github.com/astral-sh/ruff/blob/dcc451d4d2fcf76b4c930f6aaa2f66b4830f7a04/crates/ty_python_semantic/src/types/call/bind.rs#L4130-L4141) and implementing a [`.get`](https://github.com/astral-sh/ruff/blob/dcc451d4d2fcf76b4c930f6aaa2f66b4830f7a04/crates/ty_python_semantic/src/types/call/bind.rs#L4144-L4155) method to return the matching `OverloadLiteral`.

I think we could possibly use a similar approach to get the actual `OverloadLiteral` that needs to be displayed as it does [implement the `Display` trait](https://github.com/astral-sh/ruff/blob/dcc451d4d2fcf76b4c930f6aaa2f66b4830f7a04/crates/ty_python_semantic/src/types/display.rs#L757-L774). You might find it useful to go through this part of the code: https://github.com/astral-sh/ruff/blob/dcc451d4d2fcf76b4c930f6aaa2f66b4830f7a04/crates/ty_python_semantic/src/types/call/bind.rs#L2125-L2162

Another approach could be to expand the `DisplaySettings` to include the `matching_overload_index: MatchingOverloadIndex` which can then be used directly to filter the overloads at the display level. The `None` variant would display all the overloads.

crates/ty_python_semantic/src/types/ide_support.rs
carljm: I'm not sure if we'll be able to do https://github.com/astral-sh/ty/issues/1086; commented on that issue.

crates/ty_ide/src/hover.rs
Gankra: Hmm strictly speaking I don't think goto-definition should jump to an overload unless we can't find the actual implementation (in which case it's actually goto-declaration). goto-declaration should ideally jump to the overload though.

MichaReiser:
@Gankra what's the status of this PR?

Gankra:
I think this PR could be landed as-is as an incremental strict improvement. I've been locally experimenting with alternative approaches that would e.g. empower goto-declaration as well, but I keep running into issues around the fact that randomly places in the code (like type printing) don't want to talk about A Specific Overload (connected to the fact that definitions-hanging-off-types is a fundamental hack).

MichaReiser:
I'm fine with handling go-to separately. But we might want to address @dhruvmanila's feedback before landing?

Gankra:
Yeah dhruv's suggestions are what I was experimenting with locally. Hover doesn't really benefit from it (it really just needs a String, and I'm computing that String fine), but goto-decl would benefit.

Gankra:
Or well, I suppose dhruv's suggestions would ideally obsolete the more hacky stuff in #21438 

But yeah it's all kinda the same work.

PR #15665, opened 2025-01-22
[`flake8-simplify`] Also simplify annotated assignments (`SIM108`)

InSyncWithFoo:
## Summary

Resolves #15554.

## Test Plan

`cargo nextest run` and `cargo insta test`.

Issue #15554, opened 2025-01-17
`SIM108` false negative when variable has type annotation


nickdrozd:
Assignment with `if/else` gets flagged with `SIM108`, suggesting ternary operator. Good. But if the variable is annotated, not flagged:

```python
from random import randint

if randint(0, 3):
    x: int = 4  # not flagged
else:
    x = 5
```

Could be `if` var annotated or `else` var, not flagged. Also not flagged if both vars have annotation, even if annotation is the same.

MichaReiser:
That makes sense to me but is slightly more complicated because that only works if both branches infer to the same type, unless the fix removes the type annotation.

nickdrozd:
If the variable is declared with different types in the branches, Mypy flags it as a redefinition:

```python
if randint(0, 3):
    x: int = 4
else:
    x: str = 'q'  # mypy flags as redefined
```

(It's not even clear whether there is one variable `x` defined in two places or two separate variables each called `x`. Yet another reason not to define a variable in branches.)

In this case it would be fine to add a new annotation that is the union of types in the branches:

```python
x: int | str = 4 if randint(0, 3) else 'q'
```

MichaReiser:
The problem I see is more about:

```
if randint(0, 3):
    x: int = 4
else:
    x = 'q'  # mypy flags as redefined
```

Mypy might also flag this as redefinition but I'd have to check if this is true for every type checker (including red knot). But I could still see us flagging this but we may not want to provide a fix for it.

JelleZijlstra:
Right now, both ty and pyrefly allow this program:

```python
from typing import reveal_type
from random import randint

if randint(0, 3):
    x: int = 4  # not flagged
else:
    x = "x"
```

So "simplifying" this is questionable.

PR #21418, opened 2025-11-13
[`flake8-simplify`] Support annotated assignments in if block (`SIM108`)

### LINKED ISSUE #21406: SIM108 is blocked by annotations
### Summary

[`if-else-block-instead-of-if-exp` (SIM108)](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/) should apply when the assignment in the `if` block (though not the one in the `else` block) is annotated. [Example](https://play.ruff.rs/851d6427-dcb3-4325-b173-3028c0a500a1):
```console
$ cat >sim108.py <<'# EOF'
def f(x: object) -> int:
    if x:
        y: int = 1
    else:
        y = 2
    return y
# EOF

$ ruff --isolated check sim108.py --select SIM108 --unsafe-fixes --fix
All checks passed!
```

The ideal fixed version is:
```python
def f(x: object) -> int:
    y: int = 1 if x else 2
    return y
```

### Version

danparizher:
## Summary

Fixes SIM108 to apply when the if block contains an annotated assignment (e.g., `y: int = 1`) and the else block contains a regular assignment (e.g., `y = 2`), converting it to `y: int = 1 if x else 2`.

Fixes #21406.

## Problem

SIM108 only matched `Stmt::Assign` patterns in both if and else blocks, so it didn't trigger when the if block used `Stmt::AnnAssign` with a value.

## Approach

Updated pattern matching to accept both `Stmt::Assign` and `Stmt::AnnAssign` (with value) in the if block, extracted the annotation when present, and modified `assignment_ternary`, `assignment_binary_or`, and `assignment_binary_and` to accept an optional annotation and create `StmtAnnAssign` nodes when an annotation is provided.

## Test Plan

Added a test case to `SIM108.py` for the annotated assignment scenario and updated the snapshot.

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM108.py
+++ b/crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM108.py
@@ -264,37 +264,10 @@ def f(args):
     return outputs
 
 
-# SIM108 - annotated assignment with Optional type
-def f(self):
-    if self.has_tool_choice:
-        tool_choice: str | None = "any"
-    else:
-        tool_choice = None
-    return tool_choice
-
-
 # SIM108 - annotated assignment with Optional type and attribute access
 def f(stream):
     if stream:
         stream_name: str | None = stream.name
     else:
         stream_name = None
-    return stream_name
-
-
-# OK - both branches have matching Optional annotations (SIM108 does not trigger)
-def f(stream):
-    if stream:
-        stream_name: str | None = stream.name
-    else:
-        stream_name: str | None = None
-    return stream_name
-
-
-# OK - both branches have non-matching Optional annotations (SIM108 does not trigger)
-def f(stream):
-    if stream:
-        stream_name: str | None = stream.name
-    else:
-        stream_name: int | None = None
     return stream_name
--- a/crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
+++ b/crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
@@ -7,6 +7,7 @@ use ruff_text_size::{Ranged, TextRange};
 
 use crate::checkers::ast::Checker;
 use crate::fix::edits::fits;
+use crate::preview::is_sim108_annotated_assignment_enabled;
 use crate::{Edit, Fix, FixAvailability, Violation};
 
 /// ## What it does
@@ -112,7 +113,6 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
     else {
         return;
     };
-    // Handle both Stmt::Assign and Stmt::AnnAssign in the if block
     let (body_target, body_value, body_annotation) = match body.as_slice() {
         [
             Stmt::Assign(ast::StmtAssign {
@@ -133,7 +133,7 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
                 annotation: body_annotation,
                 ..
             }),
-        ] => (
+        ] if is_sim108_annotated_assignment_enabled(checker.settings()) => (
             body_target.as_ref(),
             body_value.as_ref(),
             Some(body_annotation.as_ref()),
@@ -214,8 +214,7 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
                 && !contains_effect(test_node, |id| checker.semantic().has_builtin_binding(id)) =>
         {
             let target_var = &body_target;
-            let binary =
-                assignment_binary_or(target_var, body_value, else_value.as_ref(), body_annotation);
+            let binary = assignment_binary_or(target_var, body_value, else_value, body_annotation);
             (checker.generator().stmt(&binary), AssignmentKind::Binary)
         }
         (test_node, body_node)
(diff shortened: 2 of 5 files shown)


danparizher:
Hi @dylwil3, I reread the updated version of `CONTRIBUTING.md` to make sure I'm aligning closer to PR expectations. I did this one manually; would you say this is closer to what you are looking for? Any specific feedback for how I created this one? Thanks!

dylwil3:
Thanks this is definitely easier for me to read! Could you look through the ecosystem report and add some more test coverage? For example, it would be good to have cases where both branches have the annotation (including a case where the annotations do not match). Thank you!

crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
```diff
@@ -269,45 +296,92 @@ fn assignment_ternary(
...
+            range: TextRange::default(),
+            node_index: ruff_python_ast::AtomicNodeIndex::NONE,
+        };
+        node1.into()
+    }
 }
 
-fn assignment_binary_or(target_var: &Expr, left_value: &Expr, right_value: &Expr) -> Stmt {
-    (ast::StmtAssign {
+fn assignment_binary_or(
```
ntBre, line 358: There's a lot of duplication here between `assignment_binary_or` and `assignment_binary_and`. In fact, with a bit of rearranging, I'm pretty sure the only difference on main is the `BoolOp` variant:

```rust
fn assignment_binary_and(target_var: &Expr, left_value: &Expr, right_value: &Expr) -> Stmt {
    let node = ast::ExprBoolOp {
        op: BoolOp::And,
        values: vec![left_value.clone(), right_value.clone()],
        range: TextRange::default(),
        node_index: ruff_python_ast::AtomicNodeIndex::NONE,
    };
    let node1 = ast::StmtAssign {
        targets: vec![target_var.clone()],
        value: Box::new(node.into()),
        range: TextRange::default(),
        node_index: ruff_python_ast::AtomicNodeIndex::NONE,
    };
    node1.into()
}

fn assignment_binary_or(target_var: &Expr, left_value: &Expr, right_value: &Expr) -> Stmt {
    let node = ast::ExprBoolOp {
        op: BoolOp::Or,
        values: vec![left_value.clone(), right_value.clone()],
        range: TextRange::default(),
        node_index: ruff_python_ast::AtomicNodeIndex::NONE,
    };
    let node1 = ast::StmtAssign {
        targets: vec![target_var.clone()],
        value: Box::new(node.into()),
        range: TextRange::default(),
        node_index: ruff_python_ast::AtomicNodeIndex::NONE,
    };
    node1.into()
}
```

I think it would make sense to factor out the commonality and just pass in the `BoolOp`, especially now that we have to consider the annotated case.

It seems a little harder to avoid in the ternary case.

crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM108.py
```diff
@@ -208,3 +208,93 @@ def f():
...
+    return y
+
+
+# SIM108 - annotated assignment with dict type
+def f(next_token: str | None):
+    if next_token is not None:
+        token_arg: dict[str, str] = {"nextToken": next_token}
+    else:
+        token_arg = {}
+    return token_arg
```
ntBre, line 246: Is there something special about these different annotation types? I think just covering the cases where both branches have annotations, one matching and one not, is probably sufficient since we don't really inspect the annotations, as far as I can tell.

crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
```diff
@@ -112,15 +112,33 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
...
-    let [
-        Stmt::Assign(ast::StmtAssign {
-            targets: body_targets,
-            value: body_value,
-            ..
-        }),
-    ] = body.as_slice()
-    else {
-        return;
+    // Handle both Stmt::Assign and Stmt::AnnAssign in the if block
```
ntBre: tiny nit: I don't think this comment is very helpful. The match seems straightforward enough.

crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
```diff
@@ -196,7 +214,8 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
                 && !contains_effect(test_node, |id| checker.semantic().has_builtin_binding(id)) =>
         {
             let target_var = &body_target;
-            let binary = assignment_binary_or(target_var, body_value, else_value);
+            let binary =
+                assignment_binary_or(target_var, body_value, else_value.as_ref(), body_annotation);
```
ntBre: Do we need these `as_ref` calls on `else_value`? It doesn't look like the type of `else_value` has changed, so I think it should still be okay to omit them.

crates/ruff_linter/src/rules/flake8_simplify/rules/if_else_block_instead_of_if_exp.rs
```diff
@@ -112,15 +112,33 @@ pub(crate) fn if_else_block_instead_of_if_exp(checker: &Checker, stmt_if: &ast::
...
+            (body_target, body_value.as_ref(), None)
+        }
+        [
+            Stmt::AnnAssign(ast::StmtAnnAssign {
+                target: body_target,
+                value: Some(body_value),
+                annotation: body_annotation,
+                ..
+            }),
+        ] => (
```
ntBre: I think we should preview-gate this, and this would be a good place to do so:


```suggestion
        ] if is_new_preview_function_enabled(checker.settings()) => (
```

crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM108.py
```diff
@@ -208,3 +208,66 @@ def f():
...
```
ntBre, line 246: This case is actually not triggering SIM108, but the unannotated version does in the test and in the playground: https://play.ruff.rs/ad80f8bf-776b-4243-87a9-054c21207ab2.

Also, all  of the new tests use a ternary replacement. It would be good to include tests that exercise the `or` and `and` fixes too.

ntBre:
Just commented on the issue, but I didn't realize that the issue was a duplicate of an older issue (#15554), which also had an older PR (#15665) with a `needs-decision` label. I'll close this for now, but hopefully we can use some of the code from both PRs, if we decide to make this change. Thanks for your work here!

PR #21419, opened 2025-11-13
[ty] Rename matching overload index field

dhruvmanila:
## Summary

This PR renames the `CallableBinding::matching_overload_index` field to `CallableBinding::matching_overload_after_parameter_matching` to clarify the main use case of this field which is to surface type checking errors on the matching overloads directly instead of using the `no-matching-overload` diagnostic. This can only happen after parameter matching as following steps could filter out this overload which should then result in `no-matching-overload` diagnostic.

Callers should use the `matching_overload_index` _method_ to get the matching overloads.

### DIFF
--- a/crates/ty_python_semantic/src/types/call/bind.rs
+++ b/crates/ty_python_semantic/src/types/call/bind.rs
@@ -1359,7 +1359,7 @@ impl<'db> From<Binding<'db>> for Bindings<'db> {
             dunder_call_is_possibly_unbound: false,
             bound_type: None,
             overload_call_return_type: None,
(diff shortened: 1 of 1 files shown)


sharkdp (approved):
Thank you. Given that it is potentially wrong to access this, maybe we should even choose a name that makes it more apparent that this is not the "final" matching overload? Instead of "after parameter matching", maybe it should be called "before (argument and parameter) type checking" (e.g. `matching_overload_before_type_checks`)?

dhruvmanila:
Renamed it to `matching_overload_before_type_checking`

PR #21420, opened 2025-11-13
[ty] Set `definition` modifier for parameter declarations when computing semantic tokens

lucach:
## Summary

Parameter declarations are definitions, and therefore a parameter token should be marked as a definition when computing the sematic tokens.

## Test Plan

Updated the existing tests, which were manually checked. I've also verified that the new behavior is indeed aligned with VSCode+Pylance.


MichaReiser:
Thank you

PR #21421, opened 2025-11-13
[ty] Support `type[…]` and `Type[…]` in implicit type aliases

sharkdp:
## Summary

Support `type[…]` in implicit type aliases, for example:
```py
SubclassOfInt = type[int]

reveal_type(SubclassOfInt)  # GenericAlias

def _(subclass_of_int: SubclassOfInt):
    reveal_type(subclass_of_int)  # type[int]
```

## Typing conformance

```diff
-specialtypes_type.py:138:5: error[type-assertion-failure] Argument does not have asserted type `type[Any]`
-specialtypes_type.py:140:5: error[type-assertion-failure] Argument does not have asserted type `type[Any]`
```

Two new tests passing :heavy_check_mark: 

```diff
-specialtypes_type.py:146:1: error[unresolved-attribute] Object of type `GenericAlias` has no attribute `unknown`
```

An `TA4.unknown` attribute on a PEP 613 alias (`TA4: TypeAlias = type[Any]`) is being accessed, and the conformance suite expects this to be an error. Since we currently use the inferred type for these type aliases (and possibly in the future as well), we treat this as a direct access of the attribute on `type[Any]`, which falls back to an access on `Any` itself, which succeeds. :red_circle: 

```
+specialtypes_type.py:152:16: error[invalid-type-form] `typing.TypeVar` is not a generic class
+specialtypes_type.py:156:16: error[invalid-type-form] `typing.TypeVar` is not a generic class
```

New errors because we don't handle `T = TypeVar("T"); MyType = type[T]; MyType[T]` yet. Support for this is being tracked in https://github.com/astral-sh/ty/issues/221 :red_circle:

## Ecosystem impact

Looks mostly good, a few known problems. 

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -10771,6 +10771,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                     }
                 }
             }
+            Type::SpecialForm(SpecialFormType::Type) => {
+                // Similar to the branch above that handles `type[…]`, handle `typing.Type[…]`
+                let argument_ty = self.infer_type_expression(slice);
+                return Type::KnownInstance(KnownInstanceType::TypeGenericAlias(
+                    InternedType::new(self.db(), argument_ty),
+                ));
+            }
             _ => {}
         }
 
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -607,7 +607,9 @@ def _(
     reveal_type(invalid)  # revealed: str | Unknown
 ```
 
-## `type[…]`
+## `type[…]` and `Type[…]`
+
+### `type[…]`
 
 We support implicit type aliases using `type[…]`:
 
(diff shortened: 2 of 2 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -7845,6 +7866,9 @@ pub enum KnownInstanceType<'db> {
     /// A single instance of `typing.Annotated`
     Annotated(InternedType<'db>),
 
+    /// An instance of `typing.GenericAlias` representing a `type[...]` expression.
+    TypeGenericAlias(InternedType<'db>),
```
sharkdp: Try to represent this as `Type::GenericAlias(…)` instead...

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -180,6 +180,9 @@ impl<'db> ClassBase<'db> {
                 // wrappers are just identity callables at runtime, so this sort of inheritance
                 // doesn't work and isn't allowed.
                 | KnownInstanceType::NewType(_) => None,
+                KnownInstanceType::TypeGenericAlias(_) => {
+                    Self::try_from_type(db, KnownClass::Type.to_class_literal(db), subclass)
```
sharkdp, line 184: This is mostly based on the TODO comment in `mdtest/type_of/basic.md`. Do we need to preserve more information here? I don't really understand what `class Foo(type[int]): ...` tries to achieve.

AlexWaygood: > I don't really understand what `class Foo(type[int]): ...` tries to achieve.

Nor do I. I think this is fine.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -10626,7 +10628,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 // special cases, too.
                 if class.is_tuple(self.db()) {
                     return tuple_generic_alias(self.db(), self.infer_tuple_type_expression(slice));
+                } else if class.is_known(self.db(), KnownClass::Type) {
+                    let argument_ty = self.infer_type_expression(slice);
```
AlexWaygood, line 10632: this is fine for now, but note that it results in us being too permissive when parsing the argument to `type[]`. For example, on your branch:

```py
from typing import Literal

X = type[Literal[42]]  # no error here, but there should be?

def f(
    x: type[Literal[42]],  # error: [invalid-type-form]
    y: X  # no error here either...
):
    reveal_type(x)  # revealed: @Todo(unsupported nested subscript in type[X])
    reveal_type(y)  # revealed: `<class 'int'>`
```

sharkdp: Thanks. I'll add a test with a TODO in my next PR.

PR #21422, opened 2025-11-13
[ty] Add filtering option for mdtest runner

sharkdp:
## Summary

This change to the mdtest runner makes it easy to run on a subset of tests/files. For example:

```
▶ uv run crates/ty_python_semantic/mdtest.py implicit
running 1 test
test mdtest__implicit_type_aliases ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 281 filtered out; finished in 0.83s

Ready to watch for changes...
```

Subsequent changes to either that test file or the Rust source code will also only rerun the `implicit_type_aliases` test.

Multiple arguments can be provided, and filters can either be partial file paths (`loops/for.md`, `loops/for`, `for`) or mangled test names (`loops_for`):
```
▶ uv run crates/ty_python_semantic/mdtest.py implicit binary/union

running 2 tests
test mdtest__binary_unions ... ok
test mdtest__implicit_type_aliases ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 280 filtered out; finished in 0.85s

Ready to watch for changes...
```

## Test Plan

Tested it interactively for a while

### DIFF
--- a/crates/ty_python_semantic/mdtest.py
+++ b/crates/ty_python_semantic/mdtest.py
@@ -131,13 +131,6 @@ class MDTestRunner:
             .removesuffix(".md")
         )
 
-    def _matches_filter(self, markdown_file: Path) -> bool:
-        if not self.filters:
-            return True
-
-        path_mangled = self._mangle_path(markdown_file)
-        return any(filter in path_mangled for filter in self.filters)
-
     def _run_mdtests_for_file(self, markdown_file: Path) -> None:
         path_mangled = self._mangle_path(markdown_file)
         test_name = f"mdtest__{path_mangled}"
@@ -243,8 +236,7 @@ class MDTestRunner:
                 )
 
             for path in new_md_files | changed_md_files:
-                if self._matches_filter(path):
-                    self._run_mdtests_for_file(path)
+                self._run_mdtests_for_file(path)
 
 
 def main() -> None:


crates/ty_python_semantic/mdtest.py
```diff
@@ -214,25 +230,37 @@ def watch(self) -> Never:
...
+                    self._run_mdtest(self.filters)
             elif new_md_files:
                 files = " ".join(file.as_posix() for file in new_md_files)
                 self._recompile_tests(
                     f"New Markdown test [yellow]{files}[/yellow] detected, recompiling tests..."
                 )
 
             for path in new_md_files | changed_md_files:
-                self._run_mdtests_for_file(path)
+                if self._matches_filter(path):
```
sharkdp: I guess one could even argue that the filter shouldn't apply here? That is, if a MD file is changed that doesn't match the filter, maybe it makes sense to still run it for that file.

AlexWaygood: yeah, I think it makes sense to run any mdtest that changed, even if a filter was provided?

crates/ty_python_semantic/mdtest.py
```diff
@@ -165,9 +181,9 @@ def _print_trimmed_cargo_test_output(self, output: str, test_name: str) -> None:
 
             print(line)
 
-    def watch(self) -> Never:
+    def watch(self):
```
AlexWaygood, line 177: isn't the type annotation still accurate? :P

AlexWaygood (approved):
Nice!

In case I can nerd-snipe you -- the feature I'd most like added to mdtest.py would be a way to request it to re-run our whole mdtest suite (e.g. by entering `rerun` to stdin?) without me having to terminate the script and re-run it (which usually involves a "useless" recompilation of the test binary)

crates/ty_python_semantic/mdtest.py
sharkdp: I think it was never accurate. Both ty and pyright complained. There's no way for a type checker to see that the loop runs forever.

AlexWaygood: Oh I see, because it's a `for` loop rather than a `while` loop? Makes sense.

AlexWaygood: You could add `assert False, "Should never get here"` at the bottom of the function to shut them up, I guess :P

PR #21424, opened 2025-11-13
[ty] Use the return type of `__get__` for descriptor lookups even when `__get__` is called with incorrect arguments

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/633.

Without this, we infer the type of `numpy.iinfo("int").max` and `numpy.iinfo("int").min` as `property` rather than `int` on https://github.com/astral-sh/ruff/pull/21401, which causes a _significant_ number of diagnostics across the ecosystem.

## Test Plan

Mdtests updated

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/descriptor_protocol.md
+++ b/crates/ty_python_semantic/resources/mdtest/descriptor_protocol.md
@@ -699,7 +699,7 @@ reveal_type(C.metaclass_access)  # revealed: bytes
 #
 # However, we use the return-type of `__get__` as the inferred type anyway:
 # the way to specify that the descriptor object itself is returned when the
-# attribute is accessed on the instance or the class is by overload `__get__`.
+# attribute is accessed on the instance or the class is by overloading `__get__`.
 #
 # Using the return type of `__get__` even for `__get__` calls that have invalid
 # arguments passed to them avoids false positives in situations where there are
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/descriptor_protocol.md
```diff
@@ -696,8 +696,16 @@ reveal_type(C().instance_access)  # revealed: str
 reveal_type(C.metaclass_access)  # revealed: bytes
 
 # TODO: These should emit a diagnostic
-reveal_type(C().class_object_access)  # revealed: TailoredForClassObjectAccess
-reveal_type(C.instance_access)  # revealed: TailoredForInstanceAccess
+#
+# However, we use the return-type of `__get__` as the inferred type anyway:
+# the way to specify that the descriptor object itself is returned when the
+# attribute is accessed on the instance or the class is by overload `__get__`.
```
AlexWaygood:
```suggestion
# attribute is accessed on the instance or the class is by overloading `__get__`.
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -3954,16 +3954,24 @@ impl<'db> Type<'db> {
...
+                        } else {
+                            UnionType::from_elements(db, [bindings.return_type(db), self])
+                        }
+                    }
+                    // TODO: an error when calling `__get__` will lead to a `TypeError` or similar at runtime;
+                    // we should emit a diagnostic here instead of silently ignoring the error.
+                    Err(CallError(kind, bindings)) => {
+                        if kind == CallErrorKind::NotCallable {
+                            return None;
+                        }
```
AlexWaygood: actually, not even this makes sense. If you have a `__get__` method that isn't callable, then accessing an attribute with that `__get__` method will always raise an exception at runtime. So it makes more sense for us to infer `Unknown` rather than infer the type of the bad descriptor object:

```pycon
>>> class Descriptor:
...     __get__ = None
...     
>>> class Foo:
...     desc = Descriptor()
...     
>>> Foo().desc
    Foo().desc
TypeError: 'NoneType' object is not callable
```

PR #21427, opened 2025-11-13
[ty] Press 'enter' to rerun all mdtests

sharkdp:
## Summary

Allow users of `mdtest.py` to press enter to rerun all mdtests without recompiling (thanks @AlexWaygood).

I swear I tried three other approaches (including a fully async version) before I settled on this solution. It is indeed silly, but works just fine.

## Test Plan

Interactive playing around

### DIFF
--- a/crates/ty_python_semantic/mdtest.py
+++ b/crates/ty_python_semantic/mdtest.py
@@ -13,6 +13,8 @@ import argparse
 import json
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Yes!!

carljm:
Thank you!! This solves my biggest irritation with `mdtest.py`.

PR #21430, opened 2025-11-13
[ty] diagnostic on overridden `__setattr__` and `__delattr__` in frozen dataclasses

thejchap:
## Summary
https://github.com/astral-sh/ty/issues/111

this pr adds an `invalid-dataclass-override` diagnostic when a custom `__setattr__` or `__delattr__` is defined on a dataclass where `frozen=True` ([docs](https://docs.python.org/3/library/dataclasses.html#frozen-instances))

### Runtime exception
```
  File "/Users/justinchapman/src/ty-playground/main.py", line 4, in <module>
    @dataclass(frozen=True)
     ~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/justinchapman/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/lib/python3.13/dataclasses.py", line 1295, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
                          frozen, match_args, kw_only, slots,
                          weakref_slot)
  File "/Users/justinchapman/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/lib/python3.13/dataclasses.py", line 1157, in _process_class
    func_builder.add_fns_to_class(cls)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/justinchapman/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/lib/python3.13/dataclasses.py", line 516, in add_fns_to_class
    raise TypeError(error_msg)
TypeError: Cannot overwrite attribute __setattr__ in class A
```

### Diagnostic
```
error[invalid-dataclass-override]: Cannot overwrite attribute __setattr__ in class A
 --> /Users/justinchapman/src/ty-playground/main.py:6:5
  |
4 | @dataclass(frozen=True)
5 | class A:
6 |     def __setattr__(self, name: str, value: object) -> None: ...
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
info: __setattr__
info: rule `invalid-dataclass-override` is enabled by default

Found 1 diagnostic
```

## Test Plan
- new mdtests
- e2e
- the `attrs` mypy primer diff looks to be a [true positive](https://github.com/python-attrs/attrs/blob/main/tests/test_setattr.py#L373) - the other results have been unpredictable and have changed every time i pushed new code, even if the diagnostic logic didn't change...

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -3896,8 +3896,7 @@ impl<'db> StaticClassLiteral<'db> {
             if let Some(Type::FunctionLiteral(literal)) = attr.place.ignore_possibly_undefined()
                 && matches!(name.as_str(), "__setattr__" | "__delattr__")
             {
-                if let Some(CodeGeneratorKind::DataclassLike(_)) =
-                    CodeGeneratorKind::from_static_class(db, self, None)
+                if let CodeGeneratorKind::DataclassLike(_) = field_policy
                     && self.has_dataclass_param(db, field_policy, DataclassFlags::FROZEN)
                 {
                     if let Some(builder) = context.report_lint(


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -387,10 +387,9 @@ If `__setattr__()` or `__delattr__()` is defined in the class, we should emit a
 from dataclasses import dataclass
 
 @dataclass(frozen=True)
-class MyFrozenClass:
+class MyFrozenClass:  # error: [cannot-overwrite-attribute]
```
thejchap: TODO: move to method definition

sharkdp:
Thank you for your contribution. I did not have time to review this in detail yet. One question upfront: would it be reasonable to use the existing `invalid-override` diagnostic here instead?

thejchap:
@sharkdp i am not sure - one thing that seems like it could be nice over time is to have better diagnostic info for this specific case (for example highlighting the `frozen` argument similar to the runtime error) which wouldn't be applicable to the more generic LSP violation that `invalid-method-override` is referring to

i will defer to you though - does seem reasonable!

AlexWaygood:
I think it's better to have a different error code here. `invalid-method-override` is specifically about Liskov violations, which this isn't. I've been wondering about renaming it to `unsound`-method-override, since there are obviously lots of other ways in which method overrides can be invalid.

AlexWaygood:
I think it would be good to give the error code a name that makes it clear it's specifically about dataclasses, though, which the documentation of the rule makes clear

carljm:
We may add similar rules in future for things you're not allowed to override on a tuple subclass. Would we want those to use the same rule as the dataclass case, or a distinct rule for tuples? That may influence how we name this rule.

carljm:
(Having said that, I think maybe a different rule, since the tuple case won't be things that raise at runtime, just things that could cause unsoundness. So it would be reasonable to want to disable the one rule but not the other; and they should maybe have different severities. So I think I'd support naming this rule in a dataclass-specific way.)

thejchap:
@AlexWaygood took another stab at naming based on how i interpreted your and Carl's comments

cc @carljm - i wasn't sure the best way to pull these [comments](https://github.com/astral-sh/ruff/pull/21820/files#diff-a4d813e6b59314ddbc2448d1133969548760611a9de2d07a0c36a0cbe2c93ed3R2352) over from #21820. i did refactor `has_dataclass_param` to be a method on `ClassLiteral` instead of a closure in `own_synthesized_member`, but i didn't change any of the actual logic - if you want to elaborate a bit more on the refactor you had in mind i can either try to implement it here, or just add a comment

AlexWaygood:
Thank you! Sorry for the back-and-forth on naming... maybe `invalid-dataclass-override` might be good?
- The name clearly indicates that it's dataclass-related
- But it's a bit shorter than `unsound-dataclass-method-override`
- And I'm not sure "unsound" is the best adjective to use here, because to me it implies that this is detecting sort-of an abstract theoretical issue rather than something that would raise `TypeError` at runtime. That's why I _want_ to rename Liskov violations to `unsound-method-override` -- because the error code that detects Liskov violations _is_ kinda abstract and theoretical. But this error code is _different_ to the Liskov error code -- it's not really abstract and theoretical, it's detecting something that's going to cause `TypeError` to always be raised at runtime. So the stronger adjective `invalid` seems more fitting here.

thejchap:
@AlexWaygood ah! thanks for clarifying. sounds good - done

carljm:
@thejchap sorry for the slow response time! We've been getting swamped since the beta and this got lost in my notifications. I would say don't worry about those comments for this PR, let's just get the functionality right, better to separate things when separable. Those comments are mostly about making sure we handle various kinds of dataclass-transforms correctly, and that's kind of orthogonal to this PR anyway.

thejchap:
Sounds good - I think this is ready for re-review then - let me know if I'm missing anything!

MichaReiser:
Thank you for your work on feature. Almost the entire team is out this week. It may take a few days before someone finds time to review your PR. Happy holidays.

thejchap:
@MichaReiser thanks! Happy holidays

thejchap:
@AlexWaygood @MichaReiser hi! Just wanted to check on this one - I had been interested in continuing down this road on the diagnostics for overridden comparison dunder methods on order=True dataclasses as well so was curious on any feedback on the direction in this PR

AlexWaygood:
Sorry, we still have a bit of a PR backlog from the Christmas holidays :-(

I haven't forgotten about this PR and I promise we will get to it. I've been trying to munch through review of as many PRs as possible this week.

thejchap:
Ah no worries! Thanks for the update

carljm (approved):
Looks great! Sorry again for the really slow review, we definitely aim to do better than this. This PR just hit the exact wrong time frame with beta release and then holidays :)

PR #21434, opened 2025-11-13
[ty] fix panic with direct recursive type alias

mtshiba:
## Summary

I found that trying to use a directly recursive type alias like this would result in a stack overflow:

```python
type Itself = Itself

def foo(Itself: Itself):
    x: Itself
```

This should be reported as an invalid type alias, but a stack overflow should be fixed immediately anyway. We now consider the value type to be `Unknown` in such cases.

## Test Plan

New test case in `mdtest\pep695_type_aliases.md`

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
@@ -244,6 +244,16 @@ def f(x: IntOr, y: OrInt):
         reveal_type(x)  # revealed: Never
     if not isinstance(y, int):
         reveal_type(y)  # revealed: Never
+
+# TODO emit a diagnostic on this line
+type Itself = Itself
+
+def foo(Itself: Itself):
+    x: Itself
+
+# A type alias defined with invalid recursion behaves as a dynamic type.
+foo(42)
+foo("hello")
 ```
 
 ### With legacy generic
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -11486,7 +11486,15 @@ impl<'db> PEP695TypeAliasType<'db> {
         let type_alias_stmt_node = scope.node(db).expect_type_alias();
         let definition = self.definition(db);
 
-        definition_expression_type(db, definition, &type_alias_stmt_node.node(&module).value)
+        let ty =
+            definition_expression_type(db, definition, &type_alias_stmt_node.node(&module).value);
+        // A type alias where a value type points to itself is meaningless and would lead to infinite recursion. Replace it with `Unknown`.
+        if ty == Type::TypeAlias(TypeAliasType::PEP695(self)) {
+            // TODO: Emit a diagnostic for invalid recursive type alias.
+            Type::unknown()
+        } else {
+            ty
+        }
     }
 
     fn apply_function_specialization(self, db: &'db dyn Db, ty: Type<'db>) -> Type<'db> {


crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -244,6 +244,16 @@ def f(x: IntOr, y: OrInt):
         reveal_type(x)  # revealed: Never
     if not isinstance(y, int):
         reveal_type(y)  # revealed: Never
+
+# TODO emit a diagnostic on this line
+type Itself = Itself
```
MichaReiser, line 249: Does this fix also work for compound types like list[Itself]?

carljm: Those generally already work, without this PR: https://play.ty.dev/a65c2d75-e9c1-4a74-85e3-f96db2dd9c60

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
```diff
@@ -244,6 +244,16 @@ def f(x: IntOr, y: OrInt):
...
+
+def foo(Itself: Itself):
+    x: Itself
```
carljm, line 252: IMO it's a bit odd to test that you can annotate a parameter as `Itself`, and then use that parameter again as an annotation. It may work due to how we treat `Unknown` in a type annotation, but I don't think this is something we want to commit to in a test.

I would probably just have done this instead:

```suggestion
def foo(x: Itself):
    reveal_type(x)  # Unknown
```

Is there a particular reason why you think it's important to test this "annotated value as annotation" behavior?

carljm (approved):
Thank you!

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
carljm: Another case that may be worth testing is multi-step direct recursion:

```
type A = B
type B = A
```

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
mtshiba: In fact, in that example, no panic occurs, and the revealed type is `Never`.

https://play.ty.dev/4732e30a-301c-4753-b95f-aed34cf08585

Should the original code be placed in the corpus as a regression test?

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
mtshiba: Hmm, unfortunately, it seems the panic still persists in the case of mutually recursive aliases:

```python
type A = B
type B = A

def bar(a: B):
    x: a
```

Therefore, the measure is insufficient, and we should check the consequences of eagerly expanding a type alias value type before we first use of it.
This may be an issue that needs to be resolved in #20566.

mtshiba:
Detecting mutually recursive invalid type aliases was harder than expected, and I decided to address it in #20566.

Related commits:
https://github.com/astral-sh/ruff/pull/20566/commits/ac44afa7390034422bb521393095cc02394eff6d
https://github.com/astral-sh/ruff/pull/20566/commits/2a9b9a26db73c058994f3c58d8de324ae3366005

I will close this PR.

crates/ty_python_semantic/resources/mdtest/pep695_type_aliases.md
carljm: Ah, I wondered if this part was necessary to reproduce the panic. That explains why I didn't catch this case in my previous work on recursive type aliases. Makes it much less likely people will run into this, but of course it should still be fixed.

I think it's OK for a test like this to be in mdtests, I would just expect some commentary (specifically about annotating with a value type) that "this is a very strange thing to do, but this is a regression test to ensure it doesn't panic"

PR #21436, opened 2025-11-13
[ty] Check method definitions on subclasses for Liskov violations

AlexWaygood:
## Summary

A basic implementation of Liskov assignability.

Currently we only check methods on subclasses: not attributes or properties. These should also obviously be checked as well, but they're deferred for now. Similarly, even for methods, there is much that is currently deferred: no checks are done for staticmethods or classmethods for now, and we only check definitions inferred as `Type::FunctionLiteral`s.

The main thing missing from the checks implemented here is that the diagnostics don't do a good job of explaining _why_ a particular override violates the Liskov Substitution Principle. Other type checkers do a great job at this: they'll tell you that the return annotation is invalid, or that you've changed a parameter name, etc. All of that is very difficult for us to implement until something like https://github.com/astral-sh/ruff/pull/19580 is implemented to propagate up information about _why_ an assignability or subtyping check failed. I think we'll definitely need something like that eventually, but I'm loath to do that until the new constraint solver has landed, since it'll touch the same area of code and cause a lot of merge conflicts!

Another thing to note (which looks like it comes up a lot in the ecosystem) is that other type checkers are less strict than us in some respects. For example, pyright ignores Liskov issues stemming from a parameter on a method override having the wrong name _if_ the method is a dunder method. We could consider pushing "parameter renaming issues" into a separate (disabled-by-default) error code, but once again this is very difficult to do without something like https://github.com/astral-sh/ruff/pull/19580.

## Test Plan

Mdtests and snapshots


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -548,9 +548,10 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             "Inferring deferred types should not add more deferred definitions"
         );
 
-        // TODO: Only call this function when diagnostics are enabled.
-        self.check_class_definitions();
-        self.check_overloaded_functions(node);
+        if self.db().should_check_file(self.file()) {
+            self.check_class_definitions();
+            self.check_overloaded_functions(node);
+        }
```
AlexWaygood, line 554: This is an attempt to reduce the performance hit from this PR. A few regressions went down a few percentage points as a result of this change, but I wouldn't say this made a huge impact on perf -- it seemed to improve things fairly significantly in terms of memory usage, however.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -3772,6 +3777,74 @@ impl<'db> VarianceInferable<'db> for ClassLiteral<'db> {
...
+
+        if let Some(module) = file_to_module(self.db, file) {
+            let module_name = module.name(self.db);
+            name_parts.push(module_name.as_str().to_string());
+        }
+
+        name_parts.reverse();
+        name_parts
+    }
+}
```
AlexWaygood, line 3847: (this method used to be called `qualified_name_components`, it used to be a method on `ClassLiteral`, and it used to be in `display.rs`)

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -275,52 +276,6 @@ impl<'db> ClassLiteral<'db> {
...
         }
     }
-
-    /// Returns the components of the qualified name of this class, excluding this class itself.
-    ///
-    /// For example, calling this method on a class `C` in the module `a.b` would return
-    /// `["a", "b"]`. Calling this method on a class `D` inside the namespace of a method
-    /// `m` inside the namespace of a class `C` in the module `a.b` would return
-    /// `["a", "b", "C", "<locals of function 'm'>"]`.
-    fn qualified_name_components(self, db: &'db dyn Db) -> Vec<String> {
```
AlexWaygood, line 515: (moved to `class.rs` for better reusability)

carljm:
Not sure what's up with the ConstraintSet mdtests failing only on MacOS?

crates/ty_python_semantic/resources/mdtest/attributes.md
```diff
@@ -1904,6 +1904,7 @@ we only consider the attribute assignment to be valid if the assigned attribute
 from typing import Literal
 
 class Date:
+    # error: [invalid-method-override]
```
carljm, line 1907: Not necessary to decide in this PR, but I wonder if we should special-case ignoring `object.__setattr__` for Liskov purposes (and separately implement validation of a correct `__setattr__` signature). TBH I'm not even sure why typeshed has `object.__setattr__`, and I don't think there is any soundness need for subclasses of `object` to respect Liskov compatibility with the `object.__setattr__` definition in typeshed.

AlexWaygood:
> Not sure what's up with the ConstraintSet mdtests failing only on MacOS?

I have no idea. I haven't touched any code to do with constraint sets in this PR, nor have I added any MacOS-specific code! The test also passes locally for me on my MacBook, but seems to be failing very persistently in CI on this PR.

@dcreager, any idea why this PR might be causing the test to start failing here...? The only significant way I think that this PR could impact other areas of ty is that we have to fetch the type for all attributes defined in the body of every class we check (so we're doing a lot more types in every file we check).

MichaReiser:
Do we iterate over hash maps or rely on salsa ids for ordering?

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+
+The Liskov Substitution Principle provides the basis for many of the assumptions a type checker
+generally makes about types in Python:
+
+> Subtype Requirement: Let ⁠`ϕ(x)`⁠ be a property provable about objects ⁠`x`⁠ of type `T`. Then
+> ⁠`ϕ(y)`⁠ should be true for objects ⁠`y` of type `S` where `S` is a subtype of `T`.
+
+In order for a type checker's assumptions to be sound, it is crucial for the type checker to enforce
+the Liskov Substitution Principle on code that it checks. In practice, this usually manifests as
+three checks for a type checker to perform when it checks a subclass `B` of a class `A`:
```
carljm: https://youtube.com/clip/UgkxowksZQfdnOH7rL3v9FItjHWy72y6AhrX?si=n5rja_tvodOpIM3I

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+
+1. Read-only attributes should only ever be overridden covariantly: if a property `A.p` resolves to
+    `int` when accessed, accessing `B.p` should either resolve to `int` or a subtype of `int`.
+1. Method return types should only ever be overidden covariantly: if a method `A.f` returns `int`
+    when called, calling `B.f` should also resolve to `int or a subtype of`int\`.
+1. Method parameters should only ever be overridden contravariantly: if a method `A.f` can be called
+    with an argument of type `bool`, then the method `B.f` must also be callable with type `bool`
+    (though it is permitted for the override to also accept other types)
+1. Mutable attributes should only ever be overridden invariantly: if a mutable attribute `A.attr`
+    resolves to type `str`, it can only be overidden on a subclass with exactly the same type.
```
carljm:
```suggestion
    resolves to type `str`, it can only be overridden on a subclass with exactly the same type.
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+
+```pyi
+class Super:
+    def method(self, x: int, /): ...
+
+class Sub1(Super):
+    def method(self, x: int, /): ...  # fine
+
+class Sub2(Super):
+    def method(self, x: object, /): ...  # fine: `method` still accepts any argument of type `bool`
```
carljm:
```suggestion
    def method(self, x: object, /): ...  # fine: `method` still accepts any argument of type `int`
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+    def method(self, x: int, *, extra_kw_only_arg=42): ...  # fine
+
+class Sub12(Super):
+    # Some calls permitted by the superclass are now no longer allowed
+    # (the method can no longer be passed any arguments!)
+    def method(self, /): ...  # error: [invalid-method-override]
+
+class Sub13(Super):
+    # Some calls permitted by the superclass are now no longer allowed
+    # (the method can no longer be passed with exactly one argument!)
```
carljm:
```suggestion
    # (the method can no longer be passed exactly one argument!)
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+class Sub15(Super):
+    # Some calls permitted by the superclass are now no longer allowed
+    # (x can no longer be passed any integer -- it now requires a bool!)
+    def method(self, x: bool, /): ...  # error: [invalid-method-override]
+
+class Super2:
+    def method2(self, x): ...
+
+class Sub16(Super2):
+    def method2(self, x, /): ...  # error: [invalid-method-override]
```
carljm, line 107: At this point it feels a bit arbitrary that we've stopped explaining the errors...

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,284 @@
...
+    def method(self, x, *args, y, **kwargs): ...  # error: [invalid-method-override]
+```
+
+## The entire class hierarchy is checked
+
+If a child class's method definition is Liskov-compatible with the method definition on its parent
+class, Liskov compatibility must also nonetheless be checked with respect to the method definition
+on its grandparent class. This is because type checkers will treat the child class as a subtype of
+the grandparent class just as much as they treat it as a subtype of the parent class, so
+substitutability with respect to the grandparent class is just as important:
```
carljm, line 143: Not that it needs to be specifically mentioned here, but I think the strongest case for this is gradual types, because it means you can have cases where each individual parent-child relationship passes Liskov checking, but the grandparent-child relationship does not:

```py
class Grandparent:
    x: int

class Parent(Grandparent):
    x: Any

class Child(Parent):
    x: str
```

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -0,0 +1,127 @@
...
+    },
+};
+
+pub(super) fn check_class<'db>(context: &InferContext<'db, '_>, class: ClassLiteral<'db>) {
+    let db = context.db();
+    if class.is_known(db, KnownClass::Object) {
+        return;
+    }
+
+    let class_specialized = class.identity_specialization(db);
```
carljm, line 25: I think it would be good to add some tests for method overrides involving generics. The behavior _should_ fall out from callable subtyping, but it wouldn't hurt to have some such tests for overrides specifically.

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use crate::{
+    place::Place,
+    semantic_index::place_table,
+    types::{
+        ClassBase, ClassLiteral, ClassType, KnownClass, Type,
+        class::CodeGeneratorKind,
+        context::InferContext,
+        diagnostic::report_invalid_method_override,
+        ide_support::{MemberWithDefinition, all_declarations_and_bindings},
```
carljm, line 15: I think this PR should idealy move this stuff out of `ide_support`? It's clearly no longer just for IDE support.

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    let Place::Defined(type_on_instance, _, _) =
+        Type::instance(db, class).member(db, &member.name).place
+    else {
+        return;
+    };
+
+    for supercls in class.iter_mro(db).skip(1).filter_map(ClassBase::into_class) {
+        let Place::Defined(type_on_supercls, _, _) =
+            Type::instance(db, supercls).member(db, &member.name).place
```
carljm: Is there an `own_instance_member` method we could use here instead, to short-circuit this "find and then skip if not in our own scope" dance?

It feels odd (and potentially inefficient) that if we have a long inheritance chain of length N, and a method defined at the top of it (say on `object`) and then overridden at the leaf, we will do `O(N^2)` mro traversal, as we (wastefully) traverse the full MRO of every intervening type in the inheritance chain, only to find the same super-class method on `object` every time, and then skip it because it's not defined in our own scope.

Or alternatively, if we can't use an `own_instance_member`, what about checking the place table before doing the `member` lookup?

The `break` below gives a false sense of efficiency. If the method doesn't exist anywhere in the MRO, we will only traverse the MRO once, which is good -- but the same would be true if we did the "outer" loop once and did `own_instance_member` for each type in the MRO. When the method does exist somewhere up the MRO, this way does a lot more traversal.

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -0,0 +1,127 @@
...
+        let Some(supercls_as_callable) = type_on_supercls.try_upcast_to_callable(db) else {
+            continue;
+        };
+
+        if type_on_instance.is_assignable_to(db, supercls_as_callable) {
+            continue;
+        }
+
+        // If the function was originally defined elsewhere and simply assigned
+        // in the body of the class here, we cannot use the range associated with the `FunctionType`
```
carljm: Arguably in this case we might ideally show both ranges, since you'd want to know a) what is the actual signature of the function? as well as b) why is this function in my MRO?

carljm (commented):
Nice!

crates/ty_python_semantic/src/types/liskov.rs
AlexWaygood: I think I came to the same realisation and actually pushed this change before you posted this comment :-)

crates/ty_python_semantic/resources/mdtest/liskov.md
AlexWaygood: oh no

AlexWaygood:
```suggestion
several checks for a type checker to perform when it checks a subclass `B` of a class `A`:
```

carljm: "Among the checks a type checker will perform"

crates/ty_python_semantic/resources/mdtest/attributes.md
AlexWaygood: Yes, I think you're right.

I was also wondering about special-case ignoring `object.__hash__`. It's obviously common to define unhashable classes in Python, and it's often the correct thing to do to add `__hash__ = None` if your class is mutable. Having a type checker yell at you every time you have to do that has never felt helpful to me.

I can propose them both (either together or separately) as followups. I'd rather keep them out of this PR, since either special case (while, IMO, well-motivated) would be a deviation from the behaviour of other type checkers.

crates/ty_python_semantic/src/types/liskov.rs
AlexWaygood: > Is there an `own_instance_member` method we could use here instead, to short-circuit this "find and then skip if not in our own scope" dance?

There is a method by that name, yes. But it emulates looking up an attribute directly in `self.__dict__` (it does not include class attributes defined directly in the class body). We'd have to call `own_class_member` as well as `own_instance_member` to do this correctly.

But looking at this again, this would have the advantage that it would include synthesized members, since `own_class_member` calls `own_synthesized_member`. I think we're currently incorrectly skipping over synthesized members (from dataclasses/namedtuples/etc.) on superclasses, because they don't appear in the symbol table for those superclasses.

So your suggestion probably brings correctness benefits as well as performance benefits. Thanks!

carljm: `Type::member` is Salsa-cached, but I think that doesn't help here as much as we'd hope? The internal own-member stuff that is done for each class in the MRO is not cached, so we are still doing the `N^2` walk -- but we cache the final lookup result for each class in the MRO. Which is useful if that attribute is later looked up on that superclass, but still wasted if not.

dcreager:
> @dcreager, any idea why this PR might be causing the test to start failing here...? The only significant way I think that this PR could impact other areas of ty is that we have to fetch the type for all attributes defined in the body of every class we check (so we're doing a lot more types in every file we check).

It might cause the typevar ordering to be different, which would cause the internal BDD structure to change. I'm seeing this on another PR, too, so I will try to push up a separate fix for it.

dcreager:
> It might cause the typevar ordering to be different, which would cause the internal BDD structure to change. I'm seeing this on another PR, too, so I will try to push up a separate fix for it.

https://github.com/astral-sh/ruff/pull/21524 fixes the macos CI failures I was getting on https://github.com/astral-sh/ruff/pull/21414. I was getting a slightly different test failure than you, but my hope is that they had the same underlying cause. You can cherry-pick that commit if you want to test soon, or wait till that lands on `main`.

crates/ty_python_semantic/resources/mdtest/liskov.md
AlexWaygood: I'm adding this as an example, I think it's useful!

crates/ty_python_semantic/src/types/liskov.rs
AlexWaygood: great point. Added lots more tests in https://github.com/astral-sh/ruff/pull/21436/commits/8c03f3f0ea3b631e6791dbcd0904c1d54703f53b.

crates/ty_python_semantic/src/types/liskov.rs
AlexWaygood: I do agree, but I'm a bit loath to do that in this PR because it'll dramatically increase the merge conflicts (and because this PR is already a big PR). I'd prefer to do that as an immediate followup, if that's okay

crates/ty_python_semantic/src/types/liskov.rs
AlexWaygood: Okay, I've tried to optimise this a bit. I'm a bit reticent to call `own_class_member` directly, because (at least for now) we only care about the type accessed on instances of the class. To get the class-member-as-viewed-by-instances, you'd really need to call `invoke_descriptor_protocol` (which invokes `own_class_member`) rather than `own_class_member` directly, and calling `Type::invoke_descriptor_protocol` undermines any optimisation effect we'd get from avoiding calling `Type::member()` on the nominal-instance type (it'll recurse up the class hierarchy until it finds a class that has the attribute).

For now I've moved the `place_table` check above the `Type::member` call, so that we check whether it's defined in the class body before we do the `Type::member` call. But that won't work in the future when we extend Liskov checking to also check implicit instance attributes (etc.)

AlexWaygood: The good news is that the perf regression on the latest version of this PR is indeed much reduced -- now at <7%!

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1934,6 +1939,84 @@ declare_lint! {
     }
 }
 
+declare_lint! {
+    /// ## What it does
+    /// Detects method overrides that violate the [Liskov Substitution Principle] ("LSP").
+    ///
+    /// The LSP states that any subtype should be substitutable for its supertype.
```
carljm:
```suggestion
    /// The LSP states that an instance of a subtype should be substitutable for an instance of its supertype.
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,518 @@
...
+
+The Liskov Substitution Principle provides the basis for many of the assumptions a type checker
+generally makes about types in Python:
+
+> Subtype Requirement: Let ⁠`ϕ(x)`⁠ be a property provable about objects ⁠`x`⁠ of type `T`. Then
+> ⁠`ϕ(y)`⁠ should be true for objects ⁠`y` of type `S` where `S` is a subtype of `T`.
+
+In order for a type checker's assumptions to be sound, it is crucial for the type checker to enforce
+the Liskov Substitution Principle on code that it checks. In practice, this usually manifests as
+three checks for a type checker to perform when it checks a subclass `B` of a class `A`:
```
carljm:
```suggestion
several checks for a type checker to perform when it checks a subclass `B` of a class `A`:
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,518 @@
...
+
+1. Read-only attributes should only ever be overridden covariantly: if a property `A.p` resolves to
+    `int` when accessed, accessing `B.p` should either resolve to `int` or a subtype of `int`.
+1. Method return types should only ever be overidden covariantly: if a method `A.f` returns `int`
+    when called, calling `B.f` should also resolve to `int or a subtype of`int\`.
+1. Method parameters should only ever be overridden contravariantly: if a method `A.f` can be called
+    with an argument of type `bool`, then the method `B.f` must also be callable with type `bool`
+    (though it is permitted for the override to also accept other types)
+1. Mutable attributes should only ever be overridden invariantly: if a mutable attribute `A.attr`
+    resolves to type `str`, it can only be overidden on a subclass with exactly the same type.
```
carljm:
```suggestion
    resolves to type `str`, it can only be overridden on a subclass with exactly the same type.
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,518 @@
...
+
+class D[T](A[T]):
+    def method(self, x: object) -> None: ...  # fine
+
+class E(A[int]):
+    def method(self, x: object) -> None: ...  # fine
+
+class F[T](A[T]):
+    # TODO: we should emit `invalid-method-override` on this:
+    # `str` is not necessarily a subtype of `T`!
```
carljm:
```suggestion
    # `str` is not necessarily a supertype of `T`!
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,518 @@
...
+
+`NamedTuple` classes and dataclasses both have methods generated at runtime that do not have
+source-code definitions. There are several scenarios to consider here:
+
+1. A synthesized method on a superclass is overridden by a "normal" (not synthesized) method on a
+    subclass
+1. A "normal" method on a superclass is overridden by a synthesized method on a subclass
+1. A synthesized method on a superclass is overridden by a synthesized method on a subclass
+1. No methods are overridden, but the Liskov Substitution Principle is arguably violated anyway
+    because a generated method on a superclass is incompatible with subclassing(!)
```
carljm: As I said in Discord, I think that this case falls outside the definition of Liskov that is within the scope of a type-checker to validate, so I would not even list it here.

The scope of Liskov checking for a type checker is if the types of methods or attributes are incompatible with a base class; it's based on type signatures, not arbitrary behavior of the method. If the subclass does not override the method, there cannot be a Liskov violation of incompatible method signature.

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -0,0 +1,518 @@
...
+
+# TODO: Although this class does not override any methods of `Foo`, it nonetheless
+# arguably violates the Liskov Substitution Principle. Instances of `Bar3` cannot be
+# substituted wherever an instance of `Foo` is expected, because the generated
+# `__lt__` method on `Foo` raises an error unless the r.h.s. and `l.h.s.` have exactly
+# the same `__class__` (it does not permit instances of `Foo` to be compared with
+# instances of subclasses of `Foo`). We could therefore consider treating all
+# `order=True` dataclasses as implicitly `@final` in order to enforce Liskov soundness.
+# Note that no other type checker catches this error as of 2025-11-21.
+class Bar3(Foo): ...
```
carljm: I don't think we should emit any diagnostic here; this is not a type error. If anything, we could consider a future type-aware lint rule that should have a distinct code and severity, if it seems important in practice to alert people to this runtime behavior that is outside the type system. But I don't think this should be a TODO here.

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -0,0 +1,163 @@
...
+            method_kind = match class_kind {
+                Some(CodeGeneratorKind::NamedTuple) => {
+                    MethodKind::Synthesized(SynthesizedMethodKind::NamedTuple)
+                }
+                Some(CodeGeneratorKind::DataclassLike(_)) => {
+                    MethodKind::Synthesized(SynthesizedMethodKind::Dataclass)
+                }
+                Some(CodeGeneratorKind::TypedDict) => {
+                    MethodKind::Synthesized(SynthesizedMethodKind::TypedDict)
+                }
```
carljm, line 116: It's not valid to define methods on a `TypedDict` (or subclass of a `TypedDict`) at all -- we should emit a diagnostic on that (separate PR). Seems like it then shouldn't be possible to hit this case (until we are checking non-methods on the subclass).

carljm (approved):
Nice!

There were some comments from previous review accidentlly left not addressed; I just unresolved those prior comments rather than adding new ones.

MichaReiser:
Ohh, I would have needed this feature today when working on ty_benchmark where claude messed up the overrides...

PR #21437, opened 2025-11-13
[ty] Type inference for genererator expressions

sharkdp:
## Summary

Add type inference for (async) generator expressions.

closes https://github.com/astral-sh/ty/issues/1510

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/literal/collections/generator_expressions.md
+++ b/crates/ty_python_semantic/resources/mdtest/literal/collections/generator_expressions.md
@@ -21,18 +21,17 @@ for s in squares:
     reveal_type(s)  # revealed: int
 ```
 
-When explicitly annotated, the yield type can be widened:
+`GeneratorType` is covariant in its yielded type, so it can be used where a wider yielded type is
+expected:
 
 ```py
-from types import GeneratorType
+from typing import Iterator
 
-maybe_ints: GeneratorType[int | None, None, None] = (x**2 for x in range(10))
+def process_numbers(x: Iterator[float]): ...
 
-def _():
-    reveal_type(maybe_ints)  # revealed: GeneratorType[int | None, None, None]
-
-    for s in maybe_ints:
-        reveal_type(s)  # revealed: int | None
+numbers = (x for x in range(10))
+reveal_type(numbers)  # revealed: GeneratorType[int, None, None]
+process_numbers(numbers)
 ```
 
 ## Async generators


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -7392,33 +7392,43 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
     }
 
     /// Infer the type of the `iter` expression of the first comprehension.
-    fn infer_first_comprehension_iter(&mut self, comprehensions: &[ast::Comprehension]) {
+    fn infer_first_comprehension_iter(&mut self, comprehensions: &[ast::Comprehension]) -> bool {
```
AlexWaygood: What about

```diff
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index d9894f9c54..8cef32101a 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -7392,14 +7392,19 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
     }
 
     /// Infer the type of the `iter` expression of the first comprehension.
-    fn infer_first_comprehension_iter(&mut self, comprehensions: &[ast::Comprehension]) -> bool {
+    /// Returns the evaluation mode (async or sync) of the comprehension.
+    fn infer_first_comprehension_iter(&mut self, comprehensions: &[ast::Comprehension]) -> EvaluationMode {
         let mut comprehensions_iter = comprehensions.iter();
         let Some(first_comprehension) = comprehensions_iter.next() else {
             unreachable!("Comprehension must contain at least one generator");
         };
         self.infer_standalone_expression(&first_comprehension.iter, TypeContext::default());
 
-        first_comprehension.is_async
+        if first_comprehension.is_async {
+            EvaluationMode::Async
+        } else {
+            EvaluationMode::Sync
+        }
     }
 
     fn infer_generator_expression(&mut self, generator: &ast::ExprGenerator) -> Type<'db> {
@@ -7411,7 +7416,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             parenthesized: _,
         } = generator;
 
-        let is_async = self.infer_first_comprehension_iter(generators);
+        let evaluation_mode = self.infer_first_comprehension_iter(generators);
 
         let scope_id = self
             .index
@@ -7420,7 +7425,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         let inference = infer_scope_types(self.db(), scope);
         let yield_type = inference.expression_type(elt.as_ref());
 
-        if is_async {
+        if evaluation_mode.is_async() {
             KnownClass::AsyncGeneratorType
                 .to_specialized_instance(self.db(), [yield_type, Type::none(self.db())])
         } else {
```

crates/ty_python_semantic/resources/mdtest/literal/collections/generator_expressions.md
```diff
@@ -1,6 +1,63 @@
...
+from types import GeneratorType
+
+maybe_ints: GeneratorType[int | None, None, None] = (x**2 for x in range(10))
+
+def _():
+    reveal_type(maybe_ints)  # revealed: GeneratorType[int | None, None, None]
+
+    for s in maybe_ints:
+        reveal_type(s)  # revealed: int | None
+```
```
AlexWaygood: I don't think this demonstrates anything specific about generator expressions, does it? I think this is just our general behaviour that we always respect the declared type of a variable over its narrowed local type when we observe the variable from an external scope (in this case, we're viewing the global variable from the perspective of the function scope).

If I add a `reveal_type` below the function scope, in the global scope, it looks like we reveal the narrowed type of `GeneratorType[int, None, None]` (which I think is correct, since `GeneratorType` is covariant -- there's no benefit from avoiding narrowing the type here)

AlexWaygood (approved):

crates/ty_python_semantic/resources/mdtest/literal/collections/generator_expressions.md
sharkdp: > I don't think this demonstrates anything specific about generator expressions, does it?

It demonstrates that the inferred type of `(x**2 for x in range(10))` is assignable to the declared type `GeneratorType[int | None, None, None]`. I agree that the `def _()` below could probably be removed. I can do that, if you prefer.

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Oh, yes. I forgot about my neat little enum.

crates/ty_python_semantic/resources/mdtest/literal/collections/generator_expressions.md
AlexWaygood: I thought at first this test was trying to demonstrate something to do with literal promotions. But then I read through the code and couldn't see anything to do with literal promotions or bidirectional context (which is good, we shouldn't need any for this class 😄). And then I came back to the test and saw that that the only reason why we revealed `int | None` for the yield type was because the `reveal_type` was in the nested scope.

Maybe something like this would be clearer?

```py
def f(x: Iterator[int | None]): ...

gen = (x for x in range(42))
reveal_type(gen)  # revealed: GeneratorType[int, None, None]
f(gen)  # okay; `GeneratorType` is covariant and `int` is assignable to `int | None`
```

PR #21438, opened 2025-11-13
[ty] print function names for bare Signatures opportunistically

Gankra:
This is a followup to #21417 that improves some places where we pretty-display types (hence being restricted to `multiline` printing). Even in cases where we can't get the name it behooves us to add a `def _` to fix python syntax highlighting in e.g. hovers.

### DIFF
--- a/crates/ty_ide/src/docstring.rs
+++ b/crates/ty_ide/src/docstring.rs
@@ -65,11 +65,10 @@ impl Docstring {
     /// Render the docstring for markdown display
     pub fn render_markdown(&self) -> String {
         let trimmed = documentation_trim(&self.0);
-        // TODO: now actually parse it and "render" it to markdown.
-        //
-        // For now we just wrap the content in a plaintext codeblock
-        // to avoid the contents erroneously being interpreted as markdown.
-        format!("```text\n{trimmed}\n```")
+
+        // Try to parse and render the contents as markdown,
+        // and if we fail, wrap it in a codeblock and display it raw.
+        try_render_markdown(&trimmed).unwrap_or_else(|| format!("```text\n{trimmed}\n```"))
     }
 
     /// Extract parameter documentation from popular docstring formats.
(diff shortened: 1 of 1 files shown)


Gankra:
With the two PRs combined, stdlib calls become readable on hover, if a bit innaccurate about methods:

Screenshot 2025-11-13 at 4 35 24 PM

carljm:
Signatures don't necessarily come from `def` statements (or functions), so I feel like it's kind of misleading to prepend `def _` arbitrarily to signature rendering. I understand that this gives us syntax highlighting for free :/ Probably worth it, but unfortunate.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1111,6 +1126,18 @@ impl DisplaySignature<'_> {
...
+        // If we're multiline printing and a name hasn't been emitted, try to
+        // make one up to make things more pretty
+        if multiline && !self.settings.disallow_signature_name {
+            writer.write_str("def ")?;
+            if let Some(definition) = self.definition
+                && let Some(name) = definition.name(self.db)
+            {
+                writer.write_str(&name)?;
+            } else {
+                writer.write_str("_")?;
```
MichaReiser, line 1138: What's an example where we don't have a name? Is it for `Callable` types? 

Pylance uses a different rendering for callable types:

Screenshot 2025-11-16 at 09 08 40

Which I think I prefer, as it emphasizes the fact that one is a callable rather than a specific function instance

dhruvmanila: Yes, it's for `Callable` and `lambda` expressions and we do use the same display approach.

dhruvmanila:
I'll put this in draft for now as this might not be necessary, refer to my comment here: https://github.com/astral-sh/ruff/pull/21417#discussion_r2533146691.

Gankra:
Yeah I'm gonna send this to hell for now

PR #21439, opened 2025-11-13
[ty] Improve literal promotion heuristics

ibraheemdev:
## Summary

Extends literal promotion to apply to any generic method, as opposed to only generic class constructors. This PR also improves our literal promotion heuristics to only promote literals in non-covariant position in the return type, and avoid promotion if the literal is present in non-covariant position in any argument type.

Resolves https://github.com/astral-sh/ty/issues/1357.


carljm:
It looks like this is causing a new stack overflow in the ecosystem?

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -366,7 +285,7 @@ class Z[T]:
 
 z1: Z[Any] = Z(1)
 # TODO: This should reveal `Z[Any]`.
```
carljm: Do we have a task open for this TODO? It does seem like inference of generic parameters is a case where we should fully respect the type context.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -366,7 +285,7 @@ class Z[T]:
...
-reveal_type(z1)  # revealed: Z[int]
+reveal_type(z1)  # revealed: Z[Literal[1]]
```
carljm: We don't promote here because `Z` is covariant, right?

crates/ty_python_semantic/resources/mdtest/literal_promotion.md
```diff
@@ -5,14 +5,191 @@
...
+def f6[T](x: T) -> Invariant[T] | Contravariant[T] | None: ...
+def f7[T](x: T) -> Covariant[T] | Contravariant[T] | None: ...
+def f8[T](x: T) -> Invariant[T] | Covariant[T] | None: ...
+def f9[T](x: T) -> tuple[Invariant[T], Invariant[T]] | None: ...
+
+reveal_type(Bivariant(1))  # revealed: Bivariant[Literal[1]]
+reveal_type(Covariant(1))  # revealed: Covariant[Literal[1]]
+
+reveal_type(Contravariant(1))  # revealed: Contravariant[int]
+reveal_type(Contravariant(1))  # revealed: Contravariant[int]
```
carljm: Looks like you probably meant this, rather than a duplicated line?
```suggestion
reveal_type(Invariant(1))  # revealed: Invariant[int]
```

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
ibraheemdev: Yes.

crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
ibraheemdev: This is fixed by https://github.com/astral-sh/ruff/pull/21442.

crates/ty_python_semantic/src/types.rs
```diff
@@ -3376,6 +3380,79 @@ impl<'db> Type<'db> {
...
+        tcx: TypeContext<'db>,
+        polarity: TypeVarVariance,
+        f: &mut dyn FnMut(BoundTypeVarInstance<'db>, Type<'db>, TypeVarVariance, TypeContext<'db>),
+        visitor: &SpecializationVisitor<'db>,
+    ) {
+        let Type::NominalInstance(instance) = self else {
+            match self {
+                Type::Union(union) => {
+                    for element in union.elements(db) {
+                        element.visit_specialization(db, tcx, &mut *f);
```
carljm: Should we be calling `visit_specialization_impl` and passing along `visitor` here? Otherwise I think we can still hit stack overflow if there is a cycle mediated by a union type?

crates/ty_python_semantic/src/types.rs
```diff
@@ -3376,6 +3380,79 @@ impl<'db> Type<'db> {
...
+                    }
+                }
+                Type::Intersection(intersection) => {
+                    for element in intersection.positive(db) {
+                        element.visit_specialization(db, tcx, &mut *f);
```
carljm: Same here.

crates/ty_python_semantic/src/types.rs
```diff
@@ -3376,6 +3380,79 @@ impl<'db> Type<'db> {
...
+                    }
+                }
+                Type::TypeAlias(alias) => visitor.visit(self, || {
+                    alias.value_type(db).visit_specialization(db, tcx, f);
```
carljm: And here.

carljm (approved):
This looks great!

PR #21440, opened 2025-11-13
[`flake8-use-pathlib`] Mark fixes unsafe for return type changes (`PTH104`, `PTH105`, `PTH109`, `PTH115`)

### LINKED ISSUE #21431: PTH fixes should be unsafe when they change return types
### Summary

The fixes for [`os-rename` (PTH104)](https://docs.astral.sh/ruff/rules/os-rename/), [`os-replace` (PTH105)](https://docs.astral.sh/ruff/rules/os-replace/), [`os-getcwd` (PTH109)](https://docs.astral.sh/ruff/rules/os-getcwd/), and [`os-readlink` (PTH115)](https://docs.astral.sh/ruff/rules/os-readlink/) should generally be marked unsafe because the replacements have different return types. [Example](https://play.ruff.rs/7e724491-845c-4667-bda4-cb66fd4e6783):
```console
$ cat >pth1.py <<'# EOF'
import os
import sys

# code   name               original return type -> replacement

# PTH104 os-rename          None -> Path
print(os.rename("pth1.py", "pth1.py.bak"))

# PTH105 os-replace         None -> Path
print(os.replace("pth1.py.bak", "pth1.py"))

# PTH109 os-getcwd          str -> Path
#        os-getcwdb         bytes -> Path
try: print(os.getcwd().upper())
except AttributeError as e: print(e)
try: print(os.getcwdb().upper())
except AttributeError as e: print(e)

# PTH115 os-readlink        str -> Path
try: print(os.readlink(sys.executable).upper())
except AttributeError as e: print(e)
# EOF

$ ruff --isolated check pth1.py --select PTH --preview --unsafe-fixes --fix
Found 5 errors (5 fixed, 0 remaining).

$ python pth1.py
pth1.py.bak
pth1.py
'PosixPath' object has no attribute 'upper'
'PosixPath' object has no attribute 'upper'
'PosixPath' object has no attribute 'upper'
```

The fixes are still safe in two cases. When the function call is the top-level expression in its statement, the return value is not used, so the return type does not matter, so the fix is safe. This case is common enough and easy enough to detect that I think it is worth supporting. When the original function returns a `str` or `bytes` and the replacement returns a `Path` and the function call is an argument to a function that interprets that argument as a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object) and whose return value doesn’t expose the difference, then the difference in return type doesn’t matter, so the fix is safe. This case is harder to support because it requires reviewing each such function carefully, but if the ecosystem report reveals a common function that could be safe, it might be worth supporting it.

This all applies to [`os-path-abspath` (PTH100)](https://docs.astral.sh/ruff/rules/os-path-abspath/), [`os-path-expanduser` (PTH111)](https://docs.astral.sh/ruff/rules/os-path-expanduser/), and [`os-path-dirname` (PTH120)](https://docs.astral.sh/ruff/rules/os-path-dirname/) too. Their fixes are already marked unsafe, but their documentation should explain this additional reason for unsafety.

### Version

danparizher:
## Summary

Marks fixes as unsafe when they change return types (`None` → `Path`, `str`/`bytes` → `Path`, `str` → `Path`), except when the call is a top-level expression.

Fixes #21431.

## Problem

Fixes for `os.rename`, `os.replace`, `os.getcwd`/`os.getcwdb`, and `os.readlink` were marked safe despite changing return types, which can break code that uses the return value.

## Approach

Added `is_top_level_expression_call` helper to detect when a call is a top-level expression (return value unused). Updated `check_os_pathlib_two_arg_calls` and `check_os_pathlib_single_arg_calls` to mark fixes as unsafe unless the call is a top-level expression. Updated PTH109 to use the helper for applicability determination.

## Test Plan

Updated snapshots for `preview_full_name.py`, `preview_import_as.py`, `preview_import_from.py`, and `preview_import_from_as.py` to reflect unsafe markers.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
@@ -91,17 +91,13 @@ pub(crate) fn check_os_pathlib_single_arg_calls(
 
         let edit = Edit::range_replacement(replacement, range);
 
-        let fix = match applicability {
-            Applicability::Unsafe => Fix::unsafe_edits(edit, [import_edit]),
-            _ => {
-                let determined_applicability = match applicability {
-                    Applicability::DisplayOnly => Applicability::DisplayOnly,
-                    _ if checker.comment_ranges().intersects(range) => Applicability::Unsafe,
-                    _ => applicability,
-                };
-                Fix::applicable_edits(edit, [import_edit], determined_applicability)
-            }
+        let applicability = match applicability {
+            Applicability::DisplayOnly => Applicability::DisplayOnly,
+            _ if checker.comment_ranges().intersects(range) => Applicability::Unsafe,
+            _ => applicability,
         };
+
+        let fix = Fix::applicable_edits(edit, [import_edit], applicability);
 
         Ok(fix)
     });
@@ -176,16 +172,16 @@ pub(crate) fn check_os_pathlib_two_arg_calls(
                 format!("{binding}({path_code}).{attr}({second_code})")
             };
 
-            let determined_applicability = if checker.comment_ranges().intersects(range) {
-                Applicability::Unsafe
-            } else {
-                applicability
+            let applicability = match applicability {
+                Applicability::DisplayOnly => Applicability::DisplayOnly,
+                _ if checker.comment_ranges().intersects(range) => Applicability::Unsafe,
+                _ => applicability,
             };
 
             Ok(Fix::applicable_edits(
                 Edit::range_replacement(replacement, range),
                 [import_edit],
-                determined_applicability,
+                applicability,
             ))
         });
     }
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_getcwd.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_getcwd.rs
@@ -1,12 +1,13 @@
+use ruff_diagnostics::{Applicability, Edit, Fix};
+use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::ExprCall;
+use ruff_text_size::Ranged;
+
 use crate::checkers::ast::Checker;
 use crate::importer::ImportRequest;
 use crate::preview::is_fix_os_getcwd_enabled;
 use crate::rules::flake8_use_pathlib::helpers::is_top_level_expression_call;
 use crate::{FixAvailability, Violation};
-use ruff_diagnostics::{Applicability, Edit, Fix};
-use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::ExprCall;
-use ruff_text_size::Ranged;
 
 /// ## What it does
 /// Checks for uses of `os.getcwd` and `os.getcwdb`.
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_islink.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_islink.rs
@@ -1,3 +1,4 @@
+use ruff_diagnostics::Applicability;
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_python_ast::ExprCall;
 
@@ -5,7 +6,6 @@ use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_path_islink_enabled;
 use crate::rules::flake8_use_pathlib::helpers::check_os_pathlib_single_arg_calls;
 use crate::{FixAvailability, Violation};
-use ruff_diagnostics::Applicability;
 
 /// ## What it does
 /// Checks for uses of `os.path.islink`.
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_readlink.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_readlink.rs
@@ -86,8 +86,7 @@ pub(crate) fn os_readlink(checker: &Checker, call: &ExprCall, segments: &[&str])
         return;
     }
 
-    let fix_enabled = is_fix_os_readlink_enabled(checker.settings());
-    let applicability = if fix_enabled && !is_top_level_expression_call(checker) {
+    let applicability = if !is_top_level_expression_call(checker) {
         // Unsafe because the return type changes (str/bytes -> Path)
         Applicability::Unsafe
     } else {
@@ -99,7 +98,7 @@ pub(crate) fn os_readlink(checker: &Checker, call: &ExprCall, segments: &[&str])
         call,
         "readlink()",
         "path",
-        fix_enabled,
+        is_fix_os_readlink_enabled(checker.settings()),
         OsReadlink,
         applicability,
     );
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_replace.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_replace.rs
@@ -1,3 +1,7 @@
+use ruff_diagnostics::Applicability;
+use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::ExprCall;
+
 use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_replace_enabled;
 use crate::rules::flake8_use_pathlib::helpers::{
@@ -5,9 +9,6 @@ use crate::rules::flake8_use_pathlib::helpers::{
     is_keyword_only_argument_non_default, is_top_level_expression_call,
 };
 use crate::{FixAvailability, Violation};
-use ruff_diagnostics::Applicability;
-use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::ExprCall;
 
 /// ## What it does
 /// Checks for uses of `os.replace`.
--- a/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_rename.rs
+++ b/crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_rename.rs
@@ -1,3 +1,7 @@
+use ruff_diagnostics::Applicability;
+use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::ExprCall;
+
 use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_rename_enabled;
 use crate::rules::flake8_use_pathlib::helpers::{
@@ -5,9 +9,6 @@ use crate::rules::flake8_use_pathlib::helpers::{
     is_keyword_only_argument_non_default, is_top_level_expression_call,
 };
 use crate::{FixAvailability, Violation};
-use ruff_diagnostics::Applicability;
-use ruff_macros::{ViolationMetadata, derive_message_formats};
-use ruff_python_ast::ExprCall;
 
 /// ## What it does
 /// Checks for uses of `os.rename`.
@@ -91,7 +92,7 @@ pub(crate) fn os_rename(checker: &Checker, call: &ExprCall, segments: &[&str]) {
         );
 
     // Unsafe when the fix would delete comments or change a used return value
-    let applicability = if fix_enabled && !is_top_level_expression_call(checker) {
+    let applicability = if !is_top_level_expression_call(checker) {
         // Unsafe because the return type changes (None -> Path)
         Applicability::Unsafe
     } else {


crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_getcwd.rs
```diff
@@ -85,8 +89,12 @@ pub(crate) fn os_getcwd(checker: &Checker, call: &ExprCall, segments: &[&str]) {
...
             let applicability = if checker.comment_ranges().intersects(range) {
                 Applicability::Unsafe
-            } else {
+            } else if is_top_level_expression_call(checker, call) {
+                // Safe when the call is a top-level expression (return value not used)
                 Applicability::Safe
+            } else {
+                // Unsafe because the return type changes (str/bytes -> Path)
+                Applicability::Unsafe
             };
```
ntBre: I think I would slightly prefer something like:


```suggestion
// Unsafe when the fix would delete comments or change a used return value
let applicability = if checker.comment_ranges().intersects(range) 
  || !is_top_level_expression_call(checker, call) {
                Applicability::Unsafe
            } else {
                Applicability::Safe
};
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_getcwd.rs
```diff
@@ -37,6 +38,9 @@ use ruff_text_size::Ranged;
 ///
 /// ## Fix Safety
 /// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
+/// Additionally, the fix is marked as unsafe because `os.getcwd()` and `os.getcwdb()` return `str` or `bytes`,
+/// while `Path.cwd()` returns a `Path` object. This change in return type can break code that uses the return value.
+/// The fix is safe when the function call is a top-level expression in its statement (i.e., the return value is not used).
```
ntBre: I think we could simplify this slightly to something like:

```suggestion
/// Additionally, the fix is marked as unsafe when the return value is used because the type changes
/// from `str` or `bytes` to a `Path` object.
```

I think something similar would work for the other documentation updates too.

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_full_name.py.snap
```diff
@@ -315,6 +317,7 @@ help: Replace with `Path(...).is_file()`
 22 | bbbbb = os.path.islink(p)
 23 | os.readlink(p)
 24 | os.stat(p)
+note: This is an unsafe fix and may change runtime behavior
```
ntBre: This doesn't look right. Should `isfile` be affected by these changes? I think both versions return a bool.

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_full_name.py.snap
```diff
@@ -342,6 +345,7 @@ help: Replace with `Path(...).is_symlink()`
 23 | os.readlink(p)
 24 | os.stat(p)
 25 | os.path.isabs(p)
+note: This is an unsafe fix and may change runtime behavior
```
ntBre: I think this one also shouldn't change.

crates/ruff_linter/src/rules/flake8_use_pathlib/snapshots/ruff_linter__rules__flake8_use_pathlib__tests__preview_import_as.py.snap
```diff
@@ -233,6 +233,7 @@ help: Replace with `Path(...).exists()`
 19 | bb = foo_p.expanduser(p)
 20 | bbb = foo_p.isdir(p)
 21 | bbbb = foo_p.isfile(p)
+note: This is an unsafe fix and may change runtime behavior
```
ntBre: Another one that shouldn't change, from what I can tell.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -209,3 +223,22 @@ pub(crate) fn is_argument_non_default(arguments: &Arguments, name: &str, positio
...
+    }) = checker.semantic().current_statement()
+    {
+        // Check if the call is the same expression as the statement's value
+        // We compare by checking if the call's range is contained within the child's range
+        // and if they're the same expression node
+        if let Expr::Call(call_expr) = child.as_ref() {
+            return call_expr.range() == call.range();
+        }
+    }
+    false
```
ntBre: Assuming that `call` is the `SemanticModel::current_expression`, I think this function could just be:


```suggestion
    checker.semantic().current_expression_parent().is_none()
```

danparizher:
Thanks for the feedback! I had thought that the same concepts applied to a few other rules, but it appears not after taking a closer look.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -94,12 +94,22 @@ pub(crate) fn check_os_pathlib_single_arg_calls(
...
                     Applicability::Unsafe
+                } else if applicability.is_none() {
+                    // When applicability is None, we need to determine it based on return type changes
+                    if is_top_level_expression_call(checker, call) {
+                        // Safe when the call is a top-level expression (return value not used)
+                        Applicability::Safe
+                    } else {
+                        // Unsafe because the return type changes (str/bytes -> Path or None -> Path)
+                        Applicability::Unsafe
+                    }
```
ntBre: I don't think we should do this check here. I checked all of the references for `check_os_pathlib_single_arg_calls`, and it's only used by PTH115, of the rules we're trying to fix in this PR.

I think we should just do the `is_top_level_expression_call` check in PTH115 itself.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -176,8 +186,12 @@ pub(crate) fn check_os_pathlib_two_arg_calls(
 
             let applicability = if checker.comment_ranges().intersects(range) {
                 Applicability::Unsafe
-            } else {
+            } else if is_top_level_expression_call(checker, call) {
+                // Safe when the call is a top-level expression (return value not used)
                 Applicability::Safe
+            } else {
+                // Unsafe because the return type changes (None -> Path)
+                Applicability::Unsafe
```
ntBre: As above, this helper also affects rules outside of PTH{104,105,109,115}. We should narrow the check to the affected rules and pass that applicability into the helper.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -209,3 +223,9 @@ pub(crate) fn is_argument_non_default(arguments: &Arguments, name: &str, positio
         .find_argument_value(name, position)
         .is_some_and(|expr| !expr.is_none_literal_expr())
 }
+
+/// Returns `true` if the given call is a top-level expression in its statement.
+/// This means the call's return value is not used, so return type changes don't matter.
+pub(crate) fn is_top_level_expression_call(checker: &Checker, _call: &ExprCall) -> bool {
```
ntBre: If `_call` is not used, we should remove it from the signature.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_readlink.rs
```diff
@@ -38,6 +38,8 @@ use crate::{FixAvailability, Violation};
 ///
 /// ## Fix Safety
 /// This rule's fix is marked as unsafe if the replacement would remove comments attached to the original expression.
+/// Additionally, the fix is marked as unsafe when the return value is used because the type changes
+/// from `str` to a `Path` object.
```
ntBre: nit: I think this one returns `str` or `bytes`. The annotation in the ty playground hover says [`AnyStr`](https://docs.python.org/3/library/typing.html#typing.AnyStr): https://play.ty.dev/f5f68ba2-0762-4d18-b28c-5ebef25bf9be

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_dirname.rs
```diff
@@ -42,6 +42,10 @@ use crate::{FixAvailability, Violation};
 /// As a result, code relying on the exact string returned by `os.path.dirname`
 /// may behave differently after the fix.
 ///
+/// Additionally, the fix is marked as unsafe because `os.path.dirname()` returns a `str`, while
```
ntBre: I think this is also bytes or str (`AnyStr`).

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_abspath.rs
```diff
@@ -45,6 +45,10 @@ use crate::{FixAvailability, Violation};
 /// behaviors is required, there's no existing `pathlib` alternative. See CPython issue
 /// [#69200](https://github.com/python/cpython/issues/69200).
 ///
+/// Additionally, the fix is marked as unsafe because `os.path.abspath()` returns a `str`, while
```
ntBre: This one is also `AnyStr` in ty.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_getcwd.rs
```diff
@@ -83,7 +86,10 @@ pub(crate) fn os_getcwd(checker: &Checker, call: &ExprCall, segments: &[&str]) {
                 checker.semantic(),
             )?;
 
-            let applicability = if checker.comment_ranges().intersects(range) {
+            // Unsafe when the fix would delete comments or change a used return value
+            let applicability = if checker.comment_ranges().intersects(range)
+                || !is_top_level_expression_call(checker, call)
+            {
```
ntBre, line 93: This looks good, this is the pattern I'd follow for the other rules, if possible.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_expanduser.rs
```diff
@@ -41,6 +41,10 @@ use crate::{FixAvailability, Violation};
 /// directory can't be resolved: `os.path.expanduser` returns the
 /// input unchanged, while `Path.expanduser` raises `RuntimeError`.
 ///
+/// Additionally, the fix is marked as unsafe because `os.path.expanduser()` returns a `str`, while
```
ntBre: This is also `AnyStr` in ty

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_isfile.rs
```diff
@@ -5,6 +5,7 @@ use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_path_isfile_enabled;
 use crate::rules::flake8_use_pathlib::helpers::check_os_pathlib_single_arg_calls;
 use crate::{FixAvailability, Violation};
+use ruff_diagnostics::Applicability;
```
ntBre: nit: I would group these ruff imports with the others, separate from the `crate::` imports.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_islink.rs
```diff
@@ -73,6 +74,6 @@ pub(crate) fn os_path_islink(checker: &Checker, call: &ExprCall, segments: &[&st
         "path",
         is_fix_os_path_islink_enabled(checker.settings()),
         OsPathIslink,
-        None,
+        Some(Applicability::Safe),
```
ntBre: Hmm, are there any cases left where we pass `None`? I think we could just make this a required argument. `None` and `Safe` were being treated as the same anyway, I think, so I don't see any benefit to using the Option.

ntBre (changes requested):
Thanks! This is looking better, but I had a few more suggestions.

I tried to flag all of the places the docs should say "str or bytes," but I started using [`typing.AnyStr`](https://docs.python.org/3/library/typing.html#typing.AnyStr) as a shorthand since that's what ty shows. It's technically a type variable that depends on the input, but I think it's fine just to say "str or bytes" in each of those cases.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
danparizher: Just a note on this - to have applicability passed through the helper, I had to add `#[allow(clippy::too_many_arguments)]`.

Let me know if that's fine or if you'd rather have it refactored.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
ntBre: This is not resolved.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -138,6 +139,7 @@ pub(crate) fn is_file_descriptor(expr: &Expr, semantic: &SemanticModel) -> bool
     typing::is_int(binding, semantic)
 }
 
+#[allow(clippy::too_many_arguments)]
```
ntBre: Let's use `expect` instead of `allow`. That will remind us to remove it if we ever remove arguments.


```suggestion
#[expect(clippy::too_many_arguments)]
```

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -146,6 +148,7 @@ pub(crate) fn check_os_pathlib_two_arg_calls(
     second_arg: &str,
     fix_enabled: bool,
     violation: impl Violation,
+    applicability: Option<Applicability>,
```
ntBre: I still don't see why we're using an `Option` here (https://github.com/astral-sh/ruff/pull/21440#discussion_r2539766325). If we're doing `unwrap_or(Safe)`, why don't we just have the callers pass in `Safe` to begin with?

That also applies to the single-argument helper, though I know that was pre-existing.

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_path_exists.rs
```diff
@@ -5,6 +5,7 @@ use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_path_exists_enabled;
 use crate::rules::flake8_use_pathlib::helpers::check_os_pathlib_single_arg_calls;
 use crate::{FixAvailability, Violation};
+use ruff_diagnostics::Applicability;
```
ntBre:
```suggestion
use ruff_diagnostics::Applicability;
use ruff_macros::{ViolationMetadata, derive_message_formats};
use ruff_python_ast::ExprCall;

use crate::checkers::ast::Checker;
use crate::preview::is_fix_os_path_exists_enabled;
use crate::rules::flake8_use_pathlib::helpers::check_os_pathlib_single_arg_calls;
use crate::{FixAvailability, Violation};
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_readlink.rs
```diff
@@ -5,8 +5,10 @@ use crate::checkers::ast::Checker;
 use crate::preview::is_fix_os_readlink_enabled;
 use crate::rules::flake8_use_pathlib::helpers::{
     check_os_pathlib_single_arg_calls, is_keyword_only_argument_non_default,
+    is_top_level_expression_call,
 };
 use crate::{FixAvailability, Violation};
+use ruff_diagnostics::Applicability;
```
ntBre:
```suggestion
use ruff_diagnostics::Applicability;
use ruff_macros::{ViolationMetadata, derive_message_formats};
use ruff_python_ast::{ExprCall, PythonVersion};

use crate::checkers::ast::Checker;
use crate::preview::is_fix_os_readlink_enabled;
use crate::rules::flake8_use_pathlib::helpers::{
check_os_pathlib_single_arg_calls, is_keyword_only_argument_non_default,
    is_top_level_expression_call,
};
use crate::{FixAvailability, Violation};
```

crates/ruff_linter/src/rules/flake8_use_pathlib/rules/os_replace.rs
```diff
@@ -90,6 +93,14 @@ pub(crate) fn os_replace(checker: &Checker, call: &ExprCall, segments: &[&str])
             &["src", "dst", "src_dir_fd", "dst_dir_fd"],
         );
 
+    // Unsafe when the fix would delete comments or change a used return value
+    let applicability = if fix_enabled && !is_top_level_expression_call(checker, call) {
```
ntBre: We don't need to check if the fix is enabled here. We already pass it into the helper.

crates/ruff_linter/src/rules/flake8_use_pathlib/helpers.rs
```diff
@@ -94,12 +94,13 @@ pub(crate) fn check_os_pathlib_single_arg_calls(
...
                     Applicability::Unsafe
                 } else {
-                    Applicability::Safe
+                    // applicability is Some(Applicability::Safe), use it
+                    applicability.unwrap_or(Applicability::Safe)
                 };
-                Fix::applicable_edits(edit, [import_edit], applicability)
+                Fix::applicable_edits(edit, [import_edit], determined_applicability)
             }
         };
```
ntBre: See my other comment about the `Option`, but I think we should also do something more like this to account for the possibility of `DisplayOnly` fixes, which shouldn't be "elevated" to unsafe if there are comments. This is unlikely to happen in these rules, but it's not really more code to handle it properly.

```suggestion
        let applicability = match applicability {
            Applicability::DisplayOnly => Applicability::DisplayOnly,
            _ if checker.comment_ranges().intersects(range) => Applicability::Unsafe,
            _ => applicability,
        };
```

Issue #3711, opened 2023-03-24
[feature request] Support ignore block of code with noqa


woile:
First of all, thank you so much for this tool, it's fantastic 🚀 ! 

I would like to request the possibility to support ignoring a block of code. To my surprise, this is a feature some people have needed as well from flake8, [see SO question](https://stackoverflow.com/questions/64428794/flake8-disable-linter-only-for-a-block-of-code), though it is not supported.

In my current use case, I have a bunch of templates, that are to long, and it becomes cumbersome to add `noqa` to all of them.

I was wondering if adding something like this would be possible (maybe with a different syntax):

```python
# ruff: noqa: on: E501
TEMPLATE = """
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque sit amet ligula magna. Nulla ${var1} vehicula leo, eget tincidunt leo pharetra in. Ut egestas felis et vehicula commodo. Aliquam ${var2} placerat tortor sed magna faucibus facilisis ut et mauris. Maecenas vitae sem augue. Cras egestas felis nulla, id porta nisl ${var3} vel
"""
# ruff: noqa: off: E501
```

Thanks!

charliermarsh:
I'd really like to support something like this. But we may want to couple it with a more holistic redesign of suppression comments. \cc @MichaReiser

charliermarsh:
And thank you for the kind words :)

whinee:
That the program ignores the lines under that comment.

A markdown linter named [markdownlint](https://github.com/DavidAnson/markdownlint) has [that option](https://github.com/DavidAnson/markdownlint#configuration) and if I am not mistaken, if there is no `enable` comment down the line, then the program just ignores all of it.

This program could also have an option to warm the user when an off flag/comment is unterminated (has no corresponding in switch) cuz disabling linting for the rest of the file is not recommended???? I dont knowwwwwwww

evanrittenhouse:
I'm pretty interested in this one. Obviously, pretty new to the project, but I think that the design proposed above makes sense. Not sure about any overarching design changes - @MichaReiser saw you were mentioned above, are we good to move forward on this design? Do we want to support ignoring multiple rules at once?

MichaReiser:
> @MichaReiser saw you were mentioned above, are we good to move forward on this design? Do we want to support ignoring multiple rules at once?

Technically: I recommend waiting till after #3931 lands because it significantly changes how we handle noqa comments internally. 

[...]

PR #21441, opened 2025-11-14
new module for parsing ranged suppressions

amyreese:
## Summary

This adds a new `suppression` module to the `ruff_linter` crate, similar to the suppression
module for ty, to parse comments for ruff suppression directives, such as `# ruff: disable[CODE]`.

This is just the piece for parsing and storing an IMR of the comments themselves. Next step
is to interpret parsed comments and map them to ranges, codes, etc that can be referenced
later when determining which diagnostics should be kept/filtered/whatever.

## Test Plan

- A bunch of new snapshot tests
- Added to the beginning of `check_path` so that the parsing runs on all checked files
  for performance regression testing.

Issue #3711

### DIFF
--- a/crates/ruff_linter/src/suppression.rs
+++ b/crates/ruff_linter/src/suppression.rs
@@ -1,7 +1,7 @@
 use compact_str::CompactString;
 use core::fmt;
+use ruff_python_ast::token::{TokenKind, Tokens};
 use ruff_python_ast::whitespace::indentation;
-use ruff_python_parser::{TokenKind, Tokens};
 use std::{error::Error, fmt::Formatter};
 use thiserror::Error;
 


crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,456 @@
...
+    kind: SuppressionKind,
+
+    /// The lint code being suppressed
+    code: String,
+
+    /// Range for which the suppression applies
+    range: TextRange,
+
+    /// Any comments associated with the suppression
+    comments: Vec<SuppressionComment>,
```
MichaReiser: I assume this is something that gets clearer once we have more code but it's not clear to me how a `Suppression` can have multiple comments.

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,456 @@
...
+
+use ruff_python_trivia::{CommentRanges, Cursor};
+use ruff_text_size::{TextRange, TextSize};
+use smallvec::{SmallVec, smallvec};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum SuppressionAction {
+    Disable,
+    Enable,
+    Ignore,
```
MichaReiser: I'm leaning towards skipping `ignore` for now, given that we don't have consensus on a design yet

crates/ruff_linter/src/suppression.rs
amyreese: The idea is a `Suppression` object would cover a range of the document, but for cases like `# ruff: disable` you would also need a matching `# ruff: enable` to define the range of the suppression, leading to multiple comments for a single suppression.

crates/ruff_linter/src/suppression.rs
amyreese: Sure, was not planning on supporting it in the first implementation, but wanted to keep it in mind for designing how the system/structures should work. Equally applies to `noqa` suppressions I guess.

crates/ruff_linter/src/suppression.rs
amyreese: The next step of the implementation would be looking through the list of discovered comments and start matching disable/enable comments and build a single `Suppression` object with the appropriate range, and containing references to both comments so that the future "unused suppression" diagnostic could easily find both comments to suggest removing them.

amyreese:
Updated with logic to match disable/enable suppression comments and transform them into ranged suppression objects. Currently just looks for matching indentation and codes, to get a simplified implementation that works, since looking at tokens is complicated by where dedent tokens appear relative to own-line comments at the end of blocks ([playground example](https://play.ruff.rs/3f5950bf-c969-4519-a207-1304b22d1c78)).

MichaReiser:
> Updated with logic to match disable/enable suppression comments and transform them into ranged suppression objects. Currently just looks for matching indentation and codes, to get a simplified implementation that works, since looking at tokens is complicated by where dedent tokens appear relative to own-line comments at the end of blocks ([playground example](https://play.ruff.rs/3f5950bf-c969-4519-a207-1304b22d1c78)).

Yeah, I think we have to "ignore" `Dedent`s if all that comes between the comment token and the `Dedent` token are trivia tokens (newline, comment,)

amyreese:
> Yeah, I think we have to "ignore" `Dedent`s if all that comes between the comment token and the `Dedent` token are trivia tokens (newline, comment,)

Plus we still need to do indentation checking to make a reasonable guess as to which scope the comment should be associated:

```py
def foo():
    if True:
        pass
        # here?
    # here?
# or here?
```

crates/ruff_linter/src/suppression.rs
```diff
@@ -62,101 +67,142 @@ pub(crate) struct Suppression {
...
+
+    pub(crate) fn load_from_tokens(&mut self, tokens: &Tokens) -> Suppressions {
+        let mut indents: Vec<usize> = vec![0];
+        let mut current_indent: usize = 0;
+
+        for (token_index, token) in tokens.iter().enumerate() {
+            match token {
+                token if token.kind() == TokenKind::Indent => {
+                    current_indent = token.range().len().to_usize();
+                    indents.push(current_indent);
```
MichaReiser: Could we use the `indent_level` here rather than match on the exact indent (in which case a simple `usize` counter could replace our current stack? 

Python also supports mixed space-tab indents and I never remember the exact details, but using levels over the indent-string ensures that you delegate all this to the lexer/parser.

crates/ruff_linter/src/suppression.rs
```diff
@@ -62,101 +67,142 @@ pub(crate) struct Suppression {
...
+            }
+        }
+    }
+
+    pub(crate) fn load_from_tokens(&mut self, tokens: &Tokens) -> Suppressions {
+        let mut indents: Vec<usize> = vec![0];
+        let mut current_indent: usize = 0;
+
+        for (token_index, token) in tokens.iter().enumerate() {
+            match token {
```
MichaReiser: Mathing on `token` here seems weird as it isn't an enum? I think what you want to do here is to match on `token.kind()`

amyreese: Yes

crates/ruff_linter/src/suppression.rs
amyreese: I think the problem is that if you only have an "indent level", then you can't easily match a comment to an indentation level, because indentation levels in Python can be arbitrary. This comes back to needing to determine which level a comment belongs to when it appears before a Dedent token, and I'm not aware of a way to do that other than comparing full indentation strings.  If there are helpers in ruff that can infer an indentation level to a comment, I would love to know.

> Python also supports mixed space-tab indents and I never remember the exact details

FWIW, Python does not allow mixing spaces and tabs within an indented block of code. Eg, this will produce either an `IndentationError` or a `TabError` at runtime, depending on how many spaces are used:

```
def foo():
<tab>pass
<spaces>pass
```

With modern editors and formatters, I would be surprised if mixed indents were still a problem, at least for someone at the point of using Ruff to check their code.

MichaReiser: Oh, so this is for comments at the end of a block like this:

```py
def test():
	# ruff: disable[code]
	if True:
		pass
	# ruff: enable[code] - re-enables the outer and not the inner matching ignore
```

vs 

```py

def test():
	if True:
		# ruff: disable[code]
		pass
		# ruff: enable[code] - re-enables the inner matching ignore
```

Agree, using string comparison for those is probably fine and also what we do in the formatter https://github.com/astral-sh/ruff/blob/f69f278bf12d67e17bd46b51ffcc87fb21babd3e/crates/ruff_python_formatter/src/comments/placement.rs#L603


Python does allow mixed space and tabs indent (they can even be different between statements in a block, for as long as the visual indent is the same regardless of the tab size)
https://play.ruff.rs/7a681eaf-592d-45c7-8272-1eca15252a51

amyreese: > Python does allow mixed space and tabs indent (they can even be different between statements in a block, for as long as the visual indent is the same regardless of the tab size)

gross 😅

crates/ruff_linter/src/linter.rs
```diff
@@ -138,6 +139,9 @@ pub fn check_path(
     let tokens = parsed.tokens();
     let comment_ranges = indexer.comment_ranges();
 
+    // Gather all ruff:directive suppressions
+    let _suppressions = Suppressions::from_tokens(locator.contents(), tokens);
+
```
MichaReiser: I think we should remove this before landing as it isn't used yet (but it's great to have in place up right to landing to get the benchmark resutls)

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+}
+
+impl SuppressionComment {
+    /// Return the suppressed codes as strings
+    fn codes_as_str(&self, source: &str) -> Vec<String> {
+        self.codes
+            .iter()
+            .map(|range| source.slice(range).to_string())
+            .collect()
+    }
```
MichaReiser: Allocating a `Vec` just for comparing the code and copying over the string slices is unnecessarily expensive. 

You can do something like this


```suggestion
    fn codes_as_str<'src>(&self, source: &'src str) -> impl Iterator<Item = &'src str> {
        self.codes.iter().map(|range| source.slice(range))
    }
```

And then compare the codes using `comment.codes_as_str().eq(other.codes_as_str())`


However, this will require that the codes are in the exact same order. Is this the semantics we want?

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+
+    /// Whether the comment "matches" another comment, based on indentation and suppressed codes
+    fn matches(&self, other: &SuppressionComment, source: &str) -> bool {
+        ((self.action == SuppressionAction::Enable && other.action == SuppressionAction::Disable)
```
MichaReiser: Can you make an example for why it's necessary to match `disable` comments to `enable` comments. I assumed that matching `enable` to `disable` comments should be sufficient because we always start with "enabled-by-default".

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+            || (self.action == SuppressionAction::Disable
+                && other.action == SuppressionAction::Enable))
+            && self.indent == other.indent
+            && self.codes_as_str(source) == other.codes_as_str(source)
+    }
+}
+
+#[allow(unused)]
```
MichaReiser, line 67:
```suggestion
#[expect(unused)]
```

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct Suppression {
+    /// The lint code being suppressed
+    code: String,
+
+    /// Range for which the suppression applies
+    range: TextRange,
+
+    /// Any comments associated with the suppression
+    comments: SmallVec<[SuppressionComment; 2]>,
```
MichaReiser, line 77: `SmallVec` stack-allocates up to `N` items before heap allocating a regular `Vec`. This has the advantage that it reduces the pressure on heap size. However, it can have the downside that the containing type (`Suppression`) becomes larger because the `SmallVec` requires at least `N * SuppressionComment` bytes to store the inline elements. 

The main downside of small vec is that it wastes memory when the stored elements is less than `N`. It also makes copying `Suppression` more expensive because it requires writing more bytes (e.g. when adding a `Suppression` to a `Vec`)


Given that `SuppressionComment` is a fairly chunky type (72 bytes) (rust-analyzer shows you the size on hover) and the most common case is probably to suppress exactly one code, I suggest reducing `N` to 1
```suggestion
    comments: SmallVec<[SuppressionComment; 1]>,
```

After reading the entire PR, I think this shouldn't be a `SmallVec` in the first place. Instead, I think we should use two fields:

```rust
enable_comment: Option<SuppressionComment>,
disable_comment: Option<SuppressionComment>
```

This is more explicit expresses the structure found in code.

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+            && self.indent == other.indent
+            && self.codes_as_str(source) == other.codes_as_str(source)
+    }
+}
+
+#[allow(unused)]
+#[derive(Clone, Debug)]
+pub(crate) struct Suppression {
+    /// The lint code being suppressed
+    code: String,
```
MichaReiser: Nit: Consider using `CompactStr`

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+
+    /// Range for which the suppression applies
+    range: TextRange,
+
+    /// Any comments associated with the suppression
+    comments: SmallVec<[SuppressionComment; 2]>,
+}
+
+#[derive(Clone, Debug)]
```
MichaReiser: I suggest implementing `Copy` for `InvalidSuppressionKind`. It improves the ergonomics of this type as Rust now allows you to make copys of the type without an explicit `.clone` call. `Copy` is recommended for small types (a no-brainer for anything `<= usize`; I sometimes go as high as `2 usize`) and where identity doesn't matter

```suggestion
#[derive(Copy, Clone, Debug)]
```

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+            } else {
+                self.invalid.push(InvalidSuppression {
+                    kind: InvalidSuppressionKind::Unmatched,
+                    comment: self.pending.remove(comment_index).clone(),
+                });
+            }
+        }
+    }
+
+    pub(crate) fn load_from_tokens(&mut self, tokens: &Tokens) -> Suppressions {
```
MichaReiser: Nit: I prefer to have the external visible API come before the implementation details. In this case, move `load_from_tokens` above `match_comments`. It has the advantage that, when reading the code, I can start getting an overview before I have to read the gnarly implementation details. 

I know that this might take some time getting used to coming from Python where you have to forward-declare functions and classes before you can use it. We don't need to do that in Rust, allowing us to lay out the code for better readability than the runtime requires.

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+
+impl<'a> SuppressionsBuilder<'a> {
+    pub(crate) fn new(source: &'a str) -> Self {
+        Self {
+            source,
+            ..Default::default()
+        }
+    }
+
+    fn match_comments(&mut self, current_indent: &String) {
```
MichaReiser: It's preferred to use `&str` over `&String` because you can pass a `String` reference and a static string (like `"abc"`) to a function taking a `&str`, but you can only pass a `String` reference to a function taking a `&String` argument.

```suggestion
    fn match_comments(&mut self, current_indent: &str) {
```

Overall, writing `&String` or `&Vec<T>` tends to be a code smell and you should use `&str` or `&[T]` instead

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+                    comment: self.pending.remove(comment_index).clone(),
+                });
+            }
+        }
+    }
+
+    pub(crate) fn load_from_tokens(&mut self, tokens: &Tokens) -> Suppressions {
+        let default_indent = String::new();
+        let mut current_indent: &String = &default_indent;
+        let mut indents: Vec<String> = vec![];
```
MichaReiser: We can avoid copying the idents to `String` by storing `&str`
```suggestion
        let default_indent = "";
        let mut current_indent = default_indent;
        let mut indents: Vec<&str> = vec![];
```

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+
+        // Iterate through tokens, tracking indentation, filtering trailing comments, and then
+        // looking for matching comments from the previous block when reaching a dedent token.
+        for (token_index, token) in tokens.iter().enumerate() {
+            match token.kind() {
+                TokenKind::Indent => {
+                    indents.push(self.source.slice(token).to_string());
+                    current_indent = indents.last().unwrap_or(&default_indent);
```
MichaReiser: Now that we use `&str`, which is `Copy`, we can write this as 

```suggestion
										current_indent = self.source.slice(token);
                    indents.push(current_indent);
```

Although I'd suggest removing `current_indent` altogether and instead use `indents.last().copied().unwrap_or_default()` to get the current indent. It avoids tracking the same state twice

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+            }
+        }
+
+        self.match_comments(&default_indent);
+
+        Suppressions {
+            valid: self.valid.clone(),
+            invalid: self.invalid.clone(),
+            errors: self.errors.clone(),
+        }
```
MichaReiser, line 268: You can avoid the cloning here by changing the signature of `load_from_tokens` to take a `mut self` (so that it takes an owned value, meaning, the caller won't be allowed to use `self` anymore after).

```rust
load_from_tokens(mut self, tokens: &Tokens)
```
```suggestion
        Suppressions {
            valid: self.valid,
            invalid: self.invalid,
            errors: self.errors,
        }
```

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+        self.eat_whitespace();
+
+        let action = self.eat_action()?;
+        let codes = self.eat_codes()?;
+        if codes.is_empty() {
+            return Err(ParseError::new(ParseErrorKind::MissingCodes, self.range));
+        }
+
+        self.eat_whitespace();
+        let reason = TextRange::new(self.offset(), self.range.end());
```
MichaReiser, line 343: I was a bit confused by this line. I think what you have here works because we don't need to support `# ruff:disable[RUF100] reason # fmt: off`

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+        let text_before = self.source.slice(TextRange::new(
+            self.source.line_start(self.range.start()),
+            self.range.start(),
+        ));
+        let is_own_line = text_before.chars().all(is_python_whitespace);
+        let indent = is_own_line.then(|| text_before.to_string());
```
MichaReiser: You can use `indentation` here

https://github.com/astral-sh/ruff/blob/ef8f399a65d4a6355a0140adf7ff4462e549d2ae/crates/ruff_python_ast/src/whitespace.rs#L9-L14

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+    Enable,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct SuppressionComment {
+    /// Range containing the entire suppression comment
+    range: TextRange,
+
+    /// How indented an own-line comment is, or None for trailing comments
+    indent: Option<String>,
```
MichaReiser: I'm inclined to remove `indent` from `SuppressionComment` as it is only information necessary for matching comments. I don't think we'll need it afterwards. 

It would also allow moving the indent handling out of the `SuppressionParser`, keeping the parser scope narrow to just parsing out the comment (and it's the builders responsiblity to do all the indent stuff). 


The way I would approach this is by adding a `PendingSuppressionComment` struct like this

```rust
struct PendingSuppressionComment<'src> {
	indent: &'src str,
	comment: SuppressionComment
}
```

It also has the advantage that we don't need to copy the indent to a `String`, instead we can just store the `str` slice (requires changing `pending` to `Vec<PendingSuppressionComment<'a>>`

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+        }
+    }
+
+    fn match_comments(&mut self, current_indent: &String) {
+        let mut comment_index = 0;
+
+        // for each pending comment, search for matching comments at the same indentation level,
+        // generate range suppressions for any matches, and then discard and unmatched comments
+        // from the outgoing indentation block
+        while comment_index < self.pending.len() {
```
MichaReiser: I wonder if `load_from_tokens` should store a `Vec<(&str, Vec<SuppressionComment>)` where the first item is the indent at a given level, and the second item is the suppression comments at that level.

I think this method would greatly simplify because you could just iterate over the comments at the given level instead of having to match them by indent again. 

The downside is that we allocate a few extra `Vec`'s, one for each level with a suppression comment. I think that's fine. 


Making this change should also remove the need for `PendingSuppressionComment` because we no longer need to store the `indent` anywhere (I think)

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+                let other_index = comment_index + 1 + other_index;
+                let other = &self.pending[other_index];
+
+                // record a combined range suppression from the matching comments
+                let combined_range = TextRange::new(comment.range.start(), other.range.end());
+                for code in comment.codes_as_str(self.source) {
+                    self.valid.push(Suppression {
+                        code,
+                        range: combined_range,
+                        comments: smallvec![comment.clone(), other.clone()],
```
MichaReiser: I don't think it's a big deal, but it's a bit unfortunate that we need to clone both comments here. Do we need all information from `SuppressionComment` or could we copy a subset of the information to `Suppression? (If not, as I said, I think this is completely fine)

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+# ruff: disable  # parse error!
+def bar():
+    # ruff: disable[zeta]  # unmatched
+    pass
+# ruff: enable[zeta]  # underindented
+    pass
+";
+        let parsed = parse(source, ParseOptions::from(Mode::Module)).unwrap();
+        let suppressions = Suppressions::from_tokens(source, parsed.tokens());
+        assert_debug_snapshot!(
```
MichaReiser, line 992: I find the snapshot tests here fairly hard to read due to all the byte offsets. You could consider adding a helper similar to what I added in ty that pretty prints suppression comments:

https://github.com/astral-sh/ruff/blob/ef8f399a65d4a6355a0140adf7ff4462e549d2ae/crates/ty_python_semantic/src/suppression.rs#L1087-L1159

Another alternative is to render the suppressions using our diagnostic system, similar to what we do for ty's ide tests 

https://github.com/astral-sh/ruff/blob/ef8f399a65d4a6355a0140adf7ff4462e549d2ae/crates/ty_ide/src/lib.rs#L392-L398

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,905 @@
...
+    # ruff: enable[beta,gamma]
+# ruff: enable[alpha]
+
+# ruff: disable  # parse error!
+def bar():
+    # ruff: disable[zeta]  # unmatched
+    pass
+# ruff: enable[zeta]  # underindented
+    pass
+";
```
MichaReiser, line 991: I think it would be great to have a few more tests here:

* What if we have to succinct `ruff: enable`
* What about multiple `ruff: disable` and only one `ruff: enable`
* What if the `ruff: disable` and `ruff: enable` match but the codes are in a different order?

MichaReiser (commented):
This, overall looks great. I've a few inline comment for how I would restructure the code to make it more performant or more idiomatic Rust. 


My main concern is that I see a [3% performance regression](https://codspeed.io/astral-sh/ruff/branches/suppression-parsing?utm_source=github&utm_medium=check&utm_content=details) on the linter benchmarks. I think we should try the following:

1. Can we reduce the regression by integrating the extraction into [`Indexer::from_tokens`](https://github.com/astral-sh/ruff/blob/b6bd32d9dc086e81b766019b5b99e8c343de7220/crates/ruff_python_index/src/indexer.rs#L32) by having a `visit_token` method?
2. Instead of iterating over `Tokens`, iterate over `CommentRanges`, and start iterating over tokens only the comment is a `ruff: enable` / `ruff: disable` comment. A somewhat naive implementation below

```rust
let mut indents: Vec<&str> = vec![];
        let mut errors = Vec::new();

        let mut suppressions = comments
            .iter()
            .copied()
            .filter_map(|comment_range| {
                let mut parser = SuppressionParser::new(self.source, comment_range);
                match parser.parse_comment() {
                    Ok(comment) => Some(comment),
                    Err(ParseError {
                        kind: ParseErrorKind::NotASuppression,
                        ..
                    }) => None,
                    Err(error) => {
                        errors.push(error);
                        None
                    }
                }
            })
            .peekable();

        'comments: while let Some(suppression) = suppressions.peek() {
            let last_indent = tokens
                .before(suppression.range.start())
                .iter()
                .rfind(|token| token.kind() == TokenKind::Indent)
                .map(|token| self.source.slice(token))
                .unwrap_or_default();

            indents.push(last_indent);

            let tokens = tokens.after(suppression.range.start());

            // Iterate through tokens, tracking indentation, filtering trailing comments, and then
            // looking for matching comments from the previous block when reaching a dedent token.
            for (token_index, token) in tokens.iter().enumerate() {
                let current_indent = indents.last().copied().unwrap_or_default();
                match token.kind() {
                    TokenKind::Indent => {
                        indents.push(self.source.slice(token));
                    }
                    TokenKind::Dedent => {
                        self.match_comments(current_indent);

                        indents.pop();

                        if indents.is_empty() {
                            continue 'comments;
                        }
                    }
                    TokenKind::Comment => {
                        let Some(suppression) =
                            suppressions.next_if(|suppression| suppression.range == token.range())
                        else {
                            continue;
                        };

                        let Some(indent) =
                            indentation_at_offset(suppression.range.start(), self.source)
                        else {
                            // trailing suppressions are not supported
                            self.invalid.push(InvalidSuppression {
                                kind: InvalidSuppressionKind::Trailing,
                                comment: suppression,
                            });
                            continue;
                        };

                        // comment matches current block's indentation, or precedes an indent/dedent token
                        if indent == current_indent
                            || tokens[token_index..]
                                .iter()
                                .find(|t| !t.kind().is_trivia())
                                .is_some_and(|t| {
                                    t.kind() == TokenKind::Dedent || t.kind() == TokenKind::Indent
                                })
                        {
                            self.pending.push(PendingComment {
                                comment: suppression,
                                indent,
                            });
                        } else {
                            // weirdly indented? ¯\_(ツ)_/¯
                            self.invalid.push(InvalidSuppression {
                                kind: InvalidSuppressionKind::Indentation,
                                comment: suppression,
                            });
                        }
                    }
                    _ => {}
                }
            }

            self.match_comments("");
        }

        Suppressions {
            valid: self.valid,
            invalid: self.invalid,
            errors,
        }
```

I don't have a strong preference which approach to take. I suspect that 1. is easier and probably worth trying out as it is less invasive (and also doesn't add much complexity). 2. is more complicated but I think should be doable as well. Happy to try/brainstorm on both approaches.

crates/ruff_linter/src/suppression.rs
amyreese: > I know that this might take some time getting used to coming from Python where you have to forward-declare functions and classes before you can use it. We don't need to do that in Rust, allowing us to lay out the code for better readability than the runtime requires.

That's not really true for Python. You can reference and call other functions even if they're defined later in the file, as long as at import-time you don't try to actually execute it before the runtime has gotten to the definition.

Eg, this will work just fine:

```py
def foo():
	bar()

def bar():
	...

foo()
```

It's more a style choice I picked up — for a different kind of readability — where it's easier to consume and understand a module in it's entirety if you can read it from top-to-bottom, starting with the building blocks. It helps prevent the need for jumping around in the file to understand what functions/classes are doing, because you would have read their definitions or at least their docstrings already. And if you're more interested in what the "public" API is rather than understanding the implementation, that would generally be covered by either a docstring at the top of the file, or documentation built from the source code's docstrings. 

That said, I will try to follow Ruff's/Rust's style.

MichaReiser:
As discussed in our 1:1

* I think we should move the performance optimization out of this PR. We can simply remove the call to extract the comments in this PR and address the performance concerns in a follow-up PR.
* Let's add more integration tests in https://github.com/astral-sh/ruff/pull/21623 rather than trying to improve the integration testing framework here (and extending coverage)

crates/ruff_linter/src/suppression.rs
amyreese: I wrote it both ways here because I waffled a bit on where/how to do the matching, and in which direction the search for a match would be done.

crates/ruff_linter/src/suppression.rs
amyreese: > However, this will require that the codes are in the exact same order. Is this the semantics we want?

I considered using an unordered set, or just sorting the codes found, but wasn't sure what the performance semantics of that would be when trying to match comments together, especially if there was a big file with a bunch of disable/enable comments.

From the UX perspective, I feel like "must be same order" is easy to explain, and also helps enforce the idea that you can't "mix and match" codes from various disable/enable comments.

Eg, I don't think we want to support (or make users *think* we support) this:

```
# ruff: disable[foo]
...
# ruff: disable[bar]
...
# ruff: enable[foo,bar]
```

Keeping the semantics rigid around order within matched comments seems beneficial, but I'm not opposed to allowing arbitrary orders as long as the set contents match.

MichaReiser: I'm fine keeping it as is for now and change the behavior based on user feedback.

crates/ruff_linter/src/suppression.rs
amyreese: The main reason I chose a vec rather than explicit fields was as consideration for  any logic around needing the comment ranges (like reporting unused suppressions/codes) could simply loop over any associated comments without having to know what kind the suppression is, or without needing to repeat logic within `if let Some(comment) = suppression.<kind>_comment` blocks. And in a hypothetical future where we added alternative suppression directives (eg, `#ruff:ignore`) or where we unified `#noqa` into this framework to deduplicate suppression handling logic, we wouldn't need to add new fields or further duplicate that logic.  But also maybe there's some rust tricks that make this nicer and I don't know what I'm talking about. 😅

MichaReiser: I'm fine leaving this as is in that case but a few thougths:

> The main reason I chose a vec rather than explicit fields was as consideration for any logic around needing the comment ranges (like reporting unused suppressions/codes) could simply loop over any associated comments without having to know what kind the suppression is, or without needing to repeat logic within if let Some(comment) = suppression.<kind>_comment blocks

`Suppression` could have a `comments()` method that returns an `Iterator` over al its comments for the cases where we really just want to iterate over them. I suspect that we still need special casing depending on the kind of comment when generating fixes. 

> or where we unified #noqa into this framework to deduplicate suppression handling logic, we wouldn't need to add new fields or further duplicate that logic. 

I wouldn't worry too much about that. This seems far out and there are a few options as well, E.g. we could have an `enum` depending on the comment kind that stores one or two comments (I don't think there's any case where we want to store 3?)

crates/ruff_linter/src/suppression.rs
amyreese: I'm definitely not used to thinking in terms of composing types like this. That will take more practice to reach for it more regularly.

crates/ruff_linter/src/suppression.rs
amyreese: I tried doing this earlier on and got so lost in rust compiler errors that I decided to defer it for later once I actually had the logic working 😅

crates/ruff_linter/src/suppression.rs
amyreese: > Given that SuppressionComment is a fairly chunky type (72 bytes) (rust-analyzer shows you the size on hover) and the most common case is probably to suppress exactly one code, I suggest reducing N to 1

Also I think you misunderstand the current use here. In the expected case, this vec would contain a reference to both the disable and enable `SuppressionComment` objects. Unmatched disable/enable would end up in the `invalid` pile rather than getting a `Suppression` object created.

crates/ruff_linter/src/suppression.rs
amyreese: Other than the indents — which I will move out somewhere else — the rest of the data in the comment object was intended for use by diagnostics like unused suppressions/codes.

crates/ruff_linter/src/suppression.rs
MichaReiser: > Unmatched disable/enable would end up in the invalid pile rather than getting a Suppression object created.

I thought we decided to allow implicit *enables* at the end of the block.

amyreese: Looking back at the notion doc, the summary says we decided on requiring a matching enable comment, I believe coming back to the idea of wanting to make sure that refactors didn't accidentally remove a `#ruff:enable` line and unknowingly suppress more lines than originally intended.

amyreese:
Updated with most of the suggestions, including use of `&str`, better testing snapshots, etc. Performance work to bridge comment ranges and tokens was left for a future PR. I think this should be ready to land assuming I haven't introduced other nits.

crates/ruff_linter/src/suppression.rs
MichaReiser: I'm leaning towards making this a lint rule, which we can enable by default, but giving users a way to opt out if they want to.

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,1435 @@
...
+use ruff_python_ast::whitespace::indentation;
+use ruff_python_parser::{TokenKind, Tokens};
+use std::{error::Error, fmt::Formatter};
+use thiserror::Error;
+
+use ruff_python_trivia::Cursor;
+use ruff_text_size::{Ranged, TextLen, TextRange, TextSize, TextSlice};
+use smallvec::{SmallVec, smallvec};
+
+#[allow(unused)]
```
MichaReiser, line 12: Use `expect` over `allow`. This way, rustc will remind you to remove the unused warning once the struct is used

crates/ruff_linter/src/suppression.rs
```diff
@@ -0,0 +1,1435 @@
...
+                            };
+
+                            // comment matches current block's indentation, or precedes an indent/dedent token
+                            if indent == indents.last().copied().unwrap_or_default()
+                                || tokens[token_index..]
+                                    .iter()
+                                    .find(|t| !t.kind().is_trivia())
+                                    .is_some_and(|t| {
+                                        t.kind() == TokenKind::Dedent
+                                            || t.kind() == TokenKind::Indent
```
MichaReiser: Nit
```suggestion
                                        matches!(t.kind(), TokenKind::Dedent | TokenKind::Indent)
```

crates/ruff_linter/src/suppression.rs
MichaReiser: I still think that this might be a nice simplification but up to you

MichaReiser (approved):
Thanks, this looks good. 

I suggest keeping `disable` comments without an explicit comment as valid suppressions and instead creating a new ruff dedicated lint rule that warns about this. This way, it gives users a way to opt-out from this behavior.

crates/ruff_linter/src/suppression.rs
amyreese: I tried doing that, but then I get `unfulfilled` issues because they are used by tests. 🙃  Is there a correct way to account for that?

PR #21442, opened 2025-11-14
[ty] Improve generic class constructor inference

ibraheemdev:
## Summary

We currently fail to account for the type context when inferring generic classes constructed with `__new__`, or synthesized `__init__` for dataclasses.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
+++ b/crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
@@ -324,25 +324,25 @@ class X[T]:
     def __init__(self, value: T):
         self.value = value
 
-a: X[int] = X(1)
-reveal_type(a)  # revealed: X[int]
+x1: X[int] = X(1)
+reveal_type(x1)  # revealed: X[int]
 
-b: X[int | None] = X(1)
-reveal_type(b)  # revealed: X[int | None]
+x2: X[int | None] = X(1)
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/assignment/annotations.md
```diff
@@ -324,27 +324,21 @@ class X[T]:
...
+    reveal_type(x2)  # revealed: X[T@_ | int]
 
-d: X[Any] = X(1)
-reveal_type(d)  # revealed: X[Any]
-
-def _(flag: bool):
-    # TODO: Handle unions correctly.
-    # error: [invalid-assignment] "Object of type `X[int]` is not assignable to `X[int | None]`"
-    a: X[int | None] = X(1) if flag else X(2)
-    reveal_type(a)  # revealed: X[int | None]
```
ibraheemdev: This is actually a more general issue (see https://github.com/astral-sh/ty/issues/1543), so it doesn't make sense to test it here.

crates/ty_python_semantic/src/types.rs
```diff
@@ -6340,11 +6324,19 @@ impl<'db> Type<'db> {
 
         let argument_types = infer_argument_types(bindings);
 
+        let binding_instance_ty = self_type
+            .to_instance(db)
+            .expect("type should be convertible to instance type");
```
sharkdp: Am I missing something or is this is the same as `init_ty` above?

PR #21445, opened 2025-11-14
[ty] Complete support for `ParamSpec`

dhruvmanila:
## Summary

Closes: https://github.com/astral-sh/ty/issues/157

This PR adds support for the following capabilities involving a `ParamSpec` type variable:
- Representing `P.args` and `P.kwargs` in the type system
- Matching against a callable containing `P` to create a type mapping
- Specializing `P` against the stored parameters

The value of a `ParamSpec` type variable is being represented using `CallableType` with a `CallableTypeKind::ParamSpecValue` variant. This `CallableTypeKind` is expanded from the existing `is_function_like` boolean flag. An `enum` is used as these variants are mutually exclusive.

For context, an initial iteration made an attempt to expand the `Specialization` to use `TypeOrParameters` enum that represents that a type variable can specialize into either a `Type` or `Parameters` but that increased the complexity of the code as all downstream usages would need to handle both the variants appropriately. Additionally, we'd have also need to establish an invariant that a regular type variable always maps to a `Type` while a paramspec type variable always maps to a `Parameters`.

I've intentionally left out checking and raising diagnostics when the `ParamSpec` type variable and it's components are not being used correctly to avoid scope increase and it can easily be done as a follow-up. This would also include the scoping rules which I don't think a regular type variable implements either.

## Test Plan

Add new mdtest cases and update existing test cases.

Ran this branch on pyx, no new diagnostics.

### Ecosystem analysis

There's a case where in an annotated assignment like:
```py
type CustomType[P] = Callable[...]

def value[**P](...): ...

def another[**P](...):
	target: CustomType[P] = value
```
The type of `value` is a callable and it has a paramspec that's bound to `value`, `CustomType` is a type alias that's a callable and `P` that's used in it's specialization is bound to `another`. Now, ty infers the type of `target` same as `value` and does not use the declared type `CustomType[P]`. [This is the assignment](https://github.com/mikeshardmind/async-utils/blob/0980b9d9ab2bc7a24777684a884f4ea96cbbe5f9/src/async_utils/gen_transform.py#L108) that I'm referring to which then leads to error in downstream usage. Pyright and mypy does seem to use the declared type.

There are multiple diagnostics in `dd-trace-py` that requires support for `cls`.

I'm seeing `Divergent` type for an example like which ~~I'm not sure why, I'll look into it tomorrow~~ is because of a cycle as mentioned in https://github.com/astral-sh/ty/issues/1729#issuecomment-3612279974:
```py
from typing import Callable

def decorator[**P](c: Callable[P, int]) -> Callable[P, str]: ...

@decorator
def func(a: int) -> int: ...

# ((a: int) -> str) | ((a: Divergent) -> str)
reveal_type(func)
```

I ~~need to look into why are the parameters not being specialized through multiple decorators in the following code~~ think this is also because of the cycle mentioned in https://github.com/astral-sh/ty/issues/1729#issuecomment-3612279974 and the fact that we don't support `staticmethod` properly:
```py
from contextlib import contextmanager

class Foo:
    @staticmethod
    @contextmanager
    def method(x: int):
        yield

foo = Foo()
# ty: Revealed type: `() -> _GeneratorContextManager[Unknown, None, None]` [revealed-type]
reveal_type(foo.method)
```

There's some issue related to `Protocol` that are generic over a `ParamSpec` in `starlette` which might be related to https://github.com/astral-sh/ty/issues/1635 but I'm not sure. Here's a minimal example to reproduce:

<details><summary>Code snippet:</summary>
<p>

```py
from collections.abc import Awaitable, Callable, MutableMapping
from typing import Any, Callable, ParamSpec, Protocol

P = ParamSpec("P")

Scope = MutableMapping[str, Any]
Message = MutableMapping[str, Any]
Receive = Callable[[], Awaitable[Message]]
Send = Callable[[Message], Awaitable[None]]

ASGIApp = Callable[[Scope, Receive, Send], Awaitable[None]]

_Scope = Any
_Receive = Callable[[], Awaitable[Any]]
_Send = Callable[[Any], Awaitable[None]]

# Since `starlette.types.ASGIApp` type differs from `ASGIApplication` from `asgiref`
# we need to define a more permissive version of ASGIApp that doesn't cause type errors.
_ASGIApp = Callable[[_Scope, _Receive, _Send], Awaitable[None]]


class _MiddlewareFactory(Protocol[P]):
    def __call__(
        self, app: _ASGIApp, *args: P.args, **kwargs: P.kwargs
    ) -> _ASGIApp: ...


class Middleware:
    def __init__(
        self, factory: _MiddlewareFactory[P], *args: P.args, **kwargs: P.kwargs
    ) -> None:
        self.factory = factory
        self.args = args
        self.kwargs = kwargs


class ServerErrorMiddleware:
    def __init__(
        self,
        app: ASGIApp,
        value: int | None = None,
        flag: bool = False,
    ) -> None:
        self.app = app
        self.value = value
        self.flag = flag

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None: ...


# ty: Argument to bound method `__init__` is incorrect: Expected `_MiddlewareFactory[(...)]`, found `<class 'ServerErrorMiddleware'>` [invalid-argument-type]
Middleware(ServerErrorMiddleware, value=500, flag=True)
```

</p>
</details> 

### Conformance analysis

> ```diff
> -constructors_callable.py:36:13: info[revealed-type] Revealed type: `(...) -> Unknown`
> +constructors_callable.py:36:13: info[revealed-type] Revealed type: `(x: int) -> Unknown`
> ```

Requires return type inference i.e., https://github.com/astral-sh/ruff/pull/21551

> ```diff
> +constructors_callable.py:194:16: error[invalid-argument-type] Argument is incorrect: Expected `list[T@__init__]`, found `list[Unknown | str]`
> +constructors_callable.py:194:22: error[invalid-argument-type] Argument is incorrect: Expected `list[T@__init__]`, found `list[Unknown | str]`
> +constructors_callable.py:195:4: error[invalid-argument-type] Argument is incorrect: Expected `list[T@__init__]`, found `list[Unknown | int]`
> +constructors_callable.py:195:9: error[invalid-argument-type] Argument is incorrect: Expected `list[T@__init__]`, found `list[Unknown | str]`
> ```

I might need to look into why this is happening...

> ```diff
> +generics_defaults.py:79:1: error[type-assertion-failure] Type `type[Class_ParamSpec[(str, int, /)]]` does not match asserted type `<class 'Class_ParamSpec'>`
> ```

which is on the following code
```py
DefaultP = ParamSpec("DefaultP", default=[str, int])

class Class_ParamSpec(Generic[DefaultP]): ...

assert_type(Class_ParamSpec, type[Class_ParamSpec[str, int]])
```

It's occurring because there's no equivalence relationship defined between `ClassLiteral` and `KnownInstanceType::TypeGenericAlias` which is what these types are.

Everything else looks good to me!

### DIFF
--- a/crates/ty_ide/src/hover.rs
+++ b/crates/ty_ide/src/hover.rs
@@ -2290,6 +2290,7 @@ def function():
             "#,
         );
 
+        // TODO: Should this be constravariant instead?
         assert_snapshot!(test.hover(), @r"
         P@Alias (bivariant)
         ---------------------------------------------


dhruvmanila:
There's a stack overflow in `mypy_primer` repository which I'm trying to figure out why it's happening on this branch, I've narrowed it down to the following code:

```py
from dataclasses import field

field(default_factory=dict)
```

dhruvmanila:
Hmm, not really sure where is the stack overflow being triggered, a more minimal code that behaves the same way as above but without the `dataclasses` import:

```py
from typing import Callable


def f[_T](*, default_factory: Callable[[], _T]) -> _T:
    return default_factory()


f(default_factory=dict)
```

It's important that the `default_factory` is `dict` and not any other type like `list`, `int`, etc.

dhruvmanila:
Nevermind, found the issue, it's a silly mistake on my end.

crates/ty_python_semantic/resources/mdtest/decorators.md
```diff
@@ -126,6 +126,8 @@ def custom_decorator(f) -> Callable[[int], str]:
     def wrapper(*args, **kwargs):
         print("Calling decorated function")
         return f(*args, **kwargs)
+    # TODO: This shouldn't be an error
+    # error: [invalid-return-type]
     return wrapper
```
dhruvmanila: I'm looking into this, a minimal repro:

```py
from typing import Callable


class _Wrapped[**P, R]:
    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R: ...


def update_wrapper[**P, R](wrapper: Callable[P, R]) -> _Wrapped[P, R]: ...


def wrapper(*args, **kwargs): ...


# bound method _Wrapped[(...), Unknown].__call__(**P@_Wrapped) -> Unknown
#
# This should actually reveal `(...) -> Unknown`
reveal_type(update_wrapper(wrapper).__call__)
```

dhruvmanila: Oh man, this was hard to track! I was not passing around the ParamSpec type variable correctly in all places. Fixed this in [`b358231` (#21445)](https://github.com/astral-sh/ruff/pull/21445/commits/b358231686abed80e64c6c85dd68433a9a3f0693).

crates/ty_ide/src/hover.rs
```diff
@@ -2238,12 +2236,11 @@ def function():
             "#,
         );
 
-        // TODO: This should be `P@Alias (<variance>)`
         assert_snapshot!(test.hover(), @r"
-        typing.ParamSpec
+        P@Alias (bivariant)
```
AlexWaygood, line 2295: shouldn't this be contravariant? Is that worth a TODO comment?

E.g. when we hover over `T` here, we display it as being contravariant:

```py
from typing import Callable
type X[T] = Callable[[T], int]
```

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_forms.md
```diff
@@ -66,7 +65,7 @@ def _(
     reveal_type(d)  # revealed: Unknown
 
     def foo(a_: e) -> None:
-        reveal_type(a_)  # revealed: @Todo(Support for `typing.ParamSpec`)
+        reveal_type(a_)  # revealed: @Todo(Support for `typing.ParamSpec` instances in type expressions)
```
AlexWaygood: ideally I think we'd emit a diagnostic here -- an arbitrary instance of `ParamSpec` should not be valid in a type expression. (You could add a TODO comment for now.)

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
```diff
@@ -115,3 +115,232 @@ P = ParamSpec("P", default=[A, B])
 class A: ...
 class B: ...
 ```
+
+## Validating `ParamSpec` usage
+
+`ParamSpec` is only valid as the first element to `Callable` or the final element to `Concatenate`.
```
AlexWaygood: or as a type parameter to `Protocol`/`Generic`. Maybe:

```suggestion
In type annotations, `ParamSpec` is only valid as the first element to `Callable` or the final element to `Concatenate`.
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
```diff
@@ -115,3 +115,232 @@ P = ParamSpec("P", default=[A, B])
...
+    def nested4(**kwargs: P.kwargs) -> None: ...
+
+# TODO: error
+def bar(*args: P.args, **kwargs: P.kwargs) -> None:
+    pass
+
+class Foo:
+    # TODO: error
+    def method(self, *args: P.args, **kwargs: P.kwargs) -> None: ...
+```
```
AlexWaygood, line 188: it could be worth having one test where the variadic parameters have different names, just to demonstrate that it's the _kind_ of the parameter that matters rather than the name. E.g. this is fine:

```py
class Foo(Generic[P]):
    def __call__(self, *arguments: P.args, **keyword_arguments: P.kwargs) -> None: ...
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
```diff
@@ -115,3 +115,232 @@ P = ParamSpec("P", default=[A, B])
...
+reveal_type(OnlyParamSpec[int]().attr)  # revealed: (int, /) -> None
+```
+
+But, they cannot be omitted when there are multiple type variables.
+
+```py
+reveal_type(TypeVarAndParamSpec[int, [int, str]]().attr)  # revealed: (int, str, /) -> int
+reveal_type(TypeVarAndParamSpec[int, [str]]().attr)  # revealed: (str, /) -> int
+reveal_type(TypeVarAndParamSpec[int, ...]().attr)  # revealed: (...) -> int
+reveal_type(TypeVarAndParamSpec[int, Any]().attr)  # revealed: (...) -> int
```
AlexWaygood: is this not one of the things we should emit an `[invalid-type-arguments]` diagnostic on? `Any` is not a ParamSpec, `...` or a list of types

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
```diff
@@ -115,3 +115,232 @@ P = ParamSpec("P", default=[A, B])
...
+
+# TODO: We could still specialize for `T1` as the type is valid which would reveal `(...) -> int`
+# TODO: error: paramspec is unbound
+reveal_type(TypeVarAndParamSpec[int, P2]().attr)  # revealed: (...) -> Unknown
+# error: [invalid-type-arguments] "Type argument for `ParamSpec` must be either a list of types, `ParamSpec`, `Concatenate`, or `...`"
+reveal_type(TypeVarAndParamSpec[int, int]().attr)  # revealed: (...) -> Unknown
+```
+
+Nor can they be omitted when there are more than one `ParamSpec`.
```
AlexWaygood:
```suggestion
Nor can they be omitted when there are more than one `ParamSpec`s.
```

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
```diff
@@ -115,3 +115,232 @@ P = ParamSpec("P", default=[A, B])
...
+reveal_type(p2.attr2)  # revealed: (int, /) -> None
+
+p3 = ParamSpecWithDefault4[[int], [str]]()
+reveal_type(p3.attr1)  # revealed: (int, /) -> None
+reveal_type(p3.attr2)  # revealed: (str, /) -> None
+
+# TODO: error
+# Un-ordered type variables as the default of `PAnother` is `P`
+class ParamSpecWithDefault5(Generic[PAnother, P]):
+    attr: Callable[PAnother, None]
```
AlexWaygood, line 373: it could also be worth adding a test where a class is generic over `PAnother` but _not_ generic over `P`. I think that should cause us to omit a diagnostic, since `PAnother` would have `P` as its default, but `P` wouldn't be in scope?

(It's obviously fine to just add a TODO to that test if it's not implemented yet)

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+from typing import Callable, Any
+
+def f[**P](func: Callable[P, int], *args: P.args, **kwargs: P.kwargs) -> None:
+    reveal_type(args + ("extra",))  # revealed: tuple[object, ...]
+    reveal_type(args + (1, 2, 3))  # revealed: tuple[object, ...]
+    reveal_type(args[0])  # revealed: object
+
+    reveal_type("key" in kwargs)  # revealed: bool
+    # TODO: Should be `object | None`
+    reveal_type(kwargs.get("key"))  # revealed: object
```
AlexWaygood: I don't think this TODO is accurate, because `None` is a subtype of `object`: `object | None` just immediately simplifies to `object`. The existing assertion is correct!

```suggestion
    reveal_type(kwargs.get("key"))  # revealed: object
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+reveal_type(OnlyParamSpec[int]().attr)  # revealed: (int, /) -> None
+```
+
+But, they cannot be omitted when there are multiple type variables.
+
+```py
+reveal_type(TypeVarAndParamSpec[int, [int, str]]().attr)  # revealed: (int, str, /) -> int
+reveal_type(TypeVarAndParamSpec[int, [str]]().attr)  # revealed: (str, /) -> int
+reveal_type(TypeVarAndParamSpec[int, ...]().attr)  # revealed: (...) -> int
+reveal_type(TypeVarAndParamSpec[int, Any]().attr)  # revealed: (...) -> int
```
AlexWaygood: same as in the legacy markdown test, I think this is something we should emit an error on -- `Any` is not a list of types, a ParamSpec or `...`

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+reveal_type(ParamSpecWithDefault2().attr)  # revealed: (...) -> None
+reveal_type(ParamSpecWithDefault2[int, str]().attr)  # revealed: (int, str, /) -> None
+```
+
+```py
+class ParamSpecWithDefault3[**P1, **P2 = P1]:
+    attr1: Callable[P1, None]
+    attr2: Callable[P2, None]
+
+# `P1` hasn't been specialized, so it defaults to `Unknown` gradual form
```
AlexWaygood:
```suggestion
# `P1` hasn't been specialized, so it defaults to `...` gradual form
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+def keyword_only1(*, x: int) -> int:
+    return 1
+
+def keyword_only2(*, y: int) -> int:
+    return 2
+
+# On the other hand, combining two functions with only keyword-only parameters does not have a
+# common supertype, so it should result in an error.
+# error: [invalid-argument-type] "Argument to function `multiple` is incorrect: Expected `(*, x: int) -> int`, found `def keyword_only2(*, y: int) -> int`"
+reveal_type(multiple(keyword_only1, keyword_only2))  # revealed: (*, x: int) -> bool
```
AlexWaygood, line 469: you could possibly also add an example of keyword-arguments with default values. E.g.

```py
def keyword_only_with_default_1(*, x: int = 42) -> int:
    return 1

def keyword_only_with_default_2(*, y: int = 42) -> int:
    return 2
```

In this case, the behavioural supertype of the parameter sets is `()`, since both `keyword_only_with_default_1` and `keyword_only_with_default_2` can be called with 0 arguments passed

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+    return func(*args, **kwargs)
+
+def foo2[**P2](func: Callable[P2, int], *args: P2.args, **kwargs: P2.kwargs) -> None:
+    foo1(func, *args, **kwargs)
+
+    # error: [invalid-argument-type] "Argument to function `foo1` is incorrect: Expected `P2@foo2.args`, found `Literal[1]`"
+    foo1(func, 1, *args, **kwargs)
+
+    # error: [invalid-argument-type] "Argument to function `foo1_with_extra_arg` is incorrect: Expected `str`, found `P2@foo2.args`"
+    foo1_with_extra_arg(func, *args, **kwargs)
```
AlexWaygood, line 511: and also add a call to `foo1_with_extra_arg` that _doesn't_ cause us to emit a diagnostic?

```suggestion
    # error: [invalid-argument-type] "Argument to function `foo1_with_extra_arg` is incorrect: Expected `str`, found `P2@foo2.args`"
    foo1_with_extra_arg(func, *args, **kwargs)
    
    foo1_with_extra_arg(func, "foooooo", *args, **kwargs)
```

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -62,3 +62,503 @@ Other values are invalid.
...
+
+ty will check whether the argument after `**` is a mapping type but as instance attribute are
+unioned with `Unknown`, it shouldn't error here.
+
+```py
+from typing import Callable
+
+def baz[**P](fn: Callable[P, None], foo: Foo[P]) -> None:
+    fn(*foo.args, **foo.kwargs)
+```
```
AlexWaygood, line 566: I think the "union with `Unknown`" thing is making this test a bit more confusing than it needs to be? Could you just explicitly annotate the instance attributes, so that they aren't unioned with `Unknown`?

````suggestion
```py
class Foo[**P]:
    def __init__(self, *args: P.args, **kwargs: P.kwargs) -> None:
        self.args: P.args = args
        self.kwargs: P.kwargs = kwargs

def bar[**P](foo: Foo[P]) -> None:
    reveal_type(foo)  # revealed: Foo[P@bar]
    reveal_type(foo.args)  # revealed: P@bar.args
    reveal_type(foo.kwargs)  # revealed: P@bar.kwargs
```

ty will check whether the argument after `**` is a mapping type:

```py
from typing import Callable

def baz[**P](fn: Callable[P, None], foo: Foo[P]) -> None:
    fn(*foo.args, **foo.kwargs)
    fn(*foo.kwargs, **foo.args)  # error
```
````

File: crates/ty_python_semantic/resources/mdtest/with/async.md [hunk 15]
```diff
@@ -213,7 +213,7 @@ async def connect() -> AsyncGenerator[Session]:
     yield Session()
 
 # TODO: this should be `() -> _AsyncGeneratorContextManager[Session, None]`
-reveal_type(connect)  # revealed: (...) -> _AsyncGeneratorContextManager[Unknown, None]
+reveal_type(connect)  # revealed: () -> _AsyncGeneratorContextManager[Unknown, None]
```
AlexWaygood, line 216: I guess we're nearly there on this TODO now 😅

crates/ty_python_semantic/src/types.rs
```diff
@@ -4985,11 +4985,15 @@ impl<'db> Type<'db> {
...
+                if typevar.is_paramspec(db) && matches!(name_str, "args" | "kwargs") =>
             {
-                Place::bound(todo_type!("ParamSpecArgs / ParamSpecKwargs")).into()
+                Place::declared(Type::TypeVar(match name_str {
+                    "args" => typevar.with_paramspec_attr(db, ParamSpecAttrKind::Args),
+                    "kwargs" => typevar.with_paramspec_attr(db, ParamSpecAttrKind::Kwargs),
+                    _ => unreachable!(),
+                }))
+                .into()
             }
```
AlexWaygood: nit: I think you can do this without the `unreachable!()` call:

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index c2d26b1124..a4871e093f 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -4985,14 +4985,17 @@ impl<'db> Type<'db> {
                     .into()
             }
 
-            Type::TypeVar(typevar)
-                if typevar.is_paramspec(db) && matches!(name_str, "args" | "kwargs") =>
-            {
-                Place::declared(Type::TypeVar(match name_str {
-                    "args" => typevar.with_paramspec_attr(db, ParamSpecAttrKind::Args),
-                    "kwargs" => typevar.with_paramspec_attr(db, ParamSpecAttrKind::Kwargs),
-                    _ => unreachable!(),
-                }))
+            Type::TypeVar(typevar) if name_str == "args" && typevar.is_paramspec(db) => {
+                Place::declared(Type::TypeVar(
+                    typevar.with_paramspec_attr(db, ParamSpecAttrKind::Args),
+                ))
+                .into()
+            }
+
+            Type::TypeVar(typevar) if name_str == "kwargs" && typevar.is_paramspec(db) => {
+                Place::declared(Type::TypeVar(
+                    typevar.with_paramspec_attr(db, ParamSpecAttrKind::Kwargs),
+                ))
                 .into()
             }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -8588,6 +8605,8 @@ pub struct TrackedConstraintSet<'db> {
 // The Salsa heap is tracked separately.
 impl get_size2::GetSize for TrackedConstraintSet<'_> {}
 
+// TODO: The origin is either `TypeVarInstance` or `BoundTypeVarInstance`
+
```
AlexWaygood: This TODO comment looks a bit orphaned -- what's it meant to refer to?

crates/ty_python_semantic/src/types.rs
```diff
@@ -9675,6 +9698,37 @@ impl<'db> TypeVarInstance<'db> {
...
+                    .own_tuple_spec(db)
+                    .map_or_else(Parameters::unknown, |tuple_spec| {
+                        Parameters::new(
+                            db,
+                            tuple_spec.all_elements().map(|ty| {
+                                Parameter::positional_only(None).with_annotated_type(*ty)
+                            }),
+                        )
+                    }),
+                Type::Dynamic(DynamicType::Todo(_)) => Parameters::todo(),
```
AlexWaygood: I think it would be better here to do an exhaustive match over `DynamicType` variants, because we have 3 `Todo` variants currently and we might add or remove some in the future:

```suggestion
                Type::Dynamic(dynamic) => match dynamic {
                    DynamicType::Todo(_) | DynamicType::TodoUnpack | DynamicType::TodoStarredExpression => Parameters::todo(),
                    DynamicType::Any | DynamicType::Unknown | DynamicType::Divergent(_) | DynamicType::UnknownDivergent(_) => Parameters::unknown(),
                }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9687,27 +9741,35 @@ impl<'db> TypeVarInstance<'db> {
...
-            // legacy typevar
+            // legacy typevar / ParamSpec
             DefinitionKind::Assignment(assignment) => {
                 let call_expr = assignment.value(&module).as_call_expr()?;
+                let func_ty = definition_expression_type(db, definition, &call_expr.func);
+                let known_class = func_ty.as_class_literal().and_then(|cls| cls.known(db));
                 let expr = &call_expr.arguments.find_keyword("default")?.value;
-                Some(definition_expression_type(db, definition, expr))
+                let default_type = definition_expression_type(db, definition, expr);
+                if matches!(known_class, Some(KnownClass::ParamSpec)) {
```
AlexWaygood: (optional nit)

```suggestion
                if known_class == Some(KnownClass::ParamSpec) {
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9797,14 +9874,17 @@ impl<'db> BindingContext<'db> {
 pub struct BoundTypeVarIdentity<'db> {
     pub(crate) identity: TypeVarIdentity<'db>,
     pub(crate) binding_context: BindingContext<'db>,
+    paramspec_attr: Option<ParamSpecAttrKind>,
 }
 
 /// A type variable that has been bound to a generic context, and which can be specialized to a
 /// concrete type.
 #[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
```
AlexWaygood, line 9920: for most other Salsa-interned types, we have this note about ordering:

```suggestion
/// A type variable that has been bound to a generic context, and which can be specialized to a
/// concrete type.
///
/// # Ordering
///
/// Ordering is based on the wrapped data's salsa-assigned id and not on its values.
/// The id may change between runs, or when e.g. a `BoundTypeVarInstance` was garbage-collected and recreated.
#[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
#[derive(PartialOrd, Ord)]
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9797,14 +9874,17 @@ impl<'db> BindingContext<'db> {
 pub struct BoundTypeVarIdentity<'db> {
     pub(crate) identity: TypeVarIdentity<'db>,
     pub(crate) binding_context: BindingContext<'db>,
+    paramspec_attr: Option<ParamSpecAttrKind>,
```
AlexWaygood, line 9909: nit: add some docs?

```suggestion
    /// If `Some()`, this indicates that this type variable is the `.args` or `.kwargs`
    /// attribute of a `ParamSpec`
    paramspec_attr: Option<ParamSpecAttrKind>,
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9797,14 +9874,17 @@ impl<'db> BindingContext<'db> {
...
 }
 
 /// A type variable that has been bound to a generic context, and which can be specialized to a
 /// concrete type.
 #[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
+#[derive(PartialOrd, Ord)]
 pub struct BoundTypeVarInstance<'db> {
     pub typevar: TypeVarInstance<'db>,
     binding_context: BindingContext<'db>,
+    paramspec_attr: Option<ParamSpecAttrKind>,
```
AlexWaygood, line 9926:
```suggestion
    /// If `Some()`, this indicates that this type variable is the `.args` or `.kwargs`
    /// attribute of a `ParamSpec`
    paramspec_attr: Option<ParamSpecAttrKind>,
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+    /// Returns a new bound typevar instance with the given `ParamSpec` attribute set.
+    ///
+    /// This method will also set an appropriate upper bound on the typevar, based on the
+    /// attribute kind. For `P.args`, the upper bound will be `tuple[object, ...]`, and for
+    /// `P.kwargs`, the upper bound will be `Top[dict[str, Any]]`.
+    ///
+    /// It's the caller's responsibility to ensure that this method is only called on a `ParamSpec`
+    /// type variable.
+    pub(crate) fn with_paramspec_attr(self, db: &'db dyn Db, kind: ParamSpecAttrKind) -> Self {
+        debug_assert!(self.is_paramspec(db));
```
AlexWaygood: nit: could improve the error message here if this assertion fails

```suggestion
        debug_assert!(
            self.is_paramspec(db),
            "Expected a paramspec but got a {:?}",
            self.kind(db)
        );
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+
+    /// Returns a new bound typevar instance without any `ParamSpec` attribute set.
+    ///
+    /// This method will also remove any upper bound that was set by `with_paramspec_attr`. This
+    /// means that the returned typevar will have no upper bound or constraints.
+    ///
+    /// It's the caller's responsibility to ensure that this method is only called on a `ParamSpec`
+    /// type variable.
+    pub(crate) fn without_paramspec_attr(self, db: &'db dyn Db) -> Self {
+        debug_assert!(self.is_paramspec(db));
```
```suggestion
        debug_assert!(
            self.is_paramspec(db),
            "Expected a paramspec but got a {:?}",
            self.kind(db)
        );
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+                .to_specialized_instance(db, [KnownClass::Str.to_instance(db), Type::any()])
+                .top_materialization(db),
+        });
+
+        let typevar = TypeVarInstance::new(
+            db,
+            self.typevar(db).identity(db),
+            Some(TypeVarBoundOrConstraintsEvaluation::Eager(upper_bound)),
+            None, // explicit_variance
+            None, // _default
```
```suggestion
            None, // `P.args` and `P.kwargs` cannot have defaults, even though `P` can
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+        debug_assert!(self.is_paramspec(db));
+
+        Self::new(
+            db,
+            TypeVarInstance::new(
+                db,
+                self.typevar(db).identity(db),
+                None, // Remove the upper bound set by `with_paramspec_attr`
+                None, // explicit_variance
+                None, // _default
```
```suggestion
            None, // `P.args` and `P.kwargs` cannot have defaults, even though `P` can
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+            ParamSpecAttrKind::Kwargs => KnownClass::Dict
+                .to_specialized_instance(db, [KnownClass::Str.to_instance(db), Type::any()])
+                .top_materialization(db),
+        });
+
+        let typevar = TypeVarInstance::new(
+            db,
+            self.typevar(db).identity(db),
+            Some(TypeVarBoundOrConstraintsEvaluation::Eager(upper_bound)),
+            None, // explicit_variance
```
```suggestion
            None, // ParamSpecs cannot have explicit variance
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9819,9 +9899,75 @@ impl<'db> BoundTypeVarInstance<'db> {
...
+    pub(crate) fn without_paramspec_attr(self, db: &'db dyn Db) -> Self {
+        debug_assert!(self.is_paramspec(db));
+
+        Self::new(
+            db,
+            TypeVarInstance::new(
+                db,
+                self.typevar(db).identity(db),
+                None, // Remove the upper bound set by `with_paramspec_attr`
+                None, // explicit_variance
```
```suggestion
                None, // ParamSpecs cannot have explicit variance
```

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2563,20 +2564,62 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
...
+        let variadic_type = match argument_type {
+            Some(argument_type @ Type::Union(union)) => {
+                // When accessing an instance attribute that is a `P.args`, the type we infer is
+                // `Unknown | P.args`. This needs to be special cased here to avoid calling
+                // `iterate` on it which will lose the `ParamSpec` information as it will return
+                // `object` that comes from the upper bound of `P.args`. What we want is to always
+                // use the `P.args` type to perform type checking against the parameter type. This
+                // will allow us to error when `*args: P.args` is matched against, for example,
+                // `n: int` and correctly type check when `*args: P.args` is matched against
+                // `*args: P.args`.
```
AlexWaygood: it will only be `Unknown | P.args` if the instance attribute was unannotated, as I mentioned in https://github.com/astral-sh/ruff/pull/21445#discussion_r2589923699

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -3039,6 +3117,87 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+        paramspec: BoundTypeVarInstance<'db>,
+    ) -> bool {
+        let Some(Type::Callable(callable)) = self
+            .specialization
+            .and_then(|specialization| specialization.get(self.db, paramspec))
+        else {
+            return false;
+        };
+
+        if !matches!(callable.kind(self.db), CallableTypeKind::ParamSpecValue) {
```
AlexWaygood: (optional nit)

```suggestion
        if callable.kind(self.db) != CallableTypeKind::ParamSpecValue {
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1552,100 +1574,152 @@ impl<'db> FmtDetailed<'db> for DisplaySignature<'_, 'db> {
...
-                    f.write_str(arg_separator)?;
+            ParametersKind::Gradual => {
+                // We represent gradual form as `...` in the signature, internally the parameters still
+                // contain `(*args, **kwargs)` parameters.
+                f.write_str("...")?;
+            }
+            ParametersKind::ParamSpec(typevar) => {
+                f.write_str(&format!("**{}", typevar.name(self.db)))?;
+                if let Some(name) = typevar.binding_context(self.db).name(self.db) {
+                    f.write_str(&format!("@{name}"))?;
```
```suggestion
                write!(f, "**{}", typevar.name(self.db))?;
                if let Some(name) = typevar.binding_context(self.db).name(self.db) {
                    write!(f, "@{name}")?;
```

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1380,6 +1404,10 @@ impl<'db> SpecializationBuilder<'db> {
 
         match self.types.entry(identity) {
             Entry::Occupied(mut entry) => {
+                // TODO: mypy and Pyright does this, should we keep doing it?
+                if bound_typevar.is_paramspec(self.db) {
+                    return;
+                }
```
AlexWaygood: what's "this", in this context?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2612,7 +2612,43 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                                let mut diag = builder.into_diagnostic(format_args!(
+                                    "`{name}.kwargs` is valid only in `**kwargs` annotation",
+                                ));
+                                diag.set_primary_message(format_args!(
+                                    "Did you mean `{name}.args`?"
+                                ));
+                                diagnostic::add_type_expression_reference_link(diag);
+                            }
+                            // TODO: Should this be `Unknown` instead?
+                            Type::homogeneous_tuple(self.db(), Type::unknown())
```
AlexWaygood: I think what you have currently here is correct. We do know for a fact that `*args` will _always_ be a tuple of some kind.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2612,7 +2612,43 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        }
+
+                        // `*args: P`
+                        None => {
+                            // The diagnostic for this case is handled in `in_type_expression`.
+                            // TODO: Should this be `Unknown` instead?
+                            Type::homogeneous_tuple(self.db(), Type::unknown())
```
AlexWaygood: same here, I think `tuple[Unknown, ...]` (what you currently have) is the best answer here!

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2708,11 +2744,50 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        {
+                            let name = typevar.name(self.db());
+                            let mut diag = builder.into_diagnostic(format_args!(
+                                "`{name}.args` is valid only in `*args` annotation",
+                            ));
+                            diag.set_primary_message(format_args!("Did you mean `{name}.kwargs`?"));
+                            diagnostic::add_type_expression_reference_link(diag);
+                        }
+                        // TODO: Should this be `Unknown` instead?
+                        KnownClass::Dict.to_specialized_instance(
```
AlexWaygood: and here, `dict[str, Unknown]` seems like the right answer to me; I'd just remove the TODO

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -2708,11 +2744,50 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    Some(ParamSpecAttrKind::Kwargs) => annotated_type,
+
+                    // `**kwargs: P`
+                    None => {
+                        // The diagnostic for this case is handled in `in_type_expression`.
+                        // TODO: Should this be `Unknown` instead?
+                        KnownClass::Dict.to_specialized_instance(
+                            self.db(),
+                            [KnownClass::Str.to_instance(self.db()), Type::unknown()],
+                        )
```
AlexWaygood: same here

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3341,20 +3416,81 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-            ast::Expr::EllipsisLiteral(_) => {
-                Type::single_callable(self.db(), Signature::new(Parameters::gradual_form(), None))
+    fn infer_paramspec_default(&mut self, default_expr: &ast::Expr) {
+        match default_expr {
+            ast::Expr::EllipsisLiteral(ellipsis) => {
+                // TODO: This should use `infer_type_expression` but that uses a `todo` type for
+                // inferring ellipsis literals in type expressions, which is not what we want here.
+                let ty = self.infer_ellipsis_literal_expression(ellipsis);
+                self.store_expression_type(default_expr, ty);
+                return;
```
AlexWaygood: does it _have_ to use `infer_type_expression`? We only get here in very specific situations; what you have here seems fine to me

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3341,20 +3416,81 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+            ast::Expr::EllipsisLiteral(_) => {
+                return Ok(Type::paramspec_value_callable(
+                    db,
+                    Parameters::gradual_form(),
+                ));
+            }
+
+            ast::Expr::Tuple(ast::ExprTuple { elts, .. })
+            | ast::Expr::List(ast::ExprList { elts, .. }) => {
+                // This should be taken care by the caller.
```
```suggestion
                // This should be taken care of by the caller.
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3341,20 +3416,81 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                if expr.is_tuple_expr() {
+                    debug_assert!(
+                        exactly_one_paramspec,
+                        "Inferring ParamSpec value during explicit specialization for a \
+                    tuple expression should only happen when it contains exactly one ParamSpec"
+                    );
```
AlexWaygood: nit: this assertion is probably very cheap; you could make this a non-debug assertion

```suggestion
                    assert!(
                        exactly_one_paramspec,
                        "Inferring ParamSpec value during explicit specialization for a \
                    tuple expression should only happen when it contains exactly one ParamSpec"
                    );
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3383,41 +3519,98 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+
+                match param_type {
+                    Type::TypeVar(typevar) if typevar.is_paramspec(db) => {
+                        return Ok(param_type);
+                    }
+
+                    Type::KnownInstance(known_instance)
+                        if known_instance.class(self.db()) == KnownClass::ParamSpec =>
+                    {
+                        // TODO: Raise diagnostic: "ParamSpec "P" is unbound"
```
AlexWaygood: nit

```suggestion
                        // TODO: Emit diagnostic: "ParamSpec "P" is unbound"
```

crates/ty_ide/src/hover.rs
dcreager, line 1: These edits are probably what's causing the `cargo fmt` CI failure. You'll need to revert the whitespace edits and only keep the "meaty" change below in the diff

crates/ty_python_semantic/resources/mdtest/with/async.md
dcreager: So close! I think #21551 is the rest of it!

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -318,6 +319,13 @@ impl<'db> GenericContext<'db> {
         self.variables_inner(db).values().copied()
     }
 
+    /// Returns `true` if this generic context contains exactly one `ParamSpec` type variable.
```
dcreager: To clarify — as written, this function returns whether the generic context contains exactly 1 `ParamSpec` and exactly 0 other `TypeVar`s. The comment suggests you might instead have intended to return whether it contains 1 `ParamSpec` and _0 or more_ `TypeVar`s. If you meant the former I'd suggest rewording the docs:


```suggestion
    /// Returns `true` if this generic context contains exactly one `ParamSpec`
    /// and no other type variables.
```

If you meant the latter then the method body is wrong.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -197,6 +201,69 @@ impl<'db> CallableSignature<'db> {
         tcx: TypeContext<'db>,
         visitor: &ApplyTypeMappingVisitor<'db>,
     ) -> Self {
+        if let TypeMapping::Specialization(specialization) = type_mapping
```
dcreager: We'll need this logic for `TypeMapping::PartialSpecialization` too. It might be worth pulling much of the body out into a helper method.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -3039,6 +3117,87 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+            // If we reach here, none of the arguments matched the `ParamSpec` parameter, but the
+            // `ParamSpec` could specialize to a parameter list containing some parameters.
+            self.evaluate_paramspec_sub_call(None, paramspec);
+        }
+    }
+
+    /// Try to evaluate a `ParamSpec` sub-call at the given argument index.
+    ///
+    /// If the argument at the given index matches a parameter which is a `ParamSpec`, invoke
+    /// a sub-call starting from that argument index and return `true`. Otherwise, return `false`.
```
dcreager: For these helper methods, it would be nice to have some Python snippets in the doc comment describing the particular cases that they're meant to handle.

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -2651,21 +2694,39 @@ impl<'a, 'db> ArgumentMatcher<'a, 'db> {
...
-                Some(Place::Defined(keys_method, _, Definedness::AlwaysDefined)) => keys_method
+            {
+                Place::Defined(getitem_method, _, Definedness::AlwaysDefined) => getitem_method
                     .try_call(db, &CallArguments::positional([Type::unknown()]))
                     .ok()
                     .map_or_else(Type::unknown, |bindings| bindings.return_type(db)),
                 _ => Type::unknown(),
             };
 
+            let value_type = match argument_type {
```
dcreager, line 2729: Instead of repeating the match statement here, you might consider adding this logic as a method on `VariadicArgumentType`

crates/ty_python_semantic/src/types/call/bind.rs
```diff
@@ -3081,67 +3240,92 @@ impl<'a, 'db> ArgumentTypeChecker<'a, 'db> {
...
+                )
             };
 
-            let Some(key_type) = key_type else {
-                self.errors.push(BindingError::KeywordsNotAMapping {
-                    argument_index: adjusted_argument_index,
-                    provided_ty: argument_type,
-                });
-                return;
+            let value_type = match argument_type {
```
dcreager, line 3371: Same as above — ideally we wouldn't have to repeat this match statement here, and you could add this as a method on `VariadicArgumentType`. Though there you don't have access to that type, so in this case it might not be worth the refactoring churn

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3341,20 +3416,81 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                self.store_expression_type(default_expr, ty);
+                return;
             }
             ast::Expr::List(ast::ExprList { elts, .. }) => {
+                let types = elts
+                    .iter()
+                    .map(|elt| self.infer_type_expression(elt))
+                    .collect::<Vec<_>>();
+                // N.B. We cannot represent a heterogeneous list of types in our type system, so we
+                // use a heterogeneous tuple type to represent the list of types instead.
```
dcreager, line 3428: We have considered adding a new `Type` variant to handle heterogeneous lists. I don't know if there's an open issue for it. But if we ever do that, it will probably lean heavily on `TupleSpec`, so 👍 to doing it this way for now, since that will make it most likely that this logic would carry over easily.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -3383,41 +3519,98 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    }
+
+                    // This is to handle the following case:
+                    //
+                    // ```python
+                    // from typing import ParamSpec
+                    //
+                    // class Foo[**P]: ...
+                    //
+                    // Foo[ParamSpec]  # P: (ParamSpec, /)
```
dcreager, line 3544: 🤮

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1640,6 +1668,47 @@ impl<'db> SpecializationBuilder<'db> {
                 }
             }
 
+            (Type::Callable(formal_callable), _) => {
+                if let Some(actual_callable) = actual
+                    .try_upcast_to_callable(self.db)
+                    .and_then(CallableTypes::exactly_one)
+                {
```
dcreager, line 1717: This is where the main conflict with #21551 will be. No suggested changes for this PR though!

To write down my plans for the merge conflict: #21551 adds a new `ConstraintSetAssignability` typing relation, and this logic will move over to `Type::has_relation_to_impl`, and will define what constraint set we create to describe when a callable lhs/actual is assignable to a `ParamSpec` callable rhs/formal.

dcreager (commented):
This is looking very good!

crates/ty_python_semantic/resources/mdtest/annotations/unsupported_special_forms.md
dhruvmanila: This seems easy enough to do as I just need to remove the `ParamSpec` branch in `in_type_expression`, I'll do it in this PR.

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
dhruvmanila: Yes, I should've clarified that both [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=baf1446b0bfbf475c1a3233c49d50276) and [Pyright](https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcAUCQMYA2AhgM61QBiYYA2gFTsAKANFACoBdAFxQAdBLIgApgDdp1SgH14CaQApmbInD6oYggJRA) supports specifying `Any` for a paramspec and I've also seen instances of this in the ecosystem report specifically used in `staticmethod[Any, Any]` or `classmethod[Any, Any]`.

This is similar to how `Concatenate` is allowed as well in that position, it's the `default` of a `ParamSpec` where the limitation of "ParamSpec, `...` or a list of types" exists as per the spec but I can't find any reference specifying what is allowed to pass as explicit specialization to a ParamSpec.

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
dhruvmanila: Yes, it should error because it's not in scope, I'll add the test case with a TODO as scoping rules aren't implemented in our generic implementation yet.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: See https://github.com/astral-sh/ruff/pull/21445#discussion_r2591394627, happy to revert it if you think otherwise.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: 🤦 thanks! I was a little bit confused to see this but yes it's obvious now

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: I actually meant to say that it defaults to a gradual form where the type of parameters is `Unknown` aka `(*args: Unknown, **kwargs: Unknown)` but I guess that's more of an internal details. I'll use your suggestion.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: Yeah, that makes sense, thanks!

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: You cannot use `P.args` and `P.kwargs` in other position except for to annotate `*args` and `**kwargs` parameter, so `self.args: P.args = args` isn't allowed and will raise a diagnostic.

dhruvmanila: Hmm, actually, this prompted me to try something weird because this allows you to store the "args" and "kwargs" separately and then access it outside the class like so:

```py
from typing import Callable

class Foo[**P]:
    def __init__(self, fn: Callable[P, int], *args: P.args, **kwargs: P.kwargs) -> None:
        self.fn = fn
        self.args = args
        self.kwargs = kwargs

def fn(x: int, y: str) -> int:
    return 1

foo = Foo(fn, 1, "a")

reveal_type(foo)

# Pyright reveals here `Unknown`
# ty reveals `Unknown | ((x: int, y: str))`
# mypy has an INTERNAL ERROR here: https://mypy-play.net/?mypy=master&python=3.12&gist=55fbcad1f66e947123afe9920176fd32
reveal_type(foo.args)
reveal_type(foo.kwargs)

fn(*foo.args, **foo.kwargs)
fn(*foo.args)
fn(**foo.kwargs)
```

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: Right, but you cannot annotate a variable with `P.args` or `P.kwargs`

crates/ty_python_semantic/src/types.rs
dhruvmanila: Love this, thank you!

crates/ty_python_semantic/src/types.rs
dhruvmanila: Oops, yes this was from the first iteration of this PR. I'll remove this, thanks for catching it!

crates/ty_python_semantic/src/types.rs
dhruvmanila: Makes sense, thanks!

crates/ty_python_semantic/src/types/generics.rs
dhruvmanila: Consider the following example:

```py
from typing import Callable

def foo[**P](f1: Callable[P, int], f2: Callable[P, int]) -> Callable[P, bool]: ...

def f1(x: int) -> int:
    return 1

def f2(x: str) -> int:
    return ""

reveal_type(foo(f1, f2))
```

Here, `foo` is generic over a single `ParamSpec` which is used in multiple parameters. Later, when `P` is being specialized with two different parameter list, it will create a union which isn't then mapped in the return type. So, this allows us to pick the first specialization, type check the second occurrence using that and specialize the return type using that.

Without this early return, the reveal type would be `(**P@foo) -> bool`, right now it's `(x: int) -> bool`. This is mainly a workaround for now until ty supports specializing `P` here into a supertype considering all the parameter lists. I'll update the TODO and open a new follow-up issue to keep track.

crates/ty_python_semantic/src/types/infer/builder.rs
dhruvmanila: I don't think it has to use `infer_type_expression` but isn't `...` allowed in a few other places? I guess it's fine to special case this where required like here. I'll remove the TODO comment.

crates/ty_python_semantic/src/types/infer/builder.rs
dhruvmanila: I'm not sure if this provides any benefit as this would be a programming error faced by the user.

crates/ty_ide/src/hover.rs
dhruvmanila: Yup, my editor loves removing these whitespaces!

crates/ty_python_semantic/src/types/generics.rs
dhruvmanila: Yes, I meant the former, I'll update the documentation.

crates/ty_python_semantic/src/types/signatures.rs
dhruvmanila: Ah yes, I've extracted the common part as a nested function in this function body.

crates/ty_python_semantic/src/types/infer/builder.rs
MichaReiser: I have no opinion on this but a benefit of using `assert` is that we check the assertion in mypy primer.

crates/ty_ide/src/hover.rs
dhruvmanila: And the best PR of the year award goes to 🥁 https://github.com/astral-sh/ruff/pull/21806

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: I'm not sure how would that work. The `VariadicArgumentType` is specific to `*args` but this is handling the `**kwargs` (keyword variadic). Let me see if I can think of a way to de-duplicate this, currently it's used in `match_variadic_argument`, `match_keyword_variadic` and `check_keyword_variadic_argument_type` but they have slightly different semantics. I think the keyword variadic handling can possibly be clubbed together.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
AlexWaygood: Ahh, great point! I think I was getting mixed up with `TypeVarTuple`, where you of course _are_ allowed to annotate something inside the class as `self.x: tuple[*Ts]` if the class as a whole is generic over a `TypeVarTuple` `Ts`.

The mypy crash seems to have already been reported at https://github.com/python/mypy/issues/19839. We should definitely add tests (if you haven't already) for the (invalid) case where you attempt to annotate things inside the class with `self.x: P.args` and `self.y: P.kwargs`, so that we can make sure that we at least don't crash on that 😆

You might still be able to avoid the union-with-`Unknown` behaviour if you annotated `self.args` and `self.kwargs` as `Final`?

crates/ty_python_semantic/src/types/infer/builder.rs
AlexWaygood: yeah, I have a general preference for cheap assertions to be always enabled where possible. It usually makes it easier to tell from the stack trace in bug reports where the bug is. And it can be confusing if an assertion triggers locally (using a debug build) but not on a released (non-debug) version of ty.

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
AlexWaygood: Huh. I find that quite surprising. Could you add a comment saying that this doesn't seem to be explicitly permitted by the spec, but we match mypy's/pyright's behaviour here for compatibility? (And also the same comment for the equivalent test in the pep695 mdtest suite)

crates/ty_python_semantic/src/types/call/bind.rs
AlexWaygood: but I can still cunningly force ty to infer a type of `P.args` (rather than `P.args | Unknown`) if I annotate the instance variable with `Final` on your branch:

```py
from typing import Final
from collections.abc import Callable

class F[**P]:
    def __init__(self, fn: Callable[P, int], *args: P.args, **kwargs: P.kwargs):
        self.args: Final = args
        self.kwargs: Final = kwargs

def f[**P](a: F[P]):
    reveal_type(a.args)  # revealed: P@f.args
    reveal_type(a.kwargs)  # revealed: P@f.kwargs
```

crates/ty_python_semantic/src/types/call/bind.rs
dcreager: > I'm not sure how would that work. The `VariadicArgumentType` is specific to `*args` but this is handling the `**kwargs` (keyword variadic).

Ah I missed that distinction! Yeah you could theoretically have both `VariadicArgumentType` and `KeywordVariadicArgumentType`, which would possibly look very similar to each other.

In general I like avoiding repeating these kinds of tests since it adds cognitive load to keep them in sync. But I'm fine if you decide not to do this, or mark it as a TODO. It's not worth doing if it will be a heavy lift — or at least not in this PR.

crates/ty_python_semantic/resources/mdtest/generics/legacy/paramspec.md
dhruvmanila: Yeah, added a prose in both legacy and pep695 test case and in the actual source code as well.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dhruvmanila: > We should definitely add tests (if you haven't already) for the (invalid) case where you attempt to annotate things inside the class with `self.x: P.args` and `self.y: P.kwargs`, so that we can make sure that we at least don't crash on that 😆

For sure, I've added the test case, we don't error right now as I've planned it for follow-up.

> You might still be able to avoid the union-with-`Unknown` behaviour if you annotated `self.args` and `self.kwargs` as `Final`?

I think we should have both. The one without `Final` should be testing the special-case behavior that is implemented in `call/bind.rs` (`match_variadic`, `match_keyword_variadic` and `check_keyword_variadic_argument_type`).

AlexWaygood: That sounds great, thank you!!

crates/ty_python_semantic/src/types/infer/builder.rs
dhruvmanila: That all sounds reasonable, I've changed it to use `assert`

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: Added doc comments with explanation and examples.

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: I've added this as a test case, thanks!

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: Makes sense, I'll try to take a stab at it before landing this PR but otherwise will prefer to do it as a follow-up.

crates/ty_ide/src/hover.rs
dhruvmanila: I'll add a TODO comment, I'm not actually sure about this but it does make sense for this to be Contravariant, the current variance is being inferred automatically.

crates/ty_python_semantic/src/types/call/bind.rs
dhruvmanila: I've noted this down, will do it as a follow-up instead.

crates/ty_ide/src/hover.rs
AlexWaygood: Yeah, I think it should either be contravariant or invariant... bivariant just seems definitely wrong, anyway 😆

AlexWaygood: (A TODO comment sounds great 👍)

PR #21446, opened 2025-11-14
[ty] Add synthetic members to completions on dataclasses

sharkdp:
## Summary

Add synthetic members to completions on dataclasses and dataclass instances.

Also, while we're at it, add support for `__weakref__` and `__match_args__`.

closes https://github.com/astral-sh/ty/issues/1542

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -2375,7 +2375,9 @@ impl<'db> ClassLiteral<'db> {
 
                 Some(CallableType::function_like(db, signature))
             }
-            (CodeGeneratorKind::DataclassLike(_), "__match_args__") => {
+            (CodeGeneratorKind::DataclassLike(_), "__match_args__")
+                if Program::get(db).python_version(db) >= PythonVersion::PY310 =>
+            {
                 if !has_dataclass_param(DataclassFlags::MATCH_ARGS) {
                     return None;
                 }


crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -119,13 +135,9 @@ impl<'db> AllMembers<'db> {
             ),
 
             Type::NominalInstance(instance) => {
-                let class_literal = instance.class_literal(db);
+                let (class_literal, specialization) = instance.class(db).class_literal(db);
                 self.extend_with_instance_members(db, ty, class_literal);
-
-                // If this is a NamedTuple instance, include members from NamedTupleFallback
-                if CodeGeneratorKind::NamedTuple.matches(db, class_literal, None) {
-                    self.extend_with_type(db, KnownClass::NamedTupleFallback.to_class_literal(db));
```
sharkdp, line 127: Looks like this was wrong before, because we were extending an instance type with class members. Not a huge deal, but leads to wrong types (function literals instead of bound methods). Fixed here.

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -28,6 +29,21 @@ use rustc_hash::FxHashSet;
...
+const SYNTHETIC_DATACLASS_ATTRIBUTES: &[&str] = &[
+    "__lt__",
+    "__le__",
+    "__gt__",
+    "__ge__",
+    "__replace__",
+    "__setattr__",
+    "__delattr__",
+    "__slots__",
+    "__weakref__",
```
AlexWaygood, line 43: I think there's some subtlety here where this is only added to a _slotted_ dataclass if `weakref_slot=True` in the dataclass decorator? Not hugely important, though I guess you could add a failing test for it.

Is it worth adding `__match_args__` to this list now ("prematurely")? Then the all_members test for `__match_args__` will just automatically start passing when we add support for the synthesised attribute on the type-inference side

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -28,6 +29,21 @@ use rustc_hash::FxHashSet;
...
+    "__dataclass_fields__",
```
AlexWaygood, line 45: I have a vague memory that there's also a synthesised `__dataclass_params__` attribute...

AlexWaygood (approved):
Thank you!

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -414,6 +415,36 @@ impl<'db> AllMembers<'db> {
...
+    ) {
+        match CodeGeneratorKind::from_class(db, class_literal, specialization) {
+            Some(CodeGeneratorKind::NamedTuple) => {
+                if ty.is_nominal_instance() {
+                    self.extend_with_type(db, KnownClass::NamedTupleFallback.to_instance(db));
+                } else {
+                    self.extend_with_type(db, KnownClass::NamedTupleFallback.to_class_literal(db));
+                }
+            }
+            Some(CodeGeneratorKind::TypedDict) => {}
```
sharkdp, line 436: Typed dicts are somewhat special. I tried to incorporate them here, but it only results in way more code.

crates/ty_python_semantic/src/types/ide_support.rs
AlexWaygood: I guess we could add a test to completions.rs that shows that we now display the type correctly for the suggestions? Not essential though

crates/ty_python_semantic/src/types/ide_support.rs
sharkdp: I just added proper support for both `weakref_slot` and `match_args`. All TODOs are gone, now. Added new tests for both features in `dataclasses.md` as well.

crates/ty_python_semantic/src/types/ide_support.rs
sharkdp: I added that as a new member of type `Any`.

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2368,6 +2375,37 @@ impl<'db> ClassLiteral<'db> {
...
+            (CodeGeneratorKind::DataclassLike(_), "__weakref__") => {
+                if !has_dataclass_param(DataclassFlags::WEAKREF_SLOT)
+                    || !has_dataclass_param(DataclassFlags::SLOTS)
+                {
+                    return None;
+                }
+
+                // This could probably be `weakref | None`, but it does not seem important enough to
+                // model it precisely.
+                Some(Type::any())
```
AlexWaygood: This would provide a modicum of type safety?

```suggestion
                Some(UnionType::from_elements(db, [Type::any(), Type::none(db)]))
```

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -461,7 +461,51 @@ del frozen.x  # TODO this should emit an [invalid-assignment]
 
 ### `match_args`
```
AlexWaygood, line 462: This parameter doesn't exist on 3.9. But maybe not worth worrying about since 3.9 is end-of-life

sharkdp: Fixed. Decided not add a test, though.

crates/ty_python_semantic/src/types/ide_support.rs
sharkdp: Decided that this is not important enough.

PR #21447, opened 2025-11-14
[ty] Support stringified annotations in value-position `Annotated` instances

sharkdp:
## Summary

Infer the first argument `type` inside `Annotated[type, …]` as a type expression. This allows us to support stringified annotations inside `Annotated`.

## Ecosystem

* The removed diagnostic on `prefect` shows that we now understand the `State.data` type annotation in `src/prefect/client/schemas/objects.py:230`, which uses a stringified annotation in `Annoated`. The other diagnostics are downstream changes that result from this, it seems to be a commonly used data type.
* `artigraph` does something like `Annotated[cast(Any, field_info.annotation), *field_info.metadata]` which I'm not sure we need to allow? It's unfortunate since this is probably supported at runtime, but it seems reasonable that they need to add a `# type: ignore` for that.
* `pydantic` uses something like `Annotated[(self.annotation, *self.metadata)]` but adds a `# type: ignore`

## Test Plan

New Markdown test

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/annotated.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/annotated.md
@@ -83,6 +83,11 @@ class D(Annotated[list[str], "foo"]): ...
 
 # revealed: (<class 'D'>, <class 'list[str]'>, <class 'MutableSequence[str]'>, <class 'Sequence[str]'>, <class 'Reversible[str]'>, <class 'Collection[str]'>, <class 'Iterable[str]'>, <class 'Container[str]'>, typing.Protocol, typing.Generic, <class 'object'>)
 reveal_mro(D)
+
+class E(Annotated[list["E"], "metadata"]): ...
+
+# error: [revealed-type] "Revealed MRO: (<class 'E'>, <class 'list[E]'>, <class 'MutableSequence[E]'>, <class 'Sequence[E]'>, <class 'Reversible[E]'>, <class 'Collection[E]'>, <class 'Iterable[E]'>, <class 'Container[E]'>, typing.Protocol, typing.Generic, <class 'object'>)"
+reveal_mro(E)
 ```
 
 ### Not parameterized


crates/ty_python_semantic/resources/mdtest/annotations/annotated.md
```diff
@@ -76,7 +76,13 @@ from ty_extensions import reveal_mro
...
 class C(Annotated[int, "foo"]): ...
 
-reveal_mro(C)  # revealed: (<class 'C'>, <class 'int'>, <class 'object'>)
+# revealed: (<class 'C'>, <class 'int'>, <class 'object'>)
+reveal_mro(C)
+
+class D(Annotated[list[str], "foo"]): ...
+
+# revealed: (<class 'D'>, <class 'list[str]'>, <class 'MutableSequence[str]'>, <class 'Sequence[str]'>, <class 'Reversible[str]'>, <class 'Collection[str]'>, <class 'Iterable[str]'>, <class 'Container[str]'>, typing.Protocol, typing.Generic, <class 'object'>)
+reveal_mro(D)
```
AlexWaygood, line 85: Is it worth also adding a test for a variant of this:

```py
class E(Annotated[list["E"], "metadata"]): ...
```

(which works fine at runtime, I think)

crates/ty_python_semantic/src/types/class_base.rs
```diff
@@ -183,7 +183,16 @@ impl<'db> ClassBase<'db> {
...
+                KnownInstanceType::Annotated(ty) => {
+                    // Unions are not supported in this position, so we only need to support
+                    // something like `class C(Annotated[Base, "metadata"]): ...`, which we
+                    // can do by turning the instance type (`Base` in this example) back into
+                    // a class.
+                    let annotated_ty = ty.inner(db);
+                    let instance_ty = annotated_ty.as_nominal_instance()?;
+
+                    Some(Self::Class(instance_ty.class(db)))
+                }
```
AlexWaygood, line 195: hmmmm I feel like there's probably going to be some edge case someone will come up with at some point where this lossy conversion will come back to bite us... but I'm not sure what they are, and this seems fine for now 😆

PR #21450, opened 2025-11-14
[ty] Make `__getattr__` available for `ModuleType` instances

sharkdp:
## Summary

Typeshed has a (fake) `__getattr__` method on `types.ModuleType` with a return type of `Any`. We ignore this method when accessing attributes on module *literals*, but with this PR, we respect this method when dealing with `ModuleType` itself. That is, we allow arbitrary attribute accesses on instances of `types.ModuleType`. This is useful because dynamic import mechanisms such as `importlib.import_module` use `ModuleType` as a return type.

closes https://github.com/astral-sh/ty/issues/1346

## Ecosystem

Massive reduction in diagnostics. The few new diagnostics are true positives. 

## Test Plan

Added regression test.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -336,6 +336,9 @@ bitflags! {
 
         /// Skip looking up attributes on the builtin `int` and `str` classes.
         const MRO_NO_INT_OR_STR_LOOKUP = 1 << 3;
+
+        /// Do not call `__getattr__` during member lookup.
+        const NO_GETATTR_LOOKUP = 1 << 4;
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -4512,18 +4520,7 @@ impl<'db> Type<'db> {
                 );
 
                 let custom_getattr_result = || {
-                    // Typeshed has a fake `__getattr__` on `types.ModuleType` to help out with
-                    // dynamic imports. We explicitly hide it here to prevent arbitrary attributes
-                    // from being available on modules. Same for `types.GenericAlias` - its
```
sharkdp, line 4517: I don't understand why the `GenericAlias` exception was here. No tests seem to fail when I remove it?

AlexWaygood:
Thank you!

PR #21451, opened 2025-11-14
[ty] Add panic-by-default await methods to `TestServer`

MichaReiser:
## Summary

This should help with https://github.com/astral-sh/ty/issues/1551. It doesn't 
fix the root cause but I think it's an overall improvement to the testing experience.

The `await_response`, `await_notification` and `await_request` all return a `Result` today
to account for that they can fail. However, almost all writen E2E tests only care about the
case where a request is successful. 

This PR changes the `await_` methods to panic by default when they fail and adds
`try_` counterparts that return a `Result` instead for the few cases where we 
explicitly test the error case. 

I find that this overall gives better ercognomis as we get a panic 
on the line where the request failed rather than a: Test failed because request timed out
and you then have to guess which request it was.

## Test Plan

`cargo test`

### DIFF
--- a/crates/ty_server/tests/e2e/main.rs
+++ b/crates/ty_server/tests/e2e/main.rs
@@ -34,7 +34,6 @@ mod notebook;
 mod publish_diagnostics;
 mod pull_diagnostics;
 
-use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, VecDeque};
 use std::num::NonZeroUsize;
 use std::sync::{Arc, OnceLock};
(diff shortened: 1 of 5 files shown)


crates/ty_server/tests/e2e/main.rs
```diff
@@ -372,8 +376,56 @@ impl TestServer {
...
+    /// or the server responded twice. Use [`Self::try_await_response`] if you want a non-panicking version.
+    #[track_caller]
+    pub(crate) fn await_response<R>(&mut self, id: &RequestId) -> R::Result
+    where
+        R: Request,
+    {
+        match self.try_await_response::<R>(id, None) {
+            Ok(result) => result,
+            Err(err) => match err {
+                TestServerError::ResponseError(response_error) => {
```
dhruvmanila: Should we use the `thiserror::Error` messages here to display the error here? We could possibly add the request ID as context.

crates/ty_server/tests/e2e/main.rs
```diff
@@ -417,7 +468,54 @@ impl TestServer {
...
+    ///
+    /// If the server doesn't send the notification within the default timeout or
+    /// the notification failed to deserialize. Use [`Self::try_await_notification`] for
+    /// a panic-free alternative.
+    #[track_caller]
+    pub(crate) fn await_notification<N: Notification>(&mut self) -> N::Params {
+        match self.try_await_notification::<N>(None) {
+            Ok(params) => params,
+            Err(err) => match err {
+                TestServerError::RecvTimeoutError(RecvTimeoutError::Disconnected) => {
```
dhruvmanila: Same here although this is a bit different but the response error variant could be moved first to use `unreachable` and then a catch-all `err` for the remaining error.

crates/ty_server/tests/e2e/main.rs
```diff
@@ -372,8 +376,56 @@ impl TestServer {
     /// This method will remove the response from the internal data structure, so it can only be
     /// called once per request ID.
     ///
-    /// [`send_request`]: TestServer::send_request
```
dhruvmanila, line 375: I think this is still required to reference the `send_request` function in the docs.

crates/ty_server/tests/e2e/main.rs
```diff
@@ -372,8 +376,56 @@ impl TestServer {
...
+                TestServerError::DeserializationError(error) => {
+                    panic!("Failed to deserialize response {id}: {error}")
+                }
+            },
+        }
+    }
+
+    /// Wait for a server response corresponding to the given request ID.
+    ///
+    /// This should only be called if a request was already sent to the server via [`send_request`]
```
dhruvmanila, line 422: The `send_request` requires a reference here.

crates/ty_server/tests/e2e/main.rs
```diff
@@ -462,7 +561,46 @@ impl TestServer {
...
+    ///
+    /// # Panics
+    ///
+    /// If receiving the request fails.
+    #[track_caller]
+    pub(crate) fn await_request<R: Request>(&mut self) -> (RequestId, R::Params) {
+        match self.try_await_request::<R>(None) {
+            Ok(request) => request,
+            Err(err) => match err {
+                TestServerError::RecvTimeoutError(RecvTimeoutError::Disconnected) => {
```
dhruvmanila: Same here as the `await_response` comment.

dhruvmanila (approved):
This sounds reasonable to me, I only have a few minor comments which doesn't necessarily need to be addressed.

dhruvmanila:
I believe the return type for all of the test cases can now be removed (it is `Result<()>` currently) and the dangling `Ok(())` after every test case as the panic is now part of the test server.

MichaReiser:
> I believe the return type for all of the test cases can now be removed (it is Result<()> currently) and the dangling Ok(()) after every test case as the panic is now part of the test server.

Unfortunately not. `TestServerBuilder::new` can still fail for some reason (but that's fine because the drop handler will never execute in that case)

crates/ty_server/tests/e2e/main.rs
MichaReiser: Good idea

PR #21453, opened 2025-11-14
[ty] Fixup a few details around version-specific dataclass features

AlexWaygood:
## Summary

- `slots=True` and `match_args=True` were both added in Python 3.10
- `weakref_slot=True` was added in Python 3.11

I also added a test that shows the types we display when providing autocompletions for instances of `NamedTuple` classes; that came up in https://github.com/astral-sh/ruff/pull/21446


sharkdp:
Thanks

PR #21456, opened 2025-11-14
[ty] Further improve details around which expressions should be deferred in stub files

AlexWaygood:
## Summary

- Always restore the previous `deferred_state` after parsing a type expression: we don't want that state leaking out into other contexts where we shouldn't be deferring expression inference
- Always defer the right-hand-side of a PEP-613 type alias in a stub file, allowing for forward references on the right-hand side of `T: TypeAlias = X | Y` in a stub file

Addresses @carljm's review in https://github.com/astral-sh/ruff/pull/21401#discussion_r2524260153

## Test Plan

I added a regression test for a regression that the first version of this PR introduced (we need to make sure the r.h.s. of a PEP-613 `TypeAlias`es is always deferred in a stub file)

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
@@ -1,6 +1,8 @@
 # PEP 613 type aliases
 
-We do not support PEP 613 type aliases yet. For now, just make sure that we don't panic:
+## No panics
+
+We do not fully support PEP 613 type aliases yet. For now, just make sure that we don't panic:
 
 ```py
 from typing import TypeAlias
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Lots of new sympy diagnostics, but I think that's just because we now understand some type aliases that were previously inferred as `Unknown` due to forward references...

AlexWaygood:
> ```diff
> + mypy/typeshed/stdlib/builtins.pyi:1459:54: error[too-many-positional-arguments] Too many positional arguments to class `tuple`: expected 1, got 2
> ```

not sure what's going on here but it looks like a pre-existing issue. I get lots of these diagnostics if I run `cargo run --manifest-path ../ruff/Cargo.toml -p ty check mypy/typeshed/stdlib/builtins.pyi` from mypy mypy clone with our `main` branch checked out in the `../ruff` directory

AlexWaygood:
> Lots of new sympy diagnostics, but I think that's just because we now understand some type aliases that were previously inferred as `Unknown` due to forward references...

Yes, it's because we now understand [this typeshed type alias](https://github.com/python/typeshed/blob/0c0bad88312ca103b794e52c5f2f20466a743b59/stdlib/xml/dom/minidom.pyi#L23), which is used as the upper bound for [this typeshed TypeVar](https://github.com/python/typeshed/blob/0c0bad88312ca103b794e52c5f2f20466a743b59/stdlib/xml/dom/minidom.pyi#L28), which is used in [this typeshed method annotation](https://github.com/python/typeshed/blob/0c0bad88312ca103b794e52c5f2f20466a743b59/stdlib/xml/dom/minidom.pyi#L637)

AlexWaygood:
> Might have been simpler to defer the PEP 613 part of this to the PEP 613 PR, but the conflict should be minor. Thank you!

Yeah — I initially did it without that bit but it led to a few ecosystem regressions without the r.h.s. of PEP-613 aliases being deferred!

PR #21457, opened 2025-11-14
[ty] Support class-arguments for dataclass transformers

sharkdp:
## Summary

Allow metaclass-based and baseclass-based dataclass-transformers to overwrite the default behavior using class arguments:

```py
class Person(Model, order=True):
    # ...
```

## Conformance tests

Four new tests passing!

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
@@ -356,13 +356,17 @@ model < model  # No error
 
 ### Overwriting of default parameters on the dataclass-like class
 
+In the following examples, we show how a model can overwrite the default parameters set by the
+`dataclass_transform` decorator. In particular, we change from `frozen=True` to `frozen=False`, and
+from `order=False` (default) to `order=True`:
+
 #### Using function-based transformers
 
 ```py
 from typing import dataclass_transform
 
 @dataclass_transform(frozen_default=True)
-def default_frozen_model(*, frozen: bool = True): ...
+def default_frozen_model(*, frozen: bool = True, order: bool = False): ...
 @default_frozen_model()
 class Frozen:
     name: str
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -402,12 +411,17 @@ class Frozen(DefaultFrozenModel):
...
-class Mutable(DefaultFrozenModel, frozen=False):
+Frozen(name="A") < Frozen(name="B")  # error: [unsupported-operator]
+
+class Mutable(DefaultFrozenModel, frozen=False, order=True):
     name: str
 
 m = Mutable(name="test")
-# TODO: no error here
+# TODO: This should not be an error. In order to support this, we need to implement the precise `frozen` semantics of
+# `dataclass_transform` described here: https://typing.python.org/en/latest/spec/dataclasses.html#dataclass-semantics
```
sharkdp, line 421: The reason why this doesn't work is a bit tricky. For frozen models, we generate a `__setattr__` method that returns `Never`. This allows us to emit the `invalid-assignment` diagnostic. We set `frozen=False` on `Mutable`, and with this PR, this means that we do not generate a `__setattr__` method on `Mutable` *directly* anymore. But the implicit `__setattr__` call in the assignment to `m.name` below calls also looks up `__setattr__` on the meta type of `Mutable`, i.e. `DefaultFrozenMeta`. And since `DefaultFrozenMeta` has `frozen_default=True`, we *do* generate a `__setattr__` method on the meta class. This method is what causes the right behavior for the `Frozen` model. But it turns out this is not what actually happens at runtime. https://typing.python.org/en/latest/spec/dataclasses.html#dataclass-semantics describes the precise semantics for frozen dataclasses transformers. `DefaultFrozenMeta` is actually neither frozen nor non-frozen. The `__setattr__` method should instead only be generated on `Frozen` directly.

sharkdp:
> ```diff
> - src/scikit_build_core/build/wheel.py:98:20: error[no-matching-overload] No overload of bound method `__init__` matches arguments
> ```

Is this another flaky ecosystem diff? :confused: I can not reproduce this locally, neither can ecosystem-analyzer.

MichaReiser:
> > ```diff
> 
> > - src/scikit_build_core/build/wheel.py:98:20: error[no-matching-overload] No overload of bound method `__init__` matches arguments
> 
> > ```
> 
> 
> 
> Is this another flaky ecosystem diff? :confused: I can not reproduce this locally, neither can ecosystem-analyzer.

Yes it is. It shows up on almost every PR

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
carljm: Should we file an issue for this?

sharkdp: Added to https://github.com/astral-sh/ty/issues/1327

PR #21459, opened 2025-11-14
Limit `eglot-format` hook to eglot-managed Python buffers

thamer:
Running `eglot-format` in buffers not managed by Eglot causes a `jsonrpc-error` in Emacs 30. It may also display a `documentFormattingProvider` warning when the server does not support formatting. Add checks for both.

### DIFF
--- a/docs/editors/setup.md
+++ b/docs/editors/setup.md
@@ -400,15 +400,13 @@ Ruff can be utilized as a language server via [`Eglot`](https://github.com/joaot
 To enable Ruff with automatic formatting on save, use the following configuration:
 
 ```elisp
-(add-hook 'python-mode-hook 'eglot-ensure)
 (with-eval-after-load 'eglot
   (add-to-list 'eglot-server-programs
-               '(python-mode . ("ruff" "server")))
-  (add-hook 'after-save-hook
-            (lambda ()
-              (when (and (derived-mode-p 'python-base-mode)
-                         (bound-and-true-p eglot--managed-mode))
-                (eglot-format-buffer)))))
+               '(python-base-mode . ("ruff" "server"))))
+(add-hook 'python-base-mode-hook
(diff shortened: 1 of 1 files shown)


docs/editors/setup.md
```diff
@@ -404,7 +404,11 @@ To enable Ruff with automatic formatting on save, use the following configuratio
 (with-eval-after-load 'eglot
   (add-to-list 'eglot-server-programs
                '(python-mode . ("ruff" "server")))
-  (add-hook 'after-save-hook 'eglot-format))
+  (add-hook 'after-save-hook
+            (lambda ()
+              (when (and (derived-mode-p 'python-base-mode)
+                         (bound-and-true-p eglot--managed-mode))
+                (eglot-format-buffer)))))
```
ntBre: I think another option could be to add the hook with the `LOCAL` argument within a python-mode-hook. Something like:

```suggestion
  (add-hook 'python-mode-hook
            (lambda ()
              (add-hook 'after-save-hook 'eglot-format nil t)))
```

but I haven't tested this. I'm still on lsp-mode rather than eglot and don't use format-on-save.

This could also be combined with the `python-mode-hook` above:

```elisp
  (add-hook 'python-mode-hook
            (lambda ()
              (eglot-ensure)
              (add-hook 'after-save-hook 'eglot-format nil t)))
```

thamer: That's better. But it should be `python-base-mode-hook` to also work on the tree-sitter mode.

ntBre: Ah nice, we should fix that on line 403 too then. And maybe 406?

thamer: Done.

dhruvmanila:
I'd prefer if @ntBre could look at this as I'm not familiar with Emacs plugins :)

PR #21460, opened 2025-11-14
[ty] Suppress completions when introducing names with `as`

BurntSushi:
There are a few places in Python where it is known that new names are
being introduced and thus we probably shouldn't offer completions. We
already handle this today for things like `class <CURSOR>` and `def
<CURSOR>`. But we didn't handle `as <CURSOR>`, which can appear in
`import`, `with`, `except` and `match` statements. Indeed, these are
exactly the 4 cases where the `as` keyword can occur. So we look for the
presence of `as` and suppress completions based on that.

While we're here, we also make the implementation a bit more robust with
respect to suppressing completions when the user hasn't typed anything.
Namely, previously, we'd still offer completions in a `class <CURSOR>`
context. But it looks like LSP clients (at least, VS Code) doesn't ask
for completions here, so we were "saved" incidentally. This PR detects
this case and suppresses completions there so we don't rely on LSP
client behavior to handle that case correctly.

Fixes astral-sh/ty#1287

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -244,7 +244,7 @@ pub fn completion<'db>(
     let tokens = tokens_start_before(parsed.tokens(), offset);
     let typed = find_typed_text(db, file, &parsed, offset);
 
-    if is_in_no_completions_place(db, tokens, file) {
+    if is_in_no_completions_place(db, file, tokens, typed.as_deref()) {
         return vec![];
     }
     if let Some(completions) = only_keyword_completion(tokens, typed.as_deref()) {
@@ -930,8 +930,9 @@ fn find_typed_text(
     let last = tokens.last()?;
     // It's odd to include `TokenKind::Import` here, but it
     // indicates that the user has typed `import`. This is
-    // useful to know in some contexts.
-    if !matches!(last.kind(), TokenKind::Name | TokenKind::Import) {
+    // useful to know in some contexts. And this applies also
+    // to the other keywords.
+    if !matches!(last.kind(), TokenKind::Name) && !last.kind().is_keyword() {
         return None;
     }
     // This one's weird, but if the cursor is beyond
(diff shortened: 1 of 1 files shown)


BurntSushi:
Demo:


https://github.com/user-attachments/assets/b595bf9d-1bc7-4a87-b4ac-a45d694cc36a

crates/ty_ide/src/completion.rs
```diff
@@ -930,8 +930,9 @@ fn find_typed_text(
     let last = tokens.last()?;
     // It's odd to include `TokenKind::Import` here, but it
     // indicates that the user has typed `import`. This is
-    // useful to know in some contexts.
-    if !matches!(last.kind(), TokenKind::Name | TokenKind::Import) {
+    // useful to know in some contexts. And this applies also
+    // to the other keywords.
+    if !matches!(last.kind(), TokenKind::Name) && !last.kind().is_keyword() {
```
AlexWaygood, line 935: (optional, obviously)

```suggestion
    if last.kind() != TokenKind::Name && !last.kind().is_keyword() {
```

crates/ty_ide/src/completion.rs
```diff
@@ -3158,8 +3168,13 @@ import foo as ba<CURSOR>
         // which is kind of annoying. So just assert that it
         // runs without panicking and produces some non-empty
         // output.
+        //
+        // ... some time passes ...
+        //
+        // Actually, this shouldn't offer any completions since
+        // the context here is introducing a new name.
```
AlexWaygood, line 3175: ❤️

MichaReiser (approved):

AlexWaygood:
> Fixes [astral-sh/ty#1287](https://github.com/astral-sh/ty/issues/1287)

The first three bullets in https://github.com/astral-sh/ty/issues/1287#issuecomment-3527860427 will not be fixed by this PR -- but we can open a followup issue(s) for those

crates/ty_ide/src/completion.rs
BurntSushi: I very weakly prefer `matches!` here. :)

BurntSushi:
> The ony issue I spotted from manual testing is that in this location, `as` does not seem to be suggested (and in fact, `as` should probably be the _only_ suggestion in this context?). But that's probably a distinct issue to the ones being fixed in this PR.

Yeah good catch. This is indeed a distinct issue. The `import` detection allows false positives, but not for any specifically compelling reason. I'll open a new issue for this.

BurntSushi:
I've filed https://github.com/astral-sh/ty/issues/1563 and https://github.com/astral-sh/ty/issues/1562 as additional work here.

PR #21461, opened 2025-11-14
[ty] Improve diagnostic range for `non-subscriptable` diagnostics

AlexWaygood:
## Summary

Currently our diagnostic only covers the range of the thing being subscripted:

image

But it should probably cover the _whole_ subscript expression (arguably the more "incorrect" bit is the `["foo"]` part of this expression, not the `x` part of this expression!)

## Test Plan

Added a snapshot

Co-authored-by: Brent Westbrook <36778786+ntBre@users.noreply.github.com>

### DIFF
--- a/crates/ty/tests/cli/exit_code.rs
+++ b/crates/ty/tests/cli/exit_code.rs
@@ -153,7 +153,7 @@ fn both_warnings_and_errors() -> anyhow::Result<()> {
         "#,
     )?;
 
-    assert_cmd_snapshot!(case.command().arg("--warn").arg("unresolved-reference"), @r###"
+    assert_cmd_snapshot!(case.command().arg("--warn").arg("unresolved-reference"), @r"
     success: false
     exit_code: 1
     ----- stdout -----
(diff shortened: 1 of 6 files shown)


carljm:
I think highlighting the whole expression (as you do here) is best. It is both the type of the base object (which is the type that is not subscriptable), and the fact that you are subscripting it, that's "at fault" for the error. (In the case of not-subscriptable, the actual key is arguably not involved at all! But it still makes sense to highlight the entire expression IMO.)

carljm (approved):

MichaReiser:
Similar to attribute expressions, I think it's undesired to highlight the entire object because it can get very noisy if that expression is long (think a call expression)

I think the ideal is to highlight .attribute or [value].

Related discussion https://github.com/astral-sh/ty/issues/169

I don't think this should block this PR because it at least makes it consistent but we should update said issue to also mention subscript

AlexWaygood:
Heh, well, I disagree for the same reasons I gave in https://github.com/astral-sh/ty/issues/169 ;) but I'll add a note in that issue to mention that the same discussion also applies to subscript diagnostics!

PR #21469, opened 2025-11-15
[`flake8-bandit`] Handle string literal bindings in suspicious-url-open-usage (`S310`)

### LINKED ISSUE #21462: suspicious-url-open-usage (S310) should handle string literal
### Summary

```
import urllib.request
path = "https://example.com/data.csv"
urllib.request.urlretrieve(path, "data.csv")
```
This currently raises suspicious-url-open-usage (S310), but it should not because path can be inferred to have Literal['https://example.com/data.csv'] type and that's safe.

danparizher:
## Summary

Fixes false positive in `S310` when URL arguments are assigned to variables that resolve to string literals with safe schemes (`http://` or `https://`).

Fixes #21462.

## Problem

The rule flagged `urllib.request.urlretrieve(path, ...)` when `path` was assigned a string literal like `"https://example.com/data.csv"` because it only checked direct string literals, f-strings, and concatenation, not variable bindings.

## Approach

Added `resolve_to_string_literal()` helper that uses `find_binding_value()` to resolve variable bindings to their string literal values, following the pattern in `unnecessary_regular_expression.rs`. Updated the `urlopen`/`urlretrieve` handler to check resolved string literals before falling back to `leading_chars()`.

## Test Plan

Added test case to `S310.py` for the variable binding scenario.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_bandit/snapshots/ruff_linter__rules__flake8_bandit__tests__preview__S310_S310.py.snap
+++ b/crates/ruff_linter/src/rules/flake8_bandit/snapshots/ruff_linter__rules__flake8_bandit__tests__preview__S310_S310.py.snap
@@ -6,8 +6,99 @@ source: crates/ruff_linter/src/rules/flake8_bandit/mod.rs
 +linter.preview = enabled
 
 --- Summary ---
-Removed: 0
+Removed: 8
 Added: 2
+
+--- Removed ---
+S310 Audit URL open for permitted schemes. Allowing use of `file:` or custom schemes is often unexpected.
+  --> S310.py:51:1
+   |
+49 | # https://github.com/astral-sh/ruff/issues/21462
+50 | path = "https://example.com/data.csv"
+51 | urllib.request.urlretrieve(path, "data.csv")
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+52 | url = "https://example.com/api"
+53 | urllib.request.Request(url)
+   |
+
+
+S310 Audit URL open for permitted schemes. Allowing use of `file:` or custom schemes is often unexpected.
+  --> S310.py:53:1
+   |
+51 | urllib.request.urlretrieve(path, "data.csv")
+52 | url = "https://example.com/api"
+53 | urllib.request.Request(url)
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+54 |
+55 | # Test resolved f-strings and concatenated string literals
+   |
+
+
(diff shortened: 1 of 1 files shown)


crates/ruff_linter/src/rules/flake8_bandit/rules/suspicious_function_call.rs
edgarrmondragon: It might also be worth similarly resolving the `url` argument of `urllib.request.Request`

amyreese (approved):

amyreese:
The one ecosystem change is from a project where this previously triggered no longer getting a false positive, and therefore getting an "unused suppression" from their noqa.

This might be worth gating behind preview flag.

crates/ruff_linter/src/rules/flake8_bandit/rules/suspicious_function_call.rs
```diff
@@ -1196,8 +1223,17 @@ fn suspicious_function(
                             }
                         }
 
-                        // If the `url` argument is a string literal, allow `http` and `https` schemes.
+                        // If the `url` argument is a string literal (including resolved bindings), allow `http` and `https` schemes.
                         Some(expr) => {
+                            // First try to resolve bindings to string literals
+                            if let Some(str_lit) =
+                                resolve_to_string_literal(expr, checker.semantic())
```
ntBre: This version of `resolve_to_string_literal` actually seems too restrictive in this case since `leading_chars` just takes any `Expr`. I guess we should just resolve an `Expr::Name` to whatever its binding is and pass that directly to `leading_chars`. Something like:

```rust
                        Some(expr) => {
                            let expr = resolve_name(expr);

                            if leading_chars(expr).is_some_and(has_http_prefix) {
                                return;
                            }
                        }
```

Then we could add tests showing that we handle resolved f-strings and concatenated string literals.

edgarrmondragon:
@danparizher I opened https://github.com/danparizher/ruff/pull/1 against your fork to handle calls to `urllib.request.Request` outside of `urllib.request.urlopen`, and also process concatenated string literals recursively.

ntBre:
Oh, I was waiting to see if the PR against the fork was merged, but I think the URL resolution was also fixed in the last commit. Is that right?

edgarrmondragon:
> Oh, I was waiting to see if the PR against the fork was merged, but I think the URL resolution was also fixed in the last commit. Is that right?

IIUC, the last commit in this PR fixed the `urllib.request.urlopen(urllib.request.Request(url))` case, but not `urllib.request.Request(url)` on its own.

It also didn't fully address this comment (https://github.com/astral-sh/ruff/pull/21469#discussion_r2539560233)

> handle resolved f-strings and concatenated string literals.

But my all means, feel free to merge without my updates, which I can open a new PR for.

Issue #20736, opened 2025-10-07
[feature request] option to exclude type checking imports in `ruff analyze`


gwdekker:
tach by default ignores dependencies if the import statement is wrapped by if TYPE_CHECKING:. Is that something that is possible with ruff analyze graph as well, and would you consider adding it if not yet possible? (from another thread on the issue tracker I found [Grimp](https://grimp.readthedocs.io/en/stable/usage.html), which supports this with graph = grimp.build_graph('mypackage', exclude_type_checking_imports=True).

came up in https://github.com/astral-sh/ruff/issues/20701

@seddonym

MichaReiser:
An option like this does make sense to me (which we might want to default to false). 

An alternative is that ruff always returns all imports but sets a typing_only property for imports that are in TYPE_CHECKING blocks.

jhirshman:
+1 here.  we recently changed from tach to using just ruff and are seeing the same thing

seddonym:
If you're looking for a Tach replacement right now, Import Linter supports this via the exclude-type-checking-imports setting. https://import-linter.readthedocs.io/en/stable/usage.html#top-level-configuration

gauthsvenkat:
Happy to pick this up, though I need some advice on what would be the best way to do this.

- If we do it right, we need to build a semantic model in `ruff analyze` and create a dependency on `crates/ruff_python_semantic` which has a ` is_type_checking_block` [helper function](https://github.com/astral-sh/ruffruff/blob/main/crates/ruff_python_semantic/src/analyze/typing.rs#L419-L419).  
We should be able to plug it in `collector.rs`'s `visit_stmt` [method](https://github.com/astral-sh/ruff/blob/main/crates/ruff_graph/src/collector.rs#L97) to track type checking blocks.

- The `CollectedImport` [enum](https://github.com/astral-sh/ruff/blob/main/crates/ruff_graph/src/collector.rs#L156) can be made in to a struct and have a `typing_only` boolean field for this purpose.

@MichaReiser do we go this route? Let me know if I'm overlooking something.

MichaReiser:
[...]

PR #21472, opened 2025-11-15
`analyze`: Add option to skip over imports in `TYPE_CHECKING` blocks

gauthsvenkat:
## Summary

`ruff analyze graph` naively detects all imports. This MR adds a feature where `TYPE_CHECKING` imports can be detected and optionally excluded (using a CLI flag). 

Closes: #20736 

### Main changes

- [collector.rs] `CollectedImport` is now a `struct` instead of an `enum`. This was done so that a new `type_checking` field could be added.
- [collector.rs] Refactored the logic in `Stmt::Import{,From}` into helper functions `collect_import{,_from}`. Done so they can be reused in `Stmt::If` to detect type checking imports.
- [collector.rs] Added a new helper function `is_type_checking_condition`. This can detect very simple type-checking conditions, but, practically, it would cover a lot of the cases.

## Test Plan

- Added a new test to check that type-checking imports are detected and filtered, iff `--exclude-type-checking-imports` flag is set.


gauthsvenkat:
Pipeline fails cause Clippy is complaining there are too many bools in `AnalyzeGraphCommand`. But it is a CLI option, so not sure how to deal with it.

crates/ruff_graph/src/collector.rs
```diff
@@ -30,71 +30,99 @@ impl<'a> Collector<'a> {
...
+    fn visit_stmt(&mut self, stmt: &'ast Stmt) {
+        match stmt {
+            Stmt::ImportFrom(stmt) => {
+                self.collect_import_from(stmt, false);
+            }
+            Stmt::Import(stmt) => {
+                self.collect_import(stmt, false);
+            }
+            Stmt::If(ast::StmtIf { test, body, .. }) => {
+                if is_type_checking_condition(test) {
```
MichaReiser: I think a simpler approach is to pass the `exclude_type_checking` option to `Collector` and simply skip the entire body if `test` is a `TYPE_CHECKING` condition. 

This removes the need to split `collect_import_from` and `collect_import` out of `visit_stmt` or that we need to set `type_checking` on `CollectedImport`

crates/ruff/src/args.rs
```diff
@@ -193,6 +193,9 @@ pub struct AnalyzeGraphCommand {
     /// Path to a virtual environment to use for resolving additional dependencies
     #[arg(long)]
     python: Option<PathBuf>,
+    /// Exclude imports that are only used for type checking (i.e., imports within `if TYPE_CHECKING:` blocks)
+    #[arg(long)]
+    exclude_type_checking_imports: bool,
```
MichaReiser: I think I'd name this `type_checking_only_imports` or just `type_checking_imports` where `true` includes them and `false` skips them (should default to `true`).


We should also add this as a configuration optiont to https://github.com/astral-sh/ruff/blob/6e4b15b6ca72285a2f509076f3871adca90e923a/crates/ruff_workspace/src/options.rs#L3816

crates/ruff/src/args.rs
```diff
@@ -193,6 +193,9 @@ pub struct AnalyzeGraphCommand {
     /// Path to a virtual environment to use for resolving additional dependencies
     #[arg(long)]
     python: Option<PathBuf>,
+    /// Exclude imports that are only used for type checking (i.e., imports within `if TYPE_CHECKING:` blocks)
```
MichaReiser: You can add `#[expect(clippy::struct_excessive_bools)]` to the struct definition to silence the excessive bool lint

crates/ruff_graph/src/collector.rs
gauthsvenkat: Oh yeah, that would be simpler. 

Then we don't track which imports are type-checking only for other uses. e.g., if we want to show only type-checking imports? 
Is that okay?

MichaReiser: I think that's okay for now, given that we don't show that information anywhere.

crates/ruff/src/args.rs
gauthsvenkat: I've done both. 

Adding it as a config option ended up being a bigger change than I expected. I tried to follow how it was done for `detect-string-imports`. Please let me know if that is good.

gauthsvenkat: I also just noticed that `preview` and `string_imports` config options are not boolean, but instead "enabled" or "disabled". Should I model it that way instead?

MichaReiser: Yeah, it requires a lot of boilerplate code. 

> t preview and string_imports config options are not boolean, but instead "enabled" or "disabled". Should I model it that way instead?

I think a boolean is fine.

crates/ruff_graph/src/collector.rs
```diff
@@ -91,10 +98,25 @@ impl<'ast> SourceOrderVisitor<'ast> for Collector<'_> {
...
+            Stmt::If(ast::StmtIf {
+                test,
+                body,
+                elif_else_clauses,
+                range: _,
+                node_index: _,
+            }) => {
+                // Skip TYPE_CHECKING blocks if not requested
+                if self.type_checking_imports || !is_type_checking_condition(test) {
+                    self.visit_body(body);
```
MichaReiser, line 110: We could consider handling 

```py
if !TYPE_CHECKING:
	...
else: 
	...
```

and 

```py
if condition:
	...
elif TYPE_CHECKING:
	...
```

but we can do this as a separate PR

PR #21474, opened 2025-11-15
[ty] Dataclass transform: complete set of parameters

sharkdp:
## Summary

We previously only allowed models to overwrite the `{eq,order,kw_only,frozen}_defaults` of the dataclass-transformer, but all other standard-dataclass parameters should be equally supported with the same behavior.

## Test Plan

Added regression tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
+++ b/crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
@@ -455,6 +455,82 @@ m.name = "new"  # No error
 reveal_type(Mutable(name="A") < Mutable(name="B"))  # revealed: bool
 ```
 
+## Other `dataclass` parameters
+
+Other parameters from normal dataclasses can also be set on models created using
+`dataclass_transform`.
+
+### Using function-based transformers
+
+```py
+from typing_extensions import dataclass_transform, TypeVar, Callable
+
+T = TypeVar("T", bound=type)
+
+@dataclass_transform()
+def fancy_model(*, slots: bool = False) -> Callable[[T], T]:
+    raise NotImplementedError
+
+@fancy_model()
+class NoSlots:
+    name: str
+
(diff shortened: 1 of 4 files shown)


crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
```diff
@@ -455,6 +455,82 @@ m.name = "new"  # No error
...
+
+### Using function-based transformers
+
+```py
+from typing_extensions import dataclass_transform, TypeVar, Callable
+
+T = TypeVar("T", bound=type)
+
+@dataclass_transform()
+def fancy_model(*, slots: bool = False) -> Callable[[T], T]:
```
AlexWaygood, line 471: isn't `slots=False` the default default? Wouldn't a more interesting test be this?

```suggestion
def fancy_model(*, slots: bool = True) -> Callable[[T], T]:
```

(The same applies to your other examples below!)

crates/ty_python_semantic/src/types.rs
```diff
@@ -622,6 +622,19 @@ bitflags! {
     }
 }
 
+pub const DATACLASS_FLAGS: &[(&str, DataclassFlags)] = &[
```
```suggestion
pub(crate) const DATACLASS_FLAGS: &[(&str, DataclassFlags)] = &[
```

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclass_transform.md
sharkdp: The default parameter here has no actual effect on anything. `slots=False` is the default behavior and there is no way to override this for a dataclass-transformer. There is no `slots_default` parameter for the `dataclass_transform` call.

For the parameters that can actually change their defaults (like `kw_only`), we already have tests elsewhere in this file. Here, we just test that we can also modify something like `slots` on an actual *model*/dataclass (not on the transformer / the template).

PR #21475, opened 2025-11-15
[ty] Improve diagnostics for invalid exceptions

AlexWaygood:
## Summary

Not a high-priority task... but it _is_ a weekend :P

This PR improves our diagnostics for invalid exceptions. Specifically:
- We now give a special-cased ``help: Did you mean `NotImplementedError`` subdiagnostic for `except NotImplemented`, `raise NotImplemented` and `raise <EXCEPTION> from NotImplemented`
- If the user catches a tuple of exceptions (`except (foo, bar, baz):`) and multiple elements in the tuple are invalid, we now collect these into a single diagnostic rather than emitting a separate diagnostic for each tuple element
- The explanation of why the `except`/`raise` was invalid ("must be a `BaseException` instance or `BaseException` subclass", etc.) is relegated to a subdiagnostic. This makes the top-level diagnostic summary much more concise.

## Test Plan

Lots of snapshots. And here's some screenshots:

<details>
<summary>Screenshots</summary>

image

image

image

</details>

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/exception/basic.md
+++ b/crates/ty_python_semantic/resources/mdtest/exception/basic.md
@@ -116,17 +116,18 @@ def silence2[T: (
 
 ## Invalid exception handlers
 
+<!-- snapshot-diagnostics -->
+
 ```py
 try:
     pass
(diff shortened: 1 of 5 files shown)


carljm (approved):
Nice

PR #18050, opened 2025-05-12
[ty] Improve diagnostics for `assert_type` and `assert_never`

AlexWaygood:
## Summary

Further work towards https://github.com/astral-sh/ty/issues/209.

I didn't make exactly the change suggested in that issue (only highlight the argument range) because I was worried that the behaviour of `type: ignore` comments might be confusing if the range of the diagnostic only covered part of the call. E.g. I'd expect this to work, as a user:

```py
assert_type(  # type: ignore
    very_very_very_long_variable_name,
    very_very_very_long_type,
)
```

I feel like in general we might want to separate the concepts of "range the diagnostic has for suppression purposes" and "primary range that is highlighted when the diagnostic is rendered on the terminal"? But that's out of scope for this PR.

Instead, this PR adds a subdiagnostic that draws the user's attention to the range of the argument passed in.

## Test Plan

Before:

image

After:

image

(I also added snapshots)

PR #21476, opened 2025-11-15
[ty] Better invalid-assignment diagnostics

sharkdp:
## Summary

Improve the diagnostic range for `invalid-assignment` diagnostics, and add source annotations for the value and target type.

closes https://github.com/astral-sh/ty/issues/1556

### Before

image

### After

image

## Ecosystem impact

Very good! Due to the wider diagnostic range, we now pick up more `# type: ignore` directives that were supposed to suppress an invalid assignment diagnostic.

## Test Plan

New snapshot tests

### DIFF
--- a/.github/workflows/ty-ecosystem-analyzer.yaml
+++ b/.github/workflows/ty-ecosystem-analyzer.yaml
@@ -67,7 +67,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@11aa5472cf9d6b9e019c401505a093112942d7bf"
+          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@0aff03414da5d242e97a9f43fb502e085637a4a1"
 
           ecosystem-analyzer \
             --repository ruff \
--- a/crates/ty_python_semantic/resources/mdtest/diagnostics/invalid_assignment.md
+++ b/crates/ty_python_semantic/resources/mdtest/diagnostics/invalid_assignment.md
@@ -1,0 +1,52 @@
+# Invalid assignment diagnostics
+
+<!-- snapshot-diagnostics -->
+
+## Annotated assignment
+
+```py
+x: int = "three"  # error: [invalid-assignment]
+```
+
+## Unannotated assignment
+
+```py
+x: int
+x = "three"  # error: [invalid-assignment]
+```
+
+## Named expression
+
+```py
+x: int
+
+(x := "three")  # error: [invalid-assignment]
+```
+
+## Multiline expressions
+
+```py
+# fmt: off
+
+# error: [invalid-assignment]
+x: str = (
+    1 + 2 + (
+        3 + 4 + 5
+    )
+)
+```
+
+## Multiple targets
+
+```py
+x: int
+y: str
+
+x, y = ("a", "b")  # error: [invalid-assignment]
+
+x, y = (0, 0)  # error: [invalid-assignment]
+```
+
+## Shadowing of classes and functions
+
+See [shadowing.md](./shadowing.md).
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiline_expression…_(f316976ffe72c6c7).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiline_expression…_(f316976ffe72c6c7).snap
@@ -1,0 +1,44 @@
+---
+source: crates/ty_test/src/lib.rs
+expression: snapshot
+---
+---
+mdtest name: invalid_assignment.md - Invalid assignment diagnostics - Multiline expressions
+mdtest path: crates/ty_python_semantic/resources/mdtest/diagnostics/invalid_assignment.md
+---
+
+# Python source files
+
+## mdtest_snippet.py
+
+```
+1 | # fmt: off
+2 | 
+3 | # error: [invalid-assignment]
+4 | x: str = (
+5 |     1 + 2 + (
+6 |         3 + 4 + 5
+7 |     )
+8 | )
+```
+
+# Diagnostics
+
+```
+error[invalid-assignment]: Object of type `Literal[15]` is not assignable to `str`
+ --> src/mdtest_snippet.py:4:4
+  |
+3 |   # error: [invalid-assignment]
+4 |   x: str = (
+  |  ____---___^
+  | |    |
+  | |    Declared type
+5 | |     1 + 2 + (
+6 | |         3 + 4 + 5
+7 | |     )
+8 | | )
+  | |_^ Incompatible value of type `Literal[15]`
+  |
+info: rule `invalid-assignment` is enabled by default
+
+```
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiple_targets_(e20ddfd7a91affb0).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiple_targets_(e20ddfd7a91affb0).snap
@@ -1,0 +1,55 @@
+---
+source: crates/ty_test/src/lib.rs
+expression: snapshot
+---
+---
+mdtest name: invalid_assignment.md - Invalid assignment diagnostics - Multiple targets
+mdtest path: crates/ty_python_semantic/resources/mdtest/diagnostics/invalid_assignment.md
+---
+
+# Python source files
+
+## mdtest_snippet.py
+
+```
+1 | x: int
+2 | y: str
+3 | 
+4 | x, y = ("a", "b")  # error: [invalid-assignment]
+5 | 
+6 | x, y = (0, 0)  # error: [invalid-assignment]
+```
+
+# Diagnostics
+
+```
+error[invalid-assignment]: Object of type `tuple[Literal["a"], Literal["b"]]` is not assignable to `int`
+ --> src/mdtest_snippet.py:4:1
+  |
+2 | y: str
+3 |
+4 | x, y = ("a", "b")  # error: [invalid-assignment]
+  | -      ^^^^^^^^^^ Incompatible value of type `tuple[Literal["a"], Literal["b"]]`
+  | |
+  | Declared type `int`
+5 |
+6 | x, y = (0, 0)  # error: [invalid-assignment]
+  |
+info: rule `invalid-assignment` is enabled by default
+
+```
+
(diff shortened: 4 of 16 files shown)


crates/ty_python_semantic/resources/mdtest/diagnostics/invalid_assignment.md
sharkdp, line 1: I thought about adding the following example as well, but that leads to a different error code, and takes a different code path.
```py
x = "three"
x: int
```

crates/ty_python_semantic/resources/mdtest/stubs/ellipsis.md
```diff
@@ -59,10 +59,10 @@ In a non-stub file, there's no special treatment of ellipsis literals. An ellips
 be assigned if `EllipsisType` is actually assignable to the annotated type.
 
 ```py
-# error: 7 [invalid-parameter-default] "Default value of type `EllipsisType` is not assignable to annotated parameter type `int`"
+# error: [invalid-parameter-default] "Default value of type `EllipsisType` is not assignable to annotated parameter type `int`"
 def f(x: int = ...) -> None: ...
 
-# error: 1 [invalid-assignment] "Object of type `EllipsisType` is not assignable to `int`"
+# error: [invalid-assignment] "Object of type `EllipsisType` is not assignable to `int`"
```
sharkdp, line 65: The target column changed for this diagnostic, so I removed those columns across the whole file.

sharkdp:
Oh, as the ecosystem report shows, we can't do this without touching the concise message. I really really want a separate API for concise diagnostic messages.

carljm:
I don't disagree about an API to give us more control over the concise message -- but in general I find this PR to be an improvement to the concise output as well?

sharkdp:
> I don't disagree about an API to give us more control over the concise message -- but in general I find this PR to be an improvement to the concise output as well?

I don't like that the value type is now mentioned twice.

sharkdp:
image

The math seems slightly off, because there is a weird edge case where the same diagnostic is (correctly) emitted twice for a single line in python-egglog, and both of them change their column. ecosystem-analyzer is not smart enough to understand this and emits a single change and a single removal, instead of two changes.

.github/workflows/ty-ecosystem-analyzer.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@11aa5472cf9d6b9e019c401505a093112942d7bf"
+          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@0aff03414da5d242e97a9f43fb502e085637a4a1"
```
sharkdp, line 70: Includes two fixes to ecosystem-analyzer to make the report here easier to read (only columns change in most cases).

crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
```diff
@@ -189,3 +189,40 @@ a = 10 + 4  # ty: ignore[division-by-zer]
...
+
+x1: str = 1 + 2 + 3  # ty: ignore
+
+x2: str = (  # ty: ignore
+    1 + 2 + 3
+)
+
+x4: str = (
+    1 + 2 + 3
+)  # ty: ignore
```
sharkdp, line 213: This does not work on `main`, and is the reason for the ecosystem changes here.

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2068,76 +2071,128 @@ pub(crate) fn is_invalid_typed_dict_literal(
...
+    let diagnostic_range = if let Some(value_node) = value_node {
+        // Expand the range to include parentheses around the value, if any. This allows
+        // invalid-assignment diagnostics to be suppressed on the opening or closing parenthesis:
+        // ```py
+        // x: str = ( # ty: ignore <- here
+        //     1 + 2 + 3
+        // )  # ty: ignore <- or here
+        // ```
+
+        let comment_ranges = CommentRanges::from(context.module().tokens());
```
sharkdp: This sounds potentially costly?

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2068,76 +2071,128 @@ pub(crate) fn is_invalid_typed_dict_literal(
...
+        let source = source_text(context.db(), context.file());
+        parentheses_iterator(value_node.into(), None, &comment_ranges, &source)
+            .last()
+            .unwrap_or(value_node.range())
```
sharkdp, line 2150: Thank you for the hint, Micha

crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiple_targets_(e20ddfd7a91affb0).snap
```diff
@@ -0,0 +1,55 @@
...
+
+# Diagnostics
+
+```
+error[invalid-assignment]: Object of type `tuple[Literal["a"], Literal["b"]]` is not assignable to `int`
+ --> src/mdtest_snippet.py:4:1
+  |
+2 | y: str
+3 |
+4 | x, y = ("a", "b")  # error: [invalid-assignment]
```
MichaReiser: This looks wrong but is tracked by https://github.com/astral-sh/ty/issues/1580

crates/ty_python_semantic/src/types/diagnostic.rs
MichaReiser: Hmm yeah, this sounds non-ideal but I think it would be fine merging as is but we should create a follow-up issue and rewrite (if possible) or have a second `parenthesized_ranges` function that takes `&Tokens` as arguments instead of the comment ranges and doesn't use `SimpleTokenizer` or `BackwardsTokenizer` at all.

MichaReiser: An uptimization that you could consider doing is to only take the tokens up to `value_node.start()`. 

```rust
let comment_ranges = CommentRanges::from(context.module().tokens().before(value_node.start());

File: crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiple_targets_(e20ddfd7a91affb0).snap (hunk 6 above)
Comment (sharkdp) [reply crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiple_targets_(e20ddfd7a91affb0).snap]: Yes, exactly. Forgot to add the reference here.

File: crates/ty_python_semantic/src/types/diagnostic.rs (hunk 4 above)
Comment (sharkdp) [reply crates/ty_python_semantic/src/types/diagnostic.rs]: Gave up trying to get this to work after a couple of minutes. I'll open as issue with what you wrote above.

Comment (MichaReiser) [reply crates/ty_python_semantic/src/types/diagnostic.rs]: You can create it in the ruff repo and mark it as good first issue

File: crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiline_expression…_(f316976ffe72c6c7).snap [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+  |  ____---___^
+  | |    |
+  | |    Declared type
+5 | |     1 + 2 + (
+6 | |         3 + 4 + 5
+7 | |     )
+8 | | )
+  | |_^ Incompatible value of type `Literal[15]`
+  |
+info: rule `invalid-assignment` is enabled by default
```
AlexWaygood, line 42: I would find this diagnostic more readable if the rendering did not include the parentheses:

image

I think the problem being fixed by including the parentheses in the diagnostic range is basically the same one that we previously discussed in https://github.com/astral-sh/ruff/pull/18050#discussion_r2085743755 (and elsewhere as well): the range you want highlighted in a diagnostic when it's rendered is often not the same range that you want a suppression comment to apply for.

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2068,76 +2071,128 @@ pub(crate) fn is_invalid_typed_dict_literal(
...
+    } else {
+        target_node.range()
+    };
```
AlexWaygood, line 2153: Hmm, why does it make sense to special-case this diagnostic in particular here? If we want enclosing parentheses to be part of the range for a suppression comment, wouldn't it make more sense to apply this change to our diagnostics infrastructure or suppression-comment infrastructure, so that all our diagnostics benefit from it

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2068,76 +2071,128 @@ pub(crate) fn is_invalid_typed_dict_literal(
...
+                        .message("Declared type"),
+                );
+            }
+            _ => {
+                // Otherwise, annotate the target with its declared type.
+                diag.annotate(context.secondary(target_node).message(format_args!(
+                    "Declared type `{}`",
+                    target_ty.display(context.db()),
+                )));
+            }
```
AlexWaygood, line 2184: sometimes the original declaration of the variable might be some way off if the scope is a long function or class and the variable is declared at the beginning of the scope -- pointing to the original declaration as well as (or instead of) the assignment target might be more helpful here?

AlexWaygood (commented):
Nice! This is a big improvement.

Don't feel the need to immediately followup on any of these comments; none of them are urgent

crates/ty_python_semantic/src/types/diagnostic.rs
sharkdp: Before I made that change, we saw ecosystem regressions (`# ty: ignore` comments that didn't match anymore) for this specific diagnostic. That's why @MichaReiser  and I discussed that we could fix this using `parentheses_iterator`. It's not immediately clear to me if this trick could/should be blindly applied to all diagnostics. Micha also mentioned the possibility of adding a new function to `Diagnostic` that would allow us to add additional suppression ranges. I'm not sure if that would have helped for this specific case, though.

crates/ty_python_semantic/resources/mdtest/snapshots/invalid_assignment.m…_-_Invalid_assignment_d…_-_Multiline_expression…_(f316976ffe72c6c7).snap
sharkdp: I agree. See below:

> Micha also mentioned the possibility of adding a new function to Diagnostic that would allow us to add additional suppression ranges.

PR #21479, opened 2025-11-16
[`flake8-simplify`] Fix truthiness assumption for non-iterable arguments in tuple/list/set calls (`SIM222`, `SIM223`)

### LINKED ISSUE #21473: SIM222 and SIM223 conflate truthiness with non-emptiness
### Summary

`Truthiness::from_expr` assumes that all values are iterable and that truthy is the same as non-empty, causing false positives in [`expr-or-true` (SIM222)](https://docs.astral.sh/ruff/rules/expr-or-true/) and [`expr-and-false` (SIM223)](https://docs.astral.sh/ruff/rules/expr-and-false/). [Examples](https://play.ruff.rs/581a78c0-58cb-4662-945b-bacb45bada01):
```console
$ cat >sim22.py <<'# EOF'
print(tuple(t"") or True)
print(tuple(t"") and False)
try:
    print(tuple(0) or True)
except TypeError as e:
    print(e)
try:
    print(tuple(1) and False)
except TypeError as e:
    print(e)
# EOF

$ python3.14 sim22.py
True
()
'int' object is not iterable
'int' object is not iterable

$ ruff --isolated check sim22.py --select SIM222,SIM223 --unsafe-fixes --fix
Found 4 errors (4 fixed, 0 remaining).

$ cat sim22.py
print(tuple(t""))
print(False)
try:
    print(True)
except TypeError as e:
    print(e)
try:
    print(False)
except TypeError as e:
    print(e)

$ python3.14 sim22.py
()
False
True
False
```

### Version

ruff 0.14.5 (87dafb878 2025-11-13)

danparizher:
## Summary

Fixes false positives in SIM222 and SIM223 where truthiness was incorrectly assumed for `tuple(x)`, `list(x)`, `set(x)` when `x` is not iterable.

Fixes #21473.

## Problem

`Truthiness::from_expr` recursively called itself on arguments to iterable initializers (`tuple`, `list`, `set`) without checking if the argument is iterable, causing false positives for cases like `tuple(0) or True` and `tuple("") or True`.

## Approach

Added `is_definitely_not_iterable` helper and updated `Truthiness::from_expr` to return `Unknown` for non-iterable arguments (numbers, booleans, None) and string literals when called with iterable initializers, preventing incorrect truthiness assumptions.

## Test Plan

Added test cases to `SIM222.py` and `SIM223.py` for `tuple("")`, `tuple(0)`, `tuple(1)`, `tuple(False)`, and `tuple(None)` with `or True` and `and False` patterns.

### DIFF
--- a/crates/ruff_python_ast/src/helpers.rs
+++ b/crates/ruff_python_ast/src/helpers.rs
@@ -1323,11 +1323,10 @@ impl Truthiness {
                         {
                             // Ex) `list([1, 2, 3])`
                             match argument {
-                                // Return Unknown for types with definite truthiness that might result
-                                // in empty iterables or will raise a type error:
-                                // - Non-iterable types (numbers, booleans, None, etc.) raise TypeError
-                                // - TString: always truthy but might result in empty iterables
-                                // - Lambda/Generator: always truthy but might result in empty iterables
+                                // Return Unknown for types with definite truthiness that might
+                                // result in empty iterables (t-strings and generators) or will
+                                // raise a type error (non-iterable types like numbers, booleans,
+                                // None, etc.).
                                 Expr::NumberLiteral(_)
                                 | Expr::BooleanLiteral(_)
                                 | Expr::NoneLiteral(_)


MichaReiser:
Thanks, @danparizher, for this PR. It's great to see how you jump on every Ruff issue. We appreciate your contributions greatly, and you're very productive. You're that productive that it is challenging for us to keep up with the number of PRs you submit; it can sometimes even feel a bit overwhelming to see 5 new PRs. 

That's why I'd like to ask you to give us some time to catch up. I think an ideal contribution experience for us (you as a contributor and we as a reviewer) would be to aim for a low one-digit number of open PRs that we aim to land before starting any new ones. That gives us the necessary time to properly review your PRs, as well as those from other contributors. 

I hope you won't receive this as discouraging because we really value your contributions and we would love to see more of them, maybe just have them a little more spaces out so that we can keep up :)

danparizher:
I'm not sure why this `ruff::cli analyze_graph` test is failing, I don't believe my changes are related to it.

Thanks so much for the kind words @MichaReiser! No worries at all, I completely get it—sorry for the PR flood! "Overzealous contributor" is probably a fitting title; I just get excited about Ruff 😄

I'll throttle back and stick to just a couple of open PRs at a time. I totally respect the need to keep the review pipeline manageable for you all and appreciate the heads-up!

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1328,6 +1341,19 @@ impl Truthiness {
...
+                                Self::Unknown
+                            } else if matches!(argument, Expr::StringLiteral(_)) {
+                                // For strings, tuple("") creates an empty tuple which is falsy,
+                                // but tuple("a") creates a non-empty tuple which is truthy.
+                                // However, we can't assume truthiness matches because:
+                                // - tuple("") is falsy (empty tuple), but "" is also falsy
+                                // - The issue is that we conflate truthiness with non-emptiness
+                                // - For SIM222/SIM223, we should be conservative and return Unknown
+                                //   to avoid false positives
+                                Self::Unknown
```
ntBre: I think we can omit this branch. `Self::from_expr` should already work correctly for string literals. The problem in the issue is for t-strings, among other types, which behave differently.

Removing this also fixes the pytest snapshot, which I think was a true positive.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1328,6 +1341,19 @@ impl Truthiness {
                             // recursing into Self::from_expr below, which returns Unknown for them.
                             if argument.is_generator_expr() {
                                 Self::Unknown
+                            } else if is_definitely_not_iterable(argument) {
+                                // For non-iterable arguments like numbers, booleans, None, etc.,
+                                // we can't assume truthiness. For example, tuple(0) raises TypeError.
+                                Self::Unknown
```
ntBre: I wonder if it might make more sense to list the types we explicitly want to recurse for instead of filtering out the cases where we don't want to recurse. Maybe just even a big `match` on `argument` would be easiest to reason about. What do you think?

Basically I think we need to return `Self::Unknown` here and avoid recursing for any type that has a definite `Self::Truthy` or `Self::Falsey` (or `Self::None`) type in one of the match arms above this, if the type might result in an empty iterable or will raise a type error.

As it stands, I don't think this PR actually fixes the t-string case from the issue.

crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM222.py
```diff
@@ -216,3 +216,11 @@ def get_items_list():
 
 def get_items_set():
     return tuple({item for item in items}) or None  # OK
+
+
+# https://github.com/astral-sh/ruff/issues/21473
+tuple("") or True  # OK
```
ntBre: We should add a t-string case:


```suggestion
tuple(t"") or True  # OK
```

ntBre (changes requested):

crates/ruff_python_ast/src/helpers.rs
danparizher: That approach makes sense to me, thanks!

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1329,7 +1329,63 @@ impl Truthiness {
...
+                                    | Expr::EllipsisLiteral(_) => {
+                                        // For non-iterable arguments like numbers, booleans, None, etc.,
+                                        // we can't assume truthiness. For example, tuple(0) raises TypeError.
+                                        Self::Unknown
+                                    }
+                                    // Types with definite truthiness that might result in empty iterables
+                                    Expr::StringLiteral(_)
+                                    | Expr::TString(_)
+                                    | Expr::FString(_)
+                                    | Expr::BytesLiteral(_) => {
```
ntBre: I think we can recurse for all of these except t-strings. `StringLiteral`, `BytesLiteral`, and `FString` all check if they are empty or not. Only `TString` maps to `Self::Truthy` unconditionally.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1329,7 +1329,63 @@ impl Truthiness {
                             if argument.is_generator_expr() {
                                 Self::Unknown
                             } else {
-                                Self::from_expr(argument, is_builtin)
+                                // Return Unknown for types with definite truthiness that might result
```
ntBre: We can fold this `is_generator_expr` call into the match on `argument` below.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1329,7 +1329,63 @@ impl Truthiness {
...
+                                    | Expr::YieldFrom(_)
+                                    | Expr::Compare(_) => Self::from_expr(argument, is_builtin),
+                                    // Return Unknown for types with definite truthiness that don't make sense to recurse
+                                    Expr::Lambda(_) | Expr::Generator(_) => {
+                                        // Lambda and Generator are always truthy, but when passed to tuple/list/set,
+                                        // they might result in empty iterables or raise errors. Return Unknown.
+                                        Self::Unknown
+                                    }
+                                    // Default to Unknown for any other types
+                                    _ => Self::Unknown,
```
ntBre: I think it would be more clear if we grouped all of the expressions into one recursive arm and one `Self::Unknown` arm. Is the default `_` matching anything here? It seems like most cases are enumerated.

I'm kind of second-guessing my suggestion to list them all, but let's get the groups sorted out conclusively, and then we can decide to trim it down if it seems excessive.

danparizher: The default `_` now matches the remaining `Expr` variants (collections like `List`, `Set`, `Tuple`, `Dict`; comprehensions; `Name`, `Call`, `Attribute`, etc.), which we recurse on.

crates/ruff_python_ast/src/helpers.rs
```diff
@@ -1322,14 +1322,29 @@ impl Truthiness {
...
+                                // in empty iterables or will raise a type error:
+                                // - Non-iterable types (numbers, booleans, None, etc.) raise TypeError
+                                // - String types: can't reliably determine truthiness of tuple("") from ""
+                                //   (tuple("") creates empty tuple, but tuple("a") creates non-empty tuple)
+                                // - Lambda/Generator: always truthy but might result in empty iterables
+                                Expr::NumberLiteral(_)
+                                | Expr::BooleanLiteral(_)
+                                | Expr::NoneLiteral(_)
+                                | Expr::EllipsisLiteral(_)
+                                | Expr::StringLiteral(_)
```
ntBre: We're still not recursing on string literals, and the pytest snapshot is still changing like I mentioned in https://github.com/astral-sh/ruff/pull/21479#discussion_r2543698051 and https://github.com/astral-sh/ruff/pull/21479#discussion_r2535574572.

PR #21480, opened 2025-11-16
[`parser`] Fix panic when parsing IPython escape command expressions

### LINKED ISSUE #21465: panic: `internal error: entered unreachable code: IPython escape command expression is only allowed for % and !`
### Summary

This panic was originally reported without a reproducer at https://github.com/facebook/pyrefly/issues/1559 .  I was able to track it down with some targeted fuzzing:

```
panic: Panicked at crates/ruff_python_parser/src/parser/expression.rs:2782:13 when checking `/home/user/crash.ipynb`: `internal error: entered unreachable code: IPython escape command expression is only allowed for % and !`
--> crash.ipynb:1:1
```

**crash.ipynb**
```ipython
{
  "cells": [
    {
      "cell_type": "code",
      "metadata": {},
      "outputs": [],
      "source": [
        "with a,?b\n?"
      ]
    }
  ],
  "metadata": {},
  "nbformat": 4,
  "nbformat_minor": 2
}
```

### Version

ruff 0.14.5

danparizher:
## Summary

Fixes a panic when parsing IPython escape commands with `Help` kind (`?`) in expression contexts. The parser now reports an error instead of panicking.

Fixes #21465.

## Problem

The parser panicked with `unreachable!()` in `parse_ipython_escape_command_expression` when encountering escape commands with `Help` kind (`?`) in expression contexts, where only `Magic` (`%`) and `Shell` (`!`) are allowed.

## Approach

Replaced the `unreachable!()` panic with error handling that adds a `ParseErrorType::OtherError` and continues parsing, returning a valid AST node with the error attached.

## Test Plan

Added `test_ipython_escape_command_in_with_statement` and `test_ipython_help_escape_command_as_expression` to verify the fix.


crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -2777,18 +2777,24 @@ impl<'src> Parser<'src> {
             unreachable!()
         };
 
-        if !matches!(kind, IpyEscapeKind::Magic | IpyEscapeKind::Shell) {
-            // This should never occur as the lexer won't allow it.
```
MichaReiser, line 2781: Can you explain to me why the lexer emits the `Magic` and `Shell` tokens contrary to what the comment says? Should we fix the lexer instead?

danparizher: That makes more sense, yeah. I'm pushing a change to have it fixed at the lexer level by adding an `in_expression_context` parameter to `lex_ipython_escape_command` and preventing Help kind conversion when `true` (e.g., `%foo?` stays Magic in expression contexts).

I think it makes sense to keep the parser check as a defensive measure for edge cases like `with a,?b`, where Help tokens can still be lexed at the start of logical lines. Does this approach look right?

dhruvmanila:
Thanks for putting up this PR!

I think the problem is in the way the parser is recovering. Considering your example:
```py
with a, ?b
?
```

As there's a newline before the final `?`, the lexer emits the `IpyEscapeCommand` token. Now, the parser is trying to parse a list of expressions that comprises of `with` items but (1) it tries to recover from an unexpected `?` before the `b` and then it doesn't consider the newline token (`\n`) as the end of the with items given that this is unparenthesized. I think a (better?) fix would be to try to make the parser recover from this newline token and consider that the end of with items so that the second `?` is parsed as statement which will then avoid this panic.

What do you think? This is what I'm referring to:

```diff
diff --git a/crates/ruff_python_parser/src/parser/mod.rs b/crates/ruff_python_parser/src/parser/mod.rs
index 5f35b84b04..59c1857152 100644
--- a/crates/ruff_python_parser/src/parser/mod.rs
+++ b/crates/ruff_python_parser/src/parser/mod.rs
@@ -1115,7 +1115,12 @@ impl RecoveryContextKind {
                     TokenKind::Colon => Some(ListTerminatorKind::ErrorRecovery),
                     _ => None,
                 },
-                WithItemKind::Unparenthesized | WithItemKind::ParenthesizedExpression => p
+                WithItemKind::Unparenthesized => matches!(
+                    p.current_token_kind(),
+                    TokenKind::Colon | TokenKind::Newline
+                )
+                .then_some(ListTerminatorKind::Regular),
+                WithItemKind::ParenthesizedExpression => p
                     .at(TokenKind::Colon)
                     .then_some(ListTerminatorKind::Regular),
             },
```

Edit: Oh, I think it should maybe be restricted to `WithItemKind::Unparenthesized`. I've updated the diff snippet.

dhruvmanila:
> What do you think? This is what I'm referring to:

Ok, I tried this and it seems to be working well and I think this is more correct approach. I've tried it on various code snippets to use parentheses in various places:

```py
with (a, ?b)
?
```

```py
with (a, ?b
?
```

```py
with a, ?b
?
```

crates/ruff_python_parser/src/parser/tests.rs
```diff
@@ -157,3 +157,43 @@ t"i}'"#;
...
+    // Only `%` and `!` escape commands are allowed as expressions.
+    let source = "?foo";
+
+    // Try to parse as an expression - this should not panic
+    let parsed = parse_expression(source);
+
+    // Should return an error, not panic
+    let error = parsed.unwrap_err();
+    insta::assert_debug_snapshot!(error);
+}
```
MichaReiser: Can you convert those tests to inline tests instead

https://github.com/astral-sh/ruff/blob/c4240076459bc9128fee6f83cbf0f51d134b663b/crates/ruff_python_parser/CONTRIBUTING.md#L26-L28

crates/ruff_python_parser/src/parser/tests.rs
```diff
@@ -157,3 +157,43 @@ t"i}'"#;
...
     insta::assert_debug_snapshot!(error);
 }
+
+#[test]
+fn test_ipython_escape_command_in_with_statement() {
+    // Regression test for https://github.com/astral-sh/ruff/issues/21465
+    // This should not panic when encountering a `?` escape command in a with statement
+    // The original issue was that parsing `with a,?b\n?` would panic when trying to
+    // parse `?b` as an expression. Now it should handle it gracefully with an error.
+    let source = "with a,?b\n?";
```
MichaReiser: Can you add some case arms or statements coming after `with` so that it also demonstrates that the parser recovers properly?

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -2777,18 +2784,28 @@ impl<'src> Parser<'src> {
...
-            unreachable!("IPython escape command expression is only allowed for % and !");
-        }
-
```
dhruvmanila, line 2784: I think this needs to be reverted now that the parser correctly recovers from the syntax error

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -2767,7 +2767,14 @@ impl<'src> Parser<'src> {
...
     /// If the parser isn't positioned at a `IpyEscapeCommand` token.
-    /// If the escape command kind is not `%` or `!`.
+    ///
+    /// # Note
+    ///
+    /// The lexer ensures that only `%` (Magic) and `!` (Shell) escape commands are emitted
+    /// in expression contexts (after `=`). It prevents Help kind conversion even if the command
+    /// ends with `?`. However, Help escape commands can still be lexed at the start of logical
+    /// lines (e.g., `?b` in `with a,?b`) and may then be parsed as expressions. This check
+    /// ensures we report an error instead of panicking in such cases.
```
dhruvmanila: This needs to be reverted as well

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -2777,18 +2784,28 @@ impl<'src> Parser<'src> {
...
+            // IPython escape commands like `?` (Help) are not allowed in expression contexts.
+            // Only `%` (Magic) and `!` (Shell) escape commands can be used as expressions.
+            // The lexer should prevent this, but we check here as a defensive measure.
+            self.add_error(
+                ParseErrorType::OtherError(format!(
+                    "IPython escape command expression is only allowed for % and !, found {kind:?}"
+                )),
+                &command,
+            );
+        }
```
dhruvmanila: And this, you can move the test cases above the recovery branch in `parser/mod.rs`

crates/ruff_python_parser/src/parser/statement.rs
```diff
@@ -2203,6 +2203,13 @@ impl<'src> Parser<'src> {
                 )
             }
         } else {
+            // test_err ipython_escape_command_in_with_statement
+            // # parse_options: {"mode": "ipython"}
+            // with a,?b
+            //     pass
+            // with a,?b
+            //     x = 1
+            //     y = 2
```
dhruvmanila: We should add some test cases that contains parentheses like the ones that I've mentioned in my comment

crates/ruff_python_parser/src/lexer.rs
```diff
@@ -1232,7 +1234,15 @@ impl<'src> Lexer<'src> {
...
-    fn lex_ipython_escape_command(&mut self, escape_kind: IpyEscapeKind) -> TokenKind {
+    ///
+    /// `in_expression_context` indicates whether we're in an expression context (after `=`).
+    /// In expression contexts, only `%` (Magic) and `!` (Shell) escape commands are allowed,
+    /// so we don't convert to Help kind even if the command ends with `?`.
+    fn lex_ipython_escape_command(
+        &mut self,
+        escape_kind: IpyEscapeKind,
+        in_expression_context: bool,
+    ) -> TokenKind {
```
dhruvmanila: This should be reverted as well

crates/ruff_python_parser/src/lexer.rs
```diff
@@ -1292,11 +1302,17 @@ impl<'src> Lexer<'src> {
...
+                    // conversion even if the command ends with `?`. Only `%` (Magic) and `!` (Shell)
+                    // escape commands are allowed as expressions.
+                    if in_expression_context
+                        || question_count > 2
                         || value.chars().last().is_none_or(is_python_whitespace)
                         || !matches!(self.cursor.first(), '\n' | '\r' | EOF_CHAR)
                     {
-                        // Not a help end escape command, so continue with the lexing.
+                        // Not a help end escape command (or in expression context where Help is not allowed),
+                        // so continue with the lexing.
```
dhruvmanila: And this

dhruvmanila (changes requested):
Can you revert the original solution now that the parser recovers correctly? I'd also recommend to add some more test cases especially the ones that contains parentheses.

dhruvmanila:
Hmm, actually the test cases now don't really test the bug which, I think, requires the `?` on the next line as mentioned here (https://github.com/astral-sh/ruff/pull/21480#issuecomment-3540148662). I'll update the test cases and merge.

PR #21496, opened 2025-11-17
[ty] Implicit type aliases: Add support for `Callable`

sharkdp:
## Summary

Add support for `Callable` special forms in implicit type aliases.

## Typing conformance

Four new tests are passing

## Ecosystem impact

* All of the `invalid-type-form` errors are from libraries that use `mypy_extensions` and do something like `Callable[[NamedArg("x", str)], int]`.
* A handful of new false positives because we do not support generic specializations of implicit type aliases, yet. But other
* Everything else looks like true positives or known limitations

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -839,6 +839,10 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     }
                     Type::unknown()
                 }
+                KnownInstanceType::Callable(_) => {
+                    self.infer_type_expression(slice);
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -929,6 +939,58 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         ty
     }
 
+    /// Infer the type of a `Callable[...]` type expression.
+    pub(crate) fn infer_callable_type(&mut self, subscript: &ast::ExprSubscript) -> Type<'db> {
```
sharkdp, line 937: All of the code here has just been extracted into a new function. Nothing changed.

PR #21497, opened 2025-11-17
[ty] Use `CompactStr` for `StringLiteralType`

MichaReiser:
Most `StringLiteralType` should be short (<24 bytes). We can use a `CompactStr` to avoid an extra allocation for short strings.

The benchmark show a 1-2% improvement for the wall-time benchmarks and a 1-2% regression for the instrumented benchmarks. This isn't unexpected, given that the instrumented benchmarks don't account for allocations.


sharkdp (approved):
Nice!

PR #21498, opened 2025-11-17
[ty] Custom concise diagnostic messages

sharkdp:
## Summary

This PR proposes that we add a new `set_concise_message` functionality to our `Diagnostic` construction API. When used, the concise message that is otherwise auto-generated from the main diagnostic message and the primary annotation will be overwritten with the custom message.

To understand why this is desirable, let's look at the `invalid-key` diagnostic. This is how I *want* the full diagnostic to look like:

image

However, without the change in this PR, the concise message would have the following form:

```
error[invalid-key]: Unknown key "Age" for TypedDict `Person`: Unknown key "Age" - did you mean "age"?
```

This duplication is why the full `invalid-key` diagnostic used a main diagnostic message that is only "Invalid key for TypedDict `Person`", to make that bearable:

```
error[invalid-key] Invalid key for TypedDict `Person`: Unknown key "Age" - did you mean "age"?
```

This is still less than ideal, *and* we had to make the "full" diagnostic worse. With the new API here, we have to make no such compromises. We need to do slightly more work (provide one additional custom-designed message), but we get to keep the "full" diagnostic that we actually want, and we can make the concise message more terse and readable:

```
error[invalid-key] Unknown key "Age" for TypedDict `Person` - did you mean "age"?
```

Similar problems exist for other diagnostics as well (I really want this for https://github.com/astral-sh/ruff/pull/21476). In this PR, I only changed `invalid-key` and `type-assertion-failure`.

The PR here is somewhat related to the discussion in https://github.com/astral-sh/ty/issues/1418, but note that we are solving a problem that is unrelated to sub-diagnostics.

## Test Plan

Updated tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
@@ -36,7 +36,7 @@ error[invalid-key]: Unknown key "surname" for TypedDict `Person`
 11 |     # error: [invalid-key]
 12 |     # error: [invalid-key]
 13 |     being["surname"] = "unknown"
-   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     ----- ^^^^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person` in union type `Person | Animal`
    |
@@ -51,7 +51,7 @@ error[invalid-key]: Unknown key "surname" for TypedDict `Animal`
 11 |     # error: [invalid-key]
 12 |     # error: [invalid-key]
 13 |     being["surname"] = "unknown"
-   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     ----- ^^^^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Animal` in union type `Person | Animal`
    |
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Misspelled_key_for_`…_(7cf0fa634e2a2d59).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Misspelled_key_for_`…_(7cf0fa634e2a2d59).snap
@@ -29,7 +29,7 @@ error[invalid-key]: Unknown key "Retries" for TypedDict `Config`
   |
 6 | def _(config: Config) -> None:
 7 |     config["Retries"] = 30.0  # error: [invalid-key]
-  |     ------ ^^^^^^^^^ Unknown key "Retries" - did you mean "retries"?
+  |     ------ ^^^^^^^^^ Did you mean "retries"?
   |     |
   |     TypedDict `Config`
   |
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
@@ -62,7 +62,7 @@ error[invalid-key]: Unknown key "naem" for TypedDict `Person`
    |
  7 | def access_invalid_literal_string_key(person: Person):
  8 |     person["naem"]  # error: [invalid-key]
-   |     ------ ^^^^^^ Unknown key "naem" - did you mean "name"?
+   |     ------ ^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person`
  9 |
@@ -135,7 +135,7 @@ error[invalid-key]: Unknown key "naem" for TypedDict `Person`
    |
 21 | def write_to_non_existing_key(person: Person):
 22 |     person["naem"] = "Alice"  # error: [invalid-key]
-   |     ------ ^^^^^^ Unknown key "naem" - did you mean "name"?
+   |     ------ ^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person`
 23 |
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -3167,16 +3167,21 @@ pub(crate) fn report_invalid_key_on_typed_dict<'db>(
                 });
 
                 let existing_keys = items.iter().map(|(name, _)| name.as_str());
-                let hint = if let Some(suggestion) = did_you_mean(existing_keys, key) {
-                    format!(" - did you mean \"{suggestion}\"?")
+                if let Some(suggestion) = did_you_mean(existing_keys, key) {
+                    if key_node.is_expr_string_literal() {
+                        diagnostic
+                            .set_primary_message(format_args!("Did you mean \"{suggestion}\"?"));
+                    } else {
+                        diagnostic.set_primary_message(format_args!(
+                            "Unknown key \"{key}\" - did you mean \"{suggestion}\"?",
+                        ));
+                    }
+                    diagnostic.set_concise_message(format_args!(
+                        "Unknown key \"{key}\" for TypedDict `{typed_dict_name}` - did you mean \"{suggestion}\"?",
+                    ));
                 } else {
-                    String::new()
-                };
-
-                diagnostic.set_primary_message(format!("Unknown key \"{key}\"{hint}"));
-                diagnostic.set_concise_message(format!(
-                    "Unknown key \"{key}\" for TypedDict `{typed_dict_name}`{hint}",
-                ));
+                    diagnostic.set_primary_message(format_args!("Unknown key \"{key}\""));
+                }
             }
             _ => {
                 let mut diagnostic = builder.into_diagnostic(format_args!(


sharkdp:
mypy_primer is timing out, probably because the diff is too large. Ecosystem analyzer finishes just fine.

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Unknown_key_for_all_…_(1c685d9d10678263).snap
```diff
@@ -45,7 +45,7 @@ info: rule `invalid-key` is enabled by default
 ```
 
 ```
-error[invalid-key]: Invalid key for TypedDict `Animal`
+error[invalid-key]: Unknown key "surname" for TypedDict `Animal`
   --> src/mdtest_snippet.py:13:5
    |
 11 |     # error: [invalid-key]
```
AlexWaygood: I don't love that with the default output format, the phrase `Unknown key "surname"` is now repeated in both the diagnostic summary and the primary annotation. What about something like this?

<details>
<summary>Patch</summary>

```diff
diff --git "a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Misspelled_key_for_`\342\200\246_(7cf0fa634e2a2d59).snap" "b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Misspelled_key_for_`\342\200\246_(7cf0fa634e2a2d59).snap"
index b65345fd37..ca95ce1675 100644
--- "a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Misspelled_key_for_`\342\200\246_(7cf0fa634e2a2d59).snap"
+++ "b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Misspelled_key_for_`\342\200\246_(7cf0fa634e2a2d59).snap"
@@ -29,7 +29,7 @@ error[invalid-key]: Unknown key "Retries" for TypedDict `Config`
   |
 6 | def _(config: Config) -> None:
 7 |     config["Retries"] = 30.0  # error: [invalid-key]
-  |     ------ ^^^^^^^^^ Unknown key "Retries" - did you mean "retries"?
+  |     ------ ^^^^^^^^^ Did you mean "retries"?
   |     |
   |     TypedDict `Config`
   |
diff --git "a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Unknown_key_for_all_\342\200\246_(1c685d9d10678263).snap" "b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Unknown_key_for_all_\342\200\246_(1c685d9d10678263).snap"
index 6e7a0eba6a..6c9e4de308 100644
--- "a/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Unknown_key_for_all_\342\200\246_(1c685d9d10678263).snap"
+++ "b/crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti\342\200\246_-_Subscript_assignment\342\200\246_-_Unknown_key_for_all_\342\200\246_(1c685d9d10678263).snap"
@@ -36,7 +36,7 @@ error[invalid-key]: Unknown key "surname" for TypedDict `Person`
 11 |     # error: [invalid-key]
 12 |     # error: [invalid-key]
 13 |     being["surname"] = "unknown"
-   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     ----- ^^^^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person` in union type `Person | Animal`
    |
@@ -51,7 +51,7 @@ error[invalid-key]: Unknown key "surname" for TypedDict `Animal`
 11 |     # error: [invalid-key]
 12 |     # error: [invalid-key]
 13 |     being["surname"] = "unknown"
-   |     ----- ^^^^^^^^^ Unknown key "surname" - did you mean "name"?
+   |     ----- ^^^^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Animal` in union type `Person | Animal`
    |
diff --git a/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap b/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
index e2e6ac1bdf..cdd2949031 100644
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/typed_dict.md_-_`TypedDict`_-_Diagnostics_(e5289abf5c570c29).snap
@@ -62,7 +62,7 @@ error[invalid-key]: Unknown key "naem" for TypedDict `Person`
    |
  7 | def access_invalid_literal_string_key(person: Person):
  8 |     person["naem"]  # error: [invalid-key]
-   |     ------ ^^^^^^ Unknown key "naem" - did you mean "name"?
+   |     ------ ^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person`
  9 |
@@ -78,7 +78,7 @@ error[invalid-key]: Unknown key "naem" for TypedDict `Person`
    |
 12 | def access_invalid_key(person: Person):
 13 |     person[NAME_KEY]  # error: [invalid-key]
-   |     ------ ^^^^^^^^ Unknown key "naem" - did you mean "name"?
+   |     ------ ^^^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person`
 14 |
@@ -135,7 +135,7 @@ error[invalid-key]: Unknown key "naem" for TypedDict `Person`
    |
 21 | def write_to_non_existing_key(person: Person):
 22 |     person["naem"] = "Alice"  # error: [invalid-key]
-   |     ------ ^^^^^^ Unknown key "naem" - did you mean "name"?
+   |     ------ ^^^^^^ Did you mean "name"?
    |     |
    |     TypedDict `Person`
 23 |
diff --git a/crates/ty_python_semantic/src/types/diagnostic.rs b/crates/ty_python_semantic/src/types/diagnostic.rs
index 68f42ea1a4..810a45f658 100644
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -3167,16 +3167,14 @@ pub(crate) fn report_invalid_key_on_typed_dict<'db>(
                 });
 
                 let existing_keys = items.iter().map(|(name, _)| name.as_str());
-                let hint = if let Some(suggestion) = did_you_mean(existing_keys, key) {
-                    format!(" - did you mean \"{suggestion}\"?")
+                if let Some(suggestion) = did_you_mean(existing_keys, key) {
+                    diagnostic.set_primary_message(format_args!("Did you mean \"{suggestion}\"?"));
+                    diagnostic.set_concise_message(format_args!(
+                        "Unknown key \"{key}\" for TypedDict `{typed_dict_name}` - did you mean \"{suggestion}\"?",
+                    ));
                 } else {
-                    String::new()
-                };
-
-                diagnostic.set_primary_message(format!("Unknown key \"{key}\"{hint}"));
-                diagnostic.set_concise_message(format!(
-                    "Unknown key \"{key}\" for TypedDict `{typed_dict_name}`{hint}",
-                ));
+                    diagnostic.set_primary_message(format_args!("Unknown key \"{key}\""));
+                }
             }
             _ => {
                 let mut diagnostic = builder.into_diagnostic(format_args!(
```

</details>

sharkdp: Hm, that might be slightly weird in cases where the key is not a literal string expression, right? Like if I use `MY_KEY: Final = "my_key"`, and then access the typed dict with `my_typed_dict[MY_KEY]`. It then looks like we literally want you to replace `MY_KEY` with `"correctly_spelled_key"`.

AlexWaygood: Right, fair point. You could add another check to see whether it was an `ExprStringLiteral` node that was used to subscript the `TypedDict` (which I'd expect to be the common case)

BurntSushi: You might also consider just setting the primary annotation to "unknown key" and adding the "did you mean" as a note sub-diagnostic. I guess that might still be a little weird with the `MY_KEY` example, but maybe that weirdness is best addressed by @AlexWaygood's suggestion.

AlexWaygood:
Like @sharkdp, I have on multiple occasions at this point deliberately created a worse verbose diagnostic for the sake of creating a concise diagnostic that would be usable. So I do agree that the current situation is hurting all our output formats, not just concise output formats

> this is predicated on the assumption that the quality of the concise output isn't That Important,

I have a somewhat strong opinion that it's not an acceptable outcome for us to have a concise output format that's significantly worse than the concise output format of existing, established type checkers :-)

I also think we should weigh the experience of our users over the experience of us as developers. Both are important, of course, and obviously not every possible complication to our diagnostic model would be worth it just because it improves the user experience a little. But for me, I think this one clearly meets the bar.

MichaReiser:
As long as using this override remains the exception, I think it's fine.

sharkdp:
> I realize this is framed as also improving our verbose diagnostics, but I would consider relegating concise diagnostics to "best effort" only and improve the verbose diagnostic while mostly ignoring the quality of the concise output. But this is predicated on the assumption that the quality of the concise output isn't That Important, and perhaps my instincts are wrong there.

I wouldn't worry too much about concise diagnostic messages if they were mostly for our own use (mdtests, ecosystem results) and only rarely user-facing. But concise diagnostic messages are used very prominently in IDEs, and not all of them can show annotations. Here's neovim with ty `main` ("what is the inferred type?"):

image

MichaReiser:
> I wouldn't worry too much about concise diagnostic messages if they were mostly for our own use (mdtests, ecosystem results) and only rarely user-facing. But concise diagnostic messages are used very prominently in IDEs, and not all of them can show annotations. Here's neovim with ty main ("what is the inferred type?"):

Hmm, I think this raises a good point. The LSP now uses a combination of concise + related diagnostics that isn't tested on the CLI at all. 

I think we may want to change `to_lsp_diagnostic` to use the normal primary message if the client supports related information ([see](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticClientCapabilities))

MichaReiser:
> use

Is this something you plan as a follow-up? If not, we should at least ensure that we create an issue

sharkdp:
Will open an issue later => https://github.com/astral-sh/ty/issues/1582

PR #21500, opened 2025-11-17
Mention `force-exclude` in "Configuration > Python file discovery"

thejcannon:
## Summary

This is somewhat of a footgun, so its worth mentioning here as an escape hatch for folks.

## Test Plan

N/A

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -345,8 +345,9 @@ formatting `.pyi` files, but would continue to include them in linting:
 By default, Ruff will also skip any files that are omitted via `.ignore`, `.gitignore`,
 `.git/info/exclude`, and global `gitignore` files (see: [`respect-gitignore`](settings.md#respect-gitignore)).
 
-Files that are passed to `ruff` directly are always analyzed, regardless of the above criteria.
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -345,8 +345,9 @@ formatting `.pyi` files, but would continue to include them in linting:
 By default, Ruff will also skip any files that are omitted via `.ignore`, `.gitignore`,
 `.git/info/exclude`, and global `gitignore` files (see: [`respect-gitignore`](settings.md#respect-gitignore)).
 
-Files that are passed to `ruff` directly are always analyzed, regardless of the above criteria.
-For example, `ruff check /path/to/excluded/file.py` will always lint `file.py`.
+Files that are passed to `ruff` directly are always analyzed, regardless of the above criteria,
```
thejcannon, line 348: Is this comma after "criteria" awkward? I tend to use a lot of commas so I don't mind it, but I'm also aware that I tend to use a lot of commas, soooo...

MichaReiser: I'm the wrong person to ask as a native German speaker 😆 but I think it helps readability

MichaReiser (approved):
Thank you

Issue #19226, opened 2025-07-09
Panic `assertion failed: start.raw <= end.raw` in `ruff_python_formatter/src/expression/binary_like.rs`


qarmin:
### Summary

File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
if''and(not#
0):t
```

command
```
timeout -v 300 ruff format TEST___FILE.py
```

cause this
```
warning: Detected debug build without --no-cache.
error: Panicked while formatting /opt/BROKEN_FILES_DIR/278_IDX_0_RAND_1418773717408230476762071_minimized_237.py: This indicates a bug in Ruff. If you could open an issue at:

    https://github.com/astral-sh/ruff/issues/new?title=%5BFormatter%20panic%5D

...with the relevant file contents, the `pyproject.toml` settings, and the following stack trace, we'd be very appreciative!

panicked at crates/ruff_python_formatter/src/expression/binary_like.rs:843:37:
assertion failed: start.raw <= end.raw
Backtrace:    0: ruff_db::panic::install_hook::{{closure}}::{{closure}}
             at ./ruff-main/crates/ruff_db/src/panic.rs:91:34
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/alloc/src/boxed.rs:1985:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/std/src/panicking.rs:841:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/std/src/panicking.rs:699:13
   4: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/std/src/sys/backtrace.rs:168:18
   5: __rustc::rust_begin_unwind
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/std/src/panicking.rs:697:5
   6: core::panicking::panic_fmt
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/core/src/panicking.rs:75:14
   7: core::panicking::panic
             at /rustc/a2d45f73c70d9dec57140c9412f83586eda895f8/library/core/src/panicking.rs:145:5
[...]

PR #21410, opened 2025-11-12
Fix panic when formatting comments in unary expressions

ntBre:
Summary
--

This is a second attempt that fixes #19226 based on the feedback in https://github.com/astral-sh/ruff/pull/20494#issuecomment-3467920065.

We currently mark the comment in an expression like this:

```py
if '' and (not #
0):
    pass
```

as a leading comment on `not`, eventually causing a panic in
`Operand::has_unparenthesized_leading_comments` because the end of the comment
is greater than the start of the expression.

https://github.com/astral-sh/ruff/blob/a1d9cb5830eca9b63e7fb529504fc536e99bca23/crates/ruff_python_formatter/src/expression/binary_like.rs#L843

This PR fixes the issue by instead making such a comment a dangling comment on
the unary expression.

In the third commit, I instead tried making the comment a leading comment on the
operand, which also looks pretty reasonable to me. Making it a dangling comment
seems more in line with the docs on `handle_unary_op_comments`, though.

I also tried deleting the leading comment logic in favor of the new dangling
logic in the fifth commit before reverting in the sixth. This looks okay to me
too, but the current state of the PR seems like the least invasive fix.

Test Plan
--

A new, minimized test case based on the issue. I also checked that the original
snippet from the report works now.

PR #21501, opened 2025-11-17
Fix panic when formatting comments in unary expressions

ntBre:
## Summary

This is another attempt at https://github.com/astral-sh/ruff/pull/21410 that fixes https://github.com/astral-sh/ruff/issues/19226.

@MichaReiser helped me get something working in a very helpful pairing session. I pushed one additional commit moving the comments back from leading comments to trailing comments, which I think retains more of the input formatting.

I was inspired by Dylan's PR (#21185) to make one of these tables:

<table>
                <thead>
                    <tr>
                    <th scope="col">Input</th>
                    <th scope="col">Main</th>
                    <th scope="col">PR</th>
                    </tr>
                </thead>
                <tbody>
<tr>
<td><pre lang="python">
if (
    not
    # comment
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass
</pre></td>
<td><pre lang="python">
if (
    # comment
    not aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass

</pre></td>
<td><pre lang="python">
if (
    not
    # comment
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass

</pre></td>
</tr>
<tr>
<td><pre lang="python">
if (
    # unary comment
    not
    # operand comment
    (
        # comment
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
        + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
    )
):
    pass
</pre></td>
<td><pre lang="python">
if (
    # unary comment
    # operand comment
    not (
        # comment
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
        + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
    )
):
    pass

</pre></td>
<td><pre lang="python">
if (
    # unary comment
    not
    # operand comment
    (
        # comment
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
        + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
    )
):
    pass

</pre></td>
</tr>
<tr>
<td><pre lang="python">
if (
    not # comment
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass
</pre></td>
<td><pre lang="python">
if (  # comment
    not aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass

</pre></td>
<td><pre lang="python">
if (
    not aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa  # comment
    + bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
):
    pass

</pre></td>
</tr>
</tbody>
            </table>

hopefully it helps even though the snippets are much wider here.

The two main differences are (1) that we now retain own-line comments between the unary operator and its operand instead of moving these to leading comments on the operator itself, and (2) that we move end-of-line comments between the operator and operand to dangling end-of-line comments on the operand (the last example in the table).

## Test Plan

Existing tests, plus new ones based on the issue. As I noted below, I also ran the output from main on the unary.py file back through this branch to check that we don't reformat code from main. This made me feel a bit better about not preview-gating the changes in this PR.

```shell
> git show main:crates/ruff_python_formatter/resources/test/fixtures/ruff/expression/unary.py | ruff format - | ./target/debug/ruff format --diff -
> echo $?
0
```


crates/ruff_python_formatter/tests/snapshots/format@parentheses__expression_parentheses_comments.py.snap
```diff
@@ -194,17 +193,17 @@ x = (
         "c"
     ),
     # 1
-    # 2
-    not (  # 3
+    not (  # 2  # 3
         # 4
         "d"
     ),
 )
```
ntBre, line 200: Not sure about this one. 

Input:

```py
x = (
    # 1
    not # 2
    ( # 3
        # 4
        "d"
    )
)
```

Old output:

```py
x = (
    # 1
    # 2
    not (  # 3
        # 4
        "d"
    )
)
```

and before the last commit it was like this:

```py
x = (
    # 1
    not
    # 2
    (  # 3
        # 4
        "d"
    ),
)
```

which I think might be closer to what we want, although `# 2` is an end-of-line comment in the input. That's what motivated the last commit, preserving the end-of-line status of the comments.

MichaReiser: I like the new output.

crates/ruff_python_formatter/src/comments/placement.rs
```diff
@@ -1907,24 +1908,9 @@ fn handle_unary_op_comment<'a>(
...
-            | SimpleTokenKind::Not
-            | SimpleTokenKind::Plus
-            | SimpleTokenKind::Minus
-    )));
-    let up_to = tokenizer
-        .find(|token| token.kind == SimpleTokenKind::LParen)
-        .map_or(unary_op.operand.start(), |lparen| lparen.start());
-    if comment.end() < up_to {
-        CommentPlacement::leading(unary_op, comment)
+    let up_to = operand_start(unary_op, source);
```
MichaReiser: Can you update the method description to match our new behavior

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
```diff
@@ -39,19 +42,43 @@ impl FormatNodeRule<ExprUnaryOp> for FormatExprUnaryOp {
...
+            comments.ranges(),
+            f.context().source(),
+        );
+        let has_leading_comments_before_parens = parenthesized_operand_range.is_some_and(|range| {
+            comments
+                .leading(operand.as_ref())
+                .iter()
+                .any(|comment| comment.start() < range.start())
+        });
         if comments.has_leading(operand.as_ref())
```
MichaReiser: Let's assign the leading commnts to a variable to avoid retrieving them twice

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
```diff
@@ -76,17 +103,51 @@ impl NeedsParentheses for ExprUnaryOp {
...
+        }
+
+        if is_expression_parenthesized(
             self.operand.as_ref().into(),
             context.comments().ranges(),
             context.source(),
         ) {
-            OptionalParentheses::Never
-        } else if context.comments().has(self.operand.as_ref()) {
+            return OptionalParentheses::Never;
```
MichaReiser, line 98: Do we need to change the logic here too to match the logic for when we insert a hard line break in the unary formatting?

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
```diff
@@ -76,17 +103,51 @@ impl NeedsParentheses for ExprUnaryOp {
...
+        }
+
+        let operand_start = operand_start(self, context.source());
+        if context
+            .comments()
+            .dangling(self)
+            .iter()
+            .any(|comment| comment.end() < operand_start)
```
MichaReiser: I think we can simplify this to returning `Multiline` when there's any dangling comment. 

Does this need to take precedence over the `Never` case when the `operand` is parenthesized?

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
ntBre: Do you mean something like this?

```rust
        if !context.comments().has_leading(self.operand.as_ref())
            || is_expression_parenthesized(
                self.operand.as_ref().into(),
                context.comments().ranges(),
                context.source(),
            )
        {
            return OptionalParentheses::Never;
        }
```

I played with a few variations on this and kept running into instabilities. It seems to be working okay without matching the check exactly, like on main.

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
ntBre: It seems to work in both orders, at least with our current tests.

ntBre:
> I haven't thought it through but does this change require preview gating?
> 
> I think what will help us to make this decision is if you update your summary and describe how the formatting, specifically the comment placement, changes compared to main.

I did think a little bit about this. I think it would be a bit difficult to preview-gate this since the panic fix and the new formatting seem intertwined.

I verified that taking the formatted output from `main` (the `## Output` section from the `unary.py` snapshot) and running it against this branch shows no changes. So we won't change any code that we previously formatted, at least.

I will also expand the summary, though!

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
MichaReiser: No, more like this:


```
        let parenthesized_operand_range = parenthesized_range(
            operand.into(),
            item.into(),
            comments.ranges(),
            f.context().source(),
        );
        let leading_operand_comments = comments.leading(operand.as_ref());
        let has_leading_comments_before_parens = parenthesized_operand_range.is_some_and(|range| {
            leading_operand_comments
                .iter()
                .any(|comment| comment.start() < range.start())
        });
        if !leading_operand_comments.is_empty()
            && !is_expression_parenthesized(
                operand.as_ref().into(),
                f.context().comments().ranges(),
                f.context().source(),
            )
            || has_leading_comments_before_parens
```

It's important that it exactly mirrors the case when we insert a hard line break in the formatting code because **any** line break will lead to invalid syntax if the `if` formatting doesn't add parentheses. 

Here's an example where your PR produces invalid syntax:

```py
if (
  not  
  # comment
  (a)):
    pass
```

We should add more tests that exercise the new leading comment placement (may even be true for the trailing comment placement, are there more combinations that you could test?)

MichaReiser:
Thanks for updating the summary. I think it should be safe to not preview gate this change because:

* It changes the placement of trailing operator comments, but main always moved trailing operator comments off the operator. 
* It changes the placement of leading operator comments, but main always made leading operator comments leading unary comments.

That means, any ruff formatted code can't contain any comment for which we now preserve the position

crates/ruff_python_formatter/src/expression/expr_unary_op.rs
```diff
@@ -76,17 +83,51 @@ impl NeedsParentheses for ExprUnaryOp {
         context: &PyFormatContext,
     ) -> OptionalParentheses {
         if parent.is_expr_await() {
-            OptionalParentheses::Always
-        } else if is_expression_parenthesized(
+            return OptionalParentheses::Always;
+        }
+
+        if needs_line_break(self, context) {
+            return OptionalParentheses::Multiline;
```
MichaReiser: I think we can even return `Always` here because we know it breaks over multiple lines and will need parentheses

ntBre: Ah right, thanks! And thanks for all of your help here!

PR #21502, opened 2025-11-17
[ty] Add hyperlinks to rule codes in CLI

MichaReiser:
## Summary

This PR adds hyperlink rendering to our diagnostics so that users can jump to the rule documentation by clicking on the diagnostic id.

My first implementation added an optional `url_resolver: &dyn Fn(&DiagnosticId) -> Option<String>` argument to the full and concise rendering. The main idea was to avoid constructing the URLs unnecessarily when the output format doesn't need the URL. 

However, I then realised that the LSP always uses the URL and we have other output formats (ruff only for now) that construct the URL too, namely the JSON and rdjson output formats. The overhead also seems negligible compared to that we always construct the full diagnostic, even when only using the `concise` output format. That's why I opted to store the URL as an optional field on the diagnostic instead. Not only did it simplify the implementation, it also reduced some code duplication.

The main advantage of this approach is that it requires some amount of code duplication to set the URL on the diagnostic. 

It should be very easy to leverage this functionality in Ruff too (@ntBre if you're interested).

## Test Plan
https://github.com/user-attachments/assets/9855ca70-f92d-4299-b847-360be8ea1f89

### DIFF
--- a/crates/ruff_annotate_snippets/src/renderer/display_list.rs
+++ b/crates/ruff_annotate_snippets/src/renderer/display_list.rs
@@ -261,13 +261,16 @@ impl DisplaySet<'_> {
                 if hide_severity {
                     buffer.append(
                         line_offset,
-                        &format!("{id} ", id = DisplayId(&id)),
+                        &format!("{id} ", id = fmt_with_hyperlink(id.id, id.url, stylesheet)),
                         *stylesheet.error(),
                     );
                 } else {
                     buffer.append(
                         line_offset,
-                        &format!("{annotation_type}[{id}]", id = DisplayId(&id)),
+                        &format!(
+                            "{annotation_type}[{id}]",
+                            id = fmt_with_hyperlink(id.id, id.url, stylesheet)
+                        ),
                         *color,
                     );
                 }
@@ -735,7 +738,10 @@ impl DisplaySet<'_> {
                             } else if formatted_len != 0 {
                                 formatted_len += 2;
                                 let id = match &annotation.annotation.id {
-                                    Some(id) => format!("[{id}]", id = DisplayId(&id)),
+                                    Some(id) => format!(
+                                        "[{id}]",
+                                        id = fmt_with_hyperlink(&id.id, id.url, stylesheet)
+                                    ),
                                     None => String::new(),
                                 };
                                 buffer.puts(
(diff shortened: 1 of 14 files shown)


crates/ruff_annotate_snippets/src/snippet.rs
```diff
@@ -12,13 +12,19 @@
 
 use std::ops::Range;
 
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+pub(crate) struct Id<'a> {
```
MichaReiser, line 16: This type is inspired by the new annotate snippet that also has an `Id` type

AlexWaygood:
Interesting that mypy_primer timed out but ecosystem-analyzer didn't find any changes to report (primer timeouts usually occur because it took too long trying to compute the diff between the two runs)

MichaReiser:
mypy primer timing out makes sense, given that it forces color output (every diagnostic has now a different output)

MichaReiser:
There's an argument that it's hard to notice that the id is clickable. We could decide to underline the id or add the URL as a note (similar to what we do with the rule status). This would have the benefit that we could render the URL as a string and it's up to the CLI to add the hyperlink. The downside of this is that it only works for the `full` output and not concise.

I'm also not a 100% sure if we need to restrict the hyperlink rendering to more cases, e.g. by using https://docs.rs/supports-hyperlinks/latest/src/supports_hyperlinks/lib.rs.html#12-53

AlexWaygood:
CMD+clicking on the rule code works on this branch for me in a terminal inside VSCode, but not in the Terminal app on my macbook -- any idea why? (Is this just a case of "Alex, start using iTerm2 or something instead of the default Terminal app that your macbook came with"?)

Adding the full URL as a note might resolve this, since I'm definitely able to CMD+click on full URL in a terminal (I do this all the time to open local playground deploys in a browser straight from my terminal, obviously 😄)

MichaReiser:
> CMD+clicking on the rule code works on this branch for me in a terminal inside VSCode, but not in the Terminal app on my macbook -- any idea why? (Is this just a case of "Alex, start using iTerm2 or something instead of the default Terminal app that your macbook came with"?)

I couldn't find a reference but yes, this is the behavior of terminals supporting the ANSI escape but not supporting the hyperlink feature.

MichaReiser:
I actually like the current behavior where Ghostty only renders the underline when I hit Cmd and hover the id. This makes it behave the same as other clickable links

PR #21505, opened 2025-11-17
[ty] Support inlay hint type hint location

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/1078.

[Screencast_20251117_212227.webm](https://github.com/user-attachments/assets/eb1249fe-83ad-4d2a-aed8-2ab33eb0fbcd).

To find valid identifiers in the display string we do the following.
1. iterate over each char
2. start with max length string (current char to end of string)
3. reduce string length until valid identifier is found
4. If no valid identifier is found, we append the character to the label parts.
5. if we do find one we check all global symbols (this should be sped up) to find the symbol we want.
6. Add inlay label part and navigation target.

## Test Plan

Our current snapshots for inlay hint locations are very hard to understand and frankly arent useful at all.

### DIFF
--- a/crates/ty_server/tests/e2e/inlay_hints.rs
+++ b/crates/ty_server/tests/e2e/inlay_hints.rs
@@ -47,7 +47,20 @@ y = foo(1)
             "value": ": "
           },
           {
-            "value": "int"
+            "value": "int",
+            "location": {
+              "uri": "file://ty/vendored/typeshed/f8cdc0bd526301e873cd952eb0d457bdf2554e57/stdlib/builtins.pyi",
+              "range": {
+                "start": {
+                  "line": 347,
+                  "character": 6
+                },
+                "end": {
+                  "line": 347,
+                  "character": 9
+                }
+              }
+            }
           }
         ],
         "kind": 1
--- a/crates/ty_server/tests/e2e/main.rs
+++ b/crates/ty_server/tests/e2e/main.rs
@@ -1198,6 +1198,7 @@ impl TestContext {
             r#"The system cannot find the file specified."#,
             "No such file or directory",
         );
+        settings.add_filter(r"file://[^\s]*/\.cache/", "file://");
 
         let settings_scope = settings.bind_to_scope();
 


MatthewMckee4:
There are a lot of false positives (random targets) here.

- In "Todo" annotations
- In literal string annotations.

I think this is a good test but probably not a good solution. 

Interested to see people's thoughts.

MichaReiser:
Thanks to work on this. 

It feels wrong to me to render the type to a string, and then trying to reparse the information. I think we should either:

* Accept the fact that the LSP needs more flexibility and use a separate rendering pipeline all together
* Use a visitor approach where we don't write to a `std::fmt::Formatter` but to our own formatter instead, which is a `trait` with `visit_type(&mut self, ty)`, `write_str` etc. methods. I'd have to play with the problem myself to get a good sense for how an ergonomic API for this would look like.

We do have a similar problem for showing go to targets on hover. It's slightly different than inlay but we would want to collect all types, resolve their definitions, and render the definitions in a separate list. That's why I think a visitor approach where one implementation writes to a normal formatter, another that creates the inlay hint parts, and a third that writes to a normal formatter and collects the go to definition targets seems best?

But curious to hear what @Gankra thinks. She worked on this code more than I

AlexWaygood:
> * Use a visitor approach where we don't write to a `std::fmt::Formatter` but to our own formatter instead, which is a `trait` with `visit_type(&mut self, ty)`, `write_str` etc. methods. I'd have to play with the problem myself to get a good sense for how an ergonomic API for this would look like.

For reference, we have a `TypeVisitor` trait at https://github.com/astral-sh/ruff/blob/d5a95ec8246aae0656c3be6b5b23741e84fa8a86/crates/ty_python_semantic/src/types/visitor.rs#L32, but I'm not totally sure from your description whether it would work for your purposes here

MichaReiser:
I think we would want something more specific to rendering types. I very much doubt that we need separate visitor methods for each type. We're more interested in getting "hook-in-points" to know when we enter a new type and when we exit it, as well as having a way to write the label for the current type.

MatthewMckee4:
> It feels wrong to me to render the type to a string

I agree

It also feels wrong to duplicate display logic from `display.rs`.

Could we make changes in `display.rs`? Possibly making some intermediate struct that collects all the parts of the string, importantly parts that have a definition (or just `TextRange`) that we can goto. Then implement `Display` for that.

Meaning we can get all the information we need from the intermediate struct.

MichaReiser:
> Could we make changes in display.rs? Possibly making some intermediate struct that collects all the parts of the string, importantly parts that have a definition (or just TextRange) that we can goto. Then implement Display for that.

I think this would work but has two downsides:

* We collect a lot of unnecessary information for the common case where we just want to render a type to a string
* It's not enough for go to in hover (https://github.com/astral-sh/ty/issues/1575), requiring another refactor.

The approach I would try is to define a new trait:

```rust
pub trait DisplayTypeFormatter<'db> {
	fn enter_type(&mut self, ty: Type<'db>) {}
	fn exit_type(&mut self) {}
	fn write_str(&mut self, part: &str);
}
```

I think we can even implement `std::fmt::Write` for `DisplayVisitor` so that we can keep using `write!(f, "text{interpolation}");` by either implementing it for `&dyn DisplayTypeFormatter` or generic over `T`

```rust
impl std::fmt::Write for dyn DisplayTypeFormatter<'_> {
	fn write_str(&mut self, text: &str) -> std::fmt::Result<()> {
		DisplayFormatter::write_str(self, text);
		Ok(())
	}
}
```

The implementation to formatting the type to a string is:

```rust
struct DisplayToString(String);

impl DisplayTypeFormatter<'_> for DisplayToString {
	fn write_str(&mut self, part: &str) {
		self.0.push_str(part);
	}
}
```

Extracting the inlay is a bit tricky. It will require maintaining a stack of inlay parts where you push to the stack when entering a new type and popping from the stack when exiting (the stack is your scratch pad and you move them to a finalized state once finished.


The hover implementation wraps a `DisplayTypeFormatter` and also keeps a `HashSet` of all seen `Definition`.

Note: I haven't tried any of this, so this might not work or require more methods or different signatures but I think this is worth exploring.

Gankra:
Oh how serendipitous, I was just thinking about implementing this. We already have the visitor pattern in our display code, it's used by signature_help. I would absolutely implement this feature by expanding upon this design:

https://github.com/astral-sh/ruff/blob/192c37d5401786b3043aafbfb657b4f5e7df8ac8/crates/ty_python_semantic/src/types/display.rs#L1099-L1109

https://github.com/astral-sh/ruff/blob/192c37d5401786b3043aafbfb657b4f5e7df8ac8/crates/ty_python_semantic/src/types/display.rs#L1155-L1160

MatthewMckee4:
Cool, I'll close this for now and let you do your thing, I'm happy to link it up to the inlay hints if you don't get to that.

I will also think about a better way to snapshot the targets.

Gankra:
@MatthewMckee4 done in #21533, thanks for getting the ball rolling!

PR #21506, opened 2025-11-18
[ty] Typevars should be considered equivalent even after lazy defaults/bounds/etc are forced

dcreager:
This was a lingering place where we were still using `TypeVarInstance` equality to check whether two typevars are the same. We need to fall back on comparing the typevar `identity`, since that remains the same even after evaluating lazy defaults, bounds, or constraints.

This was showing up as a bug on #20933, since the new constraint solver normalizes types as part of constructing constraint sets. That had the effect of making a `TypeVar` instance fail the argument/parameter assignability check after we apply the inferred specialization in something like

```py
def f[T = int]():
    reveal_type(T)
```

since the argument type has a lazy default, but the normalized/specialized parameter type ends up with the evaluated/eager default.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -2484,7 +2484,7 @@ impl<'db> Type<'db> {
                 disjointness_visitor,
             ),
 
-            (Type::KnownInstance(left), right) => left.instance_fallback(db).has_relation_to_impl(
+            (Type::KnownInstance(left), right) => left.has_relation_to_impl(
                 db,
                 right,
                 inferable,
@@ -2626,6 +2626,10 @@ impl<'db> Type<'db> {
 
             (Type::NominalInstance(first), Type::NominalInstance(second)) => {
                 first.is_equivalent_to_impl(db, second, inferable, visitor)
+            }
+
+            (Type::KnownInstance(first), Type::KnownInstance(second)) => {
+                ConstraintSet::from(first.is_equivalent_to(db, second))
             }
 
             (Type::Union(first), Type::Union(second)) => {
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types.rs
```diff
@@ -8121,6 +8125,78 @@ impl<'db> KnownInstanceType<'db> {
...
+                | Self::Field(_)
+                | Self::ConstraintSet(_)
+                | Self::UnionType(_)
+                | Self::Literal(_)
+                | Self::Annotated(_)
+                | Self::TypeGenericAlias(_)
+                | Self::NewType(_),
+            ) => self == other,
+        }
+    }
```
AlexWaygood, line 8160: IIUC, all of these `KnownInstance` type variants essentially represent the type of a parameterized special form in a value position. so for `T = TypeVar("T")`, we represent the type of `T` as being a `Type::KnownInstanceType(KnownInstanceType::TypeVar(..))`. Given this, why is it important to check two typevars for equivalence using `.is_same_typevar_as()`, but _not_ do the same for other special forms in value positions (which might be parameterized with typevars). E.g. when comparing whether `T | int` is equivalent to `T | int`, where `T` is a type variable -- here `T | int` will be represented in our model as a `Type::KnownInstanceType(KnownInstanceType::UnionType(..))`, where one of the elements in that union is a `Type::KnownInstanceType(KnownInstanceType::TypeVar(..))`: ISTM we need to account for the possibility of `Type::KnownInstanceType(KnownInstanceType::TypeVar(..))`s being in nested positions here, which will require recursing into the types rather than just using `==` checks

dcreager: Hmm yes I think you're right!

carljm:
Setting this back to draft given @AlexWaygood comments, to clear up our in-review PR queue.

PR #21508, opened 2025-11-18
[ty] Exit with `2` if there's any IO error

MichaReiser:
## Summary

Fixes https://github.com/astral-sh/ty/issues/354

Exit with a code of `2` if there's any IO error (because ty failed to read a file).

## Test Plan

Updated test

### DIFF
--- a/.github/workflows/ty-ecosystem-analyzer.yaml
+++ b/.github/workflows/ty-ecosystem-analyzer.yaml
@@ -67,7 +67,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@0aff03414da5d242e97a9f43fb502e085637a4a1"
+          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@e5c5f5b2d762af91b28490537fe0077334165693"
 
           ecosystem-analyzer \
             --repository ruff \
--- a/.github/workflows/ty-ecosystem-report.yaml
+++ b/.github/workflows/ty-ecosystem-report.yaml
@@ -52,7 +52,7 @@ jobs:
 
           cd ..
 
-          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@0aff03414da5d242e97a9f43fb502e085637a4a1"
+          uv tool install "git+https://github.com/astral-sh/ecosystem-analyzer@e5c5f5b2d762af91b28490537fe0077334165693"
 
           ecosystem-analyzer \
             --verbose \


MichaReiser:
Hmm, the mypy primer error is annoying

```
Pythonwin/pywin/test/_dbgscript.py: error[io] Failed to read file: stream did not contain valid UTF-8
```

The file is encoded in latin-1 and contains Umlaute
https://github.com/mhammond/pywin32/blob/main/Pythonwin/pywin/test/_dbgscript.py

I believe this is the desired behavior. But we now need a way to exclude this file 😟

.github/mypy-primer-ty.toml
```diff
@@ -6,3 +6,8 @@
 possibly-unresolved-reference = "warn"
 unused-ignore-comment = "warn"
 division-by-zero = "warn"
+
+[src]
+exclude = [
+    "Pythonwin/pywin/test/_dbgscript.py"  # not valid UTF-8
+]
```
MichaReiser: I think this won't work when running mypy primer locally unless someone uses the `mypy_primer.sh` script.

The alternative is to move `pywin` to `bad.txt`.

MichaReiser: Another alternative is to not use an IO error for non-UTF-8 files.

.github/mypy-primer-ty.toml
```diff
@@ -6,3 +6,11 @@
 possibly-unresolved-reference = "warn"
 unused-ignore-comment = "warn"
 division-by-zero = "warn"
+
+[src]
+exclude = [
+    # not valid UTF-8
+    "Pythonwin/pywin/test/_dbgscript.py",
+    "tests/appsec/iast/fixtures/aspects/str/non_utf8_content.py",
+    "tests/appsec/iast/fixtures/ast/str/non_utf8_content.py",
```
sharkdp: I think I would prefer that we patch those in mypy_primer directly. We could pass a `ty_cmd="{ty} check --exclude=… {paths}` option for the corresponding [projects](https://github.com/hauntsaninja/mypy_primer/blob/master/mypy_primer/projects.py).

This way, it would also work locally, and there would be no cross-project pollution of the settings here.

AlexWaygood: We could also just change the primer setup upstream so that primer ignores an exit code 2 from ty without aborting the whole script. It would be just another diagnostic in the diff (like it is today)

MichaReiser: Done https://github.com/hauntsaninja/mypy_primer/pull/221

sharkdp:
Running ecosystem-analyzer, just to make sure that it also picks up the mypy_primer changes.

sharkdp:
Ok, it does not automatically pick it up. I need to update the lockfile.

PR #21467, opened 2025-11-15
[ty] implement `TypedDict` structural assignment

oconnor663:
Closes https://github.com/astral-sh/ty/issues/1387.

PR #21512, opened 2025-11-18
[ty] Add Salsa caching to `ClassLiteral::fields`

AlexWaygood:
## Summary

Locally, applying this patch to @oconnor663's branch for https://github.com/astral-sh/ruff/pull/21467 speeds us up >120x on the pydantic benchmark (we go from taking a whole minute to complete the benchmark to taking <0.5s).

But I think this might provide a performance boost on `main`, which is why I'm filing it as a standalone PR.

## Test Plan

All existing tests pass

### DIFF
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -7,7 +7,6 @@ use super::{
     SpecialFormType, SubclassOfType, Truthiness, Type, TypeQualifiers, class_base::ClassBase,
     function::FunctionType, infer_expression_type, infer_unpack_types,
 };
-use crate::FxOrderMap;
 use crate::module_resolver::KnownModule;
 use crate::place::TypeOrigin;
 use crate::semantic_index::definition::{Definition, DefinitionState};
@@ -128,7 +127,7 @@ fn try_metaclass_cycle_initial<'db>(
 }
 
 /// A category of classes with code generation capabilities (with synthesized methods).
-#[derive(Clone, Copy, Debug, PartialEq, salsa::Update, get_size2::GetSize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, salsa::Update, get_size2::GetSize)]
 pub(crate) enum CodeGeneratorKind<'db> {
     /// Classes decorated with `@dataclass` or similar dataclass-like decorators
     DataclassLike(Option<DataclassTransformerParams<'db>>),
@@ -1253,7 +1252,7 @@ impl MethodDecorator {
 }
 
 /// Kind-specific metadata for different types of fields
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, salsa::Update, get_size2::GetSize)]
 pub(crate) enum FieldKind<'db> {
     /// `NamedTuple` field metadata
     NamedTuple { default_ty: Option<Type<'db>> },
@@ -1281,7 +1280,7 @@ pub(crate) enum FieldKind<'db> {
 }
 
 /// Metadata regarding a dataclass field/attribute or a `TypedDict` "item" / key-value pair.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, salsa::Update, get_size2::GetSize)]
 pub(crate) struct Field<'db> {
     /// The declared type of the field
     pub(crate) declared_ty: Type<'db>,
@@ -2329,7 +2328,8 @@ impl<'db> ClassLiteral<'db> {
                     || kw_only.unwrap_or(has_dataclass_param(DataclassFlags::KW_ONLY));
 
                 // Use the alias name if provided, otherwise use the field name
-                let parameter_name = alias.map(Name::new).unwrap_or(field_name);
+                let parameter_name =
+                    Name::new(alias.map(|alias| &**alias).unwrap_or(&**field_name));
 
                 let mut parameter = if is_kw_only {
                     Parameter::keyword_only(parameter_name)
@@ -2595,7 +2595,7 @@ impl<'db> ClassLiteral<'db> {
             (CodeGeneratorKind::TypedDict, "get") => {
                 let overloads = self
                     .fields(db, specialization, field_policy)
-                    .into_iter()
+                    .iter()
                     .flat_map(|(name, field)| {
                         let key_type =
                             Type::StringLiteral(StringLiteralType::new(db, name.as_str()));
@@ -2824,12 +2824,13 @@ impl<'db> ClassLiteral<'db> {
     /// Returns a list of all annotated attributes defined in this class, or any of its superclasses.
     ///
     /// See [`ClassLiteral::own_fields`] for more details.
+    #[salsa::tracked(returns(ref), heap_size=get_size2::GetSize::get_heap_size)]
     pub(crate) fn fields(
         self,
         db: &'db dyn Db,
         specialization: Option<Specialization<'db>>,
-        field_policy: CodeGeneratorKind,
-    ) -> FxOrderMap<Name, Field<'db>> {
+        field_policy: CodeGeneratorKind<'db>,
+    ) -> FxIndexMap<Name, Field<'db>> {
         if field_policy == CodeGeneratorKind::NamedTuple {
             // NamedTuples do not allow multiple inheritance, so it is sufficient to enumerate the
             // fields of this class only.
@@ -2877,8 +2878,8 @@ impl<'db> ClassLiteral<'db> {
         db: &'db dyn Db,
         specialization: Option<Specialization<'db>>,
         field_policy: CodeGeneratorKind,
-    ) -> FxOrderMap<Name, Field<'db>> {
-        let mut attributes = FxOrderMap::default();
+    ) -> FxIndexMap<Name, Field<'db>> {
+        let mut attributes = FxIndexMap::default();
 
         let class_body_scope = self.body_scope(db);
         let table = place_table(db, class_body_scope);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1256,6 +1256,7 @@ dependencies = [
  "compact_str",
  "get-size-derive2",
  "hashbrown 0.16.0",
+ "indexmap",
  "smallvec",
 ]
 
--- a/crates/ty/docs/rules.md
+++ b/crates/ty/docs/rules.md
@@ -39,7 +39,7 @@ def test(): -> "int":
 Default level: <a href="../rules.md#rule-levels" title="This lint has a default level of 'error'."><code>error</code></a> ·
 Added in <a href="https://github.com/astral-sh/ty/releases/tag/0.0.1-alpha.1">0.0.1-alpha.1</a> ·
 <a href="https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20call-non-callable" target="_blank">Related issues</a> ·
-<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L123" target="_blank">View source</a>
+<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L121" target="_blank">View source</a>
 </small>
 
 
@@ -63,7 +63,7 @@ Calling a non-callable object will raise a `TypeError` at runtime.
 Default level: <a href="../rules.md#rule-levels" title="This lint has a default level of 'error'."><code>error</code></a> ·
 Added in <a href="https://github.com/astral-sh/ty/releases/tag/0.0.1-alpha.1">0.0.1-alpha.1</a> ·
 <a href="https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20conflicting-argument-forms" target="_blank">Related issues</a> ·
-<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L167" target="_blank">View source</a>
+<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Ftypes%2Fdiagnostic.rs#L165" target="_blank">View source</a>
 </small>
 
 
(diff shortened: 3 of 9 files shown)


crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2824,12 +2824,13 @@ impl<'db> ClassLiteral<'db> {
     /// Returns a list of all annotated attributes defined in this class, or any of its superclasses.
     ///
     /// See [`ClassLiteral::own_fields`] for more details.
+    #[salsa::tracked(returns(ref))]
```
MichaReiser: Seems reasonable to cache this. Can you add the `heap_size` attribute so that we can get a sense of how expensive this is in terms of memory consumption?

How much does it change performance if we were to cache `own_fields` only? I'm asking because that would require slightly less memory

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -2824,12 +2824,13 @@ impl<'db> ClassLiteral<'db> {
...
     /// See [`ClassLiteral::own_fields`] for more details.
+    #[salsa::tracked(returns(ref), heap_size=ruff_memory_usage::index_map_heap_size)]
     pub(crate) fn fields(
         self,
         db: &'db dyn Db,
         specialization: Option<Specialization<'db>>,
-        field_policy: CodeGeneratorKind,
-    ) -> FxOrderMap<Name, Field<'db>> {
+        field_policy: CodeGeneratorKind<'db>,
+    ) -> FxIndexMap<Name, Field<'db>> {
```
MichaReiser, line 2833: What's the reason why we need an `IndexMap` now?

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: > Can you add the `heap_size` attribute so that we can get a sense of how expensive this is in terms of memory consumption?

Yes, I was already working on this before you posted your comment 😆

> How much does it change performance if we were to cache `own_fields` only? I'm asking because that would require slightly less memory

Not sure. Happy to try it, but I'll wait for Codspeed to finish on this version of the PR first so we can compare the two using the codspeed runners

crates/ty_python_semantic/src/types/class.rs
AlexWaygood: `salsa::Update` is implemented for `IndexMap` but not for `OrderMap`.

AlexWaygood: I can make a PR to Salsa to implement `salsa::Update` for `OrderMap` too if you'd prefer this method to continue returning `OrderMap`. But as far as I understand, we generally prefer to use `IndexMap`s where possible since `IndexMap` is a simpler type than `OrderMap` (`OrderMap` just adds `Hash` on top of `IndexMap`'s API, IIRC?). It doesn't seem necessary to use an `OrderMap` here.

MichaReiser: Oh I see. Nah, I don't care it just wasn't obvious to me why we change it

AlexWaygood:
[1-3% improvements on six benchmarks](https://codspeed.io/astral-sh/ruff/branches/alex%2Foptimize-fields?utm_source=github&utm_medium=check&utm_content=details), including a 2% improvement on pydantic, offset by a 1% regression on 9 benchmarks. No memory usage change reported by mypy_primer.

MichaReiser:
Should we rebase this on top of Jack's PR to see the real performance benefits? It seems more questionable to merge with the now very limited performance improvement

carljm:
I think the options are "land this first" or "merge it with Jack's PR" -- I don't think we can land the TypedDict PR separately without this, because it makes benchmarks time out on GitHub.

I would be inclined to land this separately just because it's easier to review as two separate PRs. But we can rebase Jack's PR on top of this first to verify that this fixes the performance issues there.

AlexWaygood:
> Should we rebase this on top of Jack's PR to see the real performance benefits? It seems more questionable to merge with the now very limited performance improvement

Hmm, I'm not _totally_ sure I understand the standard being applied here -- I feel like we've merged lots of performance improvements in the past that showed <3% improvements on our benchmarks, especially if they didn't have any measurable memory-usage regressions :-)

As I say, I've also verified locally that this yields a _huge_ speedup when I cherry pick it onto #21467, which currently can't be merged due to the performance regression on that branch.

But sure, I'm happy if @oconnor663 wants to just cherry-pick https://github.com/astral-sh/ruff/pull/21512/commits/f4075371e05fe7d48a94e1c841b37e9a6f8e3dc3 onto #21467.

> How much does it change performance if we were to cache `own_fields` only? I'm asking because that would require slightly less memory

I tried this out locally (applied to Jack's PR). It has very competitive performance to this version. But I'm inclined to go with this version, since there's no memory regression reported on this PR, and since caching `ClassLiteral::own_fields` rather than `ClassLiteral::fields` makes our APIs a fair bit less ergonomic IMO. You end up always having to assign the result of `class.fields(db)` before iterating over it, or the borrow checker complains in several places that temporary values are being dropped while borrowed, because the return type of `ClassLiteral::fields()` needs to be `FxIndexMap<&'db Name, &'db Field<'db>>` rather than `&'db FxIndexMap<Name, Field<'db>>`

<details>

```diff
diff --git a/crates/ty_python_semantic/src/types/class.rs b/crates/ty_python_semantic/src/types/class.rs
index abc54f5626..54cfd75f9c 100644
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -2593,8 +2593,9 @@ impl<'db> ClassLiteral<'db> {
                 )))
             }
             (CodeGeneratorKind::TypedDict, "get") => {
-                let overloads = self
-                    .fields(db, specialization, field_policy)
+                let fields = self.fields(db, specialization, field_policy);
+
+                let overloads = fields
                     .iter()
                     .flat_map(|(name, field)| {
                         let key_type =
@@ -2824,17 +2825,19 @@ impl<'db> ClassLiteral<'db> {
     /// Returns a list of all annotated attributes defined in this class, or any of its superclasses.
     ///
     /// See [`ClassLiteral::own_fields`] for more details.
-    #[salsa::tracked(returns(ref))]
     pub(crate) fn fields(
         self,
         db: &'db dyn Db,
         specialization: Option<Specialization<'db>>,
         field_policy: CodeGeneratorKind<'db>,
-    ) -> FxIndexMap<Name, Field<'db>> {
+    ) -> FxIndexMap<&'db Name, &'db Field<'db>> {
         if field_policy == CodeGeneratorKind::NamedTuple {
             // NamedTuples do not allow multiple inheritance, so it is sufficient to enumerate the
             // fields of this class only.
-            return self.own_fields(db, specialization, field_policy);
+            return self
+                .own_fields(db, specialization, field_policy)
+                .iter()
+                .collect();
         }
 
         let matching_classes_in_mro: Vec<_> = self
@@ -2873,11 +2876,12 @@ impl<'db> ClassLiteral<'db> {
     ///     y: str = "a"
     /// ```
     /// we return a map `{"x": (int, None), "y": (str, Some(Literal["a"]))}`.
+    #[salsa::tracked(returns(ref))]
     pub(super) fn own_fields(
         self,
         db: &'db dyn Db,
         specialization: Option<Specialization<'db>>,
-        field_policy: CodeGeneratorKind,
+        field_policy: CodeGeneratorKind<'db>,
     ) -> FxIndexMap<Name, Field<'db>> {
         let mut attributes = FxIndexMap::default();
 
diff --git a/crates/ty_python_semantic/src/types/diagnostic.rs b/crates/ty_python_semantic/src/types/diagnostic.rs
index 66316f1537..5b3cc2b62e 100644
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -3135,7 +3135,7 @@ pub(crate) fn report_invalid_key_on_typed_dict<'db>(
     typed_dict_ty: Type<'db>,
     full_object_ty: Option<Type<'db>>,
     key_ty: Type<'db>,
-    items: &FxIndexMap<Name, Field<'db>>,
+    items: &FxIndexMap<&'db Name, &'db Field<'db>>,
 ) {
     let db = context.db();
     if let Some(builder) = context.report_lint(&INVALID_KEY, key_node) {
@@ -3197,8 +3197,8 @@ pub(crate) fn report_invalid_key_on_typed_dict<'db>(
 pub(super) fn report_namedtuple_field_without_default_after_field_with_default<'db>(
     context: &InferContext<'db, '_>,
     class: ClassLiteral<'db>,
-    (field, field_def): &(Name, Option<Definition<'db>>),
-    (field_with_default, field_with_default_def): &(Name, Option<Definition<'db>>),
+    (field, field_def): (&Name, Option<Definition<'db>>),
+    (field_with_default, field_with_default_def): (&Name, Option<Definition<'db>>),
 ) {
     let db = context.db();
     let module = context.module();
diff --git a/crates/ty_python_semantic/src/types/infer/builder.rs b/crates/ty_python_semantic/src/types/infer/builder.rs
index caa925ad74..778fe849b2 100644
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -4,6 +4,7 @@ use itertools::{Either, Itertools};
 use ruff_db::diagnostic::{Annotation, DiagnosticId, Severity};
 use ruff_db::files::File;
 use ruff_db::parsed::ParsedModuleRef;
+use ruff_python_ast::name::Name;
 use ruff_python_ast::visitor::{Visitor, walk_expr};
 use ruff_python_ast::{
     self as ast, AnyNodeRef, ExprContext, HasNodeIndex, NodeIndex, PythonVersion,
@@ -613,12 +614,11 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                     ) {
                         field_with_default_encountered =
                             Some((field_name, field.single_declaration));
-                    } else if let Some(field_with_default) = field_with_default_encountered.as_ref()
-                    {
+                    } else if let Some(field_with_default) = field_with_default_encountered {
                         report_namedtuple_field_without_default_after_field_with_default(
                             &self.context,
                             class,
-                            &(field_name, field.single_declaration),
+                            (field_name, field.single_declaration),
                             field_with_default,
                         );
                     }
@@ -919,9 +919,9 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 CodeGeneratorKind::from_class(self.db(), class, None)
             {
                 let specialization = None;
+                let fields = class.fields(self.db(), specialization, field_policy);
 
-                let kw_only_sentinel_fields: Vec<_> = class
-                    .fields(self.db(), specialization, field_policy)
+                let kw_only_sentinel_fields: Vec<_> = fields
                     .iter()
                     .filter_map(|(name, field)| {
                         field.is_kw_only_sentinel(self.db()).then_some(name)
@@ -7269,7 +7269,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             }
 
             let value_ty = if let Some(Type::StringLiteral(key)) = key_ty
-                && let Some(field) = typed_dict_items.get(key.value(self.db()))
+                && let Some(field) = typed_dict_items.get(&Name::from(key.value(self.db())))
             {
                 self.infer_expression(&item.value, TypeContext::new(Some(field.declared_ty)))
             } else {
@@ -7954,7 +7954,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                     Type::TypedDict(typed_dict_ty),
                                     None,
                                     key_ty,
-                                    items,
+                                    &items,
                                 );
                                 // Return `Unknown` to prevent the overload system from generating its own error
                                 return Type::unknown();
@@ -11285,7 +11285,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                                 value_ty,
                                 None,
                                 slice_ty,
-                                typed_dict.items(db),
+                                &typed_dict.items(db),
                             );
                         } else {
                             if let Some(builder) =
diff --git a/crates/ty_python_semantic/src/types/typed_dict.rs b/crates/ty_python_semantic/src/types/typed_dict.rs
index 1b33d7111f..521329f2c9 100644
--- a/crates/ty_python_semantic/src/types/typed_dict.rs
+++ b/crates/ty_python_semantic/src/types/typed_dict.rs
@@ -58,7 +58,7 @@ impl<'db> TypedDictType<'db> {
         self.defining_class
     }
 
-    pub(crate) fn items(self, db: &'db dyn Db) -> &'db FxIndexMap<Name, Field<'db>> {
+    pub(crate) fn items(self, db: &'db dyn Db) -> FxIndexMap<&'db Name, &'db Field<'db>> {
         let (class_literal, specialization) = self.defining_class.class_literal(db);
         class_literal.fields(db, specialization, CodeGeneratorKind::TypedDict)
     }
@@ -318,7 +318,7 @@ pub(super) fn validate_typed_dict_key_assignment<'db, 'ast>(
     let items = typed_dict.items(db);
 
     // Check if key exists in `TypedDict`
-    let Some((_, item)) = items.iter().find(|(name, _)| *name == key) else {
+    let Some((_, item)) = items.iter().find(|(name, _)| **name == key) else {
         if emit_diagnostic {
             report_invalid_key_on_typed_dict(
                 context,
@@ -327,7 +327,7 @@ pub(super) fn validate_typed_dict_key_assignment<'db, 'ast>(
                 Type::TypedDict(typed_dict),
                 full_object_ty,
                 Type::string_literal(db, key),
-                items,
+                &items,
             );
         }
```

</details>

crates/ruff_memory_usage/src/lib.rs
```diff
@@ -26,3 +27,16 @@ pub fn order_map_heap_size<K: GetSize, V: GetSize, S>(map: &OrderMap<K, V, S>) -
             .map(|(k, v)| heap_size(k) + heap_size(v))
             .sum::<usize>()
 }
+
+/// An implementation of [`GetSize::get_heap_size`] for [`IndexMap`].
+pub fn index_map_heap_size<K: GetSize, V: GetSize, S>(map: &IndexMap<K, V, S>) -> usize {
+    (map.capacity() * (K::get_stack_size() + V::get_stack_size()))
+        + (map.iter())
+            .map(|(k, v)| heap_size(k) + heap_size(v))
+            .sum::<usize>()
```
carljm: Is there a reference you used to verify the actual memory use of an IndexMap that could be linked here?

carljm (approved):
Thank you!

crates/ruff_memory_usage/src/lib.rs
AlexWaygood: no, this is just the `OrderMap` implementation a few lines above directly copied and pasted. But (assuming the pre-existing `OrderMap` implementation is correct), I think that that should be correct. An `OrderMap` is [just a thin wrapper around an `IndexMap`](https://github.com/indexmap-rs/ordermap/blob/d04e267f4cd2533b6e83e98e083942becf29c778/src/map.rs#L102-L105)

ibraheemdev: get-size2 already has support for `IndexMap`, so I'm not sure why we need this. Maybe the `indexmap` feature is not enabled somewhere?

AlexWaygood: Great catch -- yes, that was it! Fixed.

oconnor663:
Thank you @AlexWaygood (and reviewers)!

PR #21514, opened 2025-11-18
Set the diagnostic URL for lint errors

ntBre:
Summary
--

This PR wires up the `Diagnostic::set_documentation_url` method from #21502 to Ruff's lint diagnostics. This enables the links for the full and concise output formats without any other changes.

I considered also including the URLs for the grouped and pylint output formats, but the grouped format is still in `ruff_linter` instead of `ruff_db`, so we'd have to export some additional functionality to wire it up with `fmt_with_hyperlink`; and the pylint format doesn't currently render with color, so I think it might actually be machine readable rather than human readable?

The other ouput formats (json, json-lines, junit, github, gitlab, rdjson, azure, sarif) seem more clearly not to need the links.

Test Plan
--

I guess you can't see my cursor or the browser opening, but it works for lint rules, which have links, and doesn't include a link for syntax errors, which don't have valid links.

out

### DIFF
--- a/crates/ruff_db/src/diagnostic/render.rs
+++ b/crates/ruff_db/src/diagnostic/render.rs
@@ -2881,6 +2881,12 @@ watermelon
             self.diag.help(message);
             self
         }
+
+        /// Set the documentation URL for the diagnostic.
+        pub(super) fn documentation_url(mut self, url: impl Into<String>) -> DiagnosticBuilder<'e> {
+            self.diag.set_documentation_url(Some(url.into()));
+            self
+        }
     }
 
     /// A helper builder for tersely populating a `SubDiagnostic`.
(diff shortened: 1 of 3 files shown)


crates/ruff_linter/src/message/mod.rs
```diff
@@ -125,6 +125,7 @@ where
     }
 
     diagnostic.set_secondary_code(SecondaryCode::new(rule.noqa_code().to_string()));
+    diagnostic.set_documentation_url(rule.url());
```
MichaReiser, line 128: Do we still need `Diagnostic::to_ruff_url` or can we replace all usages with `diagnostic.documentation_url()`?

ntBre: Oh right, yes we can remove it. I just had to set the URL in some of the rendering tests too.

PR #21515, opened 2025-11-18
[ty] Keep colorizing `mypy_primer` output

sharkdp:
## Summary

After an update to `mypy_primer`, we now need to set the environment variable ourselves.

### DIFF
--- a/.github/workflows/mypy_primer.yaml
+++ b/.github/workflows/mypy_primer.yaml
@@ -55,6 +55,7 @@ jobs:
       - name: Run mypy_primer
         env:
           PRIMER_SELECTOR: crates/ty_python_semantic/resources/primer/good.txt
+          CLICOLOR_FORCE: "1"
           DIFF_FILE: mypy_primer.diff
         run: |
           cd ruff
--- a/scripts/mypy_primer.sh
+++ b/scripts/mypy_primer.sh
@@ -20,7 +20,7 @@ cd ..
 echo "Project selector: ${PRIMER_SELECTOR}"
 # Allow the exit code to be 0 or 1, only fail for actual mypy_primer crashes/bugs
 uvx \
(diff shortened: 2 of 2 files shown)


AlexWaygood:
> Colorized output in the GitHub Actions terminal is still working

hmm, but it looks like the error code has now been stripped as well? 😄 It's now just `error[]` instead of `error[invalid-await]`

I don't mind too much if we just disable colour entirely for ty in mypy_primer. I do occasionally find colourised logs useful, but it's probably not worth spending a lot of time on right now :-)

AlexWaygood:
> hmm, but it looks like the error code has now been stripped as well?

oh, but I guess they're missing on the `main` branch too? E.g. <https://github.com/astral-sh/ruff/actions/runs/19479467091/job/55747619369?pr=21467>. I think that might have also regressed in https://github.com/astral-sh/ruff/pull/21502...

sharkdp:
> hmm, but it looks like the error code has now been stripped as well? 😄 It's now just `error[]` instead of `error[invalid-await]`

Oh, I didn't even notice in the screenshot. Well, that's not my fault. We show the unmodified output in the terminal. I guess the GitHub Actions ANSI parser can't handle hyperlinks?

> I don't mind too much if we just disable colour entirely for ty in mypy_primer. I do occasionally find colourised logs useful, but it's probably not worth spending a lot of time on right now :-)

Ok, I guess I'll just change it

MichaReiser (approved):
> Oh, I didn't even notice in the screenshot. Well, that's not my fault. We show the unmodified output in the terminal. I guess the GitHub Actions ANSI parser can't handle hyperlinks?

Hmm. Terminals should skip unsupported ansi escaped without stripping the content. 

We can also consider using a crate to dedect supported terminals so that our users can keep using color too. Happy to own that.

MichaReiser:
Sorry for breaking mypy primer, I just published https://github.com/astral-sh/ruff/pull/21519

sharkdp:
> I just published #21519

Thanks. No worries.

PR #21516, opened 2025-11-18
[ty] Only normalize constraint bounds for display

dcreager:
We were previously normalizing the upper and lower bounds of each constraint when constructing constraint sets. Like in #21463, this was for conflated reasons: It made constraint set displays nicer, since we wouldn't render multiple constraints with obviously equivalent bounds. (Think `T ≤ A & B` and `T ≤ B & A`) But it was also useful for correctness, since prior to #21463 we were (trying to) add the full transitive closure to a constraint set's BDD, and normalization gave a useful reduction in the number of nodes in a typical BDD.

Now that we don't store the transitive closure explicitly, that second reason is no longer relevant. Our sequent map can store that full transitive closure much more efficiently than the expanded BDD would have. This helps fix some false positives on #20933, where we're seeing some (incorrect, need to be fixed, but ideally not blocking this effort) assignability failures between a type and its normalization.

Normalization is still useful for display purposes, and so we do normalize the upper/lower bounds before building up our display representation of a constraint set BDD.


crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -2279,6 +2304,18 @@ impl<'db> SequentMap<'db> {
         left_constraint: ConstrainedTypeVar<'db>,
         right_constraint: ConstrainedTypeVar<'db>,
     ) {
+        // These might seem redundant with the intersection check below, since `a → b` means that
+        // `a ∧ b = a`. But we are not normalizing constraint bounds, and these clauses help us
+        // identity constraints that are identical besides e.g. ordering of union/intersection
```
sharkdp:
```suggestion
        // identify constraints that are identical besides e.g. ordering of union/intersection
```

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -2279,6 +2304,18 @@ impl<'db> SequentMap<'db> {
...
+        // elements. (For instance, these clauses would find `T ≤ τ₁ ∩ τ₂` and `T ≤ τ₂ ∩ τ₁`, and
+        // add sequents for `(T ≤ τ₁ ∩ τ₂) → (T ≤ τ₂ ∩ τ₁)` and vice versa.)
+        if left_constraint.implies(db, right_constraint) {
+            self.add_single_implication(left_constraint, right_constraint);
+        }
+        if right_constraint.implies(db, left_constraint) {
+            self.add_single_implication(right_constraint, left_constraint);
+        }
```
sharkdp, line 2320: The explanation above seems to imply that we could (should?) return early here? Because if `a → b` always implies `a ∧ b = a`, then the code below would otherwise add *two additional* `a → b` implications and one `a → a` implication, both of which seem unnecessary?

dcreager: There are other protections in place against both of those: The `single_implications` and `pair_implications` maps now have sets to store all of the consequents, so two copies of `a → b` will automatically get de-duped, no matter how they're created. And similarly, `add_single_implication` and `add_pair_implication` will skip any sequent of the form `a → a`, `a ∧ b → a`, or `a ∧ b → b`, since they are not useful.

PR #19778, opened 2025-08-06
Fix(#19757)/isc003 autofix produces incorrect code

mikeleppane:
### Problem

The ISC003 rule was flagging explicit string concatenation (`"a" + "b"`) in all contexts, but it should only apply when the concatenation is inside brackets (parentheses or braces) where implicit concatenation is possible.

### Solution

Added a new `is_inside_brackets()` helper function that:
- Locates the current statement containing the expression
- Searches for opening brackets (`(` or `{`) before the expression
- Searches for closing brackets (`)` or `}`) after the expression
- Only reports violations when both bracket types are present

### Example

Input
```python
# input
_ = "Line1\n" + (
        "Line2\n"
        + "Line3"
)
```
`ruff check --fix` =>

```python
_ = "Line1\n" + (
        "Line2\n"
         "Line3"
)
```

[ISSUE](https://github.com/astral-sh/ruff/issues/19757)

## Test Plan

```bash
cargo test
```

PR #21517, opened 2025-11-18
[`flake8-implicit-str-concat`] Avoid invalid fix generated by autofix (`ISC003`)

### LINKED ISSUE #19757: Autofix for ISC003 produces malformed code
### Summary

Autofix fix for ISC003 (Explicitly concatenated string should be implicitly concatenated) produces `TypeError: 'str' object is not callable` in the following situation.
```python
result = "Line1\n" + (
    "Line2\n"
    "Line3"
)
```
`ruff check --fix` produces:
```python
result = "Line1\n" (
    "Line2\n"
    "Line3"
)
```
Which gives a `TypeError: 'str' object is not callable` error obiously.

Link to playground: https://play.ruff.rs/a9d9ac04-ef00-4ddf-a3a6-9e61fa9f8840

### Version

0.12.7

prakhar1144:
## Summary

As reported in #19757:
While attempting ISC003 autofix for an expression with explicit string concatenation, with either operand being a string literal that wraps across multiple lines (in parentheses) - it resulted in generating a fix which caused runtime error.

Example:
```
_ = "abc" + (
    "def"
    "ghi"
)
```
was being auto-fixed to:
```
_ = "abc" (
    "def"
    "ghi"
)
```
which raised `TypeError: 'str' object is not callable`

This commit makes changes to just report diagnostic - no autofix in such cases.

Fixes #19757.

## Test Plan
Added example scenarios in `crates/ruff_linter/resources/test/fixtures/flake8_implicit_str_concat/ISC.py`.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
+++ b/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
@@ -1,12 +1,12 @@
 use ruff_macros::{ViolationMetadata, derive_message_formats};
+use ruff_python_ast::parenthesize::parenthesized_range;
 use ruff_python_ast::{self as ast, Expr, Operator};
 use ruff_python_trivia::is_python_whitespace;
 use ruff_source_file::LineRanges;
 use ruff_text_size::{Ranged, TextLen, TextRange, TextSize};
 
-use crate::AlwaysFixableViolation;
 use crate::checkers::ast::Checker;
-use crate::{Edit, Fix};
+use crate::{Edit, Fix, FixAvailability, Violation};
 
 /// ## What it does
 /// Checks for string literals that are explicitly concatenated (using the
@@ -36,14 +36,16 @@ use crate::{Edit, Fix};
 #[violation_metadata(stable_since = "v0.0.201")]
 pub(crate) struct ExplicitStringConcatenation;
 
-impl AlwaysFixableViolation for ExplicitStringConcatenation {
+impl Violation for ExplicitStringConcatenation {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
         "Explicitly concatenated string should be implicitly concatenated".to_string()
     }
 
-    fn fix_title(&self) -> String {
-        "Remove redundant '+' operator to implicitly concatenate".to_string()
+    fn fix_title(&self) -> Option<String> {
+        Some("Remove redundant '+' operator to implicitly concatenate".to_string())
     }
 }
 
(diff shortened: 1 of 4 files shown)


prakhar1144:
As per suggestion in https://github.com/astral-sh/ruff/pull/19778#discussion_r2257641741:
> I think we can probably just use this helper function to check if left and right are parenthesized and avoid the fix if either of them is

> We can also still report a diagnostic in this case, we just can't autofix it. Another option would be to try to move one of the expressions into the existing parentheses from the other expression, but that might be too involved. I'd be happy with just offering a diagnostic but no fix if one of them is parenthesized to avoid causing the runtime error.

crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
```diff
@@ -82,9 +84,25 @@ pub(crate) fn explicit(checker: &Checker, expr: &Expr) {
...
+                    parenthesized_range(
+                        expr.into(),
+                        bin_op.into(),
+                        checker.comment_ranges(),
+                        checker.source(),
+                    )
+                    .is_some()
+                };
+                // If either `left` or `right` is parenthesized, generating
+                // a fix would be too involved. Just report the diagnostic.
```
amyreese, line 100: Let's also make it clear that the existing autofix would result in broken code, and maybe include a link to the issue.

crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
```diff
@@ -82,9 +84,25 @@ pub(crate) fn explicit(checker: &Checker, expr: &Expr) {
...
+
+                let is_parenthesized = |expr: &Expr| {
+                    parenthesized_range(
+                        expr.into(),
+                        bin_op.into(),
+                        checker.comment_ranges(),
+                        checker.source(),
+                    )
+                    .is_some()
+                };
```
amyreese, line 98: I think it's preferred to have this as a separate function that takes appropriate arguments rather than a block.

I think it would also be better to have a function that takes both left and right sides at the same time, and then the code below could be something like:

```rust
if !is_either_side_parenthesized(left, right, ...) {
    diagnostic.set_fix(...);
}
```

amyreese (changes requested):

crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
prakhar1144: Extended the comment:

```diff
diff --git a/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs b/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
index 58f3db33bd..8edb7b46ad 100644
--- a/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
+++ b/crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
@@ -98,6 +98,8 @@ pub(crate) fn explicit(checker: &Checker, expr: &Expr) {
                 };
                 // If either `left` or `right` is parenthesized, generating
                 // a fix would be too involved. Just report the diagnostic.
+                // Currently, attempting `generate_fix` would result in
+                // an invalid code. See: #19757
                 if is_parenthesized(left) || is_parenthesized(right) {
                     return;
                 }
```

crates/ruff_linter/src/rules/flake8_implicit_str_concat/rules/explicit.rs
prakhar1144: > I think it would also be better to have a function that takes both left and right sides at the same time

That'd result in the new function body being:
```rs
  is_left_parenthesized = parenthesized_range(
      left.into(),
      bin_op.into(),
      checker.comment_ranges(),
      checker.source(),
  )
  .is_some();
  
   is_right_parenthesized = parenthesized_range(
      right.into(),
      bin_op.into(),
      checker.comment_ranges(),
      checker.source(),
  )
  .is_some();
  
  is_left_parenthesized || is_right_parenthesized
 ```
 
 which is a bit verbose... with repeated code ?
 
 I had thought about a separate function while writing this, but preferred this approach because:
 * It's only called twice in adjacent lines, localized scope, so closure felt natural
 * it doesn't have complex logic worth unit testing

prakhar1144:
This is ready to review.

amyreese (approved):

PR #21519, opened 2025-11-19
Only render hyperlinks for terminals known to support them

MichaReiser:
## Summary

GitHub action's renderer doesn't support hyperlink rendering and, unlike the terminals I tested,
it also isn't just ignoring the hyperlink escape sequence, instead it removes/hides the entire content between the Hyperlink escape sequence. 

This PR uses [`supports-hyperlinks`](https://docs.rs/supports-hyperlinks/3.1.0/supports_hyperlinks/) to only render hyperlinks 
on terminals that are known to support the escape sequence. 

The crate is used by cargo and miette. 

## Test Plan

Is mypy primer timing out proof enough that ty no longer renders the hyperlink ANSI escapes in CI 😅?

I tested that the hyperlinks are clickable in Ghostty

### DIFF
--- a/crates/ty_python_semantic/src/suppression.rs
+++ b/crates/ty_python_semantic/src/suppression.rs
@@ -225,7 +225,7 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
             SuppressionTarget::All => context.report_unchecked(
                 &UNUSED_IGNORE_COMMENT,
                 suppression.range,
-                format_args!("Unused `{}` directive", suppression.kind),
+                format_args!("Unused blanket `{}` directive", suppression.kind),
             ),
             SuppressionTarget::Lint(lint) => context.report_unchecked(
                 &UNUSED_IGNORE_COMMENT,


crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -729,7 +730,7 @@ print()
          ::: cell 2
         1 | # cell 2
           - import math
-        2 | 
+        2 |
```
MichaReiser: ahh, I hate that the trailing whitespace is needed here

crates/ruff_db/src/diagnostic/render/full.rs
```diff
@@ -703,52 +704,7 @@ print()
...
-        help: Remove assignment to unused variable `x`
-         ::: cell 3
-        1 | # cell 3
-        2 | def foo():
-        3 |     print()
-          -     x = 1
-        4 | 
-        note: This is an unsafe fix and may change runtime behavior
-        ");
+        insta::assert_snapshot!(env.render_diagnostics(&diagnostics));
```
MichaReiser, line 707: I made those tests "normal" snapshot tests because some snapshot lines contain trailing whitespace that Zed strips (might be a bug in Zed as I don't think an editor should strip trailing whitespace in a string!). 

I spent some time yesterday trying to fix the trailing whitespace but without success.

sharkdp (approved):
That library looks reasonable. And I don't mind if this buys us a few "false negatives".

crates/ruff_db/src/diagnostic/render/full.rs
BurntSushi: Yeah I support this. My editor (nvim) is configured to do the same.

PR #21521, opened 2025-11-19
Semantic tokens: consistently add the `DEFINITION` modifier

lucach:
## Summary

When computing semantic tokens, there are a few more statements potentially containing definitions than what is currently covered.

This commit reviews all the statements and completes the list of those requiring a custom visitor to add the `DEFINITION` modifier for the appropriate tokens.

Note: I've not added the `DEFINITION` modifier to the target of an augmented assignment. A new test makes that explicit. This is consistent with Pylance's behavior, although one could argue that given that *all* "regular" assignments are marked as definitions, perhaps also the augmented assignments should.

## Test Plan

Added new tests for each new statement with a custom visitor. When a test was already existing, I've reviewed and applied the diff.


MichaReiser:
Can you run the `ty_server` tests and update the semantic tokens snapshot

lucach:
Whoops, sorry. I even ran them, I just overlooked the output 🤦‍♂️

PR #21522, opened 2025-11-19
[ty] Add tests for generic implicit type aliases

sharkdp:
## Summary

Add a set of comprehensive tests for generic implicit type aliases to illustrate the current behavior with many flavors of `@Todo` types and false positive diagnostics.

The tests are partially based on the typing conformance suite, and the expected behavior has been checked against other type checkers.


AlexWaygood:
I guess one pathological case you could add would be something like

```py
from typing import TypeVar

T = TypeVar("T")
U = TypeVar("U")
S = TypeVar("S")

A = dict[T, U]
B = A[int, S]

def f(x: B[str]):
    reveal_type(x)  # revealed: dict[int, str]
```

sharkdp:
> I guess one pathological case

Doesn't seem too pathological. I'll add something like this.

PR #21524, opened 2025-11-19
[ty] Fix flaky tests on macos

dcreager:
We're seeing flaky test failures on macos, which seems to be caused by different Salsa ID orderings on the different platforms. Constraint set BDDs order their internal nodes based on the Salsa IDs of the interned typevar structs, and we had some code that depended on variable ordering in an unexpected way.

This patch definitely fixes the macos test failure on #21414, and hopefully fixes it on #21436, too.


MichaReiser:
I got really excited when I saw your PR title because I thought you fixed the flaky file watching tests... This is still great :)

PR #21530, opened 2025-11-19
[ty] Handle nested types when creating specializations from constraint sets

dcreager:
#21414 added the ability to create a specialization from a constraint set. It handled mutually constrained typevars just fine, e.g. given `T ≤ int ∧ U = T` we can infer `T = int, U = int`.

But it didn't handle _nested_ constraints correctly, e.g. `T ≤ int ∧ U = list[T]`. Now we do! This requires doing a fixed-point "apply the specialization to itself" step to propagate the assignments of any nested typevars, and then a cycle detection check to make sure we don't have an infinite expansion in the specialization.

This gets at an interesting nuance in our constraint set structure that @sharkdp has asked about before. Constraint sets are BDDs, and each internal node represents an _individual constraint_, of the form `lower ≤ T ≤ upper`. `lower` and `upper` are allowed to be other typevars, but only if they appear "later" in the arbitary ordering that we establish over typevars. The main purpose of this is to avoid infinite expansion for mutually constrained typevars.

However, that restriction doesn't help us here, because only applies when `lower` and `upper` _are_ typevars, not when they _contain_ typevars. That distinction is important, since it means the restriction does not affect our expressiveness: we can always rewrite `Never ≤ T ≤ U` (a constraint on `T`) into `T ≤ U ≤ object` (a constraint on `U`). The same is not true of `Never ≤ T ≤ list[U]` — there is no "inverse" of `list` that we could apply to both sides to transform this into a constraint on a bare `U`.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -254,27 +254,28 @@ impl<'db> ConstraintSet<'db> {
         }
 
         fn visit_dfs<'db>(
-            reachable_typevars: &FxHashMap<
+            reachable_typevars: &mut FxHashMap<
                 BoundTypeVarIdentity<'db>,
                 FxHashSet<BoundTypeVarIdentity<'db>>,
             >,
             discovered: &mut FxHashSet<BoundTypeVarIdentity<'db>>,
-            finished: &mut FxHashSet<BoundTypeVarIdentity<'db>>,
             bound_typevar: BoundTypeVarIdentity<'db>,
         ) -> bool {
             discovered.insert(bound_typevar);
-            for outgoing in reachable_typevars.get(&bound_typevar).into_iter().flatten() {
-                if discovered.contains(outgoing) {
+            let outgoing = reachable_typevars
+                .remove(&bound_typevar)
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -213,6 +215,104 @@ impl<'db> ConstraintSet<'db> {
...
+            discovered: &mut FxHashSet<BoundTypeVarIdentity<'db>>,
+            finished: &mut FxHashSet<BoundTypeVarIdentity<'db>>,
+            bound_typevar: BoundTypeVarIdentity<'db>,
+        ) -> bool {
+            discovered.insert(bound_typevar);
+            for outgoing in reachable_typevars.get(&bound_typevar).into_iter().flatten() {
+                if discovered.contains(outgoing) {
+                    return true;
+                }
+                if !finished.contains(outgoing) {
```
carljm: Could `finished` be eliminated in favor of removing keys from `reachable_typevars` when finished checking them?

dcreager: Hmm, yes! Have to jump through some borrow checker hoops but nothing too bad

PR #21531, opened 2025-11-19
[ty] Eagerly evaluate `types.UnionType` elements as type expressions

sharkdp:
## Summary

Eagerly evaluate the elements of a PEP 604 union in value position (e.g. `IntOrStr = int | str`) as type expressions and store the result (the corresponding `Type::Union` if all elements are valid type expressions, or the first encountered `InvalidTypeExpressionError`) on the `UnionTypeInstance`, such that the `Type::Union(…)` does not need to be recomputed every time the implicit type alias is used in a type annotation.

This might lead to performance improvements for large unions, but is also necessary for correctness, because the elements of the union might refer to type variables that need to be looked up in the scope of the type alias, not at the usage site.

## Test Plan

New Markdown tests

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
@@ -171,6 +171,31 @@ def _(x: int | str | None):
         reveal_type(x)  # revealed: int | None
     else:
         reveal_type(x)  # revealed: str
+
+ListStrOrInt = Union[list[str], int]
+
+def _(x: dict[int, str] | ListStrOrInt):
+    # TODO: this should ideally be an error
+    if isinstance(x, ListStrOrInt):
+        # TODO: this should not be narrowed
+        reveal_type(x)  # revealed: list[str] | int
+
+    # TODO: this should ideally be an error
+    if isinstance(x, Union[list[str], int]):
+        # TODO: this should not be narrowed
+        reveal_type(x)  # revealed: list[str] | int
+```
+
+## `Optional` as `classinfo`
+
+```py
+from typing import Optional
+
+def _(x: int | str | None):
+    if isinstance(x, Optional[int]):
+        reveal_type(x)  # revealed: int | None
+    else:
+        reveal_type(x)  # revealed: str
 ```
 
 ## `classinfo` is a `typing.py` special form
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -9395,6 +9395,10 @@ impl InferredAs {
 /// Contains information about a `types.UnionType` instance built from a PEP 604
 /// union or a legacy `typing.Union[…]` annotation in a value expression context,
 /// e.g. `IntOrStr = int | str` or `IntOrStr = Union[int, str]`.
+///
+/// # Ordering
+/// Ordering is based on the context's salsa-assigned id and not on its values.
+/// The id may change between runs, or when the context was garbage collected and recreated.
 #[salsa::interned(debug, heap_size=ruff_memory_usage::heap_size)]
 #[derive(PartialOrd, Ord)]
 pub struct UnionTypeInstance<'db> {


crates/ty_python_semantic/src/types.rs
```diff
@@ -6722,18 +6722,14 @@ impl<'db> Type<'db> {
...
                     fallback_type: Type::unknown(),
                 }),
-                KnownInstanceType::UnionType(list) => {
-                    let mut builder = UnionBuilder::new(db);
-                    let inferred_as = list.inferred_as(db);
-                    for element in list.elements(db) {
-                        builder = builder.add(if inferred_as.type_expression() {
-                            *element
-                        } else {
-                            element.in_type_expression(db, scope_id, typevar_binding_context)?
```
sharkdp, line 6748: The previous version was also only able to surface the first error. I'm not sure if this is a problem, or something that we can live with.

crates/ty_python_semantic/src/types.rs
```diff
@@ -6722,18 +6722,14 @@ impl<'db> Type<'db> {
...
-                        });
+                KnownInstanceType::UnionType(list) => match list {
+                    UnionTypeInstance::PEP604(instance) => {
+                        // Cloning here is cheap if the result is a `Type` (which is `Copy`). It's more
+                        // expensive if there are errors.
+                        instance.union_type(db).clone()
```
sharkdp: @carljm In the end, this didn't require a new salsa query. Instead, we're now eagerly building that `Type::Union` in case there were no errors, and interning the result. And then we only need to copy the result here when we see a use of this implicit type alias in a type expression.

This means that we do some extra work if someone defines `IntOrStr = int | str` and then never uses it as a type. But that feels okay to me?

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -212,21 +212,29 @@ impl ClassInfoConstraintFunction {
...
-                        if element.is_none(db) {
-                            self.generate_constraint(db, KnownClass::NoneType.to_class_literal(db))
-                        } else {
-                            self.generate_constraint(db, *element)
-                        }
-                    }),
-                )
+            Type::KnownInstance(KnownInstanceType::UnionType(instance)) => {
+                match instance {
+                    UnionTypeInstance::Legacy { .. } => None,
```
sharkdp: Will look into this tomorrow, but I guess a legacy `Union[int, str]` is not allowed in `isinstance`/`issubclass`. If so, we were definitely not handling it correctly before.

AlexWaygood: It actually _is_ allowed in isinstance/issubclass, but only on Python 3.10+. And I wouldn't worry too much about older versions than that since they're now end-of-life.

We should probably add some tests for this; it's an edge case but it has come up a couple of times on the mypy issue tracker.

crates/ty_python_semantic/src/types.rs
carljm: Yeah, that seems reasonable.

sharkdp: What's potentially more problematic is the fact that when we have a `LargeUnion = C1 | C2 | … | Cn`, we would build $n - 1$ union types eagerly, instead of just one. But if that turns out to be a problem, we can turn the `union_type` field into a salsa query.

crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
```diff
@@ -289,6 +288,23 @@ def _(flag: bool):
         reveal_type(x)  # revealed: Literal[1, "a"]
 ```
 
+## Generic aliases are not supported as second argument
+
+The `classinfo` argument cannot be a generic alias:
+
+```py
+def _(x: list[str] | list[int] | list[bytes]):
+    # TODO: Ideally, this would be an error
```
AlexWaygood:
```suggestion
    # TODO: Ideally, this would be an error (requires https://github.com/astral-sh/ty/issues/116)
```

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -191,13 +191,13 @@ def _(
     reveal_type(int_or_callable)  # revealed: int | ((str, /) -> bytes)
     reveal_type(callable_or_int)  # revealed: ((str, /) -> bytes) | int
     # TODO should be Unknown | int
-    reveal_type(type_var_or_int)  # revealed: T@_ | int
+    reveal_type(type_var_or_int)  # revealed: typing.TypeVar | int
```
sharkdp, line 194: We still don't support generic implicit type aliases, but this is a step in the right direction. `T@_` is just wrong.

crates/ty_python_semantic/src/types/narrow.rs
```diff
@@ -874,8 +874,6 @@ impl<'db, 'ast> NarrowingConstraintsBuilder<'db, 'ast> {
 
         let callable_ty = inference.expression_type(&*expr_call.func);
 
-        // TODO: add support for PEP 604 union types on the right hand side of `isinstance`
-        // and `issubclass`, for example `isinstance(x, str | (int | float))`.
```
sharkdp, line 878: This was recently added

crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
```diff
@@ -159,19 +159,18 @@ IntOrStr = Union[int, str]
...
+    else:
+        reveal_type(x)  # revealed: range
+
+def _(x: int | str | None):
+    if isinstance(x, Union[int, None]):
+        reveal_type(x)  # revealed: int | None
     else:
-        # TODO: Should be `range`
-        reveal_type(x)  # revealed: int | str | bytes | memoryview[int] | range
+        reveal_type(x)  # revealed: str
```
AlexWaygood, line 173: I guess you could also add a test with `Optional`, for completeness?

```pycon
>>> from typing import *
>>> if isinstance(42, Optional[int]):
...     print('hooray')
...     
hooray
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -9399,39 +9392,102 @@ impl InferredAs {
     }
 }
 
-/// A salsa-interned list of types.
-///
-/// # Ordering
-/// Ordering is based on the context's salsa-assigned id and not on its values.
-/// The id may change between runs, or when the context was garbage collected and recreated.
```
AlexWaygood, line 9406: seems like we could add this note to the new struct as well?

crates/ty_python_semantic/resources/mdtest/narrow/isinstance.md
```diff
@@ -289,6 +288,23 @@ def _(flag: bool):
...
+def _(x: list[str] | list[int] | list[bytes]):
+    # TODO: Ideally, this would be an error (requires https://github.com/astral-sh/ty/issues/116)
+    if isinstance(x, list[int]):
+        # No narrowing here:
+        reveal_type(x)  # revealed: list[str] | list[int] | list[bytes]
+
+    # error: [invalid-argument-type] "Invalid second argument to `isinstance`"
+    if isinstance(x, list[int] | list[str]):
+        # No narrowing here:
+        reveal_type(x)  # revealed: list[str] | list[int] | list[bytes]
```
AlexWaygood, line 330: could we also add an examples like:

```py
ListStrOrInt = Union[list[str], int]

def _(x: dict[int, str] | ListStrOrInt):
    if isinstance(x, ListStrOrInt):
        reveal_type(x)

    if isinstance(x, Union[list[str], int]):
        reveal_type(x)
```

Both fail at runtime, so ideally we wouldn't do any narrowing (or we'd intersect with `Unknown` rather than `list[str] | int`), and ideally we'd flag the error with a diagnostic.

But this very rarely comes up, so it's obviously fine if this fails right now and we just add TODOs to the test

sharkdp:
Looks like there are minor performance improvements

image

PR #21533, opened 2025-11-20
[ty] Implement goto-type for inlay type hints

Gankra:
This PR generalizes the signature_help system's SignatureWriter which could get the subspans of function parameters.
We now have TypeDetailsWriter which is threaded between type's display implementations via a new `fmt_detailed` method that many of the Display types now have.

With this information we can properly add goto-type targets to our inlay hints. This also lays groundwork for any future "I want to render a type but get spans" work.

Also a ton of lifetimes are introduced to avoid things getting conflated with `'db`.

This PR is broken up into a series of commits:

* Generalizing `SignatureWriter` to `TypeDetailsWriter`, but not using it anywhere else. This commit was confirmed to be a non-functional change (no test results changed)
* Introducing `fmt_detailed` everywhere to thread through `TypeDetailsWriter` and annotate various spans as "being" a given Type -- this is also where I had to reckon with a ton of erroneous `&'db self`. This commit was also confirmed to be a non-functional change.
* Finally, actually using the results for goto-type on inlay hints!
* Regenerating snapshots, fixups, etc.

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -570,7 +570,10 @@ mod tests {
                 write!(buf, "{}", diag.display(&self.db, &config)).unwrap();
             }
 
-            buf
+            // Windows path normalization for typeshed references
+            // "hey why is \x08 getting clobbered to /x08?"
+            // no it's not I don't know what you're talking about
+            buf.replace('\\', "/")
         }
     }
 
@@ -1673,7 +1676,7 @@ mod tests {
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
            |     ^^^^
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
            |
 
         info[inlay-hint-location]: Inlay Hint Target
@@ -1693,7 +1696,7 @@ mod tests {
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
            |                    ^^^
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
            |
 
         info[inlay-hint-location]: Inlay Hint Target
@@ -1711,7 +1714,7 @@ mod tests {
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
            |     ^^^^
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
            |
 
@@ -1730,7 +1733,7 @@ mod tests {
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
            |                    ^^^^^^^^
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
            |
 
@@ -1747,7 +1750,7 @@ mod tests {
            |
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
            |     ^^^^
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
@@ -1767,7 +1770,7 @@ mod tests {
            |
          8 | g[: list[Unknown | str]] = [f"{ft}", f"{ft}"]
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
            |                    ^^^^^
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
@@ -1785,7 +1788,7 @@ mod tests {
           --> main2.py:11:5
            |
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
            |     ^^^^
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
@@ -1804,7 +1807,7 @@ mod tests {
           --> main2.py:11:20
            |
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
            |                    ^^^
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
@@ -1822,7 +1825,7 @@ mod tests {
           --> main2.py:11:26
            |
          9 | h[: list[Unknown | Template]] = [t"wow %d", t"wow %d"]
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
            |                          ^^^^^
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
@@ -1839,7 +1842,7 @@ mod tests {
         info: Source
           --> main2.py:12:5
            |
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
            |     ^^^^
@@ -1857,7 +1860,7 @@ mod tests {
         info: Source
           --> main2.py:12:20
            |
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
            |                    ^^^
@@ -1874,7 +1877,7 @@ mod tests {
         info: Source
           --> main2.py:12:26
            |
-        10 | i[: list[Unknown | bytes]] = [b'\x01', b'\x02']
+        10 | i[: list[Unknown | bytes]] = [b'/x01', b'/x02']
         11 | j[: list[Unknown | int | float]] = [+1, +2.0]
         12 | k[: list[Unknown | int | float]] = [-1, -2.0]
            |                          ^^^^^


Gankra:
Have had this bee in my bonnet all week the IDE feels SO much better with these being clickable!

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -329,16 +504,17 @@ struct ClassDisplay<'db> {
...
         let qualification_level = self.settings.qualified.get(&**self.class.name(self.db));
         if qualification_level.is_some() {
             for parent in self.class.qualified_name_components(self.db) {
                 f.write_str(&parent)?;
                 f.write_char('.')?;
             }
         }
-        f.write_str(self.class.name(self.db))?;
+        f.with_detail(TypeDetail::Type(Type::ClassLiteral(self.class)))
+            .write_str(self.class.name(self.db))?;
```
Gankra, line 566: This is the single-most important type span we introduce, since it's nearly every nominal type's name.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -467,30 +659,35 @@ impl Display for DisplayRepresentation<'_> {
...
-                                    .display_with(self.db, self.settings.clone()),
-                            );
+                            if !is_first {
+                                f.write_str(separator)?;
+                            }
+                            is_first = false;
+                            signature
+                                .bind_self(self.db, Some(typing_self_ty))
+                                .display_with(self.db, self.settings.clone())
+                                .fmt_detailed(f)?;
```
Gankra: With apologies to `join` enjoyers but I had to hand-roll this gunk everywhere to invoke fmt_detailed

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -467,30 +659,35 @@ impl Display for DisplayRepresentation<'_> {
...
+                            .display_with(self.db, self.settings.singleline())
+                            .fmt_detailed(f)?;
+                        f.write_char('.')?;
+                        f.with_detail(TypeDetail::Type(self.ty))
+                            .write_str(function.name(self.db))?;
+                        type_parameters.fmt_detailed(f)?;
+                        signature
+                            .bind_self(self.db, Some(typing_self_ty))
+                            .display_with(self.db, self.settings.clone())
+                            .fmt_detailed(f)
```
Gankra, line 722: A lot of the PR is me breaking up a nice pretty `write!` into individual parts to invoke `fmt_detailed` where appropriate.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -653,11 +856,11 @@ impl Display for DisplayBoundTypeVarIdentity<'_> {
...
 
 impl<'db> TupleSpec<'db> {
-    pub(crate) fn display_with(
-        &'db self,
+    pub(crate) fn display_with<'a>(
+        &'a self,
         db: &'db dyn Db,
         settings: DisplaySettings<'db>,
-    ) -> DisplayTuple<'db> {
+    ) -> DisplayTuple<'a, 'db> {
```
Gankra, line 910: Deconflating these lifetimes was unfortunately loadbearing. I had to do that a lot.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -1169,37 +1426,40 @@ impl<'db> Signature<'db> {
...
         }
     }
 
-    /// Internal method to write signature with the signature writer
-    fn write_signature(&self, writer: &mut SignatureWriter) -> fmt::Result {
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
+        // Immediately write a marker signaling we're starting a signature
+        let _ = f.with_detail(TypeDetail::SignatureStart);
+        // When we exit this function, write a marker signaling we're ending a signature
+        let mut f = f.with_detail(TypeDetail::SignatureEnd);
```
Gankra, line 1495: An incredible victory for not wanting to have to reverse-engineer what parameters belong to nested signatures.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -641,6 +867,81 @@ mod tests {
...
+            |       ^^^
+        349 |     """int([x]) -> integer
+        350 |     int(x, base=10) -> integer
+            |
+        info: Source
+          --> main.py:9:2
+           |
+         7 | x = (1, 'abc')
+         8 | y = x
+         9 | z = (i(1), s('abc'))
```
MichaReiser: I find the tests here confusing where there's no inlay. Aren't those inlays that are hidden? Why do we render the inlay hint targets for those? I think we should omit those from the tests.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -101,6 +101,163 @@ impl<'db> DisplaySettings<'db> {
...
+/// something like `&'db self`, which, while convenient, and sometimes works, is imprecise.
+struct TypeDetailGuard<'a, 'b, 'c, 'db> {
+    inner: &'c mut TypeWriter<'a, 'b, 'db>,
+    start: Option<usize>,
+    payload: Option<TypeDetail<'db>>,
+}
+
+impl Drop for TypeDetailGuard<'_, '_, '_, '_> {
+    fn drop(&mut self) {
+        // The fallibility here is primarily retrieving `TypeWriter::Details`
```
MichaReiser, line 283: Should we bail early if `std::thread::panicking`?

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -101,6 +101,163 @@ impl<'db> DisplaySettings<'db> {
...
+        // everything else is ideally-never-fails pedantry (yay for pedantry!)
+        if let TypeWriter::Details(details) = &mut self.inner
+            && let Ok(end) = u32::try_from(details.label.len())
```
MichaReiser:
```suggestion
            && let Ok(end) = details.label.text_len()
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -101,6 +101,163 @@ impl<'db> DisplaySettings<'db> {
...
+
+/// Look on my Works, ye Mighty, and despair!
+///
+/// It's quite important that we avoid conflating any of these lifetimes, or else the
+/// borrowchecker will throw a ton of confusing errors about things not living long
+/// enough. If you get those kinds of errors, it's probably because you introduced
+/// something like `&'db self`, which, while convenient, and sometimes works, is imprecise.
+struct TypeDetailGuard<'a, 'b, 'c, 'db> {
+    inner: &'c mut TypeWriter<'a, 'b, 'db>,
+    start: Option<usize>,
```
MichaReiser:
```suggestion
    start: Option<TextSize>,
```

Seems better to error early rather than silently skipping if `TextSize::try_from` fails in the `Drop`

crates/ty_python_semantic/src/types/display.rs
MichaReiser: You could introduce a `FmtDetailed` or `DisplayDetailed` trait and then preserve the existing formatting, making the overall code more like `Display` and `Debug`


```patch
Index: crates/ty_python_semantic/src/types/display.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/crates/ty_python_semantic/src/types/display.rs b/crates/ty_python_semantic/src/types/display.rs
--- a/crates/ty_python_semantic/src/types/display.rs	(revision 13fecff48afe8065b05e012e7f0cc5301bd244c4)
+++ b/crates/ty_python_semantic/src/types/display.rs	(date 1763628882673)
@@ -190,7 +190,17 @@
             payload: Some(detail),
         }
     }
+
+    fn join<'c>(&'c mut self, separator: &'static str) -> Join<'a, 'b, 'c, 'db> {
+        Join {
+            fmt: self,
+            separator,
+            result: Ok(()),
+            seen_first: false,
+        }
+    }
 }
+
 impl std::fmt::Write for TypeWriter<'_, '_, '_> {
     fn write_str(&mut self, val: &str) -> fmt::Result {
         match self {
@@ -205,6 +215,46 @@
     }
 }
 
+trait FmtDetailed<'db> {
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result;
+}
+
+struct Join<'a, 'b, 'c, 'db> {
+    fmt: &'c mut TypeWriter<'a, 'b, 'db>,
+    separator: &'static str,
+    result: fmt::Result,
+    seen_first: bool,
+}
+
+impl<'db> Join<'_, '_, '_, 'db> {
+    fn entry(&mut self, item: &dyn FmtDetailed<'db>) -> &mut Self {
+        if self.seen_first {
+            self.result = self
+                .result
+                .and_then(|()| self.fmt.write_str(self.separator));
+        } else {
+            self.seen_first = true;
+        }
+        self.result = self.result.and_then(|()| item.fmt_detailed(self.fmt));
+        self
+    }
+
+    fn entries<I, F>(&mut self, items: I) -> &mut Self
+    where
+        I: IntoIterator<Item = F>,
+        F: FmtDetailed<'db>,
+    {
+        for item in items {
+            self.entry(&item);
+        }
+        self
+    }
+
+    fn finish(&mut self) -> fmt::Result {
+        self.result
+    }
+}
+
 pub enum TypeDetail<'db> {
     /// Dummy item to indicate a function signature's parameters have started
     SignatureStart,
@@ -412,7 +462,9 @@
             TypeWriter::Formatter(_) => unreachable!("Expected Details variant"),
         }
     }
+}
 
+impl<'db> FmtDetailed<'db> for DisplayType<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let representation = self.ty.representation(self.db, self.settings.clone());
         match self.ty {
@@ -504,7 +556,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> ClassDisplay<'db> {
+impl<'db> FmtDetailed<'db> for ClassDisplay<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let qualification_level = self.settings.qualified.get(&**self.class.name(self.db));
         if qualification_level.is_some() {
@@ -556,7 +608,7 @@
     }
 }
 
-impl<'db> DisplayRepresentation<'db> {
+impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         match self.ty {
             Type::Dynamic(dynamic) => write!(f, "{dynamic}"),
@@ -678,17 +730,15 @@
                             f.write_str("Overload[")?;
                         }
                         let separator = if self.settings.multiline { "\n" } else { ", " };
-                        let mut is_first = true;
+                        let mut join = f.join(separator);
                         for signature in signatures {
-                            if !is_first {
-                                f.write_str(separator)?;
-                            }
-                            is_first = false;
-                            signature
-                                .bind_self(self.db, Some(typing_self_ty))
-                                .display_with(self.db, self.settings.clone())
-                                .fmt_detailed(f)?;
+                            join.entry(
+                                &signature
+                                    .bind_self(self.db, Some(typing_self_ty))
+                                    .display_with(self.db, self.settings.clone()),
                         }
+                        join.finish()?;
                         if !self.settings.multiline {
                             f.write_str("]")?;
                         }
@@ -875,7 +925,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayTuple<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayTuple<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         f.write_str("tuple[")?;
         match self.tuple {
@@ -968,7 +1018,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayOverloadLiteral<'db> {
+impl<'db> FmtDetailed<'db> for DisplayOverloadLiteral<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let signature = self.literal.signature(self.db);
         let type_parameters = DisplayOptionalGenericContext {
@@ -1012,7 +1062,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayFunctionType<'db> {
+impl<'db> FmtDetailed<'db> for DisplayFunctionType<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let signature = self.ty.signature(self.db);
 
@@ -1088,7 +1138,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayGenericAlias<'db> {
+impl<'db> FmtDetailed<'db> for DisplayGenericAlias<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         if let Some(tuple) = self.specialization.tuple(self.db) {
             tuple
@@ -1160,7 +1210,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayOptionalGenericContext<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayOptionalGenericContext<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         if let Some(generic_context) = self.generic_context {
             generic_context
@@ -1187,14 +1237,6 @@
 }
 
 impl<'db> DisplayGenericContext<'_, 'db> {
-    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
-        if self.full {
-            self.fmt_full(f)
-        } else {
-            self.fmt_normal(f)
-        }
-    }
-
     fn fmt_normal(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let variables = self.generic_context.variables(self.db);
 
@@ -1229,6 +1271,16 @@
     }
 }
 
+impl<'db> FmtDetailed<'db> for DisplayGenericContext<'_, 'db> {
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
+        if self.full {
+            self.fmt_full(f)
+        } else {
+            self.fmt_normal(f)
+        }
+    }
+}
+
 impl Display for DisplayGenericContext<'_, '_> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         self.fmt_detailed(&mut TypeWriter::Formatter(f))
@@ -1276,14 +1328,6 @@
 }
 
 impl<'db> DisplaySpecialization<'db> {
-    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
-        if self.full {
-            self.fmt_full(f)
-        } else {
-            self.fmt_normal(f)
-        }
-    }
-
     fn fmt_normal(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         f.write_char('[')?;
         let types = self.specialization.types(self.db);
@@ -1320,6 +1364,16 @@
     }
 }
 
+impl<'db> FmtDetailed<'db> for DisplaySpecialization<'db> {
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
+        if self.full {
+            self.fmt_full(f)
+        } else {
+            self.fmt_normal(f)
+        }
+    }
+}
+
 impl Display for DisplaySpecialization<'_> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         self.fmt_detailed(&mut TypeWriter::Formatter(f))
@@ -1370,7 +1424,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayCallableType<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayCallableType<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         match self.signatures.overloads.as_slice() {
             [signature] => signature
@@ -1444,7 +1498,9 @@
             TypeWriter::Formatter(_) => unreachable!("Expected Details variant"),
         }
     }
+}
 
+impl<'db> FmtDetailed<'db> for DisplaySignature<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         // Immediately write a marker signaling we're starting a signature
         let _ = f.with_detail(TypeDetail::SignatureStart);
@@ -1564,7 +1620,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayParameter<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayParameter<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         if let Some(name) = self.param.display_name() {
             f.write_str(&name)?;
@@ -1666,7 +1722,7 @@
     max_when_elided: 3,
 };
 
-impl<'db> DisplayUnionType<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayUnionType<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         fn is_condensable(ty: Type<'_>) -> bool {
             matches!(
@@ -1832,7 +1888,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayIntersectionType<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayIntersectionType<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let tys = self
             .ty
@@ -1887,7 +1943,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayMaybeNegatedType<'db> {
+impl<'db> FmtDetailed<'db> for DisplayMaybeNegatedType<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         if self.negated {
             f.write_str("~")?;
@@ -1913,7 +1969,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayMaybeParenthesizedType<'db> {
+impl<'db> FmtDetailed<'db> for DisplayMaybeParenthesizedType<'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let write_parentheses = |f: &mut TypeWriter<'_, '_, 'db>| {
             f.write_char('(')?;
@@ -2001,7 +2057,7 @@
     settings: DisplaySettings<'db>,
 }
 
-impl<'db> DisplayTypeArray<'_, 'db> {
+impl<'db> FmtDetailed<'db> for DisplayTypeArray<'_, 'db> {
     fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let mut is_first = true;
         for ty in self.types {
```

MichaReiser (commented):
The implementation looks good to me. I've one suggestion that should help to reduce the necessary changes because of `Join`. 

I do find the test very confusing as we now see go to targets in places where we don't show any inlay hints. I think we should change this before landing this PR

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -245,22 +402,40 @@ pub struct DisplayType<'db> {
...
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         let representation = self.ty.representation(self.db, self.settings.clone());
         match self.ty {
             Type::IntLiteral(_)
             | Type::BooleanLiteral(_)
             | Type::StringLiteral(_)
             | Type::BytesLiteral(_)
             | Type::EnumLiteral(_) => {
-                write!(f, "Literal[{representation}]")
+                f.write_str("Literal[")?;
```
AlexWaygood, line 473: It would be cool if we could also have goto-definition for the `Literal` part of e.g. `Literal[42]`. We don't have any documentation for `typing.Literal` in our vendored typeshed stubs right now (because we currently only add docstrings for classes, functions and modules), but it would be nice to add that capability to docstring-adder at some point. Then when you goto definition for [the symbol `Literal`](https://github.com/astral-sh/ruff/blob/02c102da885922906aa57a0468fde909cbbd2ba4/crates/ty_vendored/vendor/typeshed/stdlib/typing.pyi#L351) you'd see all of this lovely explanation in the stub file to tell you how `Literal` types work (this is the docstring for the symbol at runtime):

```
% uvx python -m pydoc typing.Literal   
Help on _TypedCacheSpecialForm in typing:

typing.Literal = typing.Literal
    Special typing form to define literal types (a.k.a. value types).

    This form can be used to indicate to type checkers that the corresponding
    variable or function parameter has a value equivalent to the provided
    literal (or one of several literals)::

        def validate_simple(data: Any) -> Literal[True]:  # always returns True
            ...

        MODE = Literal['r', 'rb', 'w', 'wb']
        def open_helper(file: str, mode: MODE) -> str:
            ...

        open_helper('/some/path', 'r')  # Passes type check
        open_helper('/other/path', 'typo')  # Error in type checker

    Literal[...] cannot be subclassed. At runtime, an arbitrary value
    is allowed as type argument to Literal[...], but type checkers may
    impose restrictions.
```

The symbol `typing.Literal` itself has type `Type::SpecialForm(SpecialFormType::Literal)` in our model, so would you do that like this?

```suggestion
                f.with_detail(TypeDetail::Type(Type::SpecialForm(SpecialFormType::Literal))).write_str("Literal")?;
                f.write_char('[')?;
```

crates/ty_ide/src/inlay_hints.rs
Gankra: Fixed (the test spans referenced the original doc and not the inlayed one)

crates/ty_python_semantic/src/types/display.rs
Gankra: I don't think that's necessary?

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -463,40 +501,44 @@ mod tests {
...
 
         /// Returns the inlay hints for the given test case with custom settings.
-        fn inlay_hints_with_settings(&self, settings: &InlayHintSettings) -> String {
+        fn inlay_hints_with_settings(&mut self, settings: &InlayHintSettings) -> String {
             let hints = inlay_hints(&self.db, self.file, self.range, settings);
 
             let mut buf = source_text(&self.db, self.file).as_str().to_string();
 
-            let mut diagnostics = Vec::new();
+            let mut tbd_diagnostics = Vec::new();
```
MichaReiser, line 509: Are those to be defined diagnostics?

crates/ty_ide/src/inlay_hints.rs
MichaReiser: Okay, that makes way more sense now :)

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -370,35 +552,41 @@ struct DisplayRepresentation<'db> {
...
+        self.fmt_detailed(&mut TypeWriter::Formatter(f))
+    }
+}
+
+impl<'db> DisplayRepresentation<'db> {
+    fn fmt_detailed(&self, f: &mut TypeWriter<'_, '_, 'db>) -> fmt::Result {
         match self.ty {
-            Type::Dynamic(dynamic) => dynamic.fmt(f),
-            Type::Never => f.write_str("Never"),
+            Type::Dynamic(dynamic) => write!(f, "{dynamic}"),
```
AlexWaygood, line 611: if the type is `Type::Dynamic(DynamicType::Any)`, it would be cool if we could goto the definition for the `Any` symbol (with has type `Type::SpecialForm(SpecialFormType::Any)` in our model)

Gankra:
Micha you're god damn incredible, the join impl Just Worked.

crates/ty_ide/src/inlay_hints.rs
Gankra: I have to gather them up transiently while we create the annotated file, and then create the annotated file, and then create them properly.

crates/ty_ide/src/inlay_hints.rs
AlexWaygood, line 1: Could you also add tests for generic-alias types, e.g.

```py
class Foo[T]: ...

a [: <class 'Foo[int]'>] = Foo[int]
```

and subclass-of types, e.g.

```py
def f(x: list[str]):
    y [: type[list[str]]] = type(x)
```

For the first one I'd expect goto targets for both `Foo` and `int` in `<class 'Foo[int]>`. For the second one I'd expect goto targets for both `list `and `str` in `type[list[str]]`

AlexWaygood:
looks like you'll need to add some snapshot filters to get snapshots that include filepaths working on Windows 🙃

crates/ty_python_semantic/src/types/display.rs
Gankra: I know you literally gave me the implementation but I'd like to keep these things to followups (I'm terrified of this PR getting merge conflicted to hell).

crates/ty_python_semantic/src/types/display.rs
Gankra: (similarly leaving as a followup)

crates/ty_ide/src/inlay_hints.rs
Gankra: Cool with this also being a followup?

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: np

crates/ty_ide/src/inlay_hints.rs
AlexWaygood: of course!

Gankra:
Ughhhh

```
      299   299 │ info[inlay-hint-location]: Inlay Hint Target
      300       │-  --> stdlib/string/templatelib.pyi:10:7
            300 │+  --> stdlib\string\templatelib.pyi:10:7
```

This is really annoying because `\` can appear in like `\x08` and I think I'm just gonna ask everyone to agree to not Notice I'm clobbering those in the inlay hint tests.

Gankra:
Actually I'm extremely confused why this isn't a problem in any of the goto tests

Gankra:
(More followups for the followup pile I guess)

MatthewMckee4:
@Gankra, Sick!

Issue #21403, opened 2025-11-12
Class property without return


ashrub-holvi:
### Summary

Hi,
possible linter check - method wrapped to property decorator must have at least one explicit return, possible issue is:
```
In [1]: class A:
   ...:     @property
   ...:     def b(self):
   ...:         "Hi"
   ...: 

In [3]: A().b is None
Out[3]: True
```
of course it should be covered by tests, but sometimes is not, especially if it's some special path in the code.

ntBre:
This sounds reasonable to me. I thought one of our RET rules might activate here, but I didn't see anything in the [playground](https://play.ruff.rs/b7028262-015d-4f46-a0dd-f2f142559c37) even with an annotation. [useless-expression (B018)](https://docs.astral.sh/ruff/rules/useless-expression/#useless-expression-b018) triggers if you make the body an int or something, to avoid "Hi" looking like a docstring, but that wouldn't help in general.

@amyreese what do you think?

amyreese:
Makes sense to me. Same for `@functools.cached_property`.

ShaharNaveh:
How would the new rule handle this case:

```py
import abc


class Meta(metaclass=abc.ABCMeta):
    @property
    @abc.abstractmethod
    def prop1(self):
        ...
 
    @property
    @abc.abstractmethod
    def prop2(self):
        """
        Some docstring
        """
```

Will it be ignored if `abc.abstractmethod` is one of the decorators?

MichaReiser:
Yes, it should ignore abstract methods. We could even consider checking if `class` inherits `Protocol` (only directly) and, if so, ignore the error

Issue #14900, opened 2024-12-10
PLC0206 vs PERF102 limbo ...


oz123:
First it tells me no to use:
```
for key in dict:
      do_someting(key)
```
When I try `dict.keys` it still complains.
When I try `dict.item() and throw away value, it still complains:

```
oznt@dell-gentoo:/srv/oznt/Software/pypa/pipenv |pipenv-mRt4xz3R| [±|completely-remove-click-echo {2} U:3 ?:6 ✗|] $ruff check pipenv/routines/outdated.py 
pipenv/routines/outdated.py:50:23: PERF102 When using only the keys of a dict use the `keys()` method
   |
48 |     outdated = []
49 |     skipped = []
50 |     for package, _ in packages.items():
   |                       ^^^^^^^^^^^^^^ PERF102
51 |         norm_name = pep423_name(package)
52 |         if norm_name in updated_packages:
   |
   = help: Replace `.items()` with `.keys()`

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
oznt@dell-gentoo:/srv/oznt/Software/pypa/pipenv |pipenv-mRt4xz3R| [±|completely-remove-click-echo {2} U:3 ?:6 ✗|] $vim pipenv/routines/outdated.py 
oznt@dell-gentoo:/srv/oznt/Software/pypa/pipenv |pipenv-mRt4xz3R| [±|completely-remove-click-echo {2} U:3 ?:6 ✗|] $vim pipenv/routines/outdated.py 
oznt@dell-gentoo:/srv/oznt/Software/pypa/pipenv |pipenv-mRt4xz3R| [±|completely-remove-click-echo {2} U:2 ?:6 ✗|] $ruff check pipenv/routines/outdated.py 
pipenv/routines/outdated.py:50:5: PLC0206 Extracting value from dictionary without calling `.items()`
   |
48 |       outdated = []
49 |       skipped = []
50 |       for package in packages.keys():
   |  _____^
51 | |         norm_name = pep423_name(package)
52 | |         if norm_name in updated_packages:
53 | |             version = packages[package]
54 | |             if isinstance(version, Mapping):
55 | |                 version = parse_version(version.get("version", "").replace("==", ""))
56 | |             else:
57 | |                 version = parse_version(version.replace("==", ""))
58 | |             if updated_packages[norm_name] != version:
59 | |                 outdated.append(
[...]

PR #21535, opened 2025-11-20
Add rule to detect unnecessary class properties

ShaharNaveh:
## Summary

Fixes https://github.com/astral-sh/ruff/issues/21403

## Test Plan

Added tests


crates/ruff_linter/resources/test/fixtures/ruff/RUF066.py
```diff
@@ -0,0 +1,75 @@
+"""
+Should emit:
+RUF066 - on lines 12, 23, 48
```
ntBre: Would you mind adding inline comments marking each case as okay or an error instead of one comment at the top? That way it will appear in each of the snapshots. That makes it a bit easier to review.

And could you renumber the rule to RUF069? I think we have PRs in flight for [66](https://github.com/astral-sh/ruff/pull/9911), [67](https://github.com/astral-sh/ruff/pull/20585), and [68](https://github.com/astral-sh/ruff/pull/21079) already.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::checkers::ast::Checker;
+use crate::{FixAvailability, Violation};
+
+/// ## What it does
+/// Detects unnecessary `@property` methods.
+///
+/// An unnecessary property is a property that does not:
+/// - return
+/// - yield (or `yield from`)
+/// - raises
```
ntBre: Should this apply to yield and raises? I kind of expected it only return to suppress the diagnostic, but maybe yield and raise are more common than I thought.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,127 @@
...
+use ruff_python_ast::visitor::{Visitor, walk_expr, walk_stmt};
+use ruff_python_ast::{Expr, Stmt, StmtFunctionDef};
+use ruff_python_semantic::analyze::{function_type, visibility};
+use ruff_text_size::Ranged;
+
+use crate::checkers::ast::Checker;
+use crate::{FixAvailability, Violation};
+
+/// ## What it does
+/// Detects unnecessary `@property` methods.
```
ntBre: "Unnecessary" doesn't feel like exactly the right description to me. I think something like `property-without-return` for the rule name, along the lines of the issue title, might make more sense.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        match expr {
+            Expr::Yield(_) | Expr::YieldFrom(_) => self.necessary = true,
+            _ => walk_expr(self, expr),
+        }
+    }
+
+    fn visit_stmt(&mut self, stmt: &Stmt) {
+        if !self.necessary {
```
ntBre: nit: Can we early return here too, like in `visit_expr` above?

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,127 @@
...
+            match stmt {
+                // Sometimes a property is defined because of an ABC requiremnet but it will always raise.
+                // Thus not an unnecessary property.
```
ntBre: Ah okay, this probably answers my question above about the raise and yield. I was mostly just curious.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    fn visit_body(&mut self, body: &[Stmt]) {
+        for stmt in body {
+            self.visit_stmt(stmt);
+            if self.necessary {
+                return;
+            }
+        }
+    }
```
ntBre: I think it's _probably_ okay to skip this since we bail out of `visit_stmt` itself, but it's also okay with me either way.

ntBre (changes requested):
Thanks! This looks good to me overall, just some minor nits and a suggestion about the rule name and code.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
ShaharNaveh: Raise is more common than yields. The reason why I included `yield` is because of the ecosystem CI report that showed an [instance where apache-airflow would use `yield` inside a property called "stream"](https://github.com/apache/airflow/blob/e9d41dc60d3e88b838b8de6464a1322f2667d078/airflow-core/src/airflow/utils/log/log_stream_accumulator.py#L111-L136) and that made a lot of sense to me, so I decided to include it.

I can remove the `yield` inclusion if that's desired

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
ShaharNaveh: The initial name was something like that, but I decided it's no longer accurate after the inclusion of `raise` and `yield`, would you still want me to revert it to the old name?

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
ShaharNaveh: Good catch. I prefer to have less code as possible

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
```diff
@@ -0,0 +1,121 @@
...
+///
+/// Use instead:
+/// ```python
+/// class User:
+///     @property
+///     def full_name(self):
+///         return f"{self.first_name} {self.last_name}"
+/// ```
+#[derive(ViolationMetadata)]
+#[violation_metadata(preview_since = "0.14.6")]
```
ntBre: tiny nit since I did a release this morning:


```suggestion
#[violation_metadata(preview_since = "0.14.7")]
```

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
ntBre: Thanks, yeah that makes sense to me. I chatted with @amyreese, and I think it makes sense to her too, so it sounds good to leave all three in the rule.

crates/ruff_linter/src/rules/ruff/rules/unnecessary_property.rs
ntBre: Ah, I see what you mean, but I think it's still okay for the name and diagnostic message only to mention `return`. @amyreese had a good point that `return` will probably be the most common fix when the diagnostic fires too, so it seems good to emphasize that option.

ntBre (changes requested):
Thank you, this looks great!

I think we just need to bump the `preview_since` version and update the rule name/docs and this is good to go.

crates/ruff_linter/src/rules/ruff/rules/property_without_return.rs
```diff
@@ -0,0 +1,119 @@
...
+    visitor.visit_body(body);
+    if visitor.found {
+        return;
+    }
+
+    checker.report_diagnostic(
+        PropertyWithoutReturn {
+            name: name.to_string(),
+        },
+        function_def.range(),
```
amyreese: Sorry to add one more change needed. We should pick a smaller range for the diagnostic, rather than throwing a yellow squiggle on an entire function in an editor, as changing the diagnostic range later will potentially be a [breaking change](https://github.com/astral-sh/ruff/issues/14900#issuecomment-2543923625).

I think I'd suggest just marking the `@property` decorator itself, since that's the part that's unfulfilled by the body of the function.

MichaReiser: In some cases, we highlight the range of the function's name. You can see other usages by searching for references to the `Identifier` trait

https://github.com/astral-sh/ruff/blob/9ae698fe30cf3526f0e7ae237b800b3ed19a819f/crates/ruff_python_ast/src/identifier.rs#L18-L21

ShaharNaveh: Thank you @amyreese & @MichaReiser for the suggestions.

I agree with highlighting only one line instead of the entire function block. 

I ended up highlighting only the function name (@MichaReiser suggestion), just because it's a personal preference of mine. LMK if you want me to change that to highlight something else instead

crates/ruff_linter/src/rules/ruff/rules/property_without_return.rs
```diff
@@ -0,0 +1,119 @@
...
+    name: String,
+}
+
+impl Violation for PropertyWithoutReturn {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::None;
+
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let Self { name } = self;
+        format!("{name} is a property without a `return` statement")
```
amyreese: I think we prefer to put the name in backticks for clarity:

```suggestion
        format!("`{name}` is a property without a `return` statement")
```

MichaReiser:
Thanks this looks good. I just noticed that you recoded the rule from RUF066 to RUF069 before. Can you tell me why? It seems to me that RUF066 is still available.

MichaReiser:
Oh, never mind. It's because there are other in-flight PRs using that code. I recoded it back to RUF066, because I want to avoid holes and your PR is going to land first :)

PR #21539, opened 2025-11-20
[ty] Distinguish "unconstrained" from "constrained to any type"

dcreager:
Before, we would collapse any constraint of the form `Never ≤ T ≤ object` down to the "always true" constraint set. This is correct in terms of BDD semantics, but loses information, since "not constraining a typevar at all" is different than "constraining a typevar to take on any type". Once we get to specialization inference, we should fall back on the typevar's default for the former, but not for the latter.

This is much easier to support now that we have a sequent map, since we need to treat `¬(Never ≤ T ≤ object)` as being impossible, and prune it when we walk through BDD paths, just like we do for other impossible combinations.


crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -168,9 +170,11 @@ from typing import Any
 # fmt: off
 
 def constrained_by_gradual[T: (Base, Any)]():
```
carljm, line 173: This is a weird case, and the more I think about it, the weirder it gets.

I _think_ the principles of constrained type var should probably be:

1. We should always solve to one of the constrained types (or to `Unknown` or `Any`). This is the core invariant of constrained type variables.
2. If the constraints don't allow us to narrow down to a single constrained type, but could allow multiple, we should prefer the subtype, if one constrained type is a subtype of the other. For example, for `def f[T: (Base, Sub)](x: T)` we should prefer solving `f(typed_as sub)` to `Sub` rather than `Base`, even though both solutions would be possible.
3. If multiple constrained types are possible, and neither is a subtype of the other, we should... do what? Fall back to the typevar default?

But if we agree with these principles, then we can _never_ solve this constrained type var to `Base`, because any set of constraints that would allow us to solve to `Base` would also allow solving to `Any`, therefore we can't narrow it down to one of the choices.

[mypy](https://mypy-play.net/?mypy=latest&python=3.13&gist=25de03cec23e19bffce364d7b02e7845) seems to just always solve such a typevar to `Any`. (Mypy doesn't support typevar defaults, nor distinguish `Any` vs `Unknown`, so I can't really experiment with how it handles fallback-to-default.)

[pyright](https://pyright-play.net/?pyrightVersion=1.1.405&pythonVersion=3.13&strict=true&reportUnreachable=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcAUCQMYA2AhgM61QDKArggKYgBcJUvUCdWmSqCoAITpsAFC3YgAlNz79BwmvSbMARlIm02innwH01ogKooQbGjDYATJcdUl7bYFGABtACqcoupIANITE8lAAvOKSALpSAB7%2BPlAAPlAAcmAobJEZWQZQALQAfFB%2BRrxIHvGYDChg%2BJnZTsq81jDMIChQ8WTKINRI%2BlAASswoMFhsAKIg4CBSAET17IvyZG4eAPpSYFoAVv57%2B2zkMCG0rP6yHCFakv56bBfa19oh41xQlta2Dob9NgANzY1EoW3g7CkwF2B3k60BILBEMQ0hhlwQ8IqUGsSPBkLRUnu%2BixiNB%2BNR0Kkly0pL4uPJKKhMKxQA) also always solves to `Any`, unless the typevar is totally unconstrained, in which case it falls back to the typevar default (which I've explicitly specified to be `Base` in this example, but could also just implicitly be `Unknown`.) I'm not sure how pyright decides that `f(base)` or `f(sub)` should be `Any` rather than "the typevar default", given that those calls are compatible with either constrained type.

[pyrefly](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeS4ATrgLYAEALqcROgOZ0Q3G6UN0BBdKQA0dSjABuMVFAD6TYjAA66VQGMoqOHDoBlAK5LKiVXXN1i2uKo1addAELaYACkPGAlKfQXL123RNa30DbFdnOBhvMwsrHUDghwBVdAktBhhMHz94mzUsGDA6MABtABVEOgiXMSFSTzoAXicXAF1XfCryugAfOgA5XHQYZsHh6LoAWgA%2BOkrY8whi-E5ddFx%2BIZGcvwsJBgNKX3xAv0pUCCi6ACUDdAYuGABRSmpKV2UQDaUvz0DMEU6HJXLhsAArKpg8EwdQMMRwIxVDwwShibAuKqRGAIsLIsJie4mOipdKoTKYGK%2BfZSGTyRRuMCgiGef7U8wSaSyBTMRmuRHEVmLcS07kM1xMjFRIXskVc%2Bm8iX8sIy86ihVKJUykAiEAGR5QOAkciIEAAYhJBogTBK9zhEGG%2BVUgOKYF4NHJcnQBho2FRnSqrAYjVmdDgDBMwoOR18YC%2BAx9fuJwHwAF8vqodSAyBIwFBSIQGLQoBQLQAFUi5-NhjA4Ah0dTDSBsI7kh3oQiqC16GCjAAWDAYxDgiAA9KOc0V84ReGxRzB0KPMLh1HBR430M3W49hqOSrw6KhJJctNhYA2mxAWxcd75cMRb0bVGQGH3hlNpJQ4O2xl8AMyEAAjH%2BGboCAqa6qg9rSAAYtAMAUGgWB4EQZDgUAA) does the same thing you originally did here (always solve to a fully static type), which I'm pretty sure is wrong. It fails both the promises of a constrained type var and of the gradual guarantee, and I suspect will cause false positive errors in practice (the fact that pyright and mypy both prefer solving to `Any` strengthens this intuition.) If the `Any` constrained type means "there is really some static type here, but we don't know what it is" (imagine its an `Unknown` from a failed import of a type), then it's wrong to solve to some arbitrary fully static type, because we don't know that's what the constrained type was "supposed" to be - and it could cause a lot of cascading errors.

(I think it's fine to punt all of this, but I will suggest some different TODOs below based on this analysis.)

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -168,9 +170,11 @@ from typing import Any
...
+    # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Base]
+    reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.always()))
```
carljm, line 176: I would expect this to solve to the typevar default, in this case `Unknown`. What information allows us to pick `Base` over `Any` here? (If we are picking between `Base` and `Any`, I think `Any` is probably the better choice in practice, and is what mypy/pyright choose, but I'm not sure what heuristic they are using.)

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -168,9 +170,11 @@ from typing import Any
...
     # TODO: revealed: ty_extensions.Specialization[T@constrained_by_gradual = Any]
     # revealed: ty_extensions.Specialization[T@constrained_by_gradual = object]
-    reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.always()))
+    reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.range(Never, T, object)))
```
carljm, line 179: Here I think either `Any` or `Unknown` (typevar default) would be reasonable choices?

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
carljm, line 185: Pyright and mypy would also solve this to `Any`. Similar to above, I'm not sure why we'd pick `Base` over `Any` here as a solution.

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -180,15 +184,13 @@ def constrained_by_gradual[T: (Base, Any)]():
     # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Unrelated]
     reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.range(Never, T, Unrelated)))
 
-    # TODO: revealed: ty_extensions.Specialization[T@constrained_by_gradual = Any]
-    # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Super]
+    # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Base]
     reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.range(Never, T, Super)))
```
carljm, line 192: Same as above cases, I think this should also solve to `Any` or `Unknown`

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -180,15 +184,13 @@ def constrained_by_gradual[T: (Base, Any)]():
...
     # TODO: revealed: ty_extensions.Specialization[T@constrained_by_gradual = Any]
     # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Super]
     reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.range(Super, T, Super)))
 
-    # TODO: revealed: ty_extensions.Specialization[T@constrained_by_gradual = Any]
-    # revealed: ty_extensions.Specialization[T@constrained_by_gradual = object]
+    # revealed: ty_extensions.Specialization[T@constrained_by_gradual = Base]
     reveal_type(generic_context(constrained_by_gradual).specialize_constrained(ConstraintSet.range(Sub, T, object)))
```
carljm, line 199: And again here.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -2227,6 +2250,9 @@ impl<'db> ConstraintAssignment<'db> {
 ///
 /// We support several kinds of sequent:
 ///
+/// - `¬C₁ → false`: This indicates that `C₁` is always true. Any path that assumes it is false is
```
carljm, line 2265: Mostly just curiosity: is there an important reason why we frame this as `¬C₁ → false` rather than `C₁ → true`? Are these different in some way? The prose here suggests not.

carljm (approved):
Awesome!

crates/ty_python_semantic/src/types/constraints.rs
dcreager: "`C₁` is always true" is `true → C₁`, not `C₁ → true`. But given that change, you're right that `true → C₁` and `¬C₁ → false` are the same thing. One focuses on what is always true, and the other focuses on what's impossible. The focus of this data structure is to track situations that are impossible, so I thought it would be best to frame it with `false` on the rhs.

Though I guess that means I could go further and rewrite some the examples below ­— for instance `C₁ ∧ C₂ → D` and `C₁ ∧ C₂ ∧ ¬D → false` are equivalent in exactly the same way!

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: > But if we agree with these principles, then we can _never_ solve this constrained type var to `Base`, because any set of constraints that would allow us to solve to `Base` would also allow solving to `Any`, therefore we can't narrow it down to one of the choices.

Is it worth trying to consider which materializations of `Any` are valid specializations? If `Base` is a valid specialization but no subtypes are (if that were to violate some other requirement of the specialization), then the `Any` could not materialize to anything that would take precedence according to your rule (2).

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: > What information allows us to pick `Base` over `Any` here?

This is a side effect of the change I added from this discussion https://github.com/astral-sh/ruff/pull/21414#discussion_r2539830595. If there are multiple solutions, my original plan was to produce the union of them. But I don't think that's correct, so instead we either:

- find a single specialization that satisfies all of the solutions, or
- return `None` to indicate an ambiguous result.

Here, the constraint of `Any` can materialize to any type `T = *`, and still satisfy `ConstraintSet.always()`. The constraint of `Base` only satisfies the constraint set when `T = Base`. The intersection of those two solutions is `T = Base`, and so that's what we return.

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: Yep, this one already has `Any` as the TODO

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: Added TODO

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: Added TODO

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: Added TODO

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: (This one I think the right TODO is `Any` not `Unknown` because here we are now explicitly saying that `T` can be any type, as opposed to not having said anything about what `T` can be. The latter clearly should fall back on the default. The former I don't think should.)

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
carljm: Yes, that makes sense! So if we have a lower bound of `Base` (e.g. because we got a call that provided an instance of `Base` as argument for this typevar), then we can solve to `Base` because it must be the most precise option.

PR #21540, opened 2025-11-20
[`pyflakes`] Treat function-scope bare annotations as locals per PEP 526 (`F821`)

### LINKED ISSUE #21494: `F821`: Declared only variables
### Summary

Local variable annotations make annotated variables local even if there is no direct assignment. This is useful to create a local variable without an initial value. Ruff appears to not understand this usage. For example:
```python
from typing import Callable

def make_cell() -> tuple[Callable[[], int], Callable[[int], None]]:
    x: int

    def getter() -> int:
        return x

    def setter(xx: int) -> None:
        nonlocal x
        x = xx

    return getter, setter

getter, setter = make_cell()
setter(123)
print(getter())
```
(playground link: https://play.ruff.rs/3bd90cac-2c87-48b0-9cb4-64fe2a53bf72)

This runs fine and passes mypy, however ruff produces a `F821`:
```
$ ruff check --isolated t.py
F821 Undefined name `x`
 --> t.py:7:16
  |
6 |     def getter() -> int:
7 |         return x
  |                ^
8 |
9 |     def setter(xx: int) -> None:
```

### Version

0.14.5

Ruchir28:
## Summary

Resolves #21494 

Treat function-scope annotated variables as locals per PEP 526 , preventing F821 false positives.

## Test Plan

Added tests for these cases in `crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py`

### DIFF
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Callable
+
+
+def demonstrate_bare_local_annotation():
+    x: int
+    print(x)
+
+demonstrate_bare_local_annotation()
+
+
+def make_closure_pair() -> tuple[Callable[[], int], Callable[[int], None]]:
+    x: int
+
+    def get_value() -> int:
+        return x
+
+    def set_value(new_value: int) -> None:
+        nonlocal x
+        x = new_value
+
+    return get_value, set_value
+
+get_value, set_value = make_closure_pair()
+set_value(123)
+print(get_value())
--- a/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py
+++ b/crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py
@@ -3,13 +3,23 @@ from __future__ import annotations
 from typing import Callable
 
 
+# Error: bare annotation without assignment in same scope.
 def demonstrate_bare_local_annotation():
     x: int
     print(x)
 
-demonstrate_bare_local_annotation()
+
+# Error: bare annotation read from closure without nonlocal.
+def demonstrate_closure_without_nonlocal() -> Callable[[], int]:
+    x: int
+
+    def get_value() -> int:
+        return x
+
+    return get_value
 
 
+# OK: nonlocal declaration exists, so the annotation may be initialized.
 def make_closure_pair() -> tuple[Callable[[], int], Callable[[int], None]]:
     x: int
 
@@ -22,6 +32,14 @@ def make_closure_pair() -> tuple[Callable[[], int], Callable[[int], None]]:
 
     return get_value, set_value
 
-get_value, set_value = make_closure_pair()
-set_value(123)
-print(get_value())
+
+# OK: nonlocal declaration exists, outer scope reads after inner scope may assign.
+def demonstrate_nonlocal_rebinding_then_outer_read() -> int:
+    x: int
+
+    def set_value(new_value: int) -> None:
+        nonlocal x
+        x = new_value
+
+    set_value(1)
+    return x
--- a/crates/ruff_python_semantic/src/model.rs
+++ b/crates/ruff_python_semantic/src/model.rs
@@ -476,9 +476,19 @@ impl<'a> SemanticModel<'a> {
                     // The `name` in `print(name)` should be treated as unresolved, but the `name` in
                     // `name: str` should be treated as used.
                     //
-                    // Stub files are an exception. In a stub file, it _is_ considered valid to
-                    // resolve to a type annotation.
-                    BindingKind::Annotation if !self.in_stub_file() => continue,
+                    // There are two exceptions to this rule:
+                    // 1. Stub files. In a stub file, it _is_ considered valid to resolve to a
+                    //    type annotation.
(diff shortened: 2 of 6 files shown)


crates/ruff_linter/resources/test/fixtures/pyflakes/F821_34.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Callable
+
+
+def demonstrate_bare_local_annotation():
+    x: int
+    print(x)
```
ntBre, line 9: I still haven't looked too closely at this, but I agree that it's unfortunate not to flag this case. It might be worth looking into how ty handles this since it seems to get the expected result in both cases:

https://play.ty.dev/c6f45910-a2bd-4745-827b-f29a148f4e63

It doesn't seem to be related to `nonlocal`, though, because commenting it out doesn't change ty's diagnostics.

Ruchir28: One option to do this is then something like 
```
                    BindingKind::Annotation if !self.in_stub_file() => {
                        if self.scopes[self.bindings[binding_id].scope]
                            .kind
                            .is_function()
                        {
                            // If we're in a function scope, and the annotation is local to that scope,
                            // treat it as unbound (since it's not assigned).
                            // If it's an enclosing scope, we treat it as resolved (to support closures).
                            if index == 0 {
                                self.unresolved_references.push(
                                    name.range,
                                    self.exceptions(),
                                    UnresolvedReferenceFlags::empty(),
                                );
                                return ReadResult::UnboundLocal(binding_id);
                            }
                            
                            self.resolved_names.insert(name.into(), binding_id);
                            return ReadResult::Resolved(binding_id);
                        
                        }
                        continue;
                    }
```

if we do this then both the cases will be handled i.e. error in this case 

```
def demonstrate_bare_local_annotation():
    x: int
    print(x)
```

and no error in the closure cases i.e. the setter,getter example 


but then again we will have edge cases like these where no error will be reported [although even ty doesn't report them]

```
def demo_closure():
    x: int
    def func():
        return x
```

What do you think about this

ntBre: Thanks for looking into this, and sorry for the delay! I've looked at this a bit today, and yes, I think your suggestion here is a good idea that should handle both the existing test case and the new closure case from the issue. I'm okay with the false negative in `demo_closure` as it seems like a better trade-off.

Ruchir28: Hey @ntBre, I have a different approach for this, which also handles the edge case. Instead of resolving closure reads as Resolved, both same-scope and closure reads of bare function-scope annotations are treated as unresolved.

  To handle the nonlocal case, unresolved references now carry an annotation_binding_id. In the F821 lint phase, we use rebinding_scopes to check if any nested scope declared nonlocal on that annotation binding — if so, we suppress F821.

  F821 - UnboundLocalError at runtime

```
  def f():
      x: int
      print(x)

```
  No F821 - nonlocal rebinding exists (getter/setter pattern)

```
  def make_closure_pair():
      x: int
      def get_value():
          return x
      def set_value(new_value):
          nonlocal x
          x = new_value
      return get_value, set_value

```
  F821 — no nonlocal, previously a false negative

```
  def demo_closure():
      x: int
      def func():
          return x
```

  This eliminates the demo_closure false negative from the previous approach. The only heuristic is that we trust nonlocal presence without verifying an actual assignment -  matching Pyflakes' behavior.

  Have updated the PR.

crates/ruff_python_semantic/src/model.rs
```diff
@@ -475,9 +476,29 @@ impl<'a> SemanticModel<'a> {
                     // The `name` in `print(name)` should be treated as unresolved, but the `name` in
                     // `name: str` should be treated as used.
                     //
-                    // Stub files are an exception. In a stub file, it _is_ considered valid to
-                    // resolve to a type annotation.
-                    BindingKind::Annotation if !self.in_stub_file() => continue,
+                    // There is one exception to this rule:
+                    // 1. Stub files. In a stub file, it _is_ considered valid to resolve to a
+                    //    type annotation.
```
ntBre: I think this part of the comment would be fine to leave as-is. However, I think it would be helpful to update the earlier part of the comment now that type annotations aren't always treated as unresolved. In other words, I think I'd try to combine the comment on line 487 with the comment on 469.

ntBre (approved):
Thank you and so sorry for the long delay in review! This seems like quite an elegant fix. I just left one suggestion about updating a comment that I think is now slightly out of sync with the implementation. I'm happy to take care of that and the merge conflicts if you want, though.

crates/ruff_python_semantic/src/model.rs
Ruchir28: No worries at all, thanks for the review! 

Since you offered, I'd prefer if you took care of the comment update and the merge conflicts , I'm a little unclear on the framing myself (even in the function-scope case we're still treating the reference as unresolved, right? the only difference being that we push the binding_id along so a later nonlocal in a nested scope can suppress it), so I'd rather leave the wording to you. 

Appreciate it!

Ruchir28: Actually went ahead and took care of both - updated the comment and resolved the conflicts! Feel free to tweak the wording before merging

ntBre: Awesome, thank you!

PR #21541, opened 2025-11-20
[ty] Add more TypeDetails to the display code

Gankra:
As far as I know this change is largely non-functional, largely because of https://github.com/astral-sh/ty/issues/1601

It's possible some of these like `Type::KnownInstance` produce something useful sometimes. `LiteralString` is a new introduction, although its goto-type jumps to `str` which is a bit sad (considering that part of the SpecialForm discourse for now).

Also wrt the generics testing followup: turns out the snapshot tests were full of those already.

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -4144,6 +4144,43 @@ mod tests {
     }
 
     #[test]
+    fn test_literal_string() {
+        let mut test = inlay_hint_test(
+            r#"
+            from typing import LiteralString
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> Also wrt the generics testing followup: turns out the snapshot tests were full of those already.

Hmm, really? I was asking specifically about generic-alias and generic subclass-of types in https://github.com/astral-sh/ruff/pull/21533#discussion_r2546190060. Those are both different variants to generic nominal-instance types, and I couldn't immediately see any snapshots for them

Gankra:
Ah I see, yes those are probably missing.

PR #21546, opened 2025-11-20
[ty] Add more random TypeDetails and tests

Gankra:


crates/ty_python_semantic/src/types/display.rs
```diff
@@ -681,32 +680,48 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
-                class.display_with(self.db, self.settings.clone())
-            ),
-            Type::GenericAlias(generic) => write!(
-                f.with_detail(TypeDetail::Type(self.ty)),
-                "<class '{}'>",
-                generic.display_with(self.db, self.settings.singleline())
-            ),
+            Type::ClassLiteral(class) => {
+                let mut f = f.with_detail(TypeDetail::Type(self.ty));
+                f.write_str("<class '")?;
```
Gankra, line 675: This is theoretically useless because the span logic will prefer the inner `class.display_with` but this is effectively a fallback if it doesn't.

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -681,32 +680,48 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
                         .fmt_detailed(f)?;
-                    f.write_str("]")
+                    f.write_char(']')
+                }
+                SubclassOfInner::Dynamic(dynamic) => {
+                    f.with_detail(TypeDetail::Type(Type::SpecialForm(SpecialFormType::Type)))
+                        .write_str("type")?;
+                    f.write_char('[')?;
+                    write!(f, "{dynamic}")?;
+                    f.write_char(']')
```
Gankra, line 715: Should this invoke the Any/Unknown special cases? (Should that logic be factored out?)

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -681,32 +680,48 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
+                generic
+                    .display_with(self.db, self.settings.clone())
+                    .fmt_detailed(&mut f)?;
+                f.write_str("'>")
+            }
             Type::SubclassOf(subclass_of_ty) => match subclass_of_ty.subclass_of() {
                 SubclassOfInner::Class(ClassType::NonGeneric(class)) => {
-                    f.write_str("type[")?;
+                    f.with_detail(TypeDetail::Type(Type::SpecialForm(SpecialFormType::Type)))
+                        .write_str("type")?;
```
AlexWaygood: I'm actually not totally sold that double-clicking on `type` in `type[int]` should take you anywhere, because `type` isn't a generic class in the same way as `list[int]` or `dict[str, str]`. Though it is true that all `type[]` types are instances of `type`

If we _do_ want it to take you somewhere, though, it should take you to `builtins.type` rather than `typing.Type`, so that would be

```suggestion
                    f.with_detail(TypeDetail::Type(KnownClass::Type.to_class_literal(self.db)))
                        .write_str("type")?;
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -681,32 +680,48 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
+                    f.write_char('[')?;
                     class
                         .display_with(self.db, self.settings.clone())
                         .fmt_detailed(f)?;
-                    f.write_str("]")
+                    f.write_char(']')
                 }
                 SubclassOfInner::Class(ClassType::Generic(alias)) => {
-                    f.write_str("type[")?;
+                    f.with_detail(TypeDetail::Type(Type::SpecialForm(SpecialFormType::Type)))
```
AlexWaygood: same here

```suggestion
                    f.with_detail(TypeDetail::Type(KnownClass::Type.to_class_literal(self.db)))
```

crates/ty_python_semantic/src/types/display.rs
```diff
@@ -681,32 +680,48 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
+                        .write_str("type")?;
+                    f.write_char('[')?;
                     alias
                         .display_with(self.db, self.settings.clone())
                         .fmt_detailed(f)?;
-                    f.write_str("]")
+                    f.write_char(']')
+                }
+                SubclassOfInner::Dynamic(dynamic) => {
+                    f.with_detail(TypeDetail::Type(Type::SpecialForm(SpecialFormType::Type)))
```
AlexWaygood: and here

```suggestion
                    f.with_detail(TypeDetail::Type(KnownClass::Type.to_class_literal(self.db)))
```

crates/ty_python_semantic/src/types/display.rs
AlexWaygood: > Should this invoke the Any/Unknown special cases?

I think so, yes!

> Should that logic be factored out?

Sure!

AlexWaygood: in fact, maybe that logic should just be moved to `Type::definition()` in `types.rs`?

PR #21547, opened 2025-11-20
[ty] Improve import detection for completions and support `from ...<CURSOR>` completions

BurntSushi:
This PR has three main changes in it:

* A small refactor of how we collect completions. This centralizes
some logic and made other refactoring easier.
* All of the old piecemeal import detection heuristics have been
replaced with a giant token-based state machine that is intended to
detect all kinds of (possibly incomplete) import statements.
* We build on this new infrastructure to support module completions
for relative imports.

This is best reviewed commit-by-commit.

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -646,7 +646,16 @@ struct FromImport<'a> {
 #[derive(Clone, Debug)]
 enum FromImportKind {
     Module,
-    Submodule { parent: ModuleName },
+    Submodule {
+        parent: ModuleName,
+    },
+    Relative {
+        parent: ModuleName,
+        /// When `true`, an `import` keyword is allowed next.
+        /// For example, `from ...<CURSOR>` should offer `import`
+        /// but also submodule completions.
+        import_keyword_allowed: bool,
+    },
     Attribute,
 }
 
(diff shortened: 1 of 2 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/13c06b9a-bf99-43d9-969d-2ed745cfdb15

crates/ty_ide/src/completion.rs
```diff
@@ -678,6 +619,574 @@ struct ScopedTarget<'t> {
...
+            /// A name seen immediately after the first name. This
+            /// indicates we have an incomplete import statement like
+            /// `import foo a<CURSOR>` or `from foo imp<CURSOR>`. But
+            /// we mush on.
+            AdjacentName,
+
+            /// A state where we expect to see the start of or
+            /// continuation of a list of names following `import`.
+            /// In the grammar, this is either `dotted_as_names`
+            /// or `import_from_as_names`.
```
```suggestion
            /// In the [grammar], this is either `dotted_as_names`
            /// or `import_from_as_names`.
            ///
            /// [grammar]: https://docs.python.org/3/reference/grammar.html
```

crates/ty_ide/src/completion.rs
```diff
@@ -678,6 +619,574 @@ struct ScopedTarget<'t> {
...
+            /// token. i.e., `from name import` or `from ...name import`.
+            FromDottedName,
+            /// Occurs when we've seen an `import` followed by a
+            /// name-like token with a dot. i.e., `from .name import`
+            /// or `from ..name import`.
+            FromDottedNameDot,
+            /// A `*` was just seen, which must mean the import is of
+            /// the form `from module import *`.
+            FromStar,
+            /// A left paranthesis was just seen.
```
```suggestion
            /// A left parenthesis was just seen.
```

AlexWaygood:
One edge case I spotted while playing around: here, I'd expect module names such as `builtins` to be suggested, but auto-import suggestions are given instead: 
image

BurntSushi:
> The "backwards munching through the tokens" state machine scares me a bit 😆 but this is awesome, thank you

Yeah reverse state machines tend to be more complex in my experience. I think it's because we (humans) usually design grammars thinking about parsing them in the forward direction. But the reverse usually requires more states. (This is true in `regex` too, which compiles a forward _and_ reverse state machine for regex patterns. The reverse state machine is almost always bigger and more complicated.)

BurntSushi:
> One edge case I spotted while playing around: here, I'd expect module names such as `builtins` to be suggested, but auto-import suggestions are given instead:

Nice find! A bug in the state machine. :-) Should be fixed now.

crates/ty_ide/src/completion.rs
```diff
@@ -21,6 +21,91 @@ use crate::importer::{ImportRequest, Importer};
...
+    /// sequence of completions.
+    fn into_completions(mut self) -> Vec<Completion<'db>> {
+        self.items.sort_by(compare_suggestions);
+        self.items
+            .dedup_by(|c1, c2| (&c1.name, c1.module_name) == (&c2.name, c2.module_name));
+        self.items
+    }
+
+    /// Always adds the given completion to this collection.
+    fn add(&mut self, completion: Completion<'db>) {
```
MichaReiser: I would do the opposite here and rename add to force_add or similar and rename try_add to just add because

* try_add is what should be used in most places. If in doubt, use try_add. However, try_add is harder to discover, I'm more likely to call add
* it feels inconsistent that extend uses try_add.

crates/ty_ide/src/completion.rs
```diff
@@ -678,6 +619,577 @@ struct ScopedTarget<'t> {
...
+        let mut start = end;
+        for token in tokens.iter().rev().take(Self::LIMIT) {
+            match token.kind() {
+                TK::Name | TK::Dot | TK::Ellipsis => {
+                    start = token.range().start();
+                }
+                _ => break,
+            }
+        }
+        to_complete.push_str(&source[start.to_usize()..end.to_usize()]);
```
MichaReiser: You could consider using TextRange::new(start, end) as slice argument instead of calling to_usize on both start and end

crates/ty_ide/src/completion.rs
BurntSushi: I agree that `extend` using `try_add` is perhaps not great. Maybe it would be better to add a concrete `try_extend` and remove the `Extend` impl.

I do like keeping `try_add` though, because I think using `add` could be too misleading since it might not actually add the completion.

With that said, I'd be okay with renaming `add` to `force_add`. So we'd have `try_add` and `force_add`.

PR #21548, opened 2025-11-20
[ty] More low-hanging fruit for inlay hint goto-definition

AlexWaygood:
## Summary

Enable users to jump to `int` from the inlay for a variable inferred as `Literal[42]`, and some other easy things

## Test Plan

snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types/display.rs
+++ b/crates/ty_python_semantic/src/types/display.rs
@@ -23,9 +23,9 @@ use crate::types::signatures::{CallableSignature, Parameter, Parameters, Signatu
 use crate::types::tuple::TupleSpec;
 use crate::types::visitor::TypeVisitor;
 use crate::types::{
-    BoundTypeVarIdentity, CallableType, DynamicType, IntersectionType, KnownBoundMethodType,
-    KnownClass, MaterializationKind, Protocol, ProtocolInstanceType, SpecialFormType,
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/src/types/display.rs
```diff
@@ -847,18 +848,33 @@ impl<'db> FmtDetailed<'db> for DisplayRepresentation<'db> {
...
+                )
             }
+            // an alternative would be to use `Type::SpecialForm(SpecialFormType::LiteralString)` here,
+            // which would mean users would be able to jump to the definition of `LiteralString` from the
+            // inlay hint, but that seems less useful than the definition of `str` for a variable that is
+            // inferred as an *inhabitant* of `LiteralString` (since that variable will just be a string
+            // at runtime)
             Type::LiteralString => f
                 .with_detail(TypeDetail::Type(self.ty))
                 .write_str("LiteralString"),
```
Gankra, line 869: Hmm I think I disagree? From what I've seen LiteralString only shows up in code that's explicitly declared e.g. a function argument as LiteralString, but maybe I've just been lucky? (not a strong opinion)

AlexWaygood: hmm, you could be hovering over `os.name` or some other constant that's annotated (but not by you) as `LiteralString`? I don't have a strong opinion either; we can easily change this in response to user feedback if we get some!

PR #21549, opened 2025-11-20
[ty] suppress autocomplete suggestions during variable binding

RasmusNygren:
Statements such as `def foo(p<CURSOR>`,
`def foo[T<CURSOR>` and `for foo<CURSOR>`
should not generate any suggestions as these
cases are introducing new names.

If it's not possible to determine that suggestions should be omitted using token matching in an easy way, we turn
to traversing the AST to determine the context.

## Summary

Fixes https://github.com/astral-sh/ty/issues/1563

It keeps using the existing token matching pattern for the easy cases (nothing typed and most recent token is a definition token) and fallbacks to AST traveral for the slightly more difficult cases where token matching becomes difficult and error prone.

## Test Plan
New test cases and sanity-checking in the ty playground


BurntSushi:
Thank you!!!

PR #21550, opened 2025-11-20
[ty] Implement docstring rendering to markdown

Gankra:
## Summary

This introduces a very bad and naive python-docstring-flavoured-reStructuredText to github-flavor-markdown translator. The main goal is to try to preserve a lot of the formatting and plaintext, progressively enhance the content when we find things we know about, and escape the text when we find things that might get corrupt.

Previously I'd broken this out into rendering each different format, but with this approach you don't really need to?

## Test Plan

Lots of snapshot tests, also messed around in some random stdlib modules.

### DIFF
--- a/crates/ty_ide/src/docstring.rs
+++ b/crates/ty_ide/src/docstring.rs
@@ -269,6 +269,8 @@ fn render_markdown(docstring: &str) -> String {
         for _ in 0..line_indent {
             // If we're not in a codeblock use non-breaking spaces to preserve the indent
             if !in_any_code {
+                // TODO: would the raw unicode codepoint be handled *better* or *worse*
+                // by various IDEs? VS Code handles this approach well, at least.
                 output.push_str("&nbsp;");
             } else {
                 output.push(' ');


crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+    // and (possibly in a higher up piece of logic) try to resolve the names for
+    // cross-linking. (Similar to `TypeDetails` in the type formatting code.)
+    let mut output = String::new();
+    let mut first_line = true;
+    let mut block_indent = 0;
+    let mut in_doctest = false;
+    let mut starting_literal = false;
+    let mut in_literal = false;
+    let mut in_any_code = false;
+    for untrimmed_line in docstring.lines() {
```
Gankra, line 197: This is the worst parser I've written in a long while so if you tell me "Aria for the love of god make a Parser type with methods and enums" I will. Writing it this way was conducive to experimentation and thinking about the format.

Gankra:
I spent a while salivating over the ruff docstring formatter as a basis for this, but I ultimately concluded it would be a nightmare for both codebases to try to share that logic, and it's not *that* much logic.

Gankra:
The main risk of shipping this implementation is that the previous implementation was so wildly conservative that it "couldn't" ever mess up whereas insufficient escaping or getting confused could complete garble things and eat the docs.

On balance though this just looks so much better I think the risk is worth it (and shipping it will get me a lot of bug reports about what to fix :)

crates/ty_ide/src/docstring.rs
MichaReiser, line 197: You want to use `.universal_newlines` here to get proper `\r` support

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
     output
 }
 
+/// Given a cleaned up docstring, render it to markdown.
+///
+/// This is by no means perfect but it handles a lot of common formats well.
+///
+/// This code currently assumes docstrings are reStructuredText with significant
+/// line-breaks and indentation. Some constructs like *italic*, **bold**, and `inline code`
+/// are shared between the two formats. Other things like _italic_ are only markdown
```
MichaReiser: What are the two formats that you're referring to here?

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+                // Delete whole line
+                line = trimmed_without_lit;
+            }
+            starting_literal = true;
+        }
+
+        // Add this line's indentation.
+        // We could subtract the block_indent here but in practice it's uglier
+        for _ in 0..line_indent {
+            // If we're not in a codeblock use non-breaking spaces to preserve the indent
```
MichaReiser, line 270: Can you say why this is necessary, given that it gets rendered as markdown?

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+            } else {
+                // Delete whole line
+                line = trimmed_without_lit;
+            }
+            starting_literal = true;
+        }
+
+        // Add this line's indentation.
+        // We could subtract the block_indent here but in practice it's uglier
+        for _ in 0..line_indent {
```
MichaReiser, line 269: This won't work for code using tab indentation (which our formatter supports).

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
     output
 }
 
+/// Given a cleaned up docstring, render it to markdown.
+///
+/// This is by no means perfect but it handles a lot of common formats well.
+///
+/// This code currently assumes docstrings are reStructuredText with significant
```
MichaReiser: What happens for docstrings using the numpy or google format? will they still be rendered correctly?

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+    let mut in_literal = false;
+    let mut in_any_code = false;
+    for untrimmed_line in docstring.lines() {
+        // We can assume leading whitespace has been normalized
+        let mut line = untrimmed_line.trim_start_matches(' ');
+        let line_indent = untrimmed_line.len() - line.len();
+
+        // First thing's first, add a newline to start the new line
+        if !first_line {
+            // If we're not in a codeblock, add trailing space to the line to authentically wrap it
```
MichaReiser, line 204: I don't think I understand this. Why is it necessary to add trailing whitespace to any line? This seems very undesired?

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
 
+/// Given a cleaned up docstring, render it to markdown.
+///
+/// This is by no means perfect but it handles a lot of common formats well.
+///
+/// This code currently assumes docstrings are reStructuredText with significant
+/// line-breaks and indentation. Some constructs like *italic*, **bold**, and `inline code`
+/// are shared between the two formats. Other things like _italic_ are only markdown
+/// and should be escaped.
+fn render_markdown(docstring: &str) -> String {
```
MichaReiser, line 184: Let's add a link to [reStructuredText](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+        }
+
+        // We previously entered a literal block and we just found our first non-blank line
+        // So now we're actually in the litreal
+        if starting_literal && !line.is_empty() {
+            starting_literal = false;
+            in_literal = true;
+            in_any_code = true;
+            block_indent = line_indent;
+            // TODO: should we not be this aggressive? Let it autodetect?
```
MichaReiser, line 233: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

Defaulting to Python doesn't seem unreasonable but we should respect any override

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+        if !in_any_code && line.starts_with(">>>") {
+            block_indent = line_indent;
+            in_doctest = true;
+            in_any_code = true;
+            // TODO: is there something more specific? `pycon`?
+            output.push_str("```python\n");
+        }
+
+        // If we're not in a codeblock and we see something that signals a literal block, start one
+        if !in_any_code && let Some(without_lit) = line.strip_suffix("::") {
```
MichaReiser, line 249: I'd have to play with restructuredText but the way I read this paragraph

> Literal code blocks ([ref](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks)) are introduced by ending a paragraph with the special marker ::. The literal block must be indented

is that what comes after is only a literal block if there's at least one blank line, and text is indented. 

I think your code correctly handles the empty line. But it ends up pushing an empty markdown code block if the next paragraph isn't correctly indented.

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+    // While rendering this we should make note of all the `singletick` locations
+    // and (possibly in a higher up piece of logic) try to resolve the names for
+    // cross-linking. (Similar to `TypeDetails` in the type formatting code.)
+    let mut output = String::new();
+    let mut first_line = true;
+    let mut block_indent = 0;
+    let mut in_doctest = false;
+    let mut starting_literal = false;
+    let mut in_literal = false;
+    let mut in_any_code = false;
```
MichaReiser, line 196: I'm not sure about mixing the parsing of docstrings with the formatting logic. It seems harder to follow. I'd be inclined to parse the docstring into an intermediate representation and then do a second pass to format it.

crates/ty_ide/src/docstring.rs
Gankra: I need to signpost that better in the comments but this function assumes `documentation_trim` (in the same file) was already run, so our leading whitespace is only spaces, and our newlines are only `\n` (and unlike ruff we don't need to care about authentically preserving the flavour of whitespace).

crates/ty_ide/src/docstring.rs
Gankra: reStructuredText and markdown

crates/ty_ide/src/docstring.rs
Gankra: Yes so the idea here is that docstrings really like using indentation for structure that we do not (care to) parse but we also don't want markdown turning into code-blocks. By replacing leading whitespace with `&nbsp;` outside of codeblocks, we preserve the native indent of the docstring and stuff we don't understand still is rendered clearly and pretty.
 
Screenshot 2025-11-21 at 7 40 49 AM

crates/ty_ide/src/docstring.rs
Gankra: (another thing handled by `documentation_trim`)

crates/ty_ide/src/docstring.rs
MichaReiser: Oh, that's not clear to me from the comment

crates/ty_ide/src/docstring.rs
Gankra: Screenshot 2025-11-21 at 7 44 21 AM

Screenshot 2025-11-21 at 7 44 48 AM

Screenshot 2025-11-21 at 7 45 11 AM

crates/ty_ide/src/docstring.rs
Gankra: It's part of the approach I'm taking where I want to preserve the wrapping/indentation of the original docstring.

It prevents e.g. `param1` and `param2` from being merged together as a single line in:

```
'param1' -- The first parameter description
'param2' -- The second parameter description
            This is a continuation of param2 description.
'param3' -- A parameter without type annotation
```

crates/ty_ide/src/docstring.rs
Gankra: Haha yes you caught me this part of the parser is distinctly janky. I'll tweak it to make it Spec Compliant (although I can't help but wonder if being sloppy will be the name of the game since we need to render whatever we find).

crates/ty_ide/src/docstring.rs
Gankra: Dang, ok if the captain of the performance team wants more intermediate allocations and representations who am I to complain 😄 

(I was half expecting to be told to remove some intermediate Strings)

crates/ty_ide/src/docstring.rs
```diff
@@ -153,6 +149,128 @@ fn documentation_trim(docs: &str) -> String {
...
+                output.push_str("&nbsp;");
+            } else {
+                output.push(' ');
+            }
+        }
+
+        if !in_any_code {
+            // This is plain text, apply some escapes for reST that don't map to markdown:
+            // * underscores aren't used for italics/bold in reST (don't mess up __dunders__)
+            output.push_str(&line.replace('_', "\\_"));
```
Gankra: Found an in-retrospect-obvious place where this is broken: this mangles inline code `__init__` (I kept seeing it without any code markup, so that was my primary concern)

crates/ty_ide/src/docstring.rs
MichaReiser: Lol. I'm not too concerned here, given that we don't render markdown strings that frequently and having a more structured representation would ultimately be useful for Ruff too. Although I don't think we should aim to make it reusable for now.

crates/ty_ide/src/docstring.rs
Gankra: (Even if I had a perfect parser for this format, the only clear way to render this to markdown, imo, is by rendering these blocks as indented)

MichaReiser: Oh, I think I misread it as *in a code block*. This makes a ton of sense for indents outside code blocks.

MichaReiser: I wonder if we should use the unicode code point directly so that we don't rely on the markdown renderer supporting HTML?

PR #21551, opened 2025-11-20
[ty] Infer typevar specializations for `Callable` types

dcreager:
This is a first stab at solving https://github.com/astral-sh/ty/issues/500, at least in part, with the old solver. We add a new `TypeRelation` that lets us opt into using constraint sets to describe when a typevar is assignability to some type, and then use that to calculate a constraint set that describes when two callable types are assignable. If the callable types contain typevars, that constraint set will describe their valid specializations. We can then walk through all of the ways the constraint set can be satisfied, and record a type mapping in the old solver for each one.

### DIFF
--- a/crates/ty_python_semantic/src/types/generics.rs
+++ b/crates/ty_python_semantic/src/types/generics.rs
@@ -1937,15 +1937,30 @@ impl<'db> SpecializationBuilder<'db> {
                     return Ok(());
                 };
 
+                let formal_callable = formal_callable.signatures(self.db);
+                let formal_is_single_paramspec = formal_callable.is_single_paramspec().is_some();
+
                 for actual_callable in actual_callables.as_slice() {
-                    let when = actual_callable
-                        .signatures(self.db)
-                        .when_constraint_set_assignable_to(
-                            self.db,
-                            formal_callable.signatures(self.db),
-                            self.inferable,
-                        );
-                    self.add_type_mappings_from_constraint_set(formal, when, &mut f);
+                    if formal_is_single_paramspec {
+                        let when = actual_callable
+                            .signatures(self.db)
+                            .when_constraint_set_assignable_to(
+                                self.db,
+                                formal_callable,
+                                self.inferable,
+                            );
+                        self.add_type_mappings_from_constraint_set(formal, when, &mut f);
+                    } else {
+                        for actual_signature in &actual_callable.signatures(self.db).overloads {
+                            let when = actual_signature
+                                .when_constraint_set_assignable_to_signatures(
+                                    self.db,
+                                    formal_callable,
+                                    self.inferable,
+                                );
+                            self.add_type_mappings_from_constraint_set(formal, when, &mut f);
+                        }
+                    }
                 }
             }
 
--- a/crates/ty_python_semantic/src/types/signatures.rs
+++ b/crates/ty_python_semantic/src/types/signatures.rs
@@ -336,6 +336,12 @@ impl<'db> CallableSignature<'db> {
         )
     }
 
+    pub(crate) fn is_single_paramspec(
+        &self,
+    ) -> Option<(BoundTypeVarInstance<'db>, Option<Type<'db>>)> {
+        Self::signatures_is_single_paramspec(&self.overloads)
+    }
+
     /// Checks whether the given slice contains a single signature, and that signature is a
     /// `ParamSpec` signature. If so, returns the [`BoundTypeVarInstance`] for the `ParamSpec`,
     /// along with the return type of the signature.
@@ -388,11 +394,6 @@ impl<'db> CallableSignature<'db> {
             // the other callable's signature. We also need to compare the return types — for
             // instance, to verify in `Callable[P, int]` that the return type is assignable to
             // `int`, or in `Callable[P, T]` to bind `T` to the return type of the other callable.
-            //
-            // TODO: This logic might need to move down into `Signature`, if we need paramspecs to
-            // be able to match a _subset_ of an overloaded callable. (In that case, we need to
-            // check each signature individually, and combine together the ones that match into the
-            // overloaded callable that the paramspec binds to.)
             match (self_is_single_paramspec, other_is_single_paramspec) {
                 (
                     Some((self_bound_typevar, self_return_type)),
@@ -1095,6 +1096,65 @@ impl<'db> Signature<'db> {
         result
     }
 
+    pub(crate) fn when_constraint_set_assignable_to_signatures(
+        &self,
+        db: &'db dyn Db,
+        other: &CallableSignature<'db>,
+        inferable: InferableTypeVars<'_, 'db>,
+    ) -> ConstraintSet<'db> {
+        // If this signature is a paramspec, bind it to the entire overloaded other callable.
+        if let Some(self_bound_typevar) = self.parameters.as_paramspec()
+            && other.is_single_paramspec().is_none()
+        {
+            let upper = Type::Callable(CallableType::new(
+                db,
+                CallableSignature::from_overloads(
+                    other
+                        .overloads
+                        .iter()
+                        .map(|signature| Signature::new(signature.parameters().clone(), None)),
+                ),
+                CallableTypeKind::ParamSpecValue,
+            ));
+            let param_spec_matches =
+                ConstraintSet::constrain_typevar(db, self_bound_typevar, Type::Never, upper);
+            let return_types_match = self.return_ty.when_some_and(|self_return_type| {
+                other
+                    .overloads
+                    .iter()
+                    .filter_map(|signature| signature.return_ty)
+                    .when_any(db, |other_return_type| {
+                        self_return_type.when_constraint_set_assignable_to(
+                            db,
+                            other_return_type,
+                            inferable,
+                        )
+                    })
+            });
+            return param_spec_matches.and(db, || return_types_match);
+        }
+
+        other.overloads.iter().when_all(db, |other_signature| {
+            self.when_constraint_set_assignable_to(db, other_signature, inferable)
+        })
+    }
+
+    fn when_constraint_set_assignable_to(
+        &self,
+        db: &'db dyn Db,
+        other: &Self,
+        inferable: InferableTypeVars<'_, 'db>,
+    ) -> ConstraintSet<'db> {
+        self.has_relation_to_impl(
+            db,
+            other,
+            inferable,
+            TypeRelation::ConstraintSetAssignability,
+            &HasRelationToVisitor::default(),
+            &IsDisjointVisitor::default(),
+        )
+    }
+
     /// Implementation of subtyping and assignability for signature.
     fn has_relation_to_impl(
         &self,
@@ -1281,6 +1341,60 @@ impl<'db> Signature<'db> {
                 ),
             );
             return result;
+        }
+
+        if relation.is_constraint_set_assignability() {
+            let self_is_paramspec = self.parameters.as_paramspec();
+            let other_is_paramspec = other.parameters.as_paramspec();
+
+            // If either signature is a ParamSpec, the constraint set should bind the ParamSpec to
+            // the other signature.
+            match (self_is_paramspec, other_is_paramspec) {
+                (Some(self_bound_typevar), Some(other_bound_typevar)) => {
+                    let param_spec_matches = ConstraintSet::constrain_typevar(
+                        db,
+                        self_bound_typevar,
+                        Type::TypeVar(other_bound_typevar),
+                        Type::TypeVar(other_bound_typevar),
+                    );
+                    result.intersect(db, param_spec_matches);
+                    return result;
+                }
+
+                (Some(self_bound_typevar), None) => {
+                    let upper = Type::Callable(CallableType::new(
+                        db,
+                        CallableSignature::single(Signature::new(other.parameters.clone(), None)),
+                        CallableTypeKind::ParamSpecValue,
+                    ));
+                    let param_spec_matches = ConstraintSet::constrain_typevar(
+                        db,
+                        self_bound_typevar,
+                        Type::Never,
+                        upper,
+                    );
+                    result.intersect(db, param_spec_matches);
+                    return result;
+                }
+
+                (None, Some(other_bound_typevar)) => {
+                    let lower = Type::Callable(CallableType::new(
+                        db,
+                        CallableSignature::single(Signature::new(self.parameters.clone(), None)),
+                        CallableTypeKind::ParamSpecValue,
+                    ));
+                    let param_spec_matches = ConstraintSet::constrain_typevar(
+                        db,
+                        other_bound_typevar,
+                        lower,
+                        Type::object(),
+                    );
+                    result.intersect(db, param_spec_matches);
+                    return result;
+                }
+
+                (None, None) => {}
+            }
         }
 
         let mut parameters = ParametersZip {


crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -379,14 +379,13 @@ T = TypeVar("T")
 def invoke(fn: Callable[[A], B], value: A) -> B:
     return fn(value)
 
-def identity(x: T) -> T:
+def identity(x: T, /) -> T:
```
dcreager: This is an interesting nuance: `Callable[[A], B]` creates a signature containing positional-only parameters, so we have to make the signature here positional-only as well to make `identity` pass the assignability check for it to be a valid argument to the `fn` parameter.

(That said, I'm not convinced that's...correct? Are we mixing up the ordering of the assignability check operands somewhere?)

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
```diff
@@ -379,14 +379,13 @@ T = TypeVar("T")
...
-def head(xs: list[T]) -> T:
+def head(xs: list[T], /) -> T:
     return xs[0]
 
-# TODO: this should be `Literal[1]`
-reveal_type(invoke(identity, 1))  # revealed: Unknown
+reveal_type(invoke(identity, 1))  # revealed: Literal[1]
 
 # TODO: this should be `Unknown | int`
 reveal_type(invoke(head, [1, 2, 3]))  # revealed: Unknown
```
dcreager, line 561: This TODO is not also removed because we end up inferring this constraint set when comparing `head` to `Callable[[A], B]`:

```
(B@invoke ≤ T@head) ∧ (list[T@head] ≤ A@invoke)
```

We then try to remove `T@head` from the constraint set by calculating

```
∃T@head ⋅ (B@invoke ≤ T@head) ∧ (list[T@head] ≤ A@invoke)
```

We should be able to pick `T@head = B@invoke` and simplify that to

```
(B@invoke = *) ∧ (list[B@invoke] ≤ A@invoke)
```

which I think would then be enough to propagate through the return type to discharge this TODO. I think this would require adding more derived facts to the sequent map.

crates/ty_python_semantic/resources/mdtest/literal_promotion.md
```diff
@@ -304,7 +304,7 @@ x11: list[Literal[1] | Literal[2] | Literal[3]] = [1, 2, 3]
 reveal_type(x11)  # revealed: list[Literal[1, 2, 3]]
 
 x12: Y[Y[Literal[1]]] = [[1]]
-reveal_type(x12)  # revealed: list[Y[Literal[1]]]
+reveal_type(x12)  # revealed: list[list[Literal[1]]]
```
dcreager, line 307: This is because we're now using `specialize_recursive` instead of `specialize_partial`.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1667,6 +1660,48 @@ impl<'db> SpecializationBuilder<'db> {
...
+                        if let Type::TypeVar(lower_bound_typevar) = lower {
+                            self.add_type_mapping(
+                                lower_bound_typevar,
+                                Type::TypeVar(typevar),
+                                polarity,
+                                &mut f,
+                            );
+                        }
+                    }
+                });
```
dcreager: This is the meat of the change. We use the new `ConstraintAssignability` type relation here to calculate a constraint set describing when the two callables are assignable. That will recurse into any typevars in the callables, and find whatever constraint set allows them to unify.

We then use this new `for_each_path` method to find each way that constraint set can be satisfied, and add new typevar bindings to the old solver for whatever we find.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -995,7 +965,13 @@ impl<'db> Signature<'db> {
...
         // any other parameter list, but not a subtype or supertype of any other parameter list.
         if self.parameters.is_gradual() || other.parameters.is_gradual() {
-            return ConstraintSet::from(relation.is_assignability());
+            result.intersect(
+                db,
+                ConstraintSet::from(
+                    relation.is_assignability() || relation.is_constraint_set_assignability(),
+                ),
+            );
+            return result;
```
dcreager, line 1343: We need to combine this with `result` to hang on to any typevar mapping our constraint set has recorded for the return type comparison above. That's need to support something like

```
Callable[[int], int] ≤ Callable[..., T]
```

and have it return `int ≤ T`.

crates/ty_python_semantic/src/types.rs
```diff
@@ -1963,6 +1745,33 @@ impl<'db> Type<'db> {
...
         }
 
+        // Handle the new constraint-set-based assignability relation next. Comparisons with a
+        // typevar are translated directly into a constraint set.
+        if relation.is_constraint_set_assignability() {
+            // A typevar satisfies a relation when...it satisfies the relation. Yes that's a
+            // tautology! We're moving the caller's subtyping/assignability requirement into a
+            // constraint set. If the typevar has an upper bound or constraints, then the relation
+            // only has to hold when the typevar has a valid specialization (i.e., one that
+            // satisfies the upper bound/constraints).
```
dcreager, line 2100: This is what #20093 is trying to add to _replace_ assignability across the board. In the meantime, I've added a new `TypeRelation` that lets us opt into the new behavior only in certain places.

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
dhruvmanila: Yeah, this should work, it does when there aren't any type variables involved (https://play.ty.dev/1663d70f-4daa-492f-84d7-c35c9009fbba):

```py
from typing import Callable

def f(c: Callable[[int], int]): ...

def c(a: int) -> int: return 1

f(c)
```

dhruvmanila: Ok, I looked into this a bit more and what I'm seeing is that:

In `infer_map_impl`, the formal signature is `Callable[[A], B]` and the actual signature is the `identity` function and you invoke `formal_signature.when_constraint_set_assignable_to(..., actual_signature, ...)` which means we're checking when is `Callable[[A], B]` assignable to `identity` function which leads to checking when is the positional-only parameter (`A`) is assignable to positional-or-keyword parameter (`a: int`) which is never. This is because you cannot substitute a callable that takes a positional-or-keyword parameter with a callable that takes a positional-only parameter (https://play.ty.dev/7b31e7a0-19c8-4d2b-8bf6-507764372c0d).

This can be tested using `Protocol` to define a callable that takes a positional-or-keyword parameter:
```py
from typing import Protocol

class PositionalOrKeyword(Protocol):
    def __call__(self, a: int) -> None: ...

def positional_only(a: int, /) -> None: ...
def positional_or_keyword(a: int) -> None: ...

def test(c: PositionalOrKeyword) -> None:
    c(a=1)
```

Calling `test` using `positional_only` would fail at runtime.

crates/ty_python_semantic/resources/mdtest/call/methods.md
```diff
@@ -458,10 +458,14 @@ class C:
...
-reveal_type(C.f2(1))  # revealed: str
-reveal_type(C().f2(1))  # revealed: str
+# TODO: revealed: str
+reveal_type(C.f1(1))  # revealed: str | Divergent
+# TODO: revealed: str
+reveal_type(C().f1(1))  # revealed: str | Divergent
+# TODO: revealed: str
+reveal_type(C.f2(1))  # revealed: str | Divergent
+# TODO: revealed: str
+reveal_type(C().f2(1))  # revealed: str | Divergent
```
carljm: This definitely seems likely to impact real projects, suggesting that https://github.com/astral-sh/ty/issues/1732 is a pre-beta follow-up.

carljm:
Took a look at the conformance suite. Everything looks good except that it seems like we've lost our ability to recognize the `deprecated` decorator? See the conformance suite changes on `directives_deprecated.py`.

dcreager:
> Took a look at the conformance suite. Everything looks good except that it seems like we've lost our ability to recognize the `deprecated` decorator? See the conformance suite changes on `directives_deprecated.py`.

Yeah this showed up in the `deprecated` mdtest too. I'm looking at whether it's affected by the same cycle-merging issue as `functools.cache`.

AlexWaygood:
Looks like you need to update the number of expected diagnostics for the pandas benchmark

dcreager:
> I'm looking at whether it's affected by the same cycle-merging issue as `functools.cache`.

And this is the case...applying `@deprecated` produces a union of two function literals (one with the bottom signature and one with the real signature) due to the oscillation-preventing union. Our overload walking logic assumes that each overload will be a single, bare `Type::FunctionType`, not a union of functions, so it doesn't consider the deprecated overload to be part of the overload chain.

dcreager:
> And this is the case...applying `@deprecated` produces a union of two function literals (one with the bottom signature and one with the real signature) due to the oscillation-preventing union. Our overload walking logic assumes that each overload will be a single, bare `Type::FunctionType`, not a union of functions, so it doesn't consider the deprecated overload to be part of the overload chain.

I've pushed up a very hacky temporary fix for this — when walking overloads, if we come across a union of function literals, we pull out the one that came from the last salsa cycle iteration and use that as the type of that overload. I'm also looking at trying to eliminate the cycles entirely, but that's proving to be an invasive change, and I don't want to block this further.

dcreager:
That's a whole lot of ecosystem churn that I need to dig into

carljm:
Conformance suite results (and local testing with CLI runs) still says that even with the hacky fix, `@deprecated` is no longer working for functions. Sadly it seems to work in mdtests, but the same examples do not work when run in CLI, so the cycle problem is sensitive to something that's different in the way mdtests are run (and thus mdtests are not giving us an accurate picture of what functionality actually works, either.)

Another option for the cycle problems is to introduce hacks in `cycle_normalized` to avoid creating unions of certain known-problematic types like `FunctionLiteral` (and cross our fingers that this doesn't also reintroduce oscillation panics involving those types).

dcreager:
> Conformance suite results (and local testing with CLI runs) still says that even with the hacky fix, `@deprecated` is no longer working for functions. Sadly it seems to work in mdtests, but the same examples do not work when run in CLI, so the cycle problem is sensitive to something that's different in the way mdtests are run (and thus mdtests are not giving us an accurate picture of what functionality actually works, either.)

The first attempt at the hack only engaged when walking the overload chain, which meant that it never engaged for the implementation of an overloaded function, or for non-overloaded functions. We weren't seeing this in the mdtest suite because this wasn't enough to trigger a salsa cycle:

```py
@deprecated("use OtherClass")
def myfunc(): ...
```

Adding a parameter to `myfunc` is enough to trigger a cycle, and then we get the same failure as in the conformance suite.

> Another option for the cycle problems is to introduce hacks in `cycle_normalized` to avoid creating unions of certain known-problematic types like `FunctionLiteral` (and cross our fingers that this doesn't also reintroduce oscillation panics involving those types).

I went with this. The first hack attempt was crossing fingers like this anyway, but only for overloads. I moved the logic into `cycle_normalized` to make it engage for all parts of a function definition. Let's see how it affects the conformance suite and ecosystem...

dcreager:
The remaining conformance suite changes show us correctly inferring a specialization from the return type of a `Callable` parameter in most cases :tada:  There are still some places where we're not handling `ParamSpec`s in those callables, but that's coming soon!

[`Class7`](https://github.com/python/typing/blob/9f6d8ced7cd1c8d92687a4e9c96d7716452e471e/conformance/tests/constructors_callable.py#L151-L165) and [`Class8`](https://github.com/python/typing/blob/9f6d8ced7cd1c8d92687a4e9c96d7716452e471e/conformance/tests/constructors_callable.py#L176-L184) are the examples where we are still not inferring the _correct_ return type specialization. Both of those are generic.

carljm:
@dcreager Just came to take another review look at this -- looks like this is introducing a new fuzzer panic, and a number of other CI jobs are timing out? I'm sure you're on it, but let me know if you'd like a second pair of eyes (or just a rubber duck) on any of those issues.

dcreager:
> looks like this is introducing a new fuzzer panic, and a number of other CI jobs are timing out?

I fixed the fuzzer panic, but still seeing timeouts on some of the test repos, which is affecting the mypy_primer job and one of the wallclock jobs. `xarray` is one of the problematic repos, and in particular the `xarray/groupers.py` file in the repo. I'm letting Codex take a stab at diagnosing where the increased running time is coming from.

dcreager:
> `xarray` is one of the problematic repos, and in particular the `xarray/groupers.py` file in the repo. I'm letting Codex take a stab at diagnosing where the increased running time is coming from.

Codex helped pinpoint that the issue was [this call](https://github.com/pydata/xarray/blob/50218d01e4e0ede24c9f34c8974545b9cb0dc6fd/xarray/groupers.py#L569-L571) involving `slice`:

```py
group_indices: GroupIndices = tuple(
    list(itertools.starmap(slice, pairwise(sbins))) + [slice(sbins[-1], None)]
)
```

It turns out that `slice.__new__` is very overloaded, and many of the overloads are generic. When checking assignability of an overloaded `Callable`, I was taking the constraint sets for each overload (each _pair_ of overloads, actually, since both sides can be overloaded) and merging them together using OR. Then I would use `walk_path` on that single constraint set to pull out any assignments we should add to the old solver.

That was building up a very large BDD, that was taking a long time to iterate through. But since each of those overload pairs is independent, we can just do the path walk on each one separately. That gets rid of the combinatorial explosion. `mypy_primer` has already passed without timeout, and `xarray` worked fine locally even in the `debug` profile.

So I'm waiting for CI to confirm, but I think the timeouts have been solved now too.

dcreager:
Still in the middle of reviewing the ecosystem results, but this one stands out as interesting:

```
[error] invalid-argument-type - [:1257:20] - Argument is incorrect: Expected `int | float`, found `Literal["spam"]`
```

It looks like we're now correctly catching a [runtime `attrs` error](https://github.com/python-attrs/attrs/blob/c0111e6aad3a70080b7191aa9c7eda459a25f3dc/tests/test_validators.py#L1257)  during type-checking!

crates/ty_python_semantic/resources/mdtest/generics/pep695/variables.md
```diff
@@ -863,7 +863,7 @@ reveal_type(C[int]().y)  # revealed: int
 class D[T = T]:
     x: T
 
-reveal_type(D().x)  # revealed: T@D
+reveal_type(D().x)  # revealed: Unknown
```
dcreager, line 886: This is an error condition anyway, and I find it better to infer `Unknown` here than to let the typevar leak through when we were supposed to have solved it.

crates/ty_python_semantic/resources/mdtest/decorators.md
```diff
@@ -144,11 +144,12 @@ from functools import cache
...
-reveal_type(f)  # revealed: _lru_cache_wrapper[Unknown]
-
-# TODO: Should be `int`
-reveal_type(f(1))  # revealed: Unknown
+# TODO: revealed: _lru_cache_wrapper[int]
+# revealed: _lru_cache_wrapper[int] | _lru_cache_wrapper[Unknown]
+reveal_type(f)
+# TODO: revealed: int
+# revealed: int | Unknown
+reveal_type(f(1))
```
dcreager: The lingering `Unknown`s are because of the function definition salsa cycle issue https://github.com/astral-sh/ty/issues/1729

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -688,7 +694,11 @@ impl<'db> ConstrainedTypeVar<'db> {
     /// simplifications that we perform that operate on constraints with the same typevar, and this
     /// ensures that we can find all candidate simplifications more easily.
     fn ordering(self, db: &'db dyn Db) -> impl Ord {
-        (self.typevar(db).identity(db), self.as_id())
+        (
+            self.typevar(db).binding_context(db),
+            self.typevar(db).identity(db),
+            self.as_id(),
+        )
```
dcreager, line 738: This change makes us group all of the typevars bound by a single definition together in a constraint set BDD

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -1737,22 +1785,27 @@ impl<'db> InteriorNode<'db> {
...
             db,
-            // Remove any node that constrains `bound_typevar`, or that has a lower/upper bound of
-            // `bound_typevar`.
+            // Remove any node that constrains `bound_typevar`, or that has a lower/upper bound
+            // that mentions `bound_typevar`.
+            // TODO: This will currently remove constraints that mention a typevar, but the sequent
+            // map is not yet propagating all derived facts about those constraints. For instance,
+            // removing `T` from `T ≤ int ∧ U ≤ Sequence[T]` should produce `U ≤ Sequence[int]`.
+            // But that requires `T ≤ int ∧ U ≤ Sequence[T] → U ≤ Sequence[int]` to exist in the
+            // sequent map. It doesn't, and so we currently produce `U ≤ Unknown` in this case.
```
dcreager, line 2024: This corresponds with [this TODO](https://github.com/astral-sh/ruff/pull/21551/files#r2566717765) in the mdtests

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1748,43 +1787,19 @@ impl<'db> SpecializationBuilder<'db> {
...
 
             (Type::Callable(formal_callable), _) => {
-                if let Some(actual_callable) = actual
-                    .try_upcast_to_callable(self.db)
-                    .and_then(CallableTypes::exactly_one)
-                {
-                    // We're only interested in a formal callable of the form `Callable[P, ...]` for
-                    // now where `P` is a `ParamSpec`.
-                    // TODO: This would need to be updated once we support `Concatenate`
-                    // TODO: What to do for overloaded callables?
```
dcreager, line 1837: This logic has moved over to `CallableSignature`

dcreager:
There are quite a lot of new ecosystem diagnostics, but they are primarily due to not inferring specializations involving generic protocols yet.

carljm:
Looking at the conformance suite results:

* The new error at https://github.com/python/typing/blob/main/conformance/tests/callables_protocol.py#L199 looks pretty weird. I think it's the cycle problem... suggests that getting that fixed will be pretty important to seeing practical benefits from this PR in decorator inference?
* ... have to go, will need to pick this up later...

carljm:
In this conformance suite test (from `constructors_callable.py`), we infer the callable type `() -> Class3` (instead of the expected `(x: int) -> Class3`), and then we emit an error (on the same call to `accepts_callable` where we infer that type) that `Class3` is not assignable to `() -> Class3`. We can file a separate issue to follow up on this, just wanted to highlight it.

```py
P = ParamSpec("P")
R = TypeVar("R")

def accepts_callable(cb: Callable[P, R]) -> Callable[P, R]:
    return cb

class Class3:
    """__new__ and __init__"""

    def __new__(cls, *args, **kwargs) -> Self: ...

    def __init__(self, x: int) -> None: ...


r3 = accepts_callable(Class3)
reveal_type(r3)  # `def (x: int) -> Class3`
assert_type(r3(3), Class3)
r3()  # E
r3(y=1)  # E
r3(1, 2)  # E
```

carljm:
Similarly in this test we infer `() -> Never` instead of the expected `(*args: Any, **kwargs: Any) -> Never`:

```py
class Meta1(type):
    def __call__(cls, *args: Any, **kwargs: Any) -> NoReturn:
        raise NotImplementedError("Class not constructable")


class Class5(metaclass=Meta1):
    """Custom metaclass that overrides type.__call__"""

    def __new__(cls, *args: Any, **kwargs: Any) -> Self:
        """This __new__ is ignored for purposes of conversion"""
        return super().__new__(cls)

r5 = accepts_callable(Class5)
reveal_type(r5)  # `def (*args: Any, **kwargs: Any) -> NoReturn`
```

This looks similar to the previous issue, like we have an issue with defaulting ParamSpecs to "zero args" when inferring them from a class constructor.

carljm:
EDIT: this is clearly marked as a TODO already in the tests, we can ignore for now (but I'll leave this here and create an issue from it.)

In this case we infer a return type of `Class7[int] | Class7[str]` for both overloads, where we should have a return type of `Class7[int]` on one overload and `Class7[str]` on the other:

```py
class Class7(Generic[T]):
    @overload
    def __init__(self: "Class7[int]", x: int) -> None: ...
    @overload
    def __init__(self: "Class7[str]", x: str) -> None: ...
    def __init__(self, x: int | str) -> None:
        pass


r7 = accepts_callable(Class7)
reveal_type(
    r7
)  # overload of `def (x: int) -> Class7[int]` and `def (x: str) -> Class7[str]`
assert_type(r7(0), Class7[int])
assert_type(r7(""), Class7[str])
```

carljm:
EDIT: this is also already clearly marked as a TODO in the tests.

In both of these cases, we don't re-bind the type variable to the inferred callable type, so we aren't able to infer it in calls to that callable:

```py
class Class8(Generic[T]):
    def __new__(cls, x: list[T], y: list[T]) -> Self:
        return super().__new__(cls)


r8 = accepts_callable(Class8)
reveal_type(r8)  # `def [T] (x: list[T], y: list[T]) -> Class8[T]`
assert_type(r8([""], [""]), Class8[str])
r8([1], [""])  # E

class Class9:
    def __init__(self, x: list[T], y: list[T]) -> None:
        pass


r9 = accepts_callable(Class9)
reveal_type(r9)  # `def [T] (x: list[T], y: list[T]) -> Class9`
assert_type(r9([""], [""]), Class9)
r9([1], [""])  # E
```

carljm:
(I'm not assuming we should fix these in this PR, they can be follow-ups; this is just a convenient place to record my findings for now, can convert them to new issues when this lands.)

carljm:
I'm seeing a regression introduced by this PR in a number of ecosystem projects, where it looks like we are taking the bottom materialization of argument types when we infer a ParamSpec, where before this PR we did not. Minimal repro:

```py
from typing import Callable

def callable_identity[**P, R](func: Callable[P, R]) -> Callable[P, R]:
    return func

@callable_identity
def f(env: dict) -> None:
    pass

reveal_type(f)
```

Before this PR we reveal `(env: dict[Unknown, Unknown]) -> Unknown`.
On this PR we reveal `(env: Bottom[dict[Unknown, Unknown]]) -> None`.

This is an improvement on the return type, of course, but a major problem for the argument type, since there is no way to provide a value assignable to `Bottom[dict[Unknown, Unknown]]`.

This also manifests in some cases as wrongly inferring a callable that takes no parameters, when something like `staticmethod(...)` is called on a callable with gradual parameter list `...`. The reason is that we currently have some (wrong) logic that the bottom materialization of a gradual callable is the empty parameter list. Pretty sure this is also the root cause of some of the conformance suite issues I noted above where we end up wrongly inferring an empty argument list. 

Correct materialization of gradual callables is probably higher priority now that we have `ParamSpec` inference support, though it wouldn't matter here if we weren't wrongly taking the bottom materialization.

I'm seeing these issues in enough ecosystem projects that I think it might be a blocker here. It often comes up, for example, with use of `contextlib.contextmanager`.

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
carljm: It looks like this comment still needs follow-up? Why are we checking assignability in the wrong direction (requiring the formal signature to be assignable to the actual signature)? I don't think the change here and below to make `identity` and `head` take a positional-only parameter should be required.

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
```diff
@@ -487,14 +487,13 @@ from typing import Callable
 def invoke[A, B](fn: Callable[[A], B], value: A) -> B:
     return fn(value)
 
-def identity[T](x: T) -> T:
+def identity[T](x: T, /) -> T:
     return x
 
-def head[T](xs: list[T]) -> T:
+def head[T](xs: list[T], /) -> T:
```
carljm: Similarly here, I think these changes should not be required, and it suggests a problem in this PR if they are.

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
```diff
@@ -736,3 +735,108 @@ def f[T](x: T, y: Not[T]) -> T:
...
+class Meta(type):
+    def __call__(cls, *args: Any, **kwargs: Any) -> NoReturn:
+        raise NotImplementedError
+
+class ClassWithNoReturnMetatype(metaclass=Meta):
+    def __new__(cls, *args: Any, **kwargs: Any) -> Self:
+        raise NotImplementedError
+
+# revealed: Never
+reveal_type(accepts_callable(ClassWithNoReturnMetatype)())
```
carljm, line 823: On this test (and maybe some others, which look derived from the conformance suite) we are now getting the return type correct, but (untested here) we are regressing on the signature, due to the issue mentioned elsewhere where we are now wrongly taking bottom materializations of inferred signatures.

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
```diff
@@ -662,7 +663,7 @@ reveal_type(change_return_type(int_int))  # revealed: Overload[(x: int) -> str,
 reveal_type(change_return_type(int_str))  # revealed: Overload[(x: int) -> str, (x: str) -> str]
 
 # error: [invalid-argument-type]
-reveal_type(change_return_type(str_str))  # revealed: Overload[(x: int) -> str, (x: str) -> str]
+reveal_type(change_return_type(str_str))  # revealed: (...) -> str
```
carljm, line 682: Hmm, why do we lose the parameters here?

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -348,7 +348,8 @@ from ty_extensions import ConstraintSet, generic_context
 def mentions[T, U]():
     # (T@mentions ≤ int) ∧ (U@mentions = list[T@mentions])
     constraints = ConstraintSet.range(Never, T, int) & ConstraintSet.range(list[T], U, list[T])
-    # revealed: ty_extensions.Specialization[T@mentions = int, U@mentions = list[int]]
+    # TODO: revealed: ty_extensions.Specialization[T@mentions = int, U@mentions = list[int]]
+    # revealed: ty_extensions.Specialization[T@mentions = int, U@mentions = Unknown]
```
carljm: Do we understand the cause of this regression and what needs to happen to fix it?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -207,7 +209,7 @@ impl<'db> ConstraintSet<'db> {
                 lower.top_materialization(db),
                 upper.bottom_materialization(db),
             ),
-            TypeRelation::Assignability => (
+            TypeRelation::Assignability | TypeRelation::ConstraintSetAssignability => (
                 lower.bottom_materialization(db),
                 upper.top_materialization(db),
```
carljm: Is this where the wrong bottom materializations are coming from?

I think in the ecosystem I also saw some cases of top materializations inferred where they shouldn't be...

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -724,7 +738,7 @@ impl<'db> ConstrainedTypeVar<'db> {
 
         // If `lower ≰ upper`, then the intersection is empty, since there is no type that is both
         // greater than `lower`, and less than `upper`.
-        if !lower.is_subtype_of(db, upper) {
+        if !lower.is_constraint_set_assignable_to(db, upper) {
```
carljm, line 766: It seems like we are making this change a lot, which suggests we would be better able to handle gradual bounds now?

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1812,43 +1877,19 @@ impl<'db> SpecializationBuilder<'db> {
...
-                        },
-                        actual_signatures => Type::Callable(CallableType::new(
+                let Some(actual_callables) = actual.try_upcast_to_callable(self.db) else {
+                    return Ok(());
+                };
+
+                for actual_callable in actual_callables.as_slice() {
+                    let when = formal_callable
+                        .signatures(self.db)
+                        .when_constraint_set_assignable_to(
```
carljm: Is it backwards to check assignability of the formal callable to the actual callable?

Is this why we wrongly had to adjust some tests to make the argument callables take a positional-only parameter?

crates/ty_python_semantic/src/types.rs
```diff
@@ -912,16 +912,29 @@ impl<'db> Type<'db> {
...
             (Type::GenericAlias(prev_alias), Type::GenericAlias(curr_alias))
                 if prev_alias.origin(db) == curr_alias.origin(db) =>
             {
                 self
             }
+
+            // Similarly, don't union together two function literals, since there are several parts
+            // of our type inference machinery that assume that we infer a single FunctionLiteral
+            // type for each overload of each function definition.
+            (Type::FunctionLiteral(_), Type::FunctionLiteral(_)) => self,
```
carljm: We could check that the two function literals are coming from the same definition (though if they are the result of the same query it seems like maybe they would have to be?)

crates/ty_python_semantic/src/types.rs
```diff
@@ -1779,7 +1792,7 @@ impl<'db> Type<'db> {
                 }
             }
             Type::ClassLiteral(class_literal) => {
-                Some(class_literal.default_specialization(db).into_callable(db))
+                Some(class_literal.identity_specialization(db).into_callable(db))
```
carljm, line 1863: Is this the place where we'd want to re-bind typevars to the new callable type? Should we add a TODO for that?

carljm (commented):
Looking great overall! A few questions and issues that I see in the ecosystem report.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -337,6 +335,37 @@ impl<'db> CallableSignature<'db> {
...
+    fn signatures_is_single_paramspec(
+        signatures: &[Signature<'db>],
+    ) -> Option<(BoundTypeVarInstance<'db>, Option<Type<'db>>)> {
+        let [signature] = signatures else {
+            return None;
+        };
+        signature
+            .parameters
+            .as_paramspec()
+            .map(|bound_typevar| (bound_typevar, signature.return_ty))
```
dhruvmanila, line 358: Can you add a TODO comment here stating that this might need to be updated once `Concatenate` is supported?

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -348,6 +377,114 @@ impl<'db> CallableSignature<'db> {
         relation_visitor: &HasRelationToVisitor<'db>,
         disjointness_visitor: &IsDisjointVisitor<'db>,
     ) -> ConstraintSet<'db> {
+        if relation.is_constraint_set_assignability() {
+            // TODO: Oof, maybe ParamSpec needs to live at CallableSignature, not Signature?
```
dhruvmanila, line 389: Can you say why? Is it about overloads? I think it should still be on `Signature` given that it could be different for each overload.

I think for an overloaded function that's being matched against e.g., `Callable[P, int]`, the typevar would need to be constrained to match only those callables whose return type is assignable to `int`. There might be other examples.

crates/ty_python_semantic/src/types/signatures.rs
```diff
@@ -348,6 +377,114 @@ impl<'db> CallableSignature<'db> {
...
+            let self_is_single_paramspec = Self::signatures_is_single_paramspec(self_signatures);
+            let other_is_single_paramspec = Self::signatures_is_single_paramspec(other_signatures);
+
+            // If either callable is a ParamSpec, the constraint set should bind the ParamSpec to
+            // the other callable's signature. We also need to compare the return types — for
+            // instance, to verify in `Callable[P, int]` that the return type is assignable to
+            // `int`, or in `Callable[P, T]` to bind `T` to the return type of the other callable.
+            match (self_is_single_paramspec, other_is_single_paramspec) {
```
dhruvmanila, line 397: Would this need to be moved to `Signature` once overloads are supported?

During the initial implementation of `ParamSpec`, when I started off by trying to use the new solver, I had a similar `match` expression in `Signature::has_relation_impl_inner` after the `is_gradual` checks where I'd try to add similar constraints.

crates/ty_python_semantic/src/types.rs
```diff
@@ -12433,6 +12543,10 @@ impl<'db> CallableTypes<'db> {
         }
     }
 
+    fn as_slice(&self) -> &SmallVec<[CallableType<'db>; 1]> {
+        &self.0
+    }
```
dhruvmanila: Should this return a slice instead (`&[CallableType<'db>]`) ?

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
dcreager: Fixed with the parameter ordering change you mentioned below

crates/ty_python_semantic/resources/mdtest/generics/legacy/functions.md
dcreager: Fixed with the parameter ordering change you mention below

crates/ty_python_semantic/resources/mdtest/generics/pep695/paramspec.md
dcreager: This is an example where we're now collectively taking into account all of the constraints across the entire call site. In this case, we're correctly flagging `invalid-argument-type`, because `change_return_type` expects a callable that returns an `int`, which neither overload of `str_str` does. Before we went ahead and inferred a specialization of `P` from the callable parameter, even though it was invalid. But now, since the argument is invalid, we don't infer anything for `P`, and fall back on its `Unknown` equivalent `(...)`

crates/ty_python_semantic/src/types.rs
dcreager: Done

crates/ty_python_semantic/src/types/generics.rs
dcreager: Image

Yes, yes it was. `infer` has the parameters in opposite order from `has_relation_to`, and I copy-pasted this badly. Had to add a bit of extra logic ot `add_type_mappings_from_constraint_set`, but this is now in the correct order.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: As discussed in discord, we're going ahead and doing this in a separate PR, https://github.com/astral-sh/ruff/pull/21957. That means that as of now, constraint sets cannot be used to track subtyping queries, but we will revisit that in the future if/when we need it.

crates/ty_python_semantic/src/types.rs
dcreager: I'm not sure I follow — which rebinding do you mean?

crates/ty_python_semantic/src/types/signatures.rs
dcreager: Done

crates/ty_python_semantic/src/types/signatures.rs
dcreager: My thinking is that the following are two different things:

```py
def f(*args: P.args, **kwargs: P.kwargs) -> whatever: ...
Callable[P, whatever]
```

The first one is definitely scoped to a single overload. `P` is always in non-inferable position when you use `P.args` and `P.kwargs`.

But the second one seems like `P` is scoped to more than one overload. It's often in inferable position. When it is, and we're checking if some other callable is assignable to the `Callable[P, ...]`, we need to look at all of the overloads of the other callable to see which ones match. I'm not sure how you'd get different (inferable) `P`s for different overloads. Do you have an example?

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -1338,10 +1338,16 @@ impl<'db> Node<'db> {
...
+                let new_greatest_lower_bound = UnionBuilder::new(db)
+                    .order_elements(true)
+                    .add(greatest_lower_bound)
+                    .add(constraint.lower(db))
+                    .build();
+                let new_least_upper_bound = IntersectionBuilder::new(db)
+                    .order_elements(true)
+                    .add_positive(least_upper_bound)
+                    .add_positive(constraint.upper(db))
+                    .build();
```
AlexWaygood: `order_elements(true)` orders the union elements using `union_or_intersection_elements_ordering`. That comparison function _doesn't_ guarantee stable ordering of union elements between different runs of ty, which is what would be required for stable mdtest revealed types. It only guarantees stable ordering within the context of a single invocation of ty.

carljm: Oops that was my suggestion (on the previous PR)

AlexWaygood: hmm, yeah... not sure it's ever really correct to set `builder.order_elements(true)` outside of a `Type::normalized(_impl)` call? The uses in `constraints.rs` look a bit suspect to me. (Were those the ones that were added on the previous PR? If it made the `reveal_type` output stable in that PR, I worry that that might have just been luck/chance...)

We should add some docs to the `UnionBuilder::order_elements()` method that make this limitation clear.

carljm: The `order_elements()` method itself was added in the previous PR, in order to enable the uses in `constraints.rs`

AlexWaygood: no it wasn't, you added it in https://github.com/astral-sh/ruff/pull/20147 back in August

carljm: It was added to intersection builder in the previous PR, which is what is used in constraints.rs; previously it existed only on union builder

AlexWaygood: oh _you're_ talking about `IntersectionBuilder::order_elements()`? That one was added yesterday, yes

carljm: Yes, and that's the one this thread is about, given the code it is commenting on 😆

AlexWaygood: > It was added to intersection builder in the previous PR, which is what is used in constraints.rs

(no, constraints.rs uses both `UnionBuilder::order_elements()` _and_ `IntersectionBuilder::order_elements()`)

AlexWaygood: > Yes, and that's the one this thread is about, given the code it is commenting on 😆

um, well, the first two lines of the codeframe my original comment is attached to are

```rs
                let new_greatest_lower_bound = UnionBuilder::new(db)
                    .order_elements(true)
```

so... no, `IntersectionBuilder::order_elements()` is not what this thread was originally about? 😆

carljm: Sorry, shouldn't comment from my phone. The ordering changes in yesterday's PR were originally motivated by intersections, I didn't notice that it started ordering unions too. We should revert all uses of order_elements here (separate PR maybe) and figure out a different way to address the original instability.

dcreager: I jumped right back into the terminal window this evening without seeing this thread until now. Trying to clarify my understanding: Do we have anything currently that guarantees a stable ordering across runs? Does `normalized` do that? If so, is that because it applies `order_elements` recursively to the elements of the types as well? If not, how are we not running into these same instabilities in other mdtests? Is it because without `order_elements`, unions and intersections retain the ordering of the caller, and we're very careful to build up unions/intersections in all of those places in a reproducible way?

AlexWaygood: > Do we have anything currently that guarantees a stable ordering across runs?

Not, currently, no

> Does `normalized` do that?

No, only within a single run 

> If not, how are we not running into these same instabilities in other mdtests? Is it because without `order_elements`, unions and intersections retain the ordering of the caller, and we're very careful to build up unions/intersections in all of those places in a reproducible way?

Yes, we always retain insertion order for unions and intersections. Unions and intersections are generally created either directly as a result of a user annotation, or as a result of control flow. In either case, insertion order is always guaranteed to be stable between runs of ty

dcreager: Thanks for clarifying that, @AlexWaygood! I've got most of way done with adding a proper stable ordering on https://github.com/astral-sh/ruff/pull/21983, just a little bit more to clean up on that tomorrow.

dcreager:
> I'm seeing a regression introduced by this PR in a number of ecosystem projects, where it looks like we are taking the bottom materialization of argument types when we infer a ParamSpec, where before this PR we did not. Minimal repro:
> 
> ```python
> from typing import Callable
> 
> def callable_identity[**P, R](func: Callable[P, R]) -> Callable[P, R]:
>     return func
> 
> @callable_identity
> def f(env: dict) -> None:
>     pass
> 
> reveal_type(f)
> ```

I added this as an mdtest

crates/ty_python_semantic/src/types.rs
dcreager: Yes, thank you!

crates/ty_python_semantic/src/types/signatures.rs
dcreager: I think this is related to your suggestion above — I had this in `Signature` at first too, but had to move it up here to make some of the tests pass. (The ones where the paramspec needs to find a full overload set of the other side).

But if we need to have this match the subset of the overloads that match the paramspec, then I think you're right? I'll add a note.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: This is now moot since we allow gradual lower/upper bounds

dcreager (commented):
> In this conformance suite test (from `constructors_callable.py`), we infer the callable type `() -> Class3` (instead of the expected `(x: int) -> Class3`), and then we emit an error (on the same call to `accepts_callable` where we infer that type) that `Class3` is not assignable to `() -> Class3`. We can file a separate issue to follow up on this, just wanted to highlight it.

This is `ClassWithNewAndInit` in the mdtests. I've updated it to be fully in line with `Class3` from the conformance suite. We now infer `(...) -> Class3` as the callable that we map `**P` to, and the assignability check passes. Ideally this would map `**P` to a union of `(...) -> Class3` and `(x: int) -> Class3`, but there is an existing TODO from the paramspec implementation where we don't currently unify multiple solutions of a single paramspec.

> Similarly in this test we infer `() -> Never` instead of the expected `(*args: Any, **kwargs: Any) -> Never`:

`Class5` is `ClassWithNoReturnMetatype` in the mdtests. We're now inferring `(...) -> Unknown` — so we're getting the arguments correct now, but the return type is not being inferred. This can be follow-on work; I've added a TODO explaining why.

crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md
dcreager: See below; we're now getting the return wrong wrong, but the signature correct.

crates/ty_python_semantic/src/types/constraints.rs
dcreager: #21983 is merged; rebased this onto it

carljm:
Conformance suite is looking pretty good!

On `constructors_callable.py` the only unexpected (that is, not already TODOed) issue I see is that we are now erroring on this call itself: https://github.com/python/typing/blob/77bbb7e7366568501a8b78b36317181fc939d11d/conformance/tests/constructors_callable.py#L141 It seems like a mismatch between callable inference and callable assignability: we infer the type `() -> Any | Class6Any` from the call (which seems wrong, not sure where the `| Class6Any` part comes from), and then we don't think the class `Class6Any` is assignable to that callable type.

In `dataclasses_transform_converter.py` we are newly wrongly erroring on these two lines: https://github.com/python/typing/blob/77bbb7e7366568501a8b78b36317181fc939d11d/conformance/tests/dataclasses_transform_converter.py#L102-L103  Seems like an issue where with an overloaded callable, we end up inferring the union of the argument types from the overloads, and then (correctly) fail to actually assign to that type from the overloaded callable.

(I guess this general category of issue, where we infer a specialization and then fail to assign to that same specialization, is something that will go away when we are fully using constraint sets and get rid of the old specialization builder?)

I think both of these issues are things we can tackle as follow-ups.

carljm:
I explored the Expression ecosystem hits a bit. The issue seems to have something to do with their [`Option` type](https://github.com/dbrattli/Expression/blob/main/expression/core/option.py#L43). I tried to simplify a repro without that type, and [it works fine](https://github.com/dbrattli/Expression/blob/main/expression/core/option.py#L43).

carljm:
Hmm, no, I was doing something wrong before. This problem repros with a pretty simple case. Minimal repro:

```py
from typing import TypeVar, Generic, Callable, Mapping

Key = TypeVar("Key", bound=int)
Value = TypeVar("Value")

class MapTree(Generic[Key, Value]):
    pass

def exists(f: Callable[[Key, Value], bool], m: MapTree[Key, Value]) -> bool:
    return True

_Key = TypeVar("_Key", bound=int)
_Value = TypeVar("_Value")

def _(p: Callable[[_Key, _Value], bool], m: MapTree[_Key, _Value]):
    exists(p, m)
```

carljm:
Even more minimal repro of the Expression problem; it doesn't require that we are matching up multiple arguments:

```py
from typing import TypeVar, Generic, Callable

Key = TypeVar("Key", bound=int)
Value = TypeVar("Value")

def exists(f: Callable[[Key, Value], bool]) -> bool:
    return True

_Key = TypeVar("_Key", bound=int)
_Value = TypeVar("_Value")

def _(p: Callable[[_Key, _Value], bool]):
    exists(p)  # false positive here
```

If we remove the upper bounds on `Key` and `_Key` typevars, it's fine. But with the upper bounds, at some point we wrongly simplify a typevar to its upper bound in constraint solving, and then that causes an assignability error.

dcreager:
> If we remove the upper bounds on `Key` and `_Key` typevars, it's fine. But with the upper bounds, at some point we wrongly simplify a typevar to its upper bound in constraint solving, and then that causes an assignability error.

A-ha! I ran into this on the generic protocols PR, too. This commit 9db0e3975c3796b53ebf1f52ec8d472b558228a8 fixes that issue by tracking that the bounds/constraints of a typevar should be used to limit which specializations are allowed, but shouldn't be surfaced as a specialization directly.

carljm:
Awesome! Do you think it makes sense to backport that commit here? Or land it as a separate PR with tests specific to constraint sets?

carljm:
The new diagnostics on homeassistant look like true positives. [`EventType` is invariant](https://github.com/home-assistant/core/blob/dev/homeassistant/util/event_type.pyi#L11-L15), so `EventType[Mapping[str, Never]]` should not be assignable to `EventType[Mapping[str, object]]`. And the `Never` here is not an inference bug, it's [explicit](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/typing.py#L15).

carljm:
Actually I might have spoken too soon on the homeassistant one; the `Never` is explicit, but the `object` is inferred. It's a typevar with a gradual upper bound of `Mapping[str, Any]`. Looking closer...

dcreager:
> Awesome! Do you think it makes sense to backport that commit here? Or land it as a separate PR with tests specific to constraint sets?

That one is chunky enough that I think it does deserve a separate PR.

carljm:
Yeah, I was wrong, homeassistant does look like a bug. Other type checkers are fine with it. Minimal repro:

```py
from typing import Generic, TypeVar, Mapping, Any, Callable, Never

T = TypeVar("T", bound=Mapping[str, Any], default=Mapping[str, Any])

class Event(Generic[T]):
    pass

class EventType(Generic[T]):
    pass

STOP: EventType[Mapping[str, Never]] = EventType()

def listen(et: EventType[T], cb: Callable[[Event[T]], None]):
    pass

def mycb(event: Event) -> None:
    pass

def _():
    listen(STOP, mycb)
```

We have a false positive on the last line, we solve `T` to `Mapping[str, object]` instead of `Mapping[str, Any]`, and then fail to assign `STOP` (which is `EventType[Mapping[str, Never]]` to `EventType[T]`.

Do we still take top materializations of gradual upper bounds? That seems like it might be the problem here.

carljm:
> That one is chunky enough that I think it does deserve a separate PR.

Are you thinking to get that PR landed first and then rebase this and revisit its ecosystem report? Or would you prefer to just land this with its remaining issues and tackle them all as follow-ups?

dcreager:
> Are you thinking to get that PR landed first and then rebase this and revisit its ecosystem report? Or would you prefer to just land this with its remaining issues and tackle them all as follow-ups?

Given the time, another option would be to combine this PR and the generic protocol one and try to land it all together. I'm down to one last mdtest file with test regressions to fix.

carljm:
We can try that. Do you have a lot of other fixes in the protocol PR that might help this PR, too? If so, combining might be most expedient.

The downside might be if the combined PR is harder to review ecosystem impact and land, because its so big?

If you want, I could also pull out that commit into its own PR and add tests for it, while you keep working on the protocol PR?

dcreager:
> If you want, I could also pull out that commit into its own PR and add tests for it, while you keep working on the protocol PR?

Sure, that works! I've got it started on the `dcreager/explicit-constraints` branch

dcreager:
> We have a false positive on the last line, we solve `T` to `Mapping[str, object]` instead of `Mapping[str, Any]`, and then fail to assign `STOP` (which is `EventType[Mapping[str, Never]]` to `EventType[T]`.
> 
> Do we still take top materializations of gradual upper bounds? That seems like it might be the problem here.

For this homeassistant error, I think it's the same as this TODO: https://github.com/astral-sh/ruff/pull/21551/changes#diff-d08dc0c815ce12a46b39702c611395d769e53bc49746c3a0c8533ed8af014863R814-R815

(That same comment appears as a TODO in some other mdtests, too.)

The issue is that we're creating a constraint set of the form `Never ≤ T`, where we mean that we actually want to map `T = Never`, but we're confusing that constraint set with "`T` has no lower bound". The new explicit markings should help take care of this, but I think we can do that as a follow-on.

carljm:
Hmm, not sure about that, because the issue repros even without `Never`. If I change the `Never` to `int`, we still have the assignability error where we infer `Mapping[str, object]` and try to assign `Mapping[str, int]`.

But removing the default and changing the upper bound `Mapping[str, Any]` to `Mapping[str, object]` doesn't solve the problem, either, so it's not related to typevar defaults or to gradual upper bounds. Here's the even-more-simplified example (removing `Never` and `Any` and typevar defaults):

```py
from typing import Generic, TypeVar, Mapping, Any, Callable, Never

T = TypeVar("T", bound=Mapping[str, object])

class Event(Generic[T]):
    pass

class EventType(Generic[T]):
    pass

STOP: EventType[Mapping[str, int]] = EventType()

def listen(et: EventType[T], cb: Callable[[Event[T]], None]):
    pass

def mycb(event: Event) -> None:
    pass

def _():
    listen(STOP, mycb)
```

Other typecheckers are still happy with this; we emit an error on the bottom line yet.

Maybe this is a variant of the same problem as the Expression one, and will also be fixed by the explicit-constraints change?

If not I agree we can look into it as follow-on.

dcreager:
> Maybe this is a variant of the same problem as the Expression one, and will also be fixed by the explicit-constraints change?

Ah that's a good catch. If we're inferring the bounds, then yes I think the explicit-constraints change will help.

dcreager:
Any idea where a 101 exit code comes from? Does that happen when we panic? When I run ty locally on spack (both dev and profiling profiles), I get a successful type check.

carljm:
Just in terms of numbers, ecosystem diff is definitely improved here! We were adding almost 1200 new diagnostics before, now its under 800.

carljm:
Where do you see the 101 return code? Primer and ecosystem analyzer jobs both completed successfully...

EDIT: oh never mind, I see it once I link to the full ecosystem analyzer output... but spack checked fine on the mypy-primer job. Running ecosystem analyzer again to see if it persists.

carljm:
Spack error is gone on the second run; I suspect an infra issue. (I guess I didn't check the logs for the ecosystem analyzer run to see if there was anything useful there.)

Spot-checked the first Expression failure -- looks like a combo of generic protocol inference, plus us using narrower inferred types. Not an issue for this PR. I'll look at a couple more quick?

dcreager:
> Spot-checked the first Expression failure -- looks like a combo of generic protocol inference, plus us using narrower inferred types. Not an issue for this PR. I'll look at a couple more quick?

Sounds good, thanks!

carljm:
Second one also looks like generic protocols.

carljm:
In bokeh I see an example that boils down to this:

```py
from typing import Callable, TypeVar

T = TypeVar("T")

def infer(func: Callable[[T | str], T]) -> T:
    return func("foo")

def convert_str_seq(value: list[str] | str) -> list[str]:
    if isinstance(value, str):
        return [value]
    return value

reveal_type(infer(convert_str_seq))  # expected `list[str]` but we say `list[str] | str`
```

Basically we don't unpack the union and match it item-by-item, so the `T | str` in the formal parameter doesn't eliminate `| str` from the union in the actual argument.

But pyrefly does the same as we do here -- I think this one is fine to look at as follow-up.

carljm (approved):
Ship it!

PR #21552, opened 2025-11-21
[ty] support generic aliases in `type[...]`, like `type[C[int]]`

oconnor663:
Closes https://github.com/astral-sh/ty/issues/1101.

As I push this PR, there is [one failing test line](https://github.com/astral-sh/ruff/blob/06941c1987bd4fe2b54d99b5514784472cabc0b3/crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md?plain=1#L244) on my box, which I need help interpreting. Here's an extracted version of it:

```py
from ty_extensions import static_assert, is_assignable_to, TypeOf
from typing import Generic, TypeVar

T_co = TypeVar("T_co", covariant=True)

class Foo(Generic[T_co]): ...
class Bar(Foo[T_co], Generic[T_co]): ...

static_assert(is_assignable_to(TypeOf[Bar], type[Foo[int]]))  # error: false
```

That `static_assert` fails. The "never" result seems to originate on [this line](https://github.com/astral-sh/ruff/blob/06941c1987bd4fe2b54d99b5514784472cabc0b3/crates/ty_python_semantic/src/types/class.rs#L591), which I read as "no non-generic class literal is ever assignable/etc to a specialization" (not sure how to word that correctly). That...seems not true...?...so I'm wondering I've somehow skipped the step where the non-generic class literal, which does in fact have generic parameters, gets them implicitly specialized to `object`? I'm betting this will be super obvious to someone :)

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
+++ b/crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
@@ -241,6 +241,7 @@ class Bar(Foo[T_co], Generic[T_co]): ...
 
 static_assert(is_assignable_to(TypeOf[Bar[int]], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar[bool]], type[Foo[int]]))
+static_assert(is_assignable_to(TypeOf[Bar], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar[Any]], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar[Unknown]], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar], type[Foo]))
@@ -249,10 +250,6 @@ static_assert(is_assignable_to(TypeOf[Bar[Any]], type[Foo[int]]))
 
 static_assert(not is_assignable_to(TypeOf[Bar[int]], type[Foo[bool]]))
 static_assert(not is_assignable_to(TypeOf[Foo[bool]], type[Bar[int]]))
-
-# `TypeOf` does not implicitly default-specialize. The unspecialized class literal object `Bar` does
-# not inhabit the type `type[Foo[int]]`.
-static_assert(is_assignable_to(TypeOf[Bar], type[Foo[int]]))
 ```
 
 ## `type[]` is not assignable to types disjoint from `builtins.type`


oconnor663:
Also thanks @carljm for talking me through the implementation of this :)

carljm:
The implicit specialization should use the `default_specialization` method, and should not specialize to `object` but to `Unknown` (or the typevar default, if any, but there is none in this example).

To be honest I'm not sure if we want `TypeOf[Bar]` to implicitly default-specialize `Bar`, though. `TypeOf` is a test introspection helper designed to say "give me the type for this name", and I think it might break some other uses of it if it always implicitly default-specialized instead of giving the "raw" type of that name.

So I suspect the better option here might be to change that test assertion, with a comment saying something like "`TypeOf` does not implicitly default-specialize. The unspecialized class literal object `Bar` does not inhabit the type `type[Foo[int]]`."

And then add another test asserting explicitly that `type[Bar[Unknown]]` is assignable to `type[Foo[int]]`, if we don't already have that version.

I think the line you pointed to in `ClassType::has_relation_to_impl` is correct. What it is saying is that I am only a subclass of `other` if I inherit from a base class with matching specialization with `other` -- if `other` is generic and my base is not (or vice versa), that cannot make me a subclass of `other`. Note that we are in a `when_any(...)` iteration of `self.iter_mro(db)` there.

crates/ty_python_semantic/resources/mdtest/type_properties/is_assignable_to.md
```diff
@@ -241,15 +241,18 @@ class Bar(Foo[T_co], Generic[T_co]): ...
 
 static_assert(is_assignable_to(TypeOf[Bar[int]], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar[bool]], type[Foo[int]]))
-static_assert(is_assignable_to(TypeOf[Bar], type[Foo[int]]))
 static_assert(is_assignable_to(TypeOf[Bar[Any]], type[Foo[int]]))
+static_assert(is_assignable_to(TypeOf[Bar[Unknown]], type[Foo[int]]))
```
oconnor663, line 246: Is this new `Unknown` case redundant with the `Any` case right above it, or are is their behavior different enough that it makes sense to test both?

carljm: It's redundant, we very rarely handle `Unknown` differently from `Any`. But there's also no harm in including it, don't really care either way.

carljm (approved):
Ship it!

AlexWaygood:
> Ship it!

Looks like we're panicking on a mypy_primer project in CI?

carljm:
Nice catch :) Ok then, don't ship it quite yet...

oconnor663:
What's one little panic between friends?

oconnor663:
Progress, this seems to be a minimal repro:

```py
class Foo:
    pass

_: type[Foo[int]]
```

```
error[panic]: Panicked at crates/ty_python_semantic/src/types/infer/builder.rs:6946:14 when checking `/tmp/test.py`: `assertion `left == right` failed
  left: Some(ClassLiteral(ClassLiteral { name: Name("Foo"), body_scope: ScopeId { [salsa id]: Id(1001), file: File(System("/tmp/test.py")), file_scope_id: FileScopeId(1) }, known: None, deprecated: None, type_check_only: false, dataclass_params: None, dataclass_transformer_params: None }))
 right: None`
info: This indicates a bug in ty.
info: If you could open an issue at https://github.com/astral-sh/ty/issues/new?title=%5Bpanic%5D, we'd be very appreciative!
info: Platform: linux x86_64
info: Version: ruff/0.14.5+81 (502618634 2025-11-21)
info: Args: ["/home/jacko/astral/ruff/target-mold/debug/ty", "check", "test.py"]
info: run with `RUST_BACKTRACE=1` environment variable to show the full backtrace information
info: query stacktrace:
   0: infer_definition_types(Id(1401))
             at crates/ty_python_semantic/src/types/infer.rs:94
   1: infer_scope_types(Id(1000))
             at crates/ty_python_semantic/src/types/infer.rs:70
   2: check_file_impl(Id(c00))
             at crates/ty_project/src/lib.rs:535
```

I think this is the panic that we get when something is inferred more than once. Did it used to have a better error message than this?

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -695,6 +697,40 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
...
+                                    self.infer_explicit_callable_specialization(
+                                        subscript,
+                                        value_ty,
+                                        generic_context,
+                                        specialize,
+                                    )
+                                }
+                                None => {
+                                    // TODO: emit a diagnostic if you try to specialize a non-generic class.
+                                    self.infer_type_expression(slice);
```
carljm: This looks like the likely cause of the panic, `slice` includes `value` which we already inferred above.

oconnor663: Yep already there.

oconnor663: Interesting that the line in the wild that triggers this is https://github.com/python-jsonschema/referencing/blob/d97e65e86c95d51f01e18d89461da6d4a0c4e72a/referencing/_core.py#L200

```py
        default_specification: (
            type[Specification[D]] | Specification[D]
        ) = Specification,
```

But in this context, `Specification` _is_ generic...

carljm: Oh, that's strange and deserving of further investigation?

But doesn't need to block landing this IMO.

oconnor663:
CI is green. Looking at some of the new diagnostics. For example [in `werkzeug`](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/sansio/request.py#L75):

```py
    parameter_storage_class: type[MultiDict[str, t.Any]] = ImmutableMultiDict
```

It seems like that's a true positive, and that the annotation should be `type[MultiDict]`? (Alternatively the assigned value could be `ImmutableMultiDict[str, t.Any`, but I think they want an actual class there? Not totally sure.)

carljm:
Hmmm, no, I think those ecosystem results might suggest we need to make an adjustment here... and I might have been wrong on the original question about the failing test line. I think we need to default-specialize in that assignment and consider the class object `ImmutableMultiDict` to have the type `type[ImmutableMultiDict[Unknown, Unknown]]`, which is assignable to `type[MultiDict[str, Any]]`.

What's not totally clear to me is at which level we should do that default specialization. Should we do it on the name load? At the assignability check?

carljm:
The conformance suite results are showing a similar problem. I think the use of `assert_type` in those tests is a bit questionable, but it's showing the same problem -- that we aren't default-specializing a generic class literal when the class object is mentioned.

oconnor663:
At the risk of biting off more than I can chew, do we want to revisit [this](https://github.com/astral-sh/ruff/blob/ddc1417f22599ceaa998c1ee4053c657afb2d577/crates/ty_python_semantic/src/types/class.rs#L610-L614):

```rs
            // A non-generic class is never equivalent to a generic class.
            // Two non-generic classes are only equivalent if they are equal (handled above).
            (ClassType::NonGeneric(_), _) | (_, ClassType::NonGeneric(_)) => {
                ConstraintSet::from(false)
            }
```

I'll try it just for kicks...

carljm:
No, I don't think that's the right location for the fix. 

I think the problem is that we sometimes use `ClassType::NonGeneric` to represent an unspecialized generic class, and I think we should never(?) do that. `ClassType::NonGeneric` should always be an actual non-generic class type, and `ClassType::Generic` should always be a specialized generic class type. Because a true "class type" for a generic class must always be specialized, even if only default-specialized.

We have `Type::ClassLiteral` to represent a class literal object (non-generic or generic, but never specialized), so that when you say `MyGenericClass[X]` (an explicit specialization of `MyGenericClass`) or `MyGenericClass(X)` (a constructor call of `MyGenericClass` for which we will infer a specialization), we have some type for `MyGenericClass` in that expression that knows it isn't specialized already. So we can't just universally make all name-loads of `MyGenericClass` eagerly default-specialize themselves. But `Type::ClassLiteral` doesn't wrap `ClassType`, it directly wraps `ClassLiteral`, so I think that's still consistent with the idea that we should avoid ever constructing a `ClassType` that attempts to represent an unspecialized generic class.

I think maybe the narrow version of the fix here would be located at https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types.rs#L2381 -- that is at least one place where we are clearly constructing a `ClassType::NonGeneric` around an arbitrary `ClassLiteral`, with no regard for whether it is actually generic or not. Instead I think we should transform our `ClassLiteral` into a `ClassType` using a `ClassType` constructor (which AFAICS doesn't currently exist on `ClassType`, but I think should) which checks if the `ClassLiteral` is generic, and if so default-specializes it and constructs a `ClassType::Generic` around it, only otherwise constructing a `ClassType::NonGeneric`.

This fix is essentially saying that we consider `Type::ClassLiteral(Foo)` to be a subtype of `Type::SubclassOf(<default specialization of Foo>)` -- it's equivalent to the default specialization of `Foo` itself, excluding subclasses. I think this is a reasonable way to implement the expectation that any mention of `Foo` which _isn't_ an explicit specialization or a constructor call, should be considered the default specialization.

I suspect that fix may be enough to get the werkzeug example passing, and (depending what the rest of the ecosystem looks like) may be enough to make this landable.

The next follow-up (which could be interesting to try here, in case it just works smoothly, but if it reveals more hard problems may be better deferred) would be to actually add some assertions in `ClassType` which enforce the invariant that we never make a `ClassType::NonGeneric` around a generic `ClassLiteral`.

The next thing which we'd want to look at soon is fixing those conformance suite results. I think we may want to actually make an upstream pull request to the conformance suite to adjust those tests, because I think the way they are written is over-specified. They are asserting that the expression `Foo` (where `Foo` is a generic class) must have precisely the type `type[Foo[Unknown]]` (or whatever the default specialization of `Foo` might be). I think this is over-prescriptive, because it is valid for us to infer a narrower type here. What the conformance suite really should be asserting here is just assignability behavior, not precise types.

oconnor663:
[83d2bc2](https://github.com/astral-sh/ruff/pull/21552/commits/83d2bc237cd167dc99d0b8e5235de4a21c82455f) has cleared most of the new `werkzeug` errors. I'm playing with relaxing the conformance suite now to see how that looks (not necessarily to change in this PR, but to see where this PR stands if we were to change it).

I think the remaining new `werkzeug` error is valid:

```
+ tests/test_formparser.py:437:76: error[invalid-argument-type] Argument to bound method `__init__` is incorrect: Expected `type[MultiDict[str, Any]] | None`, found `<class 'dict'>`
```

I think the `cls=dict` argument [here](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/tests/test_formparser.py#L437) is indeed incorrect, because it's [supposed to be](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/formparser.py#L297) a `type[MultiDict[...]]`, but `MultiDict` [inherits from](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/datastructures/structures.py#L137) `TypeConversionDict` which [inherits from](https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/datastructures/structures.py#L57) `dict`. We weren't surfacing that until this PR because we didn't parse `type[MultiDict[...]]`.

oconnor663:
@AlexWaygood has clarified that this `mongo-python-driver` ecosystem hit:
image
is an instance of this known issue: https://github.com/astral-sh/ty/issues/1513

oconnor663:
@AlexWaygood again re: the `mypy` ecosystem hit:

> that diagnostic is being surfaced on mypy's vendored copy of typeshed. I think that's just us failing to recognise that mypy (like ty) is vendoring typeshed, and that therefore the tuple in mypy/typeshed/stdlib/builtins.pyi is builtins.tuple, and therefore gets to be afforded the same very special rules that the tuple class in ty_vendored/vendor/typeshed/stdlib/builtins.pyi is afforded

oconnor663:
@carljm points out that `psycopg` errors like...
```
Object of type `<class 'Cursor'>` is not assignable to attribute `cursor_factory` of type `type[Cursor[Row@Connection]]`
```
...are legitimate, because `Row` is a typevar with a `default` there, so it doesn't default to `Unknown`. Separate question whether we actually like this `ty` behavior, but it's not new in this PR. (Edit: could be something that bidirectional type inference helps with in the future.)

carljm:
A bunch of the other new diagnostics in the ecosystem derive from https://github.com/astral-sh/ty/issues/740

carljm:
I filed https://github.com/astral-sh/ty/issues/1624 for follow-up on the psycopg2 issue mentioned above

PR #21553, opened 2025-11-21
[ty] Generic types aliases (implicit and PEP 613)

sharkdp:
## Summary

Add support for generic PEP 613 type aliases and generic implicit type aliases:
```py
from typing import TypeVar

T = TypeVar("T")
ListOrSet = list[T] | set[T]

def _(xs: ListOrSet[int]):
    reveal_type(xs)  # list[int] | set[int]
```

closes https://github.com/astral-sh/ty/issues/1643
closes https://github.com/astral-sh/ty/issues/1629
closes https://github.com/astral-sh/ty/issues/1596
closes https://github.com/astral-sh/ty/issues/573
closes https://github.com/astral-sh/ty/issues/221

## Typing conformance

```diff
-aliases_explicit.py:52:5: error[type-assertion-failure] Type `list[int]` does not match asserted type `@Todo(specialized generic alias in type expression)`
-aliases_explicit.py:53:5: error[type-assertion-failure] Type `tuple[str, ...] | list[str]` does not match asserted type `@Todo(Generic specialization of types.UnionType)`
-aliases_explicit.py:54:5: error[type-assertion-failure] Type `tuple[int, int, int, str]` does not match asserted type `@Todo(specialized generic alias in type expression)`
-aliases_explicit.py:56:5: error[type-assertion-failure] Type `(int, str, /) -> str` does not match asserted type `@Todo(Generic specialization of typing.Callable)`
-aliases_explicit.py:59:5: error[type-assertion-failure] Type `int | str | None | list[list[int]]` does not match asserted type `int | str | None | list[@Todo(specialized generic alias in type expression)]`
```

New true negatives :heavy_check_mark: 

```diff
+aliases_explicit.py:41:36: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
-aliases_explicit.py:57:5: error[type-assertion-failure] Type `(int, str, str, /) -> None` does not match asserted type `@Todo(Generic specialization of typing.Callable)`
+aliases_explicit.py:57:5: error[type-assertion-failure] Type `(int, str, str, /) -> None` does not match asserted type `(...) -> Unknown`
```

These require `ParamSpec`

```diff
+aliases_explicit.py:67:24: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+aliases_explicit.py:68:24: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+aliases_explicit.py:69:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_explicit.py:70:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_explicit.py:71:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_explicit.py:102:20: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
```

New true positives :heavy_check_mark: 

```diff
-aliases_implicit.py:63:5: error[type-assertion-failure] Type `list[int]` does not match asserted type `@Todo(specialized generic alias in type expression)`
-aliases_implicit.py:64:5: error[type-assertion-failure] Type `tuple[str, ...] | list[str]` does not match asserted type `@Todo(Generic specialization of types.UnionType)`
-aliases_implicit.py:65:5: error[type-assertion-failure] Type `tuple[int, int, int, str]` does not match asserted type `@Todo(specialized generic alias in type expression)`
-aliases_implicit.py:67:5: error[type-assertion-failure] Type `(int, str, /) -> str` does not match asserted type `@Todo(Generic specialization of typing.Callable)`
-aliases_implicit.py:70:5: error[type-assertion-failure] Type `int | str | None | list[list[int]]` does not match asserted type `int | str | None | list[@Todo(specialized generic alias in type expression)]`
-aliases_implicit.py:71:5: error[type-assertion-failure] Type `list[bool]` does not match asserted type `@Todo(specialized generic alias in type expression)`
```

New true negatives :heavy_check_mark: 

```diff
+aliases_implicit.py:54:36: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
-aliases_implicit.py:68:5: error[type-assertion-failure] Type `(int, str, str, /) -> None` does not match asserted type `@Todo(Generic specialization of typing.Callable)`
+aliases_implicit.py:68:5: error[type-assertion-failure] Type `(int, str, str, /) -> None` does not match asserted type `(...) -> Unknown`
```

These require `ParamSpec`

```diff
+aliases_implicit.py:76:24: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+aliases_implicit.py:77:24: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+aliases_implicit.py:78:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_implicit.py:79:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_implicit.py:80:29: error[invalid-type-arguments] Too many type arguments: expected 1, got 2
+aliases_implicit.py:81:25: error[invalid-type-arguments] Type `str` is not assignable to upper bound `int | float` of type variable `TFloat@GoodTypeAlias12`
+aliases_implicit.py:135:20: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
```

New true positives :heavy_check_mark: 

```diff
+callables_annotation.py:172:19: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+callables_annotation.py:175:19: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+callables_annotation.py:188:25: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
+callables_annotation.py:189:25: error[invalid-type-arguments] Too many type arguments: expected 0, got 1
```

These require `ParamSpec` and `Concatenate`.

```diff
-generics_defaults_specialization.py:26:5: error[type-assertion-failure] Type `SomethingWithNoDefaults[int, str]` does not match asserted type `SomethingWithNoDefaults[int, typing.TypeVar]`
+generics_defaults_specialization.py:26:5: error[type-assertion-failure] Type `SomethingWithNoDefaults[int, str]` does not match asserted type `SomethingWithNoDefaults[int, DefaultStrT]`
```

Favorable diagnostic change :heavy_check_mark: 

```diff
-generics_defaults_specialization.py:27:5: error[type-assertion-failure] Type `SomethingWithNoDefaults[int, bool]` does not match asserted type `@Todo(specialized generic alias in type expression)`
```

New true negative :heavy_check_mark: 

```diff
-generics_defaults_specialization.py:30:1: error[non-subscriptable] Cannot subscript object of type `<class 'SomethingWithNoDefaults[int, typing.TypeVar]'>` with no `__class_getitem__` method
+generics_defaults_specialization.py:30:15: error[invalid-type-arguments] Too many type arguments: expected between 0 and 1, got 2
```

Correct new diagnostic :heavy_check_mark: 

```diff
-generics_variance.py:175:25: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:175:35: error[non-subscriptable] Cannot subscript object of type `<class 'Co[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:179:29: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:179:39: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:183:21: error[non-subscriptable] Cannot subscript object of type `<class 'Co[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:183:27: error[non-subscriptable] Cannot subscript object of type `<class 'Co[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:187:25: error[non-subscriptable] Cannot subscript object of type `<class 'Co[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:187:31: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:191:33: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:191:43: error[non-subscriptable] Cannot subscript object of type `<class 'Co[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:191:49: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:196:5: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:196:15: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
-generics_variance.py:196:25: error[non-subscriptable] Cannot subscript object of type `<class 'Contra[typing.TypeVar]'>` with no `__class_getitem__` method
```

One of these should apparently be an error, but not of this kind, so this is good :heavy_check_mark: 

```diff
-specialtypes_type.py:152:16: error[invalid-type-form] `typing.TypeVar` is not a generic class
-specialtypes_type.py:156:16: error[invalid-type-form] `typing.TypeVar` is not a generic class
```

Good, those were false positives. :heavy_check_mark: 

I skipped the analysis for everything involving `TypeVarTuple`.

## Ecosystem impact

**[Full report with detailed diff](https://david-generic-implicit-alias.ecosystem-663.pages.dev/diff)**

Previous iterations of this PR showed all kinds of problems. In it's current state, I do not see any large systematic problems, but it is hard to tell with 5k diagnostic changes.

## Performance

* There is a huge 4x regression in `colour-science/colour`, related to [this large file](https://github.com/colour-science/colour/blob/develop/colour/io/luts/tests/test_lut.py) with [many assignments of hard-coded arrays (lists of lists) to `np.NDArray` types](https://github.com/colour-science/colour/blob/83e754c8b6932532f246b9fe1727d14f3600cd83/colour/io/luts/tests/test_lut.py#L701-L781) that we now understand. We now take ~2 seconds to check this file, so definitely not great, but maybe acceptable for now.

## Test Plan

Updated and new Markdown tests

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -11113,9 +11113,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                         ));
                     }
                     return Type::KnownInstance(KnownInstanceType::Callable(
-                        CallableType::unknown(self.db())
-                            .as_callable()
-                            .expect("always returns Type::Callable"),
+                        CallableType::unknown(self.db()),
                     ));
                 }
 
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -1121,7 +1121,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
         {
             Type::single_callable(db, Signature::new(parameters, Some(return_type)))
         } else {
-            CallableType::unknown(db)
+            Type::Callable(CallableType::unknown(db))
         };
 
         // `Signature` / `Parameters` are not a `Type` variant, so we're storing
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -3726,7 +3726,7 @@ impl<'db> Type<'db> {
                     dunder_call
                         .has_relation_to_impl(
                             db,
-                            CallableType::unknown(db),
+                            Type::Callable(CallableType::unknown(db)),
                             inferable,
                             TypeRelation::Assignability,
                             relation_visitor,
@@ -7258,7 +7258,7 @@ impl<'db> Type<'db> {
                 SpecialFormType::OrderedDict => Ok(KnownClass::OrderedDict.to_instance(db)),
 
                 // TODO: Use an opt-in rule for a bare `Callable`
-                SpecialFormType::Callable => Ok(CallableType::unknown(db)),
+                SpecialFormType::Callable => Ok(Type::Callable(CallableType::unknown(db))),
 
                 // Special case: `NamedTuple` in a type expression is understood to describe the type
                 // `tuple[object, ...] & <a protocol that any `NamedTuple` class would satisfy>`.
@@ -11684,8 +11684,8 @@ impl<'db> CallableType<'db> {
     }
 
     /// Create a callable type which accepts any parameters and returns an `Unknown` type.
-    pub(crate) fn unknown(db: &'db dyn Db) -> Type<'db> {
-        Type::Callable(Self::single(db, Signature::unknown()))
+    pub(crate) fn unknown(db: &'db dyn Db) -> CallableType<'db> {
+        Self::single(db, Signature::unknown())
     }
 
     pub(crate) fn bind_self(
--- a/crates/ty_python_semantic/src/types/narrow.rs
+++ b/crates/ty_python_semantic/src/types/narrow.rs
@@ -238,7 +238,7 @@ impl ClassInfoConstraintFunction {
             Type::SpecialForm(SpecialFormType::Callable)
                 if self == ClassInfoConstraintFunction::IsInstance =>
             {
-                Some(CallableType::unknown(db).top_materialization(db))
+                Some(Type::Callable(CallableType::unknown(db)).top_materialization(db))
             }
 
             Type::SpecialForm(special_form) => special_form


crates/ty_python_semantic/resources/mdtest/async.md
```diff
@@ -79,9 +79,8 @@ async def main():
         task("B"),
     )
 
-    # TODO: these should be `int`
-    reveal_type(a)  # revealed: Unknown
-    reveal_type(b)  # revealed: Unknown
+    reveal_type(a)  # revealed: int
+    reveal_type(b)  # revealed: int
```
sharkdp, line 82: We now understand the [`_FutureLike`](https://github.com/python/typeshed/blob/3e1b90bb2d772781c2b6f318e0f44ef791cf5160/stdlib/asyncio/tasks.pyi#L78-L81) type alias.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -191,13 +191,13 @@ def _(
     reveal_type(int_or_callable)  # revealed: int | ((str, /) -> bytes)
     reveal_type(callable_or_int)  # revealed: ((str, /) -> bytes) | int
     # TODO should be Unknown | int
-    reveal_type(type_var_or_int)  # revealed: typing.TypeVar | int
+    reveal_type(type_var_or_int)  # revealed: T@TypeVarOrInt | int
```
sharkdp, line 194: I tried a couple of different approaches, but so far, I have not found a way to default-specialize type aliases that are used without an explicit specialization. It's fairly easy to turn this into `Unknown | int`, but so far, I haven't found a way that doesn't break random other things.

So these stay as TODOs with this PR

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -388,73 +390,62 @@ ListOrTuple = list[T] | tuple[T, ...]
 ListOrTupleLegacy = Union[list[T], tuple[T, ...]]
 MyCallable = Callable[P, T]
 AnnotatedType = Annotated[T, "tag"]
+TransparentAlias = T
```
AlexWaygood, line 393: Pyrefly and pyright accept `TransparentAlias` as a valid type alias, but [mypy does not](https://mypy-play.net/?mypy=latest&python=3.12&gist=32eeea189747857ec9ae7e01f0fa30db). I think we could consider adding a TODO here for us to reject it too when it appears in a type expression. `TransparentAlias[int]` fails at runtime, just like `T[int]` fails at runtime. And there's no reason to use `TransparentAlias[int]` rather than just `int`.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -388,73 +390,62 @@ ListOrTuple = list[T] | tuple[T, ...]
...
+reveal_type(DictStrTo)  # revealed: <class 'dict[str, U@DictStrTo]'>
 
 def _(
-    # TODO: No error here
-    # error: [invalid-type-form] "Invalid subscript of object of type `GenericAlias` in type expression"
     dict_str_to_int: DictStrTo[int],
 ):
-    # TODO: This should be `dict[str, int]`
-    reveal_type(dict_str_to_int)  # revealed: Unknown
+    reveal_type(dict_str_to_int)  # revealed: dict[str, int]
```
AlexWaygood, line 448: great work, this is fantastic!

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -464,43 +455,118 @@ expected:
...
+SubclassOfInt = MyType[int]
+CallableIntToStr = MyCallable[[int], str]
+
+reveal_type(IntsOrNone)  # revealed: types.UnionType
+reveal_type(IntsOrStrs)  # revealed: types.UnionType
+reveal_type(ListOfPairs)  # revealed: <class 'list[tuple[str, str]]'>
+reveal_type(ListOrTupleOfInts)  # revealed: types.UnionType
+reveal_type(AnnotatedInt)  # revealed: <typing.Annotated special form>
+reveal_type(SubclassOfInt)  # revealed: GenericAlias
+reveal_type(CallableIntToStr)  # revealed: @Todo(Callable[..] specialized with ParamSpec)
```
AlexWaygood, line 469: I guess the `Todo` message here is a bit misleading, because there's no `ParamSpec` involved here?

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -522,37 +588,184 @@ reveal_mro(Derived1)
...
 
 ```py
 ListOfInts = list[int]
 
-# TODO: this should be an error
+# error: [invalid-type-arguments] "Too many type arguments: expected 0, got 1"
 def _(doubly_specialized: ListOfInts[int]):
-    # TODO: this should be `Unknown`
-    reveal_type(doubly_specialized)  # revealed: @Todo(specialized generic alias in type expression)
+    reveal_type(doubly_specialized)  # revealed: list[int]
```
AlexWaygood, line 671: the TODO comment feels like it's still relevant here? I think `Unknown` or `list[Unknown]` would be best, rather than `list[int]`. This is going to raise an exception at runtime if they're using Python <3.14 and they don't have `from __future__ import annotations` at the top of their file. And we don't really know _what_ they intended to write: they might have actually wanted their type alias to be generic over two type variables.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -522,37 +588,184 @@ reveal_mro(Derived1)
...
+
+Similarly, if you try to specialize a union type without a binding context, we emit an error:
+
+```py
+# TODO: Better error message (of kind `invalid-type-form`)?
+# error: [invalid-type-arguments] "Too many type arguments: expected 0, got 1"
+x: (list[T] | set[T])[int]
+
+def _():
+    reveal_type(x)  # revealed: list[typing.TypeVar] | set[typing.TypeVar]
```
AlexWaygood, line 723:
```suggestion
    # TODO: `list[Unknown] | set[Unknown]` might be better
    reveal_type(x)  # revealed: list[typing.TypeVar] | set[typing.TypeVar]
```

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
```diff
@@ -522,37 +588,184 @@ reveal_mro(Derived1)
...
+if flag():
+    MyAlias = list[T]
+else:
+    MyAlias = set[T]
+
+# TODO: This should be an error like "Invalid subscript of object of type `<class 'list[T@MyAlias]'> | <class 'set[T@MyAlias]'>` in type expression"
+def _(x: MyAlias[int]):
+    # TODO: This should be `Unknown`
+    reveal_type(x)  # revealed: list[int] | set[int]
 ```
```
AlexWaygood, line 794: I would be fine with continuing to not emit errors on these; it sort-of seems like a feature that we understand them 😄 But I also don't oppose changing starting to emit diagnostics on these later if it simplifies our implementation somehow

AlexWaygood (commented):
Just a review of the tests so far. A couple more tests you could add (possibly in a followup). I haven't checked whether or not these pass on your branch, just some pathological cases I thought of while reading your tests:

```py
from typing import TypeAlias, TypeVar

T = TypeVar("T")
U = TypeVar("U")

TotallyStringifiedPEP613: TypeAlias = "dict[T, U]"
TotallyStringifiedPartiallySpecialized: TypeAlias = "TotallyStringifiedPEP613[U, int]"

def f(x: "TotallyStringifiedPartiallySpecialized[str]"):
    reveal_type(x)  # revealed: dict[str, int]
```

(Edit David: Added. This one still resolves to a `@Todo` type that was introduced for this precise case)

and

```py
from typing import TypeVar

T = TypeVar("T")
U = TypeVar("U")
V = TypeVar("V")

X = tuple[T, *tuple[U, ...], V]
Y = X[T, tuple[int, str, U], bytes]

def g(obj: Y[bool, range]):
    reveal_type(obj)  # revealed: tuple[bool, *tuple[tuple[int, str, range], ...], bytes]
```

(Edit David: Added. This one passes!)

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -4796,6 +4796,12 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 unpacked.expression_type(target)
             }
             TargetKind::Single => {
+                // This could be an implicit type alias (OptionalList = list[T] | None). Use the definition
+                // of `OptionalList` as the binding context while inferring the RHS (`list[T] | None`), in
+                // order to bind `T` to `OptionalList`.
+                let previous_typevar_binding_context =
+                    self.typevar_binding_context.replace(definition);
```
AlexWaygood, line 4803: oh, this is clever!

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -10860,6 +10875,14 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
     fn infer_subscript_load(&mut self, subscript: &ast::ExprSubscript) -> Type<'db> {
         let value_ty = self.infer_expression(&subscript.value, TypeContext::default());
+
+        // If we have an implicit type alias like `MyList = list[T]`, and if `MyList` is being
+        // used in another implicit type alias like `Numbers = MyList[int]`, then we infer the
+        // right hand side as a value expression, and need to handle the specialization here.
+        if value_ty.is_generic_alias() {
+            return self.infer_explicit_type_alias_specialization(subscript, value_ty, false);
+        }
```
AlexWaygood, line 10884: should we rename the method to just `infer_type_alias_specialization`? it seems we now call it for explicit type aliases and implicit type aliases 😄

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11065,6 +11088,37 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                    if let Some(builder) = self.context.report_lint(&INVALID_TYPE_FORM, subscript) {
+                        builder.into_diagnostic(format_args!(
+                            "The first argument to `Callable` must be either a list of types, \
+                                 ParamSpec, Concatenate, or `...`",
+                        ));
+                    }
+                    return Type::KnownInstance(KnownInstanceType::Callable(
+                        CallableType::unknown(self.db())
+                            .as_callable()
+                            .expect("always returns Type::Callable"),
```
AlexWaygood: we should refactor the method so that the return type is `CallableType` rather than `Type`, then we wouldn't need the `.expect()`. I can pick that up after this is landed.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -904,7 +977,10 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                 }
             },
             Type::Dynamic(_) => {
-                self.infer_type_expression(slice);
+                // Infer slice as a value expression to avoid false-positive
+                // `invalid-type-form` diagnostics, when we have e.g.
+                // `MyCallable[[int, str], None]` but `MyCallable` is dynamic.
+                self.infer_expression(slice, TypeContext::default());
```
AlexWaygood, line 983: I think there are probably lots more error cases in type-expression parsing where we should do similarly

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
sharkdp: > `TransparentAlias[int]` fails at runtime, just like `T[int]` fails at runtime

Ok, but it's sort-of reasonable to have it in a type annotation? I spent extra effort to support it after Carl mentioned it, but we can also remove that special handling again, I guess.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
sharkdp: Hm, not here directly, but `CallableIntToStr` really traces back to `MyCallable = Callable[P, T]` where `P` is a `ParamSpec`, and this is just how our `@Todo` types propagate?

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
sharkdp: I'll remove the `TODO`s and just mention that we might later want to change those.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
AlexWaygood: I'd personally lean towards rejecting it because i can't see any use case for it — it probably indicates that the user is confused about Python typing in some way?

But i don't have a strong opinion; don't let this slow you down when it comes to getting the PR in, if you and/or Carl disagree

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: To be fair, the `.replace` pattern is not something I invented. We already use this in other places to preserve the outer binding context.

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
AlexWaygood: Oh i see. That's fine then

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Oh, "explicit" is an overloaded term here. This is supposed to be read as: "infer *an explicit specialization* of a (implicit or PEP-613/explicit) type alias". This is in line with `infer_explicit_type_alias_type_specialization`, which was already there before (but has been slightly renamed here).

We could consider calling it `infer_explicit_specialization_of_type_alias` or similar…

crates/ty_python_semantic/src/types/infer/builder.rs
sharkdp: Done

crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
sharkdp: > I'd personally lean towards rejecting it because i can't see any use case for it — it probably indicates that the user is confused about Python typing in some way?

Maybe. But it seems like an artificial restriction to me? If `MyAlias = T | None` and `MyAlias = Annotated[T, "bla"]` are fine, then why should `MyAlias = T` be forbidden? A PEP 695 type alias `type MyAlias[T] = T` is (hopefully) also allowed.

PR #21554, opened 2025-11-21
[ty] Attach subdiagnostics to `unresolved-import` errors for relative imports as well as absolute imports

AlexWaygood:
## Summary

If we're not able to resolve an import, we attach subdiagnostics to the error saying which search paths we looked for the import in, and linking to our settings page. However, we currently only attach these subdiagnostics if `level == 0` -- i.e., if it was an _absolute_ import that was unresolved. This doesn't make much sense; if anything, relative imports are more sensitive to the details of exactly which search paths the user has or hasn't got configured. It looks to me like this is just an accident; this PR changes things so that we attach the subdiagnostic to all `unresolved-import` diagnostics, not just absolute ones.

The diff here is easiest to review with GitHub's "hide whitespace changes" feature enabled 😄

## Test Plan

Snapshot updated


zsol:
> The diff here is easiest to review with GitHub's "hide whitespace changes" feature enabled 😄

Lies! 😆

PR #21555, opened 2025-11-21
[ty] Improve debug messages when imports fail

AlexWaygood:
## Summary

- Improve a debug message (which is meant to be user-facing but wasn't written in a way that it would help users).
- Add a subdiagnostic if an unresolved relative module name could be resolved were the user to decrease the number of leading dots.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/unresolved_import.md_-_Unresolved_import_di…_-_Using_`from`_with_to…_(4b8ba6ee48180cdd).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/unresolved_import.md_-_Unresolved_import_di…_-_Using_`from`_with_to…_(4b8ba6ee48180cdd).snap
@@ -40,6 +40,8 @@ error[unresolved-import]: Cannot resolve imported module `....foo`
 2 |
 3 | stat = add(10, 15)
   |
+help: The module can be resolved if the number of leading dots is reduced
+help: Did you mean `...foo`?
 info: Searched in the following paths during module resolution:
 info:   1. /src (first-party code)
 info:   2. vendored://stdlib (stdlib typeshed stubs vendored by ty)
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -6025,7 +6025,8 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             }
             Err(ModuleNameResolutionError::TooManyDots) => {
                 tracing::debug!(
-                    "Relative module resolution `{}` failed: too many leading dots",
+                    "Relative module resolution `{}` failed: too many leading dots \
+                    (try adjusting configured search paths?)",
```
MichaReiser: I'm not sure I find this useful; it might even be misleading because it's as likely that the user just used too many dots. What would be useful here is to test if the module can be resolved by reducing the number of dots (when we generate the diagnostic) and make a suggestion based on that.

AlexWaygood: fair

AlexWaygood: Pushed a change implementing this

crates/ty_python_semantic/resources/mdtest/snapshots/unresolved_import.md_-_Unresolved_import_di…_-_Using_`from`_with_to…_(4b8ba6ee48180cdd).snap
```diff
@@ -40,6 +40,8 @@ error[unresolved-import]: Cannot resolve imported module `....foo`
 2 |
 3 | stat = add(10, 15)
   |
+help: The module *can* be resolved if the number of leading dots is reduced
```
MichaReiser: Nice :) 

I would remove the *can*. I think the question mark is enough to indicate that we aren't sure

PR #21558, opened 2025-11-21
Bump 0.14.6

ntBre:


ntBre:
macOS test flake is tracked here: https://github.com/astral-sh/ty/issues/1540. Not sure about the mypy_primer result, but I don't see how it can be related to this PR.

PR #21559, opened 2025-11-21
Set severity for non-rule diagnostics

senekor:
## Summary

This change makes ruff server emit LSP diagnostics with severity level in more situations. Specifically, if the diagnostic didn't have a "secondary code" (I'm not super familiar what that is.) ruff server would previously emit _no_ severity.

This isn't good, because we always have an internal severity, so the information is right there. Editor experience is degraded without the severity information. For example, I use Helix. Themes typically don't consider the case of diagnostics without severity, and the fallback style is very ugly on most themes. Another functional problem is that Helix has a "diagnostics picker", which can filter by severity. This allows users to focus on errors before warnings. It doesn't work without the severity information.

I also added `tags(diagnostic)` to the same code path, because the information also doesn't seem to depend on the "secondary code". However, I haven't noticed this information missing in my own use of ruff, so I didn't confirm the difference in my own testing.

## Test Plan

I saw diagnostics without severity in my editor (Helix), made the fix, installed ruff, restarted LSP in Helix, confirmed that diagnostics now have severity.

A very simple file to confirm this just contains `if`, which will emit a "Expected an expression" diagnostic. With this PR, it has the severity "error", without it it has none.

### DIFF
--- a/crates/ruff_server/src/lint.rs
+++ b/crates/ruff_server/src/lint.rs
@@ -283,24 +283,27 @@ fn to_lsp_diagnostic(
         range = diagnostic_range.to_range(source_kind.source_code(), index, encoding);
     }
 
-    let (severity, tags, code) = if let Some(code) = code {
-        let code = code.to_string();
-        (
-            Some(severity(&code)),
-            tags(diagnostic),
-            Some(lsp_types::NumberOrString::String(code)),
-        )
+    let (severity, code) = if let Some(code) = code {
+        (severity(code), code.to_string())
     } else {
-        (None, None, None)
+        (
+            match diagnostic.severity() {
+                ruff_db::diagnostic::Severity::Info => lsp_types::DiagnosticSeverity::INFORMATION,
+                ruff_db::diagnostic::Severity::Warning => lsp_types::DiagnosticSeverity::WARNING,
+                ruff_db::diagnostic::Severity::Error => lsp_types::DiagnosticSeverity::ERROR,
+                ruff_db::diagnostic::Severity::Fatal => lsp_types::DiagnosticSeverity::ERROR,
+            },
+            diagnostic.id().to_string(),
+        )
     };
 
     (
         cell,
         lsp_types::Diagnostic {
             range,
-            severity,
-            tags,
-            code,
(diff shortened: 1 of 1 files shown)


crates/ruff_server/src/lint.rs
```diff
@@ -286,19 +286,25 @@ fn to_lsp_diagnostic(
...
         let code = code.to_string();
         (
-            Some(severity(&code)),
+            severity(&code),
             tags(diagnostic),
             Some(lsp_types::NumberOrString::String(code)),
         )
     } else {
-        (None, None, None)
+        let severity = match diagnostic.severity() {
```
MichaReiser: Thank you. Hmm, this seems like an oversight to me from our diagnostic refactor. @ntBre could we always use `diagnostic.severity` here?

ntBre: Yeah it looks like it to me, it does seem like an oversight because I couldn't use the `severity` helper.

MichaReiser: Oh no we can't because all diagnostics should remain errors on the CLI, but we want to reduce the severity of some errors to warnings in the editor only

ntBre: Oh right, yes. I wasn't sure if you meant could we have used the `diagnostic.severity()` in _this branch_ all along (which is what I answered), or if we could just use `diagnostic.severity()` for both cases now. I think the changes in this PR make sense to me since we do the remapping you said.

MichaReiser: I would do a small refactor here to also show the diagnostic id in case the secondary code is missing. This should also allow us to make the two code paths more similar

```rust
let (severity, code) = if let Some(code) = code {
	(severity(code), code))
} else {
	(
		match diagnostic.severity { ... },
		&diagnostic.id
	)
};

let tags = tags(diagnostics); // or inline into the struct construction below
let code = lsp_types::NumberOrString::String(code.to_string());
```

senekor: updated

MichaReiser (approved):
Awesome, thank you

PR #21561, opened 2025-11-21
[ty] Improve diagnostics when a submodule is not available as an attribute on a module-literal type

AlexWaygood:
## Summary

It's often not obvious to users why we sometimes report `unresolved-attribute` errors for submodules-accessed-on-parent-modules, even though these attributes often are available at runtime (sometimes by sheer luck!). This PR adds a dedicated diagnostic for this case that helps guide the user towards how they can fix the error.

## Test Plan

Added snapshots

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
@@ -60,7 +60,7 @@ Y: int = 47
 import mypackage
 
 reveal_type(mypackage.imported.X)  # revealed: int
-# error: "has no member `fails`"
+# error: [unresolved-attribute] "Submodule `fails` may not be available"
 reveal_type(mypackage.fails.Y)  # revealed: Unknown
 ```
 
@@ -90,7 +90,7 @@ Y: int = 47
 import mypackage
 
 reveal_type(mypackage.imported.X)  # revealed: int
-# error: "has no member `fails`"
+# error: [unresolved-attribute] "Submodule `fails` may not be available"
 reveal_type(mypackage.fails.Y)  # revealed: Unknown
 ```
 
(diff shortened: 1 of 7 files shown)


AlexWaygood:
I skimmed the ecosystem results, and couldn't see any false positives where the new diagnostic message was inappropriate.

crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -60,7 +60,7 @@ Y: int = 47
 import mypackage
 
 reveal_type(mypackage.imported.X)  # revealed: int
-# error: "has no member `fails`"
+# error: [unresolved-attribute]
```
carljm: I think the old form of the assertion here was intentional -- it ensures we are erroring on the attribute we think we are, not e.g. the attribute `Y`. So I consider all these changes in this file to be somewhat of a regression in assertion precision. We could do this to make it even less sensitive to changes in the message, and add the code?
```suggestion
# error: [unresolved-attribute] "`fails`"
```

carljm (approved):

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -9077,10 +9077,30 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            maybe_submodule_name.extend(&relative_submodule);
+                            if resolve_module(db, &maybe_submodule_name).is_some() {
+                                let mut diag = builder.into_diagnostic(format_args!(
+                                    "Submodule `{attr_name}` may not be available as an attribute \
+                                    on module `{module_name}`"
+                                ));
+                                diag.help(format_args!(
+                                    "Consider explicitly importing `{maybe_submodule_name}`"
+                                ));
+                                return fallback();
```
AlexWaygood, line 9096: if we hit this branch, we have pretty high confidence that we know what the user _meant_ here -- we could even consider inferring the module-literal type of the submodule rather than inferring `Unknown`. (Or `<submodule type> | Unknown`...?)

And since this often will (by pure luck!) work at runtime, we could also consider downgrading the diagnostic to warning-level rather than error-level

But either of those would be separate changes from simply improving the diagnostic error message, which is all that this humble PR seeks to achieve.

PR #21564, opened 2025-11-21
[ty] Use the same snapshot handling as other tests

Gankra:
Fixes https://github.com/astral-sh/ty/issues/1605


Gankra:
(Yes it just makes it Worse but whatever it's consistent now)

Issue #13410, opened 2024-09-19
There should be an option to skip COM812 (missing-trailing-comma) on single-item calls/lists. [Feature request]


echo-lalia:
Currently, the rule `COM812` will flag any calls/lists where the final item has no comma.  
For the most part, this is useful and I like the consistency it provides. However, there are two issues with this rule that keep popping up in my code:  

Calls with only one item, that can *never* have another item added, are still flagged. This is a matter of preference, but I find this to actually harm readability, because it suggests that this is a list of items, where other items may be provided afterwards, and that is often not the case. Example:
``` Py
float(
    width
    * height
    / step,  # <- I find this comma more confusing/distracting than helpful
)
```

My second issue with this is MicroPython-specific *(which I realize may be out-of-scope for this project, but it causes problems for me so I think it's worth mentioning)*. This rule also recommends adding commas to constants, which breaks the code until the offending comma is removed. Example:
``` Py
_DISPLAY_PADDING = const(
    (_DISPLAY_HEIGHT - (6 * _SMALL_FONT_HEIGHT + _NUM_LINES * _FONT_HEIGHT))
    // (_NUM_LINES),
)
# SyntaxError: not a constant
```

<br/>

If there was an option to disable the comma rule when only a single value is present, then both of these issues would be solved.  

*Alternatively: if there was an optional list of names to ignore this rule inside of, then both `float` and `const` could be ignored, and both of these issues would be solved.*

MichaReiser:
Hi @echo-lalia 

Hmm, I can see how this is problematic but I think these two cases are exactly what the rule aims to catch

> Calls with only one item, that can never have another item added, are still flagged.

I can see how this might be desired, but it requires very involved analysis for a rule to determine if a callee has only a single parameter. Resolving the callee's signature would make the rule significantly slower. Resolving the callee will only work reliably for typed projects. 

But this also ignores the fact that code evolves. A function that only accepts a single parameter may accept multiple parameters tomorrow. Having the trailing comma in place reduces the diff where the new arguments are added because the formatting of the first argument remains unchanged. 

> My second issue with this is MicroPython-specific (which I realize may be out-of-scope for this project, but it causes problems for me so I think it's worth mentioning). This rule also recommends adding commas to constants, which breaks the code until the offending comma is removed. Example:

This certainly is frustrating. I don't know MicroPython myself and what there goal is but have you considered to open an issue and asking them to make `const` function calls python spec compliant so that they accept optional trailing commas?

[...]

PR #21567, opened 2025-11-21
[flake8-commas] Add an option for allowing single-argument functions (COM812)

### LINKED ISSUE #18258: Add an option for COM812 to disable for single-argument functions?
### Summary

I really want to be able to use the COM812 rule, but our codebase contains a ton of code like the following:

```python
some_string = gettext(
    'This is a long string which '
    'is marked for internationalization'
)

another_function(
    some_long +
    expression +
    gets_passed
)
```

While trailing commas do make a lot of sense for function calls that contain multiple arguments, for functions that take a single argument, requiring them muddies the syntax. I'd love to have an option for the flake8-commas plugin (something like `[lint.flake8-commas] allow-single-arg-function-calls = true`)

I'm happy to do the implementation for this, but I wanted to check in first to see if this is something that would be accepted.

davidt:
## Summary

Resolves #18258
See also #13410

The following patterns currently are incompatible with enabling the COM812 rule:

```python
some_string = gettext(
    'This is a long string which '
    'is marked for internationalization'
)

some_function(
    a_long +
    expression
)
```

The fundamental intent for COM812 is to minimize diffs when adding additional entries to a list, dict, or function call. In the case of the above, without an option, users are limited to one of these options:

```python
# Include a trailing comma. This is ugly, and if for some reason another
# sentence was added, the comma has to be removed.
s = gettext(
    'This is a long string '
    'which is marked for i18n.',
)

# Put the closing `)` on the same line. Adding another argument to this
# requires changing the `)` to a `,`, so there's no minimization of the
# diff here either.
some_function(
    a_long +
    expression)


# Assigning an expression to a variable and then calling the function
# works in some cases, but it bloats the code, and explicitly cannot
# work in the case of `gettext` where the scanner that extracts strings
# for the .po files doesn't know much about Python, and can only extract
# string literals inside of the `gettext(...)` calls.
var = (
    a_long +
    expression
)
some_function(var)

broken = (
    'This string will not be properly '
    'extracted by gettext scanners.'
)
s = gettext(broken)
```

In order to enable use of COM812 in codebases with this pattern, this change adds a new configuration key:
```toml
[lint.flake8-commas]
allow_single_arg_function_calls = true
```

Setting this allow the COM812 rule to flag everything it normally does, but with an exception for a function which is called with a single argument.

## Test Plan

A new test case has been added which uses the existing COM81* test fixture, and the resulting snapshot has been carefully reviewed to verify correctness. One additional case has been added to that file to verify that single-element dictionaries are not affected.


amyreese:
Are there other examples where the workaround (assigning to a variable before calling) isn't viable beyond `gettext()`? Would it be better to just have a single exception to the rule rather than adding a whole new configuration option just to allow using this rule in projects that also use gettext?

davidt:
I don't know for certain, but even in the gettext case, there are multiple potential names that get scanned for (at least `gettext()` and `_()`, historically django also provided a `ugettext()`). It's not exactly a one-and-done solution.

That said, I feel like forcing people to choose between defining a variable and adding a comma is way too strong for a linter rule where the goal of the rule is to make diffs easier to review. In my experience, a single-argument function is likely to always remain a single-argument function, and optimizing for readability is much more important than optimizing for diffs.

MichaReiser:
Thanks for looking into this.

I feel like this option defeats the purpose of the rule. The purpose is to reduce the diff size and, by enabling this rule, you opt-in that this is the top priority, including single-argument function calls.

I don't currently see enough demand for this configuration option and, therefore, close this PR. I'll leave the issue open so that other users can express their interest.

PR #21568, opened 2025-11-21
[ty] Remove brittle constraint set reveal tests

dcreager:
These were added to try to make it clearer that assignability checks will eventually return more detailed answers than true or false. However, the constraint set display rendering is still more brittle than I'd like it to be, and it's more trouble than it's worth to keep them updated with semantically identically but textually different edits. The `static_assert`s are sufficient to check correctness, and we can always add `reveal_type` when needed for further debugging.


dcreager:
This is hopefully uncontroversial, and doesn't touch any code, so I plan to merge it without review once CI goes green

PR #21569, opened 2025-11-21
[ty] Substitute for `typing.Self` when checking protocol members

dcreager:
This patch updates our protocol assignability checks to substitute for any occurrences of `typing.Self` in method signatures, replacing it with the class being checked for assignability against the protocol.

This requires a new helper method on signatures, `apply_self`, which substitutes occurrences of `typing.Self` _without_ binding the `self` parameter.

We also update the `try_upcast_to_callable` method. Before, it would return a `Type`, since certain types upcast to a _union_ of callables, not to a single callable. However, even in that case, we know that every element of the union is a callable. We now return a vector of `CallableType`. (Actually a smallvec to handle the most common case of a single callable; and wrapped in a new type so that we can provide helper methods.) If there is more than one element in the result, it represents a union of callables. This lets callers get at the `CallableType` instances in a more type-safe way. (This makes it easier for our protocol checking code to call the new `apply_self` helper.) We also provide an `into_type` method so that callers that really do want a `Type` can get the original result easily.

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1565,10 +1565,10 @@ impl<'db> Type<'db> {
                 }
             }
             Type::ClassLiteral(class_literal) => {
-                ClassType::NonGeneric(class_literal).into_callable(db)
+                Some(ClassType::NonGeneric(class_literal).into_callable(db))
             }
 
-            Type::GenericAlias(alias) => ClassType::Generic(alias).into_callable(db),
+            Type::GenericAlias(alias) => Some(ClassType::Generic(alias).into_callable(db)),
 
             Type::NewTypeInstance(newtype) => {
                 Type::instance(db, newtype.base_class_type(db)).try_upcast_to_callable(db)
@@ -1576,7 +1576,7 @@ impl<'db> Type<'db> {
 
             // TODO: This is unsound so in future we can consider an opt-in option to disable it.
             Type::SubclassOf(subclass_of_ty) => match subclass_of_ty.subclass_of() {
-                SubclassOfInner::Class(class) => class.into_callable(db),
+                SubclassOfInner::Class(class) => Some(class.into_callable(db)),
                 SubclassOfInner::Dynamic(dynamic) => {
                     Some(CallableTypes::one(CallableType::single(
                         db,
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -1052,7 +1052,7 @@ impl<'db> ClassType<'db> {
     /// Return a callable type (or union of callable types) that represents the callable
     /// constructor signature of this class.
     #[salsa::tracked(cycle_initial=into_callable_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
-    pub(super) fn into_callable(self, db: &'db dyn Db) -> Option<CallableTypes<'db>> {
+    pub(super) fn into_callable(self, db: &'db dyn Db) -> CallableTypes<'db> {
         let self_ty = Type::from(self);
         let metaclass_dunder_call_function_symbol = self_ty
             .member_lookup_with_policy(
@@ -1070,9 +1070,7 @@ impl<'db> ClassType<'db> {
             // https://typing.python.org/en/latest/spec/constructors.html#converting-a-constructor-to-callable
             // by always respecting the signature of the metaclass `__call__`, rather than
             // using a heuristic which makes unwarranted assumptions to sometimes ignore it.
-            return Some(CallableTypes::one(
-                metaclass_dunder_call_function.into_callable_type(db),
-            ));
+            return CallableTypes::one(metaclass_dunder_call_function.into_callable_type(db));
         }
 
         let dunder_new_function_symbol = self_ty.lookup_dunder_new(db);
@@ -1107,7 +1105,7 @@ impl<'db> ClassType<'db> {
             );
 
             if returns_non_subclass {
-                return Some(CallableTypes::one(dunder_new_bound_method));
+                return CallableTypes::one(dunder_new_bound_method);
             }
             Some(dunder_new_bound_method)
         } else {
@@ -1164,13 +1162,13 @@ impl<'db> ClassType<'db> {
 
         match (dunder_new_function, synthesized_dunder_init_callable) {
             (Some(dunder_new_function), Some(synthesized_dunder_init_callable)) => {
-                Some(CallableTypes::from_elements([
+                CallableTypes::from_elements([
                     dunder_new_function,
                     synthesized_dunder_init_callable,
-                ]))
+                ])
             }
             (Some(constructor), None) | (None, Some(constructor)) => {
-                Some(CallableTypes::one(constructor))
+                CallableTypes::one(constructor)
             }
             (None, None) => {
                 // If no `__new__` or `__init__` method is found, then we fall back to looking for
@@ -1186,17 +1184,17 @@ impl<'db> ClassType<'db> {
                 if let Place::Defined(Type::FunctionLiteral(new_function), _, _) =
                     new_function_symbol
                 {
-                    Some(CallableTypes::one(
+                    CallableTypes::one(
                         new_function
                             .into_bound_method_type(db, correct_return_type)
                             .into_callable_type(db),
-                    ))
+                    )
                 } else {
                     // Fallback if no `object.__new__` is found.
-                    Some(CallableTypes::one(CallableType::single(
+                    CallableTypes::one(CallableType::single(
                         db,
                         Signature::new(Parameters::empty(), Some(correct_return_type)),
-                    )))
+                    ))
                 }
             }
         }
@@ -1212,11 +1210,11 @@ impl<'db> ClassType<'db> {
 }
 
 fn into_callable_cycle_initial<'db>(
-    _db: &'db dyn Db,
+    db: &'db dyn Db,
     _id: salsa::Id,
     _self: ClassType<'db>,
-) -> Option<CallableTypes<'db>> {
-    None
+) -> CallableTypes<'db> {
+    CallableTypes::one(CallableType::bottom(db))
 }
 
 impl<'db> From<GenericAlias<'db>> for ClassType<'db> {


crates/ty_python_semantic/resources/mdtest/protocols.md
```diff
@@ -2112,8 +2112,7 @@ static_assert(not is_assignable_to(NominalLegacy, UsesSelf))
 static_assert(not is_assignable_to(NominalWithSelf, NewStyleFunctionScoped))
 static_assert(not is_assignable_to(NominalWithSelf, LegacyFunctionScoped))
 static_assert(is_assignable_to(NominalWithSelf, UsesSelf))
-# TODO: should pass
-static_assert(is_subtype_of(NominalWithSelf, UsesSelf))  # error: [static-assert-error]
+static_assert(is_subtype_of(NominalWithSelf, UsesSelf))
```
dcreager, line 2123: There are other `Self`-related tests in this file that are still TODO, since they also require being able to check against generic protocol members correctly.

dcreager:
Typing conformance result is a new true positive

dcreager:
Ecosystem results look good now, too ­— showing a more precise type in the diagnostics

AlexWaygood:
(I'd love to take a look at this one before it goes in, if that's okay, just to make sure I understand everything that's going on!)

dcreager:
> (I'd love to take a look at this one before it goes in, if that's okay, just to make sure I understand everything that's going on!)

:+1: no problem, I'll wait to merge on an okay from you

crates/ty_python_semantic/src/types.rs
```diff
@@ -1524,17 +1525,20 @@ impl<'db> Type<'db> {
         }
     }
 
-    pub(crate) fn try_upcast_to_callable(self, db: &'db dyn Db) -> Option<Type<'db>> {
+    pub(crate) fn try_upcast_to_callable(self, db: &'db dyn Db) -> CallableTypes<'db> {
```
AlexWaygood: I like that the new return type of this method encodes the fact that if this method returns a union, all union elements will be `Type::Callable`s. But I don't like that it makes it easier to miss the fact that the method is fallible. `self` might not be callable, in which case you have to check for an error. When this method returned an `Option`, you were forced to check for the possibility of an error, but now you have to always make sure you remember to check whether the returned `CallableType`s is empty whenever you call `Type::try_upcast_to_callable()`.

I'd much prefer it if this method could continue to return an `Option` if `self` is not callable (including unions in which some elements are not callable), and if the constructor of `CallableTypes` could enforce that it must always hold at least one element internally. I think that would then mean that you could make the `into_type` method infallible -- you could use `unreachable!()` for the `[]` case in the match inside that method

crates/ty_python_semantic/src/types.rs
```diff
@@ -11068,6 +11106,78 @@ impl<'db> CallableType<'db> {
...
+        CallableTypes(smallvec![callable])
+    }
+
+    pub(crate) fn from_elements(callables: impl IntoIterator<Item = CallableType<'db>>) -> Self {
+        CallableTypes(callables.into_iter().collect())
+    }
+
+    pub(crate) fn exactly_one(self) -> Option<CallableType<'db>> {
+        self.0.into_iter().exactly_one().ok()
+    }
```
AlexWaygood, line 11144: the implementation of `Itertools::exactly_one` is pretty complicated internally because it has to handle arbitrary iterators. I think we can implement this method more simply:

```suggestion
    pub(crate) fn exactly_one(self) -> Option<CallableType<'db>> {
        match self.0.as_slice() {
            [single] => Some(*single),
            _ => None,
        }
    }
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -10950,31 +10977,42 @@ impl get_size2::GetSize for CallableType<'_> {}
 impl<'db> CallableType<'db> {
     /// Create a callable type with a single non-overloaded signature.
     pub(crate) fn single(db: &'db dyn Db, signature: Signature<'db>) -> Type<'db> {
```
AlexWaygood: I sort-of feel like `CallableType::single` should return `CallableType`, and we should have a `Type::simple_callable` method that wraps that in a `Type`, similar to these helper methods we have elsewhere:

https://github.com/astral-sh/ruff/blob/e642874cf1a80598242e3d2becf684ed8d4c4597/crates/ty_python_semantic/src/types.rs#L1276-L1286

but this issue predates your PR, so don't worry about fixing it here :-)

crates/ty_python_semantic/src/types/protocol_class.rs
```diff
@@ -696,14 +693,29 @@ impl<'a, 'db> ProtocolMember<'a, 'db> {
...
+                // `other` directly. This lets us check whether things like `Literal[0]` satisfy a
+                // protocol that includes methods that have `typing.Self` annotations, without
+                // overly constraining `Self` to that specific literal.
+                //
+                // With the new solver, we should be to replace all of this with an additional
+                // constraint that enforces what `Self` can specialize to.
+                let fallback_other = other.literal_fallback_instance(db).unwrap_or(other);
+                attribute_type
+                    .try_upcast_to_callable(db)
+                    .map(|callable| callable.apply_self(db, fallback_other))
```
AlexWaygood: to make sure I understand correctly:
- we previously implicitly upcasted `attribute_type` to `Callable` as part of the `attribute_type.has_relation_to_impl(...)` check, because we have to upcast something like a `Type::FunctionLiteral` if we want to check whether it's assignable to a `Type::Callable`
- but we now `.apply_self` to the `Callable` supertype of `attribute_type` before we check whether `attribute_type <: other`, so we move the `Callable`-upcasting here

crates/ty_python_semantic/src/types.rs
dcreager: I had thought about doing something like this anyway, so I went ahead and did it

crates/ty_python_semantic/src/types.rs
dcreager: Done. This makes a few call sites need `if_some_and` now, where before the `None` case was folded into the `CallableTypes` method, but that's not too bad

crates/ty_python_semantic/src/types/protocol_class.rs
dcreager: Yes exactly!

dcreager:
> Would you be able to add a regression test for this to `protocols.md`?

Done

crates/ty_python_semantic/src/types/protocol_class.rs
AlexWaygood: awesome, thank you!

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -1052,7 +1052,7 @@ impl<'db> ClassType<'db> {
     /// Return a callable type (or union of callable types) that represents the callable
     /// constructor signature of this class.
     #[salsa::tracked(cycle_initial=into_callable_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
-    pub(super) fn into_callable(self, db: &'db dyn Db) -> Type<'db> {
+    pub(super) fn into_callable(self, db: &'db dyn Db) -> Option<CallableTypes<'db>> {
```
AlexWaygood: it looks like this method always returns `Some`. Can't we simplify this by having it return `CallableTypes` rather than `Option<CallableTypes>`? That would better reflect the invariant that a class literal should always be callable (as should a generic alias).

Well, I guess it isn't true that a class will _necessarily_ always be callable if the class's metaclass has `__call__ = None` or something strange. But I don't think we try to model that possibility at the moment.

Anyway, all tests pass if I apply this patch to your branch:

```diff
diff --git a/crates/ty_python_semantic/src/types.rs b/crates/ty_python_semantic/src/types.rs
index 8d894b6c91..2ba63d8256 100644
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -1565,10 +1565,10 @@ impl<'db> Type<'db> {
                 }
             }
             Type::ClassLiteral(class_literal) => {
-                ClassType::NonGeneric(class_literal).into_callable(db)
+                Some(ClassType::NonGeneric(class_literal).into_callable(db))
             }
 
-            Type::GenericAlias(alias) => ClassType::Generic(alias).into_callable(db),
+            Type::GenericAlias(alias) => Some(ClassType::Generic(alias).into_callable(db)),
 
             Type::NewTypeInstance(newtype) => {
                 Type::instance(db, newtype.base_class_type(db)).try_upcast_to_callable(db)
@@ -1576,7 +1576,7 @@ impl<'db> Type<'db> {
 
             // TODO: This is unsound so in future we can consider an opt-in option to disable it.
             Type::SubclassOf(subclass_of_ty) => match subclass_of_ty.subclass_of() {
-                SubclassOfInner::Class(class) => class.into_callable(db),
+                SubclassOfInner::Class(class) => Some(class.into_callable(db)),
                 SubclassOfInner::Dynamic(dynamic) => {
                     Some(CallableTypes::one(CallableType::single(
                         db,
diff --git a/crates/ty_python_semantic/src/types/class.rs b/crates/ty_python_semantic/src/types/class.rs
index bf97c60c93..7de3a86263 100644
--- a/crates/ty_python_semantic/src/types/class.rs
+++ b/crates/ty_python_semantic/src/types/class.rs
@@ -1052,7 +1052,7 @@ impl<'db> ClassType<'db> {
     /// Return a callable type (or union of callable types) that represents the callable
     /// constructor signature of this class.
     #[salsa::tracked(cycle_initial=into_callable_cycle_initial, heap_size=ruff_memory_usage::heap_size)]
-    pub(super) fn into_callable(self, db: &'db dyn Db) -> Option<CallableTypes<'db>> {
+    pub(super) fn into_callable(self, db: &'db dyn Db) -> CallableTypes<'db> {
         let self_ty = Type::from(self);
         let metaclass_dunder_call_function_symbol = self_ty
             .member_lookup_with_policy(
@@ -1070,9 +1070,7 @@ impl<'db> ClassType<'db> {
             // https://typing.python.org/en/latest/spec/constructors.html#converting-a-constructor-to-callable
             // by always respecting the signature of the metaclass `__call__`, rather than
             // using a heuristic which makes unwarranted assumptions to sometimes ignore it.
-            return Some(CallableTypes::one(
-                metaclass_dunder_call_function.into_callable_type(db),
-            ));
+            return CallableTypes::one(metaclass_dunder_call_function.into_callable_type(db));
         }
 
         let dunder_new_function_symbol = self_ty.lookup_dunder_new(db);
@@ -1107,7 +1105,7 @@ impl<'db> ClassType<'db> {
             );
 
             if returns_non_subclass {
-                return Some(CallableTypes::one(dunder_new_bound_method));
+                return CallableTypes::one(dunder_new_bound_method);
             }
             Some(dunder_new_bound_method)
         } else {
@@ -1164,13 +1162,13 @@ impl<'db> ClassType<'db> {
 
         match (dunder_new_function, synthesized_dunder_init_callable) {
             (Some(dunder_new_function), Some(synthesized_dunder_init_callable)) => {
-                Some(CallableTypes::from_elements([
+                CallableTypes::from_elements([
                     dunder_new_function,
                     synthesized_dunder_init_callable,
-                ]))
+                ])
             }
             (Some(constructor), None) | (None, Some(constructor)) => {
-                Some(CallableTypes::one(constructor))
+                CallableTypes::one(constructor)
             }
             (None, None) => {
                 // If no `__new__` or `__init__` method is found, then we fall back to looking for
@@ -1186,17 +1184,17 @@ impl<'db> ClassType<'db> {
                 if let Place::Defined(Type::FunctionLiteral(new_function), _, _) =
                     new_function_symbol
                 {
-                    Some(CallableTypes::one(
+                    CallableTypes::one(
                         new_function
                             .into_bound_method_type(db, correct_return_type)
                             .into_callable_type(db),
-                    ))
+                    )
                 } else {
                     // Fallback if no `object.__new__` is found.
-                    Some(CallableTypes::one(CallableType::single(
+                    CallableTypes::one(CallableType::single(
                         db,
                         Signature::new(Parameters::empty(), Some(correct_return_type)),
-                    )))
+                    ))
                 }
             }
         }
@@ -1212,11 +1210,11 @@ impl<'db> ClassType<'db> {
 }
 
 fn into_callable_cycle_initial<'db>(
-    _db: &'db dyn Db,
+    db: &'db dyn Db,
     _id: salsa::Id,
     _self: ClassType<'db>,
-) -> Option<CallableTypes<'db>> {
-    None
+) -> CallableTypes<'db> {
+    CallableTypes::one(CallableType::bottom(db))
 }
 
 impl<'db> From<GenericAlias<'db>> for ClassType<'db> {
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -10956,34 +10984,44 @@ pub(super) fn walk_callable_type<'db, V: visitor::TypeVisitor<'db> + ?Sized>(
 // The Salsa heap is tracked separately.
 impl get_size2::GetSize for CallableType<'_> {}
 
-impl<'db> CallableType<'db> {
+impl<'db> Type<'db> {
```
AlexWaygood, line 10987: nit: maybe move these new `Type` methods to the big `impl<'db> Type<'db>` block higher up in the file?

AlexWaygood (approved):

crates/ty_python_semantic/src/types.rs
dcreager: Ah, I thought it would be better to have them near the type that they're related to. If we ever extract this out into a separate file these are the `Type` methods that would go with `CallableType` itself.

AlexWaygood: makes sense, I don't have a strong opinion! I agree factoring out `CallableType(s)` into a separate file would be nice at some point

crates/ty_python_semantic/src/types/class.rs
dcreager: Done. I had not thought about using the bottom callable as the cycle fallback — that was why I had kept the `Option` wrapper originally

PR #21570, opened 2025-11-21
[ty] Fix subtraction overflow bug

BurntSushi:
PR #21549 introduced a subtle overflow bug that seemed impossible, but
can empirically happen. This PR fixes it by saturating to zero.

I did try to write a regression test for this, but couldn't manage it.
Instead, I'll attach before-and-after screen recordings.

### DIFF
--- a/crates/ruff_text_size/src/size.rs
+++ b/crates/ruff_text_size/src/size.rs
@@ -106,6 +106,25 @@ impl TextSize {
     pub fn checked_sub(self, rhs: TextSize) -> Option<TextSize> {
         self.raw.checked_sub(rhs.raw).map(|raw| TextSize { raw })
     }
+
+    /// Saturating addition. Returns maximum `TextSize` if overflow occurred.
+    #[inline]
+    #[must_use]
+    pub fn saturating_add(self, rhs: TextSize) -> TextSize {
+        TextSize {
+            raw: self.raw.saturating_add(rhs.raw),
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


BurntSushi:
Before:

https://github.com/user-attachments/assets/9cd9bd36-b73a-42a9-9702-a27f6606a5d8

After:

https://github.com/user-attachments/assets/740de241-9ce8-477b-bf0b-46c8be48543a

Gankra (approved):
Hmm I wonder if this is getting confused about utf8 vs utf16... 😨

RasmusNygren:
Oh that was sloppy by me, not sure how I didn't see that in the playground. :thinking: 

This seems to (at the very least) happen when the cursor is in the middle of a token at the start of the file. This seems to reproduce it:
```rust
let builder = completion_test_builder(
            "\
f<CURSOR>oo = 1
",
);
builder.build();
```

Gankra:
Oh I see, are we subtracting the size of the whole token from the cursor, rather than token-up-to-the-cursor?

RasmusNygren:
We do get the range for the entire token https://github.com/astral-sh/ruff/blob/438ef334d3e71ef0ac70927c799fb159d8f25267/crates/ty_ide/src/completion.rs#L1266

The subtraction is probably a bit off in general then and there might be some other scenarios where we don't suppress suggestions (even though we should) if the cursor is not at the end of the current token. Had a quick check but couldn't figure out a test case that trigger such a scenario even though I feel like they must exist. I'll have a think.

PR #21571, opened 2025-11-21
[ty] Don't suggest things that aren't subclasses of `BaseException` after `raise`

BurntSushi:
This only applies to items that have a type associated with them. That
is, things that are already in scope. For items that don't have a type
associated with them (i.e., suggestions from auto-import), we still
suggest them since we can't know if they're appropriate or not. It's not
quite clear on how best to improve here for the auto-import case. (Short
of, say, asking for the type of each such symbol. But the performance
implications of that aren't known yet.)

Note that because of auto-import, we were still suggesting
`NotImplemented` even though astral-sh/ty#1262 specifically cites it as
the motivating example that we *shouldn't* suggest. This was occuring
because auto-import was including symbols from the `builtins` module,
even though those are actually already in scope. So this PR also gets
rid of those suggestions from auto-import.

Overall, this means that, at least, `raise NotImpl` won't suggest
`NotImplemented`.

Fixes astral-sh/ty#1262

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -9,9 +9,10 @@ use ruff_python_ast::name::Name;
 use ruff_python_codegen::Stylist;
 use ruff_python_parser::{Token, TokenAt, TokenKind, Tokens};
 use ruff_text_size::{Ranged, TextLen, TextRange, TextSize};
+use ty_python_semantic::types::UnionType;
 use ty_python_semantic::{
-    Completion as SemanticCompletion, ModuleName, NameKind, SemanticModel,
-    types::{CycleDetector, Type},
+    Completion as SemanticCompletion, KnownModule, ModuleName, NameKind, SemanticModel,
+    types::{CycleDetector, KnownClass, Type},
 };
 
 use crate::docstring::Docstring;
@@ -81,6 +82,31 @@ impl<'db> Completions<'db> {
     /// Always adds the given completion to this collection.
     fn force_add(&mut self, completion: Completion<'db>) {
         self.items.push(completion);
+    }
+
+    /// Tags completions with whether they are known to be usable in
+    /// a `raise` context.
+    ///
+    /// It's possible that some completions are usable in a `raise`
+    /// but aren't marked by this method. That is, false negatives are
+    /// possible but false positives are not.
+    fn tag_raisable(&mut self) {
+        let raisable_type = UnionType::from_elements(
+            self.db,
+            [
+                KnownClass::BaseException.to_subclass_of(self.db),
+                KnownClass::BaseException.to_instance(self.db),
+            ],
+        );
+        for item in &mut self.items {
+            let Some(ty) = item.ty else { continue };
+            item.is_definitively_raisable = ty.is_assignable_to(self.db, raisable_type);
+        }
+    }
+
+    /// Removes any completion that doesn't satisfy the given predicate.
+    fn retain(&mut self, predicate: impl FnMut(&Completion<'_>) -> bool) {
+        self.items.retain(predicate);
     }
 }
 
@@ -153,6 +179,13 @@ pub struct Completion<'db> {
     /// Whether this item only exists for type checking purposes and
     /// will be missing at runtime
     pub is_type_check_only: bool,
+    /// Whether this item can definitively be used in a `raise` context.
+    ///
+    /// Note that this may not always be computed. (i.e., Only computed
+    /// when we are in a `raise` context.) And also note that if this
+    /// is `true`, then it's definitively usable in `raise`, but if
+    /// it's `false`, it _may_ still be usable in `raise`.
+    pub is_definitively_raisable: bool,
     /// The documentation associated with this item, if
     /// available.
     pub documentation: Option<Docstring>,
@@ -177,6 +210,7 @@ impl<'db> Completion<'db> {
             import: None,
             builtin: semantic.builtin,
             is_type_check_only,
+            is_definitively_raisable: false,
             documentation,
         }
     }
@@ -257,6 +291,7 @@ impl<'db> Completion<'db> {
             import: None,
             builtin: false,
             is_type_check_only: false,
+            is_definitively_raisable: false,
             documentation: None,
         }
     }
@@ -271,6 +306,7 @@ impl<'db> Completion<'db> {
             import: None,
             builtin: true,
             is_type_check_only: false,
+            is_definitively_raisable: false,
             documentation: None,
         }
     }
@@ -364,6 +400,20 @@ pub fn completion<'db>(
         }
     }
 
+    if is_raising_exception(tokens) {
+        completions.tag_raisable();
+
+        // As a special case, and because it's a common footgun, we
+        // specifically disallow `NotImplemented` in this context.
+        // `NotImplementedError` should be used instead. So if we can
+        // definitively detect `NotImplemented`, then we can safely
+        // omit it from suggestions.
+        completions.retain(|item| {
+            let Some(ty) = item.ty else { return true };
+            !ty.is_notimplemented(db)
+        });
+    }
+
     completions.into_completions()
 }
 
@@ -427,7 +477,8 @@ fn add_unimported_completions<'db>(
     let members = importer.members_in_scope_at(scoped.node, scoped.node.start());
 
     for symbol in all_symbols(db, &completions.query) {
-        if symbol.module.file(db) == Some(file) {
+        if symbol.module.file(db) == Some(file) || symbol.module.is_known(db, KnownModule::Builtins)
+        {
             continue;
         }
 
@@ -450,6 +501,7 @@ fn add_unimported_completions<'db>(
             builtin: false,
             // TODO: `is_type_check_only` requires inferring the type of the symbol
             is_type_check_only: false,
+            is_definitively_raisable: false,
             documentation: None,
         });
     }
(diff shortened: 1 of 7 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/94fd6788-c6d2-4bc3-9f5b-b3ed210268ee

crates/ty_ide/src/completion.rs
```diff
@@ -1358,6 +1372,30 @@ fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Opt
     })
 }
 
+/// Returns true when the cursor is after a `raise` keyword.
+fn is_raising_exception(tokens: &[Token]) -> bool {
+    /// The maximum number of tokens we're willing to
+    /// look-behind to find a `raise` keyword.
+    const LIMIT: usize = 10;
```
Gankra, line 1417: Wow I need to review more of your PRs, y'all are wildin' in here. I assume you can't more-normally walk up the AST because autocomplete has to operate in malformed ASTs more?

BurntSushi: I think it's more that I've been operating under the presumption that "if we can do things directly from the tokens, then that's probably better." In part to avoid the problem you bring up (although I've found our AST to be pretty good at dealing with malformed input) but also because it's very cheap to just look at a few tokens. It's possible that doing an AST traversal is also cheap enough. I honestly don't have a great sense of it yet, but for completions we do a lot of "check if we're in context foo, or bar, or baz, or quux..." so the faster each of those checks are (which I assume will continue to grow over time), the better. Although I did just remove a bunch of those checks for imports and consolidated it into one single check.

TL;DR - We could probably look at the AST for cases like this, but I'm starting simple.

MichaReiser: Walking tokens is certainly faster, but walking the AST also isn't crazy expensive (all Ruff lint rules just do that). That's why I'd pick whatever is easier. 

The bigger challenge with our AST is that you can't walk upwards (or sideways). At least not without building another intermediate representation that stores upward pointers for each AST node

One downside of using tokens is that there are cases where the parsed AST and token stream can disagree. E.g. the parser sometimes synthesizes name expressions if a required expression is missing (`e.g. a.` should synthesize a name expression for the attribute). However, we don't synthesize a name token in that case). 

However, our lexing is, to some extent, parser-directed during error recovery. For example, a "normal" lexer would parse the whitespace before `pass` as such whitespace because there's an unclosed `(`. This is not the case for our token stream because the parser will inform the lexer that it will start error recovery after `(` because `pass` isn't a valid argument name and that the lexer should try to lex the current token, assuming it's in a statement context, in which case the whitespace is lexed as an indent.

```py
def test(
	pass
```

https://play.ruff.rs/f1d64305-c9ff-47cc-8edf-3a7216871932

However, there are a few cases where the parsed AST and the token stream can disagree, e.g. the token stream doesn't contain a newline before `pass` even though this is parsed as `def test():\npass`


```py
def test(pass
```

https://play.ruff.rs/d9e759b6-d676-4871-b421-abbd58514fc9

crates/ty_ide/src/completion.rs
```diff
@@ -1358,6 +1372,30 @@ fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Opt
...
+
+    // This only looks for things like `raise foo.bar.baz.qu<CURSOR>`.
+    // Technically, any kind of expression is allowed after `raise`.
+    // But we may not always wanted to treat it specially. So we're
```
```suggestion
    // But we may not always want to treat it specially. So we're
```

crates/ty_ide/src/completion.rs
```diff
@@ -364,6 +369,14 @@ pub fn completion<'db>(
         }
     }
 
+    if is_raising_exception(tokens) {
+        let type_base_exception = KnownClass::BaseException.to_subclass_of(db);
+        completions.retain(|c| {
+            let Some(ty) = c.ty else { return true };
+            ty.is_assignable_to(db, type_base_exception)
+        });
+    }
```
AlexWaygood, line 415: I suppose you can also raise instances of exceptions. So if a user has something like

```py
exception_to_raise = NotImplementedError("foooooooooooooooooooooooooooo")
raise exception_to<CURSOR>
```

we should probably suggest `exception_to_raise` there. That implies that this should be

```suggestion
    if is_raising_exception(tokens) {
        let raisable_type = UnionType::from_elements(
            db,
            [KnownClass::BaseException.to_subclass_of(db), KnownClass::BaseException.to_instance(db)]
        );
        completions.retain(|c| {
            let Some(ty) = c.ty else { return true };
            ty.is_assignable_to(db, raisable_type)
        });
    }
```

basically, the same as this bit in the type-inference builder:

https://github.com/astral-sh/ruff/blob/3410041b4ca8226641b72e2f947ebd612dd2ea28/crates/ty_python_semantic/src/types/infer/builder.rs#L6003-L6025

crates/ty_ide/src/completion.rs
```diff
@@ -1358,6 +1372,30 @@ fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Opt
...
+
+    // This only looks for things like `raise foo.bar.baz.qu<CURSOR>`.
+    // Technically, any kind of expression is allowed after `raise`.
+    // But we may not always wanted to treat it specially. So we're
+    // rather conservative about what we consider "raising an
+    // exception" to be for the purposes of completions. The failure
+    // mode here is that we may wind up suggesting things that
+    // shouldn't be raised. The benefit is that when this heuristic
+    // does work, we won't suggest things that shouldn't be raised.
+    for token in tokens.iter().rev().take(LIMIT) {
```
MichaReiser, line 1427: I think you want to, at least, skip all trivia tokens and parentheses (`(`, `)`)) to properly support

```py
raise (
	a  # comment
	.Error
)
```

AlexWaygood (commented):
Thank you! This looks fantastic.

Though... the more I think about this, the more edge cases I think of. What if a user has something like this?

```py
import jsonschema.exceptions

if condition:
    raise jsons<CURSOR>
```

The user wants to type `raise jsonschema.exceptions.SchemaError`, but we refuse to suggest `jsonschema` here, and even after they've typed `jsonschema.<CURSOR>`, we refuse to suggest `jsonschema.exceptions`, since `jsonschema` and `jsonschema.exceptions` both have module-literal types, and module-literal types aren't assignable to `type[BaseException] | BaseException`.

I can think of two ways round this:
1. Make things more complicated: say that we'll also include module-literal types after `raise` keywords
2. Make things simpler: rather than trying to do a full analysis of which types make sense after a `raise` keyword in the autocompletion engine, just special-case a few symbols (like `NotImplemented`) that we _know_ don't make sense after a `raise` keyword, and make sure that they're removed.
3. Somewhere in the middle? Rather than _removing_ types that aren't assignable to `type[BaseException] | BaseException` from the list of suggestions, just make sure that they're all ranked below symbols that _are_ assignable to `type[BaseException] | BaseException`. And maybe also special-case `NotImplemented` so that it's removed entirely?

I think I'm leaning towards (3). (1) just feels like it has too many edge cases attached to it, because there are lots of ways of creating nested namespaces in Python (not just modules and submodules!). Something like this is unusual in Python, but I can imagine it would be pretty surprising if a user tried it and then found they didn't have the autocompletions they expected (they want to type `raise Namespace.Exception1`):

```py
class Namespace:
    class Exception1(ValueError): ...
    class Exception2(TypeError): ...
    class Exception3(RuntimeError): ...

raise Name<CURSOR>
```

(3) also seems like it'll have a lot less complexity associated with it than (1)

crates/ty_ide/src/completion.rs
BurntSushi: Thanks for that explanation! I didn't know some of that. I think sticking with tokens for now is fine. We can always switch to AST in response to user feedback.

BurntSushi:
@AlexWaygood Great points. I like (3) too. I think that also fits better with symbols from auto-import not having a type associated with it. So it recasts the problem more as a ranking problem than a "don't show symbols" problem.

BurntSushi:
@AlexWaygood OK, this should be updated now to do your (3) idea. Review is appreciated, particularly on how I did the check for `NotImplemented`. :-)

crates/ty_ide/src/completion.rs
```diff
@@ -364,6 +400,27 @@ pub fn completion<'db>(
...
+            db,
+            [
+                KnownClass::NotImplementedType.to_subclass_of(db),
+                KnownClass::NotImplementedType.to_instance(db),
+            ],
+        );
+        completions.retain(|item| {
+            let Some(ty) = item.ty else { return true };
+            !ty.is_assignable_to(db, not_implemented_type)
+        });
```
AlexWaygood, line 414: I think here you can simplify this a bit -- the big reason that folks mix up `NotImplemented` and `NotImplementedError` all the time is that they not only have very similar names, they're also both builtins. `type(NotImplemented)` doesn't really have the same issue, because it's not exposed as a builtin (you have to import the `NotImplementedType` class from the `types` module)

```suggestion
        // As a special case, and because it's a common footgun, we
        // specifically disallow `NotImplemented` in this context.
        // `NotImplementedError` should be used instead. So if we can
        // definitively detect `NotImplemented`, then we can safely
        // omit it from suggestions.
        completions.retain(|item| {
            let Some(ty) = item.ty else { return true };
            !ty.is_notimplemented(db)
        });
```

this will of course require making yet _another_ `ty_python_semantic` API `pub` rather than `pub(crate)` 😆

https://github.com/astral-sh/ruff/blob/a57e29131125bf05db7379e90c7616eec32624fe/crates/ty_python_semantic/src/types.rs#L895-L897

BurntSushi: Yeah I guess I was thinking about

```
class Tricksy(NotImplementedType): pass
```

But I guess it's probably not worth that and better to focus on the specific footgun?

BurntSushi: My approach also covered
```
tricksy = NotImplemented
raise tricksy
```
But probably also not worth handling that?

Especially since both of my examples will get down-ranked anyway.

AlexWaygood: yeah, I think I'd just focus on the known footgun -- apart from anything else, `NotImplementedType` cannot be subclassed, so the user will get an error about `Tricksy` at runtime and ty will also complain about the invalid class definition: https://play.ty.dev/03244fb4-7141-43e4-baf8-240d405ac1d3

AlexWaygood: > My approach also covered
> 
> ```
> tricksy = NotImplemented
> raise tricksy
> ```

that's also covered by my suggested approach! `tricksy` has the same type as the `NotImplemented` singleton here

BurntSushi: Interesting. For whatever reason, the type that completions sees for `tricksy` is `Never`:

```
[crates/ty_ide/src/completion.rs:413:17] item = Completion {
    name: Name("tricksy"),
    insert: None,
    ty: Some(
        Never,
    ),
    kind: None,
    module_name: None,
    import: None,
    builtin: false,
    is_type_check_only: false,
    is_definitively_raisable: true,
    documentation: None,
}
```

Example:

https://github.com/user-attachments/assets/fc93503a-6b8f-496c-98e5-8ee557004c8e

BurntSushi: I think it has something to do with it being used in a `raise` context?

BurntSushi: (This is for my own edification at this point. I updated the PR to use your approach. :))

AlexWaygood: > Interesting. For whatever reason, the type that completions sees for `tricksy` is `Never`:

hmm... that seems incorrect; there must be a bug somewhere. (Where it is I don't know -- probably in `ty_python_semantic` rather than `ty_ide`, though it might well be in `ide_support.rs` rather than one of the "core type-inference" parts of that crate.)

> I think it has something to do with it being used in a `raise` context?

yes, that seems likely

PR #21572, opened 2025-11-21
[ty] Improve lsp handling of hover/goto on imports

Gankra:
* Fixes https://github.com/astral-sh/ty/issues/1011
* Also fixes the fact that we didn't handle `.x` properly *at all* in hover/goto

It turns out all of our import handling completely ignored the `level` (number of relative `.`'s) in a `from ..x.y import z` statement. It was nice seeing how much my understanding of `ty` has improved -- previously this would have all been opaque to me but now it was just, completely glaring and blatant. 

Fixing this required refactoring all the import code to take the importing file into consideration. I ended up refactoring a bunch of code to pass around/require `SemanticModel` more, as it's the natural API for resolving this kind of import (it actually had an API for this that was just... dead code, whoops!).

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -992,7 +992,7 @@ fn find_module_component(
 
 /// Helper to get the module name up to the given component index
 fn import_name(module_name: &str, component_index: usize) -> &str {
-    // We want everything to the left of the nth `.`, i
+    // We want everything to the left of the nth `.`
     // If there's no nth `.` then we want the whole thing.
     let idx = module_name
         .match_indices('.')


crates/ty_ide/src/goto.rs
```diff
@@ -876,27 +868,26 @@ fn convert_resolved_definitions_to_targets(
 
 /// Shared helper to get definitions for an expr (that is presumably a name/attr)
 fn definitions_for_expression<'db>(
-    db: &'db dyn crate::Db,
-    file: ruff_db::files::File,
+    model: &SemanticModel<'db>,
```
MichaReiser, line 871: Sort of funny that `SemanticModel` survived to this day. I introduced it in one of the very earliest prototypes of a type aware linter back when ty was barely able to infer ints. I'm not sure it's the right API but good to see that it's not complete rubbish and is useful in cases like this :)

crates/ty_ide/src/goto.rs
```diff
@@ -947,29 +938,25 @@ pub(crate) fn find_goto_target(
     }
 
     let covering_node = covering_node(parsed.syntax().into(), token.range())
-        .find_first(|node| node.is_identifier() || node.is_expression())
+        .find_first(|node| {
+            node.is_identifier() || node.is_expression() || node.is_stmt_import_from()
```
MichaReiser, line 942: I wonder if it makes sense to change this to `node.is_stmt`. I don't think we ever want to go beyond the closest statement (my, now proven incorrect, assumption was that `expression` is a sufficient boundary but turns out it isn't)

MichaReiser (approved):

crates/ty_ide/src/goto.rs
Gankra: Yeah between this PR and the string annotations one I am Fully embracing it as an actually useful abstraction!

crates/ty_ide/src/goto.rs
Gankra: Hmm I'll think about it

PR #21573, opened 2025-11-21
[ty] make implicit submodule imports re-exported

Gankra:
Thus they work in `.pyi` files

Closes https://github.com/astral-sh/ty/issues/1609

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
@@ -214,8 +214,8 @@ reveal_type(mypackage.imported.X)  # revealed: Unknown
 
 ## Relative `from` Import of Nested Submodule in `__init__`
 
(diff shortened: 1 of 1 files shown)


carljm (approved):

carljm:
Ecosystem looks great to me here -- the few added diagnostics just look like cases where we now have type information instead of `Unknown`.

Gankra:
oh THERE's the missing 400 fixes I thought one of my followups would hit!

The one followup I forgot to try 😅

carljm:
WTF, is it a new rule that everything must regress performance on pydantic?

This doesn't even seem to cause any new diagnostics on pydantic...

carljm:
I tried to explore this regression and I can't even reproduce it locally. Locally when I run `main` vs this branch, single-threaded, on pydantic, with a release build, via hyperfine, I consistently see this branch as a couple percent _faster_ than main. So I'm going to acknowledge the regression and go ahead and land this.

PR #21576, opened 2025-11-22
[ty] Fix edge cases for autocomplete suppressions in variable bindings

RasmusNygren:
## Summary
This handles a few edge cases related to https://github.com/astral-sh/ty/issues/1563  that was not fully handled in https://github.com/astral-sh/ruff/pull/21549

Only using the individual `Parameter` nodes in the AST to determine parameter bindings is not enough when you're in a state where the syntax is invalid. For example when the name of the parameter clashes with a keyword. In those scenarios we now go up a level and inspect the `Parameters` node to determine if we're still binding variables.

## Test Plan

New tests and ty-playground sanity-check.

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -1407,6 +1407,24 @@ fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Opt
             type_param.name.range.contains_range(range)
         }
         ast::AnyNodeRef::StmtFor(stmt_for) => stmt_for.target.range().contains_range(range),
+        // The AST does not produce `ast::AnyNodeRef::Parameter` nodes for keywords
+        // or otherwise invalid syntax. Rather they are captured in a
+        // `ast::AnyNodeRef::Parameters` node as "empty space". To ensure
+        // we still suppress suggestions even when the syntax is technically
+        // invalid we extract the token under the cursor and check if it makes
+        // up that "empty space" inside the Parameters Node. If it does, we know
+        // that we are still binding variables, just that the current state is
+        // syntatically invalid. Hence we suppress autocomplete suggestons
+        // also in those cases.
+        ast::AnyNodeRef::Parameters(params) => {
+            if !params.range.contains_range(range) {
+                return false;
+            }
+            params
+                .iter()
+                .map(|param| param.range())
+                .all(|r| !r.contains_range(range))
+        }
         _ => false,
     })
 }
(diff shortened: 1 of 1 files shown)


crates/ty_ide/src/completion.rs
```diff
@@ -1341,7 +1341,18 @@ fn is_in_definition_place(
...
+        // middle of it (pa<CURSOR>ram). To correctly suggest completions
+        // based on the cursor's current position, we must only consider
+        // the characters in the token that are left of the cursor.
+        // E.g. we should only count "pa" if the state is "pa<CUSROR>ram".
+        let typed_len = typed
+            .text_len()
+            .min(tokens.last().map_or(typed.text_len(), |token| {
+                offset.saturating_sub(token.start())
+            }));
+        let start = offset.saturating_sub(typed_len);
```
RasmusNygren: This is technically not relevant to this PR but I realised from https://github.com/astral-sh/ruff/pull/21570 that this calculation was logically incorrect. `typed.textlen()` is the length of the entire token, even if the cursor is currently in the middle of the token which is not quite what we want.
 
 I have not managed to write a test based on suggestions that highlights this bug, but by calling `is_in_variable_binding` directly with different inputs it became obvious that it was previously incorrect.

crates/ty_ide/src/completion.rs
```diff
@@ -1341,7 +1341,18 @@ fn is_in_definition_place(
 /// E.g. naming a parameter, type parameter, or `for` <name>).
 fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Option<&str>) -> bool {
     let range = if let Some(typed) = typed {
-        let start = offset.saturating_sub(typed.text_len());
+        let tokens = tokens_start_before(parsed.tokens(), offset);
+        // `typed` contains the entire token even if the cursor sits in the
+        // middle of it (pa<CURSOR>ram). To correctly suggest completions
+        // based on the cursor's current position, we must only consider
+        // the characters in the token that are left of the cursor.
+        // E.g. we should only count "pa" if the state is "pa<CUSROR>ram".
```
BurntSushi: This actually wasn't my intent, but maybe I was wrong. In particular, rust-analyzer and pylance work how you describe: it only considers the characters before the cursor when the cursor splits a token. ty doesn't. It uses the entire token as the query.

I initially went this route because when you select a completion, clients seem to usually replace the _entire_ token with the selected suggestion. So in my mind, it made sense for the query to also correspond to the entire token. With that said, while rust-analyzer replaces the entire token, pylance does not.

I do kind of feel like only taking the text before the cursor is perhaps more intuitive and even more useful. (I'm undecided on the token replacement strategy though.)

So to that end, perhaps the code that extracts the typed text should cut itself off based on the offset? Then I think that would fix this without needing to do this extra arithmetic here.

crates/ty_ide/src/completion.rs
```diff
@@ -1354,6 +1365,24 @@ fn is_in_variable_binding(parsed: &ParsedModuleRef, offset: TextSize, typed: Opt
...
+        // also in those cases.
+        ast::AnyNodeRef::Parameters(params) => {
+            if !params.range.contains_range(range) {
+                return false;
+            }
+            params
+                .iter()
+                .map(|param| param.range())
+                .all(|r| !r.contains_range(range))
+        }
```
BurntSushi, line 1427: This makes sense to me. Nice find!

crates/ty_ide/src/completion.rs
RasmusNygren: Right, I agree with you that only considering text to the left of the cursor feels more intuitive, independently from whether the client replaces the entire token or not. Given that, it would be quite nice to only extract the relevant bit from the start, so we can be consistent without any special handling.

Would you like me to open a separate PR (maybe even an issue) for that (as I jumbled two issues into one PR here) or are you fine if I execute on that directly here?

BurntSushi: I think just doing that directly here is fine. :-) Thank you!

PR #21577, opened 2025-11-22
[ty] Implement lsp support for string annotations

Gankra:
Fixes https://github.com/astral-sh/ty/issues/1009

## Summary

This adds support for:

* semantic-tokens (syntax highlighting)
* goto-type **(partially implemented, but want to land as-is)**
* goto-declaration
* goto-definition (falls out of goto-declaration)
* hover **(limited by goto-type)**
* find-references
* rename-references (falls out of find-references)

There are 3 major things being introduced here:

* `TypeInferenceBuilder::string_annotations` is a `FxHashSet` of exprs which were determined to be string annotations during inference. It's bubbled up in `extras` to hopefully minimize the overhead as in most contexts it's empty.
  * Very happy to hear if this is too hacky and if I should do something better, but it's IMO important that we get an authoritative answer on whether something is a string annotation or not. 
* `SemanticModel::enter_string_annotation` checks if the expr was marked by `TypeInferenceBuilder::string_annotations` and then parses the subast and produces a sub-SemanticModel that sets `SemanticModel::in_string_annotation_expr`. This expr will be used by the model whenever we need to query e.g. the scope of the current expression (otherwise the code will constantly panic as the subast nodes are not in the current File's AST)
  * This hazard consequently encouraged me to refactor a bunch of code to replace uses of file/db with SemanticModel to minimize hazards (it is no longer as safe to randomly materialize a SemanticModel in the middle of analysis, you need to thread through the one you have in case it has `in_string_annotation_expr` set).
* `GotoTarget::StringAnnotationSubexpr` (and a semantic-tokens impl) which involves invoking `SemanticModel::enter_string_annotation` before invoking the same kind of subroutine a normal expression would.
  * goto-type (and consequently displaying the type in hover) is the main hole here, because we can only get the type iff the string annotation is the entire subexpression (i.e. we can get the type of `"int"` but not the parts of `"int | str"`). This is shippable IMO.

## Test Plan

Messed around in IDE, wrote a ton of tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/ide_support.rs
+++ b/crates/ty_python_semantic/src/types/ide_support.rs
@@ -990,7 +990,7 @@ pub fn call_signature_details<'db>(
     // Use into_callable to handle all the complex type conversions
     if let Some(callable_type) = func_type
         .try_upcast_to_callable(model.db())
-        .map(|callables| callables.into_type(db))
+        .map(|callables| callables.into_type(model.db()))
     {
         let call_arguments =
             CallArguments::from_arguments(&call_expr.arguments, |_, splatted_value| {


Gankra:
...why did a bunch of random comments get deleted. Oh well, it's still just a Draft 🙃

Gankra:
omg I think when I asked rust-analyzer to add missing members to unwildcard a match it deleted comments on all the existing branches 😭

crates/ty_python_semantic/src/semantic_model.rs
```diff
@@ -227,6 +232,56 @@ impl<'db> SemanticModel<'db> {
...
+            return None;
+        }
+
+        // Parse the sub-AST and create a semantic model that knows it's in a sub-AST
+        //
+        // The string_annotation will be used as the expr/node for any query that needs
+        // to look up a node in the AST to prevent panics, because these sub-AST nodes
+        // are not in the File's AST!
+        let source = source_text(self.db, self.file);
+        let string_literal = string_expr.as_single_part_string()?;
```
Gankra, line 313: Pre-empting Micha: yes this does fail for various things but the type checker also does this so it's fine

crates/ty_ide/src/goto.rs
```diff
@@ -322,14 +340,30 @@ impl GotoTarget<'_> {
...
+                let subnode = covering_node(submod.into(), *subrange).node();
+
+                // The type checker knows the type of the full annotation but nothing else
+                if AnyNodeRef::from(&*submod.body) == subnode {
+                    string_expr.inferred_type(model)
+                } else {
+                    // TODO: force the typechecker to tell us its secrets
+                    // (it computes but then immediately discards these types)
+                    return None;
+                }
```
Gankra, line 359: This is like, the one caveat of this design. Fixing this would require awful hacks or more non-trivial changes to inference.

Gankra: (awful hacks would be like... looking up types by-name)

Gankra:
I want to make it clear that this is a "high risk" change in that any bugs in this could easily lead to LSP panics. However we could potentially have this functionality toggled behind a feature flag (literally just make `enter_string_annotation` always return None). IMO it should be safe to get into this week's release, if there's issues they should be found pretty fast by people kicking the tires, and then we can trivially disable it if so.

carljm:
The new string-annotations tracking in inference looks good to me! Didn't review the rest.

Gankra:
I'm reasonably confident with the rest of the code. If we wanna get this into the release tomorrow this would be fine to merge without further review (but I'll wait for review for now).

crates/ty_python_semantic/src/semantic_model.rs
```diff
@@ -227,6 +244,82 @@ impl<'db> SemanticModel<'db> {
...
+    /// If we're analyzing a string annotation, it will return the string literal's node.
+    /// Otherwise it will return the input.
+    pub fn node_in_ast<'a>(&'a self, node: ast::AnyNodeRef<'a>) -> ast::AnyNodeRef<'a> {
+        if let Some(string_annotation) = &self.in_string_annotation_expr {
+            (&**string_annotation).into()
+        } else {
+            node
+        }
+    }
+
```
MichaReiser, line 259: I'm not sure I fully understand those methods and when I'm supposed to use them. 

Like in what sense are they safe where the in-stringified-annotation nodes aren't? When should I use this method over using the ast node directly?

crates/ty_python_semantic/src/semantic_model.rs
MichaReiser: lol

crates/ty_python_semantic/src/semantic_model.rs
```diff
@@ -315,7 +408,7 @@ pub trait HasDefinition {
 impl HasType for ast::ExprRef<'_> {
     fn inferred_type<'db>(&self, model: &SemanticModel<'db>) -> Type<'db> {
         let index = semantic_index(model.db, model.file);
-        let file_scope = index.expression_scope_id(self);
+        let file_scope = index.expression_scope_id(&model.expr_ref_in_ast(*self));
         let scope = file_scope.to_scope_id(model.db, model.file);
 
         infer_scope_types(model.db, scope).expression_type(*self)
```
MichaReiser, line 414: So just for my understanding. We now store the expression types of each expression in a stringified literal within the string-literal's scope? 

This is cool!

crates/ty_ide/src/goto.rs
MichaReiser: Okay, maybe it doesn't do yet what I thought it would.

Could we change `ExpressionNodeKey` to be an enum:

```rust
enum ExpressionNodeKey {
    Primary(NodeKey),
    StringifiedAnnotation(Box<{ parent: NodeKey, inner: NodeKey }>)
}
```

(It might even ahve to be a `ThinVec` if stringified annotationos can be nested multiple levels. 

We could then start storing the inferred expressions in type inference and look them up here


Happy to leave this as a follow up

Gankra: They cannot be nested, thankfully.

I agree this would be great, are NodeKeys this stable?

(Definitely followup material since it could have more non-trivial perf impact)

crates/ty_python_semantic/src/semantic_model.rs
Gankra: The rough heuristic in my mind is "if the node might be handed to the inference subsystem (i.e. trying to query it's scope or type)" (which SemanticModel gatekeeps). If you ever hand these sub-AST nodes to the inference subsystem it will immediately panic because the node isn't part of the file's AST.

crates/ty_ide/src/goto.rs
MichaReiser: `NodeKey`'s are stable for as long as the source text doesn't change. We simply iterate over the AST and assign an index to each node (see `parsed_module`)

PR #21578, opened 2025-11-22
[ty] Improve concise diagnostics for invalid exceptions when a user catches a tuple of objects

AlexWaygood:
## Summary

The concise diagnostic is currently just "Invalid tuple caught in an exception handler", which doesn't give the user sufficient information to debug the problem

## Test Plan

Manual testing, since we don't currently have any tests for concise diagnostics:

image


sharkdp:
> Manual testing, since we don't currently have any tests for concise diagnostics:

Mdtests?

AlexWaygood:
> > Manual testing, since we don't currently have any tests for concise diagnostics:
> 
> Mdtests?

Oh, right, of course! They're a bit of a pain to update if you assert on the error message, though

PR #21579, opened 2025-11-22
[ty] Add `from` imports to `imported_modules` *if* the module being imported is not relative to the current module

AlexWaygood:
## Summary

Make it so that we recognise that `foo` is available as an attribute on `bar` if the user has something like

```py
import bar
from foo.bar import whatever
```

## Test Plan

just manual testing so far...


AlexWaygood:
and we're back to tie-breaking hell...

PR #21580, opened 2025-11-22
[ty] Fix rendering of unused suppression diagnostic

MichaReiser:
## Summary

Set the message passed to `context.report_unchecked` as the primary diagnostic message instead of using `""`. 

## Test plan

See changed snapshot


AlexWaygood:
The pydantic benchmark seems to have become flaky recently for some reason (see also https://github.com/astral-sh/ruff/pull/21573#issuecomment-3564780095)

PR #21582, opened 2025-11-22
[ty] Add 'remove unused ignore comment' code action

MichaReiser:
## Summary

This PR adds a code action to remove unused ignore comments.

This PR also includes some infrastructure boilerplate to set up code actions in the editor:

* Extend `snapshot-diagnostics` to render fixes
* Render fixes when using `--output-format=full`
* Hook up edits and the code action request in the LSP
* Add the `Unnecessary` tag to `unused-ignore-comment` diagnostics
* Group multiple unused codes into a single diagnostic

The same fix can be used on the CLI once we add `ty fix` 

Note: `unused-ignore-comment` is currently disabled by default.

https://github.com/user-attachments/assets/f9e21087-3513-4156-85d7-a90b1a7a3489

### DIFF
--- a/crates/ty/docs/rules.md
+++ b/crates/ty/docs/rules.md
@@ -2120,7 +2120,7 @@ old_func()  # emits [deprecated] diagnostic
 Default level: <a href="../rules.md#rule-levels" title="This lint has a default level of 'warn'."><code>warn</code></a> ·
 Added in <a href="https://github.com/astral-sh/ty/releases/tag/0.0.1-alpha.1">0.0.1-alpha.1</a> ·
 <a href="https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20ignore-comment-unknown-rule" target="_blank">Related issues</a> ·
-<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L45" target="_blank">View source</a>
+<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L47" target="_blank">View source</a>
 </small>
 
 
@@ -2151,7 +2151,7 @@ a = 20 / 0  # ty: ignore[division-by-zero]
 Default level: <a href="../rules.md#rule-levels" title="This lint has a default level of 'warn'."><code>warn</code></a> ·
 Added in <a href="https://github.com/astral-sh/ty/releases/tag/0.0.1-alpha.1">0.0.1-alpha.1</a> ·
 <a href="https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20invalid-ignore-comment" target="_blank">Related issues</a> ·
-<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L70" target="_blank">View source</a>
+<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L72" target="_blank">View source</a>
 </small>
 
 
@@ -2536,7 +2536,7 @@ print(x)  # NameError: name 'x' is not defined
 Default level: <a href="../rules.md#rule-levels" title="This lint has a default level of 'ignore'."><code>ignore</code></a> ·
 Added in <a href="https://github.com/astral-sh/ty/releases/tag/0.0.1-alpha.1">0.0.1-alpha.1</a> ·
 <a href="https://github.com/astral-sh/ty/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20unused-ignore-comment" target="_blank">Related issues</a> ·
-<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L20" target="_blank">View source</a>
+<a href="https://github.com/astral-sh/ruff/blob/main/crates%2Fty_python_semantic%2Fsrc%2Fsuppression.rs#L22" target="_blank">View source</a>
 </small>
 
 


crates/ty_python_semantic/Cargo.toml
```diff
@@ -13,6 +13,7 @@ license = { workspace = true }
 [dependencies]
 ruff_db = { workspace = true }
 ruff_annotate_snippets = { workspace = true }
+ruff_diagnostics = { workspace = true }
```
MichaReiser, line 16: We should probably fold `ruff_diagnostics` into `ruff_db`. There's really not much in there but this seemed unrelated to this PR

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -634,25 +747,20 @@ impl<'a> SuppressionsBuilder<'a> {
             // `ty: ignore[a, b]`
             Some(codes) => {
                 for code_range in codes {
-                    let code = &self.source[*code_range];
-                    let range = if codes.len() == 1 {
-                        comment.range
-                    } else {
-                        *code_range
-                    };
```
MichaReiser, line 642: This is now handled by the unused suppression lint itself

crates/ty_python_semantic/resources/mdtest/suppressions/ty_ignore.md
```diff
@@ -127,7 +130,7 @@ a = 10 / 0  # ty: ignore[*-*]
 <!-- blacken-docs:off -->
 
 ```py
-a = 10 / 0  # ty: ignore[division-by-zero]       
+a = 10 / 0  # ty: ignore[division-by-zero]
             #                               ^^^^^^ trailing whitespace
```
AlexWaygood: the comment here suggests to me that the trailing whitespace might have been intentional :P

MichaReiser: Thanks, yes they are (which is why I added the comment). It's so annoying that I can't tell my editors: Don't trim trailing whitespaces here

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -184,6 +186,8 @@ fn check_invalid_suppression(context: &mut CheckSuppressionsContext) {
 ///
 /// Does nothing if the [`UNUSED_IGNORE_COMMENT`] rule is disabled.
 fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
+    use std::fmt::Write as _;
```
AlexWaygood: nit: is there a reason to import this locally rather than globally?

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -222,38 +226,138 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
...
+                            diag.sub(SubDiagnostic::new(
+                                SubDiagnosticSeverity::Help,
+                                "Remove the unused suppression code",
+                            ));
+                        } else {
+                            diag.sub(SubDiagnostic::new(
+                                SubDiagnosticSeverity::Help,
+                                "Remove the unused suppression codes",
+                            ));
+                        }
```
AlexWaygood, line 339: you can simplify the code by using the `DiagnosticGuard::help()` method:

```diff
diff --git a/crates/ty_python_semantic/src/suppression.rs b/crates/ty_python_semantic/src/suppression.rs
index caa9796295..16db479d13 100644
--- a/crates/ty_python_semantic/src/suppression.rs
+++ b/crates/ty_python_semantic/src/suppression.rs
@@ -3,8 +3,7 @@ use crate::lint::{GetLintError, Level, LintMetadata, LintRegistry, LintStatus};
 use crate::types::TypeCheckDiagnostics;
 use crate::{Db, declare_lint, lint::LintId};
 use ruff_db::diagnostic::{
-    Annotation, Diagnostic, DiagnosticId, IntoDiagnosticMessage, Severity, Span, SubDiagnostic,
-    SubDiagnosticSeverity,
+    Annotation, Diagnostic, DiagnosticId, IntoDiagnosticMessage, Severity, Span,
 };
 use ruff_db::{files::File, parsed::parsed_module, source::source_text};
 use ruff_diagnostics::{Edit, Fix};
@@ -332,17 +331,11 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
                         );
                         diag.set_fix(Fix::safe_edit(Edit::range_deletion(fix_range)));
 
-                        if unused_codes.is_empty() {
-                            diag.sub(SubDiagnostic::new(
-                                SubDiagnosticSeverity::Help,
-                                "Remove the unused suppression code",
-                            ));
+                        diag.help(if unused_codes.is_empty() {
+                            "Remove the unused suppression code"
                         } else {
-                            diag.sub(SubDiagnostic::new(
-                                SubDiagnosticSeverity::Help,
-                                "Remove the unused suppression codes",
-                            ));
-                        }
+                            "Remove the unused suppression codes"
+                        });
                     }
 
                     continue;
@@ -377,10 +370,7 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
             .unwrap()
             .push_tag(ruff_db::diagnostic::DiagnosticTag::Unnecessary);
         diag.set_fix(remove_comment_fix(suppression, &source));
-        diag.sub(SubDiagnostic::new(
-            SubDiagnosticSeverity::Help,
-            "Remove the unused suppression comment",
-        ));
+        diag.help("Remove the unused suppression comment");
     }
 }
```

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -963,6 +1079,40 @@ enum ParseErrorKind {
...
+    let comment_end = suppression.comment_range.end();
+    let comment_start = suppression.comment_range.start();
+    let after_comment = &source[comment_end.to_usize()..];
+
+    if !after_comment.starts_with(['\n', '\r']) {
+        // For example: `# ty: ignore # fmt: off`
+        // Don't remove the trailing whitespace up to the `ty: ignore` comment
+        return Fix::safe_edit(Edit::range_deletion(TextRange::new(
+            comment_start,
+            suppression.comment_range.end(),
```
AlexWaygood:
```suggestion
            comment_end,
```

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -963,6 +1079,40 @@ enum ParseErrorKind {
...
+        match c {
+            '\n' | '\r' => break,
+            c if c.is_whitespace() => leading_len += c.text_len(),
+            _ => break,
+        }
+    }
+
+    Fix::safe_edit(Edit::range_deletion(TextRange::new(
+        comment_start - leading_len,
+        suppression.comment_range.end(),
```
AlexWaygood:
```suggestion
        comment_end,
```

AlexWaygood (commented):
Very excited to see this happen! A few nits from skimming

crates/ty_python_semantic/src/suppression.rs
MichaReiser: Mainly laziness. RustRover can't auto import `Write` and having to scroll to the top and down again breaks my entire flow.

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -222,38 +226,132 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
...
+                if !includes_first_code || !includes_last_code {
+                    let mut codes = format!("'{}'", lint.name());
+                    for code in &unused_codes {
+                        let _ = write!(&mut codes, ", '{code}'", code = code.name());
+                    }
+
+                    if let Some(diag) = context.report_unchecked(
+                        &UNUSED_IGNORE_COMMENT,
+                        TextRange::new(suppression.range.start(), current.range.end()),
+                    ) {
```
AlexWaygood, line 300: you could use early-return style here to reduce the indentation:

```diff
diff --git a/crates/ty_python_semantic/src/suppression.rs b/crates/ty_python_semantic/src/suppression.rs
index ab246c1b0a..852fa5e1bd 100644
--- a/crates/ty_python_semantic/src/suppression.rs
+++ b/crates/ty_python_semantic/src/suppression.rs
@@ -294,49 +294,51 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
                         let _ = write!(&mut codes, ", '{code}'", code = code.name());
                     }
 
-                    if let Some(diag) = context.report_unchecked(
+                    let Some(diag) = context.report_unchecked(
                         &UNUSED_IGNORE_COMMENT,
                         TextRange::new(suppression.range.start(), current.range.end()),
-                    ) {
-                        let mut diag = diag.into_diagnostic(format_args!(
-                            "Unused `{kind}` directive: {codes}",
-                            kind = suppression.kind
-                        ));
-
-                        diag.primary_annotation_mut()
-                            .unwrap()
-                            .push_tag(ruff_db::diagnostic::DiagnosticTag::Unnecessary);
-
-                        // Delete everything up to the start of the next code.
-                        let trailing_len: TextSize = source[current.range.end().to_usize()..]
+                    ) else {
+                        continue;
+                    };
+
+                    let mut diag = diag.into_diagnostic(format_args!(
+                        "Unused `{kind}` directive: {codes}",
+                        kind = suppression.kind
+                    ));
+
+                    diag.primary_annotation_mut()
+                        .unwrap()
+                        .push_tag(ruff_db::diagnostic::DiagnosticTag::Unnecessary);
+
+                    // Delete everything up to the start of the next code.
+                    let trailing_len: TextSize = source[current.range.end().to_usize()..]
+                        .chars()
+                        .take_while(|c: &char| c.is_whitespace() || *c == ',')
+                        .map(TextLen::text_len)
+                        .sum();
+
+                    // If we delete the last codes before `]`, ensure we delete any trailing comma
+                    let leading_len: TextSize = if includes_last_code {
+                        source[..suppression.range.start().to_usize()]
                             .chars()
+                            .rev()
                             .take_while(|c: &char| c.is_whitespace() || *c == ',')
                             .map(TextLen::text_len)
-                            .sum();
-
-                        // If we delete the last codes before `]`, ensure we delete any trailing comma
-                        let leading_len: TextSize = if includes_last_code {
-                            source[..suppression.range.start().to_usize()]
-                                .chars()
-                                .rev()
-                                .take_while(|c: &char| c.is_whitespace() || *c == ',')
-                                .map(TextLen::text_len)
-                                .sum()
-                        } else {
-                            TextSize::default()
-                        };
-
-                        let fix_range = TextRange::new(
-                            suppression.range.start() - leading_len,
-                            current.range.end() + trailing_len,
-                        );
-                        diag.set_fix(Fix::safe_edit(Edit::range_deletion(fix_range)));
-
-                        if unused_codes.is_empty() {
-                            diag.help("Remove the unused suppression code");
-                        } else {
-                            diag.help("Remove the unused suppression codes");
-                        }
+                            .sum()
+                    } else {
+                        TextSize::default()
+                    };
+
+                    let fix_range = TextRange::new(
+                        suppression.range.start() - leading_len,
+                        current.range.end() + trailing_len,
+                    );
+                    diag.set_fix(Fix::safe_edit(Edit::range_deletion(fix_range)));
+
+                    if unused_codes.is_empty() {
+                        diag.help("Remove the unused suppression code");
+                    } else {
+                        diag.help("Remove the unused suppression codes");
                     }
```

crates/ty_python_semantic/src/suppression.rs
```diff
@@ -222,38 +226,132 @@ fn check_unused_suppressions(context: &mut CheckSuppressionsContext) {
...
+                                .sum()
+                        } else {
+                            TextSize::default()
+                        };
+
+                        let fix_range = TextRange::new(
+                            suppression.range.start() - leading_len,
+                            current.range.end() + trailing_len,
+                        );
+                        diag.set_fix(Fix::safe_edit(Edit::range_deletion(fix_range)));
```
AlexWaygood, line 333: `type: ignore` comments, unlike `ty: ignore` comments, might have other consumers apart from ty (the user might run multiple type checkers over their codebase, especially if they're maintaining a library). So I'd lean towards marking this an unsafe fix for now.

I know we've discussed that we want better configuration options around whether ty should respect `type: ignore` comments at all (or only look at `ty: ignore` comments, and if we had those configuration options I think this could possibly be made a safe fix... but we don't have those configuration options yet ;)

AlexWaygood (approved):
🚀

(Didn't review the `ty_server` code in depth)

crates/ty_python_semantic/src/suppression.rs
MichaReiser: I'd prefer to solve this as part of adding a new setting or splitting unused ignore comment into multiple rules.

crates/ty_python_semantic/src/suppression.rs
MichaReiser: I prefer how it is now. It makes it clearer that all code paths in this branch ultimately call `continue`

crates/ty_python_semantic/src/suppression.rs
AlexWaygood: I agree we should do that. To me, marking these as unsafe fixes in the meantime seems like the fairer thing to do for our users until we do that.

MichaReiser: You can't really enable `unused-ignore-comment` if you're using another type checker anyway, we don't support a `--fix` command, nor applicabilities, and triggering it from the IDE is a manual action.

crates/ty_python_semantic/Cargo.toml
Gankra: I tried this and it ran into circular reference nastiness (stuff depends on `ruff_diagnostics` that `ruff_db` in turn depends on, I guess?). I opted to just `pub use` the relevant items in `ruff_db`.

PR #14946, opened 2024-12-12
Bind top-most parent when importing nested module

dcreager:
When importing a nested module, we were correctly creating a binding for the top-most parent, but we were binding that to the nested module, not to that parent module.  Moreover, we weren't treating those submodules as members of their containing parents.  This PR addresses both issues, so that nested imports work as expected.

As discussed in ~Slack~ whatever chat app I find myself in these days :smile:, this requires keeping track of which modules have been imported within the current file, so that when we resolve member access on a module reference, we can see if that member has been imported as a submodule.  If so, we return the submodule reference immediately, instead of checking whether the parent module's definition defines the symbol.

This is currently done in a flow insensitive manner.  The `SemanticIndex` now tracks all of the modules that are imported (via `import`, not via `from...import`).  The member access logic mentioned above currently only considers module imports in the file containing the attribute expression.

PR #21583, opened 2025-11-22
[ty] Implement `from x.y import z` ensuring `x.y` resolves in the current file

### LINKED ISSUE #1654: `docker.types` reported as `possibly-missing-attribute` after `import docker; from docker.types import Ulimit`
### Question

Hello,

I have the following `test.py`:
```python
import docker
from docker.types import Ulimit

docker.types.Ulimit(name='nproc', soft=1024)
Ulimit(name='nproc', soft=1024)
```

and I have `docker==7.1.0` installed in an activte venv.

When I run `ty` (ty 0.0.1-alpha.28 (8c342496a 2025-11-25)) I get:
```
> ty check test.py
warning[possibly-missing-attribute]: Submodule `types` may not be available as an attribute on module `docker`
 --> test.py:4:1
  |
2 | from docker.types import Ulimit
3 |
4 | docker.types.Ulimit(name='nproc', soft=1024)
  | ^^^^^^^^^^^^
5 | Ulimit(name='nproc', soft=1024)
  |
help: Consider explicitly importing `docker.types`
info: rule `possibly-missing-attribute` is enabled by default

Found 1 diagnostic
```

pyright (pyright 1.1.407) also complains:
```
pyright test.py
/private/tmp/q/test.py
  /private/tmp/q/test.py:4:8 - error: "types" is not a known attribute of module "docker" (reportAttributeAccessIssue)
```

After installing stubs, mypy (mypy 1.18.2 (compiled: yes)) doesn't complain:
```
> mypy test.py
Success: no issues found in 1 source file
```

python (Python 3.12.11 on macOS) itself doesn't complain:
```
> python test.py 
>
```

Looking at https://github.com/docker/docker-py/blob/7.1.0/docker/__init__.py there is no import that create `types` but there is an adjacent `types/` directory that contains https://github.com/docker/docker-py/blob/7.1.0/docker/types/__init__.py and that imports `Ulimit`.

I have done searches https://github.com/astral-sh/ty/issues?q=unresolved-attribute and https://github.com/astral-sh/ty/issues?q=possibly-missing-attribute but I can't find an existing duplicate but I'm sure one must already exist. Can you point me in the right direction?

### Version

ty 0.0.1-alpha.28 (8c342496a 2025-11-25)

AlexWaygood:
## Summary

Implement `from x.y import z` ensuring `x.y` resolves in the current file if a `y` attribute isn't otherwise available on `x`. This is implemented by adding these kinds of imports to `available_submodule_attributes` and lowering the priority of `available_submodule_attributes`.

Keeping the priority low ensures that things like `from .y import y` in an `__init__.py` reliable exposes the function/class `y` and not the submodule `y` (an extremely common idiom).

Fixes https://github.com/astral-sh/ty/issues/1654

## Test Plan

Snapshot tests.

### DIFF
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -1536,25 +1536,22 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                     }
 
                     if node.module.is_some() {
-                        if let Ok(thispackage) = this_package
+                        if is_package
+                            && let Ok(thispackage) = this_package
+                            && self.current_scope().is_global()
                             && let Some(relative_submodule) = module_name.relative_to(&thispackage)
+                            && let Some(direct_submodule) = relative_submodule.components().next()
+                            && !self.seen_submodule_imports.contains(direct_submodule)
                         {
-                            if is_package
-                                && self.current_scope().is_global()
-                                && let Some(direct_submodule) =
-                                    relative_submodule.components().next()
-                                && !self.seen_submodule_imports.contains(direct_submodule)
-                            {
-                                self.seen_submodule_imports
-                                    .insert(direct_submodule.to_owned());
+                            self.seen_submodule_imports
+                                .insert(direct_submodule.to_owned());
 
-                                let direct_submodule_name = Name::new(direct_submodule);
-                                let symbol = self.add_symbol(direct_submodule_name);
-                                self.add_definition(
-                                    symbol.into(),
-                                    ImportFromSubmoduleDefinitionNodeRef { node },
-                                );
-                            }
+                            let direct_submodule_name = Name::new(direct_submodule);
+                            let symbol = self.add_symbol(direct_submodule_name);
+                            self.add_definition(
+                                symbol.into(),
+                                ImportFromSubmoduleDefinitionNodeRef { node },
+                            );
                         } else {
                             for name in module_name.ancestors() {
                                 self.imported_modules
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -1535,11 +1535,12 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                         is_self_import = &module_name == thispackage;
                     }
 
-                    if self.current_scope().is_global() && node.module.is_some() {
+                    if node.module.is_some() {
                         if let Ok(thispackage) = this_package
                             && let Some(relative_submodule) = module_name.relative_to(&thispackage)
                         {
                             if is_package
+                                && self.current_scope().is_global()
                                 && let Some(direct_submodule) =
                                     relative_submodule.components().next()
                                 && !self.seen_submodule_imports.contains(direct_submodule)
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -1560,6 +1560,16 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                                     .entry(name)
                                     .or_insert(ImportKind::ImportFrom);
                             }
+                            for name in &node.names {
+                                let Some(relative_name) = ModuleName::new(&name.name) else {
+                                    continue;
+                                };
+                                let mut full_name = module_name.clone();
+                                full_name.extend(&relative_name);
+                                self.imported_modules
+                                    .entry(full_name)
+                                    .or_insert(ImportKind::ImportFrom);
+                            }
                         }
                     }
                 }
--- a/crates/ty_python_semantic/src/semantic_index/builder.rs
+++ b/crates/ty_python_semantic/src/semantic_index/builder.rs
@@ -47,7 +47,7 @@ use crate::semantic_index::symbol::{ScopedSymbolId, Symbol};
 use crate::semantic_index::use_def::{
     EnclosingSnapshotKey, FlowSnapshot, ScopedEnclosingSnapshotId, UseDefMapBuilder,
 };
-use crate::semantic_index::{ExpressionsScopeMap, SemanticIndex, VisibleAncestorsIter};
+use crate::semantic_index::{ExpressionsScopeMap, ImportKind, SemanticIndex, VisibleAncestorsIter};
 use crate::semantic_model::HasTrackedScope;
 use crate::unpack::{EvaluationMode, Unpack, UnpackKind, UnpackPosition, UnpackValue};
 use crate::{Db, Program};
@@ -110,7 +110,7 @@ pub(super) struct SemanticIndexBuilder<'db, 'ast> {
     scopes_by_expression: ExpressionsScopeMapBuilder,
     definitions_by_node: FxHashMap<DefinitionNodeKey, Definitions<'db>>,
     expressions_by_node: FxHashMap<ExpressionNodeKey, Expression<'db>>,
-    imported_modules: FxHashSet<ModuleName>,
+    imported_modules: FxHashMap<ModuleName, ImportKind>,
     seen_submodule_imports: FxHashSet<String>,
     /// Hashset of all [`FileScopeId`]s that correspond to [generator functions].
     ///
@@ -150,7 +150,7 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
             expressions_by_node: FxHashMap::default(),
 
             seen_submodule_imports: FxHashSet::default(),
-            imported_modules: FxHashSet::default(),
+            imported_modules: FxHashMap::default(),
             generator_functions: FxHashSet::default(),
 
             enclosing_snapshots: FxHashMap::default(),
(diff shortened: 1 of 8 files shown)


AlexWaygood:
everything's great as long as you don't scroll down to the bit of the primer diff where the `prefect` diagnostics are listed

Gankra:
Taking a look at the code, the errors on prefect seem completely legitimate? The file doesn't import it and the `__init__.py`'s along the chain don't import it. These symbols being available must be genuine spooky-action-at-a-distance that we should probably discourage.

AlexWaygood:
> Taking a look at the code, the errors on prefect seem completely legitimate? The file doesn't import it and the `__init__.py`'s along the chain don't import it. These symbols being available must be genuine spooky-action-at-a-distance that we should probably discourage.

Yeah I agree, I'm not sure why we aren't emitting errors on this on `main`? There are other repos on which this branch does indeed produce false positives though, which I'm looking at now

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1513,33 +1517,46 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
...
-                        self.seen_submodule_imports
-                            .insert(direct_submodule.to_owned());
-
-                        let direct_submodule_name = Name::new(direct_submodule);
-                        let symbol = self.add_symbol(direct_submodule_name);
-                        self.add_definition(
-                            symbol.into(),
-                            ImportFromSubmoduleDefinitionNodeRef { node },
-                        );
+                    if self.current_scope().is_global() && node.module.is_some() {
```
Gankra: Should we add the `is_global` check to `ast::Stmt::Import` too? (It not being there bothers me...)

Gankra:
Also did you ever try not-distinguishing-import-kind? i.e. even making `import ...` weak?

The synthesis of these two issues would be interested in it (but maybe we just use the kind machinery to represent that case anyway):

* https://github.com/astral-sh/ty/issues/1490
* https://github.com/astral-sh/ty/issues/1526

AlexWaygood:
> Also did you ever try not-distinguishing-import-kind? i.e. even making `import ...` weak?

I _did_, but it directly contradicts some [longstanding tests](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/import/tracking.md#attribute-overrides-submodule) that were added in https://github.com/astral-sh/ruff/pull/14946 sooo... I got scared :P

I _do_ think it makes sense that if a user has `import a.b` at the top of their module, they almost certainly expect the attribute `a.b` to resolve to the `b` submodule of the package `a`, whereas they won't necessarily expect the same thing for the `a.b` access if all they have in their module is `from a.b import c`... but it's also the case that in tie-breaking situations, we just don't know for sure which way the tie will _actually_ be broken, and I don't know if it would be clear to users why we're distinguishing between the two kinds of imports.

The longstanding tests were also added before we added any kind of ecosystem check, so I do think it would be interesting to see the ecosystem effect from not-distinguishing-import-kind as well. I'd like to try it separately, after I've debugged some of the stranger diagnostics on this branch a bit more.

crates/ty_python_semantic/src/semantic_index/builder.rs
AlexWaygood: yeah you're right, these should definitely be consistent

crates/ty_python_semantic/src/types.rs
```diff
@@ -11765,33 +11771,55 @@ impl<'db> ModuleLiteralType<'db> {
...
+                imported_symbol(db, file, name, None).map_type(|ty| {
+                    if let Some(importing) = self.importing_file(db)
+                        && let Type::ModuleLiteral(module) = ty
+                    {
+                        Type::module_literal(db, importing, module.module(db))
+                    } else {
+                        ty
+                    }
+                })
+            })
```
AlexWaygood, line 13330: it's extremely hard to write a regression test for this (though I might try harder later), but empirically this gets rid of a false-positive diagnostic on [this line](https://github.com/DataDog/dd-trace-py/blob/7e78fe5f172442f2379ea725fe4e5ea534f00395/ddtrace/contrib/internal/trace_utils.py#L56) that was ultimately due to us thinking the `importing_file` of `ddtrace.internal` was `ddtrace/__init__.py` rather than `ddtrace/contrib/internal/trace_utils.py`

AlexWaygood:
I can't see any "obviously wrong" diagnostics in the ecosystem anymore among the added or changed diagnostics. Nearly all the _new_ diagnostics are on prefect. As noted above, I'm not sure why these aren't emitted on `main`; they seem like true positives, as do all new diagnostics on other repos. This does get rid of diagnostics across the ecosystem overall now; it's just a question of whether this behaviour is explainable to users.

Still TODO is
- ~~to just try out the bigger hammer of always giving `available_submodule_attributes` lowest priority, which would certainly be a simpler rule (and therefore easier to explain to users)~~.
  - possibly a wash? https://github.com/astral-sh/ruff/pull/21653
- ~~https://github.com/astral-sh/ruff/pull/21583#discussion_r2554196826 should probably be fixed in a standalone PR~~
  - I tried this in #21599 and didn't like it
- ~~try harder to write a standalone regression test for https://github.com/astral-sh/ruff/pull/21583#discussion_r2554271218~~
  - THE DEVIL ITSELF LIVES HERE??
- ~~figure out why none of those prefect diagnostics are reported on `main`~~
  - no longer hitting a `__getattr__` fallback

Gankra:
They say that even to this day if you put your ear against a `from ... import` statement you can hear the screams of engineers in tie-break hell.

Gankra:
Taking over this PR for Alex.

Gankra:
Oh dear, rebasing + rerunning has even worse results than last time. Gotta look at what's goin' on...

Gankra:
`prefect` mystery solved: their `__init__.py` intentionally doesn't import anything and contains a `__getattr__` that lazy-imports its submodules. We previously found this implementation out of desperation and concluded all submodule attributes resolved and were `Any`. Now that we resolve a lot of imports, we "dodge" the `__getattr__` and have way less `Any` making everything resolve everywhere.

This is an acceptable regression.

Gankra:
Ok well I can see why you had trouble reproducing the issue... THE ISSUE ONLY APPEARS IF THE CODE CONTAINS `from typing import TYPE_CHECKING`

YOU DON'T EVEN NEED TO USE IT.

WHAT.

Gankra:
More precise statement after messing around with it: `from typing import <anything that resolves>` has this issue. Importing from other modules like `typing_extensions` or `warnings` has no effect.

Gankra:
With asymmetric weak priority:

Screenshot 2025-11-26 at 8 07 45 PM


With fully weak priority:


Screenshot 2025-11-26 at 8 18 25 PM

Gankra:
Note: split the uniformly-weak-submodule-attributes out to https://github.com/astral-sh/ruff/pull/21653

Gankra:
scipy has the same magic lazy imports with `__getattr__` thing that prefect does, although it comes up a lot less.

*  [ Module `scipy.special` has no member `multigammaln` ](https://github.com/scipy/scipy/blob/1c36a69785d3d27ae279f3efbc4d66d7dd8acc7b/scipy/stats/_multivariate.py#L1821) even though [it's exported in `__all__` but maybe we get confused about the complex definition of it?](https://github.com/scipy/scipy/blob/1c36a69785d3d27ae279f3efbc4d66d7dd8acc7b/scipy/special/__init__.py#L829)
*  [Module `scipy.linalg.blas` has no member `dnrm2`](https://github.com/scipy/scipy/blob/1c36a69785d3d27ae279f3efbc4d66d7dd8acc7b/scipy/linalg/tests/test_blas.py#L1025)

Gankra:
```diff
- [warning] possibly-missing-attribute - Submodule `modin` may not be available as an attribute on module `pandera.typing`
+ [error] unresolved-attribute - Object of type `<module 'pandera.typing'> | <module 'pandera.typing'>` has no attribute `modin`
```

This is an unfortunate but kinda minor degradation (we could potentially add some diagnostic cleaning for it).

Gankra:
I think this is ready for review/discussion.

crates/ty_python_semantic/src/types.rs
AlexWaygood: update: @Gankra succeeded in writing a regression test 😃 it's the "A tale of two modules" regression test

Gankra (approved):
I guess I actually didn't end up changing the code itself, so, the code changes honestly look like what I'd expect.

AlexWaygood:
Thank you so much for picking this up, and in particular for experimenting with #21653 separately!

I'm surprised at how small the diff is on #21653 -- it's basically _all_ pwndbg, which is doing very strange and unusual things with its submodule imports (["submodule imports Georg" should be excluded as an outlier](https://www.huffingtonpost.co.uk/entry/spiders-georg-meme_n_5412861), etc.). So I wonder if we should just adopt the rule proposed by #21653, where submodule attributes always have lower priority than symbols defined in the `__init__.py` module namespace? The attraction of that is that it's a much simpler rule to explain to users; it's not great to have to explain a subtle distinction where we treat a submodule attribute introduced by `import a.b` as having a higher priority than a submodule attribute introduced by `from a.b import c`.

As per https://github.com/astral-sh/ruff/pull/14946#discussion_r1884626442, that would mean that we'd be moving towards mypy's behaviour here and away from pyright's. But empirically, the ecosystem results appear to indicate that mypy's heuristic works better much more of the time than pyright's does.

I'm curious for what others think; don't make any changes here until other folks have had a chance to chime in 😅 The disadvantage of adopting the rule in #21653 is that you can get counterintuitive situations where things like this can happen:

```py
import a.b

reveal_type(a.b)  # revealed: Literal[1]
```

But the ecosystem report suggests that situations like this are rare. And honestly, at some point you just have to say that libraries need to stop shadowing their submodules with variables in `__init__.py` if they want type checkers to understand this stuff.

AlexWaygood:
Moving back into draft because @Gankra and I want to try a few more things out (possibly after the beta)

PR #21585, opened 2025-11-22
[ruff] single-assignment-missing-final (RUF066)

### LINKED ISSUE #21584: Flag variables that 1. aren't reassigned, and 2. are missing `Final`
### Summary

I think it'd be helpful if `ruff` flagged variables that

1. aren't reassigned, and
2. are missing `Final`

hamirmahal:
Closes #21584

## Summary

This change adds a new rule that warns when variables that are only assigned once are missing a `Final` annotation.

## Test Plan

I added several test files like

```
crates/ruff_linter/resources/test/fixtures/immutability/annotated_final.py
```


hirehamir:
This initial implementation is conservative and attempts to only flag cases where a variable is definitely missing a `Final` annotation.

MichaReiser:
Thanks for your contribution. I'll close this PR for now until we've come to a decision whether we want to add this rule to Ruff. 

The high number of violations makes me very warry and highly suggests that the rule needs to be further refined.

hirehamir:
You're welcome. Would this be more appropriate for [`ty`](https://github.com/astral-sh/ty)?

ntBre:
I believe we're focused on core type checking features in ty rather than type-inference-powered lint rules at the moment. I think the issue you opened requesting the rule is probably a good place to track support for such a rule before implementing it in Ruff or ty.

Personally, I love having `let` vs `let mut` in Rust, but I agree with Micha's comment on the issue that this feels very pedantic in Python, even with type inference making the detection more reliable.

PR #21587, opened 2025-11-23
[ty] Refactor `CheckSuppressionContext` to use `DiagnosticGuard`

MichaReiser:
## Summary

This is PR is a first step towards adding a "Remove unused suppression" code action in the editor. 

Adding the fix for removing unused suppression requires setting the `Fix` on the 
diagnostic created by `CheckSuppressionContext::report_unchecked`. To make this possible,
change `report_unchecked` and `report_lint` to return a `DiagnosticGuard`.

Doing so requires extracting `DiagnosticGuard` from `InferContext`. 

I tried very hard to not having to change `CheckSuppressionContext` to store a `RefCell<TypeCheckDiagnostic>` but
I couldn't find a way to do so. I tried a `enum DiagnosticSink { Cell(RefCell<TypeCheckDiagnostics>), Mut(&mut TypeCheckDiagnostic) }` 
with a `emit` method that takes `&self`. However, the borrow checker (correctly?), disallowes me to mutate the `&mut TypeCheckDiagnostic` in that case. 

I gave up after 30min or so and I sort of like how `check_suppressions` takes the `Diagnostic`s with suppressions and returns only
the diagnostics once it's done, as the suppressions have now been handled.

## Test Plan

`cargo test`

### DIFF
--- a/crates/ty_python_semantic/src/diagnostic.rs
+++ b/crates/ty_python_semantic/src/diagnostic.rs
@@ -172,11 +172,14 @@ pub(super) struct DiagnosticGuard<'sink> {
     /// We use a [`RefCell`] here over a `&mut TypeCheckDiagnostics` to ensure the fact that
     /// `InferContext` (and other contexts with diagnostics) use a [`RefCell`] internally
     /// remains abstracted away. Specifically, we want to ensure that calling `report_lint` on
-    /// [`InferContext`] twice doesn't result in a panic:
+    /// `InferContext` twice doesn't result in a panic:
     ///
     /// ```ignore
     /// let diag1 = context.report_lint(...);
-    /// let diag2 = context.report_lint(...); // should not panic because of second mutable borrow
+    ///
+    /// // would panic if using a `&mut TypeCheckDiagnostics`
+    /// // because of a second mutable borrow.
+    /// let diag2 = context.report_lint(...);
     /// ```
     sink: &'sink RefCell<TypeCheckDiagnostics>,
 


crates/ty_python_semantic/src/diagnostic.rs
```diff
@@ -144,3 +152,153 @@ where
 
     buffer
 }
+
+/// An abstraction for mutating a diagnostic.
```
MichaReiser, line 154: This is just a move except that `DiagnosticGuard` now stores a `file` and `sink` instead of a `&InferContext`

crates/ty_python_semantic/src/types/context.rs
```diff
@@ -625,19 +540,9 @@ impl<'db, 'ctx> DiagnosticGuardBuilder<'db, 'ctx> {
...
     /// impl to `Diagnostic`.
-    pub(super) fn into_diagnostic(
-        self,
-        message: impl std::fmt::Display,
-    ) -> DiagnosticGuard<'db, 'ctx> {
-        let mut diag = Diagnostic::new(self.id, self.severity, message);
-
-        if let DiagnosticId::Lint(lint_name) = diag.id() {
-            diag.set_documentation_url(Some(lint_documentation_url(lint_name)));
-        }
```
MichaReiser, line 636: I moved this logic into `DiagnosticGuard::drop`, to reduce the code duplication

crates/ty_python_semantic/src/diagnostic.rs
```diff
@@ -144,3 +149,114 @@ where
...
+///
+/// Unlike `LintDiagnosticGuard`, this API does not guarantee that the
+/// constructed `Diagnostic` not only has a primary annotation, but its
+/// associated file is equivalent to the file being type checked. As a result,
+/// if either is violated, then the `Drop` impl on `DiagnosticGuard` will
+/// panic.
+pub(super) struct DiagnosticGuard<'sink> {
+    file: File,
+
+    sink: &'sink std::cell::RefCell<TypeCheckDiagnostics>,
```
BurntSushi: What goes wrong when you use `&mut TypeCheckDiagnostics` here?

It might be worth a brief comment saying why interior mutability is being used here.

MichaReiser: It would panic if you do:

```
let diag = context.report_lint(...);
let diag2 = context.report_lint(...); // we now borrow `&mut TypeCheckDiagnostics` twice
```

This seems *worse* as an API caller shouldn't be aware of the internal ref-cell borrow

BurntSushi: I guess I meant the higher level reason. But I think I now see: you wouldn't be able to have two guards alive at the same time?

MichaReiser: Yes, the higher level reason is that the current approach correctly isolates the `RefCell` so that callers don't need to know about it (I'll add a comment).

PR #21597, opened 2025-11-23
[ty] Improve several "Did you mean?" suggestions

AlexWaygood:
## Summary

- Make some diagnostics more concise by moving suggestions from a subdiagnostic to a primary or secondary annotation message
- Make some diagnostic summary lines more concise by moving suggestions from the diagnostic summary line to subdiagnostics
- Consistently use the `help` severity for suggestions when they do appear as subdiagnostics, rather than the `info` severity

## Test Plan

Snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8573,25 +8573,23 @@ impl<'db> InvalidTypeExpression<'db> {
                     InvalidTypeExpression::Field => {
                         f.write_str("`dataclasses.Field` is not allowed in type expressions")
                     }
-                    InvalidTypeExpression::ConstraintSet => {
-                        f.write_str("`ty_extensions.ConstraintSet` is not allowed in type expressions")
-                    }
-                    InvalidTypeExpression::GenericContext => {
-                        f.write_str("`ty_extensions.GenericContext` is not allowed in type expressions")
-                    }
-                    InvalidTypeExpression::Specialization => {
-                        f.write_str("`ty_extensions.GenericContext` is not allowed in type expressions")
-                    }
-                    InvalidTypeExpression::TypedDict => {
-                        f.write_str(
-                            "The special form `typing.TypedDict` is not allowed in type expressions. \
-                            Did you mean to use a concrete TypedDict or `collections.abc.Mapping[str, object]` instead?")
-                    }
-                    InvalidTypeExpression::TypeAlias => {
-                        f.write_str(
-                            "`typing.TypeAlias` is only allowed as the sole annotation on an annotated assignment",
-                        )
-                    }
+                    InvalidTypeExpression::ConstraintSet => f.write_str(
+                        "`ty_extensions.ConstraintSet` is not allowed in type expressions",
+                    ),
+                    InvalidTypeExpression::GenericContext => f.write_str(
+                        "`ty_extensions.GenericContext` is not allowed in type expressions",
+                    ),
+                    InvalidTypeExpression::Specialization => f.write_str(
+                        "`ty_extensions.GenericContext` is not allowed in type expressions",
+                    ),
+                    InvalidTypeExpression::TypedDict => f.write_str(
+                        "The special form `typing.TypedDict` \
+                            is not allowed in type expressions",
+                    ),
+                    InvalidTypeExpression::TypeAlias => f.write_str(
+                        "`typing.TypeAlias` is only allowed \
+                            as the sole annotation on an annotated assignment",
+                    ),
                     InvalidTypeExpression::TypeQualifier(qualifier) => write!(
                         f,
                         "Type qualifier `{qualifier}` is not allowed in type expressions \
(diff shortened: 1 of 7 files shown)


crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -148,15 +148,13 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
...
-                        if let Some(mut diag) = self.report_invalid_type_expression(
+                        self.report_invalid_type_expression(
                             expression,
                             format_args!(
                                 "Invalid binary operator `{}` in type annotation",
                                 op.as_str()
                             ),
-                        ) {
-                            diag.info("Did you mean to use `|`?");
-                        }
```
AlexWaygood, line 159: we don't have any tests for this subdiagnostic and I'm not convinced it's likely to be helpful. If a user has `int + str` in a type expression, I think it's very unclear what they meant; if they have `int - bool`, I think it's likely that they were trying to write a negation type rather than a union type; etc.

AlexWaygood:
Screenshot of some of the new diagnostics, since snapshots don't have colour: 
image

MichaReiser:
> Make some diagnostics more concise by moving suggestions from a subdiagnostic to a primary or secondary annotation message

I think I've a preference towards not doing that because it makes the `concise` output less concise and I also find it a bit confusing that the description of some underlined item is a question.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8615,35 +8618,39 @@ impl<'db> InvalidTypeExpression<'db> {
...
+            {
+                return;
+            }
 
-        // TODO: showing a diff (and even having an autofix) would be even better
-        diagnostic.info(format_args!(
-            "Did you mean to use the module's member \
-            `{module_name_final_part}.{module_name_final_part}` instead?"
-        ));
+            // TODO: showing a diff (and even having an autofix) would be even better
```
MichaReiser, line 8643: This should be "easy" once my unused suppression PR lands.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
MichaReiser: We could say something about that only `|` is allowed to define a union.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8615,35 +8618,39 @@ impl<'db> InvalidTypeExpression<'db> {
...
+            diagnostic.set_primary_message(format_args!(
+                "Did you mean to use the module's member \
+                `{module_name_final_part}.{module_name_final_part}`?"
+            ));
+        } else if let InvalidTypeExpression::TypedDict = self {
+            diagnostic.help(
+                "You might have meant to use a concrete TypedDict \
```
MichaReiser, line 8650: I think I know what it is but it might not be clear to all users what a concrete `TypedDict` is.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
AlexWaygood: We could, but we already link to the full type expression grammar in a separate subdiagnostic (which covers that), and I think it's very unclear that they were actually trying too define a union at all

crates/ty_python_semantic/src/types.rs
AlexWaygood: A type that actually is a TypedDict, rather than the special form typing.TypedDict itself.

```py
from typing import TypedDict

class MyTypedDict(TypedDict):
    x: int

obj: MyTypedDict
```

instead of 

```py
from typing import TypedDict

obj: TypedDict
```

I take your point that this might be unclear to some people, but I'm not sure what clearer wording here might be... "An actual TypedDict" feels just as bad. And I can't put that whole example in the subdiagnostic!

MichaReiser: Okay, I actually assumed it's something different 😅 

But I agree it's tricky. I tried to produce a similar error in TypeScript but they don't have this issue because it doesn't have marker types similar to `TypedDict`

sharkdp:
> > Make some diagnostics more concise by moving suggestions from a subdiagnostic to a primary or secondary annotation message
> 
> I think I've a preference towards not doing that because it makes the `concise` output less concise and I also find it a bit confusing that the description of some underlined item is a question.

If we are confident in our suggestion, I like the change here. I always look at the primary annotation before I even read the main diagnostic message, and if it's directly "actionable", that's perfect.

AlexWaygood:
Yeah, I agree with @sharkdp. If we have a direct suggestion for how to fix the diagnostic, that feels like one of the most important and useful parts of the diagnostic display to me. I think it's really useful to have that highlighted prominently, and I like that with secondary annotations we can clearly indicate that we might only be suggesting to change _part_ of the diagnostic range, not necessarily the whole range.

> it makes the `concise` output less concise

That's true of course, but to me it seems like very useful information that I'd want as part of the concise diagnostic. And I don't think the new concise diagnostic messages we see in mypy_primer are _that_ verbose?

MichaReiser:
I wonder if the solution here really is to just attach a `Fix`

AlexWaygood:
I'll land this for now as an incremental improvement. It's easy to make further adjustments later and we can easily re-evaluate if and when we add fixes for some or all of these diagnostics

PR #21598, opened 2025-11-23
[ty] Extend Liskov checks to also cover classmethods and staticmethods

AlexWaygood:
## Summary

Building on https://github.com/astral-sh/ruff/pull/21436.

There's nothing conceptually more complicated about this, it just requires its own set of tests and its own subdiagnostic hint.

I also uncovered another inconsistency between mypy/pyright/pyrefly, which is fun. In this case, I suggest we go with pyright's behaviour.

## Test Plan

mdtests/snapshots

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/liskov.md_-_The_Liskov_Substitut…_-_Staticmethods_and_cl…_(49e28aae6fdd1291).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/liskov.md_-_The_Liskov_Substitut…_-_Staticmethods_and_cl…_(49e28aae6fdd1291).snap
@@ -85,7 +85,7 @@ error[invalid-method-override]: Invalid override of method `instance_method`
  3 |     @classmethod
  4 |     def class_method(cls, x: int) -> int: ...
    |
-info: `BadChild1.instance_method` is a staticmethod but `Parent.instance_method` is a instance method
+info: `BadChild1.instance_method` is a staticmethod but `Parent.instance_method` is an instance method
 info: This violates the Liskov Substitution Principle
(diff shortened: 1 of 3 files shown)


AlexWaygood:
I skimmed the new diagnostics and couldn't see any obvious false positives. A lot of them are, once again, us being more pedantic than other type checkers when it comes to subclasses changing the names of parameters -- but as before, it's very hard for us to put those in a different error code until we're able to propagate through `Type::has_relation_to` the reason _why_ an assignability/subtyping check failed. The `unused-ignore-comment` diagnostics going away seem all to be `type: ignore[override]` comments (i.e., folks were suppressing mypy's version of this check, and those suppression comments are now no longer viewed as unused by ty -- which is good!)

crates/ty_python_semantic/resources/mdtest/snapshots/liskov.md_-_The_Liskov_Substitut…_-_Staticmethods_and_cl…_(49e28aae6fdd1291).snap
```diff
@@ -0,0 +1,220 @@
...
+   |
+  ::: src/mdtest_snippet.pyi:2:9
+   |
+ 1 | class Parent:
+ 2 |     def instance_method(self, x: int) -> int: ...
+   |         ------------------------------------ `Parent.instance_method` defined here
+ 3 |     @classmethod
+ 4 |     def class_method(cls, x: int) -> int: ...
+   |
+info: `BadChild1.instance_method` is a staticmethod but `Parent.instance_method` is a instance method
```
AlexWaygood: Ugh this needs to be "an instance method" rather than "a instance method" — will fix tomorrow

PR #21599, opened 2025-11-23
[ty] Only add `import`ed modules to `imported_modules` if the `import` statement is in the global scope

AlexWaygood:
## Summary

@Gankra pointed out in https://github.com/astral-sh/ruff/pull/21583#discussion_r2554196826 that we're currently inconsistent here between how we treat `import` statements and `from ... import ...` statements. We should probably be consistent!

## Test Plan

Added an mdtest

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/tracking.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/tracking.md
@@ -1,6 +1,6 @@
 # Tracking imported modules
 
-These tests depend on how we track which modules have been imported. There are currently two
+These tests depend on how we track which modules have been imported. There are several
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Ah, hmm, I suppose there actually is no inconsistency right now, because currently `from ... import ...` imports only ever create `Definition`s (never available submodule attributes). But https://github.com/astral-sh/ruff/pull/21583, as it currently stands, would _create_ an inconsistency (because it starts to add available submodule attributes based on `from ... import ...` imports, sometimes).

The number of ecosystem hits here makes me think I should just adjust #21583 rather than changing anything about how we do non-`from` imports.

PR #21600, opened 2025-11-24
[ty] Double click to insert inlay hint

MatthewMckee4:
## Summary

Resolves https://github.com/astral-sh/ty/issues/317#issuecomment-3567398107.

I can't get the auto import working great.

I haven't added many places where we specify that the type display is invalid syntax.

## Test Plan

Nothing yet

### DIFF
--- a/crates/ty_ide/src/inlay_hints.rs
+++ b/crates/ty_ide/src/inlay_hints.rs
@@ -318,7 +318,7 @@ impl SourceOrderVisitor<'_> for InlayHintVisitor<'_, '_> {
         match stmt {
             Stmt::Assign(assign) => {
                 self.in_assignment = !type_hint_is_excessive_for_expr(&assign.value);
-                if dont_allow_edits(assign) {
+                if !annotations_are_valid_syntax(assign) {
                     self.in_no_edits_allowed = true;
                 }
                 for target in &assign.targets {
@@ -445,9 +445,9 @@ fn type_hint_is_excessive_for_expr(expr: &Expr) -> bool {
     }
 }
 
-fn dont_allow_edits(stmt_assign: &ruff_python_ast::StmtAssign) -> bool {
+fn annotations_are_valid_syntax(stmt_assign: &ruff_python_ast::StmtAssign) -> bool {
     if stmt_assign.targets.len() > 1 {
-        return true;
+        return false;
     }
 
     if stmt_assign
@@ -455,10 +455,10 @@ fn dont_allow_edits(stmt_assign: &ruff_python_ast::StmtAssign) -> bool {
         .iter()
         .any(|target| matches!(target, Expr::Tuple(_)))
     {
-        return true;
+        return false;
     }
 
-    false
+    true
 }
 
 #[cfg(test)]
--- a/crates/ty_python_semantic/src/types/definition.rs
+++ b/crates/ty_python_semantic/src/types/definition.rs
@@ -51,17 +51,3 @@ impl TypeDefinition<'_> {
         }
     }
 }
-
-impl<'db> TypeDefinition<'db> {
-    pub fn definition<'a>(&'a self) -> Option<&'a Definition<'db>> {
-        match self {
-            Self::Module(_) => None,
-            Self::Class(definition)
-            | Self::Function(definition)
-            | Self::TypeVar(definition)
-            | Self::TypeAlias(definition)
-            | Self::SpecialForm(definition)
-            | Self::NewType(definition) => Some(definition),
-        }
-    }
-}


crates/ty_ide/src/inlay_hints.rs
```diff
@@ -62,10 +73,73 @@ impl InlayHint {
...
+
+                        let module_name = module.name(db).as_str();
+
+                        if module_name == "builtins" {
+                            continue;
+                        }
+
+                        let Some(definition_name) = definition.name(db) else {
+                            continue;
+                        };
```
MatthewMckee4: This causes issues for some definition kinds, like `Unknown`.

Gankra:
The parts that aren't the auto-import stuff look reasonable to me (just not as familiar with auto-import stuff).

Might make sense to have the auto-import stuff be a follow-up?

MatthewMckee4:
Yes, I agree. Perhaps @BurntSushi would be able to advise or help with the auto import?

I can remove the auto import for now. 

Perhaps we can put this behind a experimental flag for now also?

I will also try to add more invalid syntax places.

I also think it would be useful if we had alternative display for types that emit invalid synatx, like I'd still like to add a function signature with `Callable` notation.
This also seems like something we can just iterate on.

Gankra:
> Perhaps we can put this behind a experimental flag for now also?

I'm on the fence about it because you can just... not double-click (and undo if you do). I think we can land the minimal implementation and then disable/flag it if it's a problem for some reason.

>  This also seems like something we can just iterate on.

Yeah agreed.

Gankra:
I don't think it's strictly a blocker to not catch all the invalid syntax. It's something we'll probably burn down really fast in followups and the only stakes is a user getting mad that a feature they manually invoked did something useless.

Gankra:
Actually ok I think it's slightly higher stakes -- "user gets very confused" is probably more likely. But still, followups.

MatthewMckee4:
> I'm on the fence about it because you can just... not double-click

true, but at least for now if you double click on some type with a `todo` type, its looks horribly and not an amazing experience.

Though, we may not get the user feedback about missing invalid syntax types if we hid it.

Happy to keep it open now, and let the issues trickle in.

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -62,10 +73,73 @@ impl InlayHint {
...
+                        let import_action = importer.import(request, &members);
+
+                        if let Some(edit) = import_action.import() {
+                            if let Some(content) = &edit.content() {
+                                text_edits.push(InlayHintTextEdit {
+                                    range: edit.range(),
+                                    new_text: (*content).to_string(),
+                                });
+                            }
+                        }
```
BurntSushi: What you have here looks reasonable to me. What's going wrong?

MatthewMckee4: I will review more tonight. But off the top of my head.

```py
T[: typing.TypeVar] = TypeVar("T")
```
the definition name is `TypeVar` so we `from typing import TypeVar` for example, but this doesnt help us. as we now have the annotation `: typing.TypeVar`.

Obviously not a great example because we should not type hint `TypeVar` (also a good place to not allow insert option`.

But for other types this would cause problems.

So its mostly an issue with the names.

Gankra:
A rough heuristic for invalid syntax that will let you find most things: i believe `(`, `)`, `<`, `>`, `'` are all invalid in type syntax (idk if `'` is stricly invalid but anywhere we explicitly emit it we're doing type crimes).

MatthewMckee4:
Nice okay, do you think its worth it for us to scan over the label after getting the TypeDetails to check for invalid characters? Or should i just search for them in the display.rs file and manually set them to all be invalid

Gankra:
It's really tempting to just do the `contains` but I think it's probably best to mark up our code in display.rs yeah.

crates/ty_ide/src/inlay_hints.rs
BurntSushi: Oh I see. You might need to use `ImportAction::symbol_text`, since the import might need to be fully qualified for correctness in some cases.

With that said, I think the specific issue you're hitting is that you're always requesting a `from import` statement via `ImportRequest::import_from`. If you want to respect an existing qualified import, then you'll want `ImportRequest::import`. So you'll probably need to detect which one you have and then use the right request. (Note though that it is only a _request_. So you still need to use `ImportAction::symbol_text`.)

I built the importer API with completions in mind. It might make sense to change `ImportAction::symbol_text` to return an `Edit` instead of a `&str` for your use case.

MatthewMckee4: Okay, thanks. I'll leave this out for this PR i think, and have a go at it separately.

One thing i am thinking is that I don't know how many different formats of types we have like this. If it were just `TypeVar` and `typing.TypeVar` I could easily distinguish between them, but I don't know other ways this could show.

MatthewMckee4:
Is it ridiculous to try to make some property tests for this?

testing that after the edits have been accepted, that we have valid python syntax? And possibly no diagnostics on the file?

Gankra:
More tests is more better, I would have been lazy and just snapshot-tested like the rest of the IDE stuff 😅

MatthewMckee4:
I will try it then, see how it goes.

Gankra:
I will say we have a release tomorrow and I'd love to get some version of this PR in for that release (not a huge deal if not but I think this will really solidify inlay hints as Useful).

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -908,6 +972,20 @@ mod tests {
...
+        info[inlay-hint-edit]: File after edits
+        info: Source
+
+        def i(x: int, /) -> int:
+            return x
+        def s(x: str, /) -> str:
+            return x
+
+        x1, y1 = (1, 'abc')
+        x2: Literal[1], y2: Literal["abc"] = (x1, y1)
```
MatthewMckee4: We should not edit here.

crates/ty_ide/src/inlay_hints.rs
AlexWaygood: FWIW, `T: TypeVar = TypeVar("T")` is an invalid TypeVar declaration that [ty will complain about](https://play.ty.dev/a43cc2d3-739f-42a2-a290-d6b9a57b6e6a), so we maybe shouldn't be offering a code action to add that to the user's code 😅

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -420,13 +445,30 @@ fn type_hint_is_excessive_for_expr(expr: &Expr) -> bool {
     }
 }
 
+fn dont_allow_edits(stmt_assign: &ruff_python_ast::StmtAssign) -> bool {
+    if stmt_assign.targets.len() > 1 {
+        return true;
+    }
```
Gankra: Oh great call on this -- I might also call this `is_valid_syntax`, since that's ultimately the crux?

crates/ty_ide/src/inlay_hints.rs
```diff
@@ -5493,6 +5793,16 @@ mod tests {
...
+
+        ---------------------------------------------
+        info[inlay-hint-edit]: File after edits
+        info: Source
+
+        class F:
+            @property
+            def whatever(self): ...
+
+        ab: property = F.whatever
```
Gankra, line 5805: (can be a followup) i have a sneaking suspicion this is a fake type

crates/ty_python_semantic/src/types/definition.rs
```diff
@@ -51,3 +51,17 @@ impl TypeDefinition<'_> {
...
+            Self::Module(_) => None,
+            Self::Class(definition)
+            | Self::Function(definition)
+            | Self::TypeVar(definition)
+            | Self::TypeAlias(definition)
+            | Self::SpecialForm(definition)
+            | Self::NewType(definition) => Some(definition),
+        }
+    }
+}
```
Gankra: Is this code from the auto-import stuff?

Gankra (approved):
Rad! Just a few nits.

crates/ty_python_semantic/src/types/definition.rs
MatthewMckee4: yeah, will remove, thanks

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: yeah this didnt look great

crates/ty_ide/src/inlay_hints.rs
MatthewMckee4: I guess its "if we add an annotation is it valid syntax".

MatthewMckee4: But I'm not sure how to name that, "is_valid_syntax" doesn't seem exactly correct.

Gankra: annotations_are_valid_syntax?

Gankra:
Hell yes hell yeah let's kick this things tires

PR #21612, opened 2025-11-24
[ty] Exclude `@Todo` types from `assert_{type,never}` checks

sharkdp:
## Summary

Not sure if this a good idea, I can see arguments either way. This was prompted by a wave of type-assertion failures involving `@Todo` types in https://github.com/astral-sh/ruff/pull/21553.


sharkdp:
Opening for review to see what others think.

carljm:
I think on balance I would come down against doing this? I feel like `assert_type` and `assert_never` are important not to lie about.

carljm:
But I can see the argument in favor (silencing false positives in areas where we have known limitations), so open to discussion...

PR #21615, opened 2025-11-24
[ty] Add hint about resolved Python version when a user attempts to import a member added on a newer version

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1620. #20909 added hints if you do something like this and your Python version is set to 3.10 or lower:

```py
import typing

typing.LiteralString
```

And we also have hints if you try to do something like this and your Python version is set too low:

```py
from stdlib_module import new_submodule
```

But we don't currently have any subdiagnostic hint if you do something like _this_ and your Python version is set too low:

```py
from typing import LiteralString
```

This PR adds that hint!

## Test Plan

snapshots

### DIFF
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -3631,7 +3631,7 @@ pub(super) fn report_invalid_method_override<'db>(
 /// versions, we add a hint to the diagnostic that the user may have
 /// misconfigured their Python version.
 ///
-/// The funciton returns `true` if a hint was added, `false` otherwise.
+/// The function returns `true` if a hint was added, `false` otherwise.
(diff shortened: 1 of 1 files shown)


crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3630,30 +3630,32 @@ pub(super) fn report_invalid_method_override<'db>(
 /// *does* exist as a submodule in the standard library on *other* Python
 /// versions, we add a hint to the diagnostic that the user may have
 /// misconfigured their Python version.
+///
+/// The funciton returns `true` if a hint was added, `false` otherwise.
```
Gankra:
```suggestion
/// The function returns `true` if a hint was added, `false` otherwise.
```

PR #21616, opened 2025-11-24
Implement goto-definition and find-references for global/nonlocal statements

Gankra:
## Summary

The implementation here is to just record the idents of these statements in `scopes_by_expression` (which already supported idents but only ones that happened to appear in expressions), so that `definitions_for_name` Just Works.

goto-type (and therefore hover) notably does not work on these statements because the typechecker does not record info for them. I am tempted to just introduce `type_for_name` which runs `definitions_for_name` to find other expressions and queries the inferred type... but that's a bit whack because it won't be the computed type at the right point in the code. It probably wouldn't be particularly expensive to just compute/record the type at those nodes, as if they were a load, because global/nonlocal is so scarce?

## Test Plan

Snapshot tests added/re-enabled.

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -331,15 +331,6 @@ impl GotoTarget<'_> {
                 let module = import_name(module_name, *component_index);
                 model.resolve_module_type(Some(module), *level)?
             }
-            // TODO: Support identifier targets
-            GotoTarget::PatternMatchRest(_)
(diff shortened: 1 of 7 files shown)


MichaReiser (commented):
Hmm, the changes in this PR seem to be primarily about string annotations and not nonlocal and global?

Gankra:
Yes it's based on top of my string attr PR which I have rebased without rebasing this one 😓

Gankra:
It is now rebased on main and ready to review.

PR #21619, opened 2025-11-24
[ty] Always infer the submodule literal for submodule attribute-access, even when we emit a diagnostic

AlexWaygood:
## Summary

Closes https://github.com/astral-sh/ty/issues/1623

## Test Plan

mdtests updated

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
@@ -61,7 +61,7 @@ import mypackage
 
 reveal_type(mypackage.imported.X)  # revealed: int
 # error: [possibly-missing-attribute] "Submodule `fails` may not be available"
-reveal_type(mypackage.fails.Y)  # revealed: Unknown
+reveal_type(mypackage.fails.Y)  # revealed: int
 ```
 
 ### In Non-Stub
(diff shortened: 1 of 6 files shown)


crates/ty_python_semantic/resources/mdtest/import/nonstandard_conventions.md
```diff
@@ -430,11 +432,13 @@ import mypackage
 
 # TODO: this would be nice to support
 # error: [possibly-missing-attribute] "Submodule `submodule` may not be available"
-reveal_type(mypackage.submodule)  # revealed: Unknown
+reveal_type(mypackage.submodule)  # revealed: <module 'mypackage.submodule'>
 # error: [possibly-missing-attribute] "Submodule `submodule` may not be available"
-reveal_type(mypackage.submodule.nested)  # revealed: Unknown
+# error: [possibly-missing-attribute] "Submodule `nested` may not be available"
+reveal_type(mypackage.submodule.nested)  # revealed: <module 'mypackage.submodule.nested'>
```
AlexWaygood, line 438: the fact that we now emit two diagnostics on this line (rather than one previously) makes me pretty nervous about this. Let's see what the ecosystem says, but something like `<module 'mypackage.submodule'> & Any` might be better for the inferred type of `mypackage.submodule`; then (in theory) we wouldn't emit the second diagnostic

Gankra:
Taking over this PR for Alex.

AlexWaygood:
Some thoughts: this branch means that you can get three separate `possibly-unresolved-attribute` errors from something like `x = package.sub.sub2.sub3`. If we intersected the submodule-literal type with `Any`, that might make things better? But also just landing https://github.com/astral-sh/ruff/pull/21583 might make the impact on this look better too, so probs best to do that first

Gankra:
Horribly aware of how much "what if x AND y works well" haunts everything so just spinning that up here in parallel: https://github.com/astral-sh/ruff/pull/21654

Issue #3070, opened 2023-02-20
new cli option: --disable-noqa


onerandomusername:
There are times when I want to go through the codebase and fix every error that uses a specific error code. It would be helpful to have a cli option to disable all noqa comments for a single command invoke.

For example, if I have a noqa comment for `E501`, I should be able to view all of my existing E501 errors by running `ruff . --disable-noqa --select E501`

charliermarsh:
We already support this internally (we use this behavior for `--add-noqa`), so in theory we could expose it with a flag yeah.

onerandomusername:
> We already support this internally (we use this behavior for `--add-noqa`), so in theory we could expose it with a flag yeah.

Thank you! I chose `--disable-noqa` because its what flake8 uses, but I wouldn't be opposed to `--ignore-noqa` or any other name.

charliermarsh:
Yeah I'd probably prefer `--ignore-noqa` personally.

sciyoshi:
Should there be a way to only disable/ignore noqa for certain error codes? For example, a codebase may want to disable using noqa for specific errors altogether. Maybe as a new config in `[tool.ruff]` for this purpose, e.g. `ignore-noqa = [ "E501" ]` would prevent `# noqa: E501` from working altogether.

not-my-profile:
> Should there be a way to only disable/ignore noqa for certain error codes?

I'd rather have us implement a `forbid` or `force-deny` lint level as per https://github.com/charliermarsh/ruff/issues/1256#issuecomment-1414165258.

> Yeah I'd probably prefer `--ignore-noqa` personally.

I think the CLI option should be called something more generic since we might want to also support `# pylint:` directives in the future.

charliermarsh:
Added in #3296. I left it as `--ignore-noqa` for now since it's consistent with (e.g.) `--add-noqa` but I'm open to suggestions.

PR #21623, opened 2025-11-25
apply range suppressions to filter diagnostics

amyreese:
Builds on range suppressions from https://github.com/astral-sh/ruff/pull/21441 

Filters diagnostics based on parsed valid range suppressions.

Issue: #3711


crates/ruff_linter/src/linter.rs
```diff
@@ -329,6 +326,14 @@ pub fn check_path(
...
+    // Apply range suppressions before noqa, so that #noqa on a line already covered by a range
+    // will be reported as unused noqa rather than unused range.
+    // TODO: check if enabled?
+    {
+        let suppressions = Suppressions::from_tokens(locator.contents(), tokens);
+        suppressions.filter_diagnostics(&mut context);
+    }
+
     // Enforce `noqa` directives.
     if noqa.is_enabled()
```
MichaReiser, line 331: I think we should only apply the filtering if `noqa_is_enabled`. I'm not a 100% sure but I think this is more a terminology issue. The CLI has a `--ignore-noqa` comment which you can use to show all diagnostics even if noqa's are disabled (https://github.com/astral-sh/ruff/issues/3070). I think we want this flag to also toggle `ruff:disable`/`ruff:enable`

I also suggest reviewing the other call sites where we set `noqa::Disabled` to see if that should apply to `ruff:disable` as well

crates/ruff_linter/src/linter.rs
```diff
@@ -329,6 +326,14 @@ pub fn check_path(
         }
     }
 
+    // Apply range suppressions before noqa, so that #noqa on a line already covered by a range
+    // will be reported as unused noqa rather than unused range.
+    // TODO: check if enabled?
+    {
+        let suppressions = Suppressions::from_tokens(locator.contents(), tokens);
+        suppressions.filter_diagnostics(&mut context);
+    }
```
MichaReiser: I think we should move the logic into `check_noqa` and rename the method to `check_suppressions`. This should also avoid having to iterate over all diagnostics twice, which can be expensive if there are many

MichaReiser (commented):
I think we should add some tests now. `check_path` is called from our testing framework, meaning we should now be able to write a test similar to

https://github.com/astral-sh/ruff/blob/850398dada852843e10f6a118127a6b801ed476d/crates/ruff_linter/src/rules/ruff/mod.rs#L294-L305

This could also be used as a replacement for some of the tests I asked for in your previous PR.

I also suggest adding some CLI tests verifying that this PR doesn't break in combination with `--add-noqa`, `--remove-noqa` `--ignore-noqa` etc.

crates/ruff_linter/src/suppression.rs
```diff
@@ -101,6 +103,35 @@ impl Suppressions {
         let mut builder = SuppressionsBuilder::new(source);
         builder.load_from_tokens(tokens)
     }
+
+    /// Check reported diagnostics against the set of valid range suppressions, and remove any
+    /// diagnostics from the context that should be suppressed by those ranges.
+    pub(crate) fn filter_diagnostics(&self, context: &mut LintContext) {
+        let mut ignored: Vec<usize> = vec![];
```
MichaReiser: We should probably at a fast path here to exit immediately if `self.valid` is empty

amyreese:
> I also suggest adding some CLI tests verifying that this PR doesn't break in combination with `--add-noqa`, `--remove-noqa` `--ignore-noqa` etc.

Is there an existing set of CLI tests for noqa that I can use as an example here?

MichaReiser:
> Is there an existing set of CLI tests for noqa that I can use as an example here?

There are some here https://github.com/astral-sh/ruff/blob/81f6ec16579151d6ad41b2a071120ccc7ea1261f/crates/ruff/tests/cli/lint.rs#L1444

I couldn't find any for `--ignore-noqa`, and `--remove-noqa` doesn't exist, it's just fix and you can find some tests in the above mentioned test file

crates/ruff_linter/resources/test/fixtures/ruff/suppressions.py
```diff
@@ -0,0 +1,45 @@
...
+    # These should both be ignored by the range suppression.
+    # ruff: disable[E741]
+    # ruff: disable[F841]
+    I = 1
+    # ruff: enable[E741]
+    # ruff: enable[F841]
+
+
+def f():
+    # One should both be ignored by the range suppression, and
```
MichaReiser:
```suggestion
    # One should be ignored by the range suppression, and
```

crates/ruff_linter/resources/test/fixtures/ruff/suppressions.py
```diff
@@ -0,0 +1,45 @@
...
+    # the other logged to the user.
+    # ruff: disable[E741]
+    I = 1
+    # ruff: enable[E741]
+
+
+def f():
+    # Neither of these are ignored and warning is
```
MichaReiser:
```suggestion
    # Neither of these are ignored and a warning is
```

crates/ruff_linter/resources/test/fixtures/ruff/suppressions.py
```diff
@@ -0,0 +1,45 @@
...
+    # ruff:enable[E741,F841]
+
+
+def f():
+    # These should both be ignored by the range suppression.
+    # ruff: disable[E741]
+    # ruff: disable[F841]
+    I = 1
+    # ruff: enable[E741]
+    # ruff: enable[F841]
```
MichaReiser: I think it would be good to add/move some of the more advanced test cases involving blocks and make them integration tests instead.

MichaReiser (approved):
This looks good. But we should look into the performance regression before landing this PR (can be fixed as a separate PR but we can't land this until we fixed the regression)

amyreese:
Note: add `--preview` gating so we can land this without waiting for diagnostics etc

crates/ruff/tests/cli/lint.rs
```diff
@@ -1632,6 +1704,100 @@ def unused(x):  # noqa: ANN001, ARG001, D103
...
+"#), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    ");
+
+    let test_code =
+        fs::read_to_string(fixture.root().join("noqa.py")).expect("should read test file");
```
MichaReiser, line 1744: We try to avoid reading fixture files in CLI tests as it makes the tests depend on each other and it can also become harder to reason about what we're testing here. Would it be possible to extract the specific case you want to test from `noqa.py`?

The tests here also don't need to be exhaustive. We can add more exhaustive tests to `add_noqa` (I think we have some integration tests, right?)

crates/ruff_linter/src/linter.rs
```diff
@@ -416,6 +424,12 @@ pub fn add_noqa_to_path(
         target_version,
     );
 
+    let suppressions = if is_range_suppressions_enabled(settings) {
+        Suppressions::from_tokens(locator.contents(), parsed.tokens())
+    } else {
+        Suppressions::default()
+    };
+
```
MichaReiser: We now end up extracting the suppressions twice: Once in `check_path` and once here. We might need to pass the suppressions as part of `check_path`

crates/ruff_linter/src/noqa.rs
```diff
@@ -38,7 +39,16 @@ pub fn generate_noqa_edits(
     let file_directives = FileNoqaDirectives::extract(locator, comment_ranges, external, path);
     let exemption = FileExemption::from(&file_directives);
     let directives = NoqaDirectives::from_commented_ranges(comment_ranges, external, path, locator);
-    let comments = find_noqa_comments(diagnostics, locator, &exemption, &directives, noqa_line_for);
+    // This is called by ruff_server, which already filtered diagnostics via check_path
```
MichaReiser: We also call this function from our tests. Would it make sense to lift `Suppressions` out of `generate_noqa_edits` and instead require the caller to pass the suppressions?

crates/ruff_linter/src/rules/ruff/mod.rs
```diff
@@ -305,6 +305,35 @@ mod tests {
...
+        let diagnostics = test_path(
+            Path::new("ruff/suppressions.py"),
+            &settings::LinterSettings::for_rules(vec![
+                Rule::UnusedVariable,
+                Rule::AmbiguousVariableName,
+                Rule::UnusedNOQA,
+            ])
+            .with_preview_mode(),
+        )?;
+        assert_diagnostics!(diagnostics);
```
MichaReiser: This might be a good case for `assert_diagnostics_diff` which snapshots the difference between two settings: E.g. between preview on and off

crates/ruff/tests/cli/lint.rs
amyreese: This was a copy-paste of the tests directly above and below it that exercise `--add-noqa`.

PR #21627, opened 2025-11-25
[ty] Implement `typing.override`

AlexWaygood:
## Summary

Part of https://github.com/astral-sh/ty/issues/155. This implements the basic check (`@override`-decorated methods should override things!), but not the strict check specified in https://typing.python.org/en/latest/spec/class-compat.html#strict-enforcement-per-project, which should be a separate error code.

## Test Plan

mdtests and snapshots

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/override.md
+++ b/crates/ty_python_semantic/resources/mdtest/override.md
@@ -177,7 +177,7 @@ class Foo:
 ## Overloads
 
 The typing spec states that for an overloaded method, `@override` should only be applied to the
-implementation function. However, we nonetheless respect the decorator in this situation, even =
+implementation function. However, we nonetheless respect the decorator in this situation, even
 though we also emit `invalid-overload` on these methods.
 
 ```py
--- a/crates/ty_python_semantic/src/types/liskov.rs
+++ b/crates/ty_python_semantic/src/types/liskov.rs
@@ -129,7 +129,8 @@ fn check_class_declaration<'db>(
             .member(db, &member.name)
             .place
         else {
-            continue;
+            // If not defined on any superclass, no point in continuing to walk up the MRO
+            break;
         };
 
         let Some(superclass_type_as_callable) = superclass_type


AlexWaygood:
The typing conformance suite results are all good -- all the places where we're newly emitting diagnostics are marked with `# E`.

The ecosystem results are also excellent -- this is one of the few new diagnostics where you'd really expect _no_ new ecosystem hits, and indeed that's what we have!

crates/ty_python_semantic/resources/mdtest/override.md
```diff
@@ -0,0 +1,314 @@
...
+    @overload
+    def baz(self, x: int) -> int: ...
+    # error: [invalid-overload] "`@override` decorator should be applied only to the overload implementation"
+    # error: [explicit-override]
+    def baz(self, x: str | int) -> str | int:
+        return x
+```
+
+In a stub file, `@override` should always be applied to the first overload. Even if it isn't, we
+always emit `explicit-override` diagnostics on the first overload.
```
AlexWaygood: It's a bit annoying to implement this but it means that we pass several typing conformance-suite tests that we'd otherwise fail, so I think it's worth hit

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1545,6 +1548,42 @@ declare_lint! {
...
+    ///
+    /// class C:
+    ///     @override
+    ///     def __repr__(self): ...  # fine: overrides `object.__repr__`
+    ///
+    /// class D(A):
+    ///     @override
+    ///     def foo(self): ...  # fine: overrides `A.foo`
+    /// ```
+    pub(crate) static EXPLICIT_OVERRIDE = {
```
MichaReiser: I think I'd prefer `invalid-override` or `invalid-explicit-override`. `explicit-override` sounds like explicit overrides are a bad thing.

And `incorrect-override` for an override that doesn't match the signature of the parent class

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -143,23 +154,77 @@ fn check_class_declaration<'db>(
...
+    if !subclass_overrides_superclass_declaration && !has_dynamic_superclass {
+        if has_typeddict_in_mro {
+            if !KnownClass::TypedDictFallback
+                .to_instance(db)
+                .member(db, &member.name)
+                .place
+                .is_undefined()
+            {
+                subclass_overrides_superclass_declaration = true;
+            }
```
AlexWaygood, line 170: methods on `TypedDict` subclasses are invalid (as noted in comments in `liskov.md`)... but they exist in the wild anyway, and some of them are decorated with `@override`: https://github.com/Toufool/AutoSplit/blob/d9c5e3b48eeda52a13c08e75b5ffd61873ce5b8f/src/user_profile.py#L50-L53

crates/ty_python_semantic/src/types/diagnostic.rs
AlexWaygood: yeah, I wasn't sure what to call this. I don't want to call it `invalid-override` because that implies a Liskov violation to me (e.g. the name for the "bad method override" diagnostic is `invalid-method-override`).

`invalid-explicit-override` is okay, I can switch to that. (Edit: I pushed this rename.)

crates/ty_python_semantic/resources/mdtest/override.md
```diff
@@ -0,0 +1,314 @@
...
+    @overload
+    def baz(self, x: int) -> int: ...
+    # error: [invalid-overload] "`@override` decorator should be applied only to the overload implementation"
+    # error: [invalid-explicit-override]
+    def baz(self, x: str | int) -> str | int:
+        return x
+```
+
+In a stub file, `@override` should always be applied to the first overload. Even if it isn't, we
+always emit `invalid-explicit-override` diagnostics on the first overload.
```
AlexWaygood, line 220: It's a bit annoying to implement this but it means that we pass several typing conformance-suite tests that we'd otherwise fail, so I think it's worth it

crates/ty_python_semantic/src/types/diagnostic.rs
MichaReiser: Hmm yeah, this is tricky. `invalid-explicit-override` is a bit long.

Should `invalid-method-override` be `incorrect-method-override` to make the distinction clearer? What name do you have in mind for the strict enforcement?

* Pyrefly: bad-override

AlexWaygood: > `invalid-explicit-override` is a bit long.

not as long as some of the ones we already have, such as `subclass-of-final-class` 😄

> Should `invalid-method-override` be `incorrect-method-override` to make the distinction clearer?

I'm not sure that would make the distinction much clearer. Applying `@override` to a method that doesn't override anything is just as incorrect as overriding a method in a way that violates the Liskov Substitution Principle.

> What name do you have in mind for the strict enforcement?

The one I'm using in my head right now is `invalid-explicit-override-strict` or similar. But it's not great.

Another option here would be for the strict enforcement would actually just be the same error code, but we'd have a `tool.ty.explicit-override-enforcement = true` configuration setting (etc.) that allows you to opt into the stricter checks. That would obviate the need for a separate error code, which is nice. But the strict enforcement is really detecting a conceptually pretty different (and far more pedantic) problem -- so a separate error code does feel appropriate in some ways.

Do you have any suggestions?

AlexWaygood: Other possible names for the Liskov violation code might be:
- `unsound-method-override`
- `liskov-violating-override`
- `incompatible-method-override`

But I think even if we went with any of those for the Liskov violation error, I would still want the term "explicit" somewhere in the error code for invalid uses of `typing.override`, so that we're clear that _this_ error code isn't about overrides in general, it's about overrides that are demarcated with `typing.override`.

AlexWaygood: I do also think that the most important thing is to get this check in soon. It's very easy to change the name of an error code later, especially while we still declare ty to be alpha or beta software

crates/ty_python_semantic/resources/mdtest/override.md
```diff
@@ -0,0 +1,314 @@
...
+    @identity
+    @identity
+    @identity
+    def bar(self): ...  # error: [invalid-explicit-override]
+```
+
+## Overloads
+
+The typing spec states that for an overloaded method, `@override` should only be applied to the
+implementation function. However, we nonetheless respect the decorator in this situation, even =
```
carljm:
```suggestion
implementation function. However, we nonetheless respect the decorator in this situation, even
```

crates/ty_python_semantic/resources/mdtest/override.md
```diff
@@ -0,0 +1,314 @@
...
+class Child(ParentDataclass):
+    @override
+    def __lt__(self, other: ParentDataclass) -> bool: ...  # fine
+
+class MyNamedTuple(NamedTuple):
+    x: int
+
+    @override
+    # TODO: this raises an exception at runtime (which we should emit a diagnostic for).
+    # It shouldn't be an `invalid-explicit-override` diagnostic, however.
```
carljm, line 287: I don't think an `invalid-explicit-override` would be _wrong_ here, exactly -- `NamedTuple` isn't really a base class, so you aren't overriding anything here. But I think it's fine if we don't emit it.

crates/ty_python_semantic/src/types/diagnostic.rs
carljm: IMO `invalid-explicit-override` is fine for this code. I think for the strict check the right name would be `invalid-implicit-override`, since the nature of the strict check is "you failed to use `@override` here, but you are required to" -- that is, you are making an implicit override, but overrides are required to be explicit.

crates/ty_python_semantic/src/types/liskov.rs
```diff
@@ -73,52 +76,60 @@ fn check_class_declaration<'db>(
...
+            continue;
         }
 
         let Place::Defined(superclass_type, _, _) = Type::instance(db, superclass)
             .member(db, &member.name)
             .place
         else {
-            // If not defined on any superclass, nothing to check
-            break;
+            continue;
```
carljm: I'm not sure why this was changed? All tests pass if I change this back to `break`. The previous rationale (and comment) still seem correct to me. I think continuing to iterate the MRO after we've established that `Type::member` returns undefined (which internally already checked the rest of the MRO) is wasteful?

```suggestion
            // If not defined on any superclass, no point in continuing to walk up the MRO
            break;
```

carljm (approved):
Looks good to me!

PR #21629, opened 2025-11-25
[ty] Add "import ..." code-action for unresolved references

Gankra:
## Summary

Originally I planned to feed this in as a `fix` but I realized that we probably don't want to be trying to resolve import suggestions while we're doing type inference. Thus I implemented this as a fallback when there's no fixes on a diagnostic, which can use the full lsp machinery.

Fixes https://github.com/astral-sh/ty/issues/1552

## Test Plan

Works in the IDE, added some e2e tests.

### DIFF
--- a/playground/ty/src/Editor/Editor.tsx
+++ b/playground/ty/src/Editor/Editor.tsx
@@ -623,7 +623,10 @@ class PlaygroundServer
         continue;
       }
 
-      const codeActions = this.props.workspace.codeActions(fileHandle, diagnostic.raw);
+      const codeActions = this.props.workspace.codeActions(
+        fileHandle,
+        diagnostic.raw,
+      );
       if (codeActions == null) {
         continue;
       }
@@ -646,7 +649,6 @@ class PlaygroundServer
         });
       }
     }
-
 
     if (actions.length === 0) {
       return undefined;


crates/ty_ide/src/symbols.rs
```diff
@@ -75,13 +89,13 @@ impl QueryPattern {
     /// incorrectly. That is, it's possible that this query will match all
     /// inputs but this still returns `false`.
     pub fn will_match_everything(&self) -> bool {
-        self.re.is_none()
+        self.re.is_none() && self.original.is_empty()
     }
```
Gankra, line 93: Extremely rude bug in this extremely pedantic codepath Andrew was very clear in the comments really should never ever happen, which I have now hijacked.

Gankra:
I added a few speculative e2e tests for cases which don't work (yet) as well. Future work for someone else to implement those :)

crates/ty_ide/src/code_action.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Debug, Clone)]
+pub struct QuickFix {
+    pub title: String,
+    pub edit: Edit,
+}
+
+pub fn code_actions(
+    db: &dyn Db,
+    file: File,
+    range: TextRange,
```
MichaReiser: Nit: Maybe rename to `diagnostic_range` as it wasn't clear to me what range this is

crates/ty_server/src/server/api/requests/code_action.rs
```diff
@@ -24,45 +31,84 @@ impl BackgroundDocumentRequestHandler for CodeActionRequestHandler {
...
+            let Some(range) =
+                diagnostic
+                    .range
+                    .to_text_range(db, file, snapshot.url(), snapshot.encoding())
+            else {
+                continue;
+            };
+            let Some(fixes) = code_actions(db, file, range, diagnostic_id) else {
+                continue;
             };
```
MichaReiser: I think we want to add a `preferred: bool` field to `CodeAction` because, unlike `Diagnostic` fixes, `code_actions` can return multiple results.

MichaReiser (approved):
This is great. I'd prefer for @BurntSushi to have a look at the completion changes

MichaReiser:
https://github.com/astral-sh/ruff/pull/21655 adds code action support to the playground. It would be great if you could integrate your action too

Gankra:
Added playground support (booted it up, works great).

BurntSushi (commented):
LGTM!

PR #21633, opened 2025-11-25
Add Ability to Output to Multiple Files/Formats

### LINKED ISSUE #2388: Allow output to file with different format to output to console
Currently, when running in CI I have to do the following, e.g. in a (simplified) `.gitlab-ci.yml`:

```yaml
lint:
  script:
    - ...
    - ruff .
    - ruff --format=gitlab . > quality.json
  artifacts:
    reports:
      codequality: quality.json
```

I'd like to keep the text output in the job output, but also make use of the capabilities GitLab provides to display violations.

Now, `ruff` is super fast ⚡, so running twice isn't a pain, but it seems suboptimal. Could we enable one of the following?

- `ruff --format=text --output=gitlab:/path/to/quality.json .` ← dedicated `--output` option
- `ruff --format=text --format=gitlab:/path/to/quality.json .` ← allow multiple `--format` with filename
- `ruff --format=text,gitlab:/path/to/quality.json .` ← allow multiple values in `--format`

celestialorb:
## Summary

I have written many CI jobs where Ruff outputs in the GitLab format to a file, but this suppresses outputting to `stdout`/`stderr`. If Ruff finds an issue this will fail the CI job and it won't be immediately obvious to the developer what the lint finding was -- they'd have to go and crack open the GitLab JSON artifact, or view it in GitLab's merge request view.

This change allows us to write both a machine-readable artifact (e.g. GitLab JSON format) as well as output Ruff's findings in a human-readable format without needing to run it twice.

This PR should resolve #2388.

## Test Plan

Tested locally, and hopefully tests will also be run in a CI pipeline.

### DIFF
--- a/crates/ruff/src/args.rs
+++ b/crates/ruff/src/args.rs
@@ -246,12 +246,40 @@ pub struct CheckCommand {
 
     /// Output serialization format for violations.
     /// The default serialization format is "full".
+    ///
+    /// Note: For multiple output formats, use --output-file with format:target syntax instead.
+    /// Example: --output-file gitlab:gl-code-quality.json --output-file full:stdout
     #[arg(long, value_enum, env = "RUFF_OUTPUT_FORMAT")]
     pub output_format: Option<OutputFormat>,
 
-    /// Specify file to write the linter output to (default: stdout).
-    #[arg(short, long, env = "RUFF_OUTPUT_FILE")]
-    pub output_file: Option<PathBuf>,
+    /// Specify output destination(s) using format:target syntax.
+    /// Can be specified multiple times for multiple outputs.
+    /// Also supports comma-separated values: --output-file gitlab:file.json,full:stdout
+    ///
+    /// Examples:
+    ///   --output-file gitlab:gl-code-quality.json
+    ///   --output-file full:stdout
+    ///   --output-file concise:stderr
+    ///   --output-file gitlab:gl-code-quality.json --output-file full:stdout
+    ///   --output-file gitlab:file.json,full:stdout
+    ///
+    /// Environment variable `RUFF_OUTPUT_FILE` also supports comma-separated values:
+    ///   RUFF_OUTPUT_FILE="gitlab:gl-code-quality.json,full:stdout"
+    ///
+    /// Targets can be: stdout, stderr, or a file path.
+    /// Formats: full, concise, grouped, json, json-lines, junit, github, gitlab, pylint, rdjson, azure, sarif
+    ///
+    /// Legacy: If used with --output-format (without format: prefix), writes that format to the specified file.
+    #[arg(
+        short,
+        long,
+        action = clap::ArgAction::Append,
+        value_parser = OutputTargetPairParser,
+        env = "RUFF_OUTPUT_FILE",
+        value_delimiter = ',',
+        value_name = "FORMAT:TARGET"
+    )]
+    pub output_file: Vec<OutputTargetPairOrPath>,
     /// The minimum Python version that should be supported.
     #[arg(long, value_enum)]
     pub target_version: Option<PythonVersion>,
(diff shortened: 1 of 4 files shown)


penguinolog:
Have to added context: GitLab comment code in PR only for the most expensive subscription and not all developers immediately rushing to read CodeQuality report.
Having issues displayed at CI CLI output will be serious time saver.

MichaReiser:
I'd like to explore some more designs on the CLI argument. I don't think what's recommended in this PR has good ergonomics. 

E.g. some alternatives:

* allow `--output-file` multiple times where the value can be `<path>` or `<format:path>`. 
* match `--output-file` with their corresponding `--output-format`: `--output-file path --output-format concise`: This might be tricky to implement and understand

What bothers me a bit is that `--output-format` mostly becomes irrelevant when using `--output-file=<format:path>` 

Have you looked at what other tools do (not just pylint)?


For example, this is what vitest does:

```
--outputFile <filename/-s>                            Write test results to a file when supporter reporter is also specified, use cac's dot notation for individual outputs of multiple reporters (example: --outputFile.tap=./tap.txt)
```

**golang CLI** 

```
golangci-lint run \
  --output.text.path=stdout \
  --output.json.path=./report.json \
  --output.checkstyle.path=./report.xml
```

We might also be able to get away with something easier because what I understand is that this is a CI specific issue and the only thing people need is to allow one output file, but keeping the default CLI output. I didn't see any request where someone actually wants to write 3 outputs at once.

celestialorb:
> I'd like to explore some more designs on the CLI argument. I don't think what's recommended in this PR has good ergonomics.
> 
> E.g. some alternatives:
> 
> * allow `--output-file` multiple times where the value can be `<path>` or `<format:path>`.
> * match `--output-file` with their corresponding `--output-format`: `--output-file path --output-format concise`: This might be tricky to implement and understand
> 
> What bothers me a bit is that `--output-format` mostly becomes irrelevant when using `--output-file=<format:path>`
> 
> Have you looked at what other tools do (not just pylint)?
> 
> For example, this is what vitest does:
> 
> ```
> --outputFile <filename/-s>                            Write test results to a file when supporter reporter is also specified, use cac's dot notation for individual outputs of multiple reporters (example: --outputFile.tap=./tap.txt)
> ```
> 
> **golang CLI**
> 
> ```
> golangci-lint run \
>   --output.text.path=stdout \
>   --output.json.path=./report.json \
>   --output.checkstyle.path=./report.xml
> ```
> 
> We might also be able to get away with something easier because what I understand is that this is a CI specific issue and the only thing people need is to allow one output file, but keeping the default CLI output. I didn't see any request where someone actually wants to write 3 outputs at once.

Yeah, the main use case here is to produce two outputs, a human-readable (`stdout/stderr`) report for developers and a machine-readable report for integration into development systems (e.g. GitHub, GitLab). I can't envision someone needing to produce multiple machine-readable reports of different formats in a single job/context.

What do you think of an approach where we tweak the behavior to _always_ produce a human-readable report in the console? The invocation might be akin to:
```
ruff check
  --output-console-format=concise
  --output-file-format=gitlab
  --output-file-path=./reports/codequality.json
```

We could also introduce a new console format to disable reporting findings out to the console, should that behavior be desired. Thus we'd have a few different options for this proposed `--output-console-format` parameter: `none`, `text`, `concise`, `full`, etc.

The minor downside here is that we're introducing new parameters and we'd probably want to deprecate the old ones if we're to use something like this.

MichaReiser:
I'd prefer not to introduce new parameters if possible. I also think that `--output-format-console` is too long for a parameter that isn't that uncommonly used.

MichaReiser:
It might be worth taking a look at what basedpyright does here https://docs.basedpyright.com/latest/benefits-over-pyright/improved-ci-integration/

We also discussed in the past to automatically change the behavior based on wheter we dedect that ruff runs in CI (where the CI variable is set)

MichaReiser:
I'll close this PR as it has stalled. We can continue the design discussion on the issue.

PR #21635, opened 2025-11-26
[ty] Use `zip` to perform explicit specialization

dhruvmanila:
## Summary

This PR updates the explicit specialization logic to avoid using the call machinery.

Previously, the logic would use the call machinery by converting the list of type variables into a `Binding` with a single `Signature` where all the type variables are positional-only parameters with bounds and constraints as the annotated type and the default type as the default parameter value. This has the advantage that it doesn't need to implement any specific logic but the disadvantages are subpar diagnostic messages as it would use the ones specific to a function call. But, an important disadvantage is that the kind of type variable is lost in this translation which becomes important in #21445 where a `ParamSpec` can specialize into a list of types which is provided using list literal. For example,

```py
class Foo[T, **P]: ...

Foo[int, [int, str]]
```

This PR converts the logic to use a simple loop using `zip_longest` as all type variables and their corresponding type argument maps on a 1-1 basis. They cannot be specified using keyword argument either e.g., `dict[_VT=str, _KT=int]` is invalid.

This PR also makes an initial attempt to improve the diagnostic message to specifically target the specialization part by using words like "type argument" instead of just "argument" and including information like the type variable, bounds, and constraints. Further improvements can be made by highlighting the type variable definition or the bounds / constraints as a sub-diagnostic but I'm going to leave that as a follow-up.

## Test Plan

Update messages in existing test cases.


crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1383,6 +1384,36 @@ declare_lint! {
...
+    /// Bar[int]  # error: Too few type arguments (expected 2, got 1)
+    ///
+    /// from typing import TypeVar
+    /// T = TypeVar('T', int, str)
+    /// class Baz(Generic[T]): ...
+    /// Baz[bool]  # error: Type 'bool' does not satisfy constraints of type variable 'T'
+    /// ```
+    pub(crate) static INVALID_TYPE_ARGUMENTS = {
+        summary: "detects invalid type arguments in generic specialization",
+        status: LintStatus::stable("0.0.1-alpha.1"),
```
MichaReiser: TODO

crates/ty_python_semantic/resources/mdtest/generics/legacy/classes.md
```diff
@@ -165,11 +165,11 @@ reveal_type(Bounded[int]())  # revealed: Bounded[int]
 reveal_type(Bounded[IntSubclass]())  # revealed: Bounded[IntSubclass]
 
 # TODO: update this diagnostic to talk about type parameters and specializations
-# error: [invalid-argument-type] "Argument to class `Bounded` is incorrect: Expected `int`, found `str`"
+# error: [invalid-type-arguments] "Type `str` is not assignable to upper bound `int` of type variable `BoundedT@Bounded`"
 reveal_type(Bounded[str]())  # revealed: Unknown
```
dhruvmanila: I think this and other similar TODOs are now resolved, I'll remove them.

dhruvmanila:
Hmm, I expected the diff numbers to be same, need to see where the difference is.

MichaReiser:
I remember that @sharkdp saw something similar in one of his past PRs and he added an explanation of when this can happen but I'm unable to find that PR right now

sharkdp:
> I remember that @sharkdp saw something similar in one of his past PRs and he added an explanation of when this can happen but I'm unable to find that PR right now

Yes, see: https://github.com/astral-sh/ruff/pull/21476#issuecomment-3546898310.

dhruvmanila:
Oh, I see where the difference lies:

The source code is:

```python
_ResamplerGroupBy: TypeAlias = (
    DatetimeIndexResamplerGroupby[NDFrameT]  # ty: ignore[invalid-argument-type]
    | PeriodIndexResamplerGroupby[NDFrameT]  # ty: ignore[invalid-argument-type]
    | TimedeltaIndexResamplerGroupby[NDFrameT]  # ty: ignore[invalid-argument-type]
)
```

Here, the code has been changed from `invalid-argument-type` to `invalid-type-arguments` which results in the following new diagnostics:

```diff
+ pandas-stubs/core/groupby/groupby.pyi:74:35: error[invalid-type-arguments] Type `typing.TypeVar` is not assignable to upper bound `NDFrame` of type variable `NDFrameT@DatetimeIndexResamplerGroupby`
+ pandas-stubs/core/groupby/groupby.pyi:74:46: warning[unused-ignore-comment] Unused `ty: ignore` directive
+ pandas-stubs/core/groupby/groupby.pyi:75:35: error[invalid-type-arguments] Type `typing.TypeVar` is not assignable to upper bound `NDFrame` of type variable `NDFrameT@PeriodIndexResamplerGroupby`
+ pandas-stubs/core/groupby/groupby.pyi:75:46: warning[unused-ignore-comment] Unused `ty: ignore` directive
+ pandas-stubs/core/groupby/groupby.pyi:76:38: error[invalid-type-arguments] Type `typing.TypeVar` is not assignable to upper bound `NDFrame` of type variable `NDFrameT@TimedeltaIndexResamplerGroupby`
+ pandas-stubs/core/groupby/groupby.pyi:76:49: warning[unused-ignore-comment] Unused `ty: ignore` directive
```

These are the additional 6 diagnostics.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11167,45 +11167,175 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
         generic_context: GenericContext<'db>,
         specialize: impl FnOnce(&[Option<Type<'db>>]) -> Type<'db>,
     ) -> Type<'db> {
+        let db = self.db();
```
dcreager, line 11224: As a complete aside, I've come to the opinion that we should always be passing `db` as a parameter instead of storing it locally as a field. When we use the pattern here (saving it into a local variable), it's sometimes just to save typing from all of the places that we're doing `self.db()` (or `self.db` — another inconsistency!). But other times it's load bearing, and needed to satisfy the borrow checker! Committing to just always threading it through as a parameter would be more consistent and would eliminate the borrowck problems completely. The only places where we should store a `db` into a field is when we have to, to be able to implement some external trait.

But that's just me venting! No suggested change for this PR.

\</end of soapbox>

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11167,45 +11167,175 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-
-        specialize(overload.parameter_types())
+
+        let mut has_error = false;
+
+        for (index, item) in typevars.zip_longest(type_arguments.iter()).enumerate() {
+            match item {
+                EitherOrBoth::Both(typevar, &provided_type) => {
+                    if typevar.default_type(db).is_some() {
+                        typevar_with_defaults += 1;
```
dcreager, line 11269: This might be something you can instead track by taking the index if the first typevar with a default, since (I think) all of the defaulty typevars have to be grouped together at the end. For instance with

```py
class C[T, U = int, V]: ...
```

I don't think you can instantiate `C` with only two types (`C[str, bool]`). That would assume the `bool` lines up with `U`, and you haven't provided a type for `V`.

That said...I don't think that would be any simpler than what you've done here, so I would say leave it as is.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11167,45 +11167,175 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                                {
+                                    builder.into_diagnostic(format_args!(
+                                        "Type `{}` does not satisfy constraints `{}` \
+                                            of type variable `{}`",
+                                        provided_type.display(db),
+                                        constraints
+                                            .elements(db)
+                                            .iter()
+                                            .map(|c| c.display(db))
+                                            .join("`, `"),
```
dcreager: nit: I think you can use `Itertools::format` here instead of `Itertools::join`. It produces the same result, but as an `impl Display` instead of a `String`.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11167,45 +11167,175 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                        missing_typevars.push(typevar);
+                    } else {
+                        typevar_with_defaults += 1;
+                    }
+                    specialization_types.push(None);
+                }
+                EitherOrBoth::Right(_) => {
+                    if first_excess_type_argument_index.is_none() {
+                        first_excess_type_argument_index = Some(index);
+                    }
```
dcreager:
```suggestion
                    first_excess_type_argument_index.get_or_insert(index);
```

crates/ty_python_semantic/src/types/infer/builder.rs
dhruvmanila: Yeah, the typevar with default needs to be grouped together at the end of the list. I don't think we're raising any diagnostic for that (https://play.ty.dev/776bac63-9407-4ebf-acd7-d56d2c174f25) which is something that should be easy to do now. I can do it as a quick follow-up.

dhruvmanila: Oh wait, this is a SyntaxError in CPython for PEP 695 type variables but it's a TypeError for the legacy type variables. I'll raise a ticket for now.

dhruvmanila: https://github.com/astral-sh/ty/issues/1651 and https://github.com/astral-sh/ruff/issues/17412#issuecomment-3584088217

PR #21636, opened 2025-11-26
[ty] Semantic tokens: mark comprehension targets as definitions

lucach:
## Summary

In #21521 I went through all the _statements_ that may introduce names and marked those names as definitions when computing the semantic tokens. However, names can also be introduced by _expressions_. Some were already covered (lambdas, named expressions), but comprehensions/generators were not. This commit fixes that, and hopefully completes the set of places that can introduce names.

## Test Plan

Added a test to check that the targets of list/set/dict comprehensions and generator expressions are marked as definitions.


MichaReiser (approved):
Thank you

PR #21644, opened 2025-11-26
Improve docs on how to stop Ruff and ty disagreeing with each other

AlexWaygood:
## Summary

Lots of Ruff rules encourage you to make changes that might then cause ty to start complaining about Liskov violations. Most of these Ruff rules already refrain from complaining about a method if they see that the method is decorated with `@override`, but this usually isn't documented. This PR updates the docs of many Ruff rules to note that they refrain from complaining about `@override`-decorated methods, and also adds a similar note to the ty `invalid-method-override` documentation.

Helps with https://github.com/astral-sh/ty/issues/1644#issuecomment-3581663859

## Test Plan

- `uvx prek run -a` locally
- CI on this PR


crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -2072,7 +2072,27 @@ declare_lint! {
...
+    /// these suggestions can cause ty to start reporting an `invalid-method-override` error if
+    /// the function in question is a method on a subclass that overrides a method on a superclass,
+    /// and the change would cause the subclass method to no longer accept all argument combinations
+    /// that the superclass method accepts.
+    ///
+    /// This can usually be resolved by adding [`@typing.override`][override] to your method
+    /// definition. Ruff knows that a method decorated with `@typing.override` is intended to
+    /// override a method by the same name on a superclass, and avoids reporting rules like ARG002
+    /// for such methods; it knows that the changes recommended by ARG002 would violate the Liskov
+    /// Substitution Principle.
```
MichaReiser, line 2090: Someday this will no longer be required, because Ruff could check itself whether this is an overriden method.

MichaReiser (approved):

PR #21645, opened 2025-11-26
docs: update reference to the pre-commit hook

JumboBear:
## Summary

The reference to the pre-commit hook inside the tutorial was to the legacy alias `ruff` instead of the current `ruff-check`.

Ref: https://github.com/astral-sh/ruff-pre-commit/pull/124

## Test Plan

Not applicable.


ntBre (approved):
Thank you!

PR #21646, opened 2025-11-26
[ty] Implement `typing.final` for methods

AlexWaygood:
## Summary

A method or property on a superclass decorated with `@final` cannot be overridden on a subclass.

Fixes https://github.com/astral-sh/ty/issues/546.

This currently only implements the check for properties and function-literal types in the class body, not for methods that end up as having other types in the class body (e.g. those decorated with `@lru_cache`). I'll open a followup issue tracking that, if this is merged.

## Test Plan

mdtests/snapshots

PR #21649, opened 2025-11-26
[ty] don't iterate over a hashset

carljm:
## Summary

This caused "deterministic but chaotic" ordering of some intersection types in diagnostics. When calling a union, we infer the argument type once per matching parameter type, intersecting the inferred types for the argument expression, and we did that in an unpredictable order.

We do need a hashset here for de-duplication. Sometimes we call large unions where the type for a given parameter is the same across the union, we should infer the argument once per parameter type, not once per union element. So use an `FxIndexSet` instead of an `FxHashSet`.

## Test Plan

With this change, switching between `main` and https://github.com/astral-sh/ruff/pull/21646 no longer changes the ordering of the intersection type in the test in https://github.com/astral-sh/ruff/pull/21646/commits/cca3a8045df3a1038601e848b87b2163c81aebed

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -6894,11 +6894,14 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
                 // Infer the type of each argument once with each distinct parameter type as type context.
                 let parameter_types = overloads_with_binding
                     .iter()
-                    .filter_map(|(overload, binding)| parameter_type(overload, binding))
-                    // ensure we remove duplicates; sometimes there can be a lot of them
-                    .collect::<FxIndexSet<_>>();
+                    .filter_map(|(overload, binding)| parameter_type(overload, binding));
+
+                let mut seen = FxHashSet::default();
 
                 for parameter_type in parameter_types {
+                    if !seen.insert(parameter_type) {
+                        continue;
+                    }
                     let inferred_ty =
                         self.infer_expression(ast_argument, TypeContext::new(Some(parameter_type)));
 


ibraheemdev:
Good catch, thanks.

We could simplify this a bit further (but up to you).
```diff
- let parameter_types = overloads_with_binding
-     .iter()
-     .filter_map(|(overload, binding)| parameter_type(overload, binding));
- for parameter_type in parameter_types {
+ for (overload, binding) in overloads_with_bindings {
+     let parameter_type = parameter_type(overload, binding);
```

ibraheemdev:
Hmm.. the mypy-primer diagnostic changes are somewhat concerning.

carljm:
> We could simplify this a bit further (but up to you).

I think we can't do that, because of the `filter_map` (`parameter_type(...)` returns an option). Or, we can, but it means adding a conditional `continue`, or `if let` and extra indentation level, in the loop, not sure that's really better.

AlexWaygood:
> Hmm.. the mypy-primer diagnostic changes are somewhat concerning.

I'm also seeing them on https://github.com/astral-sh/ruff/pull/21644, which only touches documentation. I think there's been more than one new cause of nondeterminism over the last few days :-) it looks related to https://github.com/astral-sh/ruff/pull/20566

carljm:
Yes, I don't think this fixes the only issue.

The dd-trace-py change looks like this PR, the other changes less so.

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -6894,10 +6894,14 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-                    .filter_map(|(overload, binding)| parameter_type(overload, binding))
-                    .collect::<FxHashSet<_>>();
+                    .filter_map(|(overload, binding)| parameter_type(overload, binding));
+
+                let mut seen = FxHashSet::default();
 
                 for parameter_type in parameter_types {
+                    if !seen.insert(parameter_type) {
+                        continue;
+                    }
```
MichaReiser, line 6904: It would probably have been worth to adding a comment here or someone like me will come along and change this to a hash set again 😆

PR #21650, opened 2025-11-26
[ty] Support `type[T]` with type variables

ibraheemdev:
## Summary

Adds support for `type[T]`, where `T` is a type variable.

- Resolves https://github.com/astral-sh/ty/issues/501
- Resolves https://github.com/astral-sh/ty/issues/783
- Resolves https://github.com/astral-sh/ty/issues/662

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/ide_support/all_members.md
+++ b/crates/ty_python_semantic/resources/mdtest/ide_support/all_members.md
@@ -110,6 +110,11 @@ static_assert(not has_member(C(), "non_existent"))
 
 ### Class objects
 
+```toml
+[environment]
+python-version = "3.12"
+```
+
 Class-level attributes can also be accessed through the class itself:
 
 ```py
@@ -154,7 +159,13 @@ static_assert(has_member(D, "meta_attr"))
 static_assert(has_member(D, "base_attr"))
 static_assert(has_member(D, "class_attr"))
 
-def f(x: type[D]):
+def _(x: type[D]):
+    static_assert(has_member(x, "meta_base_attr"))
+    static_assert(has_member(x, "meta_attr"))
+    static_assert(has_member(x, "base_attr"))
+    static_assert(has_member(x, "class_attr"))
+
+def _[T: D](x: type[T]):
     static_assert(has_member(x, "meta_base_attr"))
     static_assert(has_member(x, "meta_attr"))
     static_assert(has_member(x, "base_attr"))


crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -1590,9 +1590,16 @@ impl<'db> SpecializationBuilder<'db> {
...
+            }
+
+            (Type::SubclassOf(subclass_of), ty) | (ty, Type::SubclassOf(subclass_of))
+                if subclass_of.is_type_var() =>
+            {
+                let formal_instance = Type::TypeVar(subclass_of.into_type_var().unwrap());
+                if let Some(actual_instance) = ty.to_instance(self.db) {
+                    return self.infer_map_impl(formal_instance, actual_instance, polarity, f);
                 }
             }
```
ibraheemdev, line 1571: This is pretty trivial so I don't think it will overlap much with the constraint solver work, but still flagging it (cc @dcreager).

crates/ty_python_semantic/resources/mdtest/type_of/generics.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+`type[T]` with a constraints `T: (A, B)` represents exactly the class object `A`, or exactly `B`:
+
+```py
+def constrained[T: (int, str)](x: type[T]) -> T:
+    reveal_type(x)  # revealed: type[T@constrained]
+
+    # TODO: This should reveal `T`.
+    reveal_type(x("hello"))  # revealed: int | Literal["hello"]
```
ibraheemdev: This doesn't currently work because we can only treat `type[T]` as a class constructor if `T` has an upper bound of a single concrete class. If there are multiple constraints, we fall back to the call binding logic, which splits `type[T]` into the union of its constraints.

I think we have a general problem of not unifying operations on constrained type variables very well (e.g. [this failing test](https://github.com/astral-sh/ruff/blob/40886963d1c01bd00f27f2b67913504a52d2d85f/crates/ty_python_semantic/resources/mdtest/generics/pep695/functions.md?plain=1#L250)), so I'm not sure this issue can be addressed by this PR.

AlexWaygood:
From the ecosystem hits, it looks like typevars with upper bounds are working great but that there might be some issues with value-constrained type variables on this branch:

```py
def f[T: (int, str), U: int](x: T, y: U) -> tuple[T, U]:
    reveal_type(type(y))  # type[U@f] (good)
    reveal_type(type(y)())  # U@f (good)

    reveal_type(type(x))  # type[T@f] (good)
    reveal_type(type(x)())  # int | Literal[""] (bad -- should be `T@f`)

    return type(x)(), type(y)()  # error[invalid-return-type]: Return type does not match returned value: expected `tuple[T@f, U@f]`, found `tuple[int | Literal[""], U@f]
```

Edit: Ah, I see you already have failing tests for this in https://github.com/astral-sh/ruff/pull/21650/files#r2566765667

AlexWaygood:
This ecosystem diagnostic looks odd, because I would expect both `type & Unknown` and `type[_T@AppKey]` to have a `__qualname__` attribute, since all instances of `type` have a `__qualname__` attribute, and all `type[]` types are subtypes of `type`

---

### aiohttp/helpers.py

[warning] possibly-missing-attribute - [:879:26](https://github.com/aio-libs/aiohttp/blob/72fadb8f1a56e8bff0fef23ccf02e2067cd87e41/aiohttp/helpers.py#L879) - Attribute `__qualname__` may be missing on object of type `(Unknown & type) | type[_T@AppKey]`

---

It looks like a minimal repro is:

```py
def f[T](x: type[T]) -> T:
    reveal_type(x.__qualname__)  # error[unresolved-attribute]: Object of type `type[T@f]` has no attribute `__qualname__`
    return x()
```

AlexWaygood:
Another false positive:

```py
from ty_extensions import Intersection, Unknown

def g[T: int](x: type | type[T]) -> T:
    return x()  # fine

def f[T: int](x: Intersection[type, Unknown] | type[T]) -> T:
    # error[invalid-return-type]: Return type does not match returned value: expected `T@f`, found `@Todo(Type::Intersection.call()) | int`
    return x()
```

I would expect the inferred type of `x()` to be `@Todo | T`, which would be assignable to `T`, rather than `@Todo | int` (which is not)

AlexWaygood:
Here's a minimal repro of one of the `cwltool` diagnostics:

```py
from typing import Callable, Any

def needs_a_callable(x: Callable[..., Any]): ...

def f[T](x: T) -> T:
    needs_a_callable(type(x))  # error: [invalid-argument-type]  "Argument to function `needs_a_callable` is incorrect: Expected `(...) -> Any`, found `type[T@f]`"
    return x
```

but that's a false positive: `type[T]` should be assignable to `Callable[..., T]`, and therefore also to `Callable[..., Any]`

AlexWaygood:
Here's a minimal repro of some of the ddtrace diagnostics:

```py
class Foo:
    def method(self):
        super(self.__class__, self)  # error[invalid-super-argument] "`type[Self@method]` is not a valid class"
```

You'll need to edit some code in https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types/bound_super.rs to get rid of that false positive

AlexWaygood:
Here's a repro of one of the discord.py diagnostics:

```py
from typing import Any

class Foo[T]: ...

def f[T: Foo[Any]](x: type[T] = Foo): ...
```

that one looks like it could possibly be tricky to fix? Not sure. Could be fine to just add a failing test for it with a TODO.

crates/ty_python_semantic/src/types.rs
```diff
@@ -2106,6 +2108,42 @@ impl<'db> Type<'db> {
                 ConstraintSet::from(true)
             }
 
+            // `type[T]` is a subtype of the class object `A` if every instance of `T` is a subtype of an instance
+            // of `A`, and vice versa.
+            (Type::SubclassOf(subclass_of), _) if subclass_of.is_type_var() => {
```
AlexWaygood: I suspect the reason why you're running into issues with `type[]` types not being assignable to `Callable` types is that these branches are above the `Type::Callable` branches (but naively moving them below the `Type::Callable` branches causes other issues 🙃)

crates/ty_python_semantic/src/types/subclass_of.rs
```diff
@@ -65,12 +102,19 @@ impl<'db> SubclassOfType<'db> {
     }
 
     /// Return a [`Type`] instance representing the type `type[Any]`.
+    #[allow(dead_code)]
     pub(crate) const fn subclass_of_any() -> Type<'db> {
         Type::SubclassOf(SubclassOfType {
             subclass_of: SubclassOfInner::Dynamic(DynamicType::Any),
         })
     }
```
AlexWaygood, line 98: eh, I think we can just delete this if we're not using it -- it's pretty trivial :P

crates/ty_python_semantic/src/types/subclass_of.rs
```diff
@@ -212,21 +297,22 @@ impl<'db> SubclassOfType<'db> {
...
             .is_some_and(|class| class.class_literal(db).0.is_typed_dict(db))
     }
 }
 
 impl<'db> VarianceInferable<'db> for SubclassOfType<'db> {
     fn variance_of(self, db: &dyn Db, typevar: BoundTypeVarInstance<'_>) -> TypeVarVariance {
         match self.subclass_of {
-            SubclassOfInner::Dynamic(_) => TypeVarVariance::Bivariant,
             SubclassOfInner::Class(class) => class.variance_of(db, typevar),
+            SubclassOfInner::Dynamic(_) | SubclassOfInner::TypeVar(_) => TypeVarVariance::Bivariant,
```
AlexWaygood, line 303: I can never remember how this trait works, but `type[]` types are covariant

crates/ty_python_semantic/src/types/subclass_of.rs
```diff
@@ -249,6 +336,7 @@ impl<'db> VarianceInferable<'db> for SubclassOfType<'db> {
 pub(crate) enum SubclassOfInner<'db> {
     Class(ClassType<'db>),
     Dynamic(DynamicType),
+    TypeVar(BoundTypeVarInstance<'db>),
 }
```
AlexWaygood, line 328: An alternative design might be to just have it be its own `Type` variant (`Type::MetaTypeVar`?). That might save you from a few of your `.unwrap()` calls in `Type::has_relation_to_impl`, for example -- there are quite a few places where you need to treat this `SubclassOfInner` variant differently to all the other ones.

I don't have a strong opinion, though. That design would probably bring its own complications.

crates/ty_python_semantic/src/types/subclass_of.rs
ibraheemdev: This is currently used in property tests. I can change it to `cfg(test)`, but seems fine either way.

crates/ty_python_semantic/src/types/subclass_of.rs
ibraheemdev: `type[]` doesn't have a type variable that we can get the variance of, so I think this is correct. `variance_of` is asking for the variance of the type variable on the generic context, not the assigned type.

crates/ty_python_semantic/src/types/subclass_of.rs
AlexWaygood: ah okay -- I'd be inclined to switch it to `#[cfg(test)]` as then readers of the code won't be confused about why it's still there (like I was, when reading the PR diff 😅)

crates/ty_python_semantic/src/types.rs
```diff
@@ -2106,6 +2108,44 @@ impl<'db> Type<'db> {
                 ConstraintSet::from(true)
             }
 
+            // `type[T]` is a subtype of the class object `A` if every instance of `T` is a subtype of an instance
+            // of `A`, and vice versa.
+            (Type::SubclassOf(subclass_of), _)
+                if subclass_of.is_type_var() && !target.is_callable_type() =>
```
AlexWaygood: I think there's a similar issue due to the `Type::ProtocolInstance()` branch appearing below this one -- there are two false positives on this snippet on your branch:

```py
from typing import Protocol

class Callback(Protocol):
    def __call__(self, *args, **kwargs) -> int: ...

class HasName(Protocol):
    __name__: str

def needs_callback(c: Callback): ...
def needs_something_named(obj: HasName): ...

def f[T: int](x: T) -> T:
    needs_callback(type(x))
    needs_callback(type(x))
    return x
```

ibraheemdev:
Thank you for the ecosystem triage @AlexWaygood! I think I addressed all of the issues you reproduced, except this one.
```py
class Foo[T]: ...
def f[T: Foo[Any]](x: type[T] = Foo): ...
```

We emit a similar error without the `type[T]`, because we treat the `T` as a non-inferable type variable. We may have to lazily check the assignability of default types after specialization, but either way I think the issue is unrelated to this PR.
```py
def f[T: int](x: T = 1): ...
```

crates/ty_python_semantic/resources/mdtest/type_of/generics.md
```diff
@@ -0,0 +1,196 @@
...
+    reveal_type(x())  # revealed: int | T@_
+```
+
+## Pivot Class
+
+```py
+class Foo:
+    def method(self):
+        super(self.__class__, self)
+```
```
AlexWaygood: nit: all our other tests like this are in https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/resources/mdtest/class/super.md, so I'd be inclined to put this one there too

crates/ty_python_semantic/src/types.rs
ibraheemdev: Hmm.. right. Unfortunately I think we just have to special case these, or alternatively bring the `Type::Callable` or `Type::Protocol` branches up (at the risk of making the match-arm ordering even more load-bearing :upside_down_face:).

Moving the `type[T]` downcasting branch down breaks things because we want to downcast *before* union expansion, so that `type[T]` is a subtype of `type[T] | None` (the ordering requirements are similar to a bare `T`).

AlexWaygood: Ideally I do indeed think that the structural-type branches would be as high as possible, to avoid issues like this! IIRC, though, weird things start happening if you move them above the set-theoretic and type-variable branches 🙃 So you may well be right that special-casing is the best we can do here.

AlexWaygood:
> We emit a similar error without the `type[T]`, because we treat the `T` as a non-inferable type variable. We may have to lazily check the assignability of default types after specialization, but either way I think the issue is unrelated to this PR.

ah, that makes sense -- thanks! It did some to come up a _fair_ bit in the ecosystem, so it might be worth adding some failing tests with TODOs?

crates/ty_python_semantic/resources/mdtest/type_of/generics.md
```diff
@@ -115,6 +110,9 @@ def _[T](_: T):
     static_assert(is_assignable_to(type[T], Callable[..., T]))
     static_assert(not is_disjoint_from(type[T], Callable[..., T]))
 
+    static_assert(is_assignable_to(type[T], Callback[..., T]))
+    static_assert(not is_disjoint_from(type[T], Callback[..., T]))
```
AlexWaygood: `Callback` isn't a generic type here, so the tests you've added are passing for the wrong reasons -- you've accidentally created lots of `@Todo` types. (We _should_ emit an error on a non-generic class being subscripted.)

```suggestion
    static_assert(is_assignable_to(type[T], Callback[))
    static_assert(not is_disjoint_from(type[T], Callback))
```

If you wanted to make `Callback` generic, you'd have to write it as

```py
T_co = TypeVar("T_co", covariant=True)

class Callback(Protocol[T_co]):
    def __call__(self, *args, **kwargs) -> T_co: ...
```

ibraheemdev:
I did a run through of the typing conformance suite changes. Most of the changes seem correct. There are a couple assertions `assert_type(cls, type[Self])` that are now failing, because we now recognize `type[Self]` but still do not recognize `cls` -- this is tracked in https://github.com/astral-sh/ty/issues/159.

There is another failure [here](https://github.com/python/typing/blob/9f6d8ced7cd1c8d92687a4e9c96d7716452e471e/conformance/tests/specialtypes_type.py#L175), because we consider `type` to be equivalent to `type[object]` and not `type[Any]`, so I believe that one is an intentional divergence from the spec.

Finally, there is one new false positive that seems incorrect,
```py
from typing import TypeVar

T = TypeVar("T")

class Meta3(type):
    def __call__(cls: type[T], *args, **kwargs) -> T:
        # error[invalid-super-argument]: `T@__call__` is not an instance or subclass of `<class 'Meta3'>` in `super(<class 'Meta3'>, T@__call__)` call
        return super().__call__(cls, *args, **kwargs)
```

but I'm not sure if that's an existing issue or not.

AlexWaygood:
> Finally, there is one new false positive that seems incorrect,
> 
> ```python
> from typing import TypeVar
> 
> T = TypeVar("T")
> 
> class Meta3(type):
>     def __call__(cls: type[T], *args, **kwargs) -> T:
>         # error[invalid-super-argument]: `T@__call__` is not an instance or subclass of `<class 'Meta3'>` in `super(<class 'Meta3'>, T@__call__)` call
>         return super().__call__(cls, *args, **kwargs)
> ```
> 
> but I'm not sure if that's an existing issue or not.

you could actually argue that's a true positive. If you change the typevar so that it has an upper bound, the error goes away:

```py
from typing import TypeVar

T = TypeVar("T", bound="Meta3")

class Meta3(type):
    def __call__(cls: type[T], *args, **kwargs) -> T:
        return super().__call__(cls, *args, **kwargs)
```

crates/ty_python_semantic/resources/mdtest/type_of/generics.md
ibraheemdev: Good catch, thanks. I assume it's a known issue that we don't error on the subscript there?

AlexWaygood: yeah. It might actually not be too hard to implement it now, we might have just not got round to it

AlexWaygood:
There's an ecosystem false positive on https://github.com/DataDog/dd-trace-py/blob/84e0f46f4da9a26aa0f741712cf2f9303794ea78/ddtrace/internal/_exceptions.py#L22, a minimal repro of which is:

```py
from typing import TypeVar

E = TypeVar("E", bound=BaseException)


def find_exception( exc: BaseException, exception_type: type[E]) -> E | None:
    if isinstance(exc, exception_type):
        return exc  # error [invalid-return-type] "Return type does not match returned value: expected `E@find_exception | None`, found `BaseException`""
    raise NotImplementedError("no")
```

you can fix the false positive with this patch

```diff
diff --git a/crates/ty_python_semantic/src/types/narrow.rs b/crates/ty_python_semantic/src/types/narrow.rs
index 5ff8b615af..0259e94877 100644
--- a/crates/ty_python_semantic/src/types/narrow.rs
+++ b/crates/ty_python_semantic/src/types/narrow.rs
@@ -171,8 +171,11 @@ impl ClassInfoConstraintFunction {
                 // e.g. `isinstance(x, list[int])` fails at runtime.
                 SubclassOfInner::Class(ClassType::Generic(_)) => None,
                 SubclassOfInner::Dynamic(dynamic) => Some(Type::Dynamic(dynamic)),
-                SubclassOfInner::TypeVar(_) => None,
-            },
+                SubclassOfInner::TypeVar(var) => match self {
+                    ClassInfoConstraintFunction::IsInstance => Some(Type::TypeVar(var)),
+                    ClassInfoConstraintFunction::IsSubclass => Some(classinfo),
+                }
+            }
             Type::Dynamic(_) => Some(classinfo),
             Type::Intersection(intersection) => {
                 if intersection.negative(db).is_empty() {
```

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -176,17 +176,19 @@ impl<'db> AllMembers<'db> {
...
+                _ => {
+                    if let Some(class_type) = subclass_of_type.subclass_of().into_class(db) {
+                        let (class_literal, specialization) = class_type.class_literal(db);
+                        self.extend_with_class_members(db, ty, class_literal);
+                        self.extend_with_synthetic_members(db, ty, class_literal, specialization);
+                        if let Type::ClassLiteral(metaclass) = class_literal.metaclass(db) {
+                            self.extend_with_class_members(db, ty, metaclass);
+                        }
+                    }
+                }
```
AlexWaygood, line 191: this change doesn't look correct -- I think you should keep this an exhaustive match. If a variable is of type `type[U]`, where `U` has an upper bound of `int`, `all_members` should return the same set of members as it would for `type[int]`

ibraheemdev: `into_class` does that conversion internally. The only case this could return `None` is for `type[T]` where `T` has multiple constraints, so there is no single class to return.

EDIT: It actually returns `object` in that case, so this should never return `None`.

AlexWaygood: Ah I see! Yes, we do want to fallback to `type` there, I think. You could also add a couple of tests to `all_members.md` to prove I'm wrong ;)

AlexWaygood:
yessss 
image

AlexWaygood:
This PR fixes https://github.com/astral-sh/ty/issues/662, so it could be worth adding the MRE in that issue as a standalone test

PR #21652, opened 2025-11-27
[ty] support `type[tuple[...]]`

carljm:
Fixes https://github.com/astral-sh/ty/issues/1649

## Summary

We missed this when adding support for `type[]` of a specialized generic.

## Test Plan

Added mdtests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/subscript/tuple.md
+++ b/crates/ty_python_semantic/resources/mdtest/subscript/tuple.md
@@ -329,6 +329,16 @@ reveal_type(tuple[int, str])  # revealed: <class 'tuple[int, str]'>
 reveal_type(tuple[int, ...])  # revealed: <class 'tuple[int, ...]'>
 ```
 
+```py
+from typing import Any
+
+def _(a: type[tuple], b: type[tuple[int]], c: type[tuple[int, ...]], d: type[tuple[Any, ...]]) -> None:
+    reveal_type(a)  # revealed: type[tuple[Unknown, ...]]
+    reveal_type(b)  # revealed: type[tuple[int]]
+    reveal_type(c)  # revealed: type[tuple[int, ...]]
+    reveal_type(d)  # revealed: type[tuple[Any, ...]]
+```
+
 ## Inheritance
 
 ```toml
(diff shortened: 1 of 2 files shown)


ibraheemdev:
Was just about to comment about the typing conformance changes being identical to https://github.com/astral-sh/ruff/pull/21652.. but then I realized this PR was stacked on that one :smile:

crates/ty_python_semantic/src/types/function.rs
```diff
@@ -1230,6 +1230,9 @@ pub enum KnownFunction {
     /// `builtins.__import__`, which returns the top-level module.
     #[strum(serialize = "__import__")]
     DunderImport,
+    /// `collections.namedtuple`
+    #[strum(serialize = "namedtuple")]
+    NamedTuple,
```
AlexWaygood, line 1283: nit: since we have the custom `strum(serialize = "namedtuple")` attribute here anyway, could we call this variant `CollectionsNamedTuple`, to disambiguate it from `SpecialFormType::NamedTuple`?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11242,7 +11242,11 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             Ok(bindings) => bindings,
             Err(CallError(_, bindings)) => {
                 bindings.report_diagnostics(&self.context, subscript.into());
-                return Type::unknown();
+                let unknowns = generic_context
+                    .variables(self.db())
+                    .map(|_| Some(Type::unknown()))
+                    .collect::<Vec<_>>();
```
```suggestion
                let unknowns =
                    vec![Some(Type::unknown()); generic_context.variables(self.db()).len()];
```

AlexWaygood:
oops, I think I just accidentally left some review comments regarding changes that already landed in https://github.com/astral-sh/ruff/commit/77f8fa6906058b935eca7ae78c29322431abb377 (I guess this needs to be rebased now that's landed 😄)

sharkdp:
Re-running the ecosystem checks to pick up the changes in #21721.

carljm:
Thanks for updating and landing this and the previous PR in stack @sharkdp !

PR #21653, opened 2025-11-27
[ty] uniformly weak imports

Gankra:
Split out from #21583 to analyze the effect of this one change.

The code in the base PR gives `import a.b` highest priority, while `from a.b import ...` gets second-lowest priority (beating only `__getattr__`). This PR removes the special handling of `import a.b` so that they get the same priority as `from a.b import...`.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/conflicts.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/conflicts.md
@@ -5,7 +5,7 @@
 ```py
 import a.b
 
-reveal_type(a.b)  # revealed: <module 'a.b'>
+reveal_type(a.b)  # revealed: int
 ```
 
 `a/__init__.py`:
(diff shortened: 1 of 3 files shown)


Gankra:
Ayup, basically just rerolls where we mess up on pwndbg

Gankra:
[pwndbg explicitly shadows it's own submodule with `None` and then in an init subroutine unshadows it](https://github.com/pwndbg/pwndbg/blob/9763fc34370d0fab9ef35c36f3c6c560991dbec8/pwndbg/aglib/__init__.py#L18-L52) with explicit reference to it doing things "so mypy understands"

This is basically the entire diff. Both in added diagnostics and removed diagnostics. Very potato-potatoh situation. pwndbg is messed up either way. I guess this is the motivating case but boy I sure wish they just had an `if TYPE_CHECKING` or something instead...

Gankra:
The other changes are:

* ✅ PyGithub has a call-non-callable fixed
* 😒 One test in dd-trace-py gets more uncertain about a submodule that is explicitly imported in the file
* 😒 A few tests in pandas get more uncertain about a submodule that is explicitly imported in the file

Gankra:
Whoops, that was a bad rebase...

PR #21654, opened 2025-11-27
[ty] Always infer the submodule literal for submodule attribute-acces…

Gankra:
This is #21619 but based on top the #21583 to see if this looks better on top of it.


Gankra:
Looks like it's nearly entirely orthogonal. The wins on mitmproxy go away, I think because the other PR already finds those wins.

PR #21657, opened 2025-11-27
[syntax-error] Default type parameter followed by non-default type parameter

11happy:
## Summary

This PR implements syntax error where a default type parameter is followed by a non-default type parameter. https://github.com/astral-sh/ruff/issues/17412#issuecomment-3584088217

## Test Plan

### DIFF
--- a/crates/ruff_python_parser/tests/snapshots/invalid_syntax@type_param_param_spec_invalid_default_expr.py.snap
+++ b/crates/ruff_python_parser/tests/snapshots/invalid_syntax@type_param_param_spec_invalid_default_expr.py.snap
@@ -375,3 +375,12 @@ Module(
 4 | type X[**P = x := int] = int
 5 | type X[**P = *int] = int
   |
+
+
+  |
+2 | type X[**P = yield x] = int
+3 | type X[**P = yield from x] = int
+4 | type X[**P = x := int] = int
+  |                   ^^^ Syntax Error: non default type parameter `int` follows default type parameter
+5 | type X[**P = *int] = int
+  |
--- a/crates/ruff_python_parser/tests/snapshots/invalid_syntax@type_param_type_var_invalid_default_expr.py.snap
+++ b/crates/ruff_python_parser/tests/snapshots/invalid_syntax@type_param_type_var_invalid_default_expr.py.snap
@@ -459,3 +459,12 @@ Module(
 5 | type X[T = x := int] = int
 6 | type X[T: int = *int] = int
   |
+
+
+  |
+3 | type X[T = (yield x)] = int
+4 | type X[T = yield from x] = int
+5 | type X[T = x := int] = int
+  |                 ^^^ Syntax Error: non default type parameter `int` follows default type parameter
+6 | type X[T: int = *int] = int
+  |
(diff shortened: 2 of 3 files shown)


crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -144,11 +144,18 @@ impl SemanticSyntaxChecker {
...
+                ..
+            }) => {
+                Self::duplicate_type_parameter_name(type_params, ctx);
+                Self::type_parameter_default_order(type_params, ctx);
+            }
+            Stmt::TypeAlias(ast::StmtTypeAlias {
+                type_params: Some(type_params),
+                ..
+            }) => {
+                Self::duplicate_type_parameter_name(type_params, ctx);
```
dhruvmanila, line 155: This check needs to be applied to type aliases as well:

```python
In [1]: type Int[T = int, U] = int
  Cell In[1], line 1
    type Int[T = int, U] = int
                      ^
SyntaxError: non-default type parameter 'U' follows default type parameter
```

crates/ruff_python_parser/src/semantic_errors.rs
```diff
@@ -611,6 +618,37 @@ impl SemanticSyntaxChecker {
...
+                ast::TypeParam::TypeVar(ast::TypeParamTypeVar { default, .. })
+                | ast::TypeParam::TypeVarTuple(ast::TypeParamTypeVarTuple { default, .. })
+                | ast::TypeParam::ParamSpec(ast::TypeParamParamSpec { default, .. }) => {
+                    default.is_some()
+                }
+            };
+
+            if seen_default && !has_default {
+                // test_err type_parameter_default_order
+                // class C[T = int, U]: ...
```
dhruvmanila, line 635: Let's add some cases where the default is not at the start of the list and it has multiple non-default type parameter following a type parameter with default like:
```py
class C[T1, T2 = int, T3, T4]: ...
```

dhruvmanila (changes requested):

crates/ruff_python_parser/src/semantic_errors.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs
11happy: done

crates/ruff_python_parser/src/semantic_errors.rs
11happy: however some additional detection's are there for example -
```
type X[*Ts = x := int] = int
```
in type params:

type_params: TypeParams { range: 6..22, node_index: NodeIndex(None), type_params: [TypeVarTuple(TypeParamTypeVarTuple { node_index: NodeIndex(None), range: 7..14, name: Identifier { id: Name("Ts"), range: 8..10, node_index: NodeIndex(None) }, default: Some(Name(ExprName { node_index: NodeIndex(None), range: 13..14, id: Name("x"), ctx: Load })) }), TypeVar(TypeParamTypeVar { node_index: NodeIndex(None), range: 18..21, name: Identifier { id: Name("int"), range: 18..21, node_index: NodeIndex(None) }, bound: None, default: None })] 

int is getting detected as seperate non default typevar, whereas I think should it be parsed as single expression ?

dhruvmanila: > int is getting detected as seperate non default typevar, whereas I think should it be parsed as single expression ?

That's unrelated to this PR as it's more about how the parser is recovering from an invalid syntax.

dhruvmanila (approved):
Thanks

crates/ruff_python_parser/src/semantic_errors.rs
dhruvmanila: Oh I see what you mean, it's related to the failing tests. I think it's fine for now as it's how the parser is recovering from an invalid syntax. Can you update the snapshots? I can merge then.

11happy: done I have updated snapshots

Issue #21640, opened 2025-11-26
Feature Request: Support formatting for `sphinx.ext.doctest` (`.. testcode::`) in docstrings


sawa3030:
## Summary
When formatting docstrings, Ruff currently does not recognize Python code inside `sphinx.ext.doctest` directive blocks (such as `.. testcode::`). It would be helpful if these doctest-style blocks were detected and formatted in the same way as other supported docstring code blocks.

## Description
`sphinx.ext.doctest` provides directives like `.. testcode::` for writing executable examples in documentation. For example:

```
class ExampleClass:
    """Example class.

    Example:
        .. testcode::
            first_example = ExampleClass()
            print(first_example)
    """
```
ref: [https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html)

At the moment, Ruff does not format the code inside the `.. testcode::` block.
Ref: [https://docs.astral.sh/ruff/formatter/#docstring-formatting](https://docs.astral.sh/ruff/formatter/#docstring-formatting)

It would be great if Ruff could also format the Python code within these blocks, similarly to other reStructuredText code directives.

I’m happy to open a PR to add support for this, if I could.

MichaReiser:
There are other directives that we should support in addition to just `testcode`, e.g. `testsetup` and `testcleanup`


I haven't taken a close look but I think this shouldn't be too hard to add (@BurntSushi feel free to correct me), given that we already support `code-block::`. All that might be needed is to add `testcode`, `testsetup`, ... here:

https://github.com/astral-sh/ruff/blob/7627d3879c26dd3e9510eea8a63239f08dae1cbe/crates/ruff_python_formatter/src/string/docstring.rs#L1025-L1041

dev-miro26:
Hi, @MichaReiser 
As a new contributor on this repo, I want to work on this issue.
If this issue still available, could you assign this to me?
Thanks

PR #21665, opened 2025-11-27
[ty] implement rendering of `.. code:: lang` in docstrings

Gankra:
## Summary

* Fixes https://github.com/astral-sh/ty/issues/1650
* Part of https://github.com/astral-sh/ty/issues/1610

We now handle:

* `.. warning::` (and friends) by bolding the line and rendering the block as normal (non-code) text
* `.. code::` (and friends) by treating it the same as `::` (fully deleted if seen, introduce a code block)
* `.. code:: lang` (and friends) by letting it set the language on the codefence
* `.. versionchanged:: 1.2.3` (and friends) by rendering it like `warning` but with the version included and italicized
* `.. dsfsdf-unknown:: (lang)` by assuming it's the same as `.. code:: (lang)`

## Test Plan

Snapshots added/updated. I also deleted a bunch of useless checks on plaintext rendering. It's important for some edge-case tests but not for the vast majority of tests.

### DIFF
--- a/crates/ty_ide/src/docstring.rs
+++ b/crates/ty_ide/src/docstring.rs
@@ -182,6 +182,11 @@ fn documentation_trim(docs: &str) -> String {
 /// </code>
 /// ```
 fn render_markdown(docstring: &str) -> String {
+    // Here lies a monumemnt to robust parsing and escaping:
+    // a codefence with SO MANY backticks that surely no one will ever accidentally
+    // break out of it, even if they're writing python documentation about markdown
+    // code fences and are showing off how you can use more than 3 backticks.
+    const FENCE: &str = "```````````";
     // TODO: there is a convention that `singletick` is for items that can
     // be looked up in-scope while ``multitick`` is for opaque inline code.
     // While rendering this we should make note of all the `singletick` locations
@@ -194,7 +199,7 @@ fn render_markdown(docstring: &str) -> String {
     let mut starting_literal = None;
     let mut in_literal = false;
     let mut in_any_code = false;
-    let mut _line;
+    let mut temp_owned_line;
     for untrimmed_line in docstring.lines() {
         // We can assume leading whitespace has been normalized
         let mut line = untrimmed_line.trim_start_matches(' ');
@@ -220,7 +225,8 @@ fn render_markdown(docstring: &str) -> String {
             in_literal = false;
             in_any_code = false;
             block_indent = 0;
-            output.push_str("```\n");
+            output.push_str(FENCE);
+            output.push('\n');
         }
 
         // We previously entered a literal block and we just found our first non-blank line
@@ -232,8 +238,8 @@ fn render_markdown(docstring: &str) -> String {
             in_literal = true;
             in_any_code = true;
             block_indent = line_indent;
-            // TODO: I hope people don't have literal blocks about markdown code fence syntax
-            output.push_str("\n```");
+            output.push('\n');
+            output.push_str(FENCE);
             output.push_str(literal);
             output.push('\n');
         }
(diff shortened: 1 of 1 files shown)


Gankra:
My awful son only gets worse

MichaReiser:
Potentially some more sphinx specific candidates https://github.com/astral-sh/ruff/issues/21640

AlexWaygood:
`.. versionadded::`, `.. versionchanged::` and `.. deprecated::` all appear a huge amount in CPython's online documentation (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#describing-changes-between-versions). They might also pop up in CPython docstrings as well from time to time -- not sure? CPython's docstrings aren't really meant to be ReST because they're intended to be viewed directly in a terminal, but sometimes ReST creeps in.

Gankra:
Screenshot 2025-11-27 at 12 39 17 PM

crates/ty_ide/src/docstring.rs
```diff
@@ -246,21 +248,74 @@ fn render_markdown(docstring: &str) -> String {
...
+            if include_colon {
+                line = line.strip_suffix(":").unwrap();
+            } else {
+                line = without_directive.trim_end();
             }
-            starting_literal = true;
+
+            starting_literal = match directive {
+                // Special directives that should be plaintext
+                #[allow(clippy::used_underscore_binding)]
```
MichaReiser: The comment suggests that it applies to the clippy allow, when it doesn't. 

I suggest using a different name for `_line`. I assumed the variable would be unused, when, in fact, it's a tricky work around lifetimes. Maybe `tmp_line` or `owned_line` or make line a `Cow`?

If you really want to use `_line`, use `expect`

crates/ty_ide/src/docstring.rs
```diff
@@ -224,16 +225,17 @@ fn render_markdown(docstring: &str) -> String {
...
-        if starting_literal && !line.is_empty() {
-            starting_literal = false;
+        if let Some(literal) = starting_literal
+            && !line.is_empty()
+        {
+            starting_literal = None;
             in_literal = true;
             in_any_code = true;
             block_indent = line_indent;
             // TODO: I hope people don't have literal blocks about markdown code fence syntax
```
MichaReiser: Doesn't markdown allow you to use an arbitrary number of backticks?

> A code fence is a sequence of at least three consecutive backtick characters (`) or tildes (~). (Tildes and backticks cannot be mixed.) A fenced code block begins with a code fence, preceded by up to three spaces of indentation.


So, we could just use a silly amount of backticks :) https://spec.commonmark.org/0.31.2/#fenced-code-blocks

crates/ty_ide/src/docstring.rs
```diff
@@ -956,19 +943,6 @@ mod tests {
...
-                print(x_y)
-            else:
-                print("too short :(")
-
-            print("done")
-        You love to see it.
-        "#);
-
         assert_snapshot!(docstring.render_markdown(), @r#"
         Check out this great example code:
```
MichaReiser, line 955: Why do we preseve the colon here but not in `literal_own_line`

crates/ty_ide/src/docstring.rs
Gankra: realtalk this was me getting Extremely flustered at the `_special_case_temp_owned_thing` idiom being flagged by clippy after 10 years of me using it. I have abandoned the underscore :(

crates/ty_ide/src/docstring.rs
Gankra: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

> The handling of the :: marker is smart:
> * If it occurs as a paragraph of its own, that paragraph is completely left out of the document.
> * If it is preceded by whitespace, the marker is removed.
> * If it is preceded by non-whitespace, the marker is replaced by a single colon.

When translating to markdown this amounts to "check if there's a space immediately before the `::`, if there is, delete it. Otherwise, turn it into `:`".

crates/ty_ide/src/docstring.rs
Gankra: This is genius.

PR #21666, opened 2025-11-27
Support nested projects within a single LSP session

MichaReiser:

### DIFF
--- a/crates/ty_project/src/metadata.rs
+++ b/crates/ty_project/src/metadata.rs
@@ -1,7 +1,9 @@
 use configuration_file::{ConfigurationFile, ConfigurationFileError};
+use ruff_db::system::walk_directory::WalkState;
 use ruff_db::system::{System, SystemPath, SystemPathBuf};
 use ruff_db::vendored::VendoredFileSystem;
 use ruff_python_ast::name::Name;
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use thiserror::Error;
 use ty_combine::Combine;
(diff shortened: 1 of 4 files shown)


MichaReiser:
One major concern with this approach is that memory usage is likely to grow significantly (~projects*files) because we parse and analyze the same file multiple time, especially because we start computing auto imports per database

Gankra:
I think we'll want to revive this for PEP-723 script support, but only register the members if the file is actually open. This way you'll only get really brutal memory usage if you have a bunch of scripts open *and* they have non-trivial dependencies (e.g. if you have a ton of scripts that import members of a monorepo, transitively importing the entire monorepo).

Notably I think we should skip scripts on workspace diagnostics by default. (...which now that I've said that, that will be annoying if your scripts import monorepo members and you're refactoring your APIs... actually christ how would `rename` even work there oh god).

MichaReiser:
> Notably I think we should skip scripts on workspace diagnostics by default. (...which now that I've said that, that will be annoying if your scripts import monorepo members and you're refactoring your APIs... actually christ how would rename even work there oh god).

Yeah, this is a good point. I'm getting more and more convinced that we need a single ty instance for all projects.

MichaReiser:
@BurntSushi after our discussions on Discord, I think we should do this but it isn't urgent. I would say completing and improving our existing LSP features is more important and I can take a stab at it when I'm back. But also feel free to pick it up if you run out of work or consider it more important

MichaReiser:
I don't think the overriding of options for nested projects is actually needed. If an outer project doesn't exclude the inner project, then the diagnostics will be reported twice, which is consistent to when you run ty on the CLI

MichaReiser:
I don't think we want to implement this with different dbs anymore. Instead, we should make use of ty's new capability to support different `Program`s in a single db.

PR #21667, opened 2025-11-27
[ty] Add IDE autofixes for two "Did you mean...?" suggestions

AlexWaygood:
## Summary

Add two autofixes. One will allow you to easily fix a typo in a `TypedDict` string-literal key; the other allows you to fix `x: datetime` to `x: datetime.datetime`.

Most of our other "Did you mean...?" suggestions are surprisingly hard to add fixes for with our current infrastructure. E.g., it would be great to add a fix that switches `raise NotImplemented` to `raise NotImplementedError`, but you first have to check whether the builtin symbol `NotImplementedError` has been shadowed anywhere in this module. It's doable, but I don't want to spend time building that infrastructure _right_ now :-)

## Test Plan

Snapshots added. And, manual testing in the playground:

https://github.com/user-attachments/assets/1b7f4c72-bed2-4508-9359-1248288bd81b

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -8910,8 +8910,6 @@ impl<'db> InvalidTypeExpression<'db> {
             {
                 return;
             }
-
-            // TODO: showing a diff (and even having an autofix) would be even better
             diagnostic.set_primary_message(format_args!(
                 "Did you mean to use the module's member \
                 `{module_name_final_part}.{module_name_final_part}`?"


crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Misspelled_key_for_`…_(7cf0fa634e2a2d59).snap
```diff
@@ -34,5 +34,11 @@ error[invalid-key]: Unknown key "Retries" for TypedDict `Config`
   |     TypedDict `Config`
   |
 info: rule `invalid-key` is enabled by default
+4 |     retries: int
+5 | 
+6 | def _(config: Config) -> None:
+  -     config["Retries"] = 30.0  # error: [invalid-key]
+7 +     config["retries"] = 30.0  # error: [invalid-key]
+note: This is an unsafe fix and may change runtime behavior
```
AlexWaygood: it feels a bit weird to me that the

```
info: rule `invalid-key` is enabled by default
```

note comes above the rendering of the fix. If I add a `help` subdiagnostic (which is currently the only way to give the quick-fix a title in the IDE), the `help` subdiagnostic comes before this note, so the `help` subdiagnostic doesn't look like a fix-title at all when it's rendered in a snapshot (or the CLI, though we don't offer fixes on the CLI yet).

crates/ty_python_semantic/src/types.rs
```diff
@@ -8910,6 +8916,10 @@ impl<'db> InvalidTypeExpression<'db> {
                 "Did you mean to use the module's member \
                 `{module_name_final_part}.{module_name_final_part}`?"
             ));
+            diagnostic.set_fix(Fix::unsafe_edit(Edit::insertion(
```
MichaReiser, line 8917: Remove the TODO on line 8914

crates/ty_test/src/lib.rs
```diff
@@ -643,7 +644,8 @@ fn create_diagnostic_snapshot(
 ) -> String {
     let display_config = DisplayDiagnosticConfig::default()
         .color(false)
-        .show_fix_diff(true);
+        .show_fix_diff(true)
+        .with_fix_applicability(Applicability::DisplayOnly);
```
MichaReiser, line 648: We should also enable this for the CLI or this will be another instance where we think this has always existed but never shipped to users, similar to showing fixes in Ruff 😆 

https://github.com/astral-sh/ruff/blob/df66946b8959589307e4278b0fd5c1d0b187e854/crates/ty/src/lib.rs#L314-L317

crates/ty_python_semantic/resources/mdtest/snapshots/assignment_diagnosti…_-_Subscript_assignment…_-_Misspelled_key_for_`…_(7cf0fa634e2a2d59).snap
MichaReiser: I agree. But that's rather tricky to fix because the diff rendering isn't really integrated into annotated-snippets. Instead, we first render the diagnostic with annotated-snippets, but the fix is rendered separately, only mimicking the annotated-snippets design.

crates/ty_python_semantic/src/types.rs
AlexWaygood: Whooops, thanks!

crates/ty_test/src/lib.rs
AlexWaygood: I opened https://github.com/astral-sh/ty/issues/1679

PR #21671, opened 2025-11-27
[ty] Implement patterns and typevars in the LSP

Gankra:
## Summary

**This is the final goto-targets with missing goto-definition/declaration implementations! 
You can now theoretically click on all the user-defined names in all the syntax. 🎉**

This adds:

* goto definition/declaration on patterns/typevars
* find-references/rename on patterns/typevars
* fixes syntax highlighting of `*rest` patterns

This notably *does not* add:

* goto-type for patterns/typevars 
* hover for patterns/typevars (because that's just goto-type for names)

Also I realized we were at the precipice of one of the great GotoTarget sins being resolved, and so I made import aliases also resolve to a ResolvedDefinition. This removes a ton of cruft and prevents further backsliding.

Note however that import aliases are, in general, completely jacked up when it comes to find-references/renames (both before and after this PR). Previously you could try to rename an import alias and it just wouldn't do anything. With this change we instead refuse to even let you try to rename it.

Sorting out why import aliases are jacked up is an ongoing thing I hope to handle in a followup.

## Test Plan

You'll surely not regret checking in 86 snapshot tests

### DIFF
--- a/crates/ty_ide/src/goto.rs
+++ b/crates/ty_ide/src/goto.rs
@@ -377,38 +377,32 @@ impl GotoTarget<'_> {
         model: &SemanticModel<'db>,
         alias_resolution: ImportAliasResolution,
     ) -> Option<Definitions<'db>> {
-        match self {
-            GotoTarget::Expression(expression) => {
-                definitions_for_expression(model, *expression).map(Definitions)
-            }
+        let definitions = match self {
+            GotoTarget::Expression(expression) => definitions_for_expression(model, *expression),
             // For already-defined symbols, they are their own definitions
-            GotoTarget::FunctionDef(function) => {
-                Some(Definitions(vec![ResolvedDefinition::Definition(
-                    function.definition(model),
-                )]))
-            }
+            GotoTarget::FunctionDef(function) => Some(vec![ResolvedDefinition::Definition(
+                function.definition(model),
+            )]),
 
-            GotoTarget::ClassDef(class) => Some(Definitions(vec![ResolvedDefinition::Definition(
+            GotoTarget::ClassDef(class) => Some(vec![ResolvedDefinition::Definition(
                 class.definition(model),
-            )])),
+            )]),
 
-            GotoTarget::Parameter(parameter) => {
-                Some(Definitions(vec![ResolvedDefinition::Definition(
-                    parameter.definition(model),
-                )]))
-            }
+            GotoTarget::Parameter(parameter) => Some(vec![ResolvedDefinition::Definition(
+                parameter.definition(model),
+            )]),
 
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1476,6 +1478,8 @@ impl<'ast> Visitor<'ast> for SemanticIndexBuilder<'_, 'ast> {
                     }
 
                     let (symbol_name, is_reexported) = if let Some(asname) = &alias.asname {
+                        self.scopes_by_expression
+                            .record_expression(asname, self.current_scope());
```
Gankra: This one is actually useless I think (I added it while thrashing and trying to get imports rename/reference working... it did not work).

crates/ty_ide/src/goto.rs
```diff
@@ -399,37 +376,39 @@ impl GotoTarget<'_> {
         &self,
         model: &SemanticModel<'db>,
         alias_resolution: ImportAliasResolution,
-    ) -> Option<DefinitionsOrTargets<'db>> {
-        use crate::NavigationTarget;
+    ) -> Option<Definitions<'db>> {
         match self {
```
MichaReiser: Nit: You could reduce the boilerplate here, I think, by doing something like:

```
let definitions = match self {
    GotoTarget::Expression(expression) => {
        definitions_for_expression(model, *expression)
    },
    GotoTarget::FunctionDef(function) => {
        Some(vec![ResolvedDefinition::Definition(
            function.definition(model),
        )])
    },
    ...
};

definitions.map(Definitions)
```

crates/ty_python_semantic/src/semantic_index/builder.rs
```diff
@@ -1064,6 +1064,8 @@ impl<'db, 'ast> SemanticIndexBuilder<'db, 'ast> {
                         ..
                     }) => (name, &None, default),
                 };
+                self.scopes_by_expression
+                    .record_expression(name, self.current_scope());
```
MichaReiser, line 1068: Should we add a `record_name` method or at least rename the struct (or update its documentation) to explain that, yes, it's mostly expressions but sometimes it's names too

crates/ty_python_semantic/src/semantic_index/builder.rs
MichaReiser: Should we remove it then? Seems easy enough to add once we know why we need it

crates/ty_python_semantic/src/semantic_index/builder.rs
Gankra: This is probably a good idea, I'll integrate it into some followup work I have planned that will involve more reckoning about ""expressions""

PR #21672, opened 2025-11-28
[ty] Implicit type aliases: Instantiation and attribute access

sharkdp:
## Summary

Allow `Annotated[MyClass, …]` instances to be instantiated.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
+++ b/crates/ty_python_semantic/resources/mdtest/implicit_type_aliases.md
@@ -11,16 +11,35 @@ valid type for use in a type expression:
 ```py
 MyInt = int
 
+reveal_type(MyInt)  # revealed: <class 'int'>
+reveal_type(MyInt(1))  # revealed: int
+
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Oh, the typing conformance suite results seem to suggest that we shouldn't do this: apparently the two lines where diagnostics are going away are places we're _supposed_ to emit diagnostics. But it does make sense to me to just emulate what happens at runtime; I don't know why that rule exists in the spec (https://typing.python.org/en/latest/spec/qualifiers.html#annotated)

sharkdp:
> I don't know why that rule exists in the spec ([typing.python.org/en/latest/spec/qualifiers.html#annotated](https://typing.python.org/en/latest/spec/qualifiers.html#annotated))

Oh. I only looked at the `typing.Annotated` documentation and forgot to check the typing spec. This one: "An attempt to call Annotated (whether parameterized or not) should be treated as a type error by type checkers"? That's strange, yes. Moving this to draft for now.

PR #21674, opened 2025-11-28
[ty] Add `db` parameter to `Parameters::new` method

dhruvmanila:
## Summary

This PR adds a new `db` parameter to `Parameters::new` for https://github.com/astral-sh/ruff/pull/21445. This change creates a large diff so thought to split it out as it's just a mechanical change.

The `Parameters::new` method not only creates the `Parameters` but also analyses the parameters to check what kind it is. For `ParamSpec` support, it's going to require the `db` to check whether the annotated type is `ParamSpec` or not. For the current set of parameters that isn't required because it's only checking whether it's dynamic or not which doesn't require `db`.


dhruvmanila:
This is a mechanical change, so I'm going to merge this without a review.

PR #21678, opened 2025-11-28
[ty] Fix subtyping of `type[Any]` / `type[T]` and protocols

sharkdp:
## Summary

This is a bugfix for subtyping of `type[Any]` / `type[T]` and protocols.

## Test Plan

Regression test that will only be really meaningful once https://github.com/astral-sh/ruff/pull/21553 lands.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/libraries/numpy.md
+++ b/crates/ty_python_semantic/resources/mdtest/libraries/numpy.md
@@ -1,0 +1,67 @@
+# numpy
+
+```toml
+[environment]
+python-version = "3.14"
+```
+
+## numpy's `dtype`
+
+numpy functions often accept a `dtype` parameter. For example, one of `np.array`'s overloads accepts
+a `dtype` parameter of type `DTypeLike | None`. Here, we build up something that resembles numpy's
+internals in order to model the type `DTypeLike`. Many details have been left out.
+
+`mini_numpy.py`:
+
+```py
+from typing import TypeVar, Generic, Any, Protocol, TypeAlias, runtime_checkable, final
+import builtins
+
+_ItemT_co = TypeVar("_ItemT_co", default=Any, covariant=True)
+
+class generic(Generic[_ItemT_co]):
+    @property
+    def dtype(self) -> _DTypeT_co:
+        raise NotImplementedError
+
+_BoolItemT_co = TypeVar("_BoolItemT_co", bound=builtins.bool, default=builtins.bool, covariant=True)
+
+class bool(generic[_BoolItemT_co], Generic[_BoolItemT_co]): ...
+
+@final
+class object_(generic): ...
+
+_ScalarT = TypeVar("_ScalarT", bound=generic)
+_ScalarT_co = TypeVar("_ScalarT_co", bound=generic, default=Any, covariant=True)
+
+@final
+class dtype(Generic[_ScalarT_co]): ...
+
+_DTypeT_co = TypeVar("_DTypeT_co", bound=dtype, default=dtype, covariant=True)
+
+@runtime_checkable
+class _SupportsDType(Protocol[_DTypeT_co]):
+    @property
+    def dtype(self) -> _DTypeT_co: ...
+
+# TODO: no errors here
+# error: [invalid-type-arguments] "Type `typing.TypeVar` is not assignable to upper bound `generic[Any]` of type variable `_ScalarT_co@dtype`"
+# error: [invalid-type-arguments] "Type `typing.TypeVar` is not assignable to upper bound `generic[Any]` of type variable `_ScalarT_co@dtype`"
+_DTypeLike: TypeAlias = type[_ScalarT] | dtype[_ScalarT] | _SupportsDType[dtype[_ScalarT]]
+
+DTypeLike: TypeAlias = _DTypeLike[Any] | str | None
+```
+
+Now we can make sure that a function which accepts `DTypeLike | None` works as expected:
+
+```py
+import mini_numpy as np
+
+def accepts_dtype(dtype: np.DTypeLike | None) -> None: ...
+
+accepts_dtype(dtype=np.bool)
+accepts_dtype(dtype=np.dtype[np.bool])
+accepts_dtype(dtype=object)
+accepts_dtype(dtype=np.object_)
+accepts_dtype(dtype="U")
+```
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/resources/mdtest/libraries/numpy.md
```diff
@@ -0,0 +1,67 @@
...
+_DTypeT_co = TypeVar("_DTypeT_co", bound=dtype, default=dtype, covariant=True)
+
+@runtime_checkable
+class _SupportsDType(Protocol[_DTypeT_co]):
+    @property
+    def dtype(self) -> _DTypeT_co: ...
+
+# TODO: no errors here
+# error: [invalid-type-arguments] "Type `typing.TypeVar` is not assignable to upper bound `generic[Any]` of type variable `_ScalarT_co@dtype`"
+# error: [invalid-type-arguments] "Type `typing.TypeVar` is not assignable to upper bound `generic[Any]` of type variable `_ScalarT_co@dtype`"
```
sharkdp, line 49: These errors prevent us from running into the problem on `main`. But on https://github.com/astral-sh/ruff/pull/21553, these errors will go away, and then we need to properly understand all of this code here, or otherwise we see `np.array` calls failing all over the place.

crates/ty_python_semantic/src/types.rs
```diff
@@ -2468,6 +2468,34 @@ impl<'db> Type<'db> {
                 })
             }
 
+            // TODO: The following two branches are a temporary workaround to prevent an important union type
+            // in numpy's internals from collapsing to a single type, which causes thousands of downstream
+            // false positive diagnostics across the ecosytem. See https://github.com/astral-sh/ty/issues/1666
+            // for details.
+            (Type::SubclassOf(self_subclass_ty), Type::ProtocolInstance(_))
+                if self_subclass_ty.is_type_var() =>
```
AlexWaygood: but `type[T]` should always be _assignable_ to a protocol-instance if `T`'s upper bound is assignable to the protocol-instance, right? 

```suggestion
            (Type::SubclassOf(self_subclass_ty), Type::ProtocolInstance(_))
                if self_subclass_ty.is_type_var() && relation.is_assignability() =>
```

crates/ty_python_semantic/src/types.rs
```diff
@@ -2468,6 +2468,34 @@ impl<'db> Type<'db> {
...
+                    _ => KnownClass::Type.to_instance(db).has_relation_to_impl(
+                        db,
+                        target,
+                        inferable,
+                        relation,
+                        relation_visitor,
+                        disjointness_visitor,
+                    ),
+                }
+            }
```
AlexWaygood, line 2489: I think the `type[Dynamic]` assignability case is already handled by the generalised member-lookup approach that `Type::satisfies_protocol()` does in the branch immediately below this one. What we really want to do in this branch is block subtyping for `type[Dynamic]`:

```suggestion
            // `type[Any]` is assignable to arbitrary protocols as it has arbitrary attributes
            // (this is handled by a lower-down branch), but it is only a subtype of
            // a given protocol if `type` is a subtype of that protocol
            (Type::SubclassOf(self_subclass_ty), Type::ProtocolInstance(_))
                if self_subclass_ty.is_dynamic() && !relation.is_assignability() =>
            {
                KnownClass::Type.to_instance(db).has_relation_to_impl(
                    db,
                    target,
                    inferable,
                    relation,
                    relation_visitor,
                    disjointness_visitor,
                ),
            }
```

crates/ty_python_semantic/src/types.rs
sharkdp: Do you mean: this branch should *return* `ConstraintSet::from(relation.is_assignability())`? Because if we guard it with `is_assignability`, the workaround would not be effective.

AlexWaygood: Ah, right, yeah -- I think I got the polarity the wrong way round?

```suggestion
            (Type::SubclassOf(self_subclass_ty), Type::ProtocolInstance(_))
                if self_subclass_ty.is_type_var() && !relation.is_assignability() =>
```

I don't think the body of the branch should be `ConstraintSet::from(relation.is_assignability())` -- if the relation is `TypeRelation::Assignability`, we only want the assignability check to hold for `type[T] <: P` if the meta-type of `T`'s upper bound is assignable to `P` (we want to fall through to a later branch that should handle that for us)

sharkdp: No, you probably meant `&& !relation.is_assignability()`

crates/ty_python_semantic/src/types.rs
```diff
@@ -2468,6 +2468,31 @@ impl<'db> Type<'db> {
...
+            {
+                KnownClass::Type.to_instance(db).has_relation_to_impl(
+                    db,
+                    target,
+                    inferable,
+                    relation,
+                    relation_visitor,
+                    disjointness_visitor,
+                )
+            }
```
AlexWaygood: and now that I look at it again, can these not be collapsed into a single branch? `type[T]` should always be a subtype of `P` if `type` is a subtype of `P`, since all `type[T]` branches will be a subtype of `type` no matter what the `T` type variable is solved to. So it's the exact same principle as with `type[Any]`.

```suggestion
            // TODO: The following two branches are a temporary workaround to prevent an important union type
            // in numpy's internals from collapsing to a single type, which causes thousands of downstream
            // false positive diagnostics across the ecosytem. See https://github.com/astral-sh/ty/issues/1666
            // for details.
            //
            // `type[Any]` is assignable to arbitrary protocols as it has arbitrary attributes
            // (this is handled by a lower-down branch), but it is only a subtype of
            // a given protocol if `type` is a subtype of that protocol. Similarly,
            // `type[T]` will always be assignable to any protocol if `type[<upper bound of T>]`
            // is assignable to that protocol (handled lower down), but it is only a
            // subtype of that protocol if `type` is a subtype of that protocol.
            (Type::SubclassOf(self_subclass_ty), Type::ProtocolInstance(_))
                if (self_subclass_ty.is_dynamic() || self_subclass_ty.is_type_var()) && !relation.is_assignability() =>
            {
                KnownClass::Type.to_instance(db).has_relation_to_impl(
                    db,
                    target,
                    inferable,
                    relation,
                    relation_visitor,
                    disjointness_visitor,
                )
            }
```

sharkdp: Thank you Alex!

PR #21682, opened 2025-11-28
[ty] Preserve quoting style when autofixing `TypedDict` keys

AlexWaygood:
## Summary

I've been away from Ruff for too long -- I forgot how complicated writing autofixes can be 🙃

This is a  minor followup to https://github.com/astral-sh/ruff/pull/21667 to make the `TypedDict` fix introduced in that PR a bit better. It now preserves the quoting style of the previous string-literal node.

## Test Plan

added a snapshot


AlexWaygood:
I'll add `internal` as this doesn't need to be included in the changelog separately from #21667

PR #21683, opened 2025-11-28
[ty] increase the limit on the number of elements in a non-recursively defined literal union

mtshiba:
## Summary

Closes https://github.com/astral-sh/ty/issues/957

As explained in https://github.com/astral-sh/ty/issues/957, literal union types for recursively defined values ​​can be widened early to speed up the convergence of fixed-point iterations.
This PR achieves this by embedding a marker in `UnionType` that distinguishes whether a value is recursively defined.

This also allows us to identify values ​​that are not recursively defined, so I've increased the limit on the number of elements in a literal union type for such values.

Edit: while this PR doesn't provide the significant performance improvement initially hoped for, it does have the benefit of allowing the number of elements in a literal union to be raised above the salsa limit, and indeed mypy_primer results revealed that a literal union of 220 elements was actually being used.

## Test Plan

`call/union.md` has been updated

### DIFF
--- a/crates/ty_python_semantic/src/types/builder.rs
+++ b/crates/ty_python_semantic/src/types/builder.rs
@@ -224,8 +224,8 @@ impl RecursivelyDefined {
 /// If the value ​​is defined recursively, widening is performed from fewer literal elements, resulting in faster convergence of the fixed-point iteration.
 const MAX_RECURSIVE_UNION_LITERALS: usize = 10;
 /// If the value ​​is defined non-recursively, the fixed-point iteration will converge in one go,
-/// so in principle we can have as many literal elements as we want, but to avoid unintended huge computational loads, we limit it to 190.
-const MAX_NON_RECURSIVE_UNION_LITERALS: usize = 190;
+/// so in principle we can have as many literal elements as we want, but to avoid unintended huge computational loads, we limit it to 256.
+const MAX_NON_RECURSIVE_UNION_LITERALS: usize = 256;
 
 pub(crate) struct UnionBuilder<'db> {
     elements: Vec<UnionElement<'db>>,


mtshiba:
Hmm, for example, when inspecting the following code, I see a clear performance improvement locally, but the change on codspeed seems less clear:

```python
class Counter:
    def __init__(self: "Counter"):
        self.count = 0

    def increment(self: "Counter"):
        self.count = self.count + 1

reveal_type(Counter().count)  # revealed: Unknown | int
```

main:
```sh
# release
$ hyperfine -i "ty check widen.py"
Benchmark 1: ty check widen.py
  Time (mean ± σ):     103.3 ms ±   8.6 ms    [User: 56.9 ms, System: 45.0 ms]
  Range (min … max):    96.4 ms … 136.6 ms    22 runs

# debug
$ hyperfine -i "ty check widen.py"
Benchmark 1: ty check widen.py
  Time (mean ± σ):     899.9 ms ±   6.2 ms    [User: 849.7 ms, System: 51.9 ms]
  Range (min … max):   889.4 ms … 910.3 ms    10 runs
```

#21683:
```sh
# release
$ hyperfine -i "ty check widen.py"
Benchmark 1: ty check widen.py
  Time (mean ± σ):      59.6 ms ±   9.2 ms    [User: 26.5 ms, System: 37.8 ms]
  Range (min … max):    52.5 ms …  95.7 ms    31 runs

# debug
$ hyperfine -i "ty check widen.py"
Benchmark 1: ty check widen.py
  Time (mean ± σ):     267.2 ms ±   4.1 ms    [User: 222.2 ms, System: 55.6 ms]
  Range (min … max):   261.6 ms … 276.5 ms    10 runs
```

mtshiba:
Conversely, the apparent performance degradation is due to increasing `MAX_NON_RECURSIVE_UNION_LITERALS`.
Even setting it to 512 did not result in a panic, which shows that the divergent cases are being correctly avoided.

The performance impact of changing `MAX_NON_RECURSIVE_UNION_LITERALS` is as follows:

190: https://codspeed.io/astral-sh/ruff/runs/compare/6929edd91e804897641888e6..6929fa9ed77b771deb1a9244
256: https://codspeed.io/astral-sh/ruff/runs/compare/6929edd91e804897641888e6..6929f4ded77b771deb1a9211
512: https://codspeed.io/astral-sh/ruff/runs/compare/6929edd91e804897641888e6..6929ef181e804897641888ec

MichaReiser:
> Hmm, for example, when inspecting the following code, I see a clear performance improvement locally, but the change on codspeed seems less clear:

There might just not be enough very large unions so that, in the grand picture, the improvement is lost in noise

mtshiba:
I think this is ready for review. Here are some comments:

The performance degradation of codspeed(many_string_assignments) is entirely expected.

There is no clear standard for what the new upper limit for literal unions should be, but I've set it to 256 for now.
mypy_primer revealed that literal unions of 220 elements are actually used, but said that of 512 elements is likely not present in practice.

crates/ty_python_semantic/resources/mdtest/call/union.md
```diff
@@ -227,17 +227,22 @@ def _(literals_2: Literal[0, 1], b: bool, flag: bool):
...
     literals_64 = 4 * literals_16 + literals_4  # Literal[0, 1, .., 63]
     literals_128 = 2 * literals_64 + literals_2  # Literal[0, 1, .., 127]
+    literals_256 = 2 * literals_128 + literals_2  # Literal[0, 1, .., 255]
 
-    # Going beyond the MAX_UNION_LITERALS limit (currently 200):
-    literals_256 = 16 * literals_16 + literals_16
-    reveal_type(literals_256)  # revealed: int
+    # Going beyond the MAX_UNION_LITERALS limit (currently 512):
+    literals_512 = 2 * literals_256 + literals_2  # Literal[0, 1, .., 511]
+    reveal_type(literals_512 if flag else 512)  # revealed: int
```
carljm, line 234: Why do we need to add the `if flag else 512` here? It wasn't in the previous version of the test, and it seems the test passes without it. Did some intermediate version of the code fail without this?

mtshiba: Oops, it seems this was written based on the old limit.
Opened #21804.

PR #21685, opened 2025-11-28
[ty] Infer type of implicit `cls` parameter in method bodies

ibraheemdev:
## Summary

Extends https://github.com/astral-sh/ruff/pull/20922 to infer unannotated `cls` parameters as `type[Self]` in method bodies.

Part of https://github.com/astral-sh/ty/issues/159.

### DIFF
--- a/crates/ruff_benchmark/benches/ty_walltime.rs
+++ b/crates/ruff_benchmark/benches/ty_walltime.rs
@@ -194,7 +194,7 @@ static SYMPY: Benchmark = Benchmark::new(
         max_dep_date: "2025-06-17",
         python_version: PythonVersion::PY312,
     },
-    13000,
+    13030,
 );
 
 static TANJUN: Benchmark = Benchmark::new(


AlexWaygood:
Here's a minimal repro of one of the new diagnostics on `artigraph` (looks like an underlying issue to do with `type[T]` rather than an issue with this PR -- it also reproduces on `main`):

```py
from typing import Callable, Any

def needs_a_callable(x: Callable[[Any], Any] | Callable[[Any, object], Any]): ...

class F:
    def __init__(self, arg) -> None: ...

def f[T: F](x: T):
    needs_a_callable(type(x))
```

It looks like we might not be implementing assignability correctly for `type[T]` vs a _union_ of callables?

AlexWaygood:
The artigraph diagnostic in `src/arti/views/__init__.py` looks like another problem caused by simplifying `T | <upper bound of T>` to `<upper bound of T>` in some situations (x-ref https://github.com/astral-sh/ty/issues/1666). So that one's a bigger issue that I don't think we should try to fix before landing this PR.

AlexWaygood:
The new attrs diagnostics are the same as https://github.com/astral-sh/ty/issues/1425, but for attrs rather than pydantic this time. We don't recognise the assignments in the class body of [`VersionInfo`](https://github.com/python-attrs/attrs/blob/43a868d703f06fffb724d5d03d4f80ac447cb27c/src/attr/_version_info.py#L12) as defining dataclass fields, because the stdlib `dataclasses` module only creates fields for _annotated_ assignments, and the assignments in `VersionInfo` are all unannotated.

AlexWaygood:
comtypes looks like it's doing some highly dynamic stuff that I don't think we should be expected to understand right now.

The new diagnostics in `homeassistant/components/recorder/db_schema.py` already all have `type: ignore` comments for mypy, so I don't think we need to worry about them (though we emit these on a different line to mypy, which is maybe slightly unfortunate).

The new diagnostics in `homeassistant/components/yeelight/scanner.py` are https://github.com/astral-sh/ty/issues/1124 -- but they're also using a generic `ClassVar`, which we should probably ban anyway 😄. Anyway, no need to worry about them for this PR.

AlexWaygood:
Here's a minimal repro for the diagnostic on `ddtrace/internal/symbol_db/symbols.py`. I actually can't repro this one on `main`, so it may be a problem with this PR?

```py
from typing import Iterable, Iterator, Self

class chain[T]:
    def __new__(cls, *its: Iterable[T]): ...

    def __iter__(self) -> Self:
        return self

    def __next__(self) -> T:
        raise NotImplementedError

class Symbol:
    @classmethod
    def g(cls) -> list[Symbol]:
        return list(chain((cls(),), (cls(),)))  # error here...
                                                # error [invalid-argument-type] "Argument to bound method `__init__` is incorrect: Expected `Iterable[Symbol]`, found `chain[Self@from_code]`"

def f[T: Symbol](x: type[T]) -> list[Symbol]:
    return list(chain((x(),), (x(),)))  # ... but no error here?
```

ibraheemdev:
That last one may be an issue with how we treat `Self`, I can reproduce it without `type[T]` as well:
```py
class Symbol:
    @classmethod
    def f(cls) -> list[Symbol]:
        y = chain((cls(),), (cls(),))
        reveal_type(y)  # chain[Self@g]
        return list(y)  # error[invalid-argument-type]

    def g(self) -> list[Symbol]:
        y = chain((self,), (self,))
        reveal_type(y)  # chain[Self@g]
        return list(y)  # error[invalid-argument-type]

def f[T: Symbol](x: type[T]) -> list[Symbol]:
    y = chain((x(),), (x(),))
    reveal_type(y)  # chain[T@g]
    return list(y)  # ok

def g[T: Symbol](x: T) -> list[Symbol]:
    y = chain((x,), (x,))
    reveal_type(y)  # chain[T@g]
    return list(y)  # ok
```

sharkdp (approved):
Thank you!

PR #21686, opened 2025-11-29
[ty] replace `FxHash{Map, Set}` with footgun-mitigated APIs

mtshiba:
## Summary

from https://github.com/astral-sh/ty/issues/1670

~~If a data structure that depends on salsa IDs is used as the key for an `FxHashMap` or the value of an `FxHashSet`, the output order of the iterator will be unstable. If a query depends on this order, the results of fixed-point iteration will be unstable.
In this case, `Fx{Index, Order}{Map, Set}` should be used, but it seems that this is not being done thoroughly.~~

~~Simply replacing all `FxHash{Map, Set}` with `Fx{Index, Order}{Map, Set}` would solve the problem, but it is generally believed that the former has slightly better performance if we are only using insertion and retrieval without using a set/map as an iterator.~~

~~Therefore, this PR proposes a compromise.~~
That is, replace all `FxHash{Map, Set}` used within ty with wrapper structs that does not implement `(Into)Iterator`, and instead define methods like `unstable_iter` to make users of these structs aware of whether iteration operations are safe.

After performing this refactoring, I discovered some suspicious parts. I hope this fix will help to resolve the issue.


crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -104,13 +103,13 @@ pub(crate) fn all_declarations_and_bindings<'db>(
 }
 
 struct AllMembers<'db> {
-    members: FxHashSet<Member<'db>>,
+    members: FxIndexSet<Member<'db>>,
```
mtshiba: It's probably better to use `FxIndexSet` here.

crates/ty_python_semantic/src/types/overrides.rs
```diff
@@ -32,7 +31,7 @@ pub(super) fn check_class<'db>(context: &InferContext<'db, '_>, class: ClassLite
     }
 
     let class_specialized = class.identity_specialization(db);
-    let own_class_members: FxHashSet<_> =
+    let own_class_members: FxIndexSet<_> =
```
mtshiba: We should probably use `FxIndexSet` here.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -1037,7 +1036,8 @@ impl<'db> Node<'db> {
             Node::Interior(_) => {}
         }
 
-        let mut typevars = FxHashSet::default();
+        // We should use `FxIndexSet` here since `BoundTypeVarInstance::{valid, required}_specializations` is query-dependent.
+        let mut typevars = FxIndexSet::default();
```
mtshiba: We should probably use `FxIndexSet` here.
The rest of this module seems okay to use unstable iterators, unless I'm overlooking something.

crates/ty_project/src/files.rs
```diff
@@ -152,7 +152,7 @@ pub struct Indexed<'db> {
 
 #[derive(Debug, get_size2::GetSize)]
 struct IndexedInner {
-    files: FxHashSet<File>,
+    files: FxIndexSet<File>,
```
mtshiba: We should probably use `FxIndexSet` here.

crates/ty_python_semantic/src/types/generics.rs
```diff
@@ -121,7 +120,7 @@ pub(crate) fn typing_self<'db>(
 #[derive(Clone, Copy, Debug)]
 pub(crate) enum InferableTypeVars<'a, 'db> {
     None,
-    One(&'a FxHashSet<BoundTypeVarIdentity<'db>>),
+    One(&'a FxIndexSet<BoundTypeVarIdentity<'db>>),
```
mtshiba: We should probably use `FxIndexSet` here.

mtshiba:
Unfortunately, it appears that non-determinism remains due to factors entirely different from those considered in this PR.
However, that doesn't mean I think these changes are useless (they appear to improve performance slightly in benchmarks of large projects).

Anyway, I'm marking this PR as ready for review.

MichaReiser:
I like the approach, but I don't think using an IndexMap is the solution. 

Iterating over two `HashMap`s, each created by inserting the same elements in the same order, will yield the same iteration order when run on the same platform. 

The issue we see with fix point is that it's no longer guaranteed that the elements are inserted in the same order. However, we'll have the exact same issue when using `IndexMap` because `IndexMap`'s iteration order is defined by insertion order, which isn't guaranteed to be deterministic within a cyclic query. 

I'm not sure what the solution here is, other than applying some sort of sorting (somewhere?).

MichaReiser:
There's also one flaky diagnostic. What I suspect is that our convergence functions are now sensitive to which query is the outer-most cycle or some query that bails early as soon as it sees the first `Divergent` type (any type), and the any type is only visible depending on the cycle nesting

crates/ruff_benchmark/benches/ty.rs
```diff
@@ -89,7 +88,7 @@ fn setup_tomllib_case() -> Case {
     });
 
     let mut db = ProjectDatabase::new(metadata, system).unwrap();
-    let mut tomllib_files = FxHashSet::default();
+    let mut tomllib_files = FxIndexSet::default();
```
MichaReiser: I don't think we should make this change. The check result should never depend on iteration order because reading a directory doesn't guarantee that the files are listed in a deterministic ordering. 

If we rely on any ordering, then we'd have to sort the files manually

crates/ty_project/src/db/changes.rs
```diff
@@ -317,7 +317,9 @@ impl ProjectDatabase {
             }
         }
 
-        let diagnostics = if let Some(walker) = ProjectFilesWalker::incremental(self, added_paths) {
+        let diagnostics = if let Some(walker) =
+            ProjectFilesWalker::incremental(self, added_paths.stable_iter())
```
MichaReiser: Same here, file watching behavior should not depend on iteration order.

crates/ty_project/src/files.rs
MichaReiser: I don't think we should because there's nothing guaranteeing that we insert files in a deterministic order. See my other comments.

crates/ty_project/src/lib.rs
```diff
@@ -56,7 +55,7 @@ pub struct Project {
     /// The files that are open in the project, [`None`] if there are no open files.
     #[returns(ref)]
     #[default]
-    open_fileset: FxHashSet<File>,
+    open_fileset: FxIndexSet<File>,
```
MichaReiser: Same here, I don't think preserving insertion order gives us anything here because we don't control insertion order. Therefore, it's important that our check result doesn't depend on in which iteration individual files are checked.

crates/ty_python_semantic/src/module_resolver/typeshed.rs
```diff
@@ -222,7 +222,7 @@ impl FromStr for TypeshedVersions {
 
 impl fmt::Display for TypeshedVersions {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let sorted_items: BTreeMap<&ModuleName, &PyVersionRange> = self.0.iter().collect();
+        let sorted_items: BTreeMap<&ModuleName, &PyVersionRange> = self.0.stable_iter().collect();
```
MichaReiser: Given that we sort, using `stable_iter` here doesn't seem important

crates/ty_python_semantic/src/semantic_index/place.rs
```diff
@@ -139,6 +139,8 @@ pub enum ScopedPlaceId {
     Member(ScopedMemberId),
 }
 
+impl crate::StableKey for ScopedPlaceId {}
```
MichaReiser: Implementing the type on the key seems problematic to me, because it isn't the key that ensures iteration order is stable. In fact, the key doesn't matter at all. 

* Iterating over a regular `HashMap` should always be considered non-consistent (it's stable but not platform agnostic). Ideally, ty doesn't depend on any platform specific ordering
* For the iteration order to be stable, it is required that all items are inserted in the exact same ordering. This isn't something the key controls. Instead, it's something that depends on how the map is built and why all maps in `SemanticIndex` are stable (because we build them in a non-cyclic query)


I like "removing" `into_iter` and `iter` from `HashSet` and `HashMap` and only exposing a `iter_unstable` and `into_iter_unstable` methods to "warn" about the possiblity that the iteration order isn't guaranteed. But I don't think we should add a `StableKey` trait as this gives a false sense of "safety". 

But I do get your idea. What you want to express is: We know that this map is built in a deterministic way, it's safe to iterate it and assume stable ordering. Unfortunately, we can't do this without using a stable hash function, which `FxHash` isn't.

crates/ty_python_semantic/src/lib.rs
```diff
@@ -56,10 +56,333 @@ mod diagnostic;
...
-type FxIndexSet<V> = indexmap::IndexSet<V, BuildHasherDefault<FxHasher>>;
+pub type FxOrderMap<K, V> = ordermap::map::OrderMap<K, V, BuildHasherDefault<FxHasher>>;
+pub type FxOrderSet<V> = ordermap::set::OrderSet<V, BuildHasherDefault<FxHasher>>;
+pub type FxIndexMap<K, V> = indexmap::IndexMap<K, V, BuildHasherDefault<FxHasher>>;
+pub type FxIndexSet<V> = indexmap::IndexSet<V, BuildHasherDefault<FxHasher>>;
+
+/// Types that implement this trait are stable as keys of `FxHash{Map, Set}`, and maps and sets can be safely used as iterators.
+pub trait StableKey {}
+
+impl StableKey for String {}
```
MichaReiser: Let's move this to its own module (`hash.rs`)

crates/ty_python_semantic/src/lib.rs
```diff
@@ -56,10 +56,333 @@ mod diagnostic;
...
+        );
+    }
+}
+
+impl<V: Ord> FxHashSet<V> {
+    pub fn sorted_ref_vec(&self) -> Vec<&V> {
+        let mut vec: Vec<&V> = self.0.iter().collect();
+        vec.sort();
+        vec
+    }
```
MichaReiser: If this is commonly required, then using a `BTreeSet` is probably better

crates/ty_python_semantic/src/lib.rs
```diff
@@ -56,10 +56,333 @@ mod diagnostic;
...
+    )]
+    pub fn keys(&self) -> std::collections::hash_map::Keys<'_, K, V> {
+        panic!(
+            "FxHashMap does not guarantee stable iteration order; use FxIndexMap or unstable_keys() instead"
+        );
+    }
+
+    #[track_caller]
+    #[deprecated(
+        note = "FxHashMap does not guarantee stable iteration order; use FxIndexMap or unstable_values() instead"
```
MichaReiser: What's the motivation for offering these methods if they always panic at runtime. It seems preferred to not have those methods in the first place (or name then `keys_unstable` etc.)

Gankra:
re: randomizing layouts: if you want to go down that route I suggest cribbing from [rustc's `-Zrandomize-layout`](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/randomize-layout.html) which lets you pass the seed in as a CLI argument (and print the seed any time one is selected so people can try to reproduce an issue).

crates/ty_python_semantic/src/semantic_index/place.rs
mtshiba: You're right. I considered `FxHashMap` for the same keys as "stable" in a narrow sense, but it certainly doesn't guarantee that it will give a consistent order in all environments and versions. I'll remove `StableKey` APIs.

crates/ty_python_semantic/src/lib.rs
mtshiba: `crate::FxHashMap` implements `Deref<Target=rustc_hash::FxHashMap>`. I wanted to avoid the hassle of wrapping all the APIs.
But then `iter`s would also be available via `Deref`, so I decided to override them.

MichaReiser: I'd be fine copying over the signatures from `FxHashMap` that we need, similar to what rustc does. I suspect that it won't be that many (mainly `get`, `entry` etc)

crates/ruff_benchmark/benches/ty.rs
```diff
@@ -2,12 +2,12 @@
 use ruff_benchmark::criterion;
 use ruff_benchmark::real_world_projects::{InstalledProject, RealWorldProject};
 
+use std::collections::BTreeSet;
```
MichaReiser: Let's not use `BTreeSet` for indexed files. It won't give us any guarantees because the deterministic ordering goes out of the window in `project.check`, where we process the files concurrently.

crates/ty_project/src/db/changes.rs
```diff
@@ -52,7 +52,7 @@ impl ProjectDatabase {
             custom_stdlib_changed: false,
         };
         // Paths that were added
-        let mut added_paths = FxHashSet::default();
+        let mut added_paths = BTreeSet::default();
```
MichaReiser: Same here. I don't see why a stable ordering would be necessary here. We sort the returned diagnostics

crates/ty_project/src/files.rs
```diff
@@ -127,7 +127,7 @@ impl<'db> LazyFiles<'db> {
     /// Sets the indexed files of a package to `files`.
     pub(super) fn set(
         mut self,
-        files: FxHashSet<File>,
+        files: BTreeSet<File>,
```
MichaReiser: Let's revert this change for the same reason as stated above

crates/ty_python_semantic/src/hash.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+}
+
+impl<V> std::ops::Deref for FxHashSet<V> {
+    type Target = rustc_hash::FxHashSet<V>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
MichaReiser: Why is it okay to `Deref` `FxHashSet` but not `FxHashMap`?

crates/ty_python_semantic/src/lint.rs
```diff
@@ -344,9 +344,9 @@ impl LintRegistryBuilder {
             ),
         };
 
+        let entry = LintEntry::Alias(*target);
         assert_eq!(
-            self.by_name
-                .insert(from.as_str(), LintEntry::Alias(*target)),
+            self.by_name.insert(from.as_str(), entry),
```
MichaReiser: What's the reason for this change?

crates/ty_python_semantic/src/types/constraints.rs
MichaReiser: Can you say more what `query-dependent` means? 

Looking at the loop below, it doesn't seem to depend on ordering as it returns `true` only if all typevars satisfy the constraints and there's no state between the type var checking, as far as I can tell

crates/ty_python_semantic/src/types/ide_support.rs
```diff
@@ -516,7 +515,7 @@ impl<'db> PartialOrd for Member<'db> {
 
 /// List all members of a given type: anything that would be valid when accessed
 /// as an attribute on an object of the given type.
-pub fn all_members<'db>(db: &'db dyn Db, ty: Type<'db>) -> FxHashSet<Member<'db>> {
+pub fn all_members<'db>(db: &'db dyn Db, ty: Type<'db>) -> FxIndexSet<Member<'db>> {
```
MichaReiser: Are we using this anywhere outside the LSP?

crates/ty_python_semantic/src/hash.rs
mtshiba: I simply forgot to revert. There is no longer any need to implement `Deref` for both `FxHash{Set, Map}`.

crates/ty_python_semantic/src/types/ide_support.rs
mtshiba: It seems to be used in ty_python_semantic as well (https://github.com/mtshiba/ruff/blob/stable-iteration/crates/ty_python_semantic/src/types/function.rs#L1464-L1467). In that case, however, an unstable iterator is not a problem.

crates/ty_python_semantic/src/lint.rs
mtshiba: Using `Deref` prevented [two-phase borrows](https://rustc-dev-guide.rust-lang.org/borrow_check/two_phase_borrows.html) from working properly. Now that we're no longer using `Deref`.

crates/ty_python_semantic/src/types/constraints.rs
mtshiba: `{valid, required}_specialization` uses `lazy_{bound, constraints}` internally. I meant that the order in which these queries are called is non-deterministic.

crates/ty_python_semantic/src/types/ide_support.rs
AlexWaygood: that is just for internal use so that we can test the function itself in our mdtests; it's not a public-facing part of ty_python_semantic

mtshiba:
> I think I'm leaning towards changing the hash maps in separate PRs and more explicitly talk about why the changes are necessary. I'm not convinced that it's necessary to use `FxIndexMap` in many cases.

I changed this PR to just replace `FxHash{Map, Set}` with new APIs. Therefore, this PR does not change any of the behavior of ty. Changes that affect behavior will be made after this PR (with reconsideration).

MichaReiser (commented):
Thanks for updating.

I'd be curious to get some more opinions on this. To me, this PR goes from one extreme to the other. I don't think we need to enforce this new API in all crates. E.g. I think it's completely fine to use the normal `FxHashMap` and `IntoIterator` in the project file discovery. 

I'm leaning towards:

* Giving those types a different name instead of overriding `FxHashSet`. I'm not entirely sure what to name them
* Update our Contribution guidelines to state that these types should be used in `ty_python_semantic` with an explanation why it's important
* Add the types to `ruff_db`

But maybe it's just me who dislikes the extra verbosity and others actually prefer to override `FxHashSet` and `FxHashMap`

mtshiba:
The purpose of this PR is to make developers aware that using `FxHash{Map, Set}` as an iterator is inherently unstable, and in a sense, it intentionally introduces noisy APIs.
However, if you feel it's troublesome to continue using `unstable_iter()` in places where it is clearly OK to use it as an iterator, one option is to give `FxHash{Map, Set}` a tag type and change its behavior depending on the tag. In this case, there is no need to prepare two types of structs.

```rust
use std::collections::HashMap;
use std::collections::hash_map::IntoIter;

/// You are sure you can always use this hash map/set as an iterator.
#[derive(Default)]
pub struct ImplIntoIterator;
#[derive(Default)]
pub struct NoIntoIterator;

#[derive(Default)]
pub struct FxHashMap<K, V, Tag=NoIntoIterator>(HashMap<K, V>, std::marker::PhantomData<Tag>);

impl<K, V> IntoIterator for FxHashMap<K, V, ImplIntoIterator> {
    type Item = (K, V);
    type IntoIter = IntoIter<K, V>;

    fn into_iter(self) -> IntoIter<K, V> {
        self.0.into_iter()
    }
}

fn main() {
    let map1: FxHashMap<u32, u32> = FxHashMap::default();
    let map2: FxHashMap<u32, u32, ImplIntoIterator> = FxHashMap::default();
    
    for _i in map1 {} // ERR
    for _i in map2 {} // OK
}
```

Using `FxHashMap<K, V, ImplIntoIterator>` means that you are declaring that it can always be used as an iterator and there is no problem with that, and using `FxHashMap<K, V, NoIntoIterator>` means that you need to make sure it is OK to use before each iteration (if it is safe you can use `unstable_iter()`).

AlexWaygood:
I'm weakly in favour of this PR.

Points I see in favour:
- We've run into multiple bugs now due to iterating over `HashSet`s when we should be using `IndexSet`s.
- It's been nontrivial to track down those bugs when we've hit them. This PR should make it much less likely for us to introduce them in the first place, since it forces us to think explicitly about whether iterating over a `HashSet` is safe in that specific location. It also makes it easier to grep for places where we're iterating over sets, which makes it easier to figure out where the bug might be when we _do_ come across one of these bugs.
- IIUC, the new abstraction should be zero-cost when it comes to performance
- The new code is all pretty well isolated in its own submodule

Points I see against:
- It does seem like a somewhat elaborate solution to the problem. It's a fair amount of new code.
- Adding APIs to the wrapper structs as and when we need them feels like it could be tedious, and it might confuse third-party contributors when they find `HashSet` APIs aren't exposed on these wrappers
- I think iterating over `HashSet`s is probably fine outside of `ty_python_semantic`? It's when the hash depends on the hash of a `Type` that things get problematic, because that changes unpredictably depending on how many `Type`s Salsa interns in which places and in which order.
- I don't think our biggest source of nondeterminism right now is the fact that we sometimes iterate over `HashSet`s, it's issues resulting from our cycle normalization

carljm:
I don't have strong feelings about this change in any direction. It seems like a lot of added code/machinery to provide compile-time protection from a problem that hasn't really ever caused us that much difficulty. Unstable iterators have only caused non-determinism a couple times that I can recall, and it hasn't been that hard to track down. The real difficult non-determinism comes from reliance on Salsa IDs, which this won't help with.

I am also concerned with the ongoing overhead of contributors being confused that methods on upstream hash-sets don't exist on our hashsets, and have to be added if we haven't used them before.

mtshiba:
The original purpose of this PR was to comprehensively identify and fix uses of unstable iterators that negatively impact deterministic results.
The refactoring itself is a side effect, and if you all don't feel a strong need to accept it, applying the YAGNI principle is probably a better choice.

PR #21695, opened 2025-11-29
[ty] Move `all_members`, and related types/routines, out of `ide_support.rs`

AlexWaygood:
## Summary

These routines and types are now used by several modules elsewhere in `ty_python_semantic` that have little to do with supporting IDE usecases. @carljm asked in https://github.com/astral-sh/ruff/pull/21436#discussion_r2539331295 for these to therefore be moved out of this module and put somewhere else. I said I'd do that as a followup, so here's the followup!

## Test Plan

Existing tests all pass

### DIFF
--- a/crates/ty_python_semantic/src/types/list_members.rs
+++ b/crates/ty_python_semantic/src/types/list_members.rs
@@ -24,7 +24,7 @@ use crate::{
 };
 
 /// Iterate over all declarations and bindings in the given scope.
-pub(crate) fn all_declarations_and_bindings<'db>(
+pub(crate) fn all_members_of_scope<'db>(
     db: &'db dyn Db,
     scope_id: ScopeId<'db>,
 ) -> impl Iterator<Item = MemberWithDefinition<'db>> + 'db {
@@ -365,7 +365,7 @@ impl<'db> AllMembers<'db> {
             .map(|class| class.class_literal(db).0)
         {
             let parent_scope = parent.body_scope(db);
-            for memberdef in all_declarations_and_bindings(db, parent_scope) {
+            for memberdef in all_members_of_scope(db, parent_scope) {
                 let result = ty.member(db, memberdef.member.name.as_str());
                 let Some(ty) = result.place.ignore_possibly_undefined() else {
                     continue;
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/src/types/list_members.rs
```diff
@@ -0,0 +1,508 @@
...
+        attribute_scopes, definition::Definition, global_scope, place_table, scope::ScopeId,
+        semantic_index, use_def_map,
+    },
+    types::{
+        ClassBase, ClassLiteral, KnownClass, KnownInstanceType, SubclassOfInner, Type,
+        TypeVarBoundOrConstraints, class::CodeGeneratorKind, generics::Specialization,
+    },
+};
+
+pub(crate) fn all_declarations_and_bindings<'db>(
```
MichaReiser: This function seems unrelated to `members`. We also use it to compute the completions for the definitions in the current scope. That's why I think `all_members.rs` isn't a great fit for it. 


I think I'd move it into `semantic_index.rs` alongside `attribute_declarations`

AlexWaygood: This function returns an Iterator of `MemberWithDefinition`, which is a type defined in this module and which is intended to be a much higher-level API than what the semantic index has to offer. So to me, it feels very much like it belongs in this module — but maybe we should rename the module? `list_members.rs` rather than `all_members.rs`?

MichaReiser: I think it depends on the terminology we want here. We should either:

* Rename the method to `all_members`
* Rename `MemberWithDefinition` to something else, making it clear it's not just class members

What's unclear to me is if `Member` should be reserved to class members only (in which case we should rename the return type) or if we're fine using it for all sort of members (in which case we should rename the method)

AlexWaygood: I've renamed the module and updated some docs. LMK what you think.

> What's unclear to me is if `Member` should be reserved to class members only (in which case we should rename the return type) or if we're fine using it for all sort of members (in which case we should rename the method)

We already use the term "member" across the codebase much more broadly than just to refer to "class members only" -- for example, the `Type::member()` method, which can be called on any type (not just class types or instance types). And `all_members()` can also be called on any type, not just a class or an instance.

MichaReiser: Should we call this `all_member_definitions`?

MichaReiser (approved):

PR #21696, opened 2025-11-29
[ty] Add `--add-ignore` CLI option

MichaReiser:
## Summary

This PR adds a new `--add-ignore` CLI option that makes ty add `ty: ignore[codes]` comments to suppress any lint diagnostics (or add the codes to existing comments).

There's quite a lot going on in this PR:

* It adds the logic to suppress all diagnostics in a single file (trickier than I first assumed)
* It extends the CLI main loop with a second mode that, instead of just printing the diagnostics, suppresses all rule diagnostics 
* It implements the logic to suppress all rule diagnostics in a project by grouping diagnostics by file, adding the suppressions, checking the file, and writing it back to disk.
* It introduces a mechanism to override the result of `source_text` so that it uses the fixed source instead of the source on disk. This is required to check whether there are any new parse diagnostics after fixing the file (and computing the remaining non-rule diagnostics). This is a case where Salsa's speculative execution would be very handy :)
* Extended `WritableSystem` to write arbitrary byte arrays (necessary for notebooks)

This PR also lays the foundation for a `--fix` option. In fact, most of it has been written so that applying arbitrary fixes should be relatively straightforward. The main wringkle of `--fix` is that ty needs to call `check_file` multiple times and apply all fixes until there are no new fixes left. 

Closes https://github.com/astral-sh/ty/issues/913

## Test Plan

I tested this new feature on a few larger projects:

* pytorch
* homeassitant
* airflow

I verified that there are:

* no new syntax errors after running `--add-ignore`
* no fatal errors
* Running the command twice doesn't add any new suppressions on the second run
* Running `check` after returns no new diagnostics.

### DIFF
--- a/crates/ruff_db/src/source.rs
+++ b/crates/ruff_db/src/source.rs
@@ -1,10 +1,11 @@
+use std::borrow::Cow;
+use std::ops::Deref;
+use std::sync::Arc;
+
 use ruff_diagnostics::SourceMap;
 use ruff_notebook::Notebook;
 use ruff_python_ast::PySourceType;
 use ruff_source_file::LineIndex;
-use std::borrow::Cow;
-use std::ops::Deref;
-use std::sync::Arc;
 
 use crate::Db;
 use crate::files::{File, FilePath};


MichaReiser:
I'm planning to pick this up again. This should now be easier with https://github.com/astral-sh/ruff/pull/22083

MichaReiser:
Okay, I'm close but there's one more challenge here and that is that we need to recompute the file's diagnostics after adding the suppressions or all offsets will be off (e.g. for revealed-type)

crates/ty/src/args.rs
```diff
@@ -54,6 +54,9 @@ pub(crate) struct CheckCommand {
     )]
     pub paths: Vec<SystemPathBuf>,
 
+    #[arg(long)]
+    pub(crate) add_ignore: bool,
```
MichaReiser, line 59: TODO: Documentation

crates/ruff_db/src/files.rs
```diff
@@ -323,6 +323,11 @@ pub struct File {
     /// the file has been deleted is to change the status to `Deleted`.
     #[default]
     status: FileStatus,
+
+    /// Allows overriding the source of a file.
+    #[default]
+    #[returns(as_deref)]
+    pub source_override: Option<Box<str>>,
```
MichaReiser: Make this take a `SourceText` and move the `override` handling into the `source_text` query (lift it one layer up)

crates/ty_project/src/fixes.rs
```diff
@@ -0,0 +1,794 @@
...
+    Io(#[from] std::io::Error),
+
+    #[error("the file has been modified")]
+    FileWasModified,
+}
+
+/// Apply a series of fixes to `File` and returns the updated source code along with the source map.
+///
+/// Returns an error if not all fixes were applied because some fixes are overlapping.
+fn apply_fixes(source: &str, mut fixes: Vec<Fix>) -> Result<FixedCode, FixedCode> {
```
MichaReiser, line 270: This is a copy of the same function in Ruff

crates/ty_project/src/fixes.rs
```diff
@@ -0,0 +1,794 @@
...
+
+        // If we got here then we've been successful. Re-check to get the diagnostics with the
+        // update source, update the fix count.
+
+        if fixable_diagnostics.len() == file_diagnostics.len() {
+            file_diagnostics.clear();
+        } else {
+            // If there are any other file level diagnostics, call `check_file` to re-compute them
+            // with updated ranges.
+            let diagnostics = project.check_file(db, file);
```
MichaReiser, line 216: We'll always need to call `check_file` when implementing `--fixes`. It will be interesting how slow that will be

crates/ruff_db/src/source.rs
```diff
@@ -1,9 +1,10 @@
-use std::ops::Deref;
-use std::sync::Arc;
-
+use ruff_diagnostics::SourceMap;
 use ruff_notebook::Notebook;
 use ruff_python_ast::PySourceType;
 use ruff_source_file::LineIndex;
+use std::borrow::Cow;
+use std::ops::Deref;
+use std::sync::Arc;
```
charliermarsh: (Nit: these moved but didn't need to.)

crates/ty_project/src/fixes.rs
charliermarsh: Isn't this somewhat uncommon, though? My understanding (likely wrong) is that these would only be... unused ignore comments, and syntax errors (or other non-suppressible diagnostics)?

crates/ruff_db/src/diagnostic/mod.rs
```diff
@@ -98,6 +98,44 @@ impl Diagnostic {
         diag
     }
 
+    /// Adds sub diagnostics that tell the user that this is a bug in ty
+    /// and asks them to open an issue on GitHub.
+    pub fn add_bug_sub_diagnostics(&mut self, url_encoded_title: &str) {
```
charliermarsh, line 103: As an API, I feel like it would make more sense if this took the un-encoded title, and the function URL-encoded it for you. But I assume that means we'd need to depend on a URL-encoding implementation that we can otherwise omit.

MichaReiser: I agree in terms of API design, but it felt overkill to introduce a new URL encoding dependency for the few places where we call this function

crates/ty_project/src/fixes.rs
MichaReiser: Today yes. These are diagnostics like `reveal_type`, `IO` errors (unlikely we ever make it here). However, this will be much more common once we have a `--fix` mode because most diagnostics don't have a fix. At the other hand, it might not be an issue because there are much fewer files that we change. So it might be fine after all.

PR #21697, opened 2025-11-29
[ty] Emit `invalid-named-tuple` on namedtuple classes that have field names starting with underscores

AlexWaygood:
## Summary

This raises `TypeError` at runtime and is caught by all other major type checkers

## Test Plan

snapshot/mdtest

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/named_tuple.md
+++ b/crates/ty_python_semantic/resources/mdtest/named_tuple.md
@@ -482,3 +482,21 @@ class F(NamedTuple):
 
 super(F, F(42))  # fine
 ```
+
+## NamedTuples cannot have field names starting with underscores
+
+<!-- snapshot-diagnostics -->
+
+```py
+from typing import NamedTuple
+
+class Foo(NamedTuple):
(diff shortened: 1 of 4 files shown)


AlexWaygood:
The new typing-conformance diagnostic is a true positive.

AlexWaygood:
The changes reported by mypy_primer are all unrelated (https://github.com/astral-sh/ty/issues/1670)

crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -408,3 +408,15 @@ class Vec2(NamedTuple):
...
+
+<!-- snapshot-diagnostics -->
+
+```py
+from typing import NamedTuple
+
+class Foo(NamedTuple):
+    # error: [invalid-named-tuple] "NamedTuple field `_bar` cannot start with an underscore"
+    _bar: int
+```
```
charliermarsh, line 502: Might be worth adding a test for this case, which seems to be allowed:
```python
from typing import NamedTuple

class Foo(NamedTuple):
    bar: int

class Baz(Foo):
    _bop: int
```

PR #21699, opened 2025-11-29
[ty] Don't introduce invalid syntax when autofixing override-of-final-method

AlexWaygood:
## Summary

This didn't actually turn out to be too hard afterall.

- If the overriding definition isn't a function, mark the autofix as display-only (we don't know what we're dealing with)
- If any function overloads don't have the `StmtClassDef` node as their immediate parent in the AST (e.g. one or more overloads is inside a `StmtIf` node), mark the fix as display-only 
- If the overriding definition is the only statement in the class body, replace the overriding definition with `pass` rather than just deleting the range of the overriding definition

Fixes https://github.com/astral-sh/ty/issues/1678

## Test Plan

Updated snapshtos

### DIFF
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -37,7 +37,7 @@ use ruff_db::{
     parsed::parsed_module,
     source::source_text,
 };
-use ruff_diagnostics::{Applicability, Edit, Fix};
+use ruff_diagnostics::{Edit, Fix};
 use ruff_python_ast::name::Name;
 use ruff_python_ast::parenthesize::parentheses_iterator;
 use ruff_python_ast::{self as ast, AnyNodeRef, StringFlags};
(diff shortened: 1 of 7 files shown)


crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -3878,30 +3878,70 @@ pub(super) fn report_overridden_final_method<'db>(
...
+            .chain(implementation)
+            .all(|overload| {
+                class_node
+                    .body
+                    .iter()
+                    .filter_map(ast::Stmt::as_function_def_stmt)
+                    .contains(overload.node(db, context.file(), context.module()))
+            });
+
+        let applicability = if should_fix {
```
MichaReiser: I'm leaning towards not showing the fix in that case because, I think,  the is_only detection uses the wrong parent in that case and the more correct fix is to delete the entire outer statement, or the statement itself.

DisplayOnlyFixes are also never shown anywhere. They're just dead code

AlexWaygood: oh, I assumed making it display-only was what you were asking for in https://github.com/astral-sh/ruff/pull/21646#discussion_r2573103184 (we don't have an `Applicability::Manual` :-)

But sure, I'll just get rid of the fix in these cases.

> DisplayOnlyFixes are also never shown anywhere. They're just dead code

What's the purpose of `Applicability::DisplayOnly` if they aren't shown anywhere? I figured that if/when we show fixes in the CLI rendering of diagnostics, the diff would be helpful for users in telling them what we were asking for (you mentioned in https://github.com/astral-sh/ruff/pull/21646 that you found it confusing that the diagnostic was only omitted on one overload, but that to fix the diagnostic you actually had to remove _all_ overloads).

MichaReiser: The idea was to only show them in the CLI but we never made it that far. There were just too many design decisions.

PR #1683, opened 2023-01-06
Automatically remove unused variables

charliermarsh:
Closes #1460.

PR #21700, opened 2025-11-30
[ty] Forbid use of `super()` in `NamedTuple` subclasses

charliermarsh:
## Summary

The exact behavior around what's allowed vs. disallowed was partly detected through trial and error in the runtime.

I was a little confused by [this comment](https://github.com/python/cpython/pull/129352) that says "`NamedTuple` subclasses cannot be inherited from" because in practice that doesn't appear to error at runtime.

Closes [#1683](https://github.com/astral-sh/ty/issues/1683).

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/named_tuple.md
+++ b/crates/ty_python_semantic/resources/mdtest/named_tuple.md
@@ -411,7 +411,7 @@ Vec2(0.0, 0.0)
 
 ## `super()` is not supported in NamedTuple methods
 
-Using `super()` in a method of a `NamedTuple` subclass will raise an exception at runtime. In Python
+Using `super()` in a method of a `NamedTuple` class will raise an exception at runtime. In Python
 3.14+, a `TypeError` is raised; in earlier versions, a confusing `RuntimeError` about
 `__classcell__` is raised.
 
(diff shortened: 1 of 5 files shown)


crates/ty_python_semantic/resources/mdtest/named_tuple.md
```diff
@@ -408,3 +408,66 @@ class Vec2(NamedTuple):
 
 Vec2(0.0, 0.0)
 ```
+
+## `super()` is not supported in NamedTuple methods
+
+Using `super()` in a method of a `NamedTuple` subclass will raise an exception at runtime. In Python
```
AlexWaygood: extremely nitpicky comment: my preferred term is "`NamedTuple` class" rather than "`NamedTuple` subclass". The generated class doesn't actually have `NamedTuple` in its MRO (because `NamedTuple` is actually a function rather than a class!) -- the generated class actually inherits directly from `tuple`. The runtime does some magic to swap out `NamedTuple` for `tuple` in the class's bases, so it feels _slightly_ incorrect to refer to it as a "subclass of" `NamedTuple`, even if `NamedTuple` is included in the class's bases.

```suggestion
Using `super()` in a method of a `NamedTuple` class will raise an exception at runtime. In Python
```

crates/ty_python_semantic/resources/mdtest/named_tuple.md
AlexWaygood, line 1: You could also add a test to verify that using `super()` on a `NamedTuple` class works fine if it occurs outside the class, e.g.

```py
class F(NamedTuple):
    x: int

super(F, F(42))  # fine
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -5546,6 +5546,20 @@ impl KnownClass {
                             return;
                         };
 
+                        // Check if the enclosing class is a `NamedTuple`, which forbids the use of `super()`.
+                        if CodeGeneratorKind::NamedTuple.matches(db, enclosing_class, None) {
+                            if let Some(builder) = context
+                                .report_lint(&SUPER_CALL_IN_NAMED_TUPLE_METHOD, call_expression)
+                            {
+                                builder.into_diagnostic(format_args!(
+                                    "Cannot use `super()` in a method of NamedTuple subclass `{}`",
```
```suggestion
                                    "Cannot use `super()` in a method of NamedTuple class `{}`",
```

crates/ty_python_semantic/src/types/class.rs
```diff
@@ -5585,6 +5599,22 @@ impl KnownClass {
...
                     }
                     [Some(pivot_class_type), Some(owner_type)] => {
+                        // Check if the enclosing class is a `NamedTuple`, which forbids the use of `super()`.
+                        if let Some(enclosing_class) = nearest_enclosing_class(db, index, scope) {
+                            if CodeGeneratorKind::NamedTuple.matches(db, enclosing_class, None) {
+                                if let Some(builder) = context
+                                    .report_lint(&SUPER_CALL_IN_NAMED_TUPLE_METHOD, call_expression)
+                                {
+                                    builder.into_diagnostic(format_args!(
+                                        "Cannot use `super()` in a method of NamedTuple subclass `{}`",
```
```suggestion
                                        "Cannot use `super()` in a method of NamedTuple class `{}`",
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1760,6 +1761,33 @@ declare_lint! {
     }
 }
 
+declare_lint! {
+    /// ## What it does
+    /// Checks for calls to `super()` inside methods of `NamedTuple` subclasses.
```
```suggestion
    /// Checks for calls to `super()` inside methods of `NamedTuple` classes.
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1760,6 +1761,33 @@ declare_lint! {
...
+    ///
+    /// ## Why is this bad?
+    /// Using `super()` in a method of a `NamedTuple` subclass will raise an exception at runtime.
```
```suggestion
    /// Using `super()` in a method of a `NamedTuple` class will raise an exception at runtime.
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1760,6 +1761,33 @@ declare_lint! {
...
+    ///
+    ///     def method(self):
+    ///         super()  # error: super() is not supported in methods of NamedTuple subclasses
+    /// ```
+    ///
+    /// ## References
+    /// - [Python documentation: super()](https://docs.python.org/3/library/functions.html#super)
+    pub(crate) static SUPER_CALL_IN_NAMED_TUPLE_METHOD = {
+        summary: "detects `super()` calls in methods of `NamedTuple` subclasses",
+        status: LintStatus::preview("1.0.0"),
```
```suggestion
        summary: "detects `super()` calls in methods of `NamedTuple` classes",
        status: LintStatus::preview("0.0.1-alpha.30"),
```

AlexWaygood:
> I was a little confused by [this comment](https://github.com/python/cpython/pull/129352) that says "`NamedTuple` subclasses cannot be inherited from" because in practice that doesn't appear to error at runtime.

yeah, I think that comment was just incorrect

PR #21702, opened 2025-11-30
parser: add `is_stub` to `ParseOptions` to fix false positive syntax error for stub files

### LINKED ISSUE #21677: false positive syntax errors when using syntax for newer python version in `.pyi` files
### Summary

```py
# foo.pyi

def foo[T](): ...

type Foo = int
```
```
Cannot use type parameter lists on Python 3.8 (syntax was added in Python 3.12)
Cannot use `type` alias statement on Python 3.8 (syntax was added in Python 3.12)
```

since `.pyi` files are not executed at runtime, they should be able to use newer syntax than the target python version

### Version

0.14.6

11happy:
## Summary

This PR fixes #21677 

## Test Plan

I have locally tested the working for main.pyi:
```
def foo[T](): ...

type Foo = int
```
image


AlexWaygood:
I would personally want these syntax errors reported on my stub files if I specified a lower Python version (https://github.com/astral-sh/ruff/issues/21677#issuecomment-3592523385). Using new syntax makes stub files less portable, which isn't what I'd want if I was publishing library stubs.

11happy:
Closing this PR in support that these warnings can be opted out , which is much better than to remove them totally.
Thank you : )

Issue #10653, opened 2024-03-29
☂️ Error recovery improvements for the new parser


dhruvmanila:
This PR keeps track of any improvements that should be made when working on error recovery for the new parser. This will be _after_ #10036 is merged and when error recovery work is started.

- [ ] Avoid including the invalid expression node for empty and unclosed container types. For example,
	```python
	[
	```
	
	Here, there are no elements and the list isn't closed. Currently, the parser would emit an invalid node and report an error. The AST would contain a list expression with a single invalid node. Refer https://github.com/astral-sh/ruff/pull/10554#discussion_r1540762104

	This should be done for list, set, dictionary, and tuple.

dhruvmanila:
- [ ] 18. Consider changing `ast::ExprSubscript` to make `slice` field `Optional` so that an empty slice (which is an error) does not need to include an empty `ExprName` node as a placeholder.

11happy:
Hello @dhruvmanila , is this issue still open ? I noticed its not marked as `help_wanted` so I just wanted to check. I am interested to work on this and would really appreciate any pointers or context to get started. Thank you : )

ntBre:
@11happy are you interested in any of the sub-tasks in particular? I think these could be pretty tricky to work on and may require some design work from us, but Dhruv would know better than me.

11happy:
I am interested to start with first one `Avoid including the invalid expression node for empty and unclosed container types` , overall I am interested on this broader task & exploring other subtasks as I go,

MichaReiser:
@11happy this seems like a good start as it doesn't require any changes to the AST. Do you have any specific questions?

11happy:
I have linked the PR for first one.
Thank you : )

11happy:
I have linked the PR for 5th one , would appreciate a review.
Thank you : )

PR #21703, opened 2025-11-30
parser: remove invalid node from empty & unclosed container types

11happy:
## Summary

This PR removes the invalid expression node for empty and unclosed container types. Part of #10653

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -2200,9 +2200,6 @@ def frob(): ...
 ",
         );
 
-        // TODO: it would be good if `bar` was included here, but
-        // the list comprehension is not yet valid and so we do not
-        // detect this as a definition of `bar`.
         assert_snapshot!(
             builder.skip_keywords().skip_builtins().build().snapshot(),
             @"bar",


crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -2000,6 +2009,16 @@ impl<'src> Parser<'src> {
...
+        {
+            return Expr::Dict(ast::ExprDict {
+                items: vec![],
+                range: self.node_range(start),
+                node_index: AtomicNodeIndex::NONE,
+            });
+        }
+
         // Nice error message when having a unclosed open brace `{`
         if self.at_ts(NEWLINE_EOF_SET) {
```
MichaReiser, line 2014: Can't we ove the return here?

11happy: `NEWLINE_EOF_SET`  contains `NewLine` & `EndofFile` Token , but in these cases with empty & unclosed the first token encountered is Unknown  after  `Lbrace`

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -1944,9 +1944,18 @@ impl<'src> Parser<'src> {
     /// See: <https://docs.python.org/3/reference/expressions.html#list-displays>
     fn parse_list_like_expression(&mut self) -> Expr {
         let start = self.node_start();
-
         self.bump(TokenKind::Lsqb);
 
+        if self.at(TokenKind::Unknown)
+            && (self.peek() == TokenKind::EndOfFile || self.peek() == TokenKind::Newline)
```
MichaReiser: While this works for `[` at the end of a file, it doesn't help if we have something like this


```py
a = [

if True:
	pass
```

I think we want to use something like `if self.at_sequence_end() {` or similar instead.

11happy: done ,
Thank you : )

MichaReiser: Please add new tests demonstrating that this now works as expected

11happy: sure

11happy: done, I have added this test
```
a = [

if True:
	pass
```
should I add similar in case of set/dict etc

11happy: also are the snapshot changes expected ones ? can you please take a look.
Thank you

crates/ty_ide/src/completion.rs
```diff
@@ -2205,7 +2205,7 @@ def frob(): ...
         // detect this as a definition of `bar`.
         assert_snapshot!(
             builder.skip_keywords().skip_builtins().build().snapshot(),
-            @"<No completions found after filtering out completions>",
+            @"bar",
```
AlexWaygood, line 2205: Please remove the TODO comment a few lines above this — this PR fixes the TODO!

11happy: done

11happy:
ping!
would appreciate a review : )

MichaReiser (commented):
Thank you. We would still add way more tests here, e.g. with an example where a `while` statement or similar follows. 

After working on https://github.com/astral-sh/ruff/pull/21895 I came to the conclusion that we shouldn't make this change because it will result in the lexer and parser state diverging after the missing `]` because we never call `re_lex`.

Fixing this will require a more holistic take at https://github.com/astral-sh/ruff/issues/21896

But thank you for looking into this

11happy:
> Thank you. We would still add way more tests here, e.g. with an example where a `while` statement or similar follows.
> 
> After working on #21895 I came to the conclusion that we shouldn't make this change because it will result in the lexer and parser state diverging after the missing `]` because we never call `re_lex`.
> 
> Fixing this will require a more holistic take at #21896
> 
> But thank you for looking into this

Sure, I understand.
Thank you

PR #21704, opened 2025-11-30
[ty] Emit a diagnostic for subclassing with `order=True`

### LINKED ISSUE #1681: Add diagnostic warning about unsound subclassing of `order=True` dataclasses
It's unsound to subclass an `order=True` dataclass, because the generated comparison methods raise an exception if you try to compare an instance of the subclass with an instance of the superclass:

```py

@dataclass(order=True)
class F:
    x: int

class G(F): ...

G(42) < F(42)  # exception raised here at runtime, but cannot detected by ty
```

We should detect attempts to subclass `order=True` dataclasses and warn users about the unsoundness (emitting a diagnostic at the point where `G` is defined). We could recommend that they use `functools.total_ordering` instead to generate their comparison methods, as `total_ordering` doesn't have the same problem.

(One way of describing the problem here is that the design of the stdlib feature violates the Liskov Substitution Principle.)

charliermarsh:
## Summary

Closes https://github.com/astral-sh/ty/issues/1681.


charliermarsh:
I believe those changes on the conformance tests are "correct" in-so-far as those are instances of the case we want to catch here.

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -349,6 +349,49 @@ GenericWithOrder[int](1) < GenericWithOrder[int](1)
 GenericWithOrder[int](1) < GenericWithOrder[str]("a")  # error: [unsupported-operator]
 ```
 
+Subclassing a dataclass with `order=True` is problematic because comparing instances of different
+classes in the inheritance hierarchy will raise a `TypeError` at runtime. This violates the Liskov
+Substitution Principle:
```
AlexWaygood:
```suggestion
classes in the inheritance hierarchy will raise a `TypeError` at runtime. The design of the stdlib
feature therefore violates the Liskov Substitution Principle:
```

crates/ty_python_semantic/resources/mdtest/liskov.md
```diff
@@ -475,37 +475,24 @@ from typing import NamedTuple
...
-# on any method definition. Note also that no other type checker complains about this
-# as of 2025-11-21.
+# Specifying `order=True` on the subclass means that a `__lt__` method is generated that
+# is incompatible with the generated `__lt__` method on the superclass. We emit a diagnostic
+# on the class definition because the design of `order=True` dataclasses themselves violates
+# the Liskov Substitution Principle.
 @dataclass(order=True)
-class Bar2(Foo):
+class Bar2(Foo):  # error: [subclass-of-dataclass-with-order]
     y: str
```
AlexWaygood, line 1176: I wonder if we should suppress the `subclass-of-dataclass-with-order` diagnostic if all the comparison methods are overridden on the subclass? (That's the case here, since `Bar2` is also decorated with `order=True`.) This case here _does_ feel like it should be flagged using the error code we use for Liskov Subsitution Principle violations, because by specifying `order=True` the user has in a sense explicitly asked for an incompatible override.

And something like the following also seems safe, if the user has taken care to override all the unsound methods from the superclass:

```py

@dataclass(order=True)
class A:
    int

class B(A):
    def __lt__(self, other: A) -> bool:
        return True

    def __le__(self, other: A) -> bool:
        return True

    __gt__ = __lt__
    __ge__ = __le__
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -743,6 +744,22 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
+                            .report_lint(&SUBCLASS_OF_DATACLASS_WITH_ORDER, &class_node.bases()[i])
+                        {
+                            builder.into_diagnostic(format_args!(
+                                "Class `{}` inherits from dataclass `{}` which has `order=True`",
+                                class.name(self.db()),
+                                base_class.name(self.db()),
+                            ));
+                        }
+                    }
+                }
```
AlexWaygood: can we add an `info` subdiagnostic here explaining why this is problematic?

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1643,6 +1644,45 @@ declare_lint! {
...
+    /// ## Example
+    ///
+    /// ```python
+    /// from dataclasses import dataclass
+    ///
+    /// @dataclass(order=True)
+    /// class Parent:
+    ///     value: int
+    ///
+    /// class Child(Parent):  # Error raised here
```
```suggestion
    /// class Child(Parent):  # Ty emits an error here
```

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -1643,6 +1644,45 @@ declare_lint! {
...
+    ///     pass
+    ///
+    /// # At runtime, this raises TypeError:
+    /// # Child(1) < Parent(2)
+    /// ```
+    ///
+    /// Consider using `functools.total_ordering` instead, which does not have this limitation.
+    ///
+    /// [Liskov Substitution Principle]: https://en.wikipedia.org/wiki/Liskov_substitution_principle
```
```suggestion
    /// Consider using [`functools.total_ordering`][total_ordering] instead, which does not have this limitation.
    ///
    /// [Liskov Substitution Principle]: https://en.wikipedia.org/wiki/Liskov_substitution_principle
    /// [total_ordering]: https://docs.python.org/3/library/functools.html#functools.total_ordering
```

AlexWaygood (commented):
Nice!

crates/ty_python_semantic/src/types/diagnostic.rs
MichaReiser: ...ty

AlexWaygood:
@charliermarsh are you still interested in getting this over the line?

charliermarsh:
Yes sir, sorry, lost track.

charliermarsh:
I think the Expression diagnostic is correct:

```
  Expression (Try inherits from Result)

  - Result is decorated with @tagged_union(frozen=True, order=True)
  - tagged_union is marked with @dataclass_transform(), so ty recognizes it as a dataclass-like decorator
  - When order=True is passed, the class gets ordering methods
  - Try inherits from Result without defining its own comparison methods
  - True positive: At runtime, Try(Success(1)) < Result(Success(2)) would raise TypeError
```

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -1060,23 +1060,58 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-                    if base_params.is_frozen() != class_is_frozen {
-                        report_bad_frozen_dataclass_inheritance(
-                            &self.context,
-                            class,
-                            class_node,
-                            base_class_literal,
-                            &class_node.bases()[i],
-                            base_params,
-                        );
+                    if let Some(class_params) = class.dataclass_params(self.db()) {
```
sharkdp: This will only for work standard library dataclasses. I haven't looked closer at this PR, but we should probably make sure that this works for `dataclass_transformer`s as well. See https://github.com/astral-sh/ruff/pull/23366 for a related PR that expands the frozen/non-frozen-inheritance checks here in the same way.

charliermarsh: Thanks, I attempted to fix it...

sharkdp: Thank you. I will get this PR merged once #23366 lands.

crates/ty_python_semantic/resources/mdtest/dataclasses/dataclasses.md
```diff
@@ -434,6 +434,103 @@ GenericWithOrder[int](1) < GenericWithOrder[int](1)
...
+
+```py
+@dataclass(order=True)
+class OrderedParent2:
+    x: int
+
+@dataclass(order=True)
+class OrderedChild2(OrderedParent2):
+    y: str
+```
```
sharkdp: But `OrderedChild2(1,2) < OrderedParent2(1)` also fails at runtime. I'm not sure I understand why we suppress the diagnostic in this case?

sharkdp (changes requested):
I wanted to get this merged today, but I think it's not quite there yet, so I'm moving this back to draft for now. I'll finish this once I get back.

It doesn't seem to be of high priority. No other type checker supports this, and it's not tested in the conformance suite.

sharkdp:
Charlie, sorry for the delay here. This is still low prio, and more complicated than it looks. Working on it...

sharkdp:
My current problem is that I think inheriting from `order=True` is basically never safe? That is, except for edge cases where you don't add new fields in the child class? Even something like this fails:
```pycon
>>> @dataclass(order=True)
... class C:
...     x: int
...
>>> @dataclass(order=True)
... class D(C):
...     y: str
...
>>> C(1) < D(2, "a")
TypeError: '<' not supported between instances of 'C' and 'D'
```

That means you can't safely use a `D` where a `C` is expected, and therefore this new rule here should flag the construction of `D`. Enforcing that would require us to emit this diagnostic in too many places, though. So I'm wondering if it's all that useful. Interested in @AlexWaygood's opinion, since you proposed this rule originally.

charliermarsh:
I'm fine to just close this and prioritize other work if you prefer, not a big deal and clearly more complicated than initially perceived.

AlexWaygood:
> My current problem is that I think inheriting from `order=True` is basically never safe? That is, except for edge cases where you don't add new fields in the child class?

Even cases where you don't add new fields in the child class are unsafe, I believe -- see the example in the original issue I wrote up!

> Enforcing that would require us to emit this diagnostic in too many places, though.

Not sure I follow... you just think it would be so many diagnostics that it would be overwhelming for users? Maybe, but I'm not sure `order=True` is a _very_ popular feature, and subclassing `order=True` dataclasses is even rarer. And there's a pretty simple solution to the diagnostic: just use `functools.total_ordering` instead of using `order=True` if you intend to subclass the dataclass. `functools.total_ordering` doesn't have the same issue. We could even add an autofix that makes the change for you in the future.

sharkdp:
> > Enforcing that would require us to emit this diagnostic in too many places, though.
> 
> Not sure I follow... you just think it would be so many diagnostics that it would be overwhelming for users?

As you can see from the ecosystem report, some dataclass-transformer libraries set `order_default=True`, and so you would basically have to set `order=False` on your model to enable any sort of subclassing. So yes, I assume this would be quite disruptive? Maybe there's a reason that no other type checker has implemented a rule like this?

But I guess we could make that rule opt-in?

AlexWaygood:
Ah I see. Sorry, I hadn't looked at the ecosystem report, and didn't consider the impact on third-party dataclass-transformer libraries.

Are we not able to distinguish between stdlib dataclasses and third-party libraries? Could we add that distinction? I'd be fine making this rule only apply to stdlib dataclasses.

We may need to add this distinction anyway, to fix issues like https://github.com/astral-sh/ty/issues/3486, where third-party dataclass-transform libraries behave differently to stdlib dataclasses.

sharkdp:
> Are we not able to distinguish between stdlib dataclasses and third-party libraries? Could we add that distinction? I'd be fine making this rule only apply to stdlib dataclasses.

It's easy to distinguish between the two. In fact, the original version here only implemented the check for stdlib dataclasses, but I argued that we should treat dataclass transformers in the same way (*"If dataclass_transform is applied to a class, dataclass-like semantics will be assumed [...]"*) 🙃 

I'll leave it up to you to decide what we do here (I will make the changes accordingly):
* flag both stdlib dataclasses and transformers, but make the rule opt-in
* only flag stdlib dataclasses, keep the rule enabled by default
* do not implement this rule at all (after all, there's only a single ecosystem hit for stdlib dataclasses, IIRC)

charliermarsh:
> only flag stdlib dataclasses, keep the rule enabled by default

I'm cool with this, though also cool with just closing for now and calling this a non-priority for stable.

AlexWaygood:
I'd also go with

> only flag stdlib dataclasses, keep the rule enabled by default

I think unfortunately it's pretty clear that third-party `dataclass_transform` libraries often differ from the stdlib `dataclasses` module in edge cases. So whereas we want to abide by this when it comes to synthesizing dataclass fields and methods, etc.:

> If dataclass_transform is applied to a class, dataclass-like semantics will be assumed [...]"

...I'm less sure that we want to abide by that principle so strictly when it comes to things like "in this specific edge case, the stdlib dataclasses module is known to raise a `TypeError`, and we'll emit a diagnostic warning you about that" (which this feels closer to).

sharkdp (approved):
Ok, this is happening! 

(very sorry this took so long)

PR #21706, opened 2025-11-30
docs: Output file option with GitLab integration

kieran-ryan:
## Summary

Document availability and usage of the `--output-file` option within the GitLab integration - over usage of `>`.

Interested whether there is rationale to retain the existing sample otherwise - in which case this issue can be closed.


MichaReiser:
thank you. I don't think there's any specific reason for using `>` over `--output-file` other than that this documentation was written before we added support for `--output-file`.

PR #21718, opened 2025-12-01
[ty] do nothing with `store_expression_type` if `inner_expression_inference_state` is `Get`

mtshiba:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1688

## Test Plan

N/A

### DIFF
--- a/crates/ty_python_semantic/resources/corpus/inner_expression_inference_state.py
+++ b/crates/ty_python_semantic/resources/corpus/inner_expression_inference_state.py
@@ -1,0 +1,6 @@
+# This is a regression test for `store_expression_type`.
+# ref: https://github.com/astral-sh/ty/issues/1688
+
+x: int
+
+type x[T] = x[T, U]


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -7093,10 +7093,13 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
 
     #[track_caller]
     fn store_expression_type(&mut self, expression: &ast::Expr, ty: Type<'db>) {
-        if self.deferred_state.in_string_annotation() {
+        if self.deferred_state.in_string_annotation()
+            || self.inner_expression_inference_state.is_get()
```
MichaReiser, line 7097: Could we use `multi_inference_state::Ignore` instead of having two modes that represent "read-only" type inference?

Gankra: drive-by context: I am likely to modify this check for `in_string_annotation` to do the "store (Expr, Expr) for string annotation types" thing we've been discussing.

mtshiba: > Could we use `multi_inference_state::Ignore` instead of having two modes that represent "read-only" type inference?

I understand that `MultiInferenceState::Ignore` is an option to perform the same calculation twice and discard the second result.
Since all type inference for expressions is skipped while in `InnerExpressionInferenceState::Get`, I think it should be used whenever possible (for example, when only diagnostics are needed, such as the handling in `infer_subscript_type_expression` for union types).
But I'm not sure whether this can be assumed in all cases where `MultiInferenceState::Ignore` is used.

carljm: I think it is necessary for `MultiInferenceState` to redo inference, since it is doing it with different type context. So I think we need both.

PR #10636, opened 2024-03-27
Add tests for dictionary comprehension

dhruvmanila:
## Summary

This PR adds test cases for dictionary comprehension.

PR #21720, opened 2025-12-01
parser: parsing doublestar as a power op without a left hand side

11happy:
## Summary

Parses `**` Doublestar as power op without lhs, Part of #10653, referred discussion https://github.com/astral-sh/ruff/pull/10636#discussion_r1542405830

## Test Plan

test case :
```python
{x: **y for x, y in data}
```

ruff-dev output before:
```
Error: Expected an expression at byte range 4..6

Caused by:
    Expected an expression
```

ruff-dev output ast after:
```
Module(
    ModModule {
        node_index: NodeIndex(None),
        range: 0..25,
        body: [
            Expr(
                StmtExpr {
                    node_index: NodeIndex(None),
                    range: 0..25,
                    value: DictComp(
                        ExprDictComp {
                            node_index: NodeIndex(None),
                            range: 0..25,
                            key: Name(
                                ExprName {
                                    node_index: NodeIndex(None),
                                    range: 1..2,
                                    id: Name("x"),
                                    ctx: Load,
                                },
                            ),
                            value: BinOp(
                                ExprBinOp {
                                    node_index: NodeIndex(None),
                                    range: 4..7,
                                    left: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 7..7,
                                            id: Name(""),
                                            ctx: Invalid,
                                        },
                                    ),
                                    op: Pow,
                                    right: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 6..7,
                                            id: Name("y"),
                                            ctx: Load,
                                        },
                                    ),
                                },
                            ),
                            generators: [
                                Comprehension {
                                    range: 8..24,
                                    node_index: NodeIndex(None),
                                    target: Tuple(
                                        ExprTuple {
                                            node_index: NodeIndex(None),
                                            range: 12..16,
                                            elts: [
                                                Name(
                                                    ExprName {
                                                        node_index: NodeIndex(None),
                                                        range: 12..13,
                                                        id: Name("x"),
                                                        ctx: Store,
                                                    },
                                                ),
                                                Name(
                                                    ExprName {
                                                        node_index: NodeIndex(None),
                                                        range: 15..16,
                                                        id: Name("y"),
                                                        ctx: Store,
                                                    },
                                                ),
                                            ],
                                            ctx: Store,
                                            parenthesized: false,
                                        },
                                    ),
                                    iter: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 20..24,
                                            id: Name("data"),
                                            ctx: Load,
                                        },
                                    ),
                                    ifs: [],
                                    is_async: false,
                                },
                            ],
                        },
                    ),
                },
            ),
        ],
    },
)


```


11happy:
> Would you mind adding a few parser tests that demonstrate that the new recovery logic works as expected?

 > E.g., the ecosystem changes suggest that we:

 > a) Fail to emit a diagnostic in that case
 > b) or we now use this code path for **kwargs when we shouldn't

I have added tests & I think its not the option b, I would appreciate any pointers to move forward

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -363,6 +363,24 @@ impl<'src> Parser<'src> {
...
+                        id: Name::empty(),
+                        ctx: ExprContext::Invalid,
+                        node_index: AtomicNodeIndex::NONE,
+                    })),
+                    op: Operator::Pow,
+                    right: Box::new(right.expr),
+                    range: self.node_range(start),
+                    node_index: AtomicNodeIndex::NONE,
+                })
+                .into();
```
MichaReiser: I'm not sure if this is the right place for the fix. Would it be possible to add it to `parse_binary_expression_or_higher` or even in `parse_binary_expression_or_higher_recursive`? If not, can you explain why?

Moving it their would give you the diagnostic for free

11happy: I think this might be correct place as we need to detect the missing left operand & in `parse_binary_expression_or_higher_recursive` the LHS is passed already as a `ParsedExpr`

11happy: done I have written the inline tests.

11happy:
ping! 
would appreciate a review : )

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -363,6 +363,29 @@ impl<'src> Parser<'src> {
         }
 
         match self.current_token_kind() {
+            TokenKind::DoubleStar => {
+                let current_start = self.node_start();
+                self.bump(TokenKind::DoubleStar);
+                let right =
+                    self.parse_binary_expression_or_higher(OperatorPrecedence::Exponent, context);
+                // test_ok missing_power_lhs
```
MichaReiser: This should be an invalid test and the parser should emit a diagnostic

11happy: done 
Thank you : )

crates/ruff_python_parser/src/parser/expression.rs
```diff
@@ -363,6 +363,41 @@ impl<'src> Parser<'src> {
...
+                // foo(**kwargs)
+
+                self.add_error(
+                    ParseErrorType::OtherError(
+                        "Power operator '**' requires a left operand".to_string(),
+                    ),
+                    &expr,
+                );
+                return expr.into();
+            }
```
MichaReiser, line 400: I kept wondering why this branch is necessary and, in fact, removing it gives us the desired parsing. So I dont' think we need to do anything more here other than adding a few tests for it.

However, the tests show that we mess up node ranges somehow:

```
Module(
    ModModule {
        node_index: NodeIndex(None),
        range: 0..22,
        body: [
            Expr(
                StmtExpr {
                    node_index: NodeIndex(None),
                    range: 0..22,
                    value: DictComp(
                        ExprDictComp {
                            node_index: NodeIndex(None),
                            range: 0..22,
                            key: Name(
                                ExprName {
                                    node_index: NodeIndex(None),
                                    range: 1..2,
                                    id: Name("x"),
                                    ctx: Load,
                                },
                            ),
                            value: BinOp(
                                ExprBinOp {
                                    node_index: NodeIndex(None),
                                    range: 4..7,
                                    left: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 3..3,
                                            id: Name(""),
                                            ctx: Invalid,
                                        },
                                    ),
                                    op: Pow,
                                    right: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 6..7,
                                            id: Name("y"),
                                            ctx: Load,
                                        },
                                    ),
                                },
                            ),
                            generators: [
                                Comprehension {
                                    range: 8..21,
                                    node_index: NodeIndex(None),
                                    target: Tuple(
                                        ExprTuple {
                                            node_index: NodeIndex(None),
                                            range: 12..16,
                                            elts: [
                                                Name(
                                                    ExprName {
                                                        node_index: NodeIndex(None),
                                                        range: 12..13,
                                                        id: Name("x"),
                                                        ctx: Store,
                                                    },
                                                ),
                                                Name(
                                                    ExprName {
                                                        node_index: NodeIndex(None),
                                                        range: 15..16,
                                                        id: Name("y"),
                                                        ctx: Store,
                                                    },
                                                ),
                                            ],
                                            ctx: Store,
                                            parenthesized: false,
                                        },
                                    ),
                                    iter: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 20..21,
                                            id: Name("z"),
                                            ctx: Load,
                                        },
                                    ),
                                    ifs: [],
                                    is_async: false,
                                },
                            ],
                        },
                    ),
                },
            ),
        ],
    },
)
```

See how the invalid `ExprName` has a range of `3..3` but the `ExprBinOp` has a range of `4..7`. Not sure why this is happening


Which I think requires addressing https://github.com/astral-sh/ruff/blob/af623e1f9a0a99a471ab5a3a066bf8f1686e894d/crates/ruff_python_parser/src/parser/mod.rs#L278-L287

11happy: would you like me to remove this branch & keep the error with tests?
however I would like to point out as you can see in snapshots
```
value: BinOp(
                                ExprBinOp {
                                    node_index: NodeIndex(None),
                                    range: 47..50,
                                    left: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 47..47,
                                            id: Name(""),
                                            ctx: Invalid,
                                        },
                                    ),
                                    op: Pow,
                                    right: Name(
                                        ExprName {
                                            node_index: NodeIndex(None),
                                            range: 49..50,
                                            id: Name("y"),
                                            ctx: Load,
                                        },
                                    ),
                                },
```
here the ranges are identified correctly with current changes,

MichaReiser: I suggest removing this branch and fixing the underlying issue why the ranges are incorrect. Although I expect this to be somewhat tricky.

11happy: sure let me investigate on the underlying issue.

PR #21722, opened 2025-12-01
[ty] WIP: sort union types during cycle recovery to stabilize output

mtshiba:
## Summary

Fixes https://github.com/astral-sh/ty/issues/1670

This PR addresses the issue of non-deterministic execution order of salsa queries causing the order of elements in union types in the output to fluctuate randomly.

We already have a comparison function for sorting union types, `union_or_intersection_elements_ordering`, but this probably won't solve the problem, since it compares based on salsa IDs rather than the contents of the interned structs.

Instead, we use the `structural_type_ordering` function, which performs comparisons by deep traversing the type's contents.


mtshiba:
Hmm, at least, this approach alone doesn't seem to eliminate the instability of the diagnostic itself, rather than the instability of the output type.

1st: https://github.com/astral-sh/ruff/actions/runs/19820273428/job/56780817916?pr=21722

```sh
scikit-build-core (https://github.com/scikit-build/scikit-build-core)
+ src/scikit_build_core/_logging.py:153:13: warning[unsupported-base] Unsupported class base with type `<class 'Mapping[str, Style]'> | <class 'Mapping[str, Divergent]'>`
- Found 41 diagnostics
+ Found 42 diagnostics
```

2nd: https://github.com/astral-sh/ruff/actions/runs/19820836584/job/56782606719?pr=21722

```sh
scikit-build-core (https://github.com/scikit-build/scikit-build-core)
- src/scikit_build_core/_logging.py:153:13: warning[unsupported-base] Unsupported class base with type `<class 'Mapping[str, Style]'> | <class 'Mapping[str, Divergent]'>`
- Found 42 diagnostics
+ Found 41 diagnostics
```

MichaReiser:
> Hmm, at least, this approach alone doesn't seem to eliminate the instability of the diagnostic itself, rather than the instability of the output type.

I think you need to create a PR targeting this PR to verify your change because mypy primer compares main vs this branch. What you want is to compare the branch to itself to see if there are any remaining instabilities

mtshiba:
> I think you need to create a PR targeting this PR to verify your change because mypy primer compares main vs this branch. What you want is to compare the branch to itself to see if there are any remaining instabilities

Ah, I see, main is unstable, so it can't be used for comparison.

MichaReiser:
The pydantic speedup is interesting

I added a CI job in https://github.com/astral-sh/ruff/pull/21864 that should make it easier to verify a fix.

mtshiba:
> I added a CI job in #21864 that should make it easier to verify a fix.

Thanks!

> The pydantic speedup is interesting

There seems to be a trade-off between the cost of sorting unions and the cost of recalculation in the fixed point iterations, and in many cases the sorting cost is higher, but in the case of pydantic, the improvement in the convergence speed of the fixed point iterations seems to outweigh it.
We may be able to improve overall performance by adding a handling that sorts all unions if the fixed point iterations don't converge within a few cycles.

mtshiba:
Sorting only the union types that appear in cycle heads does not seem to eliminate the non-determinism, likely because the non-determinism in query execution order also makes it non-deterministic which queries become cycle heads.

PR #21499, opened 2025-11-17
[ty] Fix `Todo` type for starred elements in tuple/list/set expressions

AlexWaygood:
## Summary

Fixes https://github.com/astral-sh/ty/issues/2071

PR #21723, opened 2025-12-01
[ty] Fix false positives for `class F(Generic[*Ts]): ...`

AlexWaygood:
## Summary

I believe this gets rid of around 5 `invalid-argument-type` diagnostics on `scipy-stubs` (relates to https://github.com/astral-sh/ty/issues/1685).

## Test Plan

mdtest added

### DIFF
--- a/crates/ty_python_semantic/src/types.rs
+++ b/crates/ty_python_semantic/src/types.rs
@@ -957,8 +957,13 @@ impl<'db> Type<'db> {
         self.is_instance_of(db, KnownClass::NotImplementedType)
     }
 
-    pub(crate) const fn is_todo(&self) -> bool {
-        matches!(self, Type::Dynamic(DynamicType::Todo(_)))
+    pub(crate) fn is_todo(&self) -> bool {
+        self.as_dynamic().is_some_and(|dynamic| match dynamic {
+            DynamicType::Any | DynamicType::Unknown | DynamicType::Divergent(_) => false,
+            DynamicType::Todo(_) | DynamicType::TodoStarredExpression | DynamicType::TodoUnpack => {
+                true
+            }
+        })
     }
 
     pub const fn is_generic_alias(&self) -> bool {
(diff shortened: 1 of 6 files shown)


AlexWaygood:
The conformance-suite diff looks good. Lots of false positives going away. There are a few lines marked as `# E` where we previously emitted diagnostics and no longer do, but in all those cases we were emitting a diagnostic for the wrong reasons.

AlexWaygood:
The primer report shows some false positives going away on scipy-stubs and egglog-python. It also features a bunch of pre-existing false positives that have their error messages change. Note that I have a [WIP PR](https://github.com/astral-sh/ruff/pull/21499) that will reduce the places where we infer a `Todo` type for starred expressions, and clear up some of those other pre-existing false positives. But that PR's not ready for review yet.

crates/ty_python_semantic/src/types.rs
```diff
@@ -8859,13 +8866,8 @@ impl std::fmt::Display for DynamicType {
             // `DynamicType::Todo`'s display should be explicit that is not a valid display of
             // any other type
             DynamicType::Todo(todo) => write!(f, "@Todo{todo}"),
-            DynamicType::TodoUnpack => {
-                if cfg!(debug_assertions) {
-                    f.write_str("@Todo(typing.Unpack)")
-                } else {
-                    f.write_str("@Todo")
-                }
```
sharkdp, line 8867: Not doing this anymore leads to an inconsistent display of various todo types in release mode, but I guess it doesn't hurt to have the one with the message in release builds as well (for these very special todo types)

AlexWaygood: yes. One of the criticisms of `Todo` types in the past has been that we don't tell the user what exactly is still to be done if ty has been built in release mode. It seems like a feature that we can show that information for at least _some_ TODOs 😄

PR #17584, opened 2025-04-23
Allow EXE001 & EXE002 on WSL, fix issues with other cases of mounting non-unix filesystems

MusicalNinjaDad:
I'm raising this draft-PR for review, as an attempt to handle non-unix filesystems in a way which satisfies not only WSL (#3110, #5445, #10084) but also non-WSL (#12941) cases.

> **tl;dr:** Seems to be a touch faster than current implementation on *normal* systems; simple to document & use; also supports cases like #12941 / external FAT-drives / ...

**The key changes are in: [`crates/ruff_linter/src/rules/flake8_executable/helpers.rs`](https://github.com/MusicalNinjaDad/forks-ruff/blob/wsl_shebang/crates/ruff_linter/src/rules/flake8_executable/helpers.rs)**

### Recommended Approach: Check filemode of `pyproject.toml`, fallback to a `NamedTempFile`

- Using a `OnceLock` to ensure we only run this check once
- If `project_root/pyproject.toml` exists: check the filemode. If it's not executable - then files aren't executable by default
- If it is, or doesn't exist: create a `NamedTempFile` in `project_root` and check the filemode

#### Pros

- Single approach for all cases
- Guaranteed to catch all cases
- Faster than current implementation (if pyproject.toml exists)

#### Cons

- (Probably) has a very small performance cost (consistently at the edge of what I could measure) for EXE002 in the specific case when
  - no pyproject.toml is available in the project_root, or this has been accidentally set to executable (assuming this is a rare case?)
  - file is not executable & has no shebang (usual case)
  - not WSL (usual case)
  - project is small or user is linting a single file

### Alternative 1: Manual Override

See [617cf6d39abe9b1954c2f291bb23aeacb22d8c9e](https://github.com/MusicalNinjaDad/forks-ruff/blob/617cf6d39abe9b1954c2f291bb23aeacb22d8c9e/crates/ruff_linter/src/rules/flake8_executable/mod.rs)

- Provide a environment-variable override
- Otherwise only test default filemode on WSL

#### Pros

- No risk of checking a tempfile unless using WSL
- Allows for cases like #12941
- Follows precedent of `RUFF_CACHE_DIR`

#### Cons

- Requires users to find the relevant documentation and set a specific variable
- Either invoves testing an env-var during the check (feels wrong);
[...]

PR #17548, opened 2025-04-22
[red-knot] Add more tests for protocol members

AlexWaygood:
## Summary

Add tests that cover:
- Sub-protocols inheriting and extending members from their superclasses
- Protocols with members created using weird and wacky bindings such as `for` loops, `match` statements, and more

## Test Plan

this pr is all tests

Issue #12941, opened 2024-08-16
EXE002 on file systems that do not support executable bit


Argmaster:
This issue is somewhat releated to #5445 and #10084 but applies to more generic version of reported issues.
In general on `NTFS` and on `FAT` file systems executable permissions are not avialable by default resulting in all files being executable. What I noticed is that the problem is not as simple as avoiding checking this rule on windows / wsl etc. as NTFS and FAT system may also be used for drives which are shared between Windows and Linux OS.

In my case, I have an `exFAT` drive shared between `Windows 11` and `Ubuntu 24.04` and all files appear as executable. When I am running ruff on Ubuntu 24.04 `EXE002` is triggered.

```bash
-rwxr-xr-x  1 argmaster argmaster   2543 Aug 16 19:53 conftest.py
drwxr-xr-x  3 argmaster argmaster 131072 Aug 16 20:29 examples
drwxr-xr-x 11 argmaster argmaster 131072 Aug 16 19:56 gerberx3
-rwxr-xr-x  1 argmaster argmaster     35 Aug 16 19:53 .gitignore
```

Hence I would recommend adding some additional heuristic to detect if filesystem used supports executable permissions. This should only be needed on Unix family of operating systems.

Simple, although not perfect way to detect if file system supports executable bit could be to create temporary file next to a python file we are checking and make an attempt to toggle executable bit on it. Handling symlinks could get tricky though. I am not knowledgable enough to suggest better solution unfortunately.

Python `3.12.0`
ruff `0.5.1`

MichaReiser:
Thanks for the detailed write up. Yes, this is a known limitation and we discussed some solutions in https://github.com/astral-sh/ruff/issues/10084 I'll merge the two issues

Issue #5445, opened 2023-06-29
Avoid enforcing EXE001 and EXE002 checks on WSL + Windows *file systems*


bersbersbers:
This is a follow-up to https://github.com/astral-sh/ruff/issues/3110, and I am not sure it is valid, but I wanted to at least bring it up.

Running WSL on Windows means one can easily use a Windows file system as working directory. However, such file system do not support the executable flag, so to make *anything* executable, *everything* is considered executable (see also https://github.com/microsoft/WSL/issues/936). That triggers EXE002.

So I wonder if #3111 could be extended to ignore not only Window operation system, but also Windows file systems on Linux. One way of detecting such file systems would be by creating a new file and checking if it is executable:

```
bers@hostname:~$ touch foo
bers@hostname:~$ ll foo
-rw-r--r-- 1 bers bers 0 Jun 29 20:06 foo
bers@hostname:~$ cd /mnt/c/Code/project
bers@hostname:/mnt/c/Code/project$ touch foo
bers@hostname:/mnt/c/Code/project$ ll foo
-rwxrwxrwx 1 bers bers 0 Jun 29 20:06 foo*
```

bersbersbers:
Another way might be this:
```
bers@hostname:/mnt/c/Code/project$ df -T | grep /mnt/c
drvfs          9p       498171168 427346648  70824520  86% /mnt/c
```
but this may yield false-positives since metadata can also be enabled on these file systems.

In fact, a suggested workaround is to mount `/mnt/c` with `metadata`, which in fact allows to do `chmod -x` succesfully. 

However:
```
bers@hostname:/mnt/c/Code/project$ more /etc/wsl.conf
[automount]
options = "metadata"

bers@hostname:/mnt/c/Code/project$ ll bug.py
-rw-rw-rw- 1 bers bers 1232 Jun 29 13:37 bug.py

bers@hostname:/mnt/c/Code/project$ ruff bug.py
bug.py:1:1: EXE002 The file is executable but no shebang is present
Found 1 error.
```

I had to `rm -rf .ruff_cache/` to make it work.

charliermarsh:
I think this makes sense. There's a crate we can use for this apparently: https://crates.io/crates/wsl.

PR #21724, opened 2025-12-01
Drop WSL1 special case in flake8-executable

### LINKED ISSUE #10084: EXE001 and EXE002 do not run on WSL

K900:
## Summary

WSL1 is very very dead, and WSL2 handles permissions properly. Fixes `cargo test` on Ruff itself failing on WSL2, and removes what's generally a high potential footgun.

Fixes #10084 in a simpler way than #17584.

## Test Plan

`cargo test` on a WSL2 system.

### DIFF
--- a/crates/ruff_linter/src/rules/flake8_executable/rules/shebang_missing_executable_file.rs
+++ b/crates/ruff_linter/src/rules/flake8_executable/rules/shebang_missing_executable_file.rs
@@ -28,7 +28,11 @@ use crate::rules::flake8_executable::helpers::is_executable;
 ///
 /// A file is considered executable if it has the executable bit set (i.e., its
 /// permissions mode intersects with `0o111`). As such, _this rule is only
-/// available on Unix-like systems_, and is not enforced on Windows or WSL.
+/// available on Unix-like systems_.
+///
+/// If this rule produces a significant amount of false positives, it's likely
+/// your source is stored on a filesystem that does not implement Unix file
+/// permissions (e.g. FAT variants, NTFS, CIFS network mounts, VM/WSL shared folders, etc).
 ///
 /// ## References
 /// - [Python documentation: Executable Python Scripts](https://docs.python.org/3/tutorial/appendix.html#executable-python-scripts)
@@ -47,12 +51,6 @@ impl Violation for ShebangMissingExecutableFile {
 /// EXE002
 #[cfg(target_family = "unix")]
 pub(crate) fn shebang_missing_executable_file(filepath: &Path, context: &LintContext) {
-    // WSL supports Windows file systems, which do not have executable bits.
-    // Instead, everything is executable. Therefore, we skip this rule on WSL.
-
-    if is_wsl::is_wsl() {
-        return;
-    }
     if let Ok(true) = is_executable(filepath) {
         context.report_diagnostic_if_enabled(
             ShebangMissingExecutableFile,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1671,15 +1671,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "is-docker"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "928bae27f42bc99b60d9ac7334e3a21d10ad8f1835a4e12ec3ec0464765ed1b3"
-dependencies = [
- "once_cell",
-]
-
-[[package]]
 name = "is-macro"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1700,16 +1691,6 @@ dependencies = [
  "hermit-abi",
  "libc",
  "windows-sys 0.59.0",
-]
-
-[[package]]
-name = "is-wsl"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "173609498df190136aa7dea1a91db051746d339e18476eed5ca40521f02d7aa5"
-dependencies = [
- "is-docker",
- "once_cell",
 ]
 
 [[package]]
@@ -3131,7 +3112,6 @@ dependencies = [
  "imperative",
  "insta",
  "is-macro",
- "is-wsl",
  "itertools 0.14.0",
  "jiff",
  "libcst",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,6 @@ indoc = { version = "2.0.4" }
 insta = { version = "1.35.1" }
 insta-cmd = { version = "0.6.0" }
 is-macro = { version = "0.3.5" }
-is-wsl = { version = "0.4.0" }
 itertools = { version = "0.14.0" }
 jiff = { version = "0.2.0" }
 js-sys = { version = "0.3.69" }
--- a/crates/ruff_linter/Cargo.toml
+++ b/crates/ruff_linter/Cargo.toml
@@ -41,7 +41,6 @@ globset = { workspace = true }
 hashbrown = { workspace = true }
 imperative = { workspace = true }
 is-macro = { workspace = true }
-is-wsl = { workspace = true }
 itertools = { workspace = true }
 libcst = { workspace = true }
 jiff = { workspace = true }
--- a/crates/ruff_linter/src/rules/flake8_executable/rules/shebang_not_executable.rs
+++ b/crates/ruff_linter/src/rules/flake8_executable/rules/shebang_not_executable.rs
@@ -27,7 +27,11 @@ use crate::rules::flake8_executable::helpers::is_executable;
 ///
 /// A file is considered executable if it has the executable bit set (i.e., its
 /// permissions mode intersects with `0o111`). As such, _this rule is only
-/// available on Unix-like systems_, and is not enforced on Windows or WSL.
+/// available on Unix-like systems_.
+///
+/// If this rule produces a significant amount of false positives, it's likely
+/// your source is stored on a filesystem that does not implement Unix file
+/// permissions (e.g. FAT variants, NTFS, CIFS network mounts, VM/WSL shared folders, etc).
 ///
 /// ## Example
 /// ```python
@@ -51,13 +55,6 @@ impl Violation for ShebangNotExecutable {
 /// EXE001
 #[cfg(target_family = "unix")]
 pub(crate) fn shebang_not_executable(filepath: &Path, range: TextRange, context: &LintContext) {
-    // WSL supports Windows file systems, which do not have executable bits.
-    // Instead, everything is executable. Therefore, we skip this rule on WSL.
-
-    if is_wsl::is_wsl() {
-        return;
-    }
-
     if let Ok(false) = is_executable(filepath) {
         context.report_diagnostic_if_enabled(ShebangNotExecutable, range);
     }


amyreese (approved):

ntBre:
This feels like it could be a breaking change to me (so we could try it in preview first), but I'm curious what others think. We may also need to update the rule docs, which both mention WSL.

Otherwise the change looks reasonable to me, and it's nice to drop a dependency!

cc @MichaReiser who reviewed #17584.

K900:
Oops, missed the doc change. Fixed it now.

MichaReiser:
@K900 would you mind testing the example in https://github.com/astral-sh/ruff/issues/5445


I don't use WSL myself, so I'm not very familiar with how it works but what I understand is that all files on a windows disk `/mnt/...` always have the executable bit set, in which case we'd flag every single file on that disk if they miss a shebang. 

I agree, that we should preview gate this change.

K900:
It's a little more complicated than that on WSL2. Linux permissions are persisted on Windows partitions, as NTFS xattrs, but files without any xattrs, e.g. ones created on the Windows side, are defaulted to 0777. So if you, say, run `git clone` on Windows to clone a repo and then try to access it on WSL, it will have 0777 permissions on everything, but if you `git clone` the same repo to the same path from the WSL side, it will have correct permissions on both sides. This is not entirely obvious behavior, but it will also throw off git (if used from WSL) and basically any other tool, and I don't think Ruff has to try and work around this.

K900:
To clarify:

```
~
❯ cd /mnt/d/test/

/mnt/d/test
❯ ls
╭────────────╮
│ empty list │
╰────────────╯

/mnt/d/test
❯ touch beep

/mnt/d/test
❯ /mnt/c/Windows/System32/cmd.exe /c "echo > boop"  # this escapes back to Windows and creates a file "from Windows"

/mnt/d/test
❯ ls -la
╭────┬──────┬──────┬────────┬──────────┬───────────┬───────────┬────────────────────┬──────┬───────┬──────┬─────────┬─────╮
│  # │ name │ type │ target │ readonly │   mode    │ num_links │       inode        │ user │ group │ size │ created │ ... │
├────┼──────┼──────┼────────┼──────────┼───────────┼───────────┼────────────────────┼──────┼───────┼──────┼─────────┼─────┤
│  0 │ beep │ file │        │ false    │ rw-r--r-- │         1 │ 143552238122452878 │ k900 │ users │  0 B │         │ ... │
│  1 │ boop │ file │        │ false    │ rwxrwxrwx │         1 │ 133137663984158607 │ k900 │ users │ 13 B │         │ ... │
╰────┴──────┴──────┴────────┴──────────┴───────────┴───────────┴────────────────────┴──────┴───────┴──────┴─────────┴─────╯

/mnt/d/test
❯ chmod 0666 boop

/mnt/d/test
❯ ls -la
╭────┬──────┬──────┬────────┬──────────┬───────────┬───────────┬────────────────────┬──────┬───────┬──────┬─────────┬─────╮
│  # │ name │ type │ target │ readonly │   mode    │ num_links │       inode        │ user │ group │ size │ created │ ... │
├────┼──────┼──────┼────────┼──────────┼───────────┼───────────┼────────────────────┼──────┼───────┼──────┼─────────┼─────┤
│  0 │ beep │ file │        │ false    │ rw-r--r-- │         1 │ 143552238122452878 │ k900 │ users │  0 B │         │ ... │
│  1 │ boop │ file │        │ false    │ rw-rw-rw- │         1 │ 133137663984158607 │ k900 │ users │ 13 B │         │ ... │
╰────┴──────┴──────┴────────┴──────────┴───────────┴───────────┴────────────────────┴──────┴───────┴──────┴─────────┴─────╯

**Comment by MichaReiser** [comment]
I'm not yet convinced that we should make this change but we should explain this caveat in the documentation and explain to users how they can "fix their setup" if they run into this situation (clone from within WSL)

**Comment by K900** [comment]
I don't know if I agree, to be honest. This is kind of similar levels of user error as, say, putting stuff on a noexec filesystem, or on a FAT32 or something. There's really no good way to _actually know_ if the filesystem has sane semantics. Maybe some sort of global detection that skips the check if all files come back executable? But even that feels iffy.

**Comment by MichaReiser** [comment]
I'm not sure if you're misunderstanding me. What I'm suggesting is that we make this change in preview, but we also extend the rule documentation with a `WSL` section, explaining in which situations they might see violations for every single file and what they can do to fix it. I don't see what the issue is with helping users in case they see confusing behavior and I would also find this documentation useful for myself, if a user asks for help in the future.

The use case I'm worried about is users who switch between WSL and non-WSL, creating some files in Windows and others in WSL, but flake8-executable is enforced in their organisation. They would have to change their behavior. We could consider recommending that they turn off the rule at a user-level configuration.

**Comment by K900** [comment]
The thing here is that WSL preserves permissions correctly, if set on the WSL side, so just running autofix should repair any sort of WSL setup, _even if the files are then updated from the Windows side_, because xattrs stay. (wait, does this rule even have an autofix? maybe it should if it doesn't yet) However, there are also other kinds of weird setups that are already not caught by this check, and I'm not sure there's a good way to identify those in a general enough way to try and be smart about this particular check.

**Comment by MichaReiser** [comment]
We don't necessarily have to agree here, but I see some documentation giving users a hint on how to solve this problem that users have run into with Ruff in the past, which is a prerequisite to merging.

**Comment by K900** [comment]
Sorry, to be clear, I'm not saying this exact change shouldn't have documentation, I'm kind of thinking out loud about whether there is a way to implement this that doesn't _need_ documentation. What do you think about adding an autofix for this, and then maybe explicitly documenting something like "if the autofix didn't work, your filesystem is probably being weird"?

**Comment by MichaReiser** [comment]
I see. 

I don't see an obvious way for Ruff to detect this, given that it depends on where a file was created. I'm leaning towards making the change you suggestg with better documentation and we can always introduce a `skip_on_wsl` option

**Comment by K900** [comment]
Do you mean just this change, or also adding an auto fix? I can look into doing that in a couple hours.

**Comment by MichaReiser** [comment]
Just this change. Adding an auto fix seems unrelated to this PR

**Comment by K900** [comment]
The autofix is a good way to fix a messed up repo, possibly, if you cloned from Windows and ran ruff from WSL. Also, I wonder how hard it'll be to have git support the WSL xattrs... That would solve most issues before they even happen.

**Comment by K900** [comment]
Added some documentation that is hopefully generic enough to catch most causes of false positives. Sorry for the delay, caught a cold :(

**Comment by K900** [comment]
(and yes, NTFS filesystems mounted to WSL technically _do_ support Unix permissions, but this is a can of worms that is probably out of scope for Ruff rule documentation)

File: crates/ruff_linter/src/rules/flake8_executable/rules/shebang_missing_executable_file.rs [hunk 1]
```diff
@@ -28,7 +28,11 @@ use crate::rules::flake8_executable::helpers::is_executable;
 ///
 /// A file is considered executable if it has the executable bit set (i.e., its
 /// permissions mode intersects with `0o111`). As such, _this rule is only
-/// available on Unix-like systems_, and is not enforced on Windows or WSL.
+/// available on Unix-like systems_.
+///
+/// If this rule produces a significant amount of false positives, it's likely
+/// your source is stored on a filesystem that does not implement Unix file
+/// permissions (e.g. FAT variants, NTFS, CIFS network mounts, VM/WSL shared folders, etc).
```
MichaReiser, line 35: I understood that it's less about where the file is stored but more about how the file was created (WSL side or from within windows). 

Should we also document how users can fix their file system state, e.g. by running chmod from within WSL and setting the file to its default permissions?

MichaReiser:
> Added some documentation that is hopefully generic enough to catch most causes of false positives. Sorry for the delay, caught a cold :(

No need to feel sorry. Thanks for looking into this. I find it very helpful to have someone with deep WSL understanding to talk this change through.

crates/ruff_linter/src/rules/flake8_executable/rules/shebang_missing_executable_file.rs
K900: It's kind of both, because WSL is not the only failure mode here, it's also native Linux drivers for filesystems that don't have the concept of POSIX permission bits (vfat etc). WSL in particular is extra weird because it stores POSIX permissions bits in NTFS xattrs and just assumes 0755 if the xattr isn't set, so it's extra confusing and Windows applications touching those files can also make xattrs be weird and I don't know if trying to explain all of this in a note for this lint is a good idea.

K900: Or at least I really have no idea how to communicate this in a way that's both easy to understand and _actually technically correct_, which is what made me go for the "Just Don't" hint.

MichaReiser: I see. I think my conclusion is we can't make this change and instead need a setting or we risk breaking users that use WSL in combination with such a file system.

K900: It's not just WSL, you can mount a vfat partition on normal Linux, put a repo on it and it'll all be 0755 no matter what, because vfat just makes up the permission bits. I don't know if WSL will even _let you_ access non-NTFS partitions over drvfs?

MusicalNinjaDad:
I hope it's OK to chime in here as the author of #10084 and someone with a lot of experience both using WSL and looking at potential solutions to this issue.

@ntBre, you are correct - this *is* a breaking change and will cause false positives for a subset of users, where the status quo causes false negatives for a different subset of users. (I'll add a table summary to #10084 and copy it here too)

@MichaReiser's original argumentation when discussing this was that false negatives are worse, as settings are mainly on a per-project basis - so it's hard to disable the rule only for some contributors.

@K900 - this is definitely the easiest of "solutions" and I agree with you, and initially argued myself, that it has long been a strong recommendation by MS, that WSL users avoid using mounted windows partitions for performance reasons (see the note at the end of the description on #17584 for example timings)

My personal opinion, FWIW, is that identifying default file permissions is the best: covers all use-cases, actually improves performance. (This is not what I first thought, when beginning to discuss and look at this issue; I actually agreed with the approach proposed in the PR when I first opened the issue).

crates/ruff_linter/src/rules/flake8_executable/rules/shebang_missing_executable_file.rs
MusicalNinjaDad: To be absolutely clear: the underlying problem is purely a **file-system issue** and in **no way inherent to WSL**.

The reason it comes up in relation to WSL is that this is the most common real-life scenario of a linux user storing their code on an FS without *nix permission bits *and* where the user doesn't understand the consequences. (A linux user storing code on an xFAT UBS stick is more likely to understand why they are getting this problem and simply ignore it - but see #12941)

K900:
The problem I have with identifying file permissions is that to do that we need to:

- find a place to write a test file to (we can't always do that)
- be reasonably confident that the filesystem will behave the same way as the file we just wrote (drvfs doesn't)

Edit: to be clear, #17584 does not in fact work for the "WSL, source root on drvfs, cloned from Windows" use case, because it will create a file with 644 permissions, drvfs will record that and give us back the correct permissions, but none of the other files will have the correct xattrs.

MusicalNinjaDad:
Overview of currently proposed solutions:

| | WSL user, WSL FS (#10084) | Linux user USB stick (#12941) | WSL user, Win FS (#3110, #5445) |
| -- | -- | -- | -- |
| Status Quo | :x: (false negatives) | :x: (false positives) | :x: (false negatives) |
| #17548 | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| #21724 | :white_check_mark: | :x: (false positives) | :x: (false positives) |

https://github.com/astral-sh/ruff/issues/10084#issuecomment-3734292000

MusicalNinjaDad:
> The problem I have with identifying file permissions is that to do that we need to:
> 
> * find a place to write a test file to (we can't always do that)
> * be reasonably confident that the filesystem will behave the same way as the file we just wrote (drvfs doesn't)

Did you take a look at the approach in the other PR?

1. check if pryproject.toml is executable (if not, then we can guarantee that we have a settable executable-bit)
2. touch a file in project_root (could fail if project root is read-only, although this is a more acceptable edge case than current status quo and only triggers on FS with no exe-bit)

What specifically do you have in mind by cases where we can't write a test file?

I've not tested drvfs, do you have a test setup to see whether #17584 fails in that case? My understanding is that it is only relevant for WSL1 (as you say, now dead) and the above test would disable the lints at step 1 anyway.

K900:
drvfs is still what WSL2 calls the 9p stuff internally, the case I'm imagining is: user clones repo from Windows, then runs ruff in WSL. What happens then is:

- pyproject.toml is executable, so we continue
- we touch a file in project root, which is 644 by default because umask, that's recorded in NTFS xattrs, reading back the permissions gives us 644
- all the other files are still marked executable because they have no xattrs

MusicalNinjaDad:
> drvfs is still what WSL2 calls the 9p stuff internally, the case I'm imagining is: user clones repo from Windows, then runs ruff in WSL. What happens then is:
> 
> * pyproject.toml is executable, so we continue
> * we touch a file in project root, which is 644 by default because umask, that's recorded in NTFS xattrs, reading back the permissions gives us 644
> * all the other files are still marked executable because they have no xattrs

imagining or tested?

I have tested exactly this setup both locally, and in CI (https://github.com/MusicalNinjaDad/forks-ruff/actions/runs/14663820841/job/41154168824) and in both cases the touched file is executable and the lints are disabled.

If you have a specific setup which leads to the behaviour you describe then I'd be really interested to hear the config in place, both from general curiosity and to help with this case.

(EDIT: just tested this manually, again, on WSL2 with `cd /mnt/c/Users.../Temp; touch foo; ls -al foo` and `foo` has `0777` permissions)

K900:
image

On my end, the file touched from WSL is 0644.

K900:
Also here's a run with ruff from your branch (and a few dbg's sprinkled in):

```
d/test/chinatsu-master
❯ ~/ruff/target/debug/ruff check --no-cache --select EXE
[crates/ruff_linter/src/rules/flake8_executable/helpers.rs:29:9] is_executable(&settings.project_root.join("pyproject.toml")) = Ok(
    true,
)
[crates/ruff_linter/src/rules/flake8_executable/helpers.rs:33:37] is_executable(tmpfile.path()) = Ok(
    false,
)
chinatsu/__init__.py:1:1: EXE002 The file is executable but no shebang is present
(etc)
```

K900:
```diff
diff --git a/crates/ruff_linter/src/rules/flake8_executable/helpers.rs b/crates/ruff_linter/src/rules/flake8_executable/helpers.rs
index 717c805721..47e1e7352d 100644
--- a/crates/ruff_linter/src/rules/flake8_executable/helpers.rs
+++ b/crates/ruff_linter/src/rules/flake8_executable/helpers.rs
@@ -26,11 +26,11 @@ static EXECUTABLE_BY_DEFAULT: OnceLock<bool> = OnceLock::new();
 #[cfg(target_family = "unix")]
 pub(super) fn executable_by_default(settings: &LinterSettings) -> bool {
     *EXECUTABLE_BY_DEFAULT.get_or_init(|| {
-        is_executable(&settings.project_root.join("pyproject.toml")).unwrap_or(true)
+        dbg!(is_executable(&settings.project_root.join("pyproject.toml"))).unwrap_or(true)
             // if pyproject.toml is executable or doesn't exist, run a slower check too:
             && NamedTempFile::new_in(&settings.project_root)
                 .map_err(std::convert::Into::into)
-                .and_then(|tmpfile| is_executable(tmpfile.path()))
+                .and_then(|tmpfile| dbg!(is_executable(tmpfile.path())))
                 .unwrap_or(false) // Assume a normal filesystem in case of read-only, IOError, ...
     })
 }
```

(for posterity)

MusicalNinjaDad:
Thanks for the confirmation.

Did you adjust your `[automount]` settings in `wsl.conf` to enable metadata and set a specific umask? Or are you running with default config and getting these results?

K900:
I have `options=metadata,uid=1000,gid=100`, the uid/gid stuff is just NixOS weirdness for historical reasons.

MusicalNinjaDad:
Thanks - think I've got it now... I see that NixOS [defaults metadata enabled](https://nix-community.github.io/NixOS-WSL/options.html#wslwslconfautomountenabled), unlike the [base WSL default](https://learn.microsoft.com/en-us/windows/wsl/wsl-config?source=recommendations#automount-options), and for various reasons you have chosen/need to keep your code on an NTFS partition rather than the native linux one.

1. I'm assuming from your comments above that you usually clone with linux git. So in your case your system usually behaves like the native WSL case and either of the PRs give you valid lints, status quo gives you potentially false negatives (?)

2. I don't have git for windows installed, but from looking at [your example under `d/test/chaintsu-master`](https://github.com/astral-sh/ruff/pull/21724#issuecomment-3734341168) it appears that cloning to an NTFS partition under windows doesn't add xattr info to maintain the *nix exe-bit. I'm assuming this is not your usual workflow(?)
    - I'd expect that this could get messy quickly, as any future `git pull`s inside WSL will update file permissions for added/changed files that it retrieves from the remote, and any new files will not have 0644 perms. These perms will persist thanks to the metadata setting. **<-- _Could you confirm that easily?_**
    - As you say, there's no decent heuristic available to help in this situation. So the only option would be a command-line-option/environment-variable. I looked into this and dropped it as doing it nicely and well documented affects a lot more code and forces the option to also be exposed via the config files. It's doable but at a higher cost to the maintainers ...
    - There is an argument to say that anyone working this way hopefully understands the consequences of their setup choices and will be in a small minority. Personally, I never like taking that attitude if it can be avoided ...

If I've understood things correctly this means we have the following situation:
| | WSL user, WSL FS (#10084) | Linux user USB stick (#12941) | WSL user, Win FS (#3110, #5445) | WSL user, metadata (@k900) | WSL user, metadata, cloned on Win |
| -- | -- | -- | -- | -- | -- |
| Status Quo | :exclamation: (false negatives) | :x: (false positives) | :exclamation: (false negatives) | :exclamation: (false negatives) | :exclamation: (false negatives) |
| #17548 | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :exclamation: (false negatives) |
| #21724 | :white_check_mark: | :x: (false positives) | :x: (false positives) | :white_check_mark: | :exclamation: (false negatives) |

K900:
To be clear, this is not my actual setup that I actually use, it's a setup that I assume people ran into which led to the initial change. However, it's also yet another proof that guessing "is the filesystem broken" in the general case is impossible (even ignoring all kinds of obviously-pathological things like FileStation 9000 which returns random permission bits on every read).

My normal usage of Ruff doesn't actually trigger this bug - I've run into it because I occasionally build NixOS from source on my WSL systems, and Ruff fails _in its own tests_ because those are not designed to run under WSL.

Also, Git knows nothing about the NTFS Unix-permission xattrs, so it never writes or updates them for any reason, meaning newly added files (if pulled from Windows) will be 0755, and any updated files will keep their existing permissions (unless moved, possibly?). Teaching Git about these xattrs is another thing that could probably improve things a lot, but not my can and not my worms.

MusicalNinjaDad:
Thanks. Yeah, no heuristic will be perfect for this. I'd see the "WSL user, metadata enabled, cloned on Win" as an unusual edge case which requires someone to actively adjust settings from defaults AND work against recommended practices, with consequences far beyond these lints - so reasonable to assume "they know what they're doing".

Could you confirm the below table matches your experience and I'll update the issue to reflect it:

| | WSL user, WSL FS (official MS recommendation) (#10084) | WSL user, NTFS metadata enabled (e.g. NixOs) [see #21724](https://github.com/astral-sh/ruff/pull/21724#issuecomment-3734350142) | Linux user exFAT USB stick (#12941) | WSL user, Win FS (#3110, #5445) |
| -- | -- | -- | -- | -- |
| Status Quo | :exclamation: (disabled) | :exclamation: (disabled) | :x: | :exclamation: (disabled) |
| #17548 | :white_check_mark: | :white_check_mark: | :exclamation: (disabled) | :exclamation: (disabled) |
| #21724 | :white_check_mark: | :white_check_mark: | :x: | :x: |

K900:
This seems accurate yeah. I hate how the table just keeps growing...

K900:
This seems accurate yeah. I hate how the table just keeps growing...

K900:
So folks, should we make some kind of call here? I'm still leaning towards just removing the check and documenting it better that cursed filesystem setups are the end user's problem.

MichaReiser:
> So folks, should we make some kind of call here? I'm still leaning towards just removing the check and documenting it better that cursed filesystem setups are the end user's problem.

I don't feel a strong need to make a change here, especially since this has a very high chance of breaking someone's setup. 

I'd be open to adding a config option that allows opting in on WSL, and extending the documentaiton

K900:
The other issue, and the one that actually led me to discovering this behavior, is that running `cargo test` on Ruff itself breaks on WSL, because it makes this assumption and the tests aren't designed for it. This could be fixed by skipping the tests on WSL, I guess?

MichaReiser:
> The other issue, and the one that actually led me to discovering this behavior, is that running cargo test on Ruff itself breaks on WSL, because it makes this assumption and the tests aren't designed for it. This could be fixed by skipping the tests on WSL, I guess?

Yes, I consider this to be a separate issue from changing the rule's behavior. Skipping it on WSL seems fine and something we can do in a seperate PR

K900:
OK, so here's skipping the tests, for now: https://github.com/astral-sh/ruff/pull/22721

PR #21726, opened 2025-12-01
[ty] Exhaustiveness checking for generic classes

sharkdp:
## Summary

We had tests for this already, but they used generic classes that were bivariant in their type parameter, and so this case wasn't captured.

closes https://github.com/astral-sh/ty/issues/1702

## Test Plan

Updated Markdown tests


crates/ty_python_semantic/resources/mdtest/exhaustiveness_checking.md
```diff
@@ -284,10 +284,17 @@ python-version = "3.12"
 ```py
 from typing import assert_never
 
-class A[T]: ...
```
sharkdp, line 287: bivariance strikes again

PR #21728, opened 2025-12-01
[ty] Remove `GenericAlias`-related todo type

sharkdp:
## Summary

If you manage to create an `typing.GenericAlias` instance without us knowing how that was created, then we don't know what to do with this in a type annotation. So it's better to be explicit and show an error instead of failing silently with a `@Todo` type.

## Test Plan

* New Markdown tests
* Zero ecosystem impact

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/annotations/generic_alias.md
+++ b/crates/ty_python_semantic/resources/mdtest/annotations/generic_alias.md
@@ -1,0 +1,34 @@
+# GenericAlias in type expressions
+
+We recognize if a `types.GenericAlias` instance is created by specializing a generic class. We don't
+explicitly mention it in our type display, but `list[int]` in the example below is a `GenericAlias`
+instance at runtime:
+
+```py
+Numbers = list[int]
+
+# At runtime, `Numbers` is an instance of `types.GenericAlias`. Showing
+# this as `list[int]` is more helpful, though:
+reveal_type(Numbers)  # revealed: <class 'list[int]'>
+
+def _(numbers: Numbers) -> None:
+    reveal_type(numbers)  # revealed: list[int]
+```
+
+It is also valid to create `GenericAlias` instances manually:
+
+```py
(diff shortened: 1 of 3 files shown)


crates/ty_python_semantic/resources/mdtest/annotations/generic_alias.md
```diff
@@ -0,0 +1,34 @@
...
+
+```py
+from types import GenericAlias
+
+Strings = GenericAlias(list, (str,))
+
+reveal_type(Strings)  # revealed: GenericAlias
+```
+
+However, using such a `GenericAlias` instance in a type expression is currently not supported:
```
sharkdp, line 28: No other type checker supports this. They don't emit an error, but their revealed type for `strings` (`GenericAlias`) seems… strange.

crates/ty_python_semantic/resources/mdtest/annotations/new_types.md
```diff
@@ -15,10 +15,11 @@ B = GenericAlias(A, ())
```
AlexWaygood: I'm inclined to also get rid of the TODO a couple of lines above this. I don't think it's something we actually want to do 😆

sharkdp: I don't understand this test at all, especially why it's so prominently at the top of this file :smile:

AlexWaygood: yeah, maybe just delete the `GenericAlias` bits from this test? The test was originally added in https://github.com/astral-sh/ruff/commit/63e78b41cdcb6e5469998fe4fba97260b3a3e95c (as an incidental change that was part of a bigger PR, long before we supported `NewType` properly), and the TODO was added in https://github.com/astral-sh/ruff/commit/fab7d820bd6c8ea01f4fbd5ad8f92f3d333e5506 (also an incidental change that was part of a bigger PR)

PR #21730, opened 2025-12-01
[ty] Support typevar-specialized dynamic types in generic type aliases

sharkdp:
## Summary

For a type alias like the one below, where `UnknownClass` is something with a dynamic type, we previously lost track of the fact that this dynamic type was explicitly specialized *with a type variable*. If that alias is then later explicitly specialized itself (`MyAlias[int]`), we would miscount the number of legacy type variables and emit a `invalid-type-arguments` diagnostic ([playground](https://play.ty.dev/886ae6cc-86c3-4304-a365-510d29211f85)).
```py
T = TypeVar("T")

MyAlias: TypeAlias = UnknownClass[T] | None
```
The solution implemented here is not pretty, but we can hopefully get rid of it via https://github.com/astral-sh/ty/issues/1711. Also, once we properly support `ParamSpec` and `Concatenate`, we should be able to remove some of this code.

This addresses many of the `invalid-type-arguments` false-positives in https://github.com/astral-sh/ty/issues/1685. With this change, there are still some diagnostics of this type left. Instead of implementing even more (rather sophisticated) workarounds for these cases as well, it might be much easier to wait for full `ParamSpec`/`Concatenate` support and then try again.

A disadvantage of this implementation is that we lose track of some `@Todo` types and replace them with `Unknown`. We could spend more effort and try to preserve them, but I'm unsure if this is the best use of our time right now.

## Test Plan

New Markdown tests.

### DIFF
--- a/crates/ty_python_semantic/src/types/infer/builder.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder.rs
@@ -11765,6 +11765,7 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
             }
 
             (Type::SpecialForm(SpecialFormType::Concatenate), _) => {
+                // TODO: Add proper support for `Concatenate`
                 let mut variables = FxOrderSet::default();
                 slice_ty.bind_and_find_all_legacy_typevars(
                     db,
--- a/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
+++ b/crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
@@ -946,6 +946,7 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
                     }
                 }
                 KnownInstanceType::TypeAliasType(TypeAliasType::ManualPEP695(_)) => {
+                    // TODO: support generic "manual" PEP 695 type aliases
                     let slice_ty = self.infer_expression(slice, TypeContext::default());
                     let mut variables = FxOrderSet::default();
                     slice_ty.bind_and_find_all_legacy_typevars(


carljm:
The conformance suite results here look fine. Most of the changes are around `ParamSpec`, in the direction of issuing fewer diagnostics on our failure to understand them, which makes sense since they are currently a Todo type so end up being a case of "explicitly specializing a dynamic type". In some cases we stop erroring on lines we should error on, but we were previously erroring for the wrong reason, so that's still good.

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
```diff
@@ -149,6 +149,101 @@ def _(x: MyAlias):
...
+
+def _(c: MyAlias6[str, bytes]):
+    # TODO: should be (dict[str, bytes], /) -> int
+    reveal_type(c)  # revealed: (Unknown, /) -> int
+
+ListOrDict: TypeAlias = MyList[T] | dict[str, T]
+
+def _(x: ListOrDict[int]):
+    # TODO: should be list[int] | dict[str, int]
+    reveal_type(x)  # revealed: Unknown | dict[str, int]
```
carljm, line 238: What's our barrier to getting these three TODOs right? I guess it's just that we currently don't support generics at all on "manual" PEP 695 type aliases?

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
```diff
@@ -944,8 +944,16 @@ impl<'db> TypeInferenceBuilder<'db, '_> {
...
+                    let slice_ty = self.infer_expression(slice, TypeContext::default());
+                    let mut variables = FxOrderSet::default();
+                    slice_ty.bind_and_find_all_legacy_typevars(
+                        self.db(),
+                        self.typevar_binding_context,
+                        &mut variables,
+                    );
+                    let generic_context =
+                        GenericContext::from_typevar_instances(self.db(), variables);
+                    Type::Dynamic(DynamicType::UnknownGeneric(generic_context))
```
carljm, line 959: There should be a TODO here, yeah?

And we should probably file a follow-up issue for supporting generics in "manual" PEP 695 type aliases?

crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -11707,6 +11775,17 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
 
+            (Type::SpecialForm(SpecialFormType::Concatenate), _) => {
+                let mut variables = FxOrderSet::default();
+                slice_ty.bind_and_find_all_legacy_typevars(
+                    db,
+                    self.typevar_binding_context,
+                    &mut variables,
+                );
+                let generic_context = GenericContext::from_typevar_instances(self.db(), variables);
+                Some(Type::Dynamic(DynamicType::UnknownGeneric(generic_context)))
```
carljm, line 11776: And we should probably have a TODO comment somewhere in here, too...

crates/ty_python_semantic/src/types/bound_super.rs
```diff
@@ -172,7 +172,7 @@ impl<'db> BoundSuperError<'db> {
 
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, get_size2::GetSize)]
 pub enum SuperOwnerKind<'db> {
-    Dynamic(DynamicType),
+    Dynamic(DynamicType<'db>),
```
carljm, line 175: Heh, this lifetime was just removed a week or so ago...

crates/ty_python_semantic/resources/mdtest/pep613_type_aliases.md
sharkdp: Yes, exactly. I considered this to be a (maybe lower-priority?) task separate from the generic implicit/PEP613 type alias work. I opened https://github.com/astral-sh/ty/issues/1737.

crates/ty_python_semantic/src/types/infer/builder/type_expression.rs
sharkdp: > There should be a TODO here, yeah?

Yes, I think in my mind, the whole `DynamicType::UnknownGeneric` variant was a sort of "Todo" marker, but I added an explicit TODO comment now :+1: 

> And we should probably file a follow-up issue for supporting generics in "manual" PEP 695 type aliases?

https://github.com/astral-sh/ty/issues/1737

sharkdp:
> The conformance suite results here look fine.

Thank you for checking, I was so focused on the ecosystem impact (which was erratically changing with every push), I forgot to take a look at the conformance results.

PR #21731, opened 2025-12-01
[ty] Exclude `typing_extensions` from completions unless it's really available

BurntSushi:
This works by adding a third module resolution mode that lets the caller
opt into _some_ shadowing of modules that is otherwise not allowed (for
`typing` and `typing_extensions`).

Fixes astral-sh/ty#1658

### DIFF
--- a/crates/ty_ide/src/all_symbols.rs
+++ b/crates/ty_ide/src/all_symbols.rs
@@ -1,6 +1,6 @@
 use ruff_db::files::File;
 use ty_project::Db;
-use ty_python_semantic::{Module, all_modules};
+use ty_python_semantic::{Module, ModuleName, all_modules, resolve_real_shadowable_module};
 
 use crate::symbols::{QueryPattern, SymbolInfo, symbols_for_file_global_only};
 
@@ -8,11 +8,19 @@ use crate::symbols::{QueryPattern, SymbolInfo, symbols_for_file_global_only};
 ///
 /// Returns symbols from all files in the workspace and dependencies, filtered
 /// by the query.
-pub fn all_symbols<'db>(db: &'db dyn Db, query: &QueryPattern) -> Vec<AllSymbolInfo<'db>> {
+pub fn all_symbols<'db>(
+    db: &'db dyn Db,
+    importing_from: File,
+    query: &QueryPattern,
+) -> Vec<AllSymbolInfo<'db>> {
     // If the query is empty, return immediately to avoid expensive file scanning
     if query.will_match_everything() {
         return Vec::new();
     }
+
+    let typing_extensions = ModuleName::new("typing_extensions").unwrap();
+    let is_typing_extensions_available = importing_from.is_stub(db)
+        || resolve_real_shadowable_module(db, &typing_extensions).is_some();
 
     let results = std::sync::Mutex::new(Vec::new());
     {
@@ -28,6 +36,11 @@ pub fn all_symbols<'db>(db: &'db dyn Db, query: &QueryPattern) -> Vec<AllSymbolI
                 let Some(file) = module.file(&*db) else {
                     continue;
                 };
+                // TODO: also make it available in `TYPE_CHECKING` blocks
+                // (we'd need https://github.com/astral-sh/ty/issues/1553 to do this well)
+                if !is_typing_extensions_available && module.name(&*db) == &typing_extensions {
+                    continue;
+                }
                 s.spawn(move |_| {
                     for (_, symbol) in symbols_for_file_global_only(&*db, file).search(query) {
                         // It seems like we could do better here than
@@ -143,7 +156,7 @@ ABCDEFGHIJKLMNOP = 'https://api.example.com'
 
     impl CursorTest {
         fn all_symbols(&self, query: &str) -> String {
-            let symbols = all_symbols(&self.db, &QueryPattern::fuzzy(query));
+            let symbols = all_symbols(&self.db, self.cursor.file, &QueryPattern::fuzzy(query));
 
             if symbols.is_empty() {
                 return "No symbols found".to_string();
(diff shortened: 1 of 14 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/a36545ce-b71b-42f7-9f42-df2535e00c94

crates/ty_ide/src/completion.rs
AlexWaygood, line 1: Ideally I guess we'd also include some tests where we mock out a virtual environment that _does_ include `typing_extensions`, and then check that the `typing_extensions` suggestions are still included. But that sounds a bit involved 🙃

crates/ty_ide/src/all_symbols.rs
```diff
@@ -28,6 +32,9 @@ pub fn all_symbols<'db>(db: &'db dyn Db, query: &QueryPattern) -> Vec<AllSymbolI
                 let Some(file) = module.file(&*db) else {
                     continue;
                 };
+                if !is_typing_extensions_available && module.name(&*db) == &typing_extensions {
+                    continue;
+                }
```
AlexWaygood, line 43: I think we should still include `typing_extensions` suggestions if it's a stub file, even if `typing_extensions` isn't installed into the environment. Stub files are never executed at runtime, so the typeshed version of `typing_extensions` is fine for them.

Ideally we would also include `typing_extensions` suggestions inside `if TYPE_CHECKING` blocks (which are also not executed at runtime), but that won't produce great results until/unless we implement https://github.com/astral-sh/ty/issues/1553, so I wouldn't worry about it too much right now.

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -47,8 +47,33 @@ pub fn resolve_real_module<'db>(db: &'db dyn Db, module_name: &ModuleName) -> Op
     resolve_module_query(db, interned_name)
 }
 
+/// Resolves a module name to a module (stubs not allowed, some shadowing is
+/// allowed).
+///
+/// In particular, this allows `types` and `typing_extensions` to be
```
AlexWaygood: hmm, `types` _is_ actually a stdlib module, so it's a bit different from `typing_extensions`, which is a "fake typeshed stdlib module"

crates/ty_python_semantic/src/module_resolver/resolver.rs
```diff
@@ -61,6 +86,13 @@ pub(crate) enum ModuleResolveMode {
     /// implementations. When querying searchpaths this also notably replaces typeshed with
     /// the "real" stdlib.
     StubsNotAllowed,
+    /// Like `StubsNotAllowed`, but permits some modules to be shadowed.
+    ///
+    /// In particular, this allows `types` and `typing_extensions` to be
```
AlexWaygood: same here, I'd remove the mention of `types`; it's not really relevant. That one really _is_ always available at runtime!

AlexWaygood (approved):

crates/ty_python_semantic/src/module_resolver/resolver.rs
BurntSushi: Hmmm, I was mostly just leaving what was there before. Or should `types` _always_ be considered non-shadowable?

AlexWaygood: I think `types` _probably_ should _always_ be considered non-shadowable, yes, because it is imported as part of the Python interpreter's startup process so it's very difficult to shadow it in first-party code (even though theoretically possible).

But the main thing I was actually objecting to was the comment! You say "we don't want to pretend as if these modules are always available at runtime" -- but there's no pretence with the `types` module. It just _is_ always available at runtime!

Gankra: I believe all actual stdlib modules (which includes `types`) are "always non-shadowable", yeah.

AlexWaygood: > I believe all actual stdlib modules (which includes `types`) are "always non-shadowable", yeah.

no, that's not true -- most stdlib modules can be shadowed by user code. The first-party search paths have precedence over the stdlib search path (emulating the fact that `.` is the first entry on `sys.path` at runtime).

But certain "builtin" modules have special-casing baked deep into the Python interpreter, and these modules skip normal module resolution. These "builtin" modules, a subset of the Python stdlib, must _always_ be resolved to locations in the standard library. Ty emulates this, because it accurately reflects what happens at runtime, and also adds a few other modules which aren't _actually_ builtin at runtime but where we're very likely to crash if the module is resolved to a non-stdlib location (`types`, `typing_extensions`)

BurntSushi: Got it. Comments fixed and I made `types` always non-shadowable. So only `typing_extensions` is the special snowflake here.

crates/ty_ide/src/all_symbols.rs
BurntSushi: Ah good call. That makes sense. Fixed.

crates/ty_ide/src/completion.rs
BurntSushi: Yeah we do at least have tests here with a `typing_extensions.py` in a non-std search path. So I think that covers the most critical case (which was failing before adding the shadowable configuration).

crates/ty_ide/src/all_symbols.rs
```diff
@@ -1,19 +1,27 @@
...
+    query: &QueryPattern,
+) -> Vec<AllSymbolInfo<'db>> {
     // If the query is empty, return immediately to avoid expensive file scanning
     if query.will_match_everything() {
         return Vec::new();
     }
 
+    let typing_extensions = ModuleName::new("typing_extensions").unwrap();
+    let is_typing_extensions_available = importing_from.is_stub(db)
+        || resolve_real_shadowable_module(db, &typing_extensions).is_some();
```
AlexWaygood, line 23:
```suggestion
    // TODO: also make it available in `TYPE_CHECKING` blocks
    // (we'd need https://github.com/astral-sh/ty/issues/1553 to do this well)
    let is_typing_extensions_available = importing_from.is_stub(db)
        || resolve_real_shadowable_module(db, &typing_extensions).is_some();
```

PR #21732, opened 2025-12-01
[ty] Fix find references for type defined in stub

MichaReiser:
## Summary

ty failed to return any references for a type defined in a stub where
there's also a definition in a non-stub file.

The underlying issue was that `find_references` resolved the name to
"go-to definition" whereas the visitor resolved the declaration, which
can resolve to a different file and location.

This fix uses go to declaration in both steps. I don't think it really
matters whether we use go to declaration or go to definition because
both should resolve to the same references. I picked go to declaration
because it requires less work (no stub mapping)

## Test Plan

Added test

### DIFF
--- a/crates/ty_ide/src/find_references.rs
+++ b/crates/ty_ide/src/find_references.rs
@@ -1656,4 +1656,70 @@ func<CURSOR>_alias()
           |
         ");
     }
+
+    #[test]
+    fn stub_target() {
+        let test = CursorTest::builder()
+            .source(
+                "path.pyi",
+                r#"
+                class Path:
+                    def __init__(self, path: str): ...
+            "#,
+            )
+            .source(
+                "path.py",
+                r#"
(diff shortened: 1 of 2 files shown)


Gankra:
Hmm interesting, ideally we would resolve both..? Especially if you're in a codebase that maintains the `.py` and `.pyi`. I wonder what different LSPs do...

MichaReiser:
> Hmm interesting, ideally we would resolve both..? Especially if you're in a codebase that maintains the `.py` and `.pyi`. I wonder what different LSPs do...

We only use the first navigation target to test if they're the same. 

As far as I could tell, ty now shows the same references when finding references to `Path` in `ty_benchmark` as Pylance

MichaReiser:
I'm certainly not saying this is the final solution. It's just one of the first issues I found while trying to wrap my head around how find references works

PR #21733, opened 2025-12-01
[ty] Fix auto-import code action to handle pre-existing import

BurntSushi:
Previously, the code action to do auto-import on a pre-existing symbol
assumed that the auto-importer would always generate an import
statement. But sometimes an import statement already exists.

A good example of this is the following snippet:

```py
import warnings

@deprecated
def myfunc(): pass
```

Specifically, `deprecated` exists in `warnings` but isn't currently
imported. A code action to fix this could feasibly do two
transformations here. One is:

```py
import warnings

@warnings.deprecated
def myfunc(): pass
```

Another is:

```py
from warnings import deprecated
import warnings

@deprecated
def myfunc(): pass
```

The existing auto-import infrastructure chooses the former, since it
reuses a pre-existing import statement. But this PR chooses the latter
for the case of a code action. I'm not 100% sure this is the correct
choice, but it seems to defer more strongly to what the user has typed.
That is, that they want to use it unqualified because it's what has been
typed. So we should add the necessary import statement to make that
work.

Fixes astral-sh/ty#1668

### DIFF
--- a/crates/ty_ide/src/completion.rs
+++ b/crates/ty_ide/src/completion.rs
@@ -417,7 +417,16 @@ pub fn completion<'db>(
         }
         if settings.auto_import {
             if let Some(scoped) = scoped {
-                add_unimported_completions(db, file, &parsed, scoped, &mut completions);
+                add_unimported_completions(
+                    db,
+                    file,
+                    &parsed,
+                    scoped,
+                    |module_name: &ModuleName, symbol: &str| {
+                        ImportRequest::import_from(module_name.as_str(), symbol)
+                    },
+                    &mut completions,
+                );
             }
         }
     }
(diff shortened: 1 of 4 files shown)


BurntSushi:
Demo:

https://github.com/user-attachments/assets/90294a57-3a8b-49f2-9325-af6885d2339c

crates/ty_ide/src/completion.rs
```diff
@@ -502,6 +520,7 @@ fn add_unimported_completions<'db>(
     file: File,
     parsed: &ParsedModuleRef,
     scoped: ScopedTarget<'_>,
+    create_import_request: impl for<'a> Fn(&'a ModuleName, &'a str) -> ImportRequest<'a>,
```
BurntSushi, line 523: It's always a good day when one gets to write a higher rank trait bound.

MichaReiser: It seems the only difference is `force`. Are you expecting that we'll need more arguments or flexibility around customizing in the future or could we pass a `Style` enum or similar instead of the factory function?

MichaReiser (approved):

crates/ty_ide/src/completion.rs
BurntSushi: I'm honestly not sure. But I'd rather this than creating another type for now. This is also an API internal to this module, so I don't feel too bad about it.

Gankra:
>  I'm not 100% sure this is the correct choice, but it seems to defer more strongly to what the user has typed.

I believe rust-analyzer just offers up both choices. This is much more palatable in the context of quick-fixes where you're likely to get, like, 4 options instead of 1000. I'd like us to do that in the future (this PR gets us closer to that).

(Actually I believe rust-analyzer does a two-phase thing where you pick "qualify" vs "import" and then if there are multiple possible imports/qualifications it brings up a second prompt for which you want.)

PR #21737, opened 2025-12-01
[ty] Improve diagnostics for unsupported comparison operations

AlexWaygood:
## Summary

On `main`:

image

Concise diagnostics on `main`:

```
foo.py:1:1: error[unsupported-operator] Operator `<` is not supported for types `str` and `int`, in comparing `tuple[Literal[1], Literal["foo"]]` with `tuple[Literal[1], Literal[2]]`
foo.py:2:1: error[unsupported-operator] Operator `<` is not supported for types `object` and `object`, in comparing `tuple[object]` with `tuple[object]`
```

With PR:

image

Concise diagnostics with PR:

```
foo.py:1:1: error[unsupported-operator] Operator `<` is not supported between objects of type `tuple[Literal[1], Literal["foo"]]` and `tuple[Literal[1], Literal[2]]`
foo.py:2:1: error[unsupported-operator] Operator `<` is not supported between two objects of type `tuple[object]`
```

## Test Plan

Mdtests and snapshots

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
+++ b/crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
@@ -45,53 +45,35 @@ mdtest path: crates/ty_python_semantic/resources/mdtest/comparison/intersections
 # Diagnostics
 
 ```
-error[unsupported-operator]: Unsupported comparison operation
+error[unsupported-operator]: Unsupported `in` operation
   --> src/mdtest_snippet.py:10:25
    |
  9 |             # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1 & …
 10 |             reveal_type(2 in x)  # revealed: bool
-   |                         ^^^^^^ `in` is not supported here
-11 | class Container:
-12 |     def __contains__(self, x) -> bool:
-   |
-info: Operand types
-  --> src/mdtest_snippet.py:10:25
-   |
- 9 |             # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1 & …
-10 |             reveal_type(2 in x)  # revealed: bool
-   |                         -    - Has type `NonContainer1 & NonContainer2`
-   |                         |
+   |                         -^^^^-
+   |                         |    |
+   |                         |    Has type `NonContainer1 & NonContainer2`
    |                         Has type `Literal[2]`
 11 | class Container:
 12 |     def __contains__(self, x) -> bool:
    |
-info: Operation fails because operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1`
 info: rule `unsupported-operator` is enabled by default
 
 ```
 
 ```
-error[unsupported-operator]: Unsupported comparison operation
+error[unsupported-operator]: Unsupported `in` operation
   --> src/mdtest_snippet.py:26:21
    |
 25 |         # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `~NonContainer1`"
 26 |         reveal_type(2 in x)  # revealed: bool
-   |                     ^^^^^^ `in` is not supported here
-27 |
-28 |         reveal_type(2 is x)  # revealed: bool
-   |
-info: Operand types
-  --> src/mdtest_snippet.py:26:21
-   |
-25 |         # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `~NonContainer1`"
-26 |         reveal_type(2 in x)  # revealed: bool
-   |                     -    - Has type `~NonContainer1`
-   |                     |
+   |                     -^^^^-
+   |                     |    |
+   |                     |    Has type `~NonContainer1`
    |                     Has type `Literal[2]`
 27 |
 28 |         reveal_type(2 is x)  # revealed: bool
    |
-info: Operation fails because operator `in` is not supported between objects of type `Literal[2]` and `object`
 info: rule `unsupported-operator` is enabled by default
 
 ```
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -25,6 +25,7 @@ use crate::types::string_annotation::{
     IMPLICIT_CONCATENATED_STRING_TYPE_ANNOTATION, INVALID_SYNTAX_IN_FORWARD_ANNOTATION,
     RAW_STRING_TYPE_ANNOTATION,
 };
+use crate::types::tuple::TupleSpec;
 use crate::types::{
     BoundTypeVarInstance, ClassType, DynamicType, LintDiagnosticGuard, Protocol,
     ProtocolInstanceType, SpecialFormType, SubclassOfInner, Type, TypeContext, binding_type,
@@ -4075,44 +4076,35 @@ pub(super) fn report_unsupported_comparison<'db>(
         [error.left_ty, error.right_ty, left_ty, right_ty],
     );
 
-    let mut diagnostic = if left_ty == right_ty {
-        let mut diag = diagnostic_builder
-            .into_diagnostic(format_args!("Unsupported `{}` operation", error.op));
-        diag.set_primary_message(format_args!(
+    let mut diagnostic =
+        diagnostic_builder.into_diagnostic(format_args!("Unsupported `{}` operation", error.op));
+
+    if left_ty == right_ty {
+        diagnostic.set_primary_message(format_args!(
             "Both operands have type `{}`",
             left_ty.display_with(db, display_settings.clone())
         ));
-        diag.annotate(context.secondary(left));
-        diag.annotate(context.secondary(right));
-        diag.set_concise_message(format_args!(
+        diagnostic.annotate(context.secondary(left));
+        diagnostic.annotate(context.secondary(right));
+        diagnostic.set_concise_message(format_args!(
             "Operator `{}` is not supported between two objects of type `{}`",
             error.op,
             left_ty.display_with(db, display_settings.clone())
         ));
-        diag
     } else {
-        let mut diag = diagnostic_builder.into_diagnostic("Unsupported comparison operation");
-
-        diag.set_primary_message(format_args!("`{}` is not supported here", error.op));
-
-        let mut sub = SubDiagnostic::new(SubDiagnosticSeverity::Info, "Operand types");
         for (ty, expr) in [(left_ty, left), (right_ty, right)] {
-            sub.annotate(context.secondary(expr).message(format_args!(
+            diagnostic.annotate(context.secondary(expr).message(format_args!(
                 "Has type `{}`",
                 ty.display_with(db, display_settings.clone())
             )));
         }
-        diag.sub(sub);
-
-        diag.set_concise_message(format_args!(
+        diagnostic.set_concise_message(format_args!(
             "Operator `{}` is not supported between objects of type `{}` and `{}`",
             error.op,
             left_ty.display_with(db, display_settings.clone()),
             right_ty.display_with(db, display_settings.clone())
         ));
-
-        diag
-    };
+    }
 
     // For non-atomic types like unions and tuples, we now provide context
     // on the underlying elements that caused the error.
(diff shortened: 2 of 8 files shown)


crates/ty_python_semantic/resources/mdtest/bidirectional.md
```diff
@@ -340,7 +340,7 @@ def _(a: object, b: object, flag: bool):
     else:
         x = g
 
-    # error: [unsupported-operator] "Operator `>` is not supported for types `object` and `object`"
+    # error: [unsupported-operator] "Operator `>` is not supported between two objects of type `object`"
```
MichaReiser, line 343: ❤️

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
```diff
@@ -0,0 +1,97 @@
...
+
+# Diagnostics
+
+```
+error[unsupported-operator]: Unsupported `in` operation
+  --> src/mdtest_snippet.py:10:25
+   |
+ 9 |             # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1 & …
+10 |             reveal_type(2 in x)  # revealed: bool
+   |                         ^^^^^^ Comparison may raise an exception
```
MichaReiser: I'd remove this message. This is true for almost all ty diagnostics and it doesn't provide any additional information

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
```diff
@@ -0,0 +1,97 @@
...
+12 |     def __contains__(self, x) -> bool:
+   |
+info: Operand types
+  --> src/mdtest_snippet.py:10:25
+   |
+ 9 |             # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1 & …
+10 |             reveal_type(2 in x)  # revealed: bool
+   |                         -    - Right-hand side has type `NonContainer1 & NonContainer2`
+   |                         |
+   |                         Left-hand side has type `Literal[2]`
```
MichaReiser: I would remove the `Right-hand` side and `Left-hand side`. This is viusally appearant and we also don't refer to the left-hand or right-hand side in the rest of the diagnostic message.
```suggestion
   |                         -    - has type `NonContainer1 & NonContainer2`
   |                         |
   |                         has type `Literal[2]`
```

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
```diff
@@ -0,0 +1,97 @@
...
+  --> src/mdtest_snippet.py:10:25
+   |
+ 9 |             # error: [unsupported-operator] "Operator `in` is not supported between objects of type `Literal[2]` and `NonContainer1 & …
+10 |             reveal_type(2 in x)  # revealed: bool
+   |                         ^^^^^^ Comparison may raise an exception
+11 | class Container:
+12 |     def __contains__(self, x) -> bool:
+   |
+info: Operand types
+  --> src/mdtest_snippet.py:10:25
```
MichaReiser: Do we need to add the `info` message here? Can't we add a secondary annotation to the main diagnostic? I'm asking because I don't find `Operand types` a very helpful info message

crates/ty_python_semantic/src/types/diagnostic.rs
```diff
@@ -4032,6 +4033,83 @@ pub(super) fn report_overridden_final_method<'db>(
...
+    } else {
+        diagnostic.set_concise_message(format_args!(
+            "Operator `{}` is not supported between objects of type `{}` and `{}`",
+            error.op,
+            left_ty.display_with(db, display_settings.clone()),
+            right_ty.display_with(db, display_settings.clone())
+        ));
+    }
+
+    if (error.left_ty, error.right_ty) != (left_ty, right_ty) {
```
MichaReiser, line 4117: How's this comparison different from the one on line 4078?

MichaReiser (commented):
Nice, I mainly reviewed the diagnostics and I think we should make them more concise

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
AlexWaygood: I tried it with the secondary-annotation approach initially but found the overlapping ranges very confusing to look at (the primary range covers the whole expression — correctly, in my view)

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
AlexWaygood: Fair

crates/ty_python_semantic/src/types/diagnostic.rs
AlexWaygood: I'll add a comment

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
MichaReiser: Hmm. Repeating the same code frame adds a lot of redundancy (beyond the message itself, like each diagnostic repeating the code twice). 

But I can see how overlapping ranges can look weird. The only other solution I see is to put the operand types into the primary message. Unless @BurntSushi has an idea

AlexWaygood: > The only other solution I see is to put the operand types into the primary message.

Right, but it can make for an extremely long primary message. Which was exactly what this PR was trying to fix :-)

AlexWaygood: If we had the ability to optionally specify a separate primary-highlight range that is different from the primary suppression range (which we've discussed before in various places, e.g. https://github.com/astral-sh/ruff/pull/18050#discussion_r2085275592, though I don't think we have an issue for it yet), then I would have the primary-highlight range cover only the operator in between the operands. But as it is, I'm pretty sure that would do chaotic things to people's suppression comments for this error code

MichaReiser: > Right, but it can make for an extremely long primary message. Which was exactly what this PR was trying to fix :-)

Not the diagnostic message, move it into the annotation. It might still be long, though. 

The alternative is to have to `info` messages, but I worry that this buries important information.

> If we had the ability to optionally specify a separate primary-highlight range that is different from the primary suppression range (which we've discussed before in various places, 

The case here seems different enough to me where I'd say a separate suppression range isn't the solution. In most instances where we wanted that feature is to allow a wider suppression range but without necessarily changing the line number on which the diagnostic gets reported. Here, we don't want to change the suppression range OR the line number of the diagnostic, we just want more sane rendering of overlapping ranges 😆

AlexWaygood: > Not the diagnostic message, move it into the annotation. It might still be long, though.

Yeah, this actually makes the "long sentence" problem even worse because if it's the primary annotation message then now the whole sentence is indented significantly before it even starts

MichaReiser: I think I prefer the old rendering then, given that it also matches the concise rendering. It simply presents the same information with less repetition

AlexWaygood: > The case here seems different enough to me where I'd say a separate suppression range isn't the solution. In most instances where we wanted that feature is to allow a wider suppression range but without necessarily changing the line number on which the diagnostic gets reported. Here, we don't want to change the suppression range OR the line number of the diagnostic, we just want more sane rendering of overlapping ranges 😆

If I changed the primary range of the diagnostic to be just the operator (so that it doesn't overlap with the ranges of the secondary annotations) then we _would_ want to change the suppression range. That feels exactly the same as the situation discussed in https://github.com/astral-sh/ruff/pull/18050#discussion_r2085275592, where we wanted to make the primary range of the diagnostic smaller for rendering purposes but we held off because we didn't have a way to separately specify a larger suppression range

MichaReiser: Oh I see. I'm not sure if I would do that. It can be very hard to spot the operator and, similar to the attribute expression, the error isn't the operator, it's the combination of left hand side, right hand side, and operator (you all sort of convinced me that the attribute expression ranges are correct, don't start arguing the other way ;))

BurntSushi: I like the rendering in this PR over the old rendering, even with the duplication.

But how bad does the overlapping ranges in one code frame look? That would be my instinct here.

AlexWaygood: I pushed some updates so that the diagnostic is now much more concise in the (common?) case where both operands have the same type, e.g.:

image

But if the two operands have different types, I've kept the separate code frame -- it looks much more readable to me. Here's what it looks like on this PR branch currently:

image

and here's what it looks like if I push it all into one codeframe:

image

This is the patch I applied to my PR branch to push it all into one codeframe:

<details>
<summary>Patch</summary>

```diff
diff --git a/crates/ty_python_semantic/src/types/diagnostic.rs b/crates/ty_python_semantic/src/types/diagnostic.rs
index 3493f62d17..cc98f709bb 100644
--- a/crates/ty_python_semantic/src/types/diagnostic.rs
+++ b/crates/ty_python_semantic/src/types/diagnostic.rs
@@ -4075,44 +4075,35 @@ pub(super) fn report_unsupported_comparison<'db>(
         [error.left_ty, error.right_ty, left_ty, right_ty],
     );
 
-    let mut diagnostic = if left_ty == right_ty {
-        let mut diag = diagnostic_builder
-            .into_diagnostic(format_args!("Unsupported `{}` operation", error.op));
-        diag.set_primary_message(format_args!(
+    let mut diagnostic =
+        diagnostic_builder.into_diagnostic(format_args!("Unsupported `{}` operation", error.op));
+
+    if left_ty == right_ty {
+        diagnostic.set_primary_message(format_args!(
             "Both operands have type `{}`",
             left_ty.display_with(db, display_settings.clone())
         ));
-        diag.annotate(context.secondary(left));
-        diag.annotate(context.secondary(right));
-        diag.set_concise_message(format_args!(
+        diagnostic.annotate(context.secondary(left));
+        diagnostic.annotate(context.secondary(right));
+        diagnostic.set_concise_message(format_args!(
             "Operator `{}` is not supported between two objects of type `{}`",
             error.op,
             left_ty.display_with(db, display_settings.clone())
         ));
-        diag
     } else {
-        let mut diag = diagnostic_builder.into_diagnostic("Unsupported comparison operation");
-
-        diag.set_primary_message(format_args!("`{}` is not supported here", error.op));
-
-        let mut sub = SubDiagnostic::new(SubDiagnosticSeverity::Info, "Operand types");
         for (ty, expr) in [(left_ty, left), (right_ty, right)] {
-            sub.annotate(context.secondary(expr).message(format_args!(
+            diagnostic.annotate(context.secondary(expr).message(format_args!(
                 "Has type `{}`",
                 ty.display_with(db, display_settings.clone())
             )));
         }
-        diag.sub(sub);
-
-        diag.set_concise_message(format_args!(
+        diagnostic.set_concise_message(format_args!(
             "Operator `{}` is not supported between objects of type `{}` and `{}`",
             error.op,
             left_ty.display_with(db, display_settings.clone()),
             right_ty.display_with(db, display_settings.clone())
         ));
-
-        diag
-    };
+    }
 
     // For non-atomic types like unions and tuples, we now provide context
     // on the underlying elements that caused the error.
```

</details>

BurntSushi: I'm fine to go with what you have here personally. I think it's better than the status quo.

I am not offended by the overlapping ranges in one code frame though.

AlexWaygood: Eh, I guess it's okay. It still feels a bit like a wall of text to me, but I do agree that concision is important on the CLI. I also find the tiny red `^^^` primary diagnostic in between the two yellow secondary diagnostics to look very strange, but hey.

AlexWaygood: (I'll switch to a single codeframe)

MichaReiser: I'm still not convinced by this change but I also don't think it should be blocking. I still think it makes the diagnostics unnecessarily verbose, especially in cases where the types are short, which is not uncommon. We then expand double the vertical size of the diagnostic without any real added benefit.

MichaReiser (approved):

crates/ty_python_semantic/resources/mdtest/snapshots/intersections.md_-_Comparison___Intersec…_-_Diagnostics_-_Unsupported_operator…_(27f95f68d1c826ec).snap
MichaReiser: I think I actually got this wrong and think we should keep it. Sorry. Did it show the elements that were incompatible if the type were a union or intersection? If so, I think that was useful. I misread it as that it always repeats the left and right operand types

AlexWaygood: no, you were right, it didn't show anything useful. It just said "comparison may raise an exception". I do still have an `info` subdiagnostic that tells you the incompatible elements if it's a union or tuple. (Intersections don't have the same issue, because they're much more forgiving than unions or tuples.)

AlexWaygood:
Some final screenshots now that I've addressed review comments. Much more concise than the initial version in this PR:

image

and the concise diagnostics for the same errors:

```
foo.py:1:1: error[unsupported-operator] Operator `<` is not supported between objects of type `Literal[1]` and `Literal["foo"]`
foo.py:2:1: error[unsupported-operator] Operator `<` is not supported between two objects of type `tuple[Literal[1], Literal[2], Literal[3], Literal[4], object]`
foo.py:3:1: error[unsupported-operator] Operator `<` is not supported between objects of type `tuple[Literal[1], Literal[2], Literal[3], Literal[4], Literal["foo"]]` and `tuple[Literal[1], Literal[2], Literal[3], Literal[4], Literal[5]]`
foo.py:7:5: error[unsupported-operator] Operator `<` is not supported between objects of type `int | str` and `int | list[str]`
```

PR #21738, opened 2025-12-01
Add token based `parenthesized_ranges` implementation

### LINKED ISSUE #21510: Optimized version of `parenthesized_ranges`
Rewrite (if possible) or provide alternatives to [`parenthesized_range`](https://github.com/astral-sh/ruff/blob/5ca9c15fc80070ff99f101af591a29cf1b1e11c2/crates/ruff_python_ast/src/parenthesize.rs#L60-L67) and [`parentheses_iterator`](https://github.com/astral-sh/ruff/blob/5ca9c15fc80070ff99f101af591a29cf1b1e11c2/crates/ruff_python_ast/src/parenthesize.rs#L14-L56) that take `&Tokens` as arguments instead of the `CommentRanges`, and don't use `SimpleTokenizer` or `BackwardsTokenizer` at all.

The ultimate goal of this would be to optimize calls sites such as [this one in ty](https://github.com/astral-sh/ruff/blob/5ca9c15fc80070ff99f101af591a29cf1b1e11c2/crates/ty_python_semantic/src/types/diagnostic.rs#L2146-L2150) (which uses `parentheses_iterator`).

Suggested by @MichaReiser here: https://github.com/astral-sh/ruff/pull/21476#discussion_r2537959967

denyszhak:
## Summary

Optimizes parenthesized_range by using token slice partitioning to reduce scanning overhead.

Closes #21510 

## Test Plan

1. Added unit tests for parenthesized_range and parentheses_iterator
2. Added coverage to ensure behavior matches the previous implementation

Acknowledging closed PR #21734 by @FarhanAliRaza for a couple of insights that helped shape the final version.


crates/ruff_python_parser/src/lib.rs
```diff
@@ -83,6 +83,7 @@ use ruff_text_size::{Ranged, TextRange, TextSize};
 
 mod error;
 pub mod lexer;
+pub mod parenthesize;
```
MichaReiser: I would add the method to `ruff_python_ast`. Ideally (not as part of this PR), we'd also move `Tokens` out of `ruff_python_parser` so that crates that only process the AST but don't care about how it was created don't have to depend on the parser crate. 

I suggest moving this module into `ruff_python_ast::tokens`

crates/ruff_python_parser/src/parenthesize.rs
```diff
@@ -0,0 +1,270 @@
...
+/// Mirrors the behavior of `SimpleTokenKind::is_trivia()` as closely as possible
+/// at the `TokenKind` level.
+const fn is_trivia(kind: TokenKind) -> bool {
+    matches!(
+        kind,
+        TokenKind::Comment
+            | TokenKind::Newline
+            | TokenKind::NonLogicalNewline
+            | TokenKind::Indent
+            | TokenKind::Dedent
```
MichaReiser:
```suggestion
            | TokenKind::NonLogicalNewline
```

I think we should remove `Dedent`, `Indent`, and `Newline` as all of those aren't trivia. Instead, they have semantic meaning in Python. Are there cases where you had to skip over those to make the two functions behave the same?

If they aren't needed for correctness, you can then use `token.kind().is_trivia()`

crates/ruff_python_parser/src/parenthesize.rs
```diff
@@ -0,0 +1,270 @@
...
+    } else {
+        tokens.last().map_or(expr.end(), Ranged::end)
+    };
+
+    let right_parens = tokens
+        .after(expr.end())
+        .iter()
+        .take_while(move |token| token.start() < exclusive_parent_end)
+        .filter(|token| !is_trivia(token.kind()))
+        .take_while(|token| token.kind() == TokenKind::Rpar);
```
MichaReiser: Nit: You can rewrite it like this to keep it closer to the original implementation:

```suggestion
    let after_tokens = if let Some(parent) = parent {
        // If the parent is a node that brings its own parentheses, exclude the closing parenthesis
        // from our search range. Otherwise, we risk matching on calls, like `func(x)`, for which
        // the open and close parentheses are part of the `Arguments` node.
        let exclusive_parent_end = if parent.is_arguments() {
            parent.end() - ")".text_len()
        } else {
            parent.end()
        };

        tokens.in_range(TextRange::new(expr.end(), exclusive_parent_end))
    } else {
        tokens.after(expr.end())
    };
    
    let right_parens = after_tokens
        .iter()
        .filter(|token| !token.kind().is_trivia())
        .take_while(move |token| token.kind() == TokenKind::Rpar);
```

crates/ruff_python_parser/src/parenthesize.rs
```diff
@@ -0,0 +1,270 @@
...
+/// parenthesized; or `None`, if the expression is not parenthesized.
+pub fn parenthesized_range(
+    expr: ExprRef,
+    parent: AnyNodeRef,
+    tokens: &Tokens,
+) -> Option<TextRange> {
+    parentheses_iterator(expr, Some(parent), tokens).last()
+}
+
+#[cfg(test)]
```
MichaReiser: Nice!

crates/ruff_python_ast_integration_tests/tests/parenthesize_optimized.rs
MichaReiser, line 1: I can see how those tests were useful during development to gain a better understanding of how the existing implementation works but I'm inclined to remove the tests now that you have a working implementation (or make them unit tests of `optimized_parenthesize`)

crates/ruff_python_parser/src/lib.rs
denyszhak: New method depends on Tokens, which currently lives in ruff_python_parser. If we move that method into ruff_python_ast without also moving Tokens, then ruff_python_ast would need to depend on ruff_python_parser, creating a circular dependency. Does that sound reasonable or am I interpreting it wrong?

crates/ruff_python_ast_integration_tests/tests/parenthesize_optimized.rs
denyszhak: I'm going to remove it, original unit tests provide broad coverage in general

crates/ruff_python_parser/src/lib.rs
MichaReiser: Oh hmm, that makes sense. It might be worth moving `Tokens` as a separate PR if you're up to it (I don't know how hard that would be)

denyszhak: I would love to attempt it in a separate PR and leave this one in ruff_python_parser if you are ok with that

crates/ruff_python_ast_integration_tests/tests/parenthesize_optimized.rs
denyszhak: done

crates/ruff_python_parser/src/parenthesize.rs
denyszhak: Part of additional kinds I referenced from SimpleTokenKind is_trivia, but part of it from the PR I referenced above thinking I have missed those. It doesn't influence correctness at least for the tests I added.

I changed to using `oken.kind().is_trivia()` as you suggested because now I think mimicking SimpleTokenKind.is_trivia was the wrong move

denyszhak:
@MichaReiser this one should be ready for you to review again

denyszhak:
> Thanks, this is great.
> 
> If you're interested, it would be great if we could rewrite the calls to `parenthesized_range` in `ruff_linter` to use your new implementation

I'm definitely interested, let me prepare a PR for this, I will create an issue and link it there

PR #21739, opened 2025-12-01
[ty] Fix TypeIs generic type narrowing with deferred materialization

suleymanozkeskin:
Fixes https://github.com/astral-sh/ty/issues/1703

Fixes incomplete type narrowing when using TypeIs with generic union types like: Result[T, E] = Ok[T] | Err[E].

Previously, type parameter E was materialized to 'object' before specialization could substitute it with concrete types like Exception, causing narrowing to produce Err[Unknown] instead of Err[Exception].

Changes:
  - Added is_materialized flag to TypeIsType to defer materialization
  - Implemented class-based union element matching for multi-typevar unions in generics.rs
  - Modified TypeIs creation to delay materialization until after specialization
  - Added test case for generic TypeIs narrowing to type_guards.md
  - Removed the comments which had a discussion about: https://github.com/astral-sh/ruff/pull/20591

## Summary
This pr should improve type narrowing in cases where generics are used.

## Test Plan

I did run the existing tests and added the respective case to type_guards.md to be tested, and it passed.

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
+++ b/crates/ty_python_semantic/resources/mdtest/narrow/type_guards.md
@@ -374,8 +374,8 @@ def narrowed_type_must_be_exact(a: object, b: Baz):
 from typing import Union, Generic, TypeVar
 from typing_extensions import TypeIs
 
-T = TypeVar('T', covariant=True)
-E = TypeVar('E', covariant=True)
+T = TypeVar("T", covariant=True)
+E = TypeVar("E", covariant=True)
 
 class Ok(Generic[T]):
     def __init__(self, value: T):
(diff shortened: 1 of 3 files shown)


carljm:
Thank you for the PR! I think this PR is not quite the right fix, due to an incomplete analysis of the problem.

Materialization won't transform a typevar to `object`; it will only transform a dynamic type (e.g. `Unknown` or `Any`). Materialization is not happening before specialization; it's happening after specialization, but specialization fails to infer a type for these type variables (due to limitations of our current constraint solver) and thus they are specialized to `Unknown`, after which materialization takes effect. The right fix is to improve our constraint solver to handle this case. In https://github.com/astral-sh/ty/issues/1703#issuecomment-3599797957 I showed an example that demonstrates the same issue occurring without `TypeIs`, so the fix should not require changes in `TypeIs`.

PR #21741, opened 2025-12-01
[ty] add tests for workspaces

Gankra:
Here are a bunch of (variously failing and passing) mdtests that reflect the kinds of issues people encounter when running ty over an entire workspace without sufficient hand-holding (especially because in the IDE it is unclear *how* to provide that hand-holding).

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/import/workspaces.md
+++ b/crates/ty_python_semantic/resources/mdtest/import/workspaces.md
@@ -195,7 +195,7 @@ x: int = 1
 
 `aproj/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -228,7 +228,7 @@ x: str = "2"
 
 `bproj/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -278,7 +278,7 @@ x: int = 1
 
 `a/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -316,7 +316,7 @@ x: str = "2"
 
 `b/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -369,7 +369,7 @@ x: int = 1
 
 `a/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -412,7 +412,7 @@ x: str = "2"
 
 `b/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -456,7 +456,7 @@ x: int = 1
 
 `a/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -495,7 +495,7 @@ x: str = "2"
 
 `b/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -543,7 +543,7 @@ x: int = 1
 
 `a/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -587,7 +587,7 @@ x: str = "2"
 
 `b/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -626,7 +626,7 @@ x: int = 1
 
 `a/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```
@@ -654,7 +654,7 @@ x: str = "2"
 
 `b/pyproject.toml`:
 
-```ignore
+```text
 name = "a"
 version = "0.1.0"
 ```


Gankra:
Note that some real-world examples are combinations of these. For instance "Tests Directory With Overlapping Names" actually works perfectly, but it fails in pyx because it's actually technically hitting "Current Directory Is Invalid Module Name" (because we try to resolve relative imports like `.setup` to `aproj.tests.setup`, which in pyx's case ends up being `a-proj.tests.setup`, which is rejected).

Some approaches to workspaces may dodge the invalid-module-name situation, some may not. I tried to get a good coverage of interesting cases that different solutions would show differences on.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+namespace packages or stub packages. These rules go out the window when analyzing some random local
+python file in some random workspace, and so we need to be more tolerant of situations that wouldn't
+fly in a published package, cases where we're not configured as well as we'd like, or cases where
+two projects in a monorepo have conflicting definitions (but we want to analyze both at once).
+
+## Invalid Names
+
+While you can't syntactically refer to a module with an invalid name (i.e. one with a `-`, or that
+has the same name as a keyword) there are plenty of situations where a module with an invalid name
+can be run. For instance `python my-script.py` and `python my-proj/main.py` both work, even though
```
MichaReiser, line 13: Yes, the way I think about it is that you can run a python script with an (mostly?) arbitrary name but whether you can import it depends on the segments relative to the search roots.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+we might in the course of analyzing the code compute the module name `my-script` or `my-proj.main`.
+
+Also, a sufficiently motivated programmer can technically use `importlib.import_module` which takes
+strings and does in fact allow syntactically invalid module names.
```
MichaReiser, line 17: okay, I didn't know that

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+y: int = 2
+```
+
+`mod3.py`:
+
+```py
+z: int = 2
+```
+
+## Multiple Projects
```
MichaReiser, line 142: Long term, I think we want a different solution to correctly support cases where the projects have different configurations (e.g. enabled / disabled some rules). But that shouldn't prevent us from doing better when there's no member-level configuration.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+### Tests Directory With Overlapping Names
+
+Here we have fairly typical situation where there are two projects `aproj` and `bproj` where the
+"real" packages are found under `src/` subdirectories, but each project also contains a `tests/`
+directory that contains python files which relatively import eachother and also absolutely import
+the package they test. All of these imports *should* resolve.
+
+```toml
+[environment]
+# This is similar to what we would compute for installed editables
```
MichaReiser, line 172: Those tests work because we also apply our default `environment.root` paths in mdtests?

Gankra: I believe so

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+`my-main.py`:
+
+```py
+# TODO: there should be no errors in this file
+
+# error: [unresolved-import]
+from .mod1 import x
+
+# error: [unresolved-import]
+from . import mod2
```
carljm, line 32: Both of these imports fail at runtime with `ImportError: attempted relative import with no known parent package` if you execute `python my-main.py`.

This is true (both at runtime and in ty today) even if we rename `my-main.py` to just `main.py` -- that is, it's unrelated to the use of an invalid module name. It's because you cannot relative-import top-level modules.

The only way these two imports can work is if all of these modules are imported as part of the same top-level parent package (that is, the directory containing these files is a Python package, and its parent directory is on the search path). (With the name `my-main.py`, the only way _that_ can happen is via `importlib.import_module`.)

In other words, as this test is currently written, I think the above TODO is wrong, and these imports should error. If we add an outer containing directory and place it on the search path, then I think these imports should work (and already would with a valid module name). With that change, I think the TODO to make them also work with an invalid module name would be correct.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+import mod3
```
carljm, line 33: The only scenario in which this import can work, and the two imports above also work, is overlapping search paths. The two relative imports can only work if this is all part of a containing package (as described above); this import can only work if these are all top-level modules. Overlapping search paths is the only way those two things can both be true.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+
+`mod3.py`:
+
+```py
+z: int = 2
+```
+
+## Multiple Projects
+
+It's common for a monorepo to define many separate projects that may or may not depend on eachother
```
carljm, line 144: Nit: "eachother" is not a word, it's two words. (I assumed this was a typo but then saw multiple other occurrences below. Not worth fixing alone but if we're changing this file anyway we may as well.)

```suggestion
It's common for a monorepo to define many separate projects that may or may not depend on each other
```

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+
+Here we have fairly typical situation where there are two projects `aproj` and `bproj` where the
+"real" packages are found under `src/` subdirectories, but each project also contains a `tests/`
+directory that contains python files which relatively import eachother and also absolutely import
+the package they test. All of these imports *should* resolve.
+
+```toml
+[environment]
+# This is similar to what we would compute for installed editables
+extra-paths = ["aproj/src/", "bproj/src/"]
```
carljm, line 173: The use of `extra-paths` here is significantly different from an installed editable, in that `extra-paths` are higher priority than `environment.roots`, while editable installs are lower priority. This seems like it could make a noticeable difference in the behavior of these tests, particularly those that test the interaction between same-named "outer" and "inner" packages.

It would be a bit more boilerplate, but I think these tests could better reflect a realistic scenario if they instead used `python =` config and then created `.pth` files in the fake site-packages that pointed to `aproj/src`, `bproj/src` -- that would actually get those paths added as editable installs. `import/site_packages_discovery.md` demonstrates this, minus the `.pth` file part.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+# This is similar to what we would compute for installed editables
+extra-paths = ["a/src/", "b/src/"]
+```
+
+`a/tests/test1.py`:
+
+```py
+# TODO: there should be no errors in this file.
+
+# error: [unresolved-import]
```
carljm, line 259: So I guess what happens here is we decide the package we should import is `a.tests.setup`, but then (because `./a/src/` is a higher priority search path than `.`) we go looking for `a/src/a/tests/setup.py` and don't find it.

This behavior seems correct (i.e. matches runtime)? Except that as discussed above, a more realistic test would not use `extra-paths`, which would make `.` higher priority than `./a/src/` and thus totally change the behavior of this test.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+```text
+name = "a"
+version = "0.1.0"
+```
+
+`b/src/b/__init__.py`:
+
+```py
+y: str = "20"
+```
```
carljm, line 328: This all seems totally redundant? It's testing the exact same thing as the `a` part above, in exactly the same way, and with no interaction between the two parts. Just delete it?

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
```
carljm, line 424: Same comment as above: the existence of `b` seems irrelevant to the test, we're just doing exactly the same test twice, once with the name `a` and once with the name `b`

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+# TODO: there should be no errors in this file.
+
+from .setup import x
+from . import setup
+
+# error: [unresolved-import]
+from main import y
+
+# error: [unresolved-import]
+import main
```
carljm, line 443: I don't understand why either of these imports should work. Nothing places `a/` on the search path, so `main` is not a top-level module, and these aren't relative imports.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+`b/src/b/__init__.py`:
+
+```py
+y: str = "20"
+```
+
+### Tests Package With Ambiguous Project Directories
+
+The same situation as the previous test but `tests/__init__.py` is also defined, in case that
+complicates the situation.
```
carljm, line 333: I don't think it can make a difference to this test. In theory testing various cases makes sense, but there are an infinite number of possible scenarios to test, and I'm not sure it scales to sort of randomly test little tweaks "in case", even if we haven't ever seen evidence (i.e. a failure in real code) that it makes a difference. So I'm kind of not convinced we should have both this test and the previous one, unless there's an observed behavior difference.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+
+`b/main.py`:
+
+```py
+y: str = "20"
+```
+
+### Tests Package Absolute Importing `main.py`
+
+The same as the previous case but `tests/__init__.py` exists in case that causes different issues.
```
carljm, line 511: Same comment as above.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
```diff
@@ -0,0 +1,660 @@
...
+# TODO: there should be no errors in this file.
+
+from .setup import x
+from . import setup
+
+# error: [unresolved-import]
+from main import y
+
+# error: [unresolved-import]
+import main
```
carljm, line 525: Same as above -- it seems entirely correct to me for these imports not to work.

my-script.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import annotations
```
carljm, line 1: I don't think we want this as a new top-level file in the ruff repo 😆

carljm (commented):
Thanks for adding these tests!

carljm:
I think another test we should add here is where we have overlapping search paths, where the outer path (say `.`) is a higher-priority search path than the inner path (say `./src`). If a file inside `./src` (say `./src/foo/bar.py`) does `from . import baz`, we currently reverse-resolve the module name of `./src/foo/bar.py` as `src.foo.bar` (rather than `foo.bar`) and then import `./src/foo/baz.py` as the module name `src.foo.baz`.

I think it would be better if we always preferred the inner, or more precise, matching search path instead. So `./src/foo/bar.py` would then reverse-resolve to the module name `foo.bar` (instead of `src.foo.bar`), even though `.` comes before `./src` in the search paths. And then our relative import would import `foo.baz` instead of `src.foo.baz`.

Cc @AlexWaygood since this is related to what we were discussing on https://github.com/astral-sh/ty/issues/1682

my-script.py
AlexWaygood: beat you to it ;) https://github.com/astral-sh/ruff/pull/21751

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
carljm: Ok, I guess we want these to work via https://github.com/astral-sh/ty/issues/1539 "desperation mode", where effectively we want to treat `a/` locally as an undeclared import root.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
carljm: Also here we want these to work via https://github.com/astral-sh/ty/issues/1539

carljm:
Created https://github.com/astral-sh/ty/issues/1730 to track the suggested new case separately.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: Great callout --  I've put everything but `mod3` under a `tests/` directory and renamed `my-main.py` to `my-mod.py`, and made a clarifying comment that in this case we're assuming `importlib` shenanigans.

(We in fact have had one user complain that they had a random `def.py` that they just `importlib.import_module` but we wouldn't resolve relative imports from it)

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: (Fix mentioned in previous comment also handles this)

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: ~~it's Canadian English~~

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: Also a great point, I misremembered the order of extra and first-party.

I think per discussion of "we should be auto-sorting nested search-paths" I probably want to have at least one test that uses editables and one that uses extra-paths (I imagine some users will find extra-paths, do this kind of thing manually, and it will behoove us to untangle it).

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: Having these duplicates is important, as discussed in the opening to the "Multiple Projects" section:

> Often the fact that there is both an `a` and `b` project seemingly won't matter, but many possible
solutions will misbehave under these conditions, as e.g. if both define a `main.py` and test code
has `import main`, we need to resolve each project's main as appropriate.

For instance without `b` also existing it's easy to convince yourself that, instead of doing desperate resolution, you could globally add every `pyproject.toml` as a search-path. However this would result in `a/main.py` and `b/main.py` both being the absolute module `main`.

I agree it's tedious boilerplate but I think it's really important for convincing ourselves the approach actually works in real monorepos with repetitive per-project structures.

crates/ty_python_semantic/resources/mdtest/import/workspaces.md
Gankra: This particular case was because I was perplexed by seeing this random variation in pyx's codebase and got a bit paranoid about whether it mattered if `tests` was a namespace package or a normal package. I think the `__init__.py` version is "strictly harder" for us to handle, so I should probably only have that one.

PR #21743, opened 2025-12-01
[ty] Stop testing the (brittle) constraint set display implementation

dcreager:
The `Display` implementation for constraint sets is brittle, and deserves a rethink. But later! It's perfectly fine for printf debugging; we just shouldn't be writing mdtests that depend on any particular rendering details. Most of these tests can be replaced with an equivalence check that actually validates that the _behavior_ of two constraint sets are identical.

### DIFF
--- a/crates/ty_python_semantic/src/types/constraints.rs
+++ b/crates/ty_python_semantic/src/types/constraints.rs
@@ -418,6 +418,7 @@ impl<'db> ConstraintSet<'db> {
         Self::constrain_typevar(db, typevar, lower, upper, TypeRelation::Assignability)
     }
 
+    #[expect(dead_code)] // Keep this around for debugging purposes
     pub(crate) fn display(self, db: &'db dyn Db) -> impl Display {
         self.node.simplify_for_display(db).display(db)
     }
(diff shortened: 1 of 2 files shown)


crates/ty_python_semantic/src/types/infer/builder.rs
```diff
@@ -10573,14 +10573,19 @@ impl<'db, 'ast> TypeInferenceBuilder<'db, 'ast> {
...
-                ))),
-                _ => None,
+            ) => {
+                let result = match op {
+                    ast::CmpOp::Eq => Some(
+                        left.constraints(self.db()).iff(self.db(), right.constraints(self.db()))
+                    ),
+                    ast::CmpOp::NotEq => Some(
+                        left.constraints(self.db()).iff(self.db(), right.constraints(self.db())).negate(self.db())
+                    ),
```
dcreager, line 10583: This is an important bit that lets us replace all of the `reveal_type` tests with `static_assert`s.  The `ConstraintSet.__eq__` implementation now checks whether two constraint sets are _equivalent_, not whether they are _identical_.

dcreager:
> ~I'll merge this PR to unblock the typeshed update.~ (no, that was the other PR)

The other PR was blocked on this one, though, so thank you regardless!

> We could follow the route of many other test frameworks and introduce something like a `static_assert_eq(left, right)` function that would show the display representation of the types of both arguments in case of a failure. That might be even nicer than `reveal_type` tests because we could properly align the left and right hand side types in the terminal.

That's a good idea, I opened https://github.com/astral-sh/ty/issues/1719 to track it.

PR #21744, opened 2025-12-01
[ty] Fix non-determinism in `ConstraintSet.specialize_constrained`

dcreager:
This fixes a non-determinism that we were seeing in the constraint set tests in https://github.com/astral-sh/ruff/pull/21715.

In this test, we create the following constraint set, and then try to create a specialization from it:

```
(T@constrained_by_gradual_list = list[Base])
  ∨
(Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
```

That is, `T` is either specifically `list[Base]`, or it's any `list`. Our current heuristics say that, absent other restrictions, we should specialize `T` to the more specific type (`list[Base]`).

In the correct test output, we end up creating a BDD that looks like this:

```
(T@constrained_by_gradual_list = list[Base])
┡━₁ always
└─₀ (Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
    ┡━₁ always
    └─₀ never
```

In the incorrect output, the BDD looks like this:

```
(Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
┡━₁ always
└─₀ never
```

The difference is the ordering of the two individual constraints. Both constraints appear in the first BDD, but the second BDD only contains `T is any list`. If we were to force the second BDD to contain both constraints, it would look like this:

```
(Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
┡━₁ always
└─₀ (T@constrained_by_gradual_list = list[Base])
    ┡━₁ always
    └─₀ never
```

This is the standard shape for an OR of two constraints. However! Those two constraints are not independent of each other! If `T` is specifically `list[Base]`, then it's definitely also "any `list`". From that, we can infer the contrapositive: that if `T` is not any list, then it cannot be `list[Base]` specifically. When we encounter impossible situations like that, we prune that path in the BDD, and treat it as `false`. That rewrites the second BDD to the following:

```
(Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
┡━₁ always
└─₀ (T@constrained_by_gradual_list = list[Base])
    ┡━₁ never   <-- IMPOSSIBLE, rewritten to never
    └─₀ never
```

We then would see that that BDD node is redundant, since both of its outgoing edges point at the `never` node. Our BDDs are _reduced_, which means we have to remove that redundant node, resulting in the BDD we saw above:

```
(Bottom[list[Any]] ≤ T@constrained_by_gradual_list ≤ Top[list[Any]])
┡━₁ always
└─₀ never       <-- redundant node removed
```

The end result is that we were "forgetting" about the `T = list[Base]` constraint, but only for some BDD variable orderings.

To fix this, I'm leaning in to the fact that our BDDs really do need to "remember" all of the constraints that they were created with. Some combinations might not be possible, but we now have the sequent map, which is quite good at detecting and pruning those.

So now our BDDs are _quasi-reduced_, which just means that redundant nodes are allowed. (At first I was worried that allowing redundant nodes would be an unsound "fix the glitch". But it turns out they're real! [This](https://ieeexplore.ieee.org/abstract/document/130209) is the paper that introduces them, though it's very difficult to read. Knuth mentions them in §7.1.4 of [TAOCP](https://course.khoury.northeastern.edu/csu690/ssl/bdd-knuth.pdf), and [this paper](https://par.nsf.gov/servlets/purl/10128966) has a nice short summary of them in §2.)

While we're here, I've added a bunch of `debug` and `trace` level log messages to the constraint set implementation. I was getting tired of having to add these by hands over and over. To enable them, just set `TY_LOG` in your environment, e.g.

```sh
env TY_LOG=ty_python_semantic::types::constraints::SequentMap=trace ty check ...
```

[Note, this has an `internal` label because are still not using `specialize_constrained` in anything user-facing yet.]

### DIFF
--- a/crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
+++ b/crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
@@ -41,6 +41,7 @@ def unbounded[T]():
     # revealed: None
     reveal_type(generic_context(unbounded).specialize_constrained(ConstraintSet.range(bool, T, bool) & ConstraintSet.range(Never, T, str)))
 
+    # TODO: revealed: ty_extensions.Specialization[T@unbounded = int]
     # revealed: ty_extensions.Specialization[T@unbounded = bool]
     reveal_type(generic_context(unbounded).specialize_constrained(ConstraintSet.range(Never, T, int) | ConstraintSet.range(Never, T, bool)))
     # revealed: ty_extensions.Specialization[T@unbounded = Never]


crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -41,7 +41,7 @@ def unbounded[T]():
     # revealed: None
     reveal_type(generic_context(unbounded).specialize_constrained(ConstraintSet.range(bool, T, bool) & ConstraintSet.range(Never, T, str)))
 
-    # revealed: ty_extensions.Specialization[T@unbounded = int]
+    # revealed: ty_extensions.Specialization[T@unbounded = bool]
```
dcreager, line 45: I think this test (and the other changed one below) were other possible points of nondeterminism, since the result could depend on whether `T ≤ bool` was kept in the BDD or simplified away as described in the PR body.

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
```diff
@@ -251,6 +251,30 @@ def constrained_by_gradual_list[T: (list[Base], list[Any])]():
     # revealed: ty_extensions.Specialization[T@constrained_by_gradual_list = list[Sub]]
     reveal_type(generic_context(constrained_by_gradual_list).specialize_constrained(ConstraintSet.range(list[Sub], T, list[Sub])))
 
+# Same tests as above, but with the typevar constraints in a different order, to make sure the
+# results do not depend on our BDD variable ordering.
+def constrained_by_gradual_list_reverse[T: (list[Any], list[Base])]():
```
dcreager, line 257: This test failed before these fixes, reproducing the test failure in the macOS CI job

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -780,9 +794,6 @@ impl<'db> Node<'db> {
                 root_constraint.ordering(db) > constraint.ordering(db)
             })
         );
-        if if_true == if_false {
-            return if_true;
-        }
```
dcreager, line 785: This is the clause that used to remove redundant BDD nodes. Removing it changes our BDDs from reduced to quasi-reduced.

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -934,8 +945,15 @@ impl<'db> Node<'db> {
...
-            (Node::AlwaysTrue, _) | (_, Node::AlwaysTrue) => Node::AlwaysTrue,
+            (Node::AlwaysTrue, Node::AlwaysTrue) => Node::AlwaysTrue,
             (Node::AlwaysFalse, other) | (other, Node::AlwaysFalse) => other,
+            (Node::AlwaysTrue, Node::Interior(interior))
+            | (Node::Interior(interior), Node::AlwaysTrue) => Node::new(
+                db,
+                interior.constraint(db),
+                Node::AlwaysTrue,
+                Node::AlwaysTrue,
+            ),
```
dcreager, line 956: This method skips calling `Node::new` for some inputs, so we have to change the base cases to not throw away redundant nodes. (ditto below)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -3015,7 +3169,7 @@ impl<'db> SatisfiedClauses<'db> {
         // used in test cases, so we don't need to over-optimize it.
 
         if self.clauses.is_empty() {
-            return String::from("always");
+            return String::from("never");
```
dcreager, line 3172: This was a bug in our display rendering that we couldn't trigger before, since `never` BDDs were always collapsed into the `AlwaysNever` terminal. Now we might have BDDs where all paths lead to the `never` terminal, which would hit this clause. (The intuition here is that we're displaying a DNF formula — a union of intersections — and so if the union contains no elements, the overall formula is `false`/`never`. And the new snippet just above handles where there's a single-element union of an empty intersection, which represents `true`/`always`.)

crates/ty_python_semantic/src/types/constraints.rs
```diff
@@ -3215,18 +3423,32 @@ mod tests {
     fn test_display_graph_output() {
         let expected = indoc! {r#"
             (T = str)
-            ┡━₁ (U = str)
-            │   ┡━₁ always
-            │   └─₀ (U = bool)
-            │       ┡━₁ always
-            │       └─₀ never
+            ┡━₁ (T = bool)
```
dcreager, line 3426: This looks like a big blow-up in the size of the BDD, but we still have structural sharing from the salsa interning. So for example, all of the copies of

```
(U = bool)
┡━₁ always
└─₀ never
```

are stored once as a single `InternalNode`. So the rendered size here is not a good proxy for how much memory is used internally.

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
carljm: I'm not sure I understand why we solve this to `bool`. The constraints are `T <= int | T <= bool`, which seems equivalent to `T <= int`. Why do we prefer solving to `bool` when there's the additional (seemingly redundant) constraint present?

crates/ty_python_semantic/src/types/constraints.rs
carljm: Primer doesn't report any noticeable memory increases, that's good enough for me!

crates/ty_python_semantic/resources/mdtest/generics/specialize_constrained.md
dcreager: This is because of our current heuristic for handling multiple paths in a BDD. `T ≤ int ∨ T ≤ bool` ends up looking like this now:

```
(T ≤ int)
┡━₁ (T ≤ bool)
│   ┡━₁ true
│   └─₀ true
└─₀ (T ≤ bool)
    ┡━₁ impossible
    └─₀ false
```

There are two paths to the `true` terminal, one representing `T ≤ bool` and one representing `bool < T ≤ int`. `T = bool` and `T = int` are the largest types that satisfy each respective path. Our current heuristic says that if there's a type that satisfies all paths, we choose that. (That is, if the intersection of the specializations is non-empty, use it.)

I'm very much open to changing that heuristic, but think that should be follow-on work. I'll mark this as a TODO that we'd rather produce `T = int` here.

